PR #1, opened 2017-11-21
DM-12371: Implement minimal butler Registry prototype with sqlalchemy core

pschella:


python/lsst/butler/datasets.py
```diff
@@ -0,0 +1,159 @@
+# 
+# LSST Data Management System
+#
+# Copyright 2008-2017  AURA/LSST.
```
timj: Surely not 2008?

python/lsst/butler/datasets.py
```diff
@@ -0,0 +1,159 @@
...
+# see <https://www.lsstcorp.org/LegalNotices/>.
+#
+
+from types import MappingProxyType
+from .utils import slotValuesAreEqual, slotValuesToHash
+from .storageClass import StorageClass
+from .units import DataUnitTypeSet
+
+
+class DatasetType:
```
timj: I know it's a prototype but the complete lack of doc strings makes it hard to work out what the attributes in the classes are meant to be.

python/lsst/butler/datasets.py
```diff
@@ -0,0 +1,159 @@
...
+        return self._name
+
+    @property
+    def units(self):
+        return self._units
+
+
+class DatasetRef(DatasetLabel):
+
+    __slots__ = ("_type", "_units", "_producer",
```
timj: I was under the impression that subclasses with `__slots__` couldn't reuse the same attributes. `_units` is already defined in the base class.

python/lsst/butler/datasets.py
```diff
@@ -0,0 +1,159 @@
...
+    def actualConsumers(self):
+        return MappingProxyType(self._actualConsumers)
+
+    def makePath(self, run, template=None):
+        raise NotImplementedError("TODO")
+
+
+class DatasetHandle(DatasetRef):
+
+    __slots__ = "_datasetId", "_registryId", "_uri", "_components", "_run"
```
timj: you are inconsistent with defining these. Sometimes you use parentheses and sometimes you don't.

python/lsst/butler/quantum.py
```diff
@@ -0,0 +1,85 @@
...
+
+from .datasets import DatasetRef
+from .run import Run
+
+
+class Quantum:
+
+    __slots__ = ("_quantumId", "_registryId", "_run", "_task", "_predictedInputs", "_actualInputs")
+
+    _currentId = -1
```
timj: This starts at -1 but the one above started at 0.

python/lsst/butler/quantum.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @classmethod
+    def getNewId(cls):
+        cls._currentId += 1
+        return cls._currentId
+
+    def __init__(self, run, task=None, quantumId=None, registryId=None):
+        assert isinstance(run, Run)
+        self._quantumId = quantumId
```
timj: This is using an ID but nothing is calling `getNewId`.

python/lsst/butler/quantum.py
```diff
@@ -0,0 +1,85 @@
...
+        return self._predictedInputs
+
+    @property
+    def actualInputs(self):
+        return self._actualInputs
+
+    def addPredictedInput(self, ref):
+        assert isinstance(ref, DatasetRef)
+
+        if ref.type.name not in self._actualInputs:
```
timj: Without any comments or documentation in the classes it's a bit hard to work out why the check is for `ref.type.name` but the key used is then `ref.name`.

python/lsst/butler/registry.py
```diff
@@ -0,0 +1,470 @@
...
+from .datasets import DatasetType, DatasetHandle, DatasetRef, DatasetLabel
+from .run import Run
+from .quantum import Quantum
+from .units import DataUnit, DataUnitTypeSet
+from .storageClass import StorageClass
+
+
+class Registry:
+    """Basic SQL backed registry.
+    """
```
timj: Attribute documentation is missing.

python/lsst/butler/registry.py
```diff
@@ -0,0 +1,470 @@
...
+
+        The :ref:`Quantum` that generated the :ref:`Dataset` can optionally be provided to add provenance information.
+
+        :param ref: a :ref:`DatasetRef` that identifies the :ref:`Dataset` and contains its :ref:`DatasetType`.
+
+        :param str uri: the :ref:`URI` that has been associated with the :ref:`Dataset` by a :ref:`Datastore`.
+
+        :param dict components: if the :ref:`Dataset` is a composite, a ``{name : URI}`` dictionary of its named components and storage locations.
+
+        :param Run run: the :ref:`Run` instance that produced the Dataset.  Ignored if ``producer`` is passed (:py:attr:`producer.run <Quantum.run>` is then used instead).  A Run must be provided by one of the two arguments.
```
timj: docstring lines are too long. stick to 79 characters.

python/lsst/butler/units.py
```diff
@@ -0,0 +1,992 @@
...
+        raise NotImplementedError("pure virtual")
+
+    def pkey(self):
+        return NotImplemented("pure virtual")
+
+    def invariantHash(self):
+        """
+        Compute a hash that is invariant accross Python sessions, and hence can be stored in a database.
+        """
+        return sha512(b''.join(str(v).encode() for v in self.pkey)).digest()
```
timj: If you want to be invariant I suggest you specify the encoding rather than relying on the default.

tests/SConscript
```diff
@@ -0,0 +1,3 @@
+# -*- python -*-
+from lsst.sconsUtils import scripts
+scripts.BasicSConscript.tests()
```
timj: Add `pyList=[]` as argument for `tests()`. Also add a `setup.cfg` to enable flake8 testing.

timj (commented):
I've had a quick scan of all of it. I haven't really grokked it all yet but it looks reasonable. In particular I've never really used sqlalchemy before so that's a bit hard to follow.

I think I'd like LSST concepts and CCD sensor concepts in their own files to make it clear to external adopters of butler which files contain LSST oddness and which are generic.

python/lsst/butler/datasets.py
pschella: It can, but it wastes memory. So I have removed it.

python/lsst/butler/datasets.py
pschella: That is probably because I wrote some of it, and Jim some other parts ;) I'm now using parentheses everywhere.

python/lsst/butler/quantum.py
pschella: Both are pre-increment, so should start at `-1`. Note that the whole ID thing is ugly, flaky and needs to be fixed. The only reason for it existing is because `sqlite3` doesn't allow `autoincrement` compound-primary-keys. I should probably create a ticket for this...

pschella: DM-13063 created.

python/lsst/butler/quantum.py
pschella: That is intentional (if ugly). Will be fixed on DM-13063.

python/lsst/butler/quantum.py
pschella: It should have been `ref.type.name`. Updated and added docstrings for `Quantum`.

python/lsst/butler/registry.py
pschella: Also converted to numpydoc.

PR #2, opened 2017-12-01
DM-12667: Implement minimal Butler POSIX Datastore prototype

pschella:


python/lsst/butler/datastore/datastore.py
```diff
@@ -0,0 +1,167 @@
...
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <https://www.lsstcorp.org/LegalNotices/>.
+#
+
+import yaml
+from abc import ABCMeta, abstractmethod, abstractproperty
+
+
+class DatastoreConfig(metaclass=ABCMeta):
```
SimonKrughoff, line 28: The obvious question is why you are not extending `pex_config` instead of inventing a new thing.

python/lsst/butler/datastore/datastore.py
```diff
@@ -0,0 +1,167 @@
...
+        with open(filename, 'w') as f:
+            self.dump(f)
+
+
+class Datastore(metaclass=ABCMeta):
+    """Datastore interface.
+    """
+
+    @abstractmethod
+    def get(self, uri, storageClass, parameters=None):
```
SimonKrughoff, line 82: Do we need vectorized versions of these?  I can imagine wanting to e.g. get the PSFs for all chips in a focalplane, but I'm not sure theres a benefit to vectorizing this method as opposed to just looping over the `get` in the next layer up.

python/lsst/butler/datastore/datastore.py
```diff
@@ -0,0 +1,167 @@
...
+        """Load an `InMemoryDataset` from the store.
+
+        Parameters
+        ----------
+        uri : `str`
+            a Universal Resource Identifier that specifies the location of the stored `Dataset`.
+        storageClass : `StorageClass`
+            the `StorageClass` associated with the `DatasetType`.
+        parameters : `dict`
```
SimonKrughoff, line 91: I'm a little unsure how the parameters are meant to be used.  It seems like it may be easier to see what's going on if the parameters are just parameters to another constructor, either the `StorageClass` or `FileDescriptor` or `Datastore`?

I think I'm just having trouble understanding what the parameters are supposed to be parameterizing and which class is expected to interpret them.

python/lsst/butler/datastore/datastore.py
```diff
@@ -0,0 +1,167 @@
...
+
+        Returns
+        -------
+        inMemoryDataset : `InMemoryDataset`
+            Requested `Dataset` or slice thereof as an `InMemoryDataset`.
+        """
+        raise NotImplementedError("Must be implemented by subclass")
+
+    @abstractmethod
+    def put(self, inMemoryDataset, storageClass, path, typeName=None):
```
SimonKrughoff, line 102: Should `path` be called `hint` instead?  It sort of implies a traditional filesystem if the word `path` is used.

python/lsst/butler/datastore/datastore.py
```diff
@@ -0,0 +1,167 @@
...
+        """Write a `InMemoryDataset` with a given `StorageClass` to the store.
+
+        Parameters
+        ----------
+        inMemoryDataset : `InMemoryDataset`
+            The `Dataset` to store.
+        storageClass : `StorageClass`
+            The `StorageClass` associated with the `DatasetType`.
+        path : `str`
+            A `Path` that provides a hint that the `Datastore` may use as (part of) the URI.
```
SimonKrughoff, line 112: This is a path specified as a string, right?  If so, I don't think it should be referred to as a class: i.e. path not `Path`.  May be a non-issue if we are going away from the word path in the call signature.

python/lsst/butler/datastore/datastore.py
```diff
@@ -0,0 +1,167 @@
...
+        uri : `str` 
+            The `URI` where the primary `Dataset` is stored.
+        components : `dict`, optional
+            A dictionary of URIs for the `Dataset`' components.
+            The latter will be empty if the `Dataset` is not a composite.
+        """
+        raise NotImplementedError("Must be implemented by subclass")
+
+    @abstractmethod
+    def remove(self, uri):
```
SimonKrughoff, line 128: Does this really have no return value?  Are you expecting to raise an `Exception` if this fails?

python/lsst/butler/datastore/datastore.py
```diff
@@ -0,0 +1,167 @@
...
+
+        Parameters
+        ----------
+        inputDatastore : `Datastore`
+            The external `Datastore` from which to retreive the `Dataset`.
+        inputUri : `str`
+            The `URI` of the `Dataset` in the input `Datastore`.
+        storageClass : `StorageClass`
+            The `StorageClass` associated with the `DatasetType`.
+        path : `str`
```
SimonKrughoff, line 159: Same worries about "path" as above.

python/lsst/butler/datastore/posix/datastore.py
```diff
@@ -0,0 +1,233 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <https://www.lsstcorp.org/LegalNotices/>.
+#
+
+import os
+import urllib
```
SimonKrughoff: Unused import `urllib`.

python/lsst/butler/datastore/posix/datastore.py
```diff
@@ -0,0 +1,233 @@
...
+from abc import ABCMeta, abstractmethod
```
SimonKrughoff: I believe these are unused as well.

python/lsst/butler/datastore/posix/datastore.py
```diff
@@ -0,0 +1,233 @@
...
+        filename : `str`
+            The file to dump to.
+        """
+        super().dumpToFile(filename)
+
+    @property
+    def formatters(self):
+        return self._data.formatters
+
+    def _loadDefaultConfiguration(self):
```
SimonKrughoff, line 99: I'm worried about formatters coming from code.  Doesn't that make it hard to extend the dataset definitions without modifying code?

python/lsst/butler/datastore/posix/datastore.py
```diff
@@ -0,0 +1,233 @@
...
+            if not create:
+                raise ValueError("No valid root at: {0}".format(root))
+            safeMakeDir(root)
+        self.root = root
+
+        self.locationFactory = LocationFactory(self.root)
+
+        self.config = PosixDatastoreConfig()
+        if configFile:
+            self.config.loadFromFile(configFile)
```
SimonKrughoff, line 132: I guess this answers my question above, to some extent.  Is the point that the yaml file could override the formatters in the default?

python/lsst/butler/datastore/posix/formatter.py
```diff
@@ -0,0 +1,133 @@
...
+        ----------
+        storageClass : `StorageClass`
+            Get `Formatter` associated with this `StorageClass`, unless.
+        datasetType : `DatasetType` or `str, optional
+            If given, look if an override has been specified for this `DatasetType` and,
+            if so return that instead.
+        """
+        if datasetType:
+            try:
+                typeName = self._registry[self._getName(datasetTypeName)]()
```
SimonKrughoff: `datasetTypeName` is undefined.

python/lsst/butler/datastore/posix/formatter.py
```diff
@@ -0,0 +1,133 @@
...
+        elif isinstance(typeOrName, DatasetType):
+            return typeOrName.name
+        elif issubclass(typeOrName, StorageClass):
+            return typeOrName.name
+        else:
+            raise ValueError("Cannot extract name from type")
+
+    @staticmethod
+    def _getInstanceOf(typeName):
+        cls = doImport(typeName)
```
SimonKrughoff, line 135: I didn't look in `daf_persistence` but I'm a little worried about `doImport`.  It seems like it needs to be doing an `exec` somewhere in there.  It may not be avoidable since we clearly can't put classes in the yaml config.

python/lsst/butler/datastore/posix/formatter.py
```diff
@@ -0,0 +1,133 @@
...
+        Parameters
+        ----------
+        type : `str` or `StorageClass` or `DatasetType`
+            Type for which this formatter is to be used.
+        formatter : `Formatter` subclass (not an instance)
+            The `Formatter` to use for reading and writing `Dataset`s of this type.
+        """
+        #assert issubclass(formatter, Formatter)
+        assert isinstance(formatter, str)
+        self._registry[self._getName(type)] = formatter
```
SimonKrughoff: I prefer not to shadow builtins.  I know it's not particularly dangerous here, but it's easy to simply pick a different parameter name than `type`.

python/lsst/butler/datastore/posix/formatter.py
```diff
@@ -0,0 +1,133 @@
...
+
+        Parameters
+        ----------
+        type : `str` or `StorageClass` or `DatasetType`
+            Type for which this formatter is to be used.
+        formatter : `Formatter` subclass (not an instance)
+            The `Formatter` to use for reading and writing `Dataset`s of this type.
+        """
+        #assert issubclass(formatter, Formatter)
+        assert isinstance(formatter, str)
```
SimonKrughoff: I think comparison and then raising a `ValueError` would be good here since it would give you a chance to give a more informative error message rather than just "This object isn't of type `str`".

python/lsst/butler/datastore/posix/formatter.py
```diff
@@ -0,0 +1,133 @@
...
+        return self._getInstanceOf(typeName)
+
+    def registerFormatter(self, type, formatter):
+        """Register a Formatter.
+
+        Parameters
+        ----------
+        type : `str` or `StorageClass` or `DatasetType`
+            Type for which this formatter is to be used.
+        formatter : `Formatter` subclass (not an instance)
```
SimonKrughoff: Formatter is not a subclass, it is a string indicating the class to be instantiated.

python/lsst/butler/datastore/posix/formatter.py
```diff
@@ -0,0 +1,133 @@
...
+    @staticmethod
+    def _getName(typeOrName):
+        """Extract name of `DatasetType` or `StorageClass` as string.
+        """
+        if isinstance(typeOrName, str):
+            return typeOrName
+        elif isinstance(typeOrName, DatasetType):
+            return typeOrName.name
+        elif issubclass(typeOrName, StorageClass):
+            return typeOrName.name
```
SimonKrughoff: Maybe just `elif has_attr(typeOrName, 'name'):  return typeOrName.name` to replace the last two tests.

python/lsst/butler/datastore/posix/location.py
```diff
@@ -0,0 +1,99 @@
...
+    """Factory for `Location` instances.
+    """
+
+    def __init__(self, datastoreRoot):
+        """Constructor
+
+        Parameters
+        ----------
+        datastoreRoot : `str`
+            Root directory of the `Datastore` in the filesystem.
```
SimonKrughoff: Again, I'm a little hesitant to use the word directory here.  Maybe "Root location"?

python/lsst/butler/datastore/posix/location.py
```diff
@@ -0,0 +1,99 @@
...
+        uri : `str`
+            A valid Universal Resource Identifier.
+
+        Returns
+        location : `Location`
+            The equivalent `Location`.
+        """
+        return Location(self._datastoreRoot, uri)
+
+    def fromPath(self, path):
```
SimonKrughoff, line 86: Does this belong in a `PosixLocationFactory` subclass?

python/lsst/butler/datastore/posix/datastore.py
```diff
@@ -0,0 +1,233 @@
...
+    def put(self, inMemoryDataset, storageClass, path, typeName=None):
+        """Write a `InMemoryDataset` with a given `StorageClass` to the store.
+
+        Parameters
+        ----------
+        inMemoryDataset : `InMemoryDataset`
+            The `Dataset` to store.
+        storageClass : `StorageClass`
+            The `StorageClass` associated with the `DatasetType`.
+        path : `str`
```
SimonKrughoff, line 170: As I stated above, I think `path` may be a bit overloaded and something like `hint` might be less risky.

python/lsst/butler/datastore/posix/datastore.py
```diff
@@ -0,0 +1,233 @@
...
+        inMemoryDataset : `InMemoryDataset`
+            Requested `Dataset` or slice thereof as an `InMemoryDataset`.
+        """
+        formatter = self.formatterFactory.getFormatter(storageClass)
+        location = self.locationFactory.fromUri(uri)
+        if not os.path.exists(location.path):
+            raise ValueError("No such file: {0}".format(location.uri))
+        return formatter.read(FileDescriptor(location, storageClass.type, parameters))
+
+    def put(self, inMemoryDataset, storageClass, path, typeName=None):
```
SimonKrughoff, line 161: I'm wondering if you need to allow parameters to be passed here as well.  It certainly wouldn't hurt to default it to `None`, just in case we have a data storage that needs to know something for both reads and writes.

I wonder actually if it wouldn't be good for all these access commands to just forward `kwargs` to lower level calls where they could be used or not.

python/lsst/butler/datastore/posix/datastore.py
```diff
@@ -0,0 +1,233 @@
...
+
+        Parameters
+        ----------
+        uri : `str`
+            a Universal Resource Identifier that specifies the location of the stored `Dataset`.
+
+        .. note::
+            Some Datastores may implement this method as a silent no-op to disable `Dataset` deletion through standard interfaces.
+        """
+        location = self.locationFactory.fromUri(uri)
```
SimonKrughoff, line 199: Isn't there a little more that needs to be done here if the `uri` describes a composite data set?

python/lsst/butler/datastore/posix/datastore.py
```diff
@@ -0,0 +1,233 @@
...
+
+        Parameters
+        ----------
+        inputDatastore : `Datastore`
+            The external `Datastore` from which to retreive the `Dataset`.
+        inputUri : `str`
+            The `URI` of the `Dataset` in the input `Datastore`.
+        storageClass : `StorageClass`
+            The `StorageClass` associated with the `DatasetType`.
+        path : `str`
```
SimonKrughoff, line 216: Same comment as above on `path` vs. `hint`.

tests/test_posixDatastore.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class PosixDatastoreTestCase(lsst.utils.tests.TestCase):
+
+    def setUp(self):
+        testDir = os.path.dirname(__file__)
+        catalogPath = os.path.join(testDir, "data", "basic", "source_catalog.fits")
+        self.catalog = lsst.afw.table.SourceCatalog.readFits(catalogPath)
+
+    def _assertCatalogEqual(self, inputCatalog, outputCatalog):
```
SimonKrughoff, line 41: `inputCatalog` is not used.  I guess you are planning on putting in a more comprehensive equality check in the future?

SimonKrughoff (approved):
Looks great to me.

My one general comment is that there are a lot of `assert`s that check types in the function definitions.  I think it might be better to test and raise or just let ducktyping do its thing.  If this is just a prototyping thing, ignore me.

python/lsst/butler/datastore/datastore.py
ktlim: We should not use `pex_config` for persisted in-datastore configuration.  We wouldn't want use of a dataset to require executing arbitrary code.

python/lsst/butler/datastore/posix/datastore.py
ktlim: Formatters must be checked in and versioned somewhere.  As above, arbitrary code should not be stored in the configuration.  (Plus, formatters may often involve compilation for performance.)

python/lsst/butler/datastore/posix/formatter.py
ktlim: `doImport` uses Python's import mechanism; it doesn't do an `exec` or `eval`.  It is limited to code that is on `PYTHONPATH`.

python/lsst/butler/datastore/datastore.py
SimonKrughoff: Good enough.

python/lsst/butler/datastore/posix/datastore.py
SimonKrughoff: I just meant that I don't want users to have to modify datastore code to add storage classes.  I think this is ok as it seems like Pim changed from classes in the registry to strings for just this reason.

python/lsst/butler/datastore/posix/formatter.py
SimonKrughoff: OK I'm glad to hear that.  Sorry for not looking into it myself.

python/lsst/butler/datastore/datastore.py
pschella: Yes, for some `Datastore` implementations that might be beneficial. But for this simple one probably not. We should revisit this later.

python/lsst/butler/datastore/datastore.py
pschella: Changed to raise a `FileNotFoundError`.

pschella: And documented in ABC.

python/lsst/butler/datastore/posix/datastore.py
pschella: Changed to load from config file in DM-12920.

python/lsst/butler/datastore/posix/datastore.py
pschella: Yes, but changed in DM-12920. All read from (probably versioned) config file.

python/lsst/butler/datastore/posix/datastore.py
pschella: Possibly. Although currently the API for `get` and `put` was assumed to be fixed by DMTN-056 with only lower layers to be prototyped. We should likely add parameters/flags as options, although it is a philosophical question if these should be fixed in configuration of a `Datastore` or supplied on `put/get` (e.g. compression settings, file formats, etc.). This can create problems for mismatched `get` and `put` flags.

python/lsst/butler/datastore/posix/datastore.py
pschella: There might be, although in the design we have used so far the fact that `Datastore`s don't know about composites (unless they choose to store one as a single file, in which case matters get tricky). Implementing of composites was outside the scope of this current ticket. Pushed to DM-13033

python/lsst/butler/datastore/posix/formatter.py
pschella: Granted, that is probably more idiomatic Python (stuck in C++ land here).

python/lsst/butler/datastore/posix/formatter.py
pschella: I'm not a particularly big fan of this construct either, but it seems relatively safe and the best we can do here (unless we serialize the actual objects into YAML).

tests/test_posixDatastore.py
pschella: Yes, that was the plan. I'll see if I can find the proper comparisons from where I got the test data.

python/lsst/butler/datastore/datastore.py
pschella: Yes, will find a solution for this on DM-12921.

timj:
This was merged in 85f608c

PR #3, opened 2017-12-04
DM-12668: Implement a minimal Butler prototype

pschella:


python/lsst/butler/butler.py
```diff
@@ -0,0 +1,197 @@
...
+            The run used for all outputs.
+            May be ``None`` to construct a read-only Butler.
+            The `Run`s `CollectionTag` is always used as the input collection when a `Run` is provided.
+        templates : `dict`
+            Maps `DatasetType` names to path templates, used to override `DatasetType.template` as obtained from the `Registry` when present.
+        """
+        assert collection or run
+        self.run = run
+        if run:
+            self.collection = run.tag
```
hsinfang, line 61: Is it intentional that if both `collection` and `run` are supplied, `collection` is ignored without any check?

tests/test_posixDatastore.py
```diff
@@ -30,10 +30,13 @@
 from lsst.butler.datastore import PosixDatastore
 from lsst.butler.storageClass import SourceCatalog
 
+import datasetsHelper
+
 
 class PosixDatastoreTestCase(lsst.utils.tests.TestCase):
 
     def setUp(self):
+<<<<<<< HEAD
```
hsinfang: To be cleaned up here

hsinfang (commented):
Why do we need both `DatasetLabel` and `DatasetRef`, while either can uniquely identify a dataset given a Collection? Why do some butler methods take DatasetLabel and some require DatasetRef? 

I see the `registry.expand()` method; is the only difference a datasetType object versus a datasetType name?

python/lsst/butler/butler.py
```diff
@@ -0,0 +1,197 @@
...
+        ----------
+        quantum : `Quantum`
+            The dependent `Quantum`.
+        ref : `DatasetRef`
+            The `Dataset` that is a true dependency of ``quantum``.
+        """
+        handle = self.registry.find(self.config.collection, ref)
+        self.registry.markInputUsed(handle, quantum)
+
+    def unlink(self, *labels):
```
hsinfang, line 185: What does `unlink` mean for a read-only Butler?

python/lsst/butler/butler.py
pschella: It was, but is probably safer to raise an exception.

python/lsst/butler/butler.py
pschella: Nothing. Should not be implemented in this case, or be a noop (yet to decide).

timj:
Merged in 3222b3d

PR #4, opened 2017-12-14
DM-12920: Prototype Butler configuration

pschella:


python/lsst/butler/config.py
```diff
@@ -0,0 +1,320 @@
...
+
+class _ConfigBase(collections.UserDict, yaml.YAMLObject, metaclass=_ConfigMeta):
+    pass
+
+
+class Config(_ConfigBase):
+    """Config implements a datatype that is used by Butler for configuration parameters.
+    It is essentially a dict with key/value pairs, including nested dicts (as values). In fact, it can be
+    initialized with a dict. The only caveat is that keys may NOT contain dots ('.'). This is explained next:
+    Config extends the dict api so that hierarchical values may be accessed with dot-delimited notiation.
```
timj: "notiation" -> "notation"

python/lsst/butler/config.py
```diff
@@ -0,0 +1,320 @@
...
+    That is, foo.getValue('a.b.c') is the same as foo['a']['b']['c'] is the same as foo['a.b.c'], and either
+    of these syntaxes may be used.
+
+    Storage formats supported:
+    - yaml: read and write is supported.
+    """
+
+    def __init__(self, other=None):
+        """Initialize the Config. Other can be used to initialize the Config in a variety of ways:
+        other (string) Treated as a path to a config file on disk. Must end with '.paf' or '.yaml'.
```
timj: are you really supporting paf?

python/lsst/butler/config.py
```diff
@@ -0,0 +1,320 @@
...
+
+        if isinstance(other, collections.Mapping):
+            self.update(other)
+        elif isinstance(other, Config):
+            self.data = copy.deepcopy(other.data)
+        elif isinstance(other, str):
+            # if other is a string, assume it is a file path.
+            self.__initFromFile(other)
+        else:
+            # if the config specified by other could not be loaded raise a runtime error.
```
timj: This comment is not strictly correct. This error is raised if other was not recognized, not if it could not be loaded.

python/lsst/butler/config.py
```diff
@@ -0,0 +1,320 @@
...
+            raise RuntimeError("A Config could not be loaded from other:%s" % other)
+
+    def ppprint(self):
+        """helper function for debugging, prints a config out in a readable way in the debugger.
+
+        use: pdb> print myConfigObject.pprint()
+        :return: a prettyprint formatted string representing the config
+        """
+        import pprint
```
timj: this is a standard module so are we putting it here because it will slow down imports and is rarely used?

python/lsst/butler/config.py
```diff
@@ -0,0 +1,320 @@
...
+        return pprint.pformat(self.data, indent=2, width=1)
+
+    def __repr__(self):
+        return self.data.__repr__()
+
+    def __initFromFile(self, path):
+        """Load a file from path. If path is a list, will pick one to use, according to order specified
+        by extensionPreference.
```
timj: Lists are not supported and extensionPreference is not an argument.

python/lsst/butler/config.py
```diff
@@ -0,0 +1,320 @@
...
+
+        :param path: string or list of strings, to a persisted config file.
```
timj: We have 3 different ways of documenting arguments to methods in this file. We need to pick the numpy version.

python/lsst/butler/config.py
```diff
@@ -0,0 +1,320 @@
...
+            self.__initFromYaml(f)
+
+    def __initFromYaml(self, stream):
+        """Loads a YAML config from any readable stream that contains one.
+
+        :param stream:
+        :return:
+        """
+        # will raise yaml.YAMLError if there is an error loading the file.
+        self.data = yaml.load(stream)
```
timj: Are the files too complex to use `safe_load()`?

python/lsst/butler/config.py
```diff
@@ -0,0 +1,320 @@
...
+        if isinstance(data, collections.Mapping):
+            data = Config(data)
+        return data
+
+    def __setitem__(self, name, value):
+        if isinstance(value, collections.Mapping):
+            keys = name.split('.')
+            d = {}
+            cur = d
+            for key in keys[0:-1]:
```
timj: might be clearer if you popped the last item off the `keys` list so that the -1 indexing can be removed. I don't think you need the `0` in the slice.

python/lsst/butler/config.py
```diff
@@ -0,0 +1,320 @@
...
+                cur[key] = {}
+                cur = cur[key]
+            cur[keys[-1]] = value
+            self.update(d)
+        data = self.data
+        keys = name.split('.')
```
timj: can you split the name once on entry to the method rather than doing it twice?

python/lsst/butler/config.py
```diff
@@ -0,0 +1,320 @@
...
+        for key in keys[0:-1]:
+            data = data.setdefault(key, {})
+        data[keys[-1]] = value
+
+    def __contains__(self, key):
+        d = self.data
+        keys = key.split('.')
+        for k in keys[0:-1]:
```
timj: Same indexing comment here.

python/lsst/butler/config.py
```diff
@@ -0,0 +1,320 @@
...
+
+        @param productName (string) The name of the product that the default config is installed as part of
+        @param fileName (string) The name of the config file. Can also include a path to the file relative to
+                                 the directory where the product is installed.
+        @param relativePath (string) The relative path from the directior where the product is installed to
+                                     the location where the file (or the path to the file) is found. If None
+                                     (default), the fileName argument is relative to the installation
+                                     directory.
+        """
+        basePath = lsst.utils.getPackageDir(productName)
```
timj: `getPackageDir` throws `lsst.pex.exceptions.NotFoundError` if the product can not be found so the trap on the next line is irrelevant.

python/lsst/butler/config.py
```diff
@@ -0,0 +1,320 @@
...
+        """Get the path to a default config file.
+
+        Determines a directory for the product specified by productName. Then Concatenates
+        productDir/relativePath/fileName (or productDir/fileName if relativePath is None) to find the path
+        to the default Config file
+
+        @param productName (string) The name of the product that the default config is installed as part of
+        @param fileName (string) The name of the config file. Can also include a path to the file relative to
+                                 the directory where the product is installed.
+        @param relativePath (string) The relative path from the directior where the product is installed to
```
timj: "directior"

python/lsst/butler/registry.py
```diff
@@ -30,18 +30,28 @@
...
+    def __init__(self, config):
+        self.config = RegistryConfig(config)['registry']
+        self.engine = create_engine(self.config['dbname'])
+        self.id = self.config['id']
 
         metadata.create_all(self.engine)
 
+    @staticmethod
+    def fromConfig(config):
+        from lsst.daf.persistence import doImport
```
timj: I'm not really a fan of these deferred imports as it hides the dependencies.

timj (approved):
I have quite a lot of comments on `config.py` but is that because the code came from daf_persistence? It's very inconsistent. I would have preferred some way to indicate in the commit history that this file was written by someone else before you modified it.

pschella:
Yes, that code is verbatim (minus a couple of deletions) from `daf_persistence`. I don't know how to preserve cross-package commit history. I made it a separate commit, and could add some more comments but am open to suggestions.

timj:
What I am asking for is that you don't own the commit that just copied the file across. Use the `--author` option to `git commit`. If Nate wrote it all make him the author. If it has many authors use a generic author name. This will make it more obvious where the code came from and indicate that you didn't write it. The commits to fix it up for this usage would then be yours.

TallJimbo:
This policy has been ignored basically since its inception, but it does provide a workflow that addresses this concern:

https://developer.lsst.io/processes/transferring_code.html

pschella:
Ok, I followed the procedure outlined there (except for the steps at the origin since this is a copy). Hope it is clear now.

python/lsst/butler/config.py
pschella: No, removed.

python/lsst/butler/config.py
pschella: Don't know (since I didn't write this). But moving all imports up.

python/lsst/butler/config.py
pschella: No idea. Probably. Should check.

pschella: Apparently not yet. So making it `safe_load` for now.

python/lsst/butler/config.py
pschella: Actually just removed the whole function for now, pending discussion about default configuration locations.

python/lsst/butler/registry.py
pschella: So I moved the import of `doImport` to the top. Of course it is still doing a deferred import, since that is the whole point of `doImport`. But I assume that is not what you were objecting to.

timj: Yes, it was the `from` I wanted moved up top.

timj:
Merged in commit 32c6a8e

PR #6, opened 2018-01-18
DM-13246: Synchronize DMTN-056 with Butler prototype

pschella:


python/lsst/butler/schema.py
```diff
@@ -180,8 +182,10 @@
 
 VisitRangeTable = \
     Table('VisitRange', metadata,
-          Column('visit_begin', Integer, primary_key=True, nullable=False),
-          Column('visit_end', Integer, primary_key=True, nullable=False),
+          Column('visit_begin', Integer, ForeignKey(
+              'Visit.visit_number'), primary_key=True, nullable=False),
+          Column('visit_end', Integer, ForeignKey(
+              'Visit.visit_number'), primary_key=True, nullable=False),
```
TallJimbo: I'm surprised it makes sense to actually make this a foreign key, given that it's defining a range.

TallJimbo (approved):

python/lsst/butler/schema.py
pschella: I was assuming the range had to be valid. So both need to exist. But perhaps this is not true?

TallJimbo: I think we'll want at least half-open ranges, presumably marked by null endpoints.

pschella: Ok, we'll remove the foreign key checks and hope for the best.

PR #8, opened 2018-01-25
DM-13370: Rebrand as lsst.daf.butler

timj:
In the prototype we called it `lsst.butler` but following advice from @ktlim we are rebranding as `lsst.daf.butler`.


timj:
@pschella has approved verbally.

PR #10, opened 2018-01-29
DM-13374: Deconstruct Butler prototype for redesign

pschella:


timj:
I wonder how I missed those flake8 fixes.

pschella:
Probably because they were the result of some other changes on this ticket (I intend to squash all commits).

python/lsst/daf/butler/core/datasets.py
```diff
@@ -33,7 +32,7 @@ def _safeMakeMappingProxyType(data):
     return MappingProxyType(data)
 
 
-class DatasetType:
+class DatasetType(object):
```
timj, line 35: Is this needed? Python 3 doesn't need the explicit `object` inheritance.

timj (approved):
Looks fine (I'm trying not to worry yet about code that's disappearing). Thanks for cleaning up the `type` variable name confusion.

PR #11, opened 2018-01-31
DM-13349: Define storage classes in config yaml

timj:
* Moved the formatter factory registry to a new class MappingFactory.
* Allow the factory registry to take classes as well as string definitions.
* Replace StorageClass definitions in code with a new registry and ability to dynamically create them from strings.
* Update posix data store to dynamically generate storage classes from the YAML.
* Update tests to read storage classes from factory.

### DIFF
--- a/python/lsst/daf/butler/core/formatter.py
+++ b/python/lsst/daf/butler/core/formatter.py
@@ -70,10 +70,12 @@ class Formatter(object, metaclass=ABCMeta):
         raise NotImplementedError("Type does not support writing")
 
 
-class FormatterFactory(MappingFactory):
+class FormatterFactory:
     """Factory for `Formatter` instances.
     """
-    refType = Formatter
+
+    def __init__(self):
+        self._mappingFactory = MappingFactory(Formatter)
 
     def getFormatter(self, storageClass, datasetType=None):
         """Get a new formatter instance.
@@ -86,7 +88,7 @@ class FormatterFactory(MappingFactory):
             If given, look if an override has been specified for this `DatasetType` and,
             if so return that instead.
         """
-        return self.getFromRegistry(storageClass, override=datasetType)
+        return self._mappingFactory.getFromRegistry(storageClass, override=datasetType)
 
     def registerFormatter(self, type_, formatter):
         """Register a Formatter.
@@ -103,4 +105,4 @@ class FormatterFactory(MappingFactory):
         e : `ValueError`
             If formatter does not name a valid formatter type.
         """
-        self.placeInRegistry(type_, formatter)
+        self._mappingFactory.placeInRegistry(type_, formatter)
--- a/python/lsst/daf/butler/core/formatter.py
+++ b/python/lsst/daf/butler/core/formatter.py
@@ -88,7 +88,7 @@ class FormatterFactory:
             If given, look if an override has been specified for this `DatasetType` and,
             if so return that instead.
         """
-        return self._mappingFactory.getFromRegistry(storageClass, override=datasetType)
+        return self._mappingFactory.getFromRegistry(datasetType, storageClass)
 
     def registerFormatter(self, type_, formatter):
         """Register a Formatter.
--- a/python/lsst/daf/butler/core/formatter.py
+++ b/python/lsst/daf/butler/core/formatter.py
@@ -23,7 +23,7 @@
 
 from abc import ABCMeta, abstractmethod
 
-from lsst.daf.persistence import doImport
+from .mappingFactory import MappingFactory
 
 
 class Formatter(object, metaclass=ABCMeta):
@@ -70,14 +70,10 @@ class Formatter(object, metaclass=ABCMeta):
         raise NotImplementedError("Type does not support writing")
 
 
-class FormatterFactory(object):
+class FormatterFactory(MappingFactory):
     """Factory for `Formatter` instances.
     """
-
-    def __init__(self):
-        """Constructor.
-        """
-        self._registry = {}
+    refType = Formatter
 
     def getFormatter(self, storageClass, datasetType=None):
         """Get a new formatter instance.
@@ -86,24 +82,18 @@ class FormatterFactory(object):
         ----------
         storageClass : `StorageClass`
             Get `Formatter` associated with this `StorageClass`, unless.
-        datasetType : `DatasetType` or `str, optional
+        datasetType : `DatasetType` or `str`, optional
             If given, look if an override has been specified for this `DatasetType` and,
             if so return that instead.
         """
-        if datasetType:
-            try:
-                typeName = self._registry[self._getName(datasetType)]()
-            except KeyError:
-                pass
-        typeName = self._registry[self._getName(storageClass)]
-        return self._getInstanceOf(typeName)
+        return self.getFromRegistry(storageClass, override=datasetType)
 
     def registerFormatter(self, type_, formatter):
         """Register a Formatter.
 
         Parameters
         ----------
-        type : `str` or `StorageClass` or `DatasetType`
+        type_ : `str` or `StorageClass` or `DatasetType`
             Type for which this formatter is to be used.
         formatter : `str`
             Identifies a `Formatter` subclass to use for reading and writing `Dataset`s of this type.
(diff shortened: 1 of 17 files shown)


python/lsst/daf/butler/core/formatter.py
```diff
@@ -86,24 +82,18 @@ def getFormatter(self, storageClass, datasetType=None):
...
             if so return that instead.
         """
-        if datasetType:
-            try:
-                typeName = self._registry[self._getName(datasetType)]()
-            except KeyError:
-                pass
-        typeName = self._registry[self._getName(storageClass)]
-        return self._getInstanceOf(typeName)
+        return self.getFromRegistry(storageClass, override=datasetType)
```
timj: I can switch this to using composition if you prefer since the get and put methods do not have the same names as the base class implementation

timj:
I've also relocated `doImport` from `daf_persistence` and added some tests to it.

python/lsst/daf/butler/core/mappingFactory.py
```diff
@@ -0,0 +1,144 @@
...
+    Register the mapping of some key to a python type and retrieve instances.
+
+    Enables instances of these classes to be retrieved from the factory later.
+    The class can be specified as an object, class or string.
+    If the key is an object it is converted to a string by accessing
+    a `name` attribute.
+    """
+
+    refType = None
+    """Type of instances expected to be returned from factory."""
```
pschella: Shouldn't this be documented in the `attributes` section of the class documentation instead?

tests/config/basic/butler.yaml
```diff
@@ -3,8 +3,32 @@ datastore:
...
+      components:
+        image: "Image"
+        mask: "Image"
+        variance: "Image"
+        wcs: "TablePersistable"
+        psf: "TablePersistable"
+        photoCalib: "TablePersistable"
+        visitInfo: "TablePersistable"
+        apCorr: "TablePersistable"
+        coaddInputs: "TablePersistable"
```
pschella: Are the quotes needed for the component storage classes? I'd like to specify them without quotes (or as `YAML` references), since they are internal references. Oh, and we better hope they don't become circular ;)

tests/test_formatter.py
```diff
@@ -0,0 +1,97 @@
...
+        self.factory = formatter.FormatterFactory()
+
+    def testRegistry(self):
+        """Check that formatters can be stored in the registry.
+        """
+        formatterName = "lsst.daf.butler.formatters.fitsCatalogFormatter.FitsCatalogFormatter"
+        storageClassName = "Image"
+        self.factory.registerFormatter(storageClassName, formatterName)
+        f = self.factory.getFormatter(storageClassName)
+        self.assertIsInstance(f, formatter.Formatter)
```
pschella: Also test the explicit subclass type?

tests/test_formatter.py
```diff
@@ -0,0 +1,97 @@
...
+
+        with self.assertRaises(ValueError):
+            # Try to store something that is not a Formatter but is a class
+            self.factory.registerFormatter("NotFormatter", "lsst.daf.butler.core.formatter.FormatterFactory")
```
pschella: Can be misleading if a value error is raised on the first argument instead of the second. Better reuse the same working identifier.

tests/test_formatter.py
```diff
@@ -0,0 +1,97 @@
...
+
+        with self.assertRaises(ValueError):
+            # Try to store something that is not a Formatter but is a module
+            self.factory.registerFormatter("NotFormatter", "lsst.daf.butler")
+
+        with self.assertRaises(ValueError):
+            # Try to store something that is not a Formatter but does not exist
```
pschella: and does not exist.

tests/test_formatter.py
```diff
@@ -0,0 +1,97 @@
...
+        formatterName = "lsst.daf.butler.formatters.fitsCatalogFormatter.FitsCatalogFormatter"
+        storageClassName = "TestClass"
+        sc = storageClass.makeNewStorageClass(storageClassName, dict, None)
+
+        # Store using an instance
+        self.factory.registerFormatter(sc(), formatterName)
+
+        # Retrieve using the class
+        f = self.factory.getFormatter(sc)
+        self.assertIsInstance(f, formatter.Formatter)
```
pschella: And the actual subclass type.

tests/test_formatter.py
```diff
@@ -0,0 +1,97 @@
...
+            # Try to store something that is not importable
+            self.factory.registerFormatter("NotImportable", "not a thing")
+
+        with self.assertRaises(KeyError):
+            f = self.factory.getFormatter("Missing")
+
+    def testRegistryWithStorageClass(self):
+        """Test that the registry can be given a StorageClass object.
+        """
+        formatterName = "lsst.daf.butler.formatters.fitsCatalogFormatter.FitsCatalogFormatter"
```
pschella: `formatterType` or `formatterTypeName`?

tests/test_storageClass.py
```diff
@@ -0,0 +1,86 @@
...
+        """Test that we can dynamically create storage class subclasses.
+
+        This is critical for testing the factory functions."""
+        className = "TestImage"
+        newclass = storageClass.makeNewStorageClass(className, dict, None)
+        sc = newclass()
+        self.assertIsInstance(sc, storageClass.StorageClass)
+        self.assertEqual(sc.name, className)
+        self.assertIsNone(sc.components)
+        self.assertEqual(sc.type, dict)
```
pschella: `ptype` / `pytype` here too?

tests/test_storageClass.py
```diff
@@ -0,0 +1,86 @@
...
+
+        # Check that this class is listed in the subclasses
+        self.assertIn(className, newclass.subclasses)
```
pschella: Now that we moved the registry to an explicit factory, I don't think we need `StorageClass`'s 'fancy' metaclass registration anymore (so `subclasses` and `components dict` can go I think).
It used to be responsible for mapping names to instances, but non-configurable and without overrides.

tests/test_storageClass.py
```diff
@@ -0,0 +1,86 @@
...
+import lsst.utils.tests
+
+import lsst.daf.butler.core.storageClass as storageClass
+
+"""Tests related to the StorageClass infrastructure.
+"""
+
+
+class PythonType:
+    """A dummy class to test the registory of Python types."""
```
pschella: `registry`?

python/lsst/daf/butler/core/formatter.py
pschella: Yes, that way we can also get rid of my other gripe (having `refType` specified) and move that to an argument (e.g. `self.mappingFactory = MappingFactory(Formatter)`).

python/lsst/daf/butler/core/mappingFactory.py
```diff
@@ -0,0 +1,144 @@
...
+    def getFromRegistry(self, targetClass, override=None):
+        """Get a new instance of the object stored in the registry.
+
+        Parameters
+        ----------
+        targetClass : `str` or object supporting `.name` attribute
+            Get item from registry associated with this target class, unless
+        override : `str` or object supporting `.name` attribute, optional
+            If given, look if an override has been specified for this and,
+            if so return that instead.
```
pschella: Perhaps make it slightly more generic? By taking `*targetClasses` and thus allowing for multiple overrides?

python/lsst/daf/butler/core/mappingFactory.py
timj: I spent ten minutes reading numpydoc documentation and this seemed to be the way they wanted me to document class attributes. I admit it was somewhat unclear to me (and I'm about to delete that line anyhow).

tests/config/basic/butler.yaml
timj: I can remove the quotes. Making them YAML references is something I'd have to look at (currently I need them to be strings to look up in the registry rather than items to look up from the YAML tree).

tests/test_formatter.py
timj: You mean the cases where "NotFormatter" is in the registry already? Shame I can't use `KeyError` for the "key already in use" error. What do you mean by "same working identifier"?

tests/test_formatter.py
timj: I'll use the latter then. The code is somewhat confused at the moment between names, types and classes.

tests/test_storageClass.py
timj: I did change all the `type` attributes to `pytype` at one point but then changed them back because I wasn't entirely sure that `type` was a bad name for the attribute (and keeping it meant changing less code). I'm fine with making it explicitly `pytype` so there is no confusion as to what the storage class is referring to.

tests/test_storageClass.py
timj: I might leave it in for a little just in case I have a problem where I need a storage class that has been made concrete somewhere else but is not in this registry.

python/lsst/daf/butler/core/mappingFactory.py
timj: Now that I read the numpydoc docs again I see the `attributes` section (as opposed to `parameters` for `__init__`).

PR #12, opened 2018-02-05
DM-13358: Initial composites implementation

timj:


python/lsst/daf/butler/core/storageClass.py
```diff
@@ -28,11 +28,51 @@
...
+        if cls._pytype is not None:
+            return cls._pytype
+        # Handle case where we did get a python type not string
+        if not isinstance(cls._pytypeName, str):
+            pytype = cls._pytypeName
+            cls._pytypeName = cls._pytypeName.__name__
+        elif cls._pytypeName == "dict":
+            pytype = dict
+        elif cls._pytypeName == "list":
+            pytype = list
```
pschella: Why do we have these as special cases? If we really need that, can't we use `builtins` somehow?

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -28,11 +28,51 @@
...
+    _assembler = None
+    _disassembler = None
+
     def __init__(self, name, bases, dct):
         super().__init__(name, bases, dct)
         if hasattr(self, "name"):
             StorageClass.subclasses[self.name] = self
 
+    @property
+    def pytype(cls):  # noqa N805
```
pschella: Docstring here?

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -105,24 +185,28 @@ def getStorageClass(self, storageClassName):
...
+                Map of storageClassName to `storageClass` for components.
 
         Raises
         ------
         e : `KeyError`
             If a storage class has already been registered with
             storageClassName.
         """
-        newtype = makeNewStorageClass(storageClassName, pytype, components)
+        newtype = makeNewStorageClass(storageClassName, **classAttr)
```
pschella: Given that `makeNewStorageClass` already creates the class, I think I'd rather call it outside `registerStorageClass` and pass the newly created class to it, rather than forwarding its constructor arguments.

So having: `def registerStorageClass(storageClass)` (or perhaps the type, instead of the instance).

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -66,7 +66,14 @@ def __init__(self, config):
...
                     components[cname] = self.storageClassFactory.getStorageClass(ctype)
-            self.storageClassFactory.registerStorageClass(name, info["type"], components)
+
+            # Extract scalar items from dict that are needed for StorageClass Constructor
+            scItems = {k: info[k] for k in ("pytype", "assembler", "disassembler") if k in info}
+
+            # Fill in other items
+            scItems["components"] = components
+
+            self.storageClassFactory.registerStorageClass(name, scItems)
```
pschella: Why not simply pass `**info`? After swapping out the components. Perhaps renaming it `storageClassKwargs` or something.

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -92,12 +99,38 @@ def get(self, uri, storageClass, parameters=None):
...
+        # component itself
+        comp = location.fragment
+        scComps = storageClass.components
+        if comp and scComps is not None:
+            refType = None  # Clear it since this *is* a component
+            if comp in scComps:
+                refType = scComps[comp].pytype
+
+        if refType and not isinstance(result, refType):
+            raise TypeError("Got type {} from formatter but expected {}".format(type(result), refType))
```
pschella: This check is necessary, but probably also insufficient. It fails if the component happens to be of the same type as the parent. How do we know if it actually read the component, rather than the parent?
One possible solution would be to do an (`md5`) hash or something, but this cannot be done in the `Datastore` unless it keeps a registry.

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -122,8 +155,23 @@ def put(self, inMemoryDataset, storageClass, storageHint, typeName=None):
...
+        # Check to see if this storage class has a disassembler
+        if storageClass.disassembler is not None:
+            compUris = {}
+            components = storageClass.disassembler(inMemoryDataset, storageClass)
+            for comp, info in components.items():
+                compTypeName = typeName
+                if compTypeName is not None:
+                    compTypeName = "{}.{}".format(compTypeName, comp)
+                compUris[comp], _ = self.put(info[0], info[1], "{}#{}".format(storageHint, comp),
+                                             compTypeName)
```
pschella: I'd prefer to use `urllib` (possibly via the `Location` interface) to do the URI composition. Otherwise one might get invalid URIs (e.g. with multiple fragments).

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -122,8 +155,23 @@ def put(self, inMemoryDataset, storageClass, storageHint, typeName=None):
...
+            return None, compUris
```
pschella: Only if it is a pure composite and doesn't have a parent.

tests/config/basic/butler.yaml
```diff
@@ -4,21 +4,44 @@ datastore:
   root: ./butler_test_repository
   create: true
   storageClasses:
+    StructuredDataDict:
+      pytype: dict
+      formatter: exampleFormatters.SimpleJSONFormatter
+    StructuredDataList:
+      pytype: list
+      formatter: exampleFormatters.SimpleJSONFormatter
```
pschella: Perhaps it would be nice to have this for all builtins (or at least those supported by `JSON/YAML`)? Then we could use this as a fallback, and people would be able to write almost anything with `Butler.put` in a structured way.

tests/exampleFormatters.py
```diff
@@ -0,0 +1,127 @@
...
+            A dictionary of URIs for the `MetricsExample`' components.
+            The latter will be empty if the `MetricsExample` is not a composite.
+        """
+        data = inMemoryDataset.exportAsDict()
+        with open(fileDescriptor.location.path, "w") as fd:
+            json.dump(data, fd)
+
+        baseURI = fileDescriptor.location.uri
+
+        return baseURI, {c: "{}#{}".format(baseURI, c) for c in ("output", "data", "summary")}
```
pschella: Again, prefer URI composing with `urllib`/`Location`.

tests/exampleFormatters.py
```diff
@@ -0,0 +1,127 @@
...
+    ----------
+    path : `str`
+        Path to use for reading/writing the fileDescriptor.
+    """
+
+    def getPath(self, fileDescriptor):
+        filepath = fileDescriptor.location.path
+        fragment = fileDescriptor.location.fragment
+        if fragment:
+            filepath = "{}#{}".format(filepath, fragment)
```
pschella: Same here.

tests/examplePythonTypes.py
```diff
@@ -0,0 +1,132 @@
...
+            A requested component can not be found.
+        """
+        requested = set(storageClass.components.keys())
+        components = {}
+        for c in list(requested):
+            # Remove from list
+            requested.remove(c)
+            sc = storageClass.components[c]
+            if hasattr(self, c):
+                components[c] = (getattr(self, c), sc)
```
pschella: I wonder if we can't make this dissasembler more general? Seems like extracting components based on attributes is probably almost always what we want.

tests/test_posixDatastore.py
```diff
@@ -1,7 +1,7 @@
 #
 # LSST Data Management System
 #
-# Copyright 2008-2018  AURA/LSST.
+# Copyright 2018  AURA/LSST.
```
pschella, line 4: Is this the new default?

tests/test_posixDatastore.py
```diff
@@ -25,77 +25,92 @@
...
+                                   typeName=None)
+
         # Get
-        catalogOut = datastore.get(uri, storageClass=storageClass, parameters=None)
-        datasetsHelper.assertCatalogEqual(self, catalog, catalogOut)
+        metricsOut = datastore.get(uri, storageClass=storageClass, parameters=None)
+        self.assertEqualMetrics(metrics, metricsOut)
+
+        # Get a component even though it was written without
+        summary = datastore.get("{}#{}".format(uri, "summary"), storageClass=storageClass)
```
pschella: Here it is ok I think.

tests/test_posixDatastore.py
```diff
@@ -25,77 +25,92 @@
...
+
         # These should raise
         with self.assertRaises(ValueError):
             # non-existing file
-            datastore.get(uri="file:///non_existing.fits", storageClass=storageClass, parameters=None)
+            datastore.get(uri="file:///non_existing.json", storageClass=storageClass, parameters=None)
         with self.assertRaises(ValueError):
             # invalid storage class
-            datastore.get(uri="file:///non_existing.fits", storageClass=object, parameters=None)
+            datastore.get(uri="file:///non_existing.json", storageClass=object, parameters=None)
```
pschella, line 83: We should be careful that we don't start relying on file extensions anywhere (e.g. object stores don't have them).

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -28,11 +28,51 @@
 
 class StorageClassMeta(type):
 
+    # Caches of imported code objects.
```
pschella: Do we still actually need the metaclass and its registry of subclasses? If not, we should remove it, and move its methods to regular `StorageClass` class methods, to reduce complexity.

tests/test_posixDatastore.py
timj: A new file written from scratch can't have a copyright year from before it existed.

python/lsst/daf/butler/core/storageClass.py
timj: You need a metaclass to support lazy loading of class attributes. The `subclasses` stuff is not used at the moment but I have wondered if it might be useful (will we ever need to locate a storage class that hasn't been put in the storage class registry). Can remove it easily enough once the code settles down.

python/lsst/daf/butler/core/storageClass.py
timj: We need a special case because the `else` clause tries to import them and you can't `import dict`. I hadn't thought about using `dir(builtins)` and seeing if the item exists there and checking it's a type and not a builtin_function_or_method.

python/lsst/daf/butler/core/storageClass.py
timj: Thanks. I had misunderstood how numpydoc worked with properties. I hadn't realized that I put a mention in the class docs but then the actual docs in the property.

python/lsst/daf/butler/core/storageClass.py
timj: oh, and this is metaclass lazy loading so I was wondering about how to document a metaclass at all.

python/lsst/daf/butler/core/storageClass.py
timj: and this is in the metaclass lazy loading so I wasn't sure how to document the metaclass at all.

python/lsst/daf/butler/datastores/posixDatastore.py
timj: It's a real shame we aren't using underscores rather than camel case 😄

python/lsst/daf/butler/datastores/posixDatastore.py
timj: Yes, if the parent has composites that are of the same class then this check will not be sufficient, but that is conceptually quite hard to spot.

python/lsst/daf/butler/datastores/posixDatastore.py
timj: I thought that we had agreed last week that things simplify enormously if composites are pure. As soon as we allow composites to be composed of composites plus some other stuff that is only in the parent, then that is more difficult.

tests/config/basic/butler.yaml
timj: Yes. I have no problem with moving code out of the test classes to be part of the standard set.

tests/examplePythonTypes.py
timj: Yes. `daf_persistence` has a generic disassembler that checks for standard get methods as well as properties.

tests/test_posixDatastore.py
timj: I'm not looking at file extensions anywhere at the moment. That get will fail regardless.

python/lsst/daf/butler/core/storageClass.py
timj: Definitely the type, since the mappingFactory returns the instance.

tests/exampleFormatters.py
timj: Actually, I don't think so. This one is not forming a URI, it's forming a file name. I also am requesting the fragment and path independently so there can't be a clash.

python/lsst/daf/butler/datastores/posixDatastore.py
pschella: So how do we express a new (`Exposure`) `Dataset` that consists of an existing `Dataset` with a different `Wcs`?

timj: You have URIs to all the components. Currently the datastore doesn't assemble itself (unlike disassembly which it does do) so you ask for all the bits you want individually and assemble them.

tests/test_posixDatastore.py
timj: I do now support file extensions inside formatters because some I/O libraries get upset if you don't have `.fits` (for example). The hint is treated as a hint and you get back a URI with the right extension.

python/lsst/daf/butler/assemblers/exposureAssembler.py
```diff
@@ -0,0 +1,156 @@
...
+    Returns
+    -------
+    expComps : `dict`
+        Components associated with the top level Exposure.
+    expInfoComps : `dict`
+        Components associated with the ExposureInfo
+    """
+    requested = set(storageClass.components.keys())
+    expItems = requested & EXPOSURE_COMPONENTS
+    expInfoItems = requested & EXPOSURE_INFO_COMPONENTS
```
pschella: Should we do a sanity check for requested components that are not in the union of the supported sets?

python/lsst/daf/butler/core/composites.py
```diff
@@ -0,0 +1,300 @@
...
+    ValueError
+        Some components could not be used to create the object.
+    """
+    if pytype is not None:
+        cls = pytype
+    else:
+        cls = storageClass.pytype
+
+    # Should we check that the storage class components match or are a superset
+    # of the items described in the supplied components?
```
pschella: Yes, the former I would say.

python/lsst/daf/butler/core/composites.py
```diff
@@ -0,0 +1,300 @@
...
+                    else:
+                        setter = getattr(obj, attr)
+                        setter(component)
+                    break
+            else:
+                failed.append(name)
+
+        if failed:
+            raise ValueError("Unhandled components during assembly ({})".format(failed))
+
```
pschella: Should we check that the decomposition returns the components again? Might be expensive to check though. So perhaps better done at the unittest level for each (dis)assembler.

python/lsst/daf/butler/formatters/fileFormatter.py
```diff
@@ -0,0 +1,224 @@
...
+from lsst.daf.butler.core.formatter import Formatter
+
+
+class FileFormatter(Formatter):
+    """Interface for reading and writing files on a POSIX file system.
+    """
+
+    extension = None
+    """Default file extension to use for writing files. None means that no
+    modifications will be made to the supplied file extension."""
```
pschella, line 37: Document this in the class level attributes section?

python/lsst/daf/butler/formatters/fileFormatter.py
```diff
@@ -0,0 +1,224 @@
...
+
+        Returns
+        -------
+        inMemoryDataset : `object`
+            Object of expected type `pytype`.
+        """
+        return inMemoryDataset
+
+    def read(self, fileDescriptor):
+        """Read a `Exposure` from a FITS file.
```
pschella: Why is this specific to `Exposure`?

python/lsst/daf/butler/formatters/fileFormatter.py
```diff
@@ -0,0 +1,224 @@
...
+        fileDescriptor : `FileDescriptor`
+            Identifies the file to read, type to read it into and parameters
+            to be used for reading.
+
+        Returns
+        -------
+        inMemoryDataset : `Exposure`
+            The requested `Exposure`.
+            The actual returned type will be a derived class
+            (e.g. `ExposureI` or `ExposureD`).
```
pschella: Same here.

python/lsst/daf/butler/formatters/fitsCatalogFormatter.py
```diff
@@ -20,50 +20,50 @@
...
-            A dictionary of URIs for the `Catalog`' components.
-            The latter will be empty if the `Catalog` is not a composite.
+        Raises
+        ------
+        Exception
+            The file could not be written.
         """
-        inMemoryDataset.writeFits(fileDescriptor.location.path)
-        return fileDescriptor.location.uri, {}
+        inMemoryDataset.writeFits(fileDescriptor.location.preferredPath())
```
pschella, line 69: What is the difference between `preferredPath` and `path`?

python/lsst/daf/butler/formatters/jsonFormatter.py
```diff
@@ -0,0 +1,124 @@
...
+        except FileNotFoundError:
+            data = None
+
+        return data
+
+    def _writeFile(self, inMemoryDataset, fileDescriptor):
+        """Write the in memory dataset to file on disk.
+
+        Will look for `_asdict()` method to aid JSON serialization, following
+        the approach of the simpljson module.
```
pschella: simplejson?

tests/config/basic/butler.yaml
```diff
@@ -4,31 +4,95 @@ datastore:
   root: ./butler_test_repository
   create: true
   storageClasses:
+    StructuredDataDict:
+      pytype: dict
+      formatter: lsst.daf.butler.formatters.jsonFormatter.JsonFormatter
+      # formatter: lsst.daf.butler.formatters.pickleFormatter.PickleFormatter
```
pschella, line 10: Please remove commented line

tests/config/basic/butler.yaml
```diff
@@ -4,31 +4,95 @@ datastore:
...
+    StructuredDataList:
+      pytype: list
+      formatter: lsst.daf.butler.formatters.pickleFormatter.PickleFormatter
+      # formatter: lsst.daf.butler.formatters.jsonFormatter.JsonFormatter
```
pschella, line 14: Please remove commented line

tests/config/basic/butler.yaml
```diff
@@ -4,31 +4,95 @@ datastore:
...
     StructuredData:
       # Data from a JSON or YAML file
-      type: dict
-      # formatter: "lsst.daf.butler.formatters.structuredDataFormatter.YAMLFormatter"
+      # type: dict
```
pschella: Please remove commented lines

tests/config/basic/butler.yaml
```diff
@@ -4,31 +4,95 @@ datastore:
...
+      formatter: lsst.daf.butler.formatters.jsonFormatter.JsonFormatter
+      # formatter: lsst.daf.butler.formatters.pickleFormatter.PickleFormatter
+    StructuredDataList:
+      pytype: list
+      formatter: lsst.daf.butler.formatters.pickleFormatter.PickleFormatter
+      # formatter: lsst.daf.butler.formatters.jsonFormatter.JsonFormatter
     StructuredData:
       # Data from a JSON or YAML file
-      type: dict
-      # formatter: "lsst.daf.butler.formatters.structuredDataFormatter.YAMLFormatter"
```
pschella, line 10: Would be nice to have a `YamlFormatter`, in particular if we could use it to store (part of) the config itself for bootstrapping purposes.

tests/exampleFormatters.py
```diff
@@ -0,0 +1,84 @@
...
+from lsst.daf.butler.core.formatter import Formatter
+import json
+
+
+class MetricsExampleFormatter(Formatter):
+    """Interface for reading and writing example metrics to and from JSON files.
+    """
+
+    def read(self, fileDescriptor):
+        """Read a `MetricsExample` from a JSON file.
```
pschella: Should this instead reuse `JsonFormatter`?

python/lsst/daf/butler/assemblers/exposureAssembler.py
```diff
@@ -0,0 +1,156 @@
...
+        return genericGetter(composite, componentName)
+    elif componentName in EXPOSURE_INFO_COMPONENTS:
+        return genericGetter(composite.getInfo(), componentName)
+    else:
+        raise AttributeError("Do not know how to retrieve component {} from {}".format(componentName,
+                                                                                       type(composite)))
+
+
+def validExposureComponents(composite, storageClass):
+    """Extract requested components from a composite.
```
pschella: It is unclear how this differs from `getComponentFromExposure`.

python/lsst/daf/butler/core/composites.py
```diff
@@ -0,0 +1,300 @@
...
+    Returns
+    -------
+    attrs : `tuple(str)`
+        Tuple of strings to attempt.
+    """
+    root = "get" if getter else "set"
+
+    # Capitalized name for getXxx must only capitalize first letter and not
+    # downcase the rest. getVisitInfo and not getVisitinfo
+    capitalized = "{}{}{}".format(root, componentName[0].upper(), componentName[1:])
```
pschella: This won't work for one letter getters such as`getX`, etc.

python/lsst/daf/butler/core/composites.py
```diff
@@ -0,0 +1,300 @@
...
+    return (componentName, "{}_{}".format(root, componentName), capitalized)
+
+
+def genericAssembler(storageClass, components, pytype=None):
```
pschella: The function name is ambiguous between returning an assembler or performing the assembling itself.

pschella: Also, but this may be my C++ self speaking, I'd think that it might be more natural to bundle `assemble` and `disassemble` as members of a base class.

python/lsst/daf/butler/core/composites.py
```diff
@@ -0,0 +1,300 @@
...
+        cls = storageClass.pytype
+
+    # Should we check that the storage class components match or are a superset
+    # of the items described in the supplied components?
+
+    # First try to create an instance directly using keyword args
+    try:
+        obj = cls(**components)
+    except TypeError:
+        obj = None
```
pschella: What if the constructor ignores the extra `kwargs`? Should we check that all components are in fact set?

python/lsst/daf/butler/core/composites.py
```diff
@@ -0,0 +1,300 @@
...
+            else:
+                failed.append(name)
+
+        if failed:
+            raise ValueError("Unhandled components during assembly ({})".format(failed))
+
+    return obj
+
+
+def validComponents(composite, storageClass):
```
pschella: Unclear naming. From the name I don't quite understand what it does.

python/lsst/daf/butler/core/composites.py
```diff
@@ -0,0 +1,300 @@
...
+                    comp = genericGetter(composite, c)
+                except AttributeError:
+                    pass
+                else:
+                    if comp is not None:
+                        components[c] = comp
+    return components
+
+
+def hasComponent(composite, componentName):
```
pschella: `getLikelyComponent`?

python/lsst/daf/butler/core/composites.py
```diff
@@ -0,0 +1,300 @@
...
+    """Determine if it seems likely that the composite has the component.
```
pschella: and return it.

python/lsst/daf/butler/core/composites.py
```diff
@@ -0,0 +1,300 @@
...
+
+    Parameters
+    ----------
+    composite : `object`
+        Parent composite object consisting of components to be extracted.
+    storageClass : `StorageClass`
+        `StorageClass` associated with the parent, with defined components.
+    subset : `iterable`, optional
+        Iterable containing subset of components to extract from composite.
+        Must be a subset of those defined in `storageClass`.
```
pschella: Perhaps good for flexibility, but in practice we might choose to simplify matters by simply always extracting exactly those components mentioned in the `StorageClass`.

python/lsst/daf/butler/core/composites.py
```diff
@@ -0,0 +1,300 @@
...
+
+    if subset is not None:
+        subset = set(subset)
+        diff = subset - requested
+        if diff:
+            raise ValueError("Requested subset is not a subset of supported components: {}".format(diff))
+        requested = subset
+
+    if override is not None:
+        composite = override
```
pschella: Why is this not handled outside the function?

python/lsst/daf/butler/core/composites.py
```diff
@@ -0,0 +1,300 @@
...
+
+    components = {}
+    for c in list(requested):
```
pschella: Why turn this into a list?

python/lsst/daf/butler/core/location.py
```diff
@@ -54,6 +55,81 @@ def path(self):
         """
         return os.path.join(self._datastoreRoot, self._uri.path.lstrip("/"))
 
+    @property
+    def component_path(self):
```
pschella, line 59: Why snake-case?

python/lsst/daf/butler/core/location.py
```diff
@@ -54,6 +55,81 @@ def path(self):
...
+
+        This path includes the root of the `Datastore`. Returns the empty
+        string if no fragment has been specified as part of the URI.
+        """
+        if not self.fragment:
+            return ''
+
+        # We need to insert the fragment before the file extension because
+        # some I/O libraries insist on a particular extension.
+        filepath, extension = os.path.splitext(self.path)
```
pschella, line 76: Does this include the period?

python/lsst/daf/butler/core/location.py
```diff
@@ -54,6 +55,81 @@ def path(self):
...
+        return "{}#{}{}".format(filepath, self.fragment, extension)
```
pschella, line 77: What if the user wants to store the component of a composite in a different location? For instance if the user wants to store all `Psf`s together in one location, regardless if they are independent or part of an exposure?

python/lsst/daf/butler/core/location.py
```diff
@@ -54,6 +55,81 @@ def path(self):
...
+        return (self.path, self.component_path)
+
+    def preferredPath(self):
+        """Returns component path, if set, else the path.
+
+        Returns
+        -------
+            Path for component in composite if set, else the default path.
+        """
+        return self.component_path if self.component_path else self.path
```
pschella, line 103: Why doesn't this simply return the `path` in all cases and handle component storage locations outside of it based on user configuration?

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -21,18 +21,76 @@
...
 
 from .mappingFactory import MappingFactory
 
 
 class StorageClassMeta(type):
 
+    """Metaclass used by `StorageClass`.
+
+    Implements lazy loading of class attributes, allowing datastores to
+    delay loading of external code until it is needed.
```
pschella: Still don't understand why this can't be handled in `StorageClass` itself and has to be in the metaclass.

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -43,43 +101,100 @@ class StorageClass(metaclass=StorageClassMeta):
...
+    @property
+    def assembler(self):
+        """Function object to use to create a type from components."""
+        return type(self).assembler
+
+    @property
+    def disassembler(self):
+        """Function object to use to disassembler a type into components."""
+        return type(self).disassembler
+
```
pschella: I think it would be more elegant if e.g. `SourceCatalog` and the like were instances of `StorageClass`.
Then we won't need the metaclass. Additionally we could have `assemble` and `disassemble` be instance methods on a separate object that is passed to the `StorageClass` constructor (as discussed above). Right now the structure feels a bit forced with too many layers of inheritance.

tests/examplePythonTypes.py
```diff
@@ -0,0 +1,89 @@
...
+
+        This interface is used for JSON serialization.
+
+        Returns
+        -------
+        exportDict : `dict`
+            Object contents in the form of a dict with keys corresponding
+            to object attributes.
+        """
+        return self.exportAsDict()
```
pschella, line 69: Why have both?

tests/test_posixDatastoreFits.py
```diff
@@ -0,0 +1,178 @@
...
+        if lsst.afw.table is None:
+            raise unittest.SkipTest("afw not available.")
+
+    def setUp(self):
+        self.testDir = os.path.dirname(__file__)
+        self.configFile = os.path.join(self.testDir, "config/basic/butler.yaml")
+
+    def testConstructor(self):
+        datastore = PosixDatastore(config=self.configFile)
+        self.assertIsNotNone(datastore)
```
pschella, line 54: We should probably test also that the factory function returns the proper `PosixDatastore` instance.

tests/test_posixDatastoreFits.py
```diff
@@ -0,0 +1,178 @@
...
+            self.assertIsNotNone(component)
+
+        # Get a component to check it
+        self.assertIn("wcs", comps)
+        wcs = datastore.get(comps["wcs"], storageClass=storageClass)
+
+        # Simple check of WCS
+        bbox = lsst.afw.geom.Box2I(lsst.afw.geom.Point2I(0, 0),
+                                   lsst.afw.geom.Extent2I(9, 9))
+        self.assertWcsAlmostEqualOverBBox(wcs, exposure.getWcs(), bbox)
```
pschella, line 143: Test the actual component values?

python/lsst/daf/butler/core/composites.py
timj: Done. Has to be a superset though since you should be allowed to assemble an exposure from N of the M components if the pytype can support that.

python/lsst/daf/butler/core/composites.py
timj: I think asking the assembler to call the disassembler to compare components is going to be more than we need and we should have unittests for each assembler/disassembler that does that.

python/lsst/daf/butler/formatters/fileFormatter.py
timj: That's what this is (meant) to do: https://developer.lsst.io/docs/py_docs.html?highlight=numpydoc#placement-of-constant-and-class-attribute-docstrings

python/lsst/daf/butler/formatters/fileFormatter.py
timj: Because during the code refactoring I didn't notice that. Will fix.

tests/exampleFormatters.py
timj: Actually, that entire formatter is unused following the refactoring of the file/JSON formatters. It can be deleted.

python/lsst/daf/butler/assemblers/exposureAssembler.py
timj: `getComponentFromExposure` returns the requested component. `validExposureComponents` returns all the non-`None` components.

timj: Maybe change the name to `getValidExposureComponents` and `getValidComponents`? The name is partly a hold over to the first implementation which returned a list of valid components but not the components themselves.

python/lsst/daf/butler/core/composites.py
timj: Good catch.

python/lsst/daf/butler/core/composites.py
timj: I can switch to a base class with Exposure overrides.  Assemblers would then be listed as classes in the config file rather than assembler functions.

python/lsst/daf/butler/core/composites.py
timj: I do worry about getting into a world where every assembly has to be followed by a disassembly (effectively `getValidComponents`) but I'll see what happens.

python/lsst/daf/butler/core/composites.py
timj: Function is not used at the moment since when I wrote it I later realized that in my usage what I wanted to know was whether the component was supported and not None. There is also confusion over attribute versus getter method.

python/lsst/daf/butler/core/composites.py
timj: I added this feature (and the override) when I started working in the real world of Exposure. The problem with Exposure is that some of the components are part of Exposure but others are one level down and part of `ExposureInfo`. Since I didn't have an ExposureInfo StorageClass (maybe I should have?) I couldn't tell it not to get upset when it couldn't get the calibs or whatever from the `Exposure`.

python/lsst/daf/butler/core/composites.py
timj: For the same reason subsets existed. StorageClass doesn't know about ExposureInfo and it was far simpler to be able to pass in ExposureInfo and a subset when dealing with Exposure.

python/lsst/daf/butler/core/composites.py
timj: because components are removed as the loop progresses.

python/lsst/daf/butler/core/location.py
timj: because I really want to write python code that looks like python code. I'm assuming even though daf_butler is new we aren't allowed to switch because of the "don't confuse lsst_apps people" exemption. I'll switch whilst grumbling.

python/lsst/daf/butler/core/location.py
timj: Yes
```
>>> import os.path
>>> os.path.splitext("a.fits")
('a', '.fits')
```

python/lsst/daf/butler/core/location.py
timj: I didn't think the user was allowed to control where the datastore puts files. I did it this way because I want to group all the components together.

python/lsst/daf/butler/core/location.py
timj: we never said that user configuration would control how components are split up yet, and we don't yet have templating in config files. When templating is added to datastore configurations we will likely have to revisit this. `Location` has to know the component storage scheme at the moment because `Location` is currently used by other parts of datastore to determine whether a file exists or not independent of formatter. Once we switch to datastore having to map dataIDs to filenames in a database table `Location` won't need to know any more.

python/lsst/daf/butler/core/storageClass.py
timj: It's a class attribute. You have to be able to lazy load the class type without instantiating a StorageClass instance. If you want it to be an instance attribute then I can switch it to being a standard `@property` lazy loader.

python/lsst/daf/butler/core/storageClass.py
timj: They are instances of StorageClass (that's what Factory does). If you don't need `StorageClass.pytype` to work then this can be simplified.

tests/examplePythonTypes.py
timj: History. Although if this was production code and not a test class, it might be useful to have the public interface and the private one.

tests/test_posixDatastoreFits.py
timj: This is your original test code so I haven't looked at it at all.

tests/test_posixDatastoreFits.py
timj: This is testing the Wcs values. I don't necessarily know how to test every single Exposure component but I can default to seeing if `==` works.

python/lsst/daf/butler/formatters/fitsCatalogFormatter.py
timj: I suggest we punt on this question for the review since I believe it will be addressed by the use of templates and the removal of URIs.

PR #13, opened 2018-02-05
DM-13496: Prototype Butler Registry schema definition in YAML

pschella:


config/registry/default_schema.yaml
```diff
@@ -0,0 +1,693 @@
...
+      constraints:
+      - !ForeignKeyConstraint
+        src: camera
+        tgt: Camera.camera
+  
+    - !DataUnit
+      name: Visit
+      dependencies:
+      - Camera
+      - PhysicalFilter
```
TallJimbo: I think we need some way of distinguishing between these relationships in this structure:
 - `camera` is part of `Visit`'s primary key, and needs to be provided along with the `visit` number field to uniquely identify a `Visit`.
 - `physical_filter` is not a part of `Visit`'s primary key, and is not necessary to uniquely identify a `Visit`.

config/registry/default_schema.yaml
```diff
@@ -0,0 +1,693 @@
...
+        - Visit.visit
+  
+    - !DataUnit
+      name: ExposureRange
+      dependencies:
+      - Camera
+      doc: >
+        An inclusive range of Exposures that may be open in either direction.  There
+        is no SQL representation; its fields (duplicated below) are all part of its
+        compound primary key and are hence just present in Dataset.
```
TallJimbo: We should have a machine-readable way to represent the fact that the columns here are just fields that must be present `Dataset`, not something that requires an new Table to be created.

config/registry/default_schema.yaml
```diff
@@ -0,0 +1,693 @@
...
+        doc: >
+          Link to the Run table.
+      - !Column
+        name: quantum_id
+        type: int
+        doc: >
+          Link to the Quantum table (only present for SuperTask-generated Datasets).
+      # DataUnit values
+      - !Column
+        <<: *camera
```
TallJimbo: Are these going to inherit the `primary_key` attributes these column definitions have in the `DataUnits` section?

config/registry/default_schema.yaml
```diff
@@ -0,0 +1,693 @@
...
+    - !Table
+      name: DatasetComposition
+      doc: >
+        Self-join table that relates components of a dataset to their parent.
+      columns:
+      - !Column
+        name: parent_dataset_id
+        type: int
+        primary_key: true
+        nullable: false
```
TallJimbo: In SQL, `primary_key=True` always implies `nullable=False`.  Should we take advantage of that?

config/registry/default_schema.yaml
pschella: My thought on that was that if all columns are values then no table is needed and all that exist are those columns in the `Datasets` table (e.g. the respective `DataUnit` table is skipped). Otherwise a separate table is created for this `DataUnit`.

config/registry/default_schema.yaml
pschella: Yes, unless overridden in the same way as `nullable` is.

config/registry/default_schema.yaml
pschella: Probably, although I'd want to check the standard on that to be sure (e.g. the actual primary key in `sqllite` is some auto-generated thing that you don't see so perhaps it does allow it to be `nullable`? Not sure.

config/registry/default_schema.yaml
TallJimbo: By "values", you mean "appears in the `Dataset` table"?  I don't see those indicated some other way (though I think the condition you want might actually be "if all columns are primary keys").

In any case, even that condition isn't enough, because it can't distinguish between "have a definition-only table" and "no table".

config/registry/default_schema.yaml
```diff
@@ -0,0 +1,719 @@
...
+          optional:
+          - AbstractFilter
+        doc: >
+          A filter associated with a particular camera.  Physical filters are used to
+          identify datasets that can only be associated with a single observation.
+        link:
+        - 
+          name: physical_filter
+          type: string
+          foreign_key: PhysicalFilter.physical_filter
```
TallJimbo, line 59: Should this have one `link` (`physical_filter`) or two (`physical_filter` and `camera`)?  I think you have all of the necessary information here, but just to confirm it's structured:
 - No two `DataUnits` may have the same `link`.
 - A single `DataUnits` may (rarely) have more than one `link` (e.g. `ExposureRange`).
 - A `DataUnit` is fully specified by its `link` and the `link`s of all `required` `dependencies`.  If the `DataUnit` has a table, "fully specified" means those columns correspond to its (compound) primary key.

config/registry/default_schema.yaml
```diff
@@ -0,0 +1,719 @@
...
+              - camera
+              - physical_filter
+              tgt:
+              - PhysicalFilter.camera
+              - PhysicalFilter.physical_filter
+    
+      Exposure:
+        dependencies:
+          required:
+          - Camera
```
TallJimbo, line 191: Also:
```
optional:
- Visit
```
right?

(Just making sure this an accidental omission, not something about the structure I don't understand.)

config/registry/default_schema.yaml
pschella: Yes, that is correct. I am assuming that when a link is inserted the dependencies are first expanded and also included. Thus dupplicate links are not allowed.

config/registry/default_schema.yaml
pschella: Yes, this was an accidental omission (there may be more).

PR #14, opened 2018-02-20
DM-12640:  Implement Initial Butler Registry Hierarchy

pschella:


python/lsst/daf/butler/core/registry.py
```diff
@@ -37,18 +37,25 @@ class Registry(metaclass=ABCMeta):
     """
     @staticmethod
     def fromConfig(config):
-        cls = doImport(config['registry.cls'])
+        if not isinstance(config, RegistryConfig):
+            if isinstance(config, str):
+                config = Config(config)
+            if isinstance(config, Config):
```
timj: `elif`?

python/lsst/daf/butler/core/registry.py
```diff
@@ -37,18 +37,25 @@ class Registry(metaclass=ABCMeta):
     """
     @staticmethod
     def fromConfig(config):
-        cls = doImport(config['registry.cls'])
```
timj, line 38: Add docstring.

python/lsst/daf/butler/core/registry.py
```diff
@@ -37,18 +37,25 @@ class Registry(metaclass=ABCMeta):
...
+                config = Config(config)
+            if isinstance(config, Config):
+                config = RegistryConfig(config['registry'])
+            else:
+                raise ValueError("Incompatible Registry configuration: {}".format(config))
+        cls = doImport(config['cls'])
         return cls(config=config)
 
     def __init__(self, config):
         """Constructor
```
timj: docstring for constructor goes in class docstring.

https://developer.lsst.io/docs/py_docs.html#placement-of-class-docstrings

I won't mention this again in the review.

tests/test_schema.py
```diff
@@ -0,0 +1,125 @@
+#
+# LSST Data Management System
+#
+# Copyright 2008-2018  AURA/LSST.
```
timj: Please use the updated preamble. Remove this copyright because it's in the `COPYRIGHT` file now.

tests/test_schema.py
```diff
@@ -0,0 +1,125 @@
...
+    .. warning::
+        This unittest does not verify the validaty of the schema description.
+        It only checks that the generated tables match it.
+    """
+    def setUp(self):
+        self.testDir = os.path.dirname(__file__)
+        self.schemaFile = os.path.join(self.testDir, "../config/registry/default_schema.yaml")
+        self.config = SchemaConfig(self.schemaFile)
+        self.schema = Schema(self.config)
+        self.engine = create_engine('sqlite:///:memory:')
```
timj, line 48: This is good. Means that the tests can all run in parallel.

python/lsst/daf/butler/core/schema.py
```diff
@@ -21,351 +21,148 @@
 # see <https://www.lsstcorp.org/LegalNotices/>.
```
timj: I'm not sure how this file still has the old preamble. I fixed it in 0346dbc

python/lsst/daf/butler/core/schema.py
```diff
@@ -21,351 +21,148 @@
...
-          Column('registry_id', Integer, nullable=False, primary_key=True),
-          ForeignKeyConstraint(['patch_index', 'tract_number', 'skymap_name'], [
-              'Patch.patch_index', 'Patch.tract_number', 'Patch.skymap_name']),
-          ForeignKeyConstraint(['dataset_id', 'registry_id'], [
-              'Dataset.dataset_id', 'Dataset.registry_id'])
-          )
+metadata = None  # Needed to make disabled test_hsc not fail on import
+
+
+def iterable(a):
```
timj: I think this should go in the `utils.py` package.

python/lsst/daf/butler/core/schema.py
```diff
@@ -21,351 +21,148 @@
...
+    """Make input iterable.
+
+    There are three cases, when the input is:
+    - iterable, but not a string -> iterate over elements (e.g. [i for i in a])
+    - a string -> return single element iterable (e.g. [a])
+    - not iterable -> return single elment iterable (e.g. [a]).
+    """
+    if hasattr(a, 'encode'):
```
timj: Why can't we just see if it's a `str`?

python/lsst/daf/butler/core/schema.py
```diff
@@ -21,351 +21,148 @@
...
+    def makeColumn(self, columnDescription):
+        """Make a Column entry for addition to a Table.
+
+        Parameters
+        ----------
+        columnDescription : `dict`
+            Description of the column to be created.
+            Should always contain:
+                - name, descriptive name
+                - type, valid column type
```
timj, line 101: This type is a type understood by the underlying database technology (in this case sqlaclhemy?) and not some abstract type that will be translated to something db-specific later on?

python/lsst/daf/butler/core/schema.py
```diff
@@ -21,351 +21,148 @@
...
+                - doc, docstring
+
+        Returns
+        -------
+        c : `sqlalchemy.Column`
+            The created `Column` entry.
+        
+        Raises
+        ------
+        e : `ValueError
```
timj: The docstring style guide does not include the `e :` for exceptions.

timj (changes requested):
Looks good but some cleanup required.

python/lsst/daf/butler/core/registry.py
pschella: No, because it needs to build a `RegistryConfig`.

python/lsst/daf/butler/core/schema.py
pschella: I had a debate about this with @natelust. This should be faster.

python/lsst/daf/butler/core/schema.py
pschella: Not quite. It should be a type in our type naming convention (which we haven't formalized yet).

python/lsst/daf/butler/core/schema.py
pschella: Doesn't the `a : T` syntax require a name?

python/lsst/daf/butler/core/schema.py
timj:
```
In [3]: x = "xxx"

In [4]: %timeit hasattr(x, "encode")
109 ns ± 0.692 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [5]: %timeit isinstance(x, str)
102 ns ± 0.961 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```
It doesn't look obviously different. I think we should go for the obvious implementation rather than looking for a side effect (and I can easily break yours by having a class with an encode method).

python/lsst/daf/butler/core/schema.py
timj: https://developer.lsst.io/docs/py_docs.html#raises

timj (approved):

PR #16, opened 2018-02-20
DM-13600: Add YAML formatter

timj:
Add a basic YAML formatter and update test config and tests to ensure that YAML, JSON and Pickle are tested.

### DIFF
--- a/python/lsst/daf/butler/formatters/yamlFormatter.py
+++ b/python/lsst/daf/butler/formatters/yamlFormatter.py
@@ -1,0 +1,97 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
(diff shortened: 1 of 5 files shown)


python/lsst/daf/butler/formatters/yamlFormatter.py
```diff
@@ -0,0 +1,97 @@
...
+    def _coerceType(self, inMemoryDataset, storageClass, pytype=None):
+        """Coerce the supplied inMemoryDataset to type `pytype`.
+
+        Parameters
+        ----------
+        inMemoryDataset : `object`
+            Object to coerce to expected type.
+        storageClass : `StorageClass`
+            StorageClass associated with `inMemoryDataset`.
+        pytype : `class`, optional
```
TallJimbo: `class` -> `type`?

python/lsst/daf/butler/formatters/yamlFormatter.py
```diff
@@ -0,0 +1,97 @@
...
+        Parameters
+        ----------
+        path : `str`
+            Path to use to open YAML format file.
+
+        Returns
+        -------
+        data : `object`
+            Either data as Python object read from YAML file, or None
+            if the file could not be opened.
```
TallJimbo, line 45: I'm a bit surprised that we don't raise an exception when we can't find the file.  Is this the pattern for all formatters?

TallJimbo (approved):

python/lsst/daf/butler/formatters/yamlFormatter.py
timj: Ok. I'm a bit unclear as to what we really want there so I'll switch it.

python/lsst/daf/butler/formatters/yamlFormatter.py
timj: It made `FileFormatter.read()` a little less convoluted and finding the file was the only thing that mattered. I think this code will have to be rejiggered when we start using templates for file names so I'll have another look at that time. @pschella didn't like including the component name in the filename (I think he preferred it to be under the control of the template) so this will change.

PR #17, opened 2018-02-22
DM-13350: Use templates to work out file names from data units

timj:
Templates are specified in the YAML configuration file and indexed
by dataset type.

Needs some clean up but fundamentally seems sound. Doesn't yet support fallback to a user function and I haven't got standalone tests for the template formatting.

### DIFF
--- a/python/lsst/daf/butler/core/dataUnits.py
+++ b/python/lsst/daf/butler/core/dataUnits.py
@@ -1,0 +1,39 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Code relating to DataUnits."""
+
+
+class DataUnits:
+    """Represent DataUnits specification.
+
+    Parameters
+    ----------
+    units : `dict`
+        Dictionary of data units keys and values.
+    """
+
+    def __init__(self, units):
+        self.units = units.copy()
+
+    def definedUnits(self):
+        """DataUnits with non-None values."""
+        return {k: v for k, v in self.units.items() if v is not None}
--- a/python/lsst/daf/butler/core/fileTemplates.py
+++ b/python/lsst/daf/butler/core/fileTemplates.py
@@ -1,0 +1,137 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Support for file template string expansion."""
+
+import os.path
+import string
+
+
+class FileTemplate:
+    """Format a path template into a fully expanded path.
+
+    Parameters
+    ----------
+    template : `str`
+        Template string.
+
+    Notes
+    -----
+    The templates use the standard Format Specification Mini-Language
+    with the caveat that only named fields can be used. The field names
+    are taken from the DataUnits along with two additional fields:
+    "datasettype" will be replaced with the DataSetType and "component"
+    will be replaced with the component name of a composite.
+
+    The mini-language is extended to understand a "?" in the format
+    specification. This indicates that a field is optional. If that
+    DataUnit is missing the field, along with the text before the field,
+    unless it is a path separator, will be removed from the output path.
+    """
+
+    def __init__(self, template):
+        self.template = template
+
+    def format(self, dataunits, datasettype=None, component=None):
+        """Format a template string into a full path.
+
+        Parameters
+        ----------
+        dataunits : `DataUnits`
+            DataUnits and the corresponding values.
+        datasettype : `str`, optional.
+            DataSetType name to use if needed. If it contains a "." separator
+            the type name will be split up into the main DataSetType and a
+            component.
+        component : `str`, optional
+            Component of a composite. If `datasettype` defines a component
+            this parameter will be ignored.
+
+        Returns
+        -------
+        path : `str`
+            Expanded path.
+
+        Raises
+        ------
+        KeyError
+            Requested field is not defined and the field is not optional.
+            Or, `component` is specified but "component" was not part of
+            the template.
+        """
+        units = dataunits.definedUnits()
+
+        if datasettype is not None:
+            # calexp.wcs means wcs component of a calexp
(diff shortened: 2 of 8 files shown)


python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -0,0 +1,137 @@
...
+    ----------
+    template : `str`
+        Template string.
+
+    Notes
+    -----
+    The templates use the standard Format Specification Mini-Language
+    with the caveat that only named fields can be used. The field names
+    are taken from the DataUnits along with two additional fields:
+    "datasettype" will be replaced with the DataSetType and "component"
```
pschella: `"datasetType"` and `DatasetType`.

python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -0,0 +1,137 @@
...
+    The mini-language is extended to understand a "?" in the format
+    specification. This indicates that a field is optional. If that
+    DataUnit is missing the field, along with the text before the field,
+    unless it is a path separator, will be removed from the output path.
+    """
+
+    def __init__(self, template):
+        self.template = template
+
+    def format(self, dataunits, datasettype=None, component=None):
```
pschella: `dataUnits` or `units` and `datasetType`.

python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -0,0 +1,137 @@
...
+        ----------
+        dataunits : `DataUnits`
+            DataUnits and the corresponding values.
+        datasettype : `str`, optional.
+            DataSetType name to use if needed. If it contains a "." separator
+            the type name will be split up into the main DataSetType and a
+            component.
+        component : `str`, optional
+            Component of a composite. If `datasettype` defines a component
+            this parameter will be ignored.
```
pschella: Same in the above.

python/lsst/daf/butler/core/dataUnits.py
```diff
@@ -0,0 +1,39 @@
...
+    units : `dict`
+        Dictionary of data units keys and values.
+    """
+
+    def __init__(self, units):
+        self.units = units.copy()
+
+    def definedUnits(self):
+        """DataUnits with non-None values."""
+        return {k: v for k, v in self.units.items() if v is not None}
```
pschella, line 39: Not sure why this is needed. Should be absorbed into `DataUnitTypeSet`. But we will have to do that later anyway, so fine to keep it around for now I suppose.

python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -0,0 +1,137 @@
...
+            Or, `component` is specified but "component" was not part of
+            the template.
+        """
+        units = dataunits.definedUnits()
+
+        if datasettype is not None:
+            # calexp.wcs means wcs component of a calexp
+            if "." in datasettype:
+                datasettype, component = datasettype.split(".", maxsplit=1)
+            units["datasettype"] = datasettype
```
pschella: Now the name `units` becomes confusing. Better call it something else, like `fields`.

python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -0,0 +1,137 @@
...
+                format_spec = format_spec.replace("?", "")
+            else:
+                optional = False
+
+            if field_name in units:
+                value = units[field_name]
+            elif optional:
+                # If this is optional ignore the format spec
+                # and do not include the literal text prior to the optional
+                # field unless it contains a "/" path separator
```
pschella: Will it now have two path separators? E.g. `/{visit}/{hello:?}/{camera} -> /3//hsc`.

python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -0,0 +1,137 @@
...
+            else:
+                raise KeyError("{} requested in template but not defined and not optional".format(field_name))
+
+            # Now use standard formatting
+            output = output + literal + format(value, format_spec)
+
+        # Complain if we were meant to use a component
+        if component is not None and not usedComponent:
+            raise KeyError("Component {} specified but template {} did not use it".format(component,
+                                                                                          self.template))
```
pschella, line 182: Why is this different from any other field?

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -168,9 +174,31 @@ def put(self, inMemoryDataset, storageClass, storageHint, typeName=None):
...
+            if typeName in self.templates:
+                template = self.templates[typeName]
+            elif "." in typeName:
+                baseType, component = typeName.split(".", maxsplit=1)
+                if baseType in self.templates:
+                    template = self.templates[baseType]
+
+        if template is None:
+            if "default" in self.templates:
+                template = self.templates["default"]
```
pschella: This loading could go into a separate function or `classmethod` of `Template`.

tests/config/basic/butler.yaml
```diff
@@ -3,6 +3,10 @@ datastore:
   cls: lsst.daf.butler.datastores.posixDatastore.PosixDatastore
   root: ./butler_test_repository
   create: true
+  templates:
+    default: "{datasettype}/{tract:?}/{patch:?}/{filter:?}/{visit:?}"
+    calexp: "{datasettype}.{component:?}/{datasettype}_v{visit}_f{filter}_{component:?}"
+    metric: "{datasettype}.{component:?}/{datasettype}_v{visit:08d}_f{filter}_{component:?}"
```
pschella: Per `DatasetType` templates should (probably) live in `Registry`. Given that we are dynamically adding them.

tests/test_templates.py
```diff
@@ -0,0 +1,74 @@
...
+        self.assertTemplate(tmplstr,
+                            tmplstr.format(datasettype="calexp", **self.du.units),
+                            self.du, datasettype="calexp")
+
+    def testOptional(self):
+        tmplstr = "{datasettype}.{component:?}/v{visit:05d}_f{filter}"
+        self.assertTemplate(tmplstr, "calexp.wcs/v00052_fU",
+                            self.du, datasettype="calexp.wcs")
+        self.assertTemplate(tmplstr, "calexp.psf/v00052_fU",
+                            self.du, datasettype="calexp", component="psf")
```
pschella: Without components? With multiple optionals?

tests/test_templates.py
```diff
@@ -0,0 +1,74 @@
...
+
+        with self.assertRaises(KeyError):
+            self.assertTemplate(tmplstr, "", self.du, component="wcs")
```
pschella: Different testcase?

python/lsst/daf/butler/core/fileTemplates.py
timj: no because of the path cleanup on the last line of the method.

python/lsst/daf/butler/core/fileTemplates.py
timj: because if you explicitly specify a component but the component is not used that's going to be a problem because it means the template does not specify how to distinguish components.

tests/test_templates.py
timj: I'm fine with having a separate test method for components.

tests/config/basic/butler.yaml
timj: They are in datastore because we explicitly decided they should be moved inside datastore (before we had storage hints which would have been the butler using the templates). Not all datastores should define templates so I'm not sure why registry wants to define them. Is the issue that DataSetTypes are defined in registry and it seems odd that data store needs to have its own list of them? If I have chained datastores shouldn't I be able to use different templates in each one?

python/lsst/daf/butler/datastores/posixDatastore.py
timj: Yes. Ok, I would seed it with the config object.

python/lsst/daf/butler/core/dataUnits.py
timj: I did not know we had a class for that. I asked on Slack but got no hints. Sorry.

python/lsst/daf/butler/core/fileTemplates.py
timj: I was assuming that DataUnits could be far more detailed than is needed for the file templates. That's why I didn't check that every unit was used.

python/lsst/daf/butler/core/dataUnits.py
timj: I can't see a `DataUnitTypeSet` in the codebase.

python/lsst/daf/butler/datastores/posixDatastore.py
timj: This is determining the right template to use based on the dataset type by looking up in the datastore's template registry. I think I can make this work like the Factory we use for formatters and storage classes (although we are storing instances and not classes).

timj:
@pschella I've rejigged things so that template collections are in their own class and I've extended the tests. Are you now okay with these changes? We agreed we would not move the template definitions just yet.

PR #18, opened 2018-03-10
DM-13753: Add Sphinx documentation support

timj:
* Enable Sphinx in docs/ subdirectory
* Fix numpydoc documentation problems in the Python code
* Establish a public API by adding `__init__.py` modifications.

Not perfect yet since, for example, `Quantum` does not show any methods. Also, currently every mention of `Dataset` is magically converted to a link to the h5py documentation stack. Probably still worth merging as is. Note that assemblers and formatters that depend on `afw` are not included in the documentation tree because they are effectively optional dependencies on `afw` for runtime usage but the sphinx system has to be able to import everything.

I'm not sure how we stop drift in the docs over time again as all the subtle numpydoc problems creep back.

@jonathansick can you take a look please?

### DIFF
--- a/python/lsst/daf/butler/core/__init__.py
+++ b/python/lsst/daf/butler/core/__init__.py
@@ -2,11 +2,15 @@
 Core code for butler.
 """
 
+# Do not export the utility routines from safeFileIo and utils
+
 from .composites import *
 from .config import *
 from .datasets import *
 from .datastore import *
+from .dataUnits import *
 from .fileDescriptor import *
+from .fileTemplates import *
 from .formatter import *
 from .location import *
 from .mappingFactory import *
@@ -14,8 +18,6 @@ from .quantum import *
 from .regions import *
 from .registry import *
 from .run import *
-from .safeFileIo import *
 from .schema import *
 from .storageClass import *
 from .units import *
-from .utils import *
--- a/python/lsst/daf/butler/core/__init__.py
+++ b/python/lsst/daf/butler/core/__init__.py
@@ -1,0 +1,21 @@
+"""
+Core code for butler.
+"""
+
+from .composites import *
+from .config import *
+from .datasets import *
+from .datastore import *
+from .fileDescriptor import *
+from .formatter import *
+from .location import *
+from .mappingFactory import *
+from .quantum import *
+from .regions import *
+from .registry import *
+from .run import *
+from .safeFileIo import *
+from .schema import *
+from .storageClass import *
+from .units import *
+from .utils import *
--- a/python/lsst/daf/butler/core/quantum.py
+++ b/python/lsst/daf/butler/core/quantum.py
@@ -19,13 +19,27 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+__all__ = ("Quantum",)
 
-class Quantum(object):
-    """A discrete unit of work that may depend on one or more Datasets and produces one or more `Dataset`s.
 
-    Most Quanta will be executions of a particular `SuperTask`’s `runQuantum` method,
-    but they can also be used to represent discrete units of work performed manually
-    by human operators or other software agents.
+class Quantum:
+    """A discrete unit of work that may depend on one or more Datasets and
+    produces one or more `Dataset`\ s.
+
+    Most Quanta will be executions of a particular `SuperTask`’s `runQuantum`
+    method, but they can also be used to represent discrete units of work
+    performed manually by human operators or other software agents.
+
+    Parameters
+    ----------
+    run : `Run`
+        Run this quantum is part of.
+    task : `SuperTask` or `str`, optional
+        Associated task information.
+    quantumId : ??, optional
+        ???
+    registryId : ??, optional
+        ???
     """
 
     __slots__ = ("_quantumId", "_registryId", "_run", "_task", "_predictedInputs", "_actualInputs")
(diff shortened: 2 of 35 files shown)


python/lsst/daf/butler/core/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+from .formatter import *
+from .location import *
+from .mappingFactory import *
+from .quantum import *
+from .regions import *
+from .registry import *
+from .run import *
+from .schema import *
+from .storageClass import *
+from .units import *
```
pschella, line 23: Are we sure we need to pull everything in?

python/lsst/daf/butler/core/quantum.py
```diff
@@ -19,13 +19,27 @@
...
+    Parameters
+    ----------
+    run : `Run`
+        Run this quantum is part of.
+    task : `SuperTask` or `str`, optional
+        Associated task information.
+    quantumId : ??, optional
+        ???
+    registryId : ??, optional
+        ???
```
pschella, line 42: Are question marks intentional?

python/lsst/daf/butler/core/registry.py
```diff
@@ -104,73 +107,74 @@ def addDataset(self, ref, uri, components, run, producer=None):
...
         Collection in the process.
 
         Parameters
         ----------
-        collection: `str`
-            Indicates the Collection the `Dataset`s should be associated with.
-        refs: `[DatasetRef]`
+        collection : `str`
+            Indicates the Collection the `Dataset`\ s should be associated with.
+        refs : `[DatasetRef]`
```
pschella: `list of `DatasetRef`?

python/lsst/daf/butler/core/run.py
```diff
@@ -21,8 +21,25 @@
...
 
+__all__ = ("Run", )
+
 
 class Run(object):
+    """Represent a processing run.
+
+    Parameters
+    ----------
+    runId : ??
```
pschella: int

python/lsst/daf/butler/core/run.py
```diff
@@ -21,8 +21,25 @@
...
+        ID to associate with this run.
+    registryId : ??
```
pschella: int

python/lsst/daf/butler/core/run.py
```diff
@@ -21,8 +21,25 @@
...
+        ID associated with this `Registry`.
+    collection : ??
```
pschella: string

python/lsst/daf/butler/core/run.py
```diff
@@ -21,8 +21,25 @@
...
+        Collection to use for this run.
+    environment : ??
```
pschella: string, I guess. Although later we may want to use actual `Dataset` objects for this.

python/lsst/daf/butler/core/run.py
```diff
@@ -21,8 +21,25 @@
...
+        Something about the environment.
+    pipeline : ??
```
pschella: string here also (with the same caveat).

python/lsst/daf/butler/core/units.py
```diff
@@ -22,6 +22,21 @@
...
+           "PhysicalFilter",
+           "PhysicalSensor",
+           "Visit",
+           "ObservedSensor",
+           "Snap",
+           "VisitRange",
+           "SkyMap",
+           "Tract",
+           "Patch")
+
```
pschella, line 39: This whole file is going away.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -110,40 +112,40 @@ def addDataset(self, ref, uri, components, run, producer=None):
...
-            If `True`, remove `Dataset`s from the `SqlRegistry` if they are not
+        remove : `bool`
+            If `True`, remove `Dataset`\ s from the `SqlRegistry` if they are not
             associated with any Collection (including via any composites).
 
         Returns
         -------
-        removed: `[DatasetRef]`
-            If `remove` is `True`, the `list` of `DatasetRef`s that were
+        removed : `[DatasetRef]`
```
pschella: list of `DatasetRef`?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -318,79 +322,79 @@ def merge(self, outputCollection, inputCollections):
...
-        `Dataset`s with the same `DatasetRef`.
+        `Dataset`\ s with the same `DatasetRef`.
 
         Parameters
         ----------
-        outputCollection: `str`
+        outputCollection : `str`
             collection to use for the new Collection.
-        inputCollections: `[str]`
+        inputCollections : `[str]`
```
pschella: List of `str`?

pschella (approved):

python/lsst/daf/butler/core/quantum.py
timj: They mean I have no idea what they are and need guidance.

python/lsst/daf/butler/core/registry.py
timj: Yes, I thought I had found them all but my regex clearly wasn't quite right. I'll have another pass.

python/lsst/daf/butler/core/__init__.py
timj: The big issue with the docs is working out what is public API and should be imported into `lsst.daf.butler` and which classes are entirely private API. You'll see that I didn't import the utility scripts but explicitly got those documented in a later section. This Sphinx stuff really does force us to work out public vs private.

python/lsst/daf/butler/core/units.py
timj: Ok. But for now it is there.

PR #19, opened 2018-03-12
DM-13452: Few small fixes needed for SuperTask/QuantumGraph

andy-slac:
Small fixes for things that I discovered while trying to implement
trivial tests on SuperTask side that depend on gen3 butler interfaces.

Note for reviewers: I'm not quite sure I did the right thing extending DataUnit classes with data/methods. So I did not update all DataUnit classes but only those that I needed for my SuperTask examples.

### DIFF
--- a/python/lsst/daf/butler/core/quantum.py
+++ b/python/lsst/daf/butler/core/quantum.py
@@ -42,7 +42,7 @@ class Quantum(Execution):
         The Run this Quantum is a part of.
     """
 
-    __slots__ = ("_task", "_run", "_predictedInputs", "_actualInputs")
+    __slots__ = ("_task", "_run", "_predictedInputs", "_actualInputs", "_outputs")
     __eq__ = slotValuesAreEqual
 
     def __init__(self, task, run, *args, **kwargs):
@@ -51,6 +51,7 @@ class Quantum(Execution):
         self._run = run
         self._predictedInputs = {}
         self._actualInputs = {}
+        self._outputs = {}
 
     @property
     def task(self):
@@ -93,11 +94,25 @@ class Quantum(Execution):
         """
         return self._actualInputs
 
+    @property
+    def outputs(self):
+        """A `dict` of output datasets (to be) generated for this quantum,
+        with the same form as `predictedInputs`.
+
+        Read-only; update via `addOutput()`.
+        """
+        return self._outputs
+
     def addPredictedInput(self, ref):
         """Add an input `DatasetRef` to the `Quantum`.
 
         This does not automatically update a `Registry`; all `predictedInputs`
         must be present before a `Registry.addQuantum()` is called.
+
+        Parameters
+        ----------
+        ref : `DatasetRef`
+            Reference for a Dataset to add to the Quantum's predicted inputs.
         """
         datasetTypeName = ref.datasetType.name
         if datasetTypeName not in self._actualInputs:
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/core/quantum.py
```diff
@@ -129,14 +131,32 @@ def actualInputs(self):
...
         if datasetTypeName not in self._actualInputs:
             self._predictedInputs[datasetTypeName] = [ref, ]
         else:
             self._predictedInputs[datasetTypeName].append(ref)
+
+    def addOutput(self, ref):
+        """Add an output `DatasetRef` to the `Quantum`.
+
+        This does not automatically update a `Registry`; all `outputs`
+        must be present before a `Registry.addQuantum()` is called.
```
timj, line 145: Needs `Parameters` section.

python/lsst/daf/butler/core/units.py
```diff
@@ -226,6 +226,13 @@ class Visit(DataUnit):
 
     dependencies = (Camera, PhysicalFilter)
 
+    def __init__(self, visitId):
+        self._visitId = visitId
```
timj, line 230: Please add class documentation.

python/lsst/daf/butler/core/units.py
```diff
@@ -226,6 +226,13 @@ class Visit(DataUnit):
...
+
+    @property
+    def visitId(self):
+        return self._visitId
```
timj, line 234: Docstring

python/lsst/daf/butler/core/units.py
```diff
@@ -246,12 +253,33 @@ class SkyMap(DataUnit):
 
     dependencies = ()
 
+    def __init__(self, skyMapId):
+        self._skyMapId = skyMapId
```
timj, line 257: Needs class documentation and property docstring.

python/lsst/daf/butler/core/units.py
```diff
@@ -246,12 +253,33 @@ class SkyMap(DataUnit):
...
+    def skyMapId(self):
+        return self._skyMapId
+
 
 class Tract(DataUnit):
 
     dependencies = (SkyMap,)
 
+    def __init__(self, tractId):
+        self._tractId = tractId
```
timj, line 269: Docs.

python/lsst/daf/butler/core/units.py
```diff
@@ -246,12 +253,33 @@ class SkyMap(DataUnit):
...
+    def tractId(self):
+        return self._tractId
+
 
 class Patch(DataUnit):
 
     dependencies = (SkyMap, Tract)
+
+    def __init__(self, patchId):
+        self._patchId = patchId
```
timj, line 281: Docs.

python/lsst/daf/butler/core/datasets.py
```diff
@@ -120,10 +120,10 @@ def getNewId(cls):
 
     def __init__(self, datasetType, units):
         units = datasetType.units.conform(units)
-        super().__init__(
-            datasetType.name,
-            **{unit.__class__.__name__: unit.value for unit in units}
-        )
+#         super().__init__(
```
timj: Is this a temporary removal or permanent (in which case either add a comment saying when it's coming back or delete it).

andy-slac: I was not sure why that code was there at all, so I wanted to check with @pschella . To me it looks like some left-over from previous versions of that code and it can be removed but again I'm not entirely sure.

python/lsst/daf/butler/core/quantum.py
andy-slac: Regarding this and all other docstring-related comments - I tried to do as minimal changes as I could, only what was necessary to run my dummy SuperTask example. Looking at the units code I'm not sure what Pim's plans for this are. Right now it looks more like a placeholder and I'd like to understand what is going to happen to it before spending more time on it.

timj: After last week's hack session I'm a bit sensitized to adding new code that doesn't include correct docstrings...

andy-slac: OK, I'll add something to `Quantum` methods, but `units` needs more work, there is more non-documented stuff there than what I added.

andy-slac:
I'm going to close this pull request, it's not going to be merged in its current state anyways. Branch is still sort of useful for my work on supertask, I'll keep it for now.

PR #20, opened 2018-03-15
DM-12627: Implement Butler DatasetType

pschella:


python/lsst/daf/butler/core/datasets.py
```diff
@@ -68,24 +80,11 @@ def template(self):
...
-        return self._units
-
-    @property
-    def storageClass(self):
-        """A `StorageClass` that defines how this `DatasetType` is persisted.
-        """
-        return self._storageClass
-
-    def __init__(self, name, template, units, storageClass):
+    def __init__(self, name, dataUnits, storageClass, template=None):
```
timj: Is the `dataUnits` argument here a `frozenset` and you copy it in the constructor?

python/lsst/daf/butler/core/datasets.py
```diff
@@ -46,7 +45,7 @@ class DatasetType(object):
     All arguments correspond directly to instance attributes.
```
timj: Could really do with explicit Attributes and Parameters section.

python/lsst/daf/butler/core/datasets.py
```diff
@@ -57,6 +56,19 @@ def name(self):
...
+
+    @property
+    def storageClass(self):
+        """A `StorageClass` that defines how this `DatasetType` is persisted.
+        """
+        return self._storageClass
+
     @property
     def template(self):
         """A string with `str`.format-style replacement patterns that can be
```
timj: This is not related to the file name template used by datastore. Correct? (so this template is not a `FileTemplate`).

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -56,7 +69,21 @@ def registerDatasetType(self, datasetType):
...
-        raise NotImplementedError("Must be implemented by subclass")
+        if not self._isValidDatasetType(datasetType):
+            raise ValueError("DatasetType is not valid for this registry")
+        if datasetType.name in self._datasetTypes:
+            raise KeyError("DatasetType: {} already registered".format(datasetType.name))
+        datasetTypeTable = self._schema.metadata.tables['DatasetType']
+        datasetTypeUnitsTable = self._schema.metadata.tables['DatasetTypeUnits']
+        with self._engine.begin() as connection:
+            connection.execute(datasetTypeTable.insert().values(dataset_type_name=datasetType.name,
+                                                                storage_class=datasetType.storageClass,
```
timj: `StorageClass` doesn't have a `__str__` method so what gets inserted into the table? Does it magically use `StorageClass.name`? Is it the full class name? StorageClasses are defined by the datastore so we have to work out how to verify that a registry is not specifying storage class names that a datastore doesn't know about.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -71,7 +98,28 @@ def getDatasetType(self, name):
...
+            datasetType = self._datasetTypes[name]
+        else:
+            datasetTypeTable = self._schema.metadata.tables['DatasetType']
+            datasetTypeUnitsTable = self._schema.metadata.tables['DatasetTypeUnits']
+            with self._engine.begin() as connection:
+                # Get StorageClass and template from DatasetType table
+                result = connection.execute(select([datasetTypeTable.c.storage_class,
+                                                    datasetTypeTable.c.template]).where(
+                    datasetTypeTable.c.dataset_type_name == name)).fetchone()
+                storageClass = result['storage_class']
```
timj: But `DatasetType.storageClass` is documented to be a `StorageClass` not a `str`.

tests/test_datasetType.py
```diff
@@ -0,0 +1,93 @@
...
+from lsst.daf.butler.core.datasets import DatasetType
+
+"""Tests for SqlRegistry.
+"""
+
+
+class DatasetTypeTestCase(lsst.utils.tests.TestCase):
+    """Test for DatasetType.
+    """
+    def setUp(self):
```
timj: No need to define it if it doesn't do anything.

tests/test_datasetType.py
```diff
@@ -0,0 +1,93 @@
...
+        """Test construction preserves values.
+
+        Note that construction doesn't check for valid storageClass,
+        dataUnits or template parameters.
+        These can only be verified for a particular schema.
+        """
+        datasetTypeName = "test"
+        storageClass = "StructuredData"
+        dataUnits = frozenset(("camera", "visit"))
+        template = "{datasetType}/{camera}/{visit}"
```
timj: I'm worried that we have two distinct template systems now.

tests/test_datasetType.py
```diff
@@ -0,0 +1,93 @@
...
+                            DatasetType("b", "StorageA", ("UnitA", ), "{UnitA}"))
+        self.assertNotEqual(DatasetType("a", "StorageA", ("UnitA", ), "{UnitA}"),
+                            DatasetType("a", "StorageB", ("UnitA", ), "{UnitA}"))
+        self.assertNotEqual(DatasetType("a", "StorageA", ("UnitA", ), "{UnitA}"),
+                            DatasetType("a", "StorageA", ("UnitB", ), "{UnitA}"))
+        self.assertNotEqual(DatasetType("a", "StorageA", ("UnitA", ), "{UnitA}"),
+                            DatasetType("a", "StorageA", ("UnitA", ), "{UnitB}"))
+
+    def testHashability(self):
+        types = []
```
timj: I think this test method needs a docstring to explain what is actually being tested. There's a lot going on here before the final conversion of the list to a set.

timj (approved):
Looks okay. I mainly have questions about template vs FileTemplate and StorageClass vs a string (you document it to be a class but tests use it as a string).

python/lsst/daf/butler/core/datasets.py
pschella: I thought it was policy to document these in the property docstrings?

python/lsst/daf/butler/core/datasets.py
pschella: It is related to that template. But it is unclear (to me at the moment) if we still want it here. I'll remove it for now.

python/lsst/daf/butler/core/datasets.py
pschella: It may be any iterable, but it becomes a `frozenset` in the constructor.

python/lsst/daf/butler/registries/sqlRegistry.py
pschella: It is a name of a `StorageClass`. Not an instance.

python/lsst/daf/butler/registries/sqlRegistry.py
pschella: I'll change the docstring. Should be a name (for now at least).

tests/test_datasetType.py
pschella: It should really be the same, but I'll remove it since this seems to be just a remnant of the old system.

python/lsst/daf/butler/core/datasets.py
timj: Yes. I thought that in this case there was a difference between init parameters and attributes (on phone so not checking).

PR #21, opened 2018-03-19
DM-13840: Prepare Butler for composite work to begin

pschella:
Please ignore changes to `default_schema.yaml` which are mostly cherry picked from @TallJimbo DM-12620 and will be rebased out later.

### DIFF
--- a/python/lsst/daf/butler/butler.py
+++ b/python/lsst/daf/butler/butler.py
@@ -82,7 +82,7 @@ class Butler(object):
             An identifier with `DataUnit` names and values.
         producer : `Quantum`, optional
             The producer.
-        
+
         Returns
         -------
         `DatasetRef`
@@ -99,7 +99,7 @@ class Butler(object):
             The `DatasetType`.
         dataId : `dict`
             An identifier with `DataUnit` names and values.
-        
+
         Returns
         -------
         `object`
--- a/python/lsst/daf/butler/butler.py
+++ b/python/lsst/daf/butler/butler.py
@@ -88,7 +88,10 @@ class Butler(object):
         `DatasetRef`
             A reference to the stored dataset.
         """
-        raise NotImplementedError("Not yet implemented")
+        datasetType = self.registry.getDatasetType(datasetType)
+        ref = self.registry.addDataset(datasetType, dataId, run=self.run, producer=producer)
+        self.datastore.put(obj, ref)
+        return ref
 
     def get(self, datasetType, dataId):
         """Retrieve a stored dataset.
--- a/python/lsst/daf/butler/butler.py
+++ b/python/lsst/daf/butler/butler.py
@@ -69,104 +69,40 @@ class Butler(object):
         if self.run is None:
             self.run = self.registry.makeRun(self.config['run'])
 
-    def getDirect(self, ref, parameters=None):
-        """Load a `Dataset` or a slice thereof from a `DatasetRef`.
-
-        Unlike `Butler.get`, this method allows `Datasets` outside the Butler's `Collection` to be read as
-        long as the `DatasetRef` that identifies them can be obtained separately.
+    def put(self, obj, datasetType, dataId, producer=None):
+        """Store and register a dataset.
 
         Parameters
         ----------
-        ref : `DatasetRef`
-            A pointer to the `Dataset` to load.
-        parameters : `dict`
-            `StorageClass`-specific parameters that can be used to obtain a slice of the `Dataset`.
-
+        obj : `object`
+            The dataset.
+        datasetType : `DatasetType` instance or `str`
+            The `DatasetType`.
+        dataId : `dict`
+            An identifier with `DataUnit` names and values.
+        producer : `Quantum`, optional
+            The producer.
+        
         Returns
         -------
-        inMemoryDataset : `InMemoryDataset`
-            The requested `Dataset`.
+        `DatasetRef`
+            A reference to the stored dataset.
         """
-        parent = self.datastore.get(ref.uri, ref.datasetType.storageClass, parameters) if ref.uri else None
-        children = {name: self.datastore.get(childRef, parameters)
-                    for name, childRef in ref.components.items()}
-        return ref.datasetType.storageClass.assemble(parent, children)
+        raise NotImplementedError("Not yet implemented")
 
-    def get(self, ref, parameters=None):
-        """Load a `Dataset` or a slice thereof from the Butler's `Collection`.
+    def get(self, datasetType, dataId):
+        """Retrieve a stored dataset.
 
         Parameters
         ----------
-        ref : `DatasetRef`
-            The `Dataset` to retrieve.
-        parameters : `dict`
-            A dictionary of `StorageClass`-specific parameters that can be
-            used to obtain a slice of the `Dataset`.
-
+        datasetType : `DatasetType` instance or `str`
+            The `DatasetType`.
+        dataId : `dict`
+            An identifier with `DataUnit` names and values.
+        
         Returns
         -------
-        dataset : `InMemoryDataset`
-            The requested `Dataset`.
+        `object`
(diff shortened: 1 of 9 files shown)


python/lsst/daf/butler/butler.py
```diff
@@ -69,104 +69,65 @@ def __init__(self, config):
...
-        ref : `DatasetRef`
-            A pointer to the `Dataset` to load.
-        parameters : `dict`
-            `StorageClass`-specific parameters that can be used to obtain a slice of the `Dataset`.
+        obj : `object`
+            The dataset.
+        datasetType : `DatasetType` instance or `str`
+            The `DatasetType`.
+        dataId : `dict`
+            An identifier with `DataUnit` names and values.
```
timj, line 82: Are you referring to the class here because the keys are names that convert to instances of `DataUnit` classes?

python/lsst/daf/butler/butler.py
```diff
@@ -69,104 +69,65 @@ def __init__(self, config):
...
+        producer : `Quantum`, optional
+            The producer.
 
         Returns
         -------
-        inMemoryDataset : `InMemoryDataset`
-            The requested `Dataset`.
+        `DatasetRef`
```
timj: Needs a variable name since `Returns` sections are meant to be identical to `Parameters` sections. I won't repeat the comment in each case.

python/lsst/daf/butler/butler.py
```diff
@@ -69,104 +69,65 @@ def __init__(self, config):
...
-        ----------
-        refs : `list` of `DatasetRef`
-            List of refs for `Dataset`\ s to unlink.
+        `object`
+            The dataset.
         """
-        refs = [self.registry.find(self.run.collection, ref) for ref in refs]
-        for ref in self.registry.disassociate(self.run.collection, refs, remove=True):
-            self.datastore.remove(ref.uri)
+        datasetType = self.registry.getDatasetType(datasetType)
```
timj, line 132: `getDatasetType` only takes a `str` but this method is documented to provide either a `str` or a `DatasetType`.

python/lsst/daf/butler/core/datasets.py
```diff
@@ -96,51 +96,50 @@ class DatasetRef(object):
     ----------
     datasetType : `DatasetType`
         The `DatasetType` for this `Dataset`.
-    units : `dict`
+    dataId : `dict`
```
timj, line 99: This documentation for a `dataId` is more explicit than those elsewhere.

python/lsst/daf/butler/core/datasets.py
```diff
@@ -173,14 +172,20 @@ def actualConsumers(self):
...
+    def components(self):
+        """Named `DatasetRef` components.
+
+        Read-only; update via `Registry.attachComponent()`.
+        """
+        return _safeMakeMappingProxyType(self._components)
+
+    @property
+    def assembler(self):
+        """Fully-qualified name of an importable Assembler object that can be
```
timj, line 185: Remember that assemblers are classes with assemble and disassemble methods. You have to store the assembler class name, then to assemble you create an instance and run the `assemble` method.

python/lsst/daf/butler/core/run.py
```diff
@@ -29,46 +29,32 @@ class Run(object):
...
-        Collection to use for this run.
-    environment : `str`
-        Something about the environment.
-    pipeline : `str`
-        Something about the pipeline.
+        A Collection name with which all Datasets in this Run are initially
+        associated, also used as a human-readable name for this Run.
+    environment : `int`
+        A Dataset that contains a description of
+        the software environment (e.g. versions) used for this Run.
```
timj: ie a Dataset that you can `butler.get()` using this integer? The contents of which are arbitrary?

python/lsst/daf/butler/core/run.py
```diff
@@ -29,46 +29,32 @@ class Run(object):
...
+    pipeline : `int`
+        A Dataset that contains a serialization of
```
timj: Same comment as for environment.

tests/test_sqlRegistry.py
```diff
@@ -67,6 +68,50 @@ def testDatasetType(self):
...
+        self.assertIsNone(registry.getRun(id=100))
+
+    def testAssembler(self):
+        registry = Registry.fromConfig(self.configFile)
+        datasetType = DatasetType(name="test", dataUnits=("camera",), storageClass="dummy")
+        registry.registerDatasetType(datasetType)
+        run = registry.makeRun(collection="test")
+        ref = registry.addDataset(datasetType, dataId={"camera": "DummyCam"}, run=run)
+        self.assertIsNone(ref.assembler)
+        assembler = "some.fully.qualified.assembler"  # TODO replace by actual dummy assember once implemented
```
timj, line 110: We already have assemblers that should work for your testing.

timj (approved):
Looks okay. How are you expecting storage class name to be mapped to an actual class? Is that still expected to be done inside datastore?

python/lsst/daf/butler/core/datasets.py
pschella: That is because it is wrong. In the new design `DataUnit`s will be visible as `dict`s (because the instances of `DataUnit` represent tables, not rows therein).

python/lsst/daf/butler/butler.py
pschella: It should probably be pass-through (or perhaps fill in details) for `DatasetType` instances.

python/lsst/daf/butler/butler.py
pschella: No. They are names of the tables. (e.g. `{"camera" : "HSC", "visit" : 3}`).

tests/test_sqlRegistry.py
pschella: I thought that (re)designing assemblers/dissasemblers would be part of the `Datastore` work.

python/lsst/daf/butler/core/run.py
pschella: In that case probably `DatasetRef` instances. But I'm not sure.

python/lsst/daf/butler/core/run.py
pschella: Same answer ;)

tests/test_sqlRegistry.py
timj: I wasn't foreseeing any of the assembler/disassembler classes having to change. They all seemed to be entirely consistent with your redesign.

python/lsst/daf/butler/core/datasets.py
pschella: Not according to https://confluence.lsstcorp.org/display/DM/Gen3+Butler+Composites+Design
But I'm perfectly happy for them to be so. This is part of the composite work to be done on a different ticket.

timj: When you asked me to combine free functions into classes the code got significantly cleaner. I'll be surprised if we gain by pulling everything apart again. I was really happy with the way assembler/disassembler turned out.

timj: I must admit that I hadn't taken that confluence page as gospel. I thought it was guiding principles so I haven't gone into edit it with my thoughts.

python/lsst/daf/butler/core/run.py
pschella: Yes, they should be `DatasetRef`.

python/lsst/daf/butler/core/datasets.py
TallJimbo: The code on the confluence page is absolutely intended as just pseudocode.  You're both very much encouraged to actively rethink all of it (the code parts, that is; I hope the conceptual stuff will actually stick this time around).

PR #22, opened 2018-03-19
DM-13842: Implement access to DatasetStorage table

pschella:
Will be rebased to `master` after DM-13840 is merged.

### DIFF
--- a/python/lsst/daf/butler/core/storageInfo.py
+++ b/python/lsst/daf/butler/core/storageInfo.py
@@ -1,0 +1,66 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from .utils import slotValuesAreEqual
+
+__all__ = ("StorageInfo", )
+
+
+class StorageInfo:
+    """Represents storage information.
+
+    Parameters
+    ----------
+    datastoreName : `str`
+        Name of datastore.
+    md5 : `str`
+        Checksum.
+    size : `int`
+        Size of stored object in bytes.
+    """
+    __eq__ = slotValuesAreEqual
+    __slots__ = ("_datastoreName", "_md5", "_size")
+
+    def __init__(self, datastoreName, md5=None, size=None):
+        assert isinstance(datastoreName, str)
+        self._datastoreName = datastoreName
+        assert md5 is None or isinstance(md5, str)
+        self._md5 = md5
+        assert size is None or isinstance(size, int)
+        self._size = size
+
+    @property
+    def datastoreName(self):
+        """Name of datastore (`str`).
+        """
+        return self._datastoreName
+
+    @property
+    def md5(self):
+        """Checksum (`str`).
+        """
+        return self._md5
+
+    @property
+    def size(self):
+        """Size of stored object in bytes (`int`).
+        """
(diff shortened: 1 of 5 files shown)


python/lsst/daf/butler/core/storageInfo.py
```diff
@@ -0,0 +1,51 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from .utils import slotValuesAreEqual
+
+__all__ = ("StorageInfo")
```
timj: You need the `,` to stop this being a `str`.

python/lsst/daf/butler/core/storageInfo.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class StorageInfo:
+    """Represents storage information.
+    """
```
timj: `Parameters` docstring required. Need to know the type of `md5` and the units of `size`, what do we mean by a datastore name?

python/lsst/daf/butler/core/storageInfo.py
```diff
@@ -0,0 +1,51 @@
...
+    __eq__ = slotValuesAreEqual
+    __slots__ = ("_datastoreName", "_md5", "_size")
+
+    def __init__(self, datastoreName, md5=None, size=None):
+        assert isinstance(datastoreName, str)
+        self._datastoreName = datastoreName
+        assert md5 is None or isinstance(md5, str)
```
timj: Are we using md5 because it is ubiquitous? Should it be datastore specific or must it be defined by gen3 butler to be md5? blake2 is meant to be as fast as md5 but better (and is part of python3.6). Does it need to be md5 because @MichelleGower requires that from NCSA?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -314,6 +315,79 @@ def getRun(self, collection=None, id=None):
...
+        datastoreName : `str`
+            What datastore association to update.
+        storageInfo : `StorageInfo`
+            Storage information about the dataset.
+        """
+        datasetStorageTable = self._schema.metadata.tables['DatasetStorage']
+        with self._engine.begin() as connection:
+            connection.execute(datasetStorageTable.update().where(and_(
+                datasetStorageTable.c.dataset_id == ref.id,
+                datasetStorageTable.c.datastore_name == datastoreName)).values(
```
timj, line 361: I'm confused by the relationship between `datastoreName` and `storageInfo.datastoreName`. This may partly be because I'm not sure what a datastore name really is. In this code you update the row that has the supplied `datastoreName` by changing it to be the name stored in the storage info?

timj (approved):
Looks good. Some minor comments and tweaks.

python/lsst/daf/butler/core/storageInfo.py
pschella: We are using `md5` because that is what @TallJimbo put in the schema description. But it could/should probably be `checksum`.

python/lsst/daf/butler/registries/sqlRegistry.py
pschella: `datastoreName` is the identifier of the particular `Datastore` used to save the file (i.e. we haven't decided what that should be yet). The difference between `datastoreName` and `storageInfo.datastoreName` in `update` is that we may want to change the `datastoreName` to a new one for an existing entry. Thus we need to supply both the current datastore name (part of the primary key) to look it up (this is `datastoreName`) and the new one which is part of `storageInfo` just as everything else we want to update is.

pschella: By "we haven't decided yet" I mean that I don't really know off the top of my head. But @TallJimbo probably has some Idea.

pschella: The comment in the schema just says: "Name of the Datastore this entry corresponds to.".

python/lsst/daf/butler/core/storageInfo.py
TallJimbo: I have no preference as to what kind of checksum we use.  Probably best to make the name generic.

python/lsst/daf/butler/registries/sqlRegistry.py
TallJimbo: Registry definitely needs some kind of identifier for each associated Datastore.  I added a string column for this to the Registry schema for this and called it `datastore_name` in SQL, but I would have no objection to making it an integer (assuming we have an approach for assigning them) or changing the name.

PR #23, opened 2018-03-22
DM-13869: Implement basic Registry.find and related functionality

pschella:


timj:
I'd be wary of running autopep8 on the entire code base when we know we are already flake8 clean.

timj (approved):
I think this is fine. I'd like us to stop running autopep8 in the future given that we already pass flake8. DM-13869 says it is meant to be working on Butler.get but I don't think this is (and I'm working on that).

pschella:
Running autopep8 is my lazy way of being flake8 compatible (at least for the multi-line sql annoyance). But I'll stop running it on a non-file-to-file basis now.
As for DM-13869, it doesn't work for `Butler.get` (and `put`) because it assumes a `Datastore` interface that doesn't exist yet. But it shouldn't not work because something is missing in `Registry`. If that is not the case there is a bug in my code.

timj:
Can you enable flake8 checking in your editor? It's just I'm worried about autopep8 fiddling with things that currently are fine.

PR #26, opened 2018-03-28
DM-13814: Revamp how datastores and composites work

timj:
Following the new scheme, DatasetRefs are now the means by which a datastore get/put work. Complete reorganization of how that happens and composite disassembly now happens at the butler level. All tests now pass. Currently uses a transient in-memory store in Datastore for keeping track of formatter that was used for a dataset.

Still need to add tests for butler put/get.


python/lsst/daf/butler/assemblers/exposureAssembler.py
```diff
@@ -159,6 +160,50 @@ def disassemble(self, composite):
...
+            All the components from which to construct the Exposure.
+            Some can be missing.
+
+        Returns
+        -------
+        exposure : `~lsst.afw.image.Exposure`
+            Assembled exposure.
+        """
+        micomps = {}
+        hasMI = False
```
pschella: Prefer longer names: `maskedImageComponents`. Annoying to type, but easier to read.

python/lsst/daf/butler/assemblers/exposureAssembler.py
```diff
@@ -159,6 +160,50 @@ def disassemble(self, composite):
...
+        for c in ("image", "variance", "mask"):
+            v = None
+            if c in components:
+                hasMI = True
+                v = components[c]
+            micomps[c] = v
```
pschella: Here too: `c -> component`, `v -> value`.

python/lsst/daf/butler/assemblers/exposureAssembler.py
```diff
@@ -159,6 +160,50 @@ def disassemble(self, composite):
...
+
+            if not isinstance(exposure, pytype):
+                raise RuntimeError("Internal inconsistency in Exposure assembly")
+
+        else:
+            exposure = pytype()
+            if wcs is not None:
+                exposure.setWcs(wcs)
+
+        # Set other components
```
pschella, line 210: ?

python/lsst/daf/butler/assemblers/exposureAssembler.py
```diff
@@ -159,6 +160,50 @@ def disassemble(self, composite):
...
+        if "wcs" in components:
+            wcs = components["wcs"]
+
+        pytype = self.storageClass.pytype
+        if hasMI:
+            maskedImage = makeMaskedImage(**micomps)
+            exposure = makeExposure(maskedImage, wcs=wcs)
+
+            if not isinstance(exposure, pytype):
+                raise RuntimeError("Internal inconsistency in Exposure assembly")
```
pschella: Not a very descriptive error message.

python/lsst/daf/butler/butler.py
```diff
@@ -91,7 +94,30 @@ def put(self, obj, datasetType, dataId, producer=None):
...
-        # self.datastore.put(obj, ref)
+
+        # Look up storage class in factory
+        storageClass = self.storageClasses.getStorageClass(datasetType.storageClass)
+
+        # Check to see if this storage class has a disassembler
+        # and also has components
+        if storageClass.assemblerClass.disassemble is not None and storageClass.components:
+            components = storageClass.assembler().disassemble(obj)
+            if components:  # Probably need to always have all components
```
pschella: I would think so.

python/lsst/daf/butler/butler.py
```diff
@@ -91,7 +94,30 @@ def put(self, obj, datasetType, dataId, producer=None):
...
+                for comp, info in components.items():
+                    compTypeName = "{}.{}".format(datasetType.name, comp)
+
+                    compRef = self.put(info.component, compTypeName, dataId, producer)
```
pschella: No run?

python/lsst/daf/butler/butler.py
```diff
@@ -110,9 +136,22 @@ def getDirect(self, ref):
         obj : `object`
             The dataset.
         """
-        # Currently a direct pass-through to `Datastore.get` but this should
-        # change for composites.
-        return self.datastore.get(ref)
+        print("Ref: {}".format(ref))
```
pschella: Leftover?

python/lsst/daf/butler/core/composites.py
```diff
@@ -137,6 +141,7 @@ def assemble(self, components, pytype=None):
 
             failed = []
             for name, component in components.items():
+                print("{}: {} - {}".format(obj, name, component))
```
pschella: Leftover?

python/lsst/daf/butler/core/composites.py
```diff
@@ -145,6 +150,7 @@ def assemble(self, components, pytype=None):
                             setattr(obj, attr, component)
                         else:
                             setter = getattr(obj, attr)
+                            print("Trying setter: {}".format(setter))
```
pschella: ?

python/lsst/daf/butler/core/datasets.py
```diff
@@ -32,7 +33,7 @@ def _safeMakeMappingProxyType(data):
 
 
 class DatasetType(object):
-    """A named category of Datasets that defines how they are organized,
+    r"""A named category of Datasets that defines how they are organized,
```
pschella: Why raw? Here and below.

python/lsst/daf/butler/core/datasets.py
```diff
@@ -190,3 +209,10 @@ def assembler(self):
         Read-only; update via `Registry.setAssembler()`.
         """
         return self._assembler
+
+    def __str__(self):
+        comp = ""
+        if self.components:
+            comp = ", components=[" + ", ".join(self.components) + "]"
+        return "DatasetRef({}, id={}, dataId={} {})".format(self.datasetType.name,
+                                                            self.id, self.dataId, comp)
```
pschella: I don't really like abbreviating `components` or `component` with `comp`. Confusion about plurality and about comparison.

python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -133,7 +133,9 @@ def format(self, dataUnits, datasetType=None, component=None):
             Or, `component` is specified but "component" was not part of
             the template.
         """
-        fields = dataUnits.definedUnits()
+        # Extract defined non-None units from the dataId
+        print(dataId)
```
pschella: ?

python/lsst/daf/butler/core/formatter.py
```diff
@@ -30,6 +30,13 @@ class Formatter(metaclass=ABCMeta):
     """Interface for reading and writing `Dataset`\ s with a particular
     `StorageClass`.
     """
+
+    @classmethod
+    def name(cls):
+        """Returns the fully qualified name of the formatter.
+        """
+        return cls.__module__ + "." + cls.__qualname__
```
pschella: Sounds like a job for a general utility function?

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -46,13 +50,20 @@ class StorageClass:
...
     # We set a default assembler so that a class is guaranteed to support
     # something.
     _pytype = None
     _assembler = CompositeAssembler
 
+    # @property
+    # def name(self):
+    #    """Name of the StorageClass (`str`)
+    #    """
+    #    return self._name
```
pschella: ?

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -147,12 +167,91 @@ def makeNewStorageClass(name, pytype=None, components=None, assembler=None):
...
+            ``storageClasses`` key.
+        """
+        sconfig = StorageClassConfig(config)['storageClasses']
+        self._configs.append(sconfig)
+
+        for name, info in sconfig.items():
+            if name == "config" or (isinstance(info, str) and info.endswith(".yaml")):
+                # This seems to be a location of another file so process that
+                self.addFromConfig(sconfig["config"])
+                continue
```
pschella, line 231: This looks like duplicate from `Config`.

python/lsst/daf/butler/core/utils.py
```diff
@@ -147,3 +147,31 @@ def doImport(pythonType):
...
+
+class Singleton(type):
+    """Metaclass to convert a class to a Singleton."""
+
+    _instances = {}
+
+    def __call__(cls, *args, **kwargs):  # noqa N805
+        if cls not in cls._instances:
+            cls._instances[cls] = super(Singleton, cls).__call__(*args, **kwargs)
+        return cls._instances[cls]
```
pschella: What does this do about multiple instances generated with different arguments? Now you confusingly get back an instance constructed with different arguments then you supplied.

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -112,127 +186,203 @@ def get(self, uri, storageClass, parameters=None):
...
+
+        """
+
+        # if this has never been written then we have to guess
+        if ref.id is None:
+            datasetType = ref.datasetType
+            typeName = datasetType.name
+            template = self.templates.getTemplate(typeName)
+            location = self.locationFactory.fromPath(template.format(ref.dataId,
+                                                     datasetType=typeName) + "#predicted")
```
pschella: I assume we don't use `#component` anymore? So this won't clash?

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -112,127 +186,203 @@ def get(self, uri, storageClass, parameters=None):
...
+            Reference to the required Dataset in the input data store.
+
+        """
+        assert inputDatastore is not self  # unless we want it for renames?
+        inMemoryDataset = inputDatastore.get(ref)
+        return self.put(inMemoryDataset, ref)
+
+    @staticmethod
+    def calcChecksum(filename, algorithm="blake2b", block_size=8192):
+        """Calculate the checksum of the supplied file.
```
pschella: Is `calc` our convention for this?

tests/test_butler.py
```diff
@@ -19,54 +19,103 @@
...
+        # Test getDirect
+        metricOut = butler.getDirect(ref)
+        self.assertEqual(metric, metricOut)
+        # Test get
+        metricOut = butler.get(datasetTypeName, dataId)
+        self.assertEqual(metric, metricOut)
+
+        # Get a component
+        summary = butler.get("{}.{}".format(datasetTypeName, "summary"), dataId)
+        self.assertEqual(summary, metric.summary)
```
pschella: What about the other components?

pschella (commented):
Looks ok, modulo comments. But I'd like a second look at it by Jim

python/lsst/daf/butler/assemblers/exposureAssembler.py
timj: I was wondering if something in `obs_base` already knows how to create an exposure from its components so I hadn't finished populating it. Maybe @TallJimbo  knows.

python/lsst/daf/butler/butler.py
timj: This gets back to the "if a component is defined to be None, does it really exist". In this case I think disassembly must always return something unless it's a completely blank Exposure.

python/lsst/daf/butler/butler.py
timj: there are a few debug prints lying around. I'll removed them when I squash everything.

python/lsst/daf/butler/core/datasets.py
timj: Because `\ ` is not an allowed escape sequence in a python string.

python/lsst/daf/butler/core/utils.py
timj: you only ever get the first thing you constructed back. The `**kwargs` are there to allow this metaclass to be generally useful. I never call the `StorageClassFactory` constructor with arguments now because it no longer makes any sense because you don't know if you are the first one. I could remove the arguments from this definition if you want to force our uses of `Singleton` to only work if there are no arguments.

python/lsst/daf/butler/datastores/posixDatastore.py
timj: Correct. I should remove that code from `Location`.

python/lsst/daf/butler/datastores/posixDatastore.py
timj: I don't know. `computeChecksum`?

tests/test_butler.py
timj: ok. I'll use a for loop and `getattr`.

python/lsst/daf/butler/butler.py
timj: run is not an argument to `butler.put`. `run` seems to be set per-butler.

python/lsst/daf/butler/core/storageClass.py
timj: I don't know. I wrote that from scratch. Does `Config` know that a `storageClasses` entry in a config can recursively load another config file if it sees the key `config`?

python/lsst/daf/butler/butler.py
pschella: Ah yes, I got confused with `addDataset` above.

python/lsst/daf/butler/core/storageClass.py
pschella: No, but it does have some check for a YAML file argument to its constructor that then builds it.

python/lsst/daf/butler/core/utils.py
pschella: Yes, I think that would be safer.

python/lsst/daf/butler/core/storageClass.py
timj: I don't want to create a storageClass called `extraStorageClassesHere.yaml` and I thought that you'd want the option of specifying multiple sources of extra definitions.

python/lsst/daf/butler/datastores/posixDatastore.py
pschella: See rule 3-26. Abbreviations in names SHOULD be avoided in https://developer.lsst.io/cpp/style.html (which, given that it is about naming) I assume also applies to Python. `compute` is a specific example.

python/lsst/daf/butler/core/datasets.py
```diff
@@ -50,15 +51,38 @@ class DatasetType(object):
...
 
     __slots__ = ("_name", "_dataUnits", "_storageClass")
     __eq__ = slotValuesAreEqual
     __hash__ = slotValuesToHash
 
+    def __str__(self):
+        return "DatasetType({}, {}, {})".format(self.name, self.storageClass.name, self.dataUnits)
+
+    @staticmethod
+    def nameWithComponent(datasetTypeName, componentName):
```
timj, line 67: @pschella you might have an opinion on `nameWithComponent`, `component` and `componentTypeName` in `DatasetType`. They contain the logic for "calexp.wcs" naming convention so I wanted it to be in the one place in this class.

python/lsst/daf/butler/assemblers/exposureAssembler.py
TallJimbo: No, nothing like this already exists.  The afw APIs probably ought to be updated someday, so you can just pass kwargs with components to the `Exposure` constructor.

python/lsst/daf/butler/assemblers/exposureAssembler.py
```diff
@@ -159,6 +160,74 @@ def disassemble(self, composite):
...
+        if "calib" in components:
+            exposure.setCalib(components.pop("calib"))
+
+        info = exposure.getInfo()
+        if "visitInfo" in components:
+            info.setVisitInfo(components.pop("visitInfo"))
+        if "apCorrMap" in components:
+            info.setApCorrMap(components.pop("calib"))
+        if "coaddInputs" in components:
+            info.setCoaddInputs(components.pop("coaddInputs"))
```
TallJimbo: Not necessarily advocating this, but if you're interested, *most* of these `ExposureInfo` setters can take `None` (I think `VisitInfo` may be the only exception), so you could use a pattern like
```
info.setX(components.pop("x", None))
```

python/lsst/daf/butler/butler.py
```diff
@@ -91,7 +94,27 @@ def put(self, obj, datasetType, dataId, producer=None):
         """
         datasetType = self.registry.getDatasetType(datasetType)
         ref = self.registry.addDataset(datasetType, dataId, run=self.run, producer=producer)
-        # self.datastore.put(obj, ref)
+
+        # Look up storage class to see if this is a composite
+        storageClass = datasetType.storageClass
+
+        # Check to see if this storage class has a disassembler
+        if storageClass.assemblerClass.disassemble is not None and storageClass.components:
```
TallJimbo, line 102: Whether this `disassemble` method is `None` depends on the `Butler` config, right?  How does that work if the `storageClass` instance is being retrieved from the Registry?  (I'd have thought the `assembler` couldn't be attached to the `storageClass` for this reason, or if it was, it'd have to be attached by `Butler` itself, and then only if it knows it's the only `Butler` using that particular `StorageClass` instance).

That said, what we need from the config probably isn't the disassembler itself (since that should always be the same, which of course why it's nice to attach it to `StorageClass` intrinsically) - we just need a flag in the config that says whether we should use it, to use in this if statement.

(Or maybe I just don't understand how it works, of course).

python/lsst/daf/butler/core/storedFileInfo.py
```diff
@@ -0,0 +1,76 @@
...
+
+    Parameters
+    ----------
+    formatter : `str` or `Formatter`
+        Full name of formatter to use to read this Dataset or a `Formatter`
+        instance.
+    path : `str`
+        Path to Dataset, relative to `Datastore` root.
+    storageClass : `StorageClass`
+        `StorageClass` used when writing the file.
```
TallJimbo: I think I understand how this works from our conversation yesterday, but this would be a good place to describe how/why/when it can differ from the `StorageClass` used to read the file.

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -44,66 +69,115 @@ class PosixDatastore(Datastore):
...
+        Returns
+        -------
+        exists : `bool`
+            `True` if the entity exists in the `Datastore`.
+        """
+        # Get the file information (this will fail if no file)
+        try:
+            storedFileInfo = self.getStoredFileInfo(ref)
+        except KeyError:
+            return False
```
TallJimbo, line 175: This currently returns `False` for virtual composites.  I think that *might* be a problem; we definitely need a way to distinguish that state from "it used to exist, but has been deleted", which I don't think we have right now, but the solution doesn't have to be having this return `True` for virtual composites.

python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -106,13 +106,13 @@ class FileTemplate:
     def __init__(self, template):
         self.template = template
 
-    def format(self, dataUnits, datasetType=None, component=None):
+    def format(self, dataId, datasetType=None, component=None):
```
TallJimbo: I think this may not need a `component` argument anymore, since that's packed into the `datasetType`.

In the future, it will probably need to have a `run` argument, as we'll need templates to be unique across runs as well.

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -112,127 +186,200 @@ def get(self, uri, storageClass, parameters=None):
...
+            Dataset does not exist in the datastore, the URI may be a guess
+            and will include a URI fragment "#predicted".
+            If the datastore does not have entities that relate well
+            to the concept of a URI the returned URI string will be
+            descriptive. The returned URI is not guaranteed to be obtainable.
+
+        """
+
+        # if this has never been written then we have to guess
+        if ref.id is None:
```
TallJimbo: This check only tells you whether *some* `Datastore` has written it, not whether this `Datastore` has, and I think that's what you want to be checking.

That said, I think users actually may want to specify (e.g. with a boolean arg) whether they're asking for a guess URI for something they haven't written yet or a URI for something that should exist, so they can get an exception when they don't get what they want.

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -112,127 +186,200 @@ def get(self, uri, storageClass, parameters=None):
...
         """
-        location = self.locationFactory.fromUri(uri)
-        if not os.path.exists(location.preferredPath()):
+        # Get file metadata and internal metadata
+        storedFileInfo = self.getStoredFileInfo(ref)
+        location = self.locationFactory.fromPath(storedFileInfo.path)
+        if not os.path.exists(location.path):
             raise FileNotFoundError("No such file: {0}".format(location.uri))
-        os.remove(location.preferredPath())
+        os.remove(location.path)
```
TallJimbo, line 379: Delete from `StoredFileInfo`, too?

python/lsst/daf/butler/formatters/fileFormatter.py
```diff
@@ -99,46 +99,54 @@ def _coerceType(self, inMemoryDataset, storageClass, pytype=None):
...
+        if readStorageClass != fileDescriptor.storageClass:
+            if comp is None:
+                raise ValueError("Storage class inconsistency ({} vs {}) but no"
+                                 " component requested".format(readStorageClass.name,
+                                                               fileDescriptor.storageClass.name))
+
+            # Concrete composite written as a single file (we hope)
+            try:
+                data = fileDescriptor.storageClass.assembler().getComponent(data, comp)
+            except AttributeError:
```
TallJimbo, line 143: I'm surprised the expected exception is an `AttributeError`; is that a detail of the default implementation?

TallJimbo (approved):
Looks good.  I had one structural comment about where disassemblers are associated with StorageClasses that I suspect is just a misunderstanding on my part, and a bunch of smaller things that I think are just minor cleanups.

I also think we're missing a piece of the lookup for components of deferred virtual composites in *Registry* (@pschella); there's an `if` statement in the confluence page pseudocode for that that doesn't seem to be implemented yet here.

python/lsst/daf/butler/butler.py
timj: storage classes are constructed via StorageClassFactory from config file. That's a singleton (as we discussed last week). The definition includes a definition of the assembler class. This is a name at class construction time. When you ask for an assembler instance from the storageclass it does a runtime import of the assembler (so you only need afw Exposure knowledge when you are actively wanting to deal with an Exposure). This is all hidden from registry and butler so registry is not involved other than asking the factory to convert the name into an instance when creating the DatasetType, and butler can just assume that the assembler will be there. You always have an assembler defined, but a disassemble method can be nulled out if you don't want to split up the in memory dataset into components.

Does that help?

python/lsst/daf/butler/datastores/posixDatastore.py
timj: Good point. What does existence mean...? I was treating existence as "can I retrieve this thing" not "has it ever existed".

python/lsst/daf/butler/core/fileTemplates.py
timj: Yes. Now that I've encoded the dot syntax into `DatasetType` I don't think a separate component is needed any more. I think I need to use the new `DatasetType` splitter here now.

python/lsst/daf/butler/datastores/posixDatastore.py
timj: When I wrote this I didn't have an exists method. I'll add an option to only retrieve things that actually exist.

python/lsst/daf/butler/butler.py
TallJimbo: That all makes sense if the `disassemble` method is either nulled out or not the original config file that defines the `StorageClass`, and can never change after that.

Let's say we're talking about a `StorageClass` that's define in the global configuration.

Can per-butler override configuration null the disassemble method?

python/lsst/daf/butler/formatters/fileFormatter.py
timj: https://github.com/lsst/daf_butler/blob/master/python/lsst/daf/butler/core/composites.py#L205

You prefer `KeyError`? I throw `AttributeError` because I'm testing with `hasattr` in there.

TallJimbo: I don't have a strong opinion; I just wanted to make sure this was being considered as part of the `CompositeAssembler` public API rather than being an accident (and I see now that the exception type is documented there as the one derived classes should raise, so I'm happy).

python/lsst/daf/butler/butler.py
timj: StorageClassFactory is a singleton so you always retrieve the same StorageClass with a given name so if you have multiple butlers instantiated you do indeed end up with no way to enable or disable the disassembler per-butler. StorageClass doesn't have any instance data either so you can't override it by configuration. In the previous implementation this wasn't a problem because each butler got its own StorageClassFactory and the StorageClass definitions in the factory were allowed to be different even if they had the same fully qualified name. The switch to a singleton does require me to think some more about this.

TallJimbo: Sounds good.  I don't mind merging this as is; I don't think fixing it in the future will break anything structural.

python/lsst/daf/butler/datastores/posixDatastore.py
timj: this is also a datastore.exists and not a butler.exists. If butler splits up an exposure and writes it as separate items datastore never sees the parent DatasetRef so can't understand that it does indeed have all the bits needed to make the composite. The only way this would really work is if butler could register the parent DatasetRef with the datastore after it's put all the components.

python/lsst/daf/butler/core/fileTemplates.py
timj: Actually, I think it should just take a `DatasetRef` directly (it has all the information).

python/lsst/daf/butler/datastores/posixDatastore.py
TallJimbo: Keeping Datastore from knowing about these parents is probably a good idea.  I think that just means we need a way for Registry to know that no Datastore entry *should* exist.  That was going to be the `assembler` field in Dataset, and I think we can now just make that into a bool.

PR #28, opened 2018-04-02
DM-13963: Enable foreign_key integrity checking on Butler SqlRegistry

pschella:


python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -21,7 +21,11 @@
 
 import itertools
 
+from sqlite3 import Connection as SQLite3Connection
```
timj: Any chance we could have the sqlite knowledge in a subclass?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -36,6 +40,14 @@
 __all__ = ("SqlRegistryConfig", "SqlRegistry")
 
 
+@event.listens_for(Engine, "connect")
+def _enableForeignKeysOnSqlite3(dbapiConnection, connectionRecord):
+    if isinstance(dbapiConnection, SQLite3Connection):
+        cursor = dbapiConnection.cursor()
+        cursor.execute("PRAGMA foreign_keys=ON;")
```
timj: Do you need a `try/finally` here to ensure that `cursor.close()` is called? Or the internet gives the alternative form of:

```python
from contextlib import closing
with closing(self.connection.cursor()) as cur:
```

timj (approved):
Overall is fine. Can I please have some reasoning in the commit message that switches the datastore tests to a dummy registry rather than the in-memory registry?

python/lsst/daf/butler/registries/sqlRegistry.py
pschella: We probably should. But I'm not sure we want to do that right now?

timj: would it be anything other than copying 20 lines to another file and changing the registry name in the yaml file?

python/lsst/daf/butler/registries/sqlRegistry.py
pschella: Hmm, we don't want the connection to be closed at least (because the whole point of the listener is to do this on every opened connection). I'm not sure about the cursor. I can test though.

PR #29, opened 2018-04-05
DM-14012: Add helper container class supporting topologically sorted iteration over elements

pschella:
Undoubtedly not optimally efficient, and perhaps wrongly named, but at least useful.

### DIFF
--- a/python/lsst/daf/butler/core/utils.py
+++ b/python/lsst/daf/butler/core/utils.py
@@ -19,6 +19,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from collections import namedtuple
+
 
 def iterable(a):
     """Make input iterable.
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/core/utils.py
```diff
@@ -203,3 +203,87 @@ def __call__(cls):  # noqa N805
         if cls not in cls._instances:
             cls._instances[cls] = super(Singleton, cls).__call__()
         return cls._instances[cls]
+
+
+class ConnectedSet:
```
timj: Does it need a `__len__` to make it properly look like a `collections.Set`?

pschella: Probably. I'll add it. And perhaps some of the other methods too. But I'm not sure supporting anything other than iteration, length and perhaps `union` is worth the effort at the moment.

pschella:
Oh, and what do we think about naming?

- TopologicalSet?
- ConnectedSet?
- ...

python/lsst/daf/butler/core/utils.py
```diff
@@ -203,3 +203,87 @@ def __call__(cls):  # noqa N805
...
+        Raises
+        ------
+        KeyError
+            When either element is not already in the set.
+        ValueError
+            When a dupplicate connection is inserted.
+        """
+        for element in (sourceElement, targetElement):
+            if element not in self._elements:
+                raise KeyError('{} not in set'.format(element))
```
natelust: maybe make this:
`
missing = {sourceElement, TargetElement} - self._elements
if missing: raise KeyError('element(s) {x} not in set'.format(x=missing))
`

python/lsst/daf/butler/core/utils.py
```diff
@@ -203,3 +203,87 @@ def __call__(cls):  # noqa N805
...
+    sorted order.
+
+    Parameters
+    ----------
+    elements : `iterable`
+        Any iterable with elements to insert.
+    """
+    def __init__(self, elements):
+        self._elements = set(elements)
+        self._connections = {e: set() for e in elements}
```
natelust: This probably should iterate over self._elements, as it is not guaranteed all the entries in the elements iterable will be unique.  While this behaviorally wont make any difference, it will save you time on duplicate iteration and construction time on sets that will just be discarded the next time a duplicate e is encountered.

python/lsst/daf/butler/core/utils.py
```diff
@@ -203,3 +203,87 @@ def __call__(cls):  # noqa N805
...
+        KeyError
+            When either element is not already in the set.
+        ValueError
+            When a dupplicate connection is inserted.
+        """
+        for element in (sourceElement, targetElement):
+            if element not in self._elements:
+                raise KeyError('{} not in set'.format(element))
+        if element in self._connections[sourceElement]:
+            raise ValueError('Dupplicate connection: ({}, {})'.format(sourceElement, targetElement))
```
natelust: Do you really care that there is a duplicate connection, vs just making it a no-op? if the user just says I want these connected, why should they need a try except block to catch something when the result of the operation they are asking for is done. It seems similar to a caching function.

python/lsst/daf/butler/core/utils.py
```diff
@@ -203,3 +203,87 @@ def __call__(cls):  # noqa N805
...
+        ------
+        ValueError
+            If a cycle is found and hence no topological order exists.
+        """
+        # Iteration needs to modify the number of incomming connections
+        _nIncommingConnections = self._nIncommingConnections.copy()
+        _nConnections = self._nConnections
+        # Set of all nodes with no incoming connections
+        startPoints = set(e for (e, n) in _nIncommingConnections.items() if n == 0)
+        while len(startPoints) > 0:
```
natelust: just have this be:
`
while startPoints:
`
an empty set is false-y

python/lsst/daf/butler/core/utils.py
```diff
@@ -203,3 +203,87 @@ def __call__(cls):  # noqa N805
...
+        targetElement : `object`
+            The target element.
+
+        Raises
+        ------
+        KeyError
+            When either element is not already in the set.
+        ValueError
+            When a dupplicate connection is inserted.
+        """
```
natelust: You should add a check to make sure that sourceElement != targetElement

natelust: Though I guess you would pick that up in your cycle check, but it is probably better to stop the user shooting themselves in the foot by accident

natelust (commented):
Some things to consider, and maybe try the algorithm we talked about in person

pschella:
Took your first suggestion (to use DFS instead to have less jumping all over the place) and your `namedtuple` suggestion (which I don't really like normally, but works well here).
If you have some suggestion to get rid of the need for both `seen` and `finished` that would be great, but I don't want to spend too much time on it now.

python/lsst/daf/butler/core/utils.py
```diff
@@ -290,3 +292,93 @@ def __iter__(self):
...
+        Raises
+        ------
+        KeyError
+            When either element is not already in the set.
+        ValueError
+            If a self connections between elements would be created.
+        """
+        if sourceElement == targetElement:
+            raise ValueError('Cannot connect {} to itself'.format(sourceElement))
+        for element in (sourceElement, targetElement):
```
natelust, line 255: I still think:
`
missing = {sourceElement, TargetElement} - self._elements 
if missing: raise KeyError('element(s) {x} not in set'.format(x=missing))
`
Would be better than a for loop. It would also have the benefit of telling the user if point points are invalid rather than bailing at the first

PR #30, opened 2018-04-07
DM-12635: Subpackage for converting Gen2->Gen3.

TallJimbo:
The conversion code here has two high-level drivers:
 - The `ConversionWalker` class (`walker.py`) extracts everything we'll need for the conversion from a Gen2 repo.  It's basically complete, and I've done some one-off testing it on `ci_hsc` (check out `ci_hsc`, setup and build, and start by calling `ConversionWalker.tryRoot` on `$CI_HSC_DIR/DATA`).
 - The `ConversionWriter` class (`writer.py`) adds entries to a `Registry` and `Datastore` - or it would if all of the actual calls to do so weren't just "TODO" comment placeholders

As for those TODO placeholders, there are a ton of them, and I think they cover most of what needs to be done to get things working.  Beyond that, we'll also eventually need a high-level driver that can be run easily from the command-line, but that doesn't necessarily need to be done right now.

### DIFF
--- a/config/gen2convert.yaml
+++ b/config/gen2convert.yaml
@@ -1,0 +1,51 @@
+mappers:
+  HscMapper:
+    # Gen3 Camera name associated with this Gen2 Mapper
+    camera: HSC
+    # VisitInfo objects are used to populate Gen3 Visit and Exposure DataUnit
+    # entries; options here say how to get those from a Gen2 repo.
+    VisitInfo:
+      # The Gen2 DatasetType to read when trying to create a VisitInfo.
+      # (we actually add a "_md" suffix, because we just read the metadata).
+      DatasetType: raw
+      # Gen2 Data ID keys for the above DatasetType that are used
+      # to construct Gen3 Exposure/Visit identifiers.
+      # At present, only one key can be provided and this is used directly
+      # as the Gen3 value.
+      uniqueKeys:
+        - visit
+collections:
+  substitutions:
+    # Regular expression patterns and replacement strings (passed directly
+    # to Python's re.sub) applied in order to all Gen2 absolute repository
+    # paths in order to construct names for the Collections they go into.
+    # This can be used to merge Gen2 repositories into a single Collection
+    # by making them reduce to the same name.
+    -
+      pattern: "^(.+)/rerun/private/"
+      repl: "u/"
+    -
+      pattern: "^(.+)/rerun/"
+      repl: "shared/"
+  overrides:
+    # A dictionary of DatasetType -> Collection mappings that force all
+    # Datasets with that DatasetType into a particular Collection.
+    # Collection names can be str.format patterns that utilize any of
+    # the Gen3 DataUnits or Gen2 DataIds associated with the Dataset.
+    raw: raw/{camera}
+    ref_cat: ref/{name}
+    ref_cat_config: ref/{name}
+runs:
+  # Names of Collections (after processing via the above section)
+  # that should be assigned to a particular Run.
+  # Note that the *first* Collection a Dataset is added to determines
+  # its Run; this will be the one corresponding to the Gen2 repository
+  # that originally contained the file, unless that has been overridden.
+  raw/HSC: 1
+  ref/ps1_pv3_3pi_20170110: 2
+
+skymaps: {}  # dictionary mapping repository roots to Gen3 SkyMap names
+
+storageClasses:
+  # dictionary mapping Gen2 Mapping.persistable to Gen3 StorageClass name
+  {}
--- a/config/registry/default_schema.yaml
+++ b/config/registry/default_schema.yaml
@@ -262,6 +262,12 @@ dataUnits:
             Local sideral time on the meridian (equivalent, but not
             equal, to Local Mean Sidereal Time).
         -
+          name: seeing
+          type: float
+          doc: >
+            Average seeing, measured as the FWHM of the Gaussian with the same
+            effective area (arcsec).
+        -
           name: region
           type: blob
           doc: >
--- a/config/registry/default_schema.yaml
+++ b/config/registry/default_schema.yaml
@@ -31,19 +31,6 @@ dataUnits:
     -
       name: abstract_filter
       type: string
-      #foreign_key: AbstractFilter.abstract_filter
-    tables:
-      AbstractFilter:
-        doc: >
-          A definition table containing recognized AbstractFilter names.
-        columns:
-        -
-          name: abstract_filter
-          type: string
-          primary_key: true
-          doc: >
-            The name of the filter, typically a single letter
-            (e.g. "r" or "i").
 
   PhysicalFilter:
     dependencies:
@@ -88,9 +75,6 @@ dataUnits:
         -
           src: camera
           tgt: Camera.camera
-        -
-          src: abstract_filter
-          tgt: AbstractFilter.abstract_filter
 
   Sensor:
     dependencies:
--- a/config/registry/default_schema.yaml
+++ b/config/registry/default_schema.yaml
@@ -330,6 +330,12 @@ dataUnits:
           doc: >
             The bandpass filter used for all exposures in this Visit.
         -
+          name: physical_filter
+          type: string
+          nullable: false
+          doc: >
+            The bandpass filter used for all exposures in this Visit.
+        -
           name: snap
           type: int
           doc: >
--- a/config/registry/default_schema.yaml
+++ b/config/registry/default_schema.yaml
@@ -196,13 +196,20 @@ dataUnits:
             same as the datetime_begin of the first Exposure associated
             with this Visit.
         -
+          name: datetime_end
+          type: datetime
+          doc: >
+            Timestamp of the end of the Visit.  This should be the
+            same as the datetime_end of the last Exposure associated
+            with this Visit.
+        -
           name: exposure_time
           type: float
           doc: >
             The total exposure time of the Visit in seconds.  This should
             be equal to the sum of the exposure_time values for all
-            constituent Exposures (i.e. it should not include time with
-            the shutter closed).
+            constituent Exposures (i.e. it should not include time between
+            Exposures).
         -
           name: region
           type: blob
--- a/config/registry/default_schema.yaml
+++ b/config/registry/default_schema.yaml
@@ -600,7 +600,8 @@ dataunit_joins:
     rhs:
     - ExposureRange
     sql:
-      where: (lhs.camera = rhs.camera) AND (lhs.exposure BETWEEN rhs.valid_first AND rhs.valid_last)
+      (lhs.camera = rhs.camera) AND
+      (lhs.exposure BETWEEN rhs.valid_first AND rhs.valid_last)
   MultiCameraExposureJoin:
     doc: >
       A join table that relates Exposures from different Cameras, with
@@ -680,6 +681,7 @@ dataunit_joins:
         -
           name: sensor
           type: int
+          nullable: false
           doc: Sensor ID
         -
           name: skypix
--- a/config/registry/default_schema.yaml
+++ b/config/registry/default_schema.yaml
@@ -157,6 +157,8 @@ dataUnits:
       A sequence of observations processed together, comprised of one or
       more Exposures from the same Camera with the same pointing and
       PhysicalFilter.
+      The Visit table contains metadata that is both meaningful only for
+      science Exposures and the same for all Exposures in a Visit.
     link:
     -
       name: visit
@@ -210,6 +212,55 @@ dataUnits:
             be equal to the sum of the exposure_time values for all
             constituent Exposures (i.e. it should not include time between
             Exposures).
+        -
+          name: earth_rotation_angle
+          type: float
+          doc: Earth rotation angle in degrees.
+        -
+          name: boresight_ra
+          type: float
+          doc: Position of the boresight in right ascension (decimal degrees).
+        -
+          name: boresight_dec
+          type: float
+          doc: Position of the boresight in declination (decimal degrees).
+        -
+          name: boresight_alt
+          type: float
+          doc: Position of the boresight in altitude (decimal degrees).
+        -
+          name: boresight_az
+          type: float
+          doc: Position of the boresight in azimuth (decimal degrees).
+        -
+          name: boresight_hour_angle
+          type: float
+          doc: Hour angle at the boresight.
+        -
+          name: boresight_parallactic_angle
+          type: float
+          doc: >
+            Equal to the angle between the North celestial pole and Zenith at
+            the boresight. Or, the angular separation between two great circle
+            arcs that meet at the object, one passing through the North
+            celestial pole, and the other through zenith. For an object on the
+            meridian the angle is zero if it is South of zenith and pi if it is
+            North of zenith The angle is positive for objects East of the
+            meridian, and negative for objects to the West.
+        -
+          name: boresight_airmass
+          type: float
+          doc: Airmass at the boresight, relative to zenith at sea level.
+        -
+          name: rot_angle
+          type: float
+          doc: Rotation angle of the focal plane w.r.t. nominal.
+        -
+          name: local_era
+          type: float
+          doc: >
+            Local sideral time on the meridian (equivalent, but not
+            equal, to Local Mean Sidereal Time).
         -
           name: region
           type: blob
@@ -299,6 +350,10 @@ dataUnits:
       don't have multiple Exposures per Visit.  As a result, Cameras that
       don't have multiple Exposures per Visit will typically have Visit
       entries that are essentially duplicates of their Exposure entries.
+
+      The Exposure table contains metadata entries that are relevant for
+      calibration Exposures, and does not duplicate entries in Visit that
+      would be the same for all Exposures within a Visit.
     link:
     -
       name: exposure
@@ -356,8 +411,23 @@ dataUnits:
         -
           name: exposure_time
           type: float
-          doc: >
-            Duration of the Exposure in seconds.
+          doc: Duration of the Exposure with shutter open (seconds).
+        -
+          name: dark_time
+          type: float
+          doc: Duration of the Exposure with shutter closed (seconds).
+        -
+          name: rot_angle
+          type: float
+          doc: Rotation angle of the focal plane w.r.t. nominal
+        -
+          name: boresight_alt
+          type: float
+          doc: Position of the boresight in altitude (decimal degrees).
+        -
+          name: boresight_az
+          type: float
+          doc: Position of the boresight in azimuth (decimal degrees).        
         foreignKeys:
         -
           src: camera
--- a/python/lsst/daf/butler/datastores/posixDatastore.py
+++ b/python/lsst/daf/butler/datastores/posixDatastore.py
@@ -283,6 +283,23 @@ class PosixDatastore(Datastore):
         path = formatter.write(inMemoryDataset, FileDescriptor(location,
                                                                storageClass=storageClass))
 
+        self.ingest(path, ref, formatter=formatter)
+
+    def ingest(self, path, ref, formatter=None):
+        """Record that a Dataset with the given `DatasetRef` exists in the store.
+
+        Parameters
+        ----------
+        path : `str`
+            File path, relative to the repository root.
+        ref : `DatasetRef`
+            Reference to the associated Dataset.
+        formatter : `Formatter` (optional)
+            Formatter that should be used to retreive the Dataset.
+        """
+        if formatter is None:
+            formatter = self.formatterFactory.getFormatter(ref.datasetType.storageClass,
+                                                           ref.datasetType.name)
         # Create Storage information in the registry
         ospath = os.path.join(self.root, path)
         checksum = self.computeChecksum(ospath)
@@ -292,7 +309,7 @@ class PosixDatastore(Datastore):
         self.registry.addStorageInfo(ref, info)
 
         # Associate this dataset with the formatter for later read.
-        fileInfo = StoredFileInfo(formatter, path, storageClass)
+        fileInfo = StoredFileInfo(formatter, path, ref.datasetType.storageClass)
         self.addStoredFileInfo(ref, fileInfo)
 
         # Register all components with same information
--- a/python/lsst/daf/butler/gen2convert/extractor.py
+++ b/python/lsst/daf/butler/gen2convert/extractor.py
@@ -1,0 +1,165 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import re
+from collections import OrderedDict
+
+from .structures import Gen2Dataset, Gen2DatasetType
+
+__all__ = ("Extractor",)
+
+# Regular expression that matches a single substitution in
+# Gen2 CameraMapper template, such as "%(tract)04d".
+TEMPLATE_RE = re.compile(r'\%\((?P<name>\w+)\)[^\%]*?(?P<type>[idrs])')
+
+
+class FilePathParser:
+    """A callable object that extracts Gen2Dataset instances from filenames
+    corresponding to a particular Gen2 DatasetType.
+
+    External code should use the `fromMapping` method to construct instances.
+
+    Parameters
+    ----------
+    datasetType : `Gen2DatasetType`
+        Information about the DatasetType this parser processes.
+    regex : regular expression object
+        Regular expression pattern with named groups for all data ID keys.
+    """
+
+    @classmethod
+    def fromMapping(cls, mapping):
+        """Construct a FilePathParser instance from a Gen2
+        `lsst.obs.base.Mapping` instance.
+        """
+        try:
+            template = mapping.template
+        except RuntimeError:
+            return None
+        datasetType = Gen2DatasetType(name=mapping.datasetType,
+                                      keys={},
+                                      persistable=mapping.persistable,
+                                      python=mapping.python)
+        # The template string is something like
+        # "deepCoadd/%(tract)04d-%(patch)s/%(filter)s"; each step of this
+        # iterator corresponds to a %-tagged substitution string.
+        # Our goal in all of this parsing is to turn the template into a regex
+        # we can use to extract the associated values when matching strings
+        # generated with the template.
+        last = 0
+        terms = []
+        allKeys = mapping.keys()
+        for match in TEMPLATE_RE.finditer(template):
+            # Copy the (escaped) regular string between the last substitution
+            # and this one to the terms that will form the regex.
+            terms.append(re.escape(template[last:match.start()]))
+            # Pull out the data ID key from the name used in the
+            # subistution string.  Use that and the substition
+            # type to come up with the pattern to use in the regex.
+            name = match.group("name")
+            if name == "patch":
+                pattern = r"\d+,\d+"
+            elif match.group("type") in "id":  # integers
+                pattern = r"0*\d+"
+            else:
+                pattern = ".+"
+            # only use named groups for the first occurence of a key
+            if name not in datasetType.keys:
+                terms.append(r"(?P<%s>%s)" % (name, pattern))
+                datasetType.keys[name] = allKeys[name]
+            else:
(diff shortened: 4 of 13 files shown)


timj:
Important note: datastore doesn't yet write the formatter information to somewhere permanent. It uses a shim at the moment that passes tests but forgets immediately the process ends. We need a more permanent fix in there.

TallJimbo:
Good to know.  I think we'd get a lot of the benefits from just getting this to write to `Registry`, as that'd let us easily put together realistically complex test datasets for it.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -508,6 +510,7 @@ def addRun(self, run):
                                                         collection=run.collection,
                                                         environment_id=None,  # TODO add environment
                                                         pipeline_id=None))    # TODO add pipeline
+        # TODO: set given Run's 'id' attribute.
```
pschella, line 571: No, this is already done in the `Execution` part.

python/lsst/daf/butler/core/registry.py
```diff
@@ -135,6 +135,7 @@ def addDataset(self, ref, uri, components, run, producer=None):
             If a Dataset with the given `DatasetRef` already exists in the
             given Collection.
         """
+        # TODO: this signature no longer agrees with SqlRegistry.addDataset.
```
pschella: One of many I'm afraid. I'm planning to copy them all over after things have settled a bit.

python/lsst/daf/butler/gen2convert/extractor.py
```diff
@@ -0,0 +1,123 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import re
+from collections import OrderedDict
+
+from .structures import Gen2Dataset
+
+TEMPLATE_RE = re.compile(r'\%\((?P<name>\w+)\).*?(?P<type>[idrs])')
```
pschella: Would be nice to add an example comment for what this matches.

python/lsst/daf/butler/gen2convert/writer.py
```diff
@@ -0,0 +1,174 @@
...
+        return converted
+
+    def run(self, registry, datastore):
+        """Main driver for ConversionWriter.
+        """
+        self.insertCameras(registry)
+        self.insertSkyMaps(registry)
+        self.insertObservations(registry)
+        self.insertDatasets(registry, datastore)
+        # TODO: associate parent repo datasets with child repo Collections
```
pschella: Is this `Registry.merge`? Or just `Registry.associate`?

TallJimbo:
This is now as far along as I think make sense on this ticket.  Next steps will be making `Datastore` records persistent and adding concrete `StorageClasses` and `Formatters`.

config/gen2convert.yaml
```diff
@@ -0,0 +1,51 @@
+mappers:
+  HscMapper:
+    # Gen3 Camera name associated with this Gen2 Mapper
+    camera: HSC
+    # VisitInfo objects are used to populate Gen3 Visit and Exposure DataUnit
+    # entries; options here say how to get those from a Gen2 repo.
+    VisitInfo:
+      # The Gen2 DatasetType to read when trying to create a VisitInfo.
+      # (we actually add a "_md" suffix, because we just read the metadata).
```
pschella, line 9: Something we don't (yet) have in Gen3. Unless you mean `findDataUnitEntry` type things here.

config/gen2convert.yaml
```diff
@@ -0,0 +1,51 @@
...
+    ref_cat: ref/{name}
+    ref_cat_config: ref/{name}
+runs:
+  # Names of Collections (after processing via the above section)
+  # that should be assigned to a particular Run.
+  # Note that the *first* Collection a Dataset is added to determines
+  # its Run; this will be the one corresponding to the Gen2 repository
+  # that originally contained the file, unless that has been overridden.
+  raw/HSC: 1
+  ref/ps1_pv3_3pi_20170110: 2
```
pschella, line 45: We may want to allow for a `None` default that will generate a new `Run`.

config/registry/default_schema.yaml
```diff
@@ -195,14 +181,76 @@ dataUnits:
             Timestamp of the beginning of the Visit.  This should be the
             same as the datetime_begin of the first Exposure associated
             with this Visit.
+        -
+          name: datetime_end
+          type: datetime
+          doc: >
+            Timestamp of the end of the Visit.  This should be the
+            same as the datetime_end of the last Exposure associated
+            with this Visit.
```
pschella, line 190: How is this enforced? Should it be?

config/registry/default_schema.yaml
```diff
@@ -343,8 +401,23 @@ dataUnits:
...
-            Duration of the Exposure in seconds.
+          doc: Duration of the Exposure with shutter open (seconds).
+        -
+          name: dark_time
+          type: float
+          doc: Duration of the Exposure with shutter closed (seconds).
+        -
+          name: rot_angle
+          type: float
+          doc: Rotation angle of the focal plane w.r.t. nominal
```
pschella, line 412: period

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -283,6 +283,23 @@ def put(self, inMemoryDataset, ref):
         path = formatter.write(inMemoryDataset, FileDescriptor(location,
                                                                storageClass=storageClass))
 
+        self.ingest(path, ref, formatter=formatter)
+
+    def ingest(self, path, ref, formatter=None):
+        """Record that a Datset with the given `DatasetRef` exists in the store.
```
pschella: Dataset

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -292,7 +309,7 @@ def put(self, inMemoryDataset, ref):
         self.registry.addStorageInfo(ref, info)
 
         # Associate this dataset with the formatter for later read.
-        fileInfo = StoredFileInfo(formatter, path, storageClass)
+        fileInfo = StoredFileInfo(formatter, path, ref.datasetType.storageClass)
```
pschella, line 312: How was this working before?

python/lsst/daf/butler/gen2convert/extractor.py
```diff
@@ -0,0 +1,159 @@
...
+
+import re
+from collections import OrderedDict
+
+from .structures import Gen2Dataset, Gen2DatasetType
+
+__all__ = ("Extractor",)
+
+
+TEMPLATE_RE = re.compile(r'\%\((?P<name>\w+)\).*?(?P<type>[idrs])')
```
pschella: Comment explaining what it matches?

python/lsst/daf/butler/gen2convert/extractor.py
```diff
@@ -0,0 +1,159 @@
...
+                    datasetType.keys[name] = allKeys[name]
+                else:
+                    terms.append(r"(%s)" % pattern)
+                m2 = iterator.__next__()
+                terms.append(re.escape(template[m1.end():m2.start()]))
+                m1 = m2
+        except StopIteration:
+            pass
+        terms.append(re.escape(template[m1.end():]))
+        return cls(datasetType=datasetType, regex=re.compile("".join(terms)))
```
pschella: Ok, this function definitely would benefits from some more comments. In particular describing the role of `m1` and `m2`. I also suspect that this would be clearer when setting `terms=None` for the first case and then using a single `for else` block for the looping of both cases.

pschella: Also with parsing in general it is always convenient to have an example of the to-be-matched string in the documentation somewhere.

config/gen2convert.yaml
```diff
@@ -0,0 +1,51 @@
...
+    # to Python's re.sub) applied in order to all Gen2 absolute repository
+    # paths in order to construct names for the Collections they go into.
+    # This can be used to merge Gen2 repositories into a single Collection
+    # by making them reduce to the same name.
+    -
+      pattern: "^(.+)/rerun/private/"
+      repl: "u/"
+    -
+      pattern: "^(.+)/rerun/"
+      repl: "shared/"
```
pschella, line 29: Are we inventing a namespacing convention for collections? Michelle might want input on this.

python/lsst/daf/butler/gen2convert/extractor.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+class Extractor:
+    """An object that parses Gen2 paths into Gen2Dataset instance for a
+    particular Gen2 data repository.
+
+    Parameters
+    ----------
+    repo : `Gen2Repo`
+        Structure describing the repository this Extractor will process.
```
pschella, line 127: `Attributes` too?

python/lsst/daf/butler/gen2convert/extractor.py
```diff
@@ -0,0 +1,159 @@
...
+        m = self.regex.fullmatch(filePath)
+        if m is None:
+            return None
+        dataId = {k: v(m.group(k)) for k, v in self.datasetType.keys.items()}
+        return Gen2Dataset(datasetType=self.datasetType, dataId=dataId,
+                           filePath=filePath, root=root)
+
+
+class Extractor:
+    """An object that parses Gen2 paths into Gen2Dataset instance for a
```
pschella: instance? instances?

pschella: callable object?

python/lsst/daf/butler/gen2convert/translators.py
```diff
@@ -0,0 +1,275 @@
...
+    """Base class for Translator helpers that each handle just one Gen3 Data ID key.
+
+    Parameters
+    ----------
+    gen3key : `str`
+        Name of the Gen3 Data ID key (DataUnit link field name) populated by this
+        handler.
+    gen3unit : `str`
+        Name of the Gen3 DataUnit associated with `gen3key`.
+    """
```
pschella: An example would be nice. (e.g. `camera`, `Camera`).

python/lsst/daf/butler/gen2convert/translators.py
```diff
@@ -0,0 +1,275 @@
...
+
+    Parameters
+    ----------
+    handlers : `list`
+        A list of KeyHandlers this Translator should use.
+    skyMap : `BaseSkyMap`
+        SkyMap instance used to define any Tract or Patch DataUnits.
+    skyMapName : `str`
+        Gen3 SkyMap DataUnit name to be associated with any Tract or Patch
+        DataUnits.
```
pschella, line 130: Why is this not `skyMap.name`?

python/lsst/daf/butler/gen2convert/translators.py
```diff
@@ -0,0 +1,275 @@
...
+    """
+
+    __slots__ = ("handlers", "skyMap", "skyMapName")
+
+    # Rules used to match Handlers when constring a Translator.
+    # outer key is camera name, or None for any
+    # inner key is DatasetType name, or None for any
+    # values are 2-tuples of (frozenset(gen2keys), handler, consume)
```
pschella: How is this a 2-tuple?

python/lsst/daf/butler/gen2convert/translators.py
```diff
@@ -0,0 +1,275 @@
...
+    _rules = {
+        None: {
+            None: []
+        }
+    }
+
+    @classmethod
+    def addRule(cls, handler, camera=None, datasetTypeName=None, gen2keys=(), consume=True):
+        """Add a Handler and an associated matching rule.
```
pschella: `KeyHandler`

python/lsst/daf/butler/gen2convert/translators.py
```diff
@@ -0,0 +1,275 @@
...
+        """The Gen3 data ID keys populated by this Translator (`frozenset`)."""
+        return frozenset(h.gen3key for h in self.handlers)
+
+    @property
+    def gen3units(self):
+        """The Gen3 DataUnit (names) populated by this Translator (`frozenset`)."""
+        return frozenset(h.gen3unit for h in self.handlers)
+
+
+class ConstantKeyHandler(KeyHandler):
```
pschella: You may want to group these with the `KeyHandler`? Without the `Translator` in between.

python/lsst/daf/butler/gen2convert/translators.py
```diff
@@ -0,0 +1,275 @@
...
+    def __init__(self, gen3key, gen3unit, gen2key=None):
+        super().__init__(gen3key, gen3unit)
+        self.gen2key = gen2key if gen2key is not None else gen3key
+
+    def extract(self, gen2id, skyMap, skyMapName):
+        return gen2id[self.gen2key]
+
+
+class PatchKeyHandler(KeyHandler):
+    """A KeyHandler for Patchs."""
```
pschella: Patches?

python/lsst/daf/butler/gen2convert/walker.py
```diff
@@ -0,0 +1,329 @@
...
+
+__all__ = ("ConversionWalker",)
+
+
+class ConversionWalker:
+    """A class that walks Gen2 Data Repositories to extract the information
+    necessary to create a Gen3 'snapshot' view to them.
+
+    ConversionWalkers attempt to only include state that does not involve any user
+    choices about to perform the conversion; anything that does is handled
```
pschella: about how to

python/lsst/daf/butler/gen2convert/walker.py
```diff
@@ -0,0 +1,329 @@
...
+            for root in todo:
+                self.scanRepo(self.found[root])
+            todo = self.found.keys() - self.scanned.keys() - self.ignored
+
+    def tryRoot(self, root):
+        """Attempt to identify a Gen2 Data Repository at the given root path.
+
+        If the path appears to be a Gen2 Data Repository, a Gen2Data instance
+        representing it will be added to `self.found` (unless it is already there)
+        and returned.  `tryRoot()` also be called recursively on all parent
```
pschella: can be? will be? called recursively...

python/lsst/daf/butler/gen2convert/walker.py
```diff
@@ -0,0 +1,329 @@
...
+            return repo
+        if root in self.ignored:
+            return True
+
+        MapperClass = None
+        parentPaths = []
+        repoCfgPath = os.path.join(root, "repositoryCfg.yaml")
+        if os.path.exists(repoCfgPath):
+            with open(repoCfgPath, 'r') as f:
+                repoCfg = yaml.load(f)
```
pschella, line 106: Why is this not loaded with a `Config`?

python/lsst/daf/butler/gen2convert/walker.py
```diff
@@ -0,0 +1,329 @@
...
+        else:
+            parentLinkPath = os.path.join(root, "_parent")
+            if os.path.exists(parentLinkPath):
+                parentPaths.append(os.readlink(parentLinkPath))
+            mapperFilePath = os.path.join(root, "_mapper")
+            if os.path.exists(mapperFilePath):
+                with open(mapperFilePath, 'r') as f:
+                    terms = f.read().strip().split(".")
+                    module = importlib.import_module(".".join(terms[:-1]))
+                    MapperClass = getattr(module, terms[-1])
```
pschella: Why not use `utils.doImport`?

python/lsst/daf/butler/gen2convert/walker.py
```diff
@@ -0,0 +1,329 @@
...
+        repo = Gen2Repo(root, MapperClass)
+        self.found[root] = repo
+
+        try:
+            # Recursively construct (or look up) Gen2Repo objects for parents.
+            for path in parentPaths:
+                parent = self.tryRoot(path)
+                if parent is None:
+                    raise ValueError("%s is a parent of %s, but is not a repository." % (parent, root))
+                repo.parents.append(parent)
```
pschella: Also add if ignored?

python/lsst/daf/butler/gen2convert/walker.py
```diff
@@ -0,0 +1,329 @@
...
+            # Make sure there are no cycles in the graph of parents.
+            repo._findCycles()
```
pschella: Doesn't this repeat the recursive cycle check for all it's parents? E.g. can't you use that `tryRoot` on it's parents has already not found any cycles? Probably not worth it, but still. Basically you want to know that the sets of parents and the current instance are disjoint recursively.

python/lsst/daf/butler/gen2convert/walker.py
```diff
@@ -0,0 +1,329 @@
...
+            return existing
+
+        log.info("%s: making extractor.", repo.root)
+        extractor = Extractor(repo)
+        repo.datasetTypes.update(extractor.getDatasetTypes())
+
+        log.info("%s: walking datasets.", repo.root)
+        for dirPath, dirNames, fileNames in os.walk(repo.root):
+            dirNames[:] = [dirName for dirName in dirNames
+                           if not self.tryRoot(os.path.join(dirPath, dirName))]
```
pschella, line 188: This looks quite expensive. If you are already in a repo, can a subdirectory still be a root?

python/lsst/daf/butler/gen2convert/walker.py
```diff
@@ -0,0 +1,329 @@
...
+        with the latter a Gen2 filter name.
+        """
+        return self._visitInfo
+
+    def _ensureMapperClass(self, repo):
+        """Make sure self.MapperClass is defined.
+
+        If self.MapperClass is None, set it from the MapperClass used by
+        parent repositories.  Raises ValueError if this is there is no single
+        MapperClass used by all parents.
```
pschella: Better to also have this in the standard docstring format with `Raises` section.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -60,6 +60,33 @@ def __init__(self, config):
...
+
+        Yields
+        -------
+        row : `dict`
+            The next row result from executing the query.
+
+        """
+        t = text(sql)
+        with self._engine.begin() as connection:
+            for row in connection.execute(t, **params):
```
pschella: This seems rather unsafe. What if the user doesn't read the comment properly and writes something other than a `SELECT` statement?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -193,7 +221,8 @@ def addDataset(self, datasetType, dataId, run, producer=None):
                                                                      **dataId))
             datasetRef = DatasetRef(datasetType=datasetType, dataId=dataId, id=result.inserted_primary_key[0])
             # A dataset is always associated with its Run collection
-            self.associate(run.collection, [datasetRef])
+            connection.execute(datasetCollectionTable.insert(),
+                               [{'dataset_id': datasetRef.id, 'collection': run.collection}])
```
pschella: :(

tests/test_sqlDatabaseDict.py
```diff
@@ -0,0 +1,108 @@
...
+            d[0]
+
+    def testKeyInValue(self):
+        """Test when the value includes all fields, including the key."""
+        value = namedtuple("TestValue", ["x", "y", "z"])
+        data = {
+            0: value(x=0, y="zero", z=0.0),
+            1: value(x=1, y="one", z=0.1),
+        }
+        self.checkDatabaseDict(value, data)
```
pschella: Let's name `value` -> `TestValue` since it is the type.

python/lsst/daf/butler/gen2convert/writer.py
```diff
@@ -0,0 +1,316 @@
...
+        Runs all steps to create a Gen3 Repo, aside from Camera registration
+        (we merely check that the needed Cameras, Sensors, and PhysicalFilters
+        have already been registered).
+        """
+        self.checkCameras(registry)
+        self.insertSkyMaps(registry)
+        self.insertObservations(registry)
+        self.insertDatasetTypes(registry)
+        self.insertDatasets(registry, datastore)
+        # TODO: associate parent repo datasets with child repo Collections
```
pschella, line 184: Is this still open? Seems reasonably essential.

python/lsst/daf/butler/gen2convert/writer.py
```diff
@@ -0,0 +1,316 @@
...
+    def insertSkyMaps(self, registry):
+        """Add all necessary SkyMap DataUnits (and associated Tracts and
+        Patches) to the Registry.
+        """
+        log = Log.getLogger("lsst.daf.butler.gen2convert")
+        for sha1, skyMap in self.skyMaps.items():
+            skyMapName = self.skyMapNames.get(sha1, None)
+            try:
+                existing, = registry.query("SELECT skymap FROM SkyMap WHERE sha1=:sha1",
+                                           sha1=sha1)
```
pschella, line 210: In general we probably don't want to rely on `query` too much, when other API calls should exist.

python/lsst/daf/butler/gen2convert/writer.py
```diff
@@ -0,0 +1,316 @@
...
+            except ValueError:
+                # No SkyMap with this sha1 exists, so we need to insert it.
+                pass
+            if skyMapName is None:
+                raise LookupError(
+                    ("SkyMap with sha1={} has no name "
+                     "and does not already exist in the Registry.").format(sha1.hex())
+                )
+            log.info("Inserting SkyMap '%s' with sha1=%s", skyMapName, sha1.hex())
+            skyMap.register(skyMapName, registry)
```
pschella, line 230: Do we want the same skymap to have different names? Otherwise I don't really get why the call can't be `skyMap.register(registry)`.

pschella: Or directly `addDataUnitEntry(...)`.

python/lsst/daf/butler/gen2convert/writer.py
```diff
@@ -0,0 +1,316 @@
...
+                    "dark_time": visitInfo.getDarkTime()
+                })
+                visitValues = commonValues.copy()
+                visitValues.update({
+                    "datetime_end": mid + offset,
+                    "earth_rotation_angle": visitInfo.getEra().asDegrees(),
+                    "boresight_ra": visitInfo.getBoresightRaDec().getLongitude().asDegrees(),
+                    "boresight_dec": visitInfo.getBoresightRaDec().getLatitude().asDegrees(),
+                    "boresight_parallactic_angle": visitInfo.getBoresightParAngle().asDegrees(),
+                    "local_era": visitInfo.getLocalEra().asDegrees(),
```
pschella, line 270: Do we loose information with respect to `visitInfo` here?

python/lsst/daf/butler/gen2convert/writer.py
```diff
@@ -0,0 +1,316 @@
...
+                registry.addDataUnitEntry("Visit", visitValues)
+                registry.addDataUnitEntry("Exposure", exposureValues)
+
+    def insertDatasetTypes(self, registry):
+        """Add all necessary DatasetType registrations to the Registry.
+        """
+        log = Log.getLogger("lsst.daf.butler.gen2convert")
+        for datasetType in self.datasetTypes.values():
+            # TODO: should put this "just make sure it exists" logic
+            # into registerDatasetType itself, and make it a bit more careful.
```
pschella, line 282: Definitely!

pschella (commented):
I'm assuming we can't really unittest any of this?

python/lsst/daf/butler/core/registry.py
TallJimbo: Yes, this TODO has already been removed.

config/gen2convert.yaml
TallJimbo: I was thinking we'd handle the metadata as another `Exposure` component dataset.

config/gen2convert.yaml
TallJimbo: We're inventing a corner of a provisional one.  We can revisit later.

config/gen2convert.yaml
TallJimbo: That default is already in place; if nothing here matches, we'll make a new `Run`, and use the repo root at its `Collection` name.

config/registry/default_schema.yaml
TallJimbo: The most important place to enforce it is probably the raw data ingest script.  Since it (and a number of other constraints) can probably be verified via SQL, it might make sense to have a script we can run explicitly to check these things.  I doubt trying to add complex table constraints is worthwhile; I would guess those are pretty RDBMS-specific.

python/lsst/daf/butler/datastores/posixDatastore.py
TallJimbo: This code was moved out of `put`, which had a `storageClass` local variable pointing to the same object.

python/lsst/daf/butler/gen2convert/extractor.py
TallJimbo: I had previously convinced myself that I couldn't use a for loop here because I needed to have both the current and next element available together, but of course that's silly.  Thanks for the extra push to simplify it; it's much better now.

python/lsst/daf/butler/gen2convert/extractor.py
TallJimbo: Does that actually work?  I thought we had to do that nonsense with the class attributes that get overridden by instance attributes (and frankly I don't think that's worth it).

python/lsst/daf/butler/gen2convert/translators.py
TallJimbo: Maybe it should be, in the future, but `BaseSkyMap` (and its subclasses) are not new, and I'm a little wary of adding a `name` attribute since it can't be derived from the SkyMap's actual contents.

python/lsst/daf/butler/gen2convert/translators.py
TallJimbo: Historically, before I added `consume` :-)

python/lsst/daf/butler/gen2convert/walker.py
TallJimbo: `will be` (fixed)

python/lsst/daf/butler/gen2convert/walker.py
TallJimbo: I didn't want to assume that Gen3 config would remain compatible with Gen2 Config.

python/lsst/daf/butler/gen2convert/walker.py
TallJimbo: I was just copying the code from Gen2; I'll try `utils.doImport` and see if it works.

python/lsst/daf/butler/gen2convert/walker.py
TallJimbo: I think we can't really handle the case where a parent is ignored; I'll make that an explicit exception.

TallJimbo: Actually, I think we have other checks for when the parent was necessary; I'll just remove ignored parents from the list here.

python/lsst/daf/butler/gen2convert/walker.py
TallJimbo: Because `tryRoot` short-circuits if `repo` is already in `self.found`, and we add the current `repo` to `self.found` before recursing, it won't find cycles.

python/lsst/daf/butler/gen2convert/walker.py
TallJimbo: Yes, though in practice this only happens under the `rerun` subdirectory of the repo root.  We can take advantage of that if we need to to speed things up in the future.

python/lsst/daf/butler/registries/sqlRegistry.py
TallJimbo: It certainly is, but I'd rather not deal with it now.  I'll add a TODO.

python/lsst/daf/butler/registries/sqlRegistry.py
TallJimbo: I've added a TODO for this, too.

TallJimbo (commented):
I think I've addressed all review comments.  I'll squash and merge by Monday.

As for tests, I don't think it's a good use of time to try to cook up the kind of test data we'd need for small scale unit testing of this code in daf_butler.  Instead, I'm planning to add running the conversion to ci_hsc's SConstruct, and add checks in that package that we can access the datasets using the Gen3 butler and that this is the same as Gen2 access.  We need at least DM-14225 before that's worthwhile, though, and may want to add a few more concrete StorageClasses/Formatters first as well.

python/lsst/daf/butler/gen2convert/writer.py
TallJimbo: Yes, I'm afraid it is, and I remember it being hard but don't remember exactly why.  We can work around it with a custom call to `query` in the ci_hsc drivers for the conversion for now.

python/lsst/daf/butler/gen2convert/writer.py
TallJimbo: I completely agree.  I've been meaning to talk to you about how to add alternate unique lookup APIs for some DataUnits (SkyMap `sha1`, Patch `cell_x` and `cell_y`, Sensor `name`, probably some Camera-specific `Visit` or `Exposure` identifiers).

python/lsst/daf/butler/gen2convert/writer.py
TallJimbo: Possibly, but I don't want to think too hard about this until we've gone through the schema review.

PR #31, opened 2018-04-16
DM-14114: Add interface for camera specialization

pschella:


python/lsst/daf/butler/instrument.py
```diff
@@ -0,0 +1,70 @@
...
+        List of PhysicalFilter entries (each entry being a dict).
+    sensors : `list`
+        List of Sensor entries (each entry being a dict).
+    """
+    @property
+    def camera(self):
+        raise NotImplementedError('Must be specified by derived class')
+
+    physicalFilters = []
+    sensors = []
```
TallJimbo: It doesn't really make sense to me that `camera` is an instance property while the others are class attributes.

I think I'd probably just make `camera` a class attribute that's set to `None`, and document in the class that all that matters is that subclass instances have these attributes; that will let subclasses that need to override them at the instance level while allowing those that don't to override them at the class level.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -672,17 +673,25 @@ def markInputUsed(self, quantum, ref):
...
 
-        unit : `DataUnit`
-            The `DataUnit` to add or replace.
-        replace : `bool`
-            If `True`, replace any matching `DataUnit` that already exists
-            (updating its non-unique fields) instead of raising an exception.
+        dataUnitName : `str`
+            Name of the `DataUnit` (e.g. ``"Camera"``).
+        values : `dict`
+            Dictionary of ``columnName, value`` pairs.
```
TallJimbo: Should document that `value` is itself a dictionary.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -672,17 +673,25 @@ def markInputUsed(self, quantum, ref):
                 raise KeyError("{} is not a predicted consumer for {}".format(ref, quantum))
             quantum._markInputUsed(ref)
 
-    def addDataUnit(self, unit, replace=False):
-        """Add a new `DataUnit`, optionally replacing an existing one
-        (for updates).
+    def addDataUnitEntry(self, dataUnitName, value):
```
TallJimbo: Should there be an abstract version of this in the base `Registry` class?

TallJimbo (approved):
I've left a few minor comments, but I think this is enough for me to rebase on and start making some progress.  You probably have a better idea than I do now of what's missing here in order to get the conversion scripts working (since I've been away from that for a week).  If you want to add anything else to the PR before merging, just ping me if you'd like me to take a look.

python/lsst/daf/butler/registries/sqlRegistry.py
pschella: Yes, but at this point it is easier to copy the `SqlRegistry` API back into `Registry` after the dust settles.

pschella: Assuming nobody else is working on an alternative `Registry` implementation yet.

python/lsst/daf/butler/registries/sqlRegistry.py
pschella: It is? I had assumed it looked something like: `{"camera": "HSC", "abstract_filter": "i", "physical_filter": "HSC_i", "visit": 3}`. Did you have something else in mind?

TallJimbo: Oh, I was confused by the fact that the argument is called `values` in the doc but `value` in the code (and I was looking at the line where you do `**value` below).  Just rename the argument and I think it'll all make sense.

PR #33, opened 2018-04-19
One line per copyright statement

timj:
It seems that GitHub requires this to spot the correct license.


timj:
The need for this is discussed in benbalter/licensee#285

PR #34, opened 2018-04-19
DM-12624: Implement Butler DataUnits

pschella:


python/lsst/daf/butler/core/__init__.py
```diff
@@ -21,4 +21,3 @@
 from .storageClass import *
 from .storageInfo import *
 from .storedFileInfo import *
-from .units import *
```
timj, line 24: I think `dataUnit` has to be added to this file

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+class DataUnit:
+    """A discrete abstract unit of data that can be associated with
+    metadata and used to label datasets.
+
+    `DataUnit` instances represent concrete units such as e.g. `Camera`,
+    `Sensor`, `Visit` and `SkyMap`.
+
+    Attributes
```
timj: Init parameters need to be documented separately.

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -0,0 +1,271 @@
...
+        ----------
+        dataId : `dict`
+            A `dict` of `DataUnit` column name, value pairs.
+
+        Raises
+        ------
+        ValueError
+            If a value for a required dependency is missing.
+        """
+        for columnName in self.primaryKey:
```
timj: maybe do this using sets so that the error message shows all keys that are bad rather than the first.

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -0,0 +1,271 @@
...
+
+        Build a `DataUnitRegistry` instance from a `Config` object and an
+        (optional) `SchemaBuilder`.
+
+        Parameters
+        ----------
+        config : `SchemaConfig`
+            `Registry` schema configuration describing `DataUnit` relations.
+        builder : `SchemaBuilder`, optional
+            When given, create `Table` entries for every `DataUnit` table.
```
timj: There is no `Table` class in `daf_butler`.

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -0,0 +1,271 @@
...
+
+        Returns
+        -------
+        primaryKeyNames : `list`
+            All primary-key column names for the given ``dataUnitNames``.
+        """
+        primaryKeyNames = []
+        for dataUnitName in dataUnitNames:
+            primaryKeyNames.extend(self[dataUnitName].primaryKey)
+        return primaryKeyNames
```
timj:
```python
return [self[name].primarykey for name in dataUnitNames]
```
? Why a list and not a tuple?

python/lsst/daf/butler/core/schema.py
```diff
@@ -20,43 +20,18 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from .utils import iterable
-from .config import Config
```
timj: Why did this move down? If you are grouping by local versus external imports shouldn't the `.utils` import also move?

tests/test_butler.py
```diff
@@ -79,7 +79,7 @@ def testBasicPutGet(self):
         butler = Butler(self.configFile)
         # Create and register a DatasetType
         datasetTypeName = "test_metric"
-        dataUnits = ("camera", "visit")
+        dataUnits = ("Camera", "Visit")
```
timj, line 82: should they be case sensitive?

tests/test_sqlRegistry.py
```diff
@@ -189,7 +189,7 @@ def testStorageInfo(self):
     def testAssembler(self):
         registry = Registry.fromConfig(self.configFile)
         storageClass = makeNewStorageClass("testAssembler")()
-        datasetType = DatasetType(name="test", dataUnits=("camera",), storageClass=storageClass)
+        datasetType = DatasetType(name="test", dataUnits=("Camera",), storageClass=storageClass)
```
timj, line 192: `Camera` here and `camera` a few lines later so maybe they are case insensitive?

timj (approved):
I've had a quick look (I'm traveling) and it looks reasonably okay. I think you can now remove `daf_persistence` from the table file.

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -0,0 +1,271 @@
...
+        Related `DataUnit` instances on which existence this `DataUnit`
+        instance depends.
+    optionalDependencies : `frozenset`
+        Related `DataUnit` instances that may also be provided (and when they
+        are, they must be kept in sync).
+    dependencies : `frozenset`
+        The union of `requiredDependencies` and `optionalDependencies`.
+    table : `sqlalchemy.core.Table`, optional
+        When not ``None`` the primary table entry corresponding to this
+        `DataUnit`.
```
TallJimbo: Would this preclude the table actually being implemented as a view?

python/lsst/daf/butler/core/dataUnit.py
TallJimbo: And I think the attributes themselves need to be documented as the docstrings of the properties themselves.

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -0,0 +1,271 @@
...
+    @property
+    def dependencies(self):
+        return self.requiredDependencies.union(self.optionalDependencies)
+
+    @property
+    def table(self):
+        if hasattr(self, '_table'):
+            return self._table
+        else:
+            return None
```
TallJimbo: `getattr(self, '_table', None)`

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -0,0 +1,271 @@
...
+            If a value for a required dependency is missing.
+        """
+        for columnName in self.primaryKey:
+            if columnName not in dataId:
+                raise ValueError("Required key {} missing from {} for DataUnit {}".format(
+                    columnName, dataId, self.name))
+
+
+class DataUnitRegistry:
+    """Instances of this class keep track of `DataUnit` relations.
```
TallJimbo: Are there actually multiple instances of this class (or, if so, a need for multiple instances)?

And as much as it's the obvious name, given that `Registry` means something very important and specific in daf_butler, we should probably find a better name for this class.

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def __getitem__(self, dataUnitName):
+        return self._dataUnits[dataUnitName]
+
+    def __setitem__(self, dataUnitName, dataUnit):
+        assert isinstance(dataUnit, DataUnit)
+        self._dataUnits[dataUnitName] = dataUnit
+
+    def __iter__(self):
+        yield from self._dataUnitNames
```
TallJimbo: Is this different from / better than `iter(self._dataUnitNames)`?

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def keys(self):
+        yield from self._dataUnitNames
```
TallJimbo: `keys` should return a set-like view, not just an iterator, I think.  Is it problematic to just return `self._dataUnitNames` directly?

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def values(self):
+        return (self[dataUnitName] for dataUnitName in self._dataUnitNames)
```
TallJimbo: Should also return a view object as well.   I think a list or tuple copy is preferable to an iterator if `collections.abc` doesn't have a way to give you a real view (though I bet it does, especially if you inherit from `collections.abc.Mapping` and let the mixin define these methods).

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def items(self):
+        for dataUnitName in self._dataUnitNames:
+            yield (dataUnitName, self[dataUnitName])
```
TallJimbo, line 204: Should also be a view or at least a sequence.

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -0,0 +1,271 @@
...
+
+        Parameters
+        ----------
+        config : `SchemaConfig`
+            Schema configuration describing `DataUnit` relations.
+        """
+        self._dataUnitNames = TopologicalSet(config)
+        for dataUnitName, dataUnitDescription in config.items():
+            if 'dependencies' in dataUnitDescription:
+                dependencies = dataUnitDescription['dependencies']
```
TallJimbo, line 221: Not a big deal, but I prefer the
```
dependencies = dataUnitDescription.get('dependencies', None)
if dependencies is not None:
    ...
```
pattern for this.   But maybe I've just spent too much time in C++ thinking about how to avoid multiple `std::map` tree traversals.

Though, in this case, I suppose you could also just do:
```
dependencies = dataUnitDescription.get('dependencies', {}}
for category in ('required', 'optional'):
    for dependency in dependencies.get('category', ()):
        ...
```

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -0,0 +1,271 @@
...
+            requiredDependencies = ()
+            optionalDependencies = ()
+            table = None
+            link = ()
+            if 'dependencies' in dataUnitDescription:
+                dependencies = dataUnitDescription['dependencies']
+                if 'required' in dependencies:
+                    requiredDependencies = (self[name] for name in dependencies['required'])
+                if 'optional' in dependencies:
+                    optionalDependencies = (self[name] for name in dependencies['optional'])
```
TallJimbo, line 249: Maybe my brain just isn't functional enough, because I don't think this code is actually incorrect, but I think the code would be a bit less fragile if you just made these collections of some sort here (probably `frozenset`) rather than passing deferred iterators to the `DataUnit` constructor.

python/lsst/daf/butler/core/registry.py
```diff
@@ -19,7 +19,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from abc import ABCMeta, abstractmethod
+from abc import ABCMeta
```
TallJimbo, line 22: A code comment explaining that we'll put the extra methods back into `Registry` when we're done prototyping would be welcome.

python/lsst/daf/butler/core/schema.py
```diff
@@ -70,25 +45,38 @@ class Schema:
...
-        The sqlalchemy schema description
-    dataUnits : `dict`
-        Columns that represent dataunit links.
+        The sqlalchemy schema description.
+    """
+    def __init__(self, config):
+        if isinstance(config, str):
+            config = SchemaConfig(config)
+        self.config = config
+        self.builder = SchemaBuilder()
```
TallJimbo, line 54: I find it a little strange that the `builder` remains attached to the `Schema` even after the `Schema` has been built.  That might be a perfectly reasonable relationship, but I think if it is the `Builder` class might be better off with a different name, or we should at least have some docstrings that explain what their relationship is in detail.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -214,8 +214,8 @@ def getDataset(self, id):
             # the corresponding sqlalchemy.core.Column entry to index the result
             # because the name of the key may not be the name of the name of the
             # DataUnit link.
-            dataId = {dataUnitName: result[self._schema.dataUnits[dataUnitName]]
-                      for dataUnitName in datasetType.dataUnits}
+            dataId = {dataUnitName: result[self._schema.dataUnits.links[dataUnitName]]
+                      for dataUnitName in self._schema.dataUnits.getPrimaryKeyNames(datasetType.dataUnits)}
```
TallJimbo, line 218: This should probably be deferred to another ticket if we do it at all, but this seems like the wrong time and place to merge the primary key names for all of the DataUnits associated with a `DatasetType`; could we do that instead when the `DatasetType` instance is constructed?

That was, of course, part of the motivation behind having a custom `DataUnitTypeSet` class for `DatasetType.dataUnits`, though we don't have to do it that way.

tests/test_butler.py
TallJimbo: I've been wondering about this, and I think the best approach is probably to be case (and underscore) insensitive, at least for data ID dictionary keys - the same words appear as both DataUnit type names (upper/camel case) and database field names (lower/snake case), and I think it'll be frustrating if we don't accept both in the dictionaries where the keys effectively refer to both.

That's probably out of scope for this ticket, though.

tests/test_sqlRegistry.py
TallJimbo: Note that these are slightly different contexts; this one probably should be case-sensitive and require upper/camel case, since it's definitely a DataUnit type, not a field in a DataUnit table.  The one a few lines later is where I think we ought to be case insensitive (as discussed in the comment at https://github.com/lsst/daf_butler/pull/34/files#diff-f06c7b82277ef9b866c66ef83c24c846R82), but probably do so by translating user inputs to lower/snake case.

TallJimbo (approved):
Looks good; nothing looks awry from a structural perspective, though I do have a few design comments and of course a couple of style comments.

python/lsst/daf/butler/core/__init__.py
pschella: It is not clear to me that we want to pull up all these submodules into the `core` scope. But as long as we do that we should be consistent.

python/lsst/daf/butler/core/dataUnit.py
pschella: No reason. I'll change it.

python/lsst/daf/butler/core/schema.py
pschella: Don't know. I assume it can move back up.

tests/test_butler.py
pschella: Yes, this was tricky for me to decide. Right now `DataUnit` names are `CamelCase` and column names / dataId dict keys are `snake_case`. But it is probably nicer to allow both.

tests/test_sqlRegistry.py
pschella: Hmm, ok then I missunderstand what you say above. The only place `CamelCase` is used is when referring to either `DataUnit` types or their tables. All other cases are `snake_case` because they map directly to column names in said tables. So for now I'll leave it as is and will make a separate ticket to decide / fix this.

python/lsst/daf/butler/core/dataUnit.py
pschella: Well, the comment might, but the structure doesn't as far as I think of it now.

pschella: I can change the comment if that is better. I assume one can duck-type whatever entity represents a view in `sqlalchemy` with the few things required (`columns`) in our API.

python/lsst/daf/butler/core/dataUnit.py
pschella: Yes, I don't like the name either. But I couldn't think of a better one. Also note that we have a number of similar (but not identical enough) things in the design that basically act (or could act) like a `map` of some kind. There may be multiple instances, although I don't know a need for it at this moment. Perhaps when there are multiple Butler instances with different units? I don't know.

python/lsst/daf/butler/core/dataUnit.py
pschella: Good question. I suspect in this case it does the same (since it is not part of a nested for loop). I'll change it to `iter`.

python/lsst/daf/butler/core/schema.py
pschella: I only did that such that we could add more `DataUnit` tables later. But indeed, it might make more sense to separate it. Building is basically a one-off thing.

python/lsst/daf/butler/registries/sqlRegistry.py
pschella: Yes, I like your suggestion to refactor it. This was a bit quick and dirty. No need to build this over and over again. I'll file a ticket (possibly resurrecting `DataUnitTypeSet`).

python/lsst/daf/butler/core/__init__.py
timj: If you don't they won't be documented (as currently set up)

PR #35, opened 2018-04-30
DM-14225: Make PosixDatastore's internal records persistent

TallJimbo:
The approach I've taken here is:
 - Add a `DatabaseDict` ABC that just provides a mechanism for constructing persistent dict-like objects from Config.
 - Add a SQL implementation thereof, and a way to get an instance of that SQL-backed dict from a Registry.
 - Replace the `internalRegistry` in PosixDatastore with DatabaseDict it constructs from config.

Because `DatabaseDict` uses `namedtuple` objects for its value type (since, being backed by a database, that needs to be known up front), there's a bit of redundancy between `StoredFileInfo` and the new `namedtuple` `PosixDatastore.RecordTuple` (basically, they only differ on whether to hold a `StorageClass` instance or name).  I imagine we'll want to address that eventually, but just letting them both exist for now minimizes the changes to `PosixDatastore`, and I think that's a good thing.

### DIFF
--- a/python/lsst/daf/butler/core/databaseDict.py
+++ b/python/lsst/daf/butler/core/databaseDict.py
@@ -1,0 +1,105 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from collections.abc import MutableMapping
+
+from .utils import doImport
+
+
+class DatabaseDict(MutableMapping):
+    """An abstract base class for dict-like objects with a specific key type
+    and namedtuple values, backed by a database.
+
+    DatabaseDict subclasses must implement the abstract ``__getitem__``,
+    ``__setitem__``, ``__delitem__`, ``__iter__``, and ``__len__`` abstract
+    methods defined by `MutableMapping`.
+
+    They must also provide a constructor that takes the same arguments as that
+    of `DatabaseDict` itself, *unless* they are constructed solely by
+    `Registry.makeDatabaseDict` (in which case any constructor arguments are
+    permitted).
+
+    Parameters
+    ----------
+    config : `Config`
+        Configuration used to identify and construct a subclass.
+    types : `dict`
+        A dictionary mapping `str` field names to type objects, containing
+        all fields to be held in the database.
+    key : `str`
+        The name of the field to be used as the dictionary key.  Must not be
+        present in ``value._fields``.
+    value : `type`
+        The type used for the dictionary's values, typically a `namedtuple`.
+        Must have a ``_fields`` class attribute that is a tuple of field names
+        (i.e. as defined by `namedtuple`); these field names must also appear
+        in the ``types`` arg, and a `_make` attribute to construct it from a
+        sequence of values (again, as defined by `namedtuple`).
+    """
+
+    @staticmethod
+    def fromConfig(config, types, key, value, registry=None):
+        """Create a `DatabaseDict` subclass instance from `config`.
+
+        If ``config`` contains a class ``cls`` key, this will be assumed to
+        be the fully-qualified name of a DatabaseDict subclass to construct.
+        If not, ``registry.makeDatabaseDict`` will be called instead.
+
+        Parameters
+        ----------
+        config : `Config`
+            Configuration used to identify and construct a subclass.
+        types : `dict`
+            A dictionary mapping `str` field names to type objects, containing
+            all fields to be held in the database.
+        key : `str`
+            The name of the field to be used as the dictionary key.  Must not
+            be present in ``value._fields``.
+        value : `type`
+            The type used for the dictionary's values, typically a
+            `namedtuple`. Must have a ``_fields`` class attribute that is a
+            tuple of field names (i.e. as defined by `namedtuple`); these
+            field names must also appear in the ``types`` arg, and a `_make`
+            attribute to construct it from a sequence of values (again, as
+            defined by `namedtuple`).
+        registry : `Registry`
+            A registry instance from which a `DatabaseDict` subclass can be
+            obtained.  Ignored if ``config["cls"]`` exists; may be None if
+            it does.
+
+        Returns
+        -------
+        dictionary : `DatabaseDict` (subclass)
+            A new `DatabaseDict` subclass instance.
+        """
+        if 'cls' in config:
+            cls = doImport(config['cls'])
+            return cls(config=config, types=types, key=key, value=value)
+        else:
+            table = config['table']
+            if registry is None:
+                raise ValueError("Either config['cls'] or registry must be provided.")
+            return registry.makeDatabaseDict(table, types=types, key=key, value=value)
+
+    def __init__(self, config, types, key, value):
+        # This constructor is currently defined just to clearly document the
+        # interface subclasses should conform to.
+        pass
--- a/python/lsst/daf/butler/core/databaseDict.py
+++ b/python/lsst/daf/butler/core/databaseDict.py
@@ -61,7 +61,9 @@ class DatabaseDict(MutableMapping):
 
         If ``config`` contains a class ``cls`` key, this will be assumed to
         be the fully-qualified name of a DatabaseDict subclass to construct.
-        If not, ``registry.makeDatabaseDict`` will be called instead.
+        If not, ``registry.makeDatabaseDict`` will be called instead, and
+        ``config`` must contain a ``table`` key with the name of the table
+        to use.
 
         Parameters
         ----------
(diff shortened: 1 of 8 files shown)


tests/test_sqlDatabaseDict.py
```diff
@@ -0,0 +1,123 @@
...
+        """Test when there are fields not in the value or the key."""
+        value = namedtuple("TestValue", ["x", "y"])
+        data = {
+            0: value(x=0, y="zero"),
+            1: value(x=1, y="one"),
+        }
+        d = DatabaseDict.fromConfig(self.config, key=self.key, fields=self.fields, value=value)
+        self.checkDatabaseDict(d, data)
+
+    def testFromRegistry(self):
```
timj: Missing docstring.

tests/test_sqlDatabaseDict.py
```diff
@@ -0,0 +1,123 @@
...
+        value = namedtuple("TestValue", ["y", "z"])
+        data = {
+            0: value(y="zero", z=0.0),
+            1: value(y="one", z=0.1),
+        }
+        d = DatabaseDict.fromConfig(self.config, key=self.key, fields=self.fields, value=value)
+        self.checkDatabaseDict(d, data)
+
+    def testExtraFields(self):
+        """Test when there are fields not in the value or the key."""
```
timj: I don't understand this docstring. `x` and `y` are allowed so I don't see what "ExtraFields" means. "Extra" implies to me you define a `namedtuple` with `a` and `b` in it and try to store that when the `DatabaseDict` only expects `x`, `y` and `z`.

tests/test_sqlDatabaseDict.py
```diff
@@ -0,0 +1,123 @@
...
+        with self.assertRaises(KeyError):
+            del d[0]
+        with self.assertRaises(KeyError):
+            d[0]
+
+    def testKeyInValue(self):
+        """Test when the value includes all fields, including the key."""
+        value = namedtuple("TestValue", ["x", "y", "z"])
+        data = {
+            0: value(x=0, y="zero", z=0.0),
```
timj: I'd like a test where the datatypes don't match those used in `self.fields`.

tests/test_sqlDatabaseDict.py
```diff
@@ -0,0 +1,123 @@
...
+    def testFromRegistry(self):
+        testDir = os.path.dirname(__file__)
+        configFile = os.path.join(testDir, "config/basic/butler.yaml")
+        registry = Registry.fromConfig(configFile)
+        value = namedtuple("TestValue", ["x", "y", "z"])
+        data = {
+            0: value(x=0, y="zero", z=0.0),
+            1: value(x=1, y="one", z=0.1),
+        }
+        d = registry.makeDatabaseDict(table="TestTable", key=self.key, fields=self.fields, value=value)
```
timj: Call this table a different name to ensure that nothing odd is happening relating to the config version?

python/lsst/daf/butler/core/databaseDict.py
```diff
@@ -0,0 +1,101 @@
...
+        Configuration used to identify and construct a subclass.
+    fields : `dict`
+        A dictionary mapping `str` field names to type objects, containing
+        all fields to be held in the database.
+    key : `str` or `tuple`
+        The name of the field to be used as keys in the dictionary, or
+        a tuple thereof.
+    value : `type`
+        The type used for the dictionary's values, typically a `namedtuple`.
+        Must have a ``_field`` class attribute that is a tuple of field
```
timj: Is this `_fields` rather than `_field`? I found this all a bit confusing until I worked out that `_fields` and `_make` came from `namedtuple`.

python/lsst/daf/butler/core/databaseDict.py
```diff
@@ -0,0 +1,101 @@
...
+        -------
+        dictionary : `DatabaseDict` (subclass)
+            A new `DatabaseDict` subclass instance.
+        """
+        if 'cls' in config:
+            cls = doImport(config['cls'])
+            return cls(config=config, fields=fields, key=key, value=value)
+        else:
+            table = config['table']
+            return registry.makeDatabaseDict(table, fields=fields, key=key, value=value)
```
timj: maybe check if `registry is not None`?

python/lsst/daf/butler/core/sqlDatabaseDict.py
```diff
@@ -0,0 +1,135 @@
...
+            try:
+                if self._key not in self._value._fields:
+                    kwds[self._key] = key
+                connection.execute(self._table.insert(), **kwds)
+                return
+            except IntegrityError:
+                pass
+        # If we fail due to an IntegrityError (i.e. duplicate primary key values),
+        # try to do an update instead.
+        kwds.discard(self._key)
```
timj: Is there a test for this? I think `kwds` is a dict (`OrderedDict` probably) and that does not have a `discard()` method. I think this should be `pop()` or `del`.

python/lsst/daf/butler/core/sqlDatabaseDict.py
```diff
@@ -0,0 +1,135 @@
...
+            return self._value._make(row)
+
+    def __setitem__(self, key, value):
+        assert isinstance(value, self._value)
+        # Try insert first, as we expect that to be the most commmon usage
+        # pattern.
+        kwds = value._asdict()
+        with self._engine.begin() as connection:
+            try:
+                if self._key not in self._value._fields:
```
timj: Should it be documented that if the key field is found in value, that the supplied key will be ignored? Should we test if key is supplied and it's different to the key field from value?

python/lsst/daf/butler/core/sqlDatabaseDict.py
```diff
@@ -0,0 +1,135 @@
...
+                    kwds[self._key] = key
+                connection.execute(self._table.insert(), **kwds)
+                return
+            except IntegrityError:
+                pass
+        # If we fail due to an IntegrityError (i.e. duplicate primary key values),
+        # try to do an update instead.
+        kwds.discard(self._key)
+        with self._engine.begin() as connection:
+            connection.execute(self._updateSql, key=key, **kwds)
```
timj: Is it inconsistent that set could use the internal key from value but if that fails update is done here with the supplied key?

python/lsst/daf/butler/core/databaseDict.py
```diff
@@ -0,0 +1,101 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from collections.abc import MutableMapping
+
+from .utils import doImport
+
+
+class DatabaseDict(MutableMapping):
+    """An abstract base class for dict-like objects backed by a database.
```
timj: "dict-like objects" implies something a bit simpler than what this generally is. It's not simple key/value pairs in a dict that are persisted to a database, it's more of a python model of a database table. I feel like we are close to reimplementing an astropy.table or afw table database serialization system.

python/lsst/daf/butler/core/sqlDatabaseDict.py
```diff
@@ -0,0 +1,135 @@
...
+    def __delitem__(self, key):
+        with self._engine.begin() as connection:
+            result = connection.execute(self._delSql, key=key)
+            if result.rowcount == 0:
+                raise KeyError("{} not found".format(key))
+
+    def __iter__(self):
+        with self._engine.begin() as connection:
+            for row in connection.execute(self._keysSql).fetchall():
+                yield row[0]
```
timj, line 143: I'm a bit surprised that sqlalchemy doesn't have a built in iterator return. On the other hand from the docs for `fetchall()` I have no idea what it really returns.

timj (approved):
I think there needs to be a few more tests. I've made some comments about that and I think there is a bad method call discard vs del. I was initially confused by key/value being everywhere but I think I've got the hang of it.

python/lsst/daf/butler/core/databaseDict.py
TallJimbo: Yeah, "dict-like object" would be both simpler and more flexible than than what this.  But from the API perspective I think the big difference is the strong types for the key and value, especially the `namedtuple` values.  The implementation in `SqlDatabaseDict` is getting a bit towards a Python model of a database table, but the very limited dict-like lookup we support is  still a lot simpler.  I'll update the docstring with a focus on the typing and use of `namedtuples`.

python/lsst/daf/butler/core/databaseDict.py
TallJimbo: Yes, `_fields`.  I've updated the documentation to clarify where `_fields` and `_make` come from.

python/lsst/daf/butler/core/sqlDatabaseDict.py
TallJimbo: I suppose this question illustrates that we don't actually need to support the case where the value tuple includes the key.  I think it'd simplify the code and the tests to just reject that possibility at construction.

python/lsst/daf/butler/core/sqlDatabaseDict.py
TallJimbo: Good catch - yes, the behavior I was looking for would be `kwds.pop(self._key, None)`, and it does need a test.

python/lsst/daf/butler/core/sqlDatabaseDict.py
TallJimbo: Another reason I'm glad I removed support for having the key duplicated in the value.

python/lsst/daf/butler/core/sqlDatabaseDict.py
TallJimbo: I think it does have a built-in iterator that yields full rows, but I think it does need to be adapted as above to obtain an iterator over a single column from each row (even if the row only has one column).

tests/test_sqlDatabaseDict.py
TallJimbo: This test is to check that it's okay if the table has additional fields the DatabaseDict is not expected to set or retrieve.  I've renamed it to `testExtraFieldsInTable`.  I'll add a new `testExtraFieldsInValue` method for the case you described.

TallJimbo (commented):
I believe I've addressed all review comments.  I also renamed the "fields" argument to "types", since its purpose is really to provide the types of the fields provided in the "key" and "value".  That makes the docs a lot clearer.

timj:
Looks great. Much clearer now. Thank you.

PR #36, opened 2018-05-04
DM-14321: Make StorageClass pickleable (by making instances instead of subclasses)

pschella:


timj (approved):
Certainly looks significantly easier to understand what's going on.

PR #37, opened 2018-05-07
DM-13452: Few small fixes needed for SuperTask/QuantumGraph

andy-slac:
Reincarnation of #19 PR after some changes - I dropped changes to `units` as that module disappeared. `Quantum` changes are still there, and it is simple addition of `outputs` DatasetRefs. All comments regarding Quantum from old PR were taken into account. @pschella already said that Quantum changes look OK, so I'll probably just merge this unless I get more comments today-tomorrow.

### DIFF
--- a/python/lsst/daf/butler/core/quantum.py
+++ b/python/lsst/daf/butler/core/quantum.py
@@ -42,7 +42,7 @@ class Quantum(Execution):
         The Run this Quantum is a part of.
     """
 
-    __slots__ = ("_task", "_run", "_predictedInputs", "_actualInputs")
+    __slots__ = ("_task", "_run", "_predictedInputs", "_actualInputs", "_outputs")
     __eq__ = slotValuesAreEqual
 
     def __init__(self, task, run, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


python/lsst/daf/butler/core/quantum.py
```diff
@@ -93,11 +94,25 @@ def actualInputs(self):
         """
         return self._actualInputs
 
+    @property
+    def outputs(self):
+        """A `dict` of output datasets (to be) generated for this quantum,
+        with the same form as `predictedInputs`.
+
+        Read-only; update via `addOutput()`.
```
timj, line 102: I think you may well need two backticks here for numpydoc because it (and `predictedInputs` text above) are not classes. I think the alternative is to write something like `~lsst.daf.core.butler.Quantum.addOutput()` (or whatever); ie give a fully qualified method name and the `~` means to truncate the output to the final element. I haven't tested this though.

andy-slac: BTW, I almost literally copied docstring from other method in this class, should I change it everywhere?

timj: Ah. Leave it then. I'll fix it next time I try to run the doc gen and see all the warnings appear.

PR #39, opened 2018-05-09
DM-14191: Merge butler configuration files with defaults

timj:
* Also rewrite `doImport`


python/lsst/daf/butler/core/butlerConfig.py
```diff
@@ -0,0 +1,104 @@
...
+        self.update(butlerConfig)
+
+    def _validate(self, config=None):
+        """Check a butler config contains mandatory keys.
+
+        Parameters
+        ----------
+        config : `Config`, optional
+            By default checks itself, but if ``config`` is given, this
+            config will be checked instead.
```
pschella: Better to make `validate` a public class method instead?

python/lsst/daf/butler/core/datastore.py
```diff
@@ -27,13 +27,14 @@
...
+from .config import ConfigSubset
 
 __all__ = ("DatastoreConfig", "Datastore")
 
 
-class DatastoreConfig(Config):
-    pass
+class DatastoreConfig(ConfigSubset):
+    component = "datastore"
+    requiredKeys = ("cls",)
```
pschella, line 37: Isn't this check now duplicated in `Butler(Config)`?

python/lsst/daf/butler/core/registry.py
```diff
@@ -57,13 +62,11 @@ def fromConfig(config):
...
-            if isinstance(config, str):
-                config = Config(config)
-            if isinstance(config, Config):
-                config = RegistryConfig(config['registry'])
+            if isinstance(config, str) or isinstance(config, Config):
+                config = RegistryConfig(config)
             else:
                 raise ValueError("Incompatible Registry configuration: {}".format(config))
-        cls = doImport(config['cls'])
+        cls = doImport(config['registry.cls'])
```
pschella: If `config` now is a `RegistryConfig` doesn't it have just the `registry` subset?

pschella: Ah, I see, because we want the full Butler config for it. Maybe then explicitly specify that?

python/lsst/daf/butler/core/butlerConfig.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def __init__(self, other):
+
+        # Create an empty config for us to populate
+        super().__init__()
+
+        # Find the butler configs
+        self.defaultsDir = None
+        if "DAF_BUTLER_DIR" in os.environ:
+            self.defaultsDir = os.path.join(os.environ["DAF_BUTLER_DIR"], "config")
```
TallJimbo: `utils.getPackageDir`?

python/lsst/daf/butler/core/butlerConfig.py
```diff
@@ -0,0 +1,104 @@
...
+
+    The configuration is read and merged with default configurations for
+    the particular classes. The defaults are read from
+    ``$DAF_BUTLER_DIR/config`` using names specified by the appropriate classes
+    for registry and datastore, and including the standard schema and storage
+    class definitions.
+
+    Parameters
+    ----------
+    other : `str`
```
TallJimbo: Might want to rename this argument to something more descriptive.

config/datastores/posixDatastore.yaml
```diff
@@ -0,0 +1,5 @@
+datastore:
+  cls: lsst.daf.butler.datastores.posixDatastore.PosixDatastore
+  records:
+    table: PosixDatastoreRecords
+  create: true
```
TallJimbo: I'm a bit surprised to not see the default mapping from StorageClass to Formatter here.  Those are specific for PosixDatastore, right?

I don't actually care if they go here or elsewhere, but we need an easy way to construct a Butler that knows about those defaults.

python/lsst/daf/butler/core/butlerConfig.py
```diff
@@ -0,0 +1,104 @@
...
+        # defaults to use.
+        butlerConfig = Config(other)
+
+        # Check that fundamental keys are present
+        self._validate(butlerConfig)
+
+        # Look for class specific defaults
+        for section in ("datastore", "registry"):
+            k = f"{section}.cls"
+            print("Checking {}: {}".format(k, butlerConfig[k]))
```
TallJimbo: Remove or use logging.

python/lsst/daf/butler/core/butlerConfig.py
```diff
@@ -0,0 +1,104 @@
...
+        if "DAF_BUTLER_DIR" in os.environ:
+            self.defaultsDir = os.path.join(os.environ["DAF_BUTLER_DIR"], "config")
+
+        # Storage classes
+        storageClasses = Config(os.path.join(self.defaultsDir, "storageClasses.yaml"))
+        self.update(storageClasses)
+
+        # Default schema
+        schema = Config(os.path.join(self.defaultsDir, "registry", "default_schema.yaml"))
+        self.update(schema)
```
TallJimbo: Should we be using `RegistryConfig` here?  ("No" is a perfectly good answer; I'm just curious why not.)

TallJimbo (approved):
Looks good; mostly very minor comments.  I would like to make sure we figure out where to put the StorageClass->Formatter mapping defaults so those don't have to be repeated in each repository (and that raises the question of which config values, if any, should always go in the repo config instead of being pulled from the defaults).

pschella:
Since they are all defaults we may want to rename `default_schema.yaml` to something else.

python/lsst/daf/butler/core/butlerConfig.py
pschella: I still wonder if having just one level of overrides is future proof enough. As opposed to a (colon-separated) `$DAF_BUTLER_CONFIG_PATH`.

timj: I can do that. I thought we agreed that once we have the merging code in one place we can easily add it but since I'm there I'll add this other variable as well.

python/lsst/daf/butler/core/datastore.py
timj: ButlerConfig doesn't inherit from ConfigSubset so no. I could move the validation code in `ConfigSubset.__init__()` to `Config.__init__()` and always have it as an option if you set `requiredKeys`.

python/lsst/daf/butler/core/butlerConfig.py
timj: See my comment below on validation of `ConfigSubset` but would you prefer class attributes `Config.requiredKeys` and a public method in `Config.validate` which used `requiredKeys` but could be overridden? Why a class method and not an instance method?

python/lsst/daf/butler/core/registry.py
timj: Right. I think it's better for RegistryConfig to only have registry stuff and SchemaConfig to only have schema stuff. I'll change that.

python/lsst/daf/butler/core/butlerConfig.py
timj: I called it that because that's what it was called in `Config` (which came from `daf_persistence`). I'll have a think.

config/datastores/posixDatastore.yaml
timj: I wasn't entirely sure what defaults would mean since the formatters really do lock in the file format, but you are right that FITS is where we are so our default for "how to write an Exposure" should be in this file.

python/lsst/daf/butler/core/butlerConfig.py
timj: Yes. Sorry. Should have deleted it.

python/lsst/daf/butler/core/butlerConfig.py
timj: Probably. Before I added `ConfigSubset` there wasn't any difference. Now there is so I'll revisit that.

timj: Although, these configs have to fit in the global structure so should not have the `schema` stripped. The advantage of using `SchemaConfig` is you get the validation so I have to see if I can read the subset and bind it to the larger config hierarchy.

python/lsst/daf/butler/core/butlerConfig.py
timj: I wondered about that. `utils` would work fine but I do wonder about adding LSST dependencies for no real gain. We only use `utils` in tests and I would argue that we don't need that dependency. That leaves only `lsst.log` and `sphgeom` stopping us from putting this package directly on pypi for everyone to use without needing any non-standard dependencies. (I'm thinking about world domination here...)

timj:
Regarding `default_schema.yaml` I'm happy to delete the `registry` directory and call the file `registry_schema.yaml`.

python/lsst/daf/butler/core/butlerConfig.py
timj: Do you envisage the `$DAF_BUTLER_CONFIG_PATH` containing a set of search paths that look identical to the on in `daf_butler/config` (so we look for `storageClasses.yaml`, etc)? And that the defaults should be filled in by reading butler configs first, then the path in reverse order?

TallJimbo: I have some of the same thoughts about world domination (see RFC-430 and RFC-460 vis a vis getting that sphgeom dependency out on pypi).  But I think using `DAF_BUTLER_DIR` here just avoids an explicit dependency on `utils` via an implicit dependency on `EUPS` and perhaps `sconsUtils`, and I think the `utils` version may be slightly easier to rip out and replace with something unencumbered later.

TallJimbo: I think we should hold off on this until we have a use case for it.  I tend to think that having extra paths that aren't connected to either a repo or a software version could do more harm than good.

timj: I implemented this last night... How about I push the change this morning as a new commit and you can take a look.

config/datastores/posixDatastore.yaml
TallJimbo: A lot of those defaults currently live in the `gen3` module of `ci_hsc` (on the `gen3-middleware` branch).  I'd love to get those moved out of Python code into YAML.  I'd also be happy to take care of making a ci_hsc branch compatible with this ticket once you're done with the daf_butler changes.

python/lsst/daf/butler/core/butlerConfig.py
```diff
@@ -51,54 +63,64 @@ def __init__(self, other):
...
+
+        # Search each directory for each of the default files
+        for pathDir in self.defaultsPaths:
+            for configClass, configFile in defaultsFiles:
+                # Assume external paths have same config files as global config
+                # directory. Absolute paths are possible for external
+                # code.
+                # Should this be a log message? Are we using lsst.log?
+                # print("Checking path {} for {} ({})".format(pathDir, configClass, configFile))
+                if not os.path.isabs(configFile):
```
ktlim: Doesn't this apply absolute paths over and over?  Maybe that's not a problem.

timj: Yes. It does. Hadn't thought of that. I don't think it's a problem as such, but it will slow things down.

python/lsst/daf/butler/core/datastore.py
pschella: That sounds good.

python/lsst/daf/butler/core/butlerConfig.py
```diff
@@ -0,0 +1,126 @@
...
+class ButlerConfig(Config):
+    """Contains the configuration for a `Butler`
+
+    The configuration is read and merged with default configurations for
+    the particular classes. The defaults are read from
+    ``$DAF_BUTLER_DIR/config`` and ``$DAF_BULTER_CONFIG_PATH``. The defaults
+    are constructed by reading first the global defaults, and then adding
+    in overrides from each entry in the colon-separated
+    ``$DAF_BUTLER_CONFIG_PATH`` in reverse order such that the entries ahead
+    in the list take priority. The registry and datastore configurations
```
pschella: Is this opposite behavior to shell `$PATH`? If so why?

python/lsst/daf/butler/core/butlerConfig.py
```diff
@@ -0,0 +1,126 @@
...
+
+        # This is a list of all the config files to search
+        defaultsFiles = [(c, c.defaultConfigFile) for c in configComponents]
+
+        # Components that derive their configurations from implementation
+        # classes rather than configuration classes
+        specializedConfigs = (DatastoreConfig, RegistryConfig)
+
+        # Look for class specific default files to be added to the file list
+        # These class definitions must come from the suppliedbutler config and
```
pschella: supplied butler.

python/lsst/daf/butler/core/butlerConfig.py
```diff
@@ -0,0 +1,126 @@
...
+        # are not defaulted.
+        for componentConfig in specializedConfigs:
```
pschella: Why not simply loop over all components and have those subclasses determine if they are special?

pschella: E.g. by having a `component` member.

python/lsst/daf/butler/core/butlerConfig.py
```diff
@@ -0,0 +1,126 @@
...
+
+        # Create an empty config for us to populate
+        super().__init__()
+
+        # Read the supplied config so that we can work out which other
+        # defaults to use.
+        butlerConfig = Config(other)
+
+        # All the configs that can be associated with default files
+        configComponents = (SchemaConfig, StorageClassConfig, DatastoreConfig, RegistryConfig)
```
pschella: Perhaps make this a class member?

python/lsst/daf/butler/core/butlerConfig.py
pschella: Why not simply invert the loop order?

python/lsst/daf/butler/core/config.py
```diff
@@ -94,7 +94,7 @@ def __init__(self, other=None):
     def ppprint(self):
         """helper function for debugging, prints a config out in a readable way in the debugger.
 
-        use: pdb> print myConfigObject.pprint()
+        use: pdb> print myConfigObject.ppprint()
```
pschella: `ppprint`?

python/lsst/daf/butler/core/config.py
```diff
@@ -328,3 +328,45 @@ def dumpToFile(self, path):
...
+    requiredKeys = ()
+    """Keys that are required to be specified in the configuration.
+    """
+
+    defaultConfigFile = None
+    """Name of the file containing defaults for this config class.
+    """
+
+    def __init__(self, other=None):
+        super().__init__(other)
```
pschella: What if `other` is a `ConfigSubset`?

tests/test_dataUnit.py
```diff
@@ -33,7 +34,8 @@ class DataUnitRegistryTestCase(lsst.utils.tests.TestCase):
     """
     def setUp(self):
         self.testDir = os.path.dirname(__file__)
-        self.schemaFile = os.path.join(self.testDir, "../config/registry/default_schema.yaml")
+        self.schemaFile = os.path.join(lsst.utils.getPackageDir("daf_butler"), "config",
+                                       SchemaConfig.defaultConfigFile)
         self.config = SchemaConfig(self.schemaFile)
```
pschella: Do we still need to specify the `schemaFile`? Or does calling `SchemaConfig()` load the default file?

tests/test_instrument.py
```diff
@@ -51,7 +52,7 @@ def setUp(self):
         self.configFile = os.path.join(self.testDir, "config/basic/butler.yaml")
```
pschella, line 52: Should we have a default file for `Butler` too perhaps?

pschella: If only for users to copy. But it might be useful to do `butler = Butler()` and simply get something working with a local `SqliteRegistry` and `PosixDatastore` in a default local subdirectory.

pschella: Perhaps even more useful if that then can write the default `butler.yaml` config in it (something like `git init .`) that you can subsequently modify.

pschella: But maybe the default constructor is not the best for that. And instead you want an explicit class method e.g. `butler = Butler.makeDefault()` (with an optional path).

tests/test_posixDatastoreFits.py
```diff
@@ -132,10 +132,10 @@ def testTransfer(self):
         ref = self.makeDatasetRef("calexp", dataUnits, storageClass, dataId)
 
         inputConfig = DatastoreConfig(self.configFile)
-        inputConfig['datastore.root'] = os.path.join(self.testDir, "./test_input_datastore")
+        inputConfig['root'] = os.path.join(self.testDir, "./test_input_datastore")
         inputPosixDatastore = PosixDatastore(config=inputConfig, registry=self.registry)
         outputConfig = inputConfig.copy()
-        outputConfig['datastore.root'] = os.path.join(self.testDir, "./test_output_datastore")
+        outputConfig['root'] = os.path.join(self.testDir, "./test_output_datastore")
```
pschella, line 138: Are we now assuming that `Datastore` and `Registry` always have the same `root`? What if (for a particular datastore) there is no such thing?

python/lsst/daf/butler/core/config.py
TallJimbo: I think `pdb` requires parentheses after `print` these days.

python/lsst/daf/butler/core/config.py
```diff
@@ -328,3 +328,45 @@ def dumpToFile(self, path):
...
+
+class ConfigSubset(Config):
+    """Config which can be constructed from a global config but which uses a
+    subset.
+
+    Sometimes you want to instantiate a ``Config`` object but you know that
+    the configuration supplied might contain more keys than you are interested
+    in and you only want one bit of it. For example, your config might contain
+    ``schema`` if it's part of a global config but it might be the contents
+    of ``schema``.
```
TallJimbo: Missing the end of this sentence, I think.

python/lsst/daf/butler/core/config.py
```diff
@@ -328,3 +328,45 @@ def dumpToFile(self, path):
...
+    """
+
+    component = None
+    """Component to use from supplied config. Can be None. If specified the
+    key is not required. Can be a full dot-separated path to a component.
+    """
+
+    requiredKeys = ()
+    """Keys that are required to be specified in the configuration.
+    """
```
TallJimbo, line 390: Are these relative to the top level or the component?  (should be documented)

python/lsst/daf/butler/core/butlerConfig.py
```diff
@@ -0,0 +1,126 @@
...
+                # code.
+                # Should this be a log message? Are we using lsst.log?
+                # print("Checking path {} for {} ({})".format(pathDir, configClass, configFile))
+                if not os.path.isabs(configFile):
+                    configFile = os.path.join(pathDir, configFile)
+                if os.path.exists(configFile):
+                    # this checks a specific part of the tree
+                    # We may need to turn off validation since we do not
+                    # require that each defaults file found is fully
+                    # consistent.
```
TallJimbo: I do think we'll need to defer validation until the full config has been loaded and merged.

python/lsst/daf/butler/core/utils.py
```diff
@@ -145,30 +146,38 @@ def doImport(pythonType):
     ------
     TypeError
         pythonType is not a `str`.
-    ValueError
-        pythonType can not be imported.
+    ModuleNotFoundError
+        No module in the supplied import string could be found.
+    ImportError
+        pythonType is found but can not be imported.
```
TallJimbo: I think it might be best to catch and re-raise any of these as `ImportError`; it's much more likely that the caller will want to catch anything (without having a too-general `except Exception` block) than try to distinguish between them.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -50,14 +50,21 @@ class SqlRegistry(Registry):
...
+    schemaConfig : `SchemaConfig` or `str`
+        Definition of the schema to use.
     """
 
-    def __init__(self, config):
-        super().__init__(config)
+    defaultConfigFile = None
+    """Path to configuration defaults. Relative to $DAF_BUTLER_DIR/config or
+    absolute path. Can be None if no defaults specified.
+    """
```
TallJimbo, line 60: Do we not actually have a default for `SqlRegistry` configuration?

TallJimbo (approved):
Looks good.  I've made some minor comments, but like @pschella, from this point I'm happy for you to merge when you think it's in good shape.
If you do want to merge before moving the StorageClass->Formatter lookup defaults to a non-test config file, I'll go ahead and add those on DM-14378 (there are a number of additional mappings I want to move to daf_butler from ci_hsc anyway, once I have a place to put them in daf_butler).

python/lsst/daf/butler/core/utils.py
timj: `ModuleNotFoundError` is a subclass of `ImportError`. I was pleased to discover that the simplification of `doImport` is only possible because we are using python 3.6 (previous versions did not have `ModuleNotFoundError`). I was trying to keep the interface looking like `import`. I suppose I could convert `AttributeError` to `ImportError` (although I thought I had managed to get `import` to throw that at one point, maybe via `from`).

TallJimbo: Yeah, using a hierarchy of exceptions is the best scenario.  If you can convert the `AttributeError` to `ImportError`, I do think that'd be best, even if that is a bit unlike `import` itself.

tests/test_instrument.py
TallJimbo: I think I'd go with an explicit class method with a required path argument (which could accept `"."`, I suppose).

TallJimbo:
> Regarding default_schema.yaml I'm happy to delete the registry directory and call the file registry_schema.yaml.

Should we just call it `schema.yaml`, so overrides can use the same filename?   (Overrides should be rare, but probably permitted in the same way StorareClass overrides are).

timj:
Everything can be overridden using `$DAF_BUTLER_CONFIG_PATH` if they use the same file name as is in the butler config directory. I'm okay with `schema.yaml`.

python/lsst/daf/butler/core/butlerConfig.py
timj: It's not meant to be the opposite order. In `$PATH` things early in the path take precedence over things later in the path, so I thought I had to reverse the order when reading in the defaults so that earlier ones in the path override defaults appearing late in the path.

python/lsst/daf/butler/core/butlerConfig.py
timj: yes. since the configs read from the paths are subsets of the config I can loop over configs and then paths.

python/lsst/daf/butler/core/butlerConfig.py
timj: Are you saying I should look through all the top level keys in the supplied butler config looking for `.cls` members, then ask the imported classes for their config files and associated config classes? (the latter they don't know but they could).

python/lsst/daf/butler/core/config.py
timj: I didn't write the method...

python/lsst/daf/butler/core/config.py
timj: What am I missing?
```
>>> from lsst.daf.butler import ConfigSubset
>>> x = ConfigSubset({"A": 5, "b": 6})
>>> x
{'A': 5, 'b': 6}
>>> y = ConfigSubset(x)
>>> y
{'A': 5, 'b': 6}
>>> x["c"] = 3
>>> x
{'A': 5, 'b': 6, 'c': 3}
>>> y
{'A': 5, 'b': 6}
```

tests/test_dataUnit.py
timj: At the moment, the only defaulting is in `ButlerConfig`. I can see how each subclass doing its own defaulting and getting the search path from a shared routine might be helpful (and more obvious).

tests/test_posixDatastoreFits.py
timj: I'm not sure I understand. This is a `DatastoreConfig` so by definition it's a datastore config not a registry config. Or are you asking a different question?

python/lsst/daf/butler/core/config.py
timj: Relative to the Config root of this config (which does not include the component name).

python/lsst/daf/butler/core/config.py
```diff
@@ -328,3 +341,201 @@ def dumpToFile(self, path):
...
+        # Create a blank object to receive the defaults
+        # Once we have the defaults we then update with the external values
+        super().__init__()
+
+        # Create a standard Config rather than subset
+        externalConfig = Config(other)
+
+        # Select the part we need from it
+        if self.component is not None and self.component in externalConfig:
+            externalConfig.data = externalConfig.data[self.component]
```
TallJimbo, line 407: Does this mean we merge the wrong `externalConfig` values when these conditions are not met, or does that mean we're assuming that `externalConfig` already corresponds to just the stuff under this component?

python/lsst/daf/butler/core/config.py
```diff
@@ -328,3 +341,201 @@ def dumpToFile(self, path):
...
+        # Find the butler configs
+        defaultsPaths.append(os.path.join(lsst.utils.getPackageDir("daf_butler"), "config"))
+
+        return defaultsPaths
+
+    def _updateWithConfigsFromPath(self, searchPaths, configFile):
+        """Search the supplied paths, merging the configuration values
+
+        The values read will override values currently stored in the object.
+        Every file found in the path will be read. The priority
```
TallJimbo: Orphaned sentence.

TallJimbo (approved):
Looks good; one question and one doc bug.

Also, could you rename `registry_schema.yaml` to just `schema.yaml`?

python/lsst/daf/butler/core/config.py
timj: It means that if you give a config with a root of "datastore" and the config knows it's a datastore then your config will take the contents of the "datastore" part of the tree. If there is no "datastore" in the tree it assumes it's all meant for this config (and will then check for mandatory keys). ie you can give it a datastore.yaml that doesn't have the "datastore" part.

PR #40, opened 2018-05-14
DM-14378: fixes/improvements for issues discovered in ci_hsc conversion testing

TallJimbo:


python/lsst/daf/butler/butler.py
```diff
@@ -61,17 +62,54 @@ class Butler:
...
+            if collection is None:
+                collection = run
+        # n.b. at this point, 'run' is a str or None; 'self.run' is a Run or None
+        if collection is None:  # didn't get a collection from collection or run *args*
+            collection = self.config.get("collection", None)
+            if collection is None:  # didn't get a collection from config['collection']
+                collection = run    # get collection from run found in config
+        if collection is None:
+            raise ValueError("No run or collection provided.")
+        if run is not None and collection != run:
```
pschella: Maybe `runCollection` is a better name for `run` when it actually refers to a collection name associated with a `Run`?

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
```diff
@@ -50,7 +50,7 @@ def _readFile(self, path, pytype):
         if not os.path.exists(path):
             return None
 
-        return pytype.readFits(path)
+        return pytype(path)
```
pschella, line 53: Why is this better? Just curious.

python/lsst/daf/butler/formatters/pexConfigFormatter.py
```diff
@@ -50,7 +51,21 @@ def _readFile(self, path, pytype):
...
+        # because that'd be a huge proliferation of StorageClasses.
+        # Instead, we use a bit of a hack: try to load using the base-class
+        # Config, inspect the exception message to obtain the class we should
+        # have used, import that, and try it instead.
+        # TODO: clean this up, somehow.
+        try:
+            instance.load(path)
+            return instance
+        except AssertionError as err:
+            actualPyTypeStr = str(err).split()[-1]
```
pschella: I suppose the possibility of an `AssertionError` with an unexpectedly differently formatted string being raised is rather low, but still potentially a very confusing error message.

python/lsst/daf/butler/gen2convert/writer.py
```diff
@@ -326,3 +326,12 @@ def insertDatasets(self, registry, datastore):
                     log.debug("Adding Dataset %s as %s in %s", dataset.filePath, gen3id, repo.run)
                     ref = registry.addDataset(datasetType, gen3id, run)
                     datastore.ingest(path=os.path.relpath(dataset.fullPath, start=datastore.root), ref=ref)
+                    refs.append(ref)
+            # Add Datasets to collections associated with any child repos to similate Gen2 parent lookups.
```
pschella: simulate

python/lsst/daf/butler/gen2convert/writer.py
```diff
@@ -326,3 +326,12 @@ def insertDatasets(self, registry, datastore):
...
+            # TODO: only associated parent Datasets with DataUnits associated with DataUnits used by child
+            #       repo Datasets.
```
pschella: ?

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
TallJimbo: I personally think it's worse, but it matches what the afw APIs are and hence actually works.

python/lsst/daf/butler/butler.py
TallJimbo: I'll give it a try, but I'll want to see how it works before committing; I'd like to leave the *argument* (as opposed to the local variable) as just `run` because it actually can be a `Run`.

python/lsst/daf/butler/formatters/pexConfigFormatter.py
TallJimbo: Yeah, I imagine I can at least do a `startswith` test on the error message.

python/lsst/daf/butler/gen2convert/writer.py
TallJimbo: First "associated" should just be "associate".  Will fix.

The problem is this: someone runs a SuperTask that processes one visit, using an input collection that includes all LSST raw data ever taken.  Should their output collection also contain all LSST raw data ever taken (the Gen2 behavior, strictly speaking, and what the converter currently does), or just the raw data of the single visit that was processed (what I think we should do in Gen3, and what this TODO is about).

python/lsst/daf/butler/butler.py
```diff
@@ -49,17 +55,95 @@ class Butler:
...
+            This (mostly) decouples the repository from the default
+            configuration insulating it from changes (which may be good or
+            bad, depending on the nature of the changes).
+
+        Note that when ``expand=False`` (the default), the configuration
+        search path (see `ConfigSubset.defaultSearchPaths`) that was used to
+        construct the repository should also be used to construct any Butlers
+        to it to avoid configuration inconsistencies.
+        """
+        if isinstance(config, ButlerConfig):
```
timj: Maybe also include `ConfigSubset` in this test to be completely sure.

python/lsst/daf/butler/butler.py
```diff
@@ -49,17 +55,95 @@ class Butler:
     ----------
     config : `Config`
         Configuration.
+    collection : `str` or `None`
```
timj: I think the convention is to say `, optional` at the end for keyword args.

timj (approved):

python/lsst/daf/butler/butler.py
```diff
@@ -49,17 +55,95 @@ class Butler:
...
+            If True, write all expanded defaults, not just customized or
+            repository-specific settings.
+            This (mostly) decouples the repository from the default
+            configuration insulating it from changes (which may be good or
+            bad, depending on the nature of the changes).
+
+        Note that when ``expand=False`` (the default), the configuration
+        search path (see `ConfigSubset.defaultSearchPaths`) that was used to
+        construct the repository should also be used to construct any Butlers
+        to it to avoid configuration inconsistencies.
```
pschella: Why does this not return a `Butler` (in which case it should also be a `classmethod`)?

python/lsst/daf/butler/butler.py
```diff
@@ -49,17 +55,95 @@ class Butler:
...
+    ValueError
+        Raised if neither 'collection' nor 'run' are provided by argument or
+        config, or if both are provided and are inconsistent.
     """
 
-    def __init__(self, config=None):
+    @staticmethod
+    def makeRepo(root, config=None, expand=False):
+        """Create an empty data repository by adding a butler.yaml config
+        to a repository root directory.
```
pschella: Does it also make the `root` if it doesn't exist? If not then `initRepo(sitory)` might be a better name? Like in `git init`.

python/lsst/daf/butler/butler.py
```diff
@@ -49,17 +55,95 @@ class Butler:
...
+
+        Parameters
+        ----------
+        root : `str`
+            Filesystem path to the root of the new repository.
+        config : `Config` or `None`
+            Configuration to write to the repository, after setting any
+            root-dependent Registry or Datastore config options.
```
pschella: `Config`? Or `ButlerConfig`?

pschella: If it is `None`, what does it do? Write nothing, or write just the default config?

python/lsst/daf/butler/butler.py
```diff
@@ -49,17 +55,95 @@ class Butler:
...
+        expand : `bool`
```
pschella: `expandDefaults` might be better?

python/lsst/daf/butler/butler.py
```diff
@@ -49,17 +55,95 @@ class Butler:
...
-    def __init__(self, config=None):
+    @staticmethod
+    def makeRepo(root, config=None, expand=False):
+        """Create an empty data repository by adding a butler.yaml config
+        to a repository root directory.
+
+        Parameters
+        ----------
+        root : `str`
+            Filesystem path to the root of the new repository.
```
pschella: Should it exist? What happens if it doesn't, or if it does but isn't empty (or even already is an existing repo)?

python/lsst/daf/butler/butler.py
```diff
@@ -49,17 +55,95 @@ class Butler:
...
+            raise ValueError("makeRepo must be passed a regular Config without defaults applied.")
+        root = os.path.abspath(root)
+        config = Config(config)
+        full = ButlerConfig(config)  # this applies defaults
+        datastoreClass = doImport(full["datastore.cls"])
+        datastoreClass.setConfigRoot(root, config, full)
+        registryClass = doImport(full["registry.cls"])
+        registryClass.setConfigRoot(root, config, full)
+        if expand:
+            config.merge(full)
```
pschella: Isn't this more costly (and potentially error prone) than conditionally dumping `config` or `full`?

python/lsst/daf/butler/butler.py
TallJimbo: It does not currently create the directory, but I think it probably should.  Will fix.

python/lsst/daf/butler/butler.py
TallJimbo: No, `ButlerConfig` is not permitted (and `ConfigSubset` shouldn't be either, as per Tim's comment below).  I'm documenting this in a new `Raises` section of the docs.

If `config is None`, It writes the defaults; now documented.

python/lsst/daf/butler/butler.py
TallJimbo: I still like `expand` better; I think brevity is more valuable for keyword argument names that regular local variables in terms of overall readability.  @timj, want to break the tie? :-)

timj: I think I'd prefer `withDefaults` to `expandDefaults`. You are asking for a fully qualified standalone configuration to be written, rather than the bare minimum based on what you gave in `config` and the root-specific entries. You aren't really expanding as such, you are filling it in to make it explicit. `complete` or `standalone` ?

python/lsst/daf/butler/butler.py
```diff
@@ -49,17 +55,95 @@ class Butler:
...
+    Raises
+    ------
+    ValueError
+        Raised if neither 'collection' nor 'run' are provided by argument or
+        config, or if both are provided and are inconsistent.
     """
 
-    def __init__(self, config=None):
+    @staticmethod
+    def makeRepo(root, config=None, expand=False):
```
timj: I am wondering whether we should add `searchPaths` here to match those I added to the `ConfigSubset` constructor (and which I didn't add to `ButlerConfig` but should). At the very least it helps testing.

python/lsst/daf/butler/registries/sqliteRegistry.py
```diff
@@ -47,5 +47,27 @@ class SqliteRegistry(SqlRegistry):
...
+            A Butler-level config object to update (but not a
+            `ButlerConfig`, to avoid included expanded defaults).
+        full : `ButlerConfig`
+            A complete Butler config with all defaults expanded;
+            repository-specific options that should not be obtained
+            from defaults when Butler instances are constructed
+            should be copied from `full` to `Config`.
+        """
+        config["registry.db"] = "sqlite:///{}/gen3.sqlite3".format(root)
+        for key in ("registry.cls",):
```
timj, line 69: I do wonder whether we could save some code duplication by having these keys as class attributes and then a routine that reads the keys and copies them (and maybe attempts to run `format(root)` on them.

python/lsst/daf/butler/butler.py
TallJimbo: Ooh, I like `standalone`.

python/lsst/daf/butler/butler.py
TallJimbo: I just don't think it's that useful for it to return a `Butler`; usually I imagine this being run by a standalone command-line tool that doesn't do anything else.  Returning the `config` object doesn't cost anything, though, so I'll do that.

python/lsst/daf/butler/butler.py
TallJimbo: `full` doesn't have entries like `datastore.root` and `registry.db` at this stage.

python/lsst/daf/butler/registries/sqliteRegistry.py
TallJimbo: Perhaps a bit, but I'm content to wait until we see what other concrete Datastore/Registry implementations actually need to do here before worrying about them duplicating code.

python/lsst/daf/butler/butler.py
TallJimbo: I'm actually more worried about flexibility in search paths making it easy to use the wrong search path, and I'm content to leave testing of the search path functionality to lower-level tests for now.

PR #42, opened 2018-05-21
DM-14003: Create minimal ShimButler for v2 on top of v3

pschella:


timj:
Is this PR moribund?

TallJimbo:
> Is this PR moribund?

Yes; closing.  I don't think we'll be following this approach to integration unless something goes really badly in PipelineTask-land.

PR #44, opened 2018-05-22
DM-14545: Add test for composite calexp to Butler

pschella:


tests/config/basic/butler.yaml
```diff
@@ -4,7 +4,7 @@ datastore:
   root: ./butler_test_repository
   templates:
     default: "{datasetType}/{tract:?}/{patch:?}/{filter:?}/{camera:?}_{visit:?}"
-    calexp: "{datasetType}.{component:?}/{datasetType}_v{visit}_f{filter}_{component:?}"
+    calexp: "{datasetType}.{component:?}/{datasetType}_v{visit}_f{filter:?}_{component:?}"
```
TallJimbo, line 7: This may be fine for a test, but real-world calexps should have `sensor` and should not *need* to have `filter`.

tests/test_butler.py
```diff
@@ -159,6 +159,33 @@ def testMakeRepo(self):
         self.assertIn("datastore.formatters", full)
         self.assertNotIn("datastore.formatters", limited)
 
+    def testExposureCompositePutGet(self):
```
timj: This test needs to skip if afw does not exist. (I deliberately kept Exposure out of this file) We need to be clear about our optional formatters. The datastore tests are explicitly isolated for optional afw dependency.

tests/test_butler.py
```diff
@@ -159,6 +159,33 @@ def testMakeRepo(self):
...
+        dataId = {"visit": 23, "camera": "HSC"}
+        butler.put(exposure, "calexp", dataId)
+        # Get the full thing
+        full = butler.get("calexp", dataId)  # noqa F841
+        # TODO enable check for equality (fix for Exposure type)
+        # self.assertEqual(full, exposure)
+        # Get a component
+        compsRead = {}
+        for compName in ("wcs", "image", "mask", "coaddInputs", "psf"):
+            component = butler.get("calexp.{}".format(compName), dataId)
```
timj: Use `DatsetType.nameWithComponent()`?

timj (changes requested):
I'd like to ensure that these tests work when afw is missing. See for example `test_posixDatastoreFits.py`.

tests/config/basic/butler.yaml
pschella: The problem with `calexp` was not for the test, but rather for `ci_hsc` where `sensor` does not get passed I think.

pschella: I can remove it here and (temporarily) override it in the `ShimButler` config.

TallJimbo: I don't this config is used at all in the ci_hsc conversion.  The one at `config/datastores/posixDatastore.yaml` is what you need to modify to have any effect on ci_hsc.

pschella: Sorry, I wasn't specific enough, it isn't used in the conversion, but it is used by the `ShimButler`. Anyway, I'll revert the change here.

PR #45, opened 2018-05-22
DM-14543: Fix DatasetType registration

pschella:


python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -117,15 +117,28 @@ def registerDatasetType(self, datasetType):
 
         Raises
         ------
-        KeyError
-            Dataset is already registered.
         ValueError
-            DatasetType is not valid for this registry.
+            DatasetType is not valid for this registry or is already registered
+            but not identical.
```
TallJimbo, line 122: Might be useful to return a bool indicating whether it was inserted or found.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -117,15 +117,28 @@ def registerDatasetType(self, datasetType):
...
         """
+        # Check if datasetType is valid for this Registry
```
TallJimbo: This comment isn't useful as-is; it just repeats what the code below says.  But it might be useful to have a comment that says what "valid for this Registry" means (or at least mentions a way a DatasetType may not be valid).

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -117,15 +117,28 @@ def registerDatasetType(self, datasetType):
...
+            existingDatasetType = self.getDatasetType(datasetType.name)
+        except KeyError:
+            # No registered DatasetType with this name exists, move on to inserting it
+            pass
+        else:
+            # A DatasetType with this name exists, check if is equal
+            if datasetType == existingDatasetType:
+                return
+            else:
+                raise ValueError("DatasetType: {} already registered but different".format(datasetType.name))
```
TallJimbo: I guarantee we'll want an error message that shows how they differ.  Might be a to-do item rather than something to do on this ticket.

python/lsst/daf/butler/registries/sqlRegistry.py
pschella: True, it was more of a remnant of a TODO.

PR #47, opened 2018-05-24
DM-14563: Add command-line script for repo creation.

TallJimbo:
ci_hsc needs to make a new repo, and it's quite a bit more convenient if it can run a command-line script.  I also think this will be useful in other contexts.

### DIFF
--- a/bin.src/makeButlerRepo.py
+++ b/bin.src/makeButlerRepo.py
@@ -1,0 +1,41 @@
+#!/usr/bin/env python
+
+# This file is part of ci_hsc.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


bin.src/makeButlerRepo.py
```diff
@@ -0,0 +1,41 @@
...
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Create an empty Gen3 Butler repository.")
+    parser.add_argument("root",
+                        help=("Filesystem path for the new repository.  "
+                              "Will be created if it does not exist."))
+    parser.add_argument("-c", "--config",
+                        help=("Path to an existing YAML config file to apply (on top of defaults)."))
+    parser.add_argument("--standalone", action="store_true", default=False,
+                        help=("Include all defaults in the config file in the repo, insulating  "
+                              "the repo from changes in package defaults."))
```
pschella: Why the two trailing spaces?

pschella (approved):

bin.src/makeButlerRepo.py
TallJimbo: Copy/paste relic.  Fixed.

PR #48, opened 2018-05-25
DM-14467: Make SqlRegistry more transactional

pschella:


python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -68,6 +80,17 @@ def __init__(self, registryConfig, schemaConfig):
...
         self._datasetTypes = {}
+        self._connection = self._engine.connect()
+
+    @contextlib.contextmanager
+    def transaction(self):
+        trans = self._connection.begin()
+        try:
+            yield
+            trans.commit()
+        except Exception:
```
timj: Are we sure that this should not be `BaseException`? If I do a Ctrl-C what happens?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -68,6 +80,17 @@ def __init__(self, registryConfig, schemaConfig):
...
         self._schema.metadata.create_all(self._engine)
         self._datasetTypes = {}
+        self._connection = self._engine.connect()
+
+    @contextlib.contextmanager
+    def transaction(self):
+        trans = self._connection.begin()
+        try:
+            yield
+            trans.commit()
```
timj, line 90: Should the commit be in a `finally` block?

timj (approved):
Looks like a good start.

python/lsst/daf/butler/registries/sqlRegistry.py
pschella: No, because we don't want to commit in case an exception is raised.

timj: I'm sorry. I meant an `else` block. So that we don't try to run the rollback if the commit itself failed.

PR #49, opened 2018-06-08
DM-14638: Make Datastore optionally transactional

pschella:


python/lsst/daf/butler/core/datastore.py
```diff
@@ -98,6 +160,24 @@ def __init__(self, config, registry):
...
+        Transactions can be nested, and are to be used in combination with
+        `Registry.transaction`.
+        """
+        self._transaction = DatastoreTransaction(self._transaction)
+        try:
+            yield
+        except BaseException:
+            self._transaction.rollback()
+            raise
+        finally:
```
timj: I know I usually get confused by this, but shouldn't this be `else:` rather than `finally`? You don't want the commit to be called after the rollback has run, you only want the commit to happen if the except did not happen.

timj (approved):
Are we sure that `remove` can always be run without itself getting upset if you ask to `remove()` in the rollback when nothing was `put()` in the first place?

PR #50, opened 2018-06-09
DM-14639: Make Butler optionally transactional

pschella:


pschella:
Will be rebased on (and merged into) master once DM-14638 is merged.

pschella:
Note that if we do end up keeping both `SqlDatabaseDict` and `SqlRegistryDatabaseDict` the latter should probably become a subclass of the former.

python/lsst/daf/butler/core/sqlRegistryDatabaseDict.py
```diff
@@ -0,0 +1,143 @@
...
+        assert isinstance(value, self._value)
+        # Try insert first, as we expect that to be the most commmon usage
+        # pattern.
+        kwds = value._asdict()
+        kwds[self._key] = key
+        with self.registry._connection.begin():
+            try:
+                self.registry._connection.execute(self._table.insert(), **kwds)
+                return
+            except IntegrityError:
```
timj, line 109: I think there should be some comment here explaining why `IntegrityError` is perfectly okay to ignore. It seems scary, and you explain later on in the code why it's probably fine, but people are already wondering before they get further down.

timj (approved):

PR #51, opened 2018-06-12
DM-14760: Merge StorageFileInfo into StorageInfo

pschella:


python/lsst/daf/butler/core/storedFileInfo.py
```diff
@@ -58,6 +58,10 @@ def __init__(self, formatter, path, storageClass):
         self._path = path
         assert isinstance(storageClass, StorageClass)
         self._storageClass = storageClass
+        assert checksum is None or isinstance(checksum, str)
```
TallJimbo, line 61: `str` or `bytes`?

python/lsst/daf/butler/core/storedFileInfo.py
```diff
@@ -76,3 +80,19 @@ def storageClass(self):
...
+
+    @property
+    def size(self):
+        """Size of stored object in bytes (`int`).
+        """
+        return self._size
+
+    def __repr__(self):
+        return f"{type(self).__qualname__}(path=\"{self.path}\", formatter=\"{self.formatter}\"" \
+            f" size={self.size}, checksum=\"{self.checksum}\", storageClass=\"{self.storageClass.name}\")"
```
TallJimbo: This is pretty hard to read with all the escaping - seems like a solid reason to use single-quotes in some of these places.

TallJimbo (approved):
Looks good.  Only really minor comments.

python/lsst/daf/butler/core/storedFileInfo.py
timj: Currently `computeChecksum` returns a hex digest as a str. We don't have byte checksums flying around.

PR #52, opened 2018-06-15
DM-14771: Add support for views to Schema

pschella:


TallJimbo:
@pschella, note that I also already started on the dataUnit.py parts of this on DM-14428.  Looks like we went in almost the same direction, and you got farther than I did, so there's certainly no need to roll any of that back.  But once you've got the rest of this working I'd be happy to take over that part - I know you're due to disappear at virtually any moment.

python/lsst/daf/butler/core/sqlalchemy_views.py
```diff
@@ -0,0 +1,71 @@
...
+    metadata : `sqlalchemy.Metadata`
+        Metadata to attach the view to.
+    selectable : `sqlalchemy.sql.select` or `str`
+        The SQL ``SELECT`` clause used to generate the view.
+    """
+    t = table(name)
+
+    if isinstance(selectable, str):
+        selectable = select([text(selectable.lstrip("SELECT"))])
+
```
andy-slac: I wonder if you can do something simple like

    t = alias(selectable, name=name)

instead?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -77,7 +77,7 @@ def __init__(self, registryConfig, schemaConfig):
         self.config = SqlRegistryConfig(registryConfig)
         self.storageClasses = StorageClassFactory()
         self._schema = Schema(schemaConfig)
-        self._engine = create_engine(self.config['db'])
+        self._engine = create_engine(self.config['db'], echo=True)
```
andy-slac: I guess this is for debugging, it should not go into master?

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -150,6 +150,60 @@ def validateId(self, dataId):
...
+    @property
+    def summarizes(self):
+        return self._summarizes
+
+    @property
+    def table(self):
+        """When not ``None`` the primary table entry corresponding to this
+        `DataUnitJoin` (`sqlalchemy.core.TableClause`, optional).
+        """
+        return getattr(self, '_table', None)
```
andy-slac, line 204: Why not just `self._table`?

python/lsst/daf/butler/core/sqlalchemy_views.py
andy-slac: I guess I was thinking about something different here. Ignore my comment if it confuses you :confused:

config/schema.yaml
```diff
@@ -694,19 +694,20 @@ schema:
...
+    #     A calculated join between Datasets identified with an Exposure
+    #     (typically raw science frames) and Datasets identified with ExposureRange
+    #     (typically master calibrations).
+    #   lhs:
+    #   - Exposure
+    #   rhs:
+    #   - ExposureRange
+    #   sql:
+    #     (lhs.camera = rhs.camera) AND
+    #     (lhs.exposure BETWEEN rhs.valid_first AND rhs.valid_last)
```
TallJimbo: Instead of commenting this out, let's just leave this as it was, and ignore it when building the schema for now.  I don't think we will ever want to make it a view; I think we'll instead want to find some way to inject this join expression into preflight queries.

I think since this "sql" entry is not inside a "tables" entry, this shouldn't be parsed by the DataUnitJoin code; if it is, we should fix that instead.

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -150,6 +150,60 @@ def validateId(self, dataId):
...
+    ----------
+    name : `str`
+        Name of this `DataUnit` (`str`, read-only).
+        Also assumed to be the name of the primary table (if present).
+    lhs : `tuple`
+        Left-hand-side of the join.
+    rhs : `tuple`
+        Right-hand-side of the join.
+    summarizes : `DataUnit`
+        Summarizes this other `DataUnit`.
```
TallJimbo: This is a `DataUnitJoin`, not a `DataUnit`, right?

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -275,6 +331,31 @@ def _initDataUnits(self, config, builder):
...
+                    table = builder.addTable(tableName, tableDescription)
+            lhs = dataUnitJoinDescription.get('lhs', None)
+            rhs = dataUnitJoinDescription.get('rhs', None)
+            summarizes = dataUnitJoinDescription.get('summarizes', None)
+            dataUnitJoin = DataUnitJoin(name=dataUnitJoinName,
+                                        lhs=lhs,
+                                        rhs=rhs,
+                                        summarizes=summarizes,
+                                        table=table)
+            self.joins[dataUnitJoinName] = dataUnitJoin
```
TallJimbo: I don't think this needs to be changed here, but as an FYI, on DM-14428 I'll probably switch the keys to be `(lhs, rhs)`,  as I think that's what we'll want to use for lookup most of the time.  Please shout if you think that'll be problematic.

TallJimbo (approved):
Looks good!  A few minor comments.

python/lsst/daf/butler/core/dataUnit.py
pschella: I think that is fine.

config/schema.yaml
pschella: Yes, it is skipped.

pschella: I just commented it out such that the skipping is more explicit.

PR #53, opened 2018-06-15
DM-14334: Add initial implementation of preflight solver.

andy-slac:
This is an incomlpete implementation, it only works on either camera
utits or skymap units but does not know how to join two groups together
(and join tabes/views are missing from registry so I cannot yet start
implementing joins). Renamed makeDataGraph() method into selectDataUnits()
to reflect better what it does. Combining units into quanta is done in
pipe_supertask as it needs to know about pipeline tasks.

### DIFF
--- a/python/lsst/daf/butler/registries/sqlPreFlight.py
+++ b/python/lsst/daf/butler/registries/sqlPreFlight.py
@@ -1,0 +1,188 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+__all__ = ("SqlPreFlight")
+
+import itertools
+import copy
+
+from sqlalchemy.sql import select, and_, text
+
+import lsst.log as lsstLog
+
+
+_LOG = lsstLog.Log.getLogger(__name__)
+
+
+class SqlPreFlight:
+    """Class implementing part of preflight solver which extracts
+    units data from registry.
+
+    This is an implementation detail only to be used by SqlRegistry class,
+    not supposed to be used anywhere else.
+
+    Parameters
+    ----------
+    schema : `Schema`
+        Schema instance
+    connection : `sqlalchmey.Connection`
+        Connection to use for database access.
+    """
+    def __init__(self, schema, connection):
+        self._schema = schema
+        self._connection = connection
+
+    def selectDataUnits(self, collections, expr, neededDatasetTypes, futureDatasetTypes):
+        """Evaluate a filter expression and lists of
+        `DatasetTypes <DatasetType>` and return a set of data unit values.
+
+        Returned set consists of combinations of units participating in data
+        transformation from ``neededDatasetTypes`` to ``futureDatasetTypes``,
+        restricted by existing data and filter expression.
+
+        Parameters
+        ----------
+        collections : `list` of `str`
+            An ordered `list` of collections indicating the Collections to
+            search for Datasets.
+        expr : `str`
+            An expression that limits the `DataUnits <DataUnit>` and
+            (indirectly) the Datasets returned.
+        neededDatasetTypes : `list` of `DatasetType`
+            The `list` of `DatasetTypes <DatasetType>` whose DataUnits will
+            be included in the returned column set. Output is limited by the
+            the Datasets of these DatasetTypes which already exist in the
+            registry.
+        futureDatasetTypes : `list` of `DatasetType`
+            The `list` of `DatasetTypes <DatasetType>` whose DataUnits will
+            be included in the returned column set. It is expected that
+            Datasets for these DatasetTypes do not exist in the registry,
+            but presently this is not checked.
+
+        Returns
+        -------
+        header : `tuple` of `tuple`
+            Length of tuple equals the number of columns in the returned
+            result set. Each item is a tuple with two elements - DataUnit
+            name (e.g. "Visit") and unit value name (e.g. "visit").
+        rows : iterable of `tuple`
+            Result set, this can be a single-pass iterator. Each tuple
+            contains unit values corresponding to units in a header.
+        """
+
+        # for now only a single collection is supported
+        if len(collections) != 1:
+            raise ValueError("Only single collection is supported by makeDataGraph()")
+        collection = collections[0]
+
+        # Collect unit names in both input and output dataset types
+        allUnits = set(itertools.chain.from_iterable(dsType.dataUnits for dsType in neededDatasetTypes))
+        allUnits.update(itertools.chain.from_iterable(dsType.dataUnits for dsType in futureDatasetTypes))
+        _LOG.debug("allUnits: %s", allUnits)
+
+        # All DataUnit instances in a subset that we need
+        allDataUnits = {unitName: self._schema.dataUnits[unitName] for unitName in allUnits}
+
+        # potentially we have to extend units set with the "required" superset
+        # from schema, for now just assume that set is full.
+
+        # Temporary workaround for AbstractFilter "view" which is not defined in schema yet,
+        # make a subquery for it and use it as an alias for that sub-query
+        dataUnit = allDataUnits.get("AbstractFilter")
+        if dataUnit and dataUnit.table is None:
+            # make a copy, do not change object in schema
+            dataUnit = copy.copy(dataUnit)
+            filterTable = self._schema.dataUnits["PhysicalFilter"].table
+            subquery = select([filterTable.c["abstract_filter"]]).distinct()
+            dataUnit._table = subquery.alias("AbstractFilter")
+            allDataUnits["AbstractFilter"] = dataUnit
+
+        # Build select column list
+        selectColumns = []
+        header = []            # tuple (UnitName, link_name) for each returned column
+        for dataUnit in allDataUnits.values():
+            if dataUnit.table is not None:
+                # take link column names, usually there is one
+                for link in dataUnit.link:
+                    header.append((dataUnit.name, link))
+                    selectColumns.append(dataUnit.table.c[link])
+        _LOG.debug("selectColumns: %s", selectColumns)
+        _LOG.debug("header: %s", header)
+
+        # joins for all unit tables
+        where = []
+        for dataUnit in allDataUnits.values():
+            table = dataUnit.table
+            if table is None:
+                continue
+            _LOG.debug("add unit table: %s", table)
+
+            for other in dataUnit.requiredDependencies:
+                _LOG.debug("joining to: %s", other)
+                for link in other.link:
+                    _LOG.debug("joining on link: %s", link)
+                    where.append(table.c[link] == other.table.c[link])
+
+            for other in dataUnit.optionalDependencies:
+                if other.name in allDataUnits:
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -0,0 +1,185 @@
...
+
+    This is an implementation detail only to be used by SqlRegistry class,
+    not supposed to be used anywhere else.
+
+    Parameters
+    ----------
+    schema : `Schema`
+        Schema instance
+    engine : `sqlalchmey.engine`
+        Engine to use for database access.
```
TallJimbo: Looks (from the variable names) that this is now a connection, not an engine.

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -0,0 +1,185 @@
...
+            The `list` of `DatasetType`\ s whose instances may be added to the
+            graph later, which requires that their `DataUnit` types must be
+            present in the graph.
+
+        Returns
+        -------
+        header : `tuple` of `tuple`
+            Length of tuple equals the number of columns in the returned
+            result set. Each item is a tuple with two elements - DataUnit
+            name (e.g. "Visit") and unit value name (e.g. "visit").
```
TallJimbo, line 85: As a heads up, there is one DataUnit with two value columns (`ExposureRange`).

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -947,9 +949,13 @@ def merge(self, outputCollection, inputCollections):
...
 
-    def makeDataGraph(self, collections, expr, neededDatasetTypes, futureDatasetTypes):
+    def selectDataUnits(self, collections, expr, neededDatasetTypes, futureDatasetTypes):
         r"""Evaluate a filter expression and lists of `DatasetType`\ s and
-        return a `QuantumGraph`.
+        return a set of data unit values.
+
+        Returned set consists of combinations of units participating in data
+        transformation from ``neededDatasetTypes`` to ``futureDatasetTypes``,
+        restricted by existing data and filter expression.
```
TallJimbo, line 1039: Why is this an `r"..."` string?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -969,11 +975,17 @@ def makeDataGraph(self, collections, expr, neededDatasetTypes, futureDatasetType
...
-        graph : `QuantumGraph`
-            A `QuantumGraph` instance with a `QuantumGraph.units` attribute
-            that is not `None`.
+        header : `tuple` of `tuple`
+            Length of tuple equals the number of columns in the returned
+            result set. Each item is a tuple with two elements - DataUnit
+            name (e.g. "Visit") and unit value name (e.g. "visit").
+        rows : sequence of `tuple`
+            Result set, this can be a single-pass iterator. Each tuple
+            contains unit values corresponding to units in a header.
```
TallJimbo, line 1065: If it can be a single-pass iterator, docs should probably say `iterable` rather than `sequence`.

Also, just a heads-up: after [DM-15034](https://jira.lsstcorp.org/browse/DM-15034) it might make sense to make `rows` an iterable of Data ID objects instead.

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -0,0 +1,185 @@
...
+        restricted by existing data and filter expression.
+
+        Parameters
+        ----------
+        collections : `list` of `str`
+            An ordered `list` of collections indicating the Collections to
+            search for Datasets.
+        expr : `str`
+            An expression that limits the `DataUnit`\ s and (indirectly) the
+            Datasets returned.
```
TallJimbo: Should point to some documentation for the expressions syntax (when it exists; I understand if it does not yet).

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -0,0 +1,185 @@
...
+        neededDatasetTypes : `list` of `DatasetType`
+            The `list` of `DatasetType`\ s whose instances should be included
+            in the graph and limit its extent.
```
TallJimbo: Since this method no longer returns anything like a graph, this docstring (and the one below) should be changed.

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -0,0 +1,185 @@
...
+        futureDatasetTypes : `list` of `DatasetType`
+            The `list` of `DatasetType`\ s whose instances may be added to the
+            graph later, which requires that their `DataUnit` types must be
+            present in the graph.
```
TallJimbo: We might want to provide another argument that allows the caller to provide a list of additional DataUnits to be included.  We could even use that to pass the quantum units in for completeness, but I can't think of a case where a SuperTask's quantum units would not be a subset of its DatasetType units.

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -0,0 +1,185 @@
...
+        # Collect unit names in both input and output dataset types
+        allUnits = set(itertools.chain.from_iterable(dsType.dataUnits for dsType in neededDatasetTypes))
+        allUnits.update(itertools.chain.from_iterable(dsType.dataUnits for dsType in futureDatasetTypes))
+        _LOG.debug("allUnits: %s", allUnits)
+
+        # All DataUnit instances in a subset that we need
+        allDataUnits = {unitName: self._schema.dataUnits[unitName] for unitName in allUnits}
+
+        # potentially we have to extend units set with the "required" superset
+        # from schema, for now just assume that set is full.
```
TallJimbo, line 105: If all of the units are coming from `DatasetType` objects, the guarantee should happen there.  I don't think this is done yet but it probably will be on [DM-15034](https://jira.lsstcorp.org/browse/DM-15034).

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -0,0 +1,185 @@
...
+        # Temporary workaround for AbstractFilter "view" which is not defined in schema yet,
+        # make a subquery for it and use it as an alias for that sub-query
+        dataUnit = allDataUnits.get("AbstractFilter")
+        if dataUnit and dataUnit.table is None:
+            # make a copy, do not change object in schema
+            dataUnit = copy.copy(dataUnit)
+            filterTable = self._schema.dataUnits["PhysicalFilter"].table
+            subquery = select([filterTable.c["abstract_filter"]]).distinct()
+            dataUnit._table = subquery.alias("AbstractFilter")
+            allDataUnits["AbstractFilter"] = dataUnit
```
TallJimbo, line 116: This workaround can be removed after you rebase on master; an `AbstractFilter` view was added on [DM-14428](https://jira.lsstcorp.org/browse/DM-14334).

tests/test_sqlPreFlight.py
```diff
@@ -0,0 +1,240 @@
...
+
+import lsst.utils.tests
+
+from lsst.daf.butler.core.butlerConfig import ButlerConfig
+from lsst.daf.butler.core.datasets import DatasetType
+from lsst.daf.butler.core.registry import Registry
+from lsst.daf.butler.core.storageClass import StorageClass
+
+"""Tests for SqlRegistry.
+"""
```
TallJimbo: This docstring seems superfluous.  I'm also not sure it works as a docstring anyway if it's below all of the imports.

tests/test_sqlPreFlight.py
```diff
@@ -0,0 +1,240 @@
...
+                                                       expr="Visit.visit > 1000",
+                                                       neededDatasetTypes=[rawType],
+                                                       futureDatasetTypes=[calexpType])
+        self.assertEqual(len(headers), 4)  # headers must be the same
+        rows = list(rows)
+        self.assertEqual(len(rows), 0)
+
+        # Selecting by PhysicalFilter, this should be done via
+        # PhysicalFilter.physical_filter but that is not supported currently,
+        # instead we can do Visit.physical_filter (or Exposure.physical_filter)
```
TallJimbo, line 130: I don't see where this limitation comes from in the implementation.

tests/test_sqlPreFlight.py
```diff
@@ -0,0 +1,240 @@
...
+        self.assertEqual(len(headers), 4)  # headers must be the same
+        rows = list(rows)
+        self.assertEqual(len(rows), 2*3)   # 2 exposures times 3 sensors
+
+    def testPreFlightSkyMapUnits(self):
+        """Test involving only SkyMap units, no joins to Camera"""
+        registry = self.registry
+
+        # need a bunch of units and datasets for test, we want 'AbstractFilter'
+        # in the test so also have to add PhysicslPhilter units
```
TallJimbo: I think this workaround can also now be removed.

TallJimbo (approved):
Looks good.  I've left some minor cleanup comments, and some thoughts about the future.

My only big-picture comment is that I don't see a good reason to make `SqlPreFlight` a distinct class from `SqlRegistry`; its attributes are things the `SqlRegistry` already has.  I could imagine it being nice to keep it in a separate source file, at least temporarily, to avoid merge/rebase conflicts while everything is under active development, but I think a free function would work as well and would be easier to integrate into `SqlRegistry` itself when development slows down.

python/lsst/daf/butler/registries/sqlRegistry.py
timj: It has a `\` in it and this triggers a warning without either escaping it or using `r"""`.

TallJimbo: Better to write that as `DatasetTypes <DatasetType>`, I think?

python/lsst/daf/butler/registries/sqlPreFlight.py
andy-slac: These will appear as two separate tuples in the `header` but they both will have "ExposureRange" as first item.

python/lsst/daf/butler/registries/sqlRegistry.py
andy-slac: I copied this from elsewhere but I agree `DatasetTypes <DatasetType>` should be better and we can drop `r` prefix.

python/lsst/daf/butler/registries/sqlPreFlight.py
andy-slac: I think `str` here will have to be replaced by some new expression class ~soon.

python/lsst/daf/butler/registries/sqlPreFlight.py
andy-slac: I'm not sure I understand the use case for this but we can certainly extend it later if necessary.

python/lsst/daf/butler/registries/sqlPreFlight.py
andy-slac: I was not 100% sure that is always going to be true. Let's keep the comment, we can clean it up later.

tests/test_sqlPreFlight.py
andy-slac: `PhysicalFilter` is not in the units for the `RAW` DatasetType that I defined above, so PhysicalFilter table does not appear in generated SQL.  I will need to analyze user expression to see that PhysicalFilter DataUnit is there and then add a join for that table. For that I'll need user expression to be something better than dumb string, something I plan to do later.

TallJimbo: Ah, this gets at the point above about dependent DataUnits (though not required ones).  "raw" does depend on a DataUnit that has a foreign key to PhysicalFilter, and that's another way it could be brought in rather than via the expression.

tests/test_sqlPreFlight.py
andy-slac: Hmm, I think AbstractFilter is a view on PhysicalFilter in the new implementation too, so I need to have some PhysicalFilter units in the registry, no?

andy-slac:
Thanks for review @TallJimbo, I have indeed started preflight implementation as a method in SqlRegistry, but that class is already quite big and preflight will grow (much) larger than it is now so I decided to move to a separate class, which also helps with parallel development. I'm not even sure that I want to merge it back but we will see later what is the best.
I think I responded to all comments, going to rebase, re-test and merge my branch today.

tests/test_sqlPreFlight.py
TallJimbo: That is correct: AbstractFilter is just a SELECT DISTINCT on PhysicalFilter.

PR #54, opened 2018-06-18
DM-14811: Update Butler makeView to use DDLEvents.

pschella:


TallJimbo (approved):
This may be out of scope for this issue, but since we know which tables any of our views will depend on, could we make the creation of just those tables trigger view creation?

Possibly related: what happens if we trigger view creation *before* the creation of the tables on which it depends?

pschella:
This should be possible in principle, however I'm not sure (yet) how to trigger after a fixed set of tables has been created except after each one (as is done currently). I'll have a quick look to see if that is obvious and if so will implement it, if not I'll merge as is and file a follow-up ticket.

PR #55, opened 2018-06-19
DM-14428: Spatial joins tables and views

TallJimbo:


python/lsst/daf/butler/core/registry.py
```diff
@@ -69,7 +69,8 @@ def setConfigRoot(cls, root, config, full):
             from defaults when Butler instances are constructed
             should be copied from `full` to `Config`.
         """
-        raise NotImplementedError()
+        for key in ("registry.skypix.name", "registry.skypix.level"):
```
andy-slac: Should it be `registry.skypix.cls` instead of `registry.skypix.name`?

config/schema.yaml
```diff
@@ -694,6 +655,66 @@ schema:
...
+            type: int
+            primary_key: true
+            nullable: false
+            doc: Visit ID
+          -
+            name: sensor
+            type: int
+            primary_key: true
+            nullable: false
+            doc: visit
```
andy-slac: Sensor ID?

config/schema.yaml
```diff
@@ -694,6 +655,66 @@ schema:
...
+          -
+            name: region
+            type: string
```
andy-slac: type is `string` in this table but `blob` in Tract/Patch tables. Should we instead introduce special `region` column type and use type magic name rather than column magic name in Python code?

config/schema.yaml
```diff
@@ -694,6 +655,66 @@ schema:
...
+            doc: >
+              A spatial region on the sky that bounds the area associated with
+              this Visit+Sensor combination.  This is expected to be more
+              precise than the SkyPixels associated with the Visit+Sensor (see
+              VisitSensorSkyPixJoin), but may still be larger than the true
+              region as long as it fully covers it.
```
andy-slac, line 707: Do we describe string representation for regions anywhere or do we depend on sphgeom for it? If latter them maybe add couple of words here.

config/schema.yaml
```diff
@@ -694,6 +655,66 @@ schema:
...
+          foreignKeys:
+          -
+            src: camera
+            tgt: Camera.camera
```
andy-slac: I'm not sure this foreign key is useful at all.

config/schema.yaml
```diff
@@ -694,6 +655,66 @@ schema:
...
+          -
+            src:
+            - camera
+            - visit
```
andy-slac: I am confused by YAML indentation, I see in this document two styles:

    src:
    - camera
    - visit

and

    src:
      - camera
      - visit

Do they mean the same thing?

config/schema.yaml
```diff
@@ -813,9 +835,9 @@ schema:
         VisitSensorSkyPixJoin.
 
       lhs:
-      - Visit
+        - Visit
       rhs:
-      - SkyPix
+        - SkyPix
       summarizes:
       - VisitSensorSkyPixJoin
```
andy-slac: Indent this too?

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -331,6 +373,25 @@ def _initDataUnits(self, config, builder):
                                 link=link)
             self[dataUnitName] = dataUnit
 
+    def _initDataUnitRegions(self, config, builder):
+        """Initialize tables that associate regions with multiple DataUnits.
+
+        Parameters
+        ----------
+        config : `Config`
+            The `DataUnit_regions` component of a `SchemaConfig`.
```
andy-slac: `dataUnitRegions`?

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -227,13 +236,16 @@ def fromConfig(cls, config, builder=None):
         Parameters
         ----------
         config : `SchemaConfig`
-            `Registry` schema configuration describing `DataUnit` relations.
+            `Registry` schema configuration containing 'DataUnits',
+            'DataUnit_regions', and 'DataUnit_joins' entries.
```
andy-slac: `dataUnitRegions` and `dataUnitJoins`?

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -331,6 +373,25 @@ def _initDataUnits(self, config, builder):
...
+        Parameters
+        ----------
+        config : `Config`
+            The `DataUnit_regions` component of a `SchemaConfig`.
+        builder : `SchemaBuilder`, optional
+            When given, create `sqlalchemy.core.Table` entries.
+        """
+        for description in config:
+            dataUnitNames = frozenset(description["relates"])
+            (tableName, tableDescription), = description["tables"].items()
```
andy-slac: :grimacing:
`[(tableName, tableDescription)] = description["tables"].items()` may be tiny bit better but still causing me eye strain :eyes:

config/schema.yaml
```diff
@@ -694,6 +655,66 @@ schema:
         type: string
         doc: A string value composed only of letters, numbers, and underscores.
 
+  dataUnitRegions:
+    -
```
andy-slac, line 668: Is there a reason why this is a list rather than associative array like `dataUnits` and `dataUnitJoins`?

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -260,6 +272,36 @@ def items(self):
         for dataUnitName in self._dataUnitNames:
             yield (dataUnitName, self[dataUnitName])
 
+    def getRegionTable(self, *dataUnitNames):
+        """Return the region table that holds regions for the given combination
+        of DataUnits.
+        """
+        if len(dataUnitNames) == 1:
+            return self[dataUnitNames[0]].table
```
andy-slac, line 280: `self._dataUnits[dataUnitNames[0]].table`?

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -270,7 +312,7 @@ def _initDataUnitNames(self, config):
         Parameters
         ----------
         config : `SchemaConfig`
-            Schema configuration describing `DataUnit` relations.
+            The `DataUnit` component of a `SchemaConfig`.
```
andy-slac: DataUnits (or dataUnits)?

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -286,8 +328,8 @@ def _initDataUnits(self, config, builder):
 
         Parameters
         ----------
-        config : `SchemaConfig`
-            Schema configuration describing `DataUnit` relations.
+        config : `Config`
+            The `DataUnit` component of a `SchemaConfig`.
```
andy-slac: Here too.

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -343,16 +404,19 @@ def _initDataUnitJoins(self, config, builder):
         """
         for dataUnitJoinName, dataUnitJoinDescription in config.items():
             table = None
+            isView = None
             if 'tables' in dataUnitJoinDescription and builder is not None:
                 for tableName, tableDescription in dataUnitJoinDescription['tables'].items():
                     table = builder.addTable(tableName, tableDescription)
+                    isView = "sql" in tableDescription
```
andy-slac, line 411: Could "sql" in YAML potentially mean something else than view?

python/lsst/daf/butler/core/registry.py
```diff
@@ -115,5 +116,7 @@ def fromConfig(registryConfig, schemaConfig=None, create=False):
     def __init__(self, registryConfig, schemaConfig=None, create=False):
         assert isinstance(registryConfig, RegistryConfig)
         self.config = registryConfig
+        pixelizationCls = doImport(self.config["skypix.cls"])
```
andy-slac: Make `self.pixelization` a property so that import is deferred until it's actually needed?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -788,6 +788,12 @@ def addDataUnitEntry(self, dataUnitName, values):
         values : `dict`
             Dictionary of ``columnName, columnValue`` pairs.
 
+        If ``values`` includes a "region" key, `setDataUnitRegion` will
+        automatically be called to set it any any associated spatial join
```
andy-slac: any any

andy-slac (commented):
@TallJimbo, I'm not completely done reviewing yet, but have some comments you may want to start looking at while I'm finishing my part.

config/schema.yaml
```diff
@@ -769,7 +790,8 @@ schema:
       lhs:
         - Visit
         - Sensor
-      rhs: SkyPix
+      rhs:
+        - SkyPix
```
andy-slac: For simple cases like this could we possibly use more compact inline style?

    lhs: [Visit, Sensor]
    rhs: [SkyPix]

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
```diff
@@ -19,52 +19,106 @@
...
         Parameters
         ----------
-        path : `str`
-            Path to use to open the file.
-        pytype : `class`
-            Class to use to read the FITS file.
+        fileDescriptor : `FileDescriptor`
+            Identifies the file to read, type to read it into and parameters
+            to be used for reading.
+        comp : `str`, optional
```
andy-slac: Can you spell this parameter name completely? `comp` is such a common prefix, hard to tell if it means compare/computer/compost/compassion :wink:
Also, base class `read()` method does not seem to have this parameter, I can't figure out how it can be used.

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
```diff
@@ -19,52 +19,106 @@
...
+            # If we're reading a non-image component, just read in a single-pixel image for efficiency.
+            kwds = {}
+            if comp in ("image", "variance", "mask"):
+                kwds["bbox"] = Box2I(minimum=Point2I(0, 0), maximum=Point2I(0, 0))
+                kwds["origin"] = LOCAL
+            elif fileDescriptor.parameters is not None:
+                # Just pass parameters into kwargs for constructor, but check that we recognize them.
+                kwds.update(fileDescriptor.parameters)
+                if not self.parameters.issuperset(kwds.keys()):
+                    raise KeyError("Unrecognized parameter key(s): {}".format(self.parameters - kwds.keys()))
```
andy-slac: Should it be `kwds.keys() - self.parameters` instead?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -800,12 +806,81 @@ def addDataUnitEntry(self, dataUnitName, values):
...
+            keyColumns.update(dataUnit.primaryKeyColumns)
+        table = self._schema.dataUnits.getRegionTable(*dataUnitNames)
+        if table is None:
+            raise TypeError("No region table found for '{}'.".format(dataUnitNames))
+        # If a region record for these DataUnits already exists, use an update
+        # query. That could happen either because those DataUnits have been
+        # inserted previously and this is an improved region for them, or
+        # because the region is associated with a single DataUnit instance and
+        # is hence part of that DataUnit's main table.
+        if not new or (len(dataUnitNames) == 1 and table == dataUnit.table):
```
andy-slac, line 853: I guess `dataUnit` here originates from the loop above? Not illegal but potentially fragile.

andy-slac (changes requested):
I'm done with review, it looks OK but I think there may be couple of thing that need change, please see my comments.

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
timj: This looks like it's my fault because I call it `comp` in `fileFormatter`. posixDatastore does use that form of `formatter.read()`. You are right I forgot to change the abstract base class. The change here is to allow Exposure components to be read without requiring the entire Exposure to be read.

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
```diff
@@ -19,52 +19,106 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import os.path
-
-from lsst.daf.butler.formatters.fileFormatter import FileFormatter
+from ..core.formatter import Formatter
```
timj: I was deliberately using the public formatter interface rather than doing a relative import, since this is what external formatters would look like and it seemed wrong to use internals for them.

ups/daf_butler.table
```diff
@@ -9,6 +9,7 @@ setupRequired(utils)
 setupRequired(daf_persistence)
 setupRequired(afw)
 
+envPrepend(PATH, ${PRODUCT_DIR}/bin)
 envPrepend(LD_LIBRARY_PATH, ${PRODUCT_DIR}/lib)
```
timj: whereby I notice that we are setting library search paths when we should not be since this is pure python package.

python/lsst/daf/butler/registries/sqliteRegistry.py
```diff
@@ -65,6 +65,7 @@ def setConfigRoot(cls, root, config, full):
             from defaults when Butler instances are constructed
             should be copied from `full` to `Config`.
         """
+        super(SqliteRegistry, cls).setConfigRoot(root, config, full)
```
timj: If `super().setConfigRoot()` is not going to work it might be worth having a comment here explaining why you are using the explicit form.

config/schema.yaml
TallJimbo: I felt like the conceptual key for one of these items is the "relates" entry, which is itself a list and hence can't be used as a key directly.  So a list seemed like the next-best thing.

config/schema.yaml
TallJimbo: I've made it a blob here too.  Is there a way to use magic Python conversions in a way that is truly RDBMS-independent?  (Note I'd like to avoid even exposing that we're using SQLAlchemy publicly).

config/schema.yaml
TallJimbo: We do rely on sphgeom for this.  I'll add some docs.

config/schema.yaml
TallJimbo: It is indeed redundant with other foreign key constraints.  I'll remove it.

config/schema.yaml
TallJimbo: Yes, they do.  I suspect it's just two different editors.  It would be nice to standardize, but I don't think we should do that on this ticket.

config/schema.yaml
TallJimbo: Sure (I don't have a preference - it's a question of whether compactness or consistency is more valuable).

config/schema.yaml
TallJimbo: (just switched to compact style)

python/lsst/daf/butler/core/dataUnit.py
TallJimbo: Seems better to have private methods use the public API unless there's a good reason not to.

python/lsst/daf/butler/core/dataUnit.py
TallJimbo: I had no idea that syntax worked.  Will do.

I do wish there was a better way to write this as well, but I can't think of anything that isn't super verbose.

python/lsst/daf/butler/core/dataUnit.py
TallJimbo: In general, yes (see `ExposureRangeJoin`), inside a `tables` section it always means a view.

config/schema.yaml
andy-slac: I don't think we need magic here, we are reading this schema ourselves and transform it into backed-specific schema so we have complete freedom. My worry is that using low-level (physical) type here instead of logical (`region`) we limit ourselves to a particular choice of storage type. I can imagine for some backends BLOB may be too inefficient and you may want to store regions (which should be relatively short strings of bytes) in more efficient storage type.

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
TallJimbo: I've renamed this to `component` and added it to the `Formatter` base class.

python/lsst/daf/butler/registries/sqlRegistry.py
TallJimbo: Right; it looks a bit weird to my C++-trained eyes, too, but I think this is an example of exactly why Python allows loop variables to live beyond their loops.

python/lsst/daf/butler/registries/sqliteRegistry.py
TallJimbo: Ah, I thought `super()` wouldn't work in a `classmethod`, but it seemed I was wrong about that.

config/schema.yaml
TallJimbo: Ah, I just misunderstood what you were suggesting.  I'll go ahead and make that change (while still mapping `region` columns to `sqlalchemy.LargeBinary` for now).

TallJimbo:
@andy-slac, I believe I've addressed all comments.  Please take another look and approve if you agree.

python/lsst/daf/butler/core/dataUnit.py
andy-slac: I missed `__getitem__`, sorry for the noise.

andy-slac (approved):

andy-slac:
Approved, github says "andy-slac approved these changes 17 seconds from now" but I approved it 20 seconds ago :smile:

PR #58, opened 2018-07-24
DM-13363: In memory datastore

timj:
* Add an in memory datastore.
* Reconfigure datastore tests to run with multiple datastores.


python/lsst/daf/butler/core/utils.py
```diff
@@ -323,3 +324,46 @@ def inner(self, *args, **kwargs):
         with self.transaction():
             return func(self, *args, **kwargs)
     return inner
+
+
+def get_object_size(obj, seen=None):
```
TallJimbo: -> camelCase :-/

python/lsst/daf/butler/core/utils.py
```diff
@@ -323,3 +324,46 @@ def inner(self, *args, **kwargs):
...
+    # Important mark as seen *before* entering recursion to gracefully handle
+    # self-referential objects
+    seen.add(obj_id)
+    if isinstance(obj, dict):
+        size += sum([get_object_size(v, seen) for v in obj.values()])
+        size += sum([get_object_size(k, seen) for k in obj.keys()])
+    elif hasattr(obj, '__dict__'):
+        size += get_object_size(obj.__dict__, seen)
+    elif hasattr(obj, '__iter__') and not isinstance(obj, (str, bytes, bytearray)):
+        size += sum([get_object_size(i, seen) for i in obj])
```
TallJimbo: If this works at all for `Image` objects or other C++-based things (which is where most of our memory is going to go), it's gonna be awfully slow.  I'm inclined to say we should just have `InMemoryDatastore` leave the size field blank until we have a use case for it doing something else, and only then think about what kinds of actual objects we need to support to estimate overall memory use.  I suspect we'll need something much more intrusive to make it useful.

python/lsst/daf/butler/datastores/inMemoryDatastore.py
```diff
@@ -0,0 +1,386 @@
...
+            Reference to the associated Dataset.
+        """
+
+        datasetType = ref.datasetType
+        storageClass = datasetType.storageClass
+
+        # Sanity check
+        if not isinstance(inMemoryDataset, storageClass.pytype):
+            raise ValueError("Inconsistency between supplied object ({}) "
+                             "and storage class type ({})".format(type(inMemoryDataset), storageClass.pytype))
```
TallJimbo: Seems like a `TypeError` to me, especially because that's what its inverse in `get` is.

python/lsst/daf/butler/datastores/inMemoryDatastore.py
```diff
@@ -0,0 +1,386 @@
...
+
+        Raises
+        ------
+        FileNotFoundError
+            Attempt to remove a dataset that does not exist.
+
+        Notes
+        -----
+        Some Datastores may implement this method as a silent no-op to
+        disable Dataset deletion through standard interfaces.
```
TallJimbo: Probably don't want this `Note` in a concrete (and presumably final) implementation.

python/lsst/daf/butler/core/config.py
```diff
@@ -403,8 +454,16 @@ def __init__(self, other=None, validate=True, mergeDefaults=True, searchPaths=No
...
+        # To simplify the use of !include we also check for the existence of
+        # component.component (since the included files can themselves
+        # include the component name)
+        if self.component is not None:
+            doubled = "{0}.{0}".format(self.component)
+            # Must check for double depth first
+            if doubled in externalConfig:
+                externalConfig = externalConfig[doubled]
+            elif self.component in externalConfig:
+                externalConfig.data = externalConfig.data[self.component]
```
TallJimbo, line 479: What happens if we have a parent YAML file with something like:
```
datastore:
    cls: lsst.daf.butler.ParentDatastore
    child1: !include posixDatastore.yaml
    child2: !include someOtherDatastore.yaml
```
(i.e. instead of using "datastore" as the key for the `!include` value).  I think we'd want that to also strip out the `datastore` from the included file.

I can think of two solutions to this problem (there may be others):
 - We stop putting the top-level key in each file, and adjust all consumers of those files (i.e. so datastore config files would *never* start with `datastore:`; they'd start with the level below that).
 - We give `!include` the ability take an argument that specifies a key whose value to extract from the file, so the previous example would be written as something like:
```
datastore:
    cls: lsst.daf.butler.ParentDatastore
    child1: !include posixDatastore.yaml[datastore]
    child2: !include someOtherDatastore.yaml[datastore]
```
Given that more verbose syntax, I think I prefer the first option, unless there's some other problem the second one would solve.  And maybe there's a third option that's better than either, but I don't think we can assume we want to remove the child key if any only if it's a duplicate of the parent key.

tests/config/basic/butler.yaml
```diff
@@ -1,65 +1,3 @@
...
-      output: StructuredDataDictJson
-      data: StructuredDataListJson
-  StructuredCompositeTestB:
-    <<: *StructuredComposite
-    components:
-      summary: StructuredDataDictPickle
-      output: StructuredDataDictJson
-      data: StructuredDataListYaml
+datastore: !include posixDatastore.yaml
+storageClasses: !include storageClasses.yaml
```
TallJimbo, line 3: I assume this means that `!include` is relative to the directory of the file that uses it?

python/lsst/daf/butler/core/config.py
```diff
@@ -53,6 +54,56 @@ class _ConfigBase(collections.UserDict, yaml.YAMLObject, metaclass=_ConfigMeta):
...
+           data = yaml.load(f, Loader=Loader)
+
+    Notes
+    -----
+    See https://davidchall.github.io/yaml-includes.html
+    """
+
+    def __init__(self, stream):
+        self._root = os.path.split(stream.name)[0]
+        super().__init__(stream)
```
TallJimbo: Seeing the delegation to `super` in the second line of this method made me immediately wonder why it wasn't the first line.  If there's no reason (I don't see one), it'd read slightly more easily to make it the first one.

python/lsst/daf/butler/datastores/inMemoryDatastore.py
```diff
@@ -219,7 +237,11 @@ def get(self, ref, parameters=None):
         storedItemInfo = self.getStoredItemInfo(ref)
         writeStorageClass = storedItemInfo.storageClass
 
-        inMemoryDataset = self.datasets[ref.id]
+        # We might need a parent
```
TallJimbo: Please expand this comment a bit: I should remember under what conditions we do this kind of thing, but it took me a moment to remind myself by reading the code, and someone without that background wouldn't have a chance.

python/lsst/daf/butler/datastores/inMemoryDatastore.py
```diff
@@ -0,0 +1,397 @@
...
+        self.storageClassFactory = StorageClassFactory()
+
+        # Name ourselves with the timestamp the datastore
+        # was created.
+        self.name = "InMemoryDatastore@{}".format(time.time())
+        logging.debug("Creating datastore %s", self.name)
+
+        # Storage of datasets, keyed by dataset_id
+        self.datasets = {}
+        self.records = {}
```
TallJimbo: It took me a while to figure out that the fundamental reason you need `self.records` in addition to `self.datasets` is because `records` includes entries for components of concrete composites while `datasets` does not (before I figured that out, I was going to argue that you could remove `records`, because there was no public API to get at the extra information it contained); please add a comment to that effect.

TallJimbo (approved):
Looks good.  A few minor comments on the PR, and I'd like to discuss what to do about `!include` producing `datastore: datastore` entries and the like.

python/lsst/daf/butler/core/utils.py
timj: I did check that it worked for numpy arrays, but I haven't checked what happens with a full size `Exposure`.  I'll leave the code in but remove the usage from the datastore. My use case was pondering whether we might want to cleanup the cache based on size as well as age.

python/lsst/daf/butler/datastores/inMemoryDatastore.py
timj: It's a `ValueError` because that's what we used in `PosixDataStore` and this is a copy and paste. I'll fix up both places.

python/lsst/daf/butler/core/config.py
timj: I think in your example the right thing would happen because I strip `datastore.datastore` but the child will be `datastore.child1.datastore` and the `DatastoreConfig` constructor would be given `datastore.child1` as an argument and would correctly handle it. I was assuming chained datastores would recursively call datastore constructors with `DatastoreConfig` rather than having the parent datastore try to parse everything itself. I might be misunderstanding your point though. I was worrying this morning that `datastore.datastore` might be deliberate 

The code should already work without including the `datastore:` level in datastore config files. I'm okay with requiring that we don't have that in files that are `!include`ed but I was thinking it's clearer for people if they can see from the contents of the file that it's a datastore config rather than relying on the file name or parent directory name.

tests/config/basic/butler.yaml
timj: Yes. I think the YAML parser chdirs to the directory containing the file. I think relative to the current YAML file is the right answer.

python/lsst/daf/butler/core/config.py
timj: I'll have a look. This is "Code From the Internet" :-)

python/lsst/daf/butler/datastores/inMemoryDatastore.py
timj: Ok. It took me a while to work that out myself (wondering why all the tests of concrete composites were failing).

python/lsst/daf/butler/core/config.py
TallJimbo: I was just thinking that the `child1` in `datastore.child1.datastore` would be an unnecessary single-item dict that it'd be nice to strip out, especially because we *do* strip out a conceptually similar single-item dict in `datastore.datastore` -> `datastore` (i.e. I don't think a composite Datastore's config should have a different structure based on whether it calls its child "datastore" or something else).  But I suppose it's a judgement call whether that's more or less important than having datastore configs start with their headings; that's nice too.

timj: How about we punt this discussion until after I see how chained datastore configs really work?

TallJimbo: Works for me.

python/lsst/daf/butler/core/utils.py
timj: It doesn't work at all for `Exposure` because `Exposure` does not have any way of iterating through its attributes -- all those getter methods are essentially opaque to Python and I can't try all the getter methods because I'm not sure which ones are representing actual content and which are derived properties (like the dimensions) and which ones need arguments (and so aren't pure getters). I've tested some code that runs all the getters and includes them in the sum but it's obviously not getting everything -- we could really use native support in Exposure for reporting size (why isn't there a `__size__` method?).

PR #59, opened 2018-07-24
DM-15189: core butler support for raw data ingest

TallJimbo:
@timj, this is by no means done yet, but it occurred to me that we might be doing conflicting things with `Datastores`, so I wanted to get this out there in case you'd like me to spin it off to a new issue and merge sooner.  I also need to write some tests for it, but since you've just refactored `Datastore` tests I definitely didn't want to do that before rebasing on that work.

### DIFF
--- a/config/datastores/posixDatastore.yaml
+++ b/config/datastores/posixDatastore.yaml
@@ -5,7 +5,7 @@ datastore:
     table: PosixDatastoreRecords
   create: true
   templates:
-    default: "{datasetType}/{tract:?}/{patch:?}/{filter:?}/{camera:?}_{visit:?}"
+    default: "{collection}/{datasetType}/{tract:?}/{patch:?}/{filter:?}/{camera:?}_{visit:?}"
   formatters:
     TablePersistable: lsst.daf.butler.formatters.fitsCatalogFormatter.FitsCatalogFormatter
     TablePersistableWcs: lsst.daf.butler.formatters.fitsCatalogFormatter.FitsCatalogFormatter
--- a/config/storageClasses.yaml
+++ b/config/storageClasses.yaml
@@ -66,6 +66,7 @@ storageClasses:
       visitInfo: TablePersistableVisitInfo
       apCorrMap: TablePersistableApCorr
       coaddInputs: TablePersistableCoaddInputs
+      transmissionCurve: TablePersistableTransmissionCurve
       metadata: PropertyList
   ExposureF: &ExposureF
     <<: *Exposure
--- a/config/storageClasses.yaml
+++ b/config/storageClasses.yaml
@@ -19,12 +19,24 @@ storageClasses:
     pytype: lsst.afw.image.Calib
   TablePersistableTransmissionCurve:
     pytype: lsst.afw.image.TransmissionCurve
+  Image: &Image
+    pytype: lsst.afw.image.Image
   ImageF:
+    <<: *Image
     pytype: lsst.afw.image.ImageF
+  ImageI:
+    <<: *Image
+    pytype: lsst.afw.image.ImageI
   ImageU:
+    <<: *Image
     pytype: lsst.afw.image.ImageU
+  DecoratedImage: &DecoratedImage
+    pytype: lsst.afw.image.DecoratedImage
   DecoratedImageU:
+    <<: *DecoratedImage
     pytype: lsst.afw.image.DecoratedImageU
+  Mask:
+    pytype: lsst.afw.image.Mask
   MaskX:
     pytype: lsst.afw.image.MaskX
   Catalog:
@@ -35,6 +47,8 @@ storageClasses:
     pytype: lsst.afw.table.SourceCatalog
   SkyMap:
     pytype: lsst.skymap.BaseSkyMap
+  PropertySet:
+    pytype: lsst.daf.base.PropertySet
   PropertyList:
     pytype: lsst.daf.base.PropertyList
   PropertySet:
@@ -43,9 +57,9 @@ storageClasses:
     pytype: lsst.afw.image.Exposure
     assembler: lsst.daf.butler.assemblers.exposureAssembler.ExposureAssembler
     components:
-      image: ImageF
-      mask: MaskX
-      variance: ImageF
+      image: Image
+      mask: Mask
+      variance: Image
       wcs: TablePersistableWcs
       psf: TablePersistablePsf
       calib: TablePersistableCalib
@@ -56,9 +70,35 @@ storageClasses:
   ExposureF: &ExposureF
     <<: *Exposure
     pytype: lsst.afw.image.ExposureF
+    components:
+      image: ImageF
+      mask: MaskX
+      variance: ImageF
+      wcs: TablePersistableWcs
+      psf: TablePersistablePsf
+      calib: TablePersistableCalib
+      visitInfo: TablePersistableVisitInfo
+      apCorrMap: TablePersistableApCorr
+      coaddInputs: TablePersistableCoaddInputs
+      transmissionCurve: TablePersistableTransmissionCurve
+      metadata: PropertyList
   ExposureI: &ExposureI
     <<: *Exposure
     pytype: lsst.afw.image.ExposureI
+    components:
+      image: ImageI
+      mask: MaskX
+      variance: ImageF
+      wcs: TablePersistableWcs
+      psf: TablePersistablePsf
+      calib: TablePersistableCalib
+      visitInfo: TablePersistableVisitInfo
+      apCorrMap: TablePersistableApCorr
+      coaddInputs: TablePersistableCoaddInputs
+      transmissionCurve: TablePersistableTransmissionCurve
+      metadata: PropertyList
+  ExposureComposite:
+    <<: *Exposure
   ExposureCompositeF:
     <<: *ExposureF
   ExposureCompositeI:
--- a/python/lsst/daf/butler/butler.py
+++ b/python/lsst/daf/butler/butler.py
@@ -324,6 +324,9 @@ class Butler:
         # Always lookup the DatasetRef, even if one is given, to ensure it is
         # present in the current collection.
         ref = self.registry.find(self.collection, datasetType, dataId)
+        if ref is None:
+            raise LookupError("Dataset {} with data ID {} could not be found in {}".format(
+                              datasetType.name, dataId, self.collection))
         if idNumber is not None and idNumber != ref.id:
             raise ValueError("DatasetRef.id does not match id in registry")
         return self.getDirect(ref)
--- a/python/lsst/daf/butler/butler.py
+++ b/python/lsst/daf/butler/butler.py
@@ -237,26 +237,23 @@ class Butler:
             if dataId is None:
                 raise ValueError("Must provide a dataId if first argument is not a DatasetRef")
             datasetType = self.registry.getDatasetType(datasetRefOrType)
-        ref = self.registry.addDataset(datasetType, dataId, run=self.run, producer=producer)
 
-        # Look up storage class to see if this is a composite
-        storageClass = datasetType.storageClass
+        isVirtualComposite = self.composites.doDisassembly(datasetType)
+
+        # Add Registry Dataset entry.  If not a virtual composite, add
+        # and attach components at the same time.
+        ref = self.registry.addDataset(datasetType, dataId, run=self.run, producer=producer,
+                                       recursive=not isVirtualComposite)
 
         # Check to see if this datasetType requires disassembly
-        if self.composites.doDisassembly(datasetType):
-            components = storageClass.assembler().disassemble(obj)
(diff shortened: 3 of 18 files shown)


TallJimbo:
Most of the code originally here was moved to DM-15212, so some earlier comments may no longer make sense.

TallJimbo:
The full ticket is finally ready for review.  In this package (as well as obs_subaru), I recommend looking at each commit separately; this package in particular just contains a lot of miscellaneous improvements necessary for making higher-level code in obs_base and obs_subaru work.

config/storageClasses.yaml
```diff
@@ -19,12 +19,21 @@ storageClasses:
     pytype: lsst.afw.image.Calib
   TablePersistableTransmissionCurve:
     pytype: lsst.afw.image.TransmissionCurve
+  Image: &Image
+    pytype: lsst.afw.image.Image
   ImageF:
+    <<: *Image
```
timj, line 25: Since `Image` only has one field and that field is being over-ridden, I'm not sure this line does anything (and the other similar changes). Are you worried that `Image` might get extra fields?

config/datastores/posixDatastore.yaml
```diff
@@ -5,7 +5,7 @@ datastore:
     table: PosixDatastoreRecords
   create: true
   templates:
-    default: "{datasetType}/{tract:?}/{patch:?}/{filter:?}/{camera:?}_{visit:?}"
+    default: "{collection}/{datasetType}/{tract:?}/{patch:?}/{filter:?}/{camera:?}_{visit:?}"
```
timj, line 8: General comment on defaulting filenames: I'm a bit worried that with this in the base definition it's not possible for people to remove defaulting in their own datastores. Maybe I should allow `default: None` to disable defaulting.

config/storageClasses.yaml
```diff
@@ -41,31 +50,50 @@ storageClasses:
...
+      calib: TablePersistableCalib
+      visitInfo: TablePersistableVisitInfo
+      apCorrMap: TablePersistableApCorr
+      coaddInputs: TablePersistableCoaddInputs
+      transmissionCurve: TablePersistableTransmissionCurve
+      metadata: PropertyList
   ExposureComposite: &ExposureComposite
     <<: *Exposure
     assembler: lsst.daf.butler.assemblers.exposureAssembler.ExposureAssembler
   ExposureCompositeF:
```
timj: This change won't be needed once #76 is merged because `ExposureCompositeF` becomes a simple label that allows you to control concrete/virtual at the butler level.

config/storageClasses.yaml
```diff
@@ -41,31 +50,50 @@ storageClasses:
...
+      mask: MaskX
+      variance: ImageF
+      wcs: TablePersistableWcs
+      psf: TablePersistablePsf
+      calib: TablePersistableCalib
+      visitInfo: TablePersistableVisitInfo
+      apCorrMap: TablePersistableApCorr
+      coaddInputs: TablePersistableCoaddInputs
+      transmissionCurve: TablePersistableTransmissionCurve
+      metadata: PropertyList
```
timj, line 85: `getMetadata` is documented to return a `PropertySet` though?

python/lsst/daf/butler/butler.py
```diff
@@ -266,6 +266,9 @@ def getDirect(self, ref):
         ----------
         ref : `DatasetRef`
             Reference to an already stored dataset.
+        parameters : `dict`
+            Additional StorageClass-defined options to control reading,
+            typically used to efficiently only a subset of the dataset.
```
timj: Missing word between "efficiently" and "only". Maybe I should work on parameters next (inMemoryDatastore will mess up with any tests where you use parameters).

python/lsst/daf/butler/butler.py
```diff
@@ -274,20 +277,25 @@ def getDirect(self, ref):
...
+            return self.datastore.get(ref, parameters=parameters)
         elif ref.components:
             # Reconstruct the composite
             components = {}
             for compName, compRef in ref.components.items():
-                components[compName] = self.datastore.get(compRef)
+                if parameters is None:
+                    compParams = None
+                else:
+                    compParams = {k: v for k, v in parameters.items()
```
timj: Parameters are nested by component name?

python/lsst/daf/butler/butler.py
```diff
@@ -300,6 +308,9 @@ def get(self, datasetRefOrType, dataId=None):
             within a Collection.
             When `None` a `DatasetRef` should be supplied as the second
             argument.
+        parameters : `dict`
+            Additional StorageClass-defined options to control reading,
+            typically used to efficiently only a subset of the dataset.
```
timj: Same missing word here.

python/lsst/daf/butler/butler.py
```diff
@@ -318,9 +329,12 @@ def get(self, datasetRefOrType, dataId=None):
         # Always lookup the DatasetRef, even if one is given, to ensure it is
         # present in the current collection.
         ref = self.registry.find(self.collection, datasetType, dataId)
+        if ref is None:
```
timj, line 337: Our tests never tickled this?

tests/test_templates.py
```diff
@@ -37,7 +37,7 @@ def makeDatasetRef(self, datasetTypeName, dataId=None):
             self.datasetTypes[datasetTypeName] = DatasetType(datasetTypeName, list(dataId.keys()),
                                                              StorageClass(None))
         datasetType = self.datasetTypes[datasetTypeName]
-        return DatasetRef(datasetType, dataId)
+        return DatasetRef(datasetType, dataId, run=Run(id=2, collection="run2"))
```
timj, line 40: I don't see a test in here for `collection` only, both `collection` and `run`, or for where `run` and `collection` are not specified in the template.

python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -162,6 +172,11 @@ def format(self, ref):
             else:
                 optional = False
 
+            if field_name in ("run", "collection"):
+                usedRunOrCollection = True
+                if optional:
+                    raise KeyError("'run' and 'collection' may not be optional.")
```
timj: Are you allowed to specify an optional `run` if `collection` is mandatory? (not that it's possible for `run` to be optional of course). Maybe the error here should be "can not be optional" ("may" sounds like you might be able to be optional). Do you have a test for optional `collection` below?

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
```diff
@@ -33,6 +33,53 @@ class FitsExposureFormatter(Formatter):
...
+
+    def readInfoComponent(self, fileDescriptor, component):
+        """Read a component held by ExposureInfo."""
+        from lsst.afw.image import LOCAL
+        from lsst.geom import Box2I, Point2I
+        parameters = dict(bbox=Box2I(minimum=Point2I(0, 0), maximum=Point2I(0, 0)), origin=LOCAL)
+        tiny = self.readFull(fileDescriptor, parameters)
+        return fileDescriptor.storageClass.assembler().getComponent(tiny, component)
+
+    def readFull(self, fileDescriptor, parameters=None):
```
timj, line 115: You are defining the parameters for `get` to be any parameters understood by the Exposure constructor?

python/lsst/daf/butler/formatters/fitsRawFormatterBase.py
```diff
@@ -0,0 +1,139 @@
...
+
+from abc import ABCMeta, abstractmethod
+
+from .fitsExposureFormatter import FitsExposureFormatter
+
+
+__all__ = ("FitsRawFormatterBase",)
+
+
+class FitsRawFormatterBase(FitsExposureFormatter, metaclass=ABCMeta):
```
timj, line 30: All of the methods here need parameters and returns sections to the numpydocs.

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
```diff
@@ -33,6 +33,53 @@ class FitsExposureFormatter(Formatter):
 
     parameters = frozenset(("bbox", "origin"))
 
+    def readImageComponent(self, fileDescriptor, component):
```
timj, line 36: Please add parameters/returns sections to docstrings.

python/lsst/daf/butler/instrument.py
```diff
@@ -65,6 +71,76 @@ def _addPhysicalFilters(self, registry):
...
+    snap : `int`
+        Snap index of the Exposure.
+
+    Returns
+    -------
+    entry : `dict`
+        A dictionary containing all fields in the Exposure table.
+    """
+    avg = visitInfo.getDate()
+    begin = DateTime(int(avg.nsecs(DateTime.TAI) - 0.5E9*visitInfo.getExposureTime()), DateTime.TAI)
```
timj, line 98: You mean we can't get the value of `DATE-OBS` from the object but need to guess it by subtracting half the exposure time?

timj (approved):
In general looks good. I have quite a few comments.

config/storageClasses.yaml
TallJimbo: Yeah, it doesn't actually do anything; I just felt that using `<<:` to indicate inheritance even when it was a no-op was a useful bit of code-as-documentation.

timj: It's not too late to add a `parent:` keyword and get proper pythonic inheritance for StorageClasses (I think I can sort out print through of those class properties).

config/storageClasses.yaml
TallJimbo: Yeah, we discussed this on the other ticket.  I'll convert to `PropertySet` here.

timj: I actually left it alone in #76 since it worked regardless.

python/lsst/daf/butler/butler.py
TallJimbo: No, I'm just filtering out parameters that aren't supported by that component.  So, for instance, if you pass a box for a subimage when getting a virtual composite Exposure, it gets passed down to `image`, `mask`, and `variance`, but not `wcs`.

I'll add a code comment.

python/lsst/daf/butler/butler.py
TallJimbo: I'm not sure; the old behavior definitely raised an exception, but it was a pretty incomprehensible one.  My goal here was just it more comprehensible.

python/lsst/daf/butler/core/fileTemplates.py
TallJimbo: :+1: on "can be optional".  Will add a test.

TallJimbo: Oh, actually, yeah, I suppose you could make one optional if the other is mandatory.  Probably best to just remove this particular check, as there is another one to verify that at least one of those is present.

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
TallJimbo: Not at the moment - just `bbox` and `origin`.  I suppose we need a place to document this at the StorageClass level.  I'm open to ideas.

python/lsst/daf/butler/instrument.py
TallJimbo: AFAIK, yes.  I don't know much about `VisitInfo`.  I stole this recipe from `obs.base.makeRawVisitInfo.offsetDate`, and I'd like to replace both that and `VisitInfo` itself, but this seemed like an easy way to get the correct (or at least consistent with Gen2) answer.

config/storageClasses.yaml
TallJimbo: Let's see what happens with [DM-15238](https://jira.lsstcorp.org/browse/DM-15238) first.

config/storageClasses.yaml
TallJimbo: Ok, I think I'll stick with `PropertyList` here, too, then.

TallJimbo:
I think I've addressed all comments (and rebased on latest master).

One question from that rebase (which I should have asked on a previous review, probably): do we still need the `ExposureComposite` StorageClasses?

timj:
Re ExposureComposite: I think we should move it to the test configs so that the tests can try a virtual and concrete composite.

TallJimbo:
> I think we should move it to the test configs so that the tests can try a virtual and concrete composite.

I gave this a quick try on this ticket, but it's not easy because aliases apparently can't be used outside the file where they're defined (so you can't easily make `ExposureComposite` inherit attributes from `Exposure` when they're in different files).

timj:
I was worrying about that cross-yaml file inheritance this morning. Would trivially be fixed if we had true python inheritance as well.  Leave ExposureComposite in there for now -- I really do need to make a test for Exposure that uses virtual and one that uses concrete.

PR #61, opened 2018-07-26
DM-15204: Chained datastores

timj:
Simple chained datastore supporting multiple datastores where a put is sent to all and a get is retrieved from the first item.

### DIFF
--- a/python/lsst/daf/butler/core/config.py
+++ b/python/lsst/daf/butler/core/config.py
@@ -439,6 +439,12 @@ class Config(_ConfigBase):
             Allows for named parameters to be set to new values in bulk, and
             for other values to be set by copying from a reference config.
 
+            Assumes that the supplied config is compatible with ```configType``
+            and will attach the updated values to the supplied config by
+            looking for the related component key.  It is assumed that
+            ``config`` and ``full`` are from the same part of the
+            configuration hierarchy.
+
             Parameters
             ----------
             configType : `ConfigSubset`
@@ -474,6 +480,14 @@ class Config(_ConfigBase):
             if toupdate is None and tocopy is None:
                 raise ValueError("One of toupdate or tocopy parameters must be set.")
 
+            # If this is a parent configuration then we need to ensure that
+            # the supplied config has the relevant component key in it.
+            # If this is a parent configuration we add in the stub entry
+            # so that the ConfigSubset constructor will do the right thing.
+            # We check full for this since that is guaranteed to be complete.
+            if configType.component in full and configType.component not in config:
+                config[configType.component] = {}
+
             # Extract the part of the config we wish to update
             localConfig = configType(config, mergeDefaults=False, validate=False)
 
@@ -486,7 +500,7 @@ class Config(_ConfigBase):
                 for key in tocopy:
                     localConfig[key] = localFullConfig[key]
 
-            # Reattach to parent
+            # Reattach to parent if this is a child config
             if configType.component in config:
                 config[configType.component] = localConfig
             else:
--- a/python/lsst/daf/butler/core/config.py
+++ b/python/lsst/daf/butler/core/config.py
@@ -242,15 +242,31 @@ class Config(_ConfigBase):
             self.update(d)
         data = self.data
         for key in keys:
-            data = data.setdefault(key, {})
-        data[last] = value
+            # data could be a list
+            if isinstance(data, collections.Sequence):
+                data = data[int(key)]
+            else:
+                data = data.setdefault(key, {})
+        try:
+            data[last] = value
+        except TypeError:
+            data[int(last)] = value
 
     def __contains__(self, key):
         d = self.data
         keys = key.split('.')
         last = keys.pop()
         for k in keys:
-            if k in d:
+            if isinstance(d, collections.Sequence):
+                # Check sequence first because for lists
+                # __contains__ checks whether value is found in list
+                # not whether the index exists in list. When we traverse
+                # the hierarchy we are interested in the index.
+                try:
+                    d = d[int(k)]
+                except IndexError:
+                    return False
+            elif k in d:
                 d = d[k]
             else:
                 return False
@@ -280,9 +296,19 @@ class Config(_ConfigBase):
                         r = doUpdate(d.get(k, {}), v)
                         d[k] = r
                     else:
-                        d[k] = u[k]
+                        d[k] = v
                 else:
-                    d = {k: u[k]}
+                    if isinstance(d, collections.Sequence):
+                        try:
+                            intk = int(k)
+                        except ValueError:
+                            # This will overwrite the entire sequence
+                            d = {k: v}
+                        else:
+                            r = doUpdate(d[intk], v)
+                            d[intk] = r
+                    else:
+                        d = {k: v}
             return d
         doUpdate(self.data, other)
 
--- a/python/lsst/daf/butler/core/config.py
+++ b/python/lsst/daf/butler/core/config.py
@@ -232,6 +232,8 @@ class Config(_ConfigBase):
     def __setitem__(self, name, value):
         keys = name.split('.')
         last = keys.pop()
+        if isinstance(value, Config):
+            value = copy.deepcopy(value.data)
         data = self.data
         for key in keys:
             # data could be a list
--- a/python/lsst/daf/butler/core/config.py
+++ b/python/lsst/daf/butler/core/config.py
@@ -432,6 +432,66 @@ class Config(_ConfigBase):
         with open(path, 'w') as f:
             self.dump(f)
 
+    @staticmethod
+    def overrideConfigParameters(configType, config, full, toupdate=None, tocopy=None):
+            """Generic helper function for overriding specific config parameters
+
+            Allows for named parameters to be set to new values in bulk, and
+            for other values to be set by copying from a reference config.
+
+            Parameters
+            ----------
+            configType : `ConfigSubset`
+                Config type to use to extract relevant items from ``config``.
+            config : `Config`
+                A `Config` to update. Only the subset understood by
+                the supplied `ConfigSubset` will be modified. Default values
+                will not be inserted and the content will not be validated
+                since mandatory keys are allowed to be missing until
+                populated later by merging.
+            full : `Config`
+                A complete config with all defaults expanded that can be
+                converted to a ``configType``. Read-only and will not be
+                modified by this method. Values are read from here if
+                ``tocopy`` is defined.
+
+                Repository-specific options that should not be obtained
+                from defaults when Butler instances are constructed
+                should be copied from `full` to `Config`.
+            toupdate : `dict`, optional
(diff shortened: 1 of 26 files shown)


timj:
@TallJimbo I think it's worth you having a look at the current implementation now. The tests now work with a ChainedDatastore and I've had to rewrite the config system a bit and overhaul setConfigRoot to get this to all work. StorageClassFactory had a bit of an overhaul as well to handle the order of definitions not being guaranteed (I am leaving the equality test for new definitions to the other ticket).

python/lsst/daf/butler/core/config.py
```diff
@@ -406,6 +426,80 @@ def dumpToFile(self, path):
         with open(path, 'w') as f:
             self.dump(f)
 
+    @staticmethod
+    def overrideConfigParameters(configType, config, full, toupdate=None, tocopy=None):
```
TallJimbo: Given that the class is already called `Config`, including "Config" in the name of the method seems redundant.

`toupdate`, `tocopy` -> camelCase.

python/lsst/daf/butler/core/datastore.py
```diff
@@ -142,6 +142,11 @@ class Datastore(metaclass=ABCMeta):
     absolute path. Can be None if no defaults specified.
     """
 
+    containerKey = None
+    """Name of the key containing a list of subconfigurations that also
+    need to be merged with defaults and may use different Python types
+    (but the same config class). Assumed to be a list of configs."""
```
TallJimbo: Does this actually need to be here?  Seems like it's something only some subclasses need.  Or was this supposed to be part of `DatastoreConfig` rather than `Datastore`?

python/lsst/daf/butler/core/sqlDatabaseDict.py
```diff
@@ -55,12 +57,13 @@ class SqlDatabaseDict(DatabaseDict):
...
-        Must have a ``_fields`` class attribute that is a tuple of field names
-        (i.e. as defined by `namedtuple`); these field names must also appear
+    value : `type` (`~collections.namedtuple`)
+        The type used for the dictionary's values, typically a
+        `~collections.namedtuple`. Must have a ``_fields`` class attribute
+        that is a tuple of field names (i.e. as defined by
+        `~collections.namedtuple`); these field names must also appear
         in the ``types`` arg, and a `_make` attribute to construct it from a
-        sequence of values (again, as defined by `namedtuple`).
+        sequence of values (again, as defined by `~collections.namedtuple`).
```
TallJimbo, line 66: Just for my own education: what does the `~` do here?

python/lsst/daf/butler/core/utils.py
```diff
@@ -24,10 +24,10 @@
 import importlib
 from collections import namedtuple
 
-
 __all__ = ("iterable", "allSlots", "slotValuesAreEqual", "slotValuesToHash",
            "getFullTypeName", "doImport", "getInstanceOf", "Singleton",
-           "TopologicalSet", "transactional", "getObjectSize")
+           "TopologicalSet", "TopologicalSetNode", "transactional",
```
TallJimbo, line 29: Did you add an external use of `TopologicalSetNode` I missed?  I thought it was an implementation detail.

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -0,0 +1,342 @@
...
+        return
+
+    def __init__(self, config, registry=None):
+        super().__init__(config, registry)
+
+        self.storageClassFactory = StorageClassFactory()
+
+        # Name ourselves with the timestamp the datastore
+        # was created.
+        self.name = "{}@{}".format(type(self).__qualname__, time.time())
```
TallJimbo: If this is timestamp, it's probably not useful as a publicly-visible name; should this be a private attribute?

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -0,0 +1,342 @@
...
+            Reference to the associated Dataset.
+
+        Raises
+        ------
+        TypeError
+            Supplied object and storage class are inconsistent.
+        """
+
+        for datastore in self.datastores:
+            datastore.put(inMemoryDataset, ref)
```
TallJimbo: What does a `Datastore` normally do when it isn't configured to support a particular `DatasetRef` when it's asked to `put` (e.g. a `PosixDatastore` that doesn't have a formatter configured)?

If the answer is "raise an exception", we probably either want to change the behavior of `ChainedDatastore` itself to ignore those failures, or add a new trivial forwarding `Datastore` in between that does.  I think it'll be pretty common to want to write only some `DatasetTypes` to certain `Datastores` in the chain.

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -0,0 +1,342 @@
...
+    def ingest(self, *args, **kwargs):
+        """Add an on-disk file with the given `DatasetRef` to the store,
+        possibly transferring it.
+
+        This method is forwarded to each of the chained datastores, trapping
+        cases where a datastore has not implemented file ingest and ignoring
+        them.
+
+        A transfer mode of None is not supported since that requires the
+        file to have been previously copied to each individual datastore.
```
TallJimbo: Seems like `None` ought to be okay if only one `Datastore` in the chain supports ingest, and the file is already in that one.  I think you'll get the right behavior if you just remove that special check.

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -0,0 +1,342 @@
...
+        ref : `DatasetRef`
+            Reference to the required Dataset.
+
+        Raises
+        ------
+        FileNotFoundError
+            Attempt to remove a dataset that does not exist.
+        """
+        for datastore in self.datastores:
+            datastore.remove(ref)
```
TallJimbo: Given that `get` works if only one `Datastore` has the dataset, this should probably behave like `ingest` and raise only if it wasn't removed anywhere.

TallJimbo (approved):
Looks good!  A few comments, but nothing big or structural.

python/lsst/daf/butler/core/sqlDatabaseDict.py
timj: It writes the documentation as `namedtuple` but tries to link to `collections.namedtuple`.

python/lsst/daf/butler/core/utils.py
timj: Without this the sphinx build did not work. I'll give it another try.

python/lsst/daf/butler/datastores/chainedDatastore.py
timj: The name has to be unique (the name is used in registry) so this makes it unique. I can remove this if the system ensures that only one inMemoryDatastore can ever be created in memory.

python/lsst/daf/butler/datastores/chainedDatastore.py
timj: My real problem here was to understand how a file turned up in one of the data stores in the first place. The test infrastructure doesn't actually know that a chained datastore is being used so I wasn't sure how much special casing I should be including to copy that file in to the right place during the tests (it would have to know to copy the test file to `.datastores[1].root`). I now realize that I should add a second posix datastore to the chaining tests since that would demonstrate that ingest with transfer=move would fail on the second datastore. Then we have to decide whether a single ingest working is good enough or if all implemented ingests should work (transfer=move could work if the chained ingest copied the file itself to a temporary area so that each datastore could move it itself)

python/lsst/daf/butler/datastores/chainedDatastore.py
timj: Oh yes. That shows that my tests need to be extended.

python/lsst/daf/butler/core/datastore.py
timj: `ConfigSubset` needs to know the container key used by each class, so it's definitely a datastore property and not a config property: we only have one DatastoreConfig but each datastore class has a different content in those configs: we could have DatastoreConfig subclasses but we don't do it that way at the moment). Saying that, I can remove this from here because `ConfigSubset` traps the case where the value hasn't been set for the class -- it was here mainly as a documentation aid for the property.

python/lsst/daf/butler/datastores/chainedDatastore.py
TallJimbo: Will this go away when Registry supports ephemeral datastores directly, then?

Also, why does a constraint on `InMemoryDatastore` affect `ChainedDatastore`?

timj: Re InMemoryDatastore: Because I got my wires crossed (InMemoryDatastore does need the timestamp to make it unique). You are right that this name is not used directly in registry because the puts are done by the contained datastores (which need unique names). I could just call this `ChainedDatastore` or else call it `ChainedDatastore[datastore1.name, datastore2.name]`

TallJimbo: Either is fine with me - you probably have a better sense than I do whether we want a more unique and descriptive name here or something more concise.

python/lsst/daf/butler/datastores/chainedDatastore.py
TallJimbo: What about making it so `ingest` is only supported if exactly one `Datastore` accepts the file?  I can't think of a good use case for anything more complex than that (though we would probably want to let different `Datastores` in the chain accept different files).

timj: Hmm. It seems a bit odd to me that ingest would stop at the first success (or continue trying in order to ensure that they all subsequently failed) but `put` sends to each child datastore.

python/lsst/daf/butler/datastores/chainedDatastore.py
timj: I agree that if a datastore put fails in a specific way, we can skip that datastore and allow the chained put to work if at least one put worked (possibly to a non-ephemeral datastore -- we could make that configurable).

python/lsst/daf/butler/core/utils.py
timj: Trying again, when I build the docs I get this:
```
lsst.daf.butler.core.utils.TopologicalSetNode.rst:: WARNING: document isn't included in any toctree```
If I leave out `TopologicalSetNode`. It's only used internally in utils.py so I'm not sure why Sphinx insists it should be listed in `__all__`. Maybe @jonathansick has an idea.

timj:
@TallJimbo do you want to take another look? ChainedDatastore tries to be a bit more robust against child datastores not handling certain datasets. Still no per-datasettype configuration on in-memory or chained datastore handling or cache expiry (which we should discuss punting to a different ticket).

PR #62, opened 2018-07-27
DM-15212: better transactionality and ingest-with-transfer for Datastores

TallJimbo:
Also some includes some minor test refactoring and namespacing fixes to benefit the doc build.

### DIFF
--- a/python/lsst/daf/butler/core/datastore.py
+++ b/python/lsst/daf/butler/core/datastore.py
@@ -104,7 +104,7 @@ class DatastoreTransaction:
             except BaseException as e:
                 # Deliberately swallow error that may occur in unrolling
                 log = logging.getLogger(__name__)
-                log.debug("Exception: %s caught while unrolling: %s", e, name)
+                log.warn("Exception: %s caught while unrolling: %s", e, name)
                 pass
 
     def commit(self):
--- a/python/lsst/daf/butler/core/datastore.py
+++ b/python/lsst/daf/butler/core/datastore.py
@@ -76,6 +76,24 @@ class DatastoreTransaction:
         """
         self._log.append(self.Event(name, undoFunc, args, kwargs))
 
+    @contextlib.contextmanager
+    def undoWith(self, name, undoFunc, *args, **kwargs):
+        """A context manager that calls `registerUndo` if the nested operation
+        does not raise an exception.
+
+        This can be used to wrap individual undo-able statements within a
+        DatastoreTransaction block.  Multiple statements that can fail
+        separately should not be part of the same `undoWith` block.
+
+        All arguments are forwarded directly to `registerUndo`.
+        """
+        try:
+            yield None
+        except BaseException:
+            raise
+        else:
+            self.registerUndo(name, undoFunc, *args, **kwargs)
+
     def rollback(self):
         """Roll back all events in this transaction.
         """
@@ -172,7 +190,7 @@ class Datastore(metaclass=ABCMeta):
         """
         self._transaction = DatastoreTransaction(self._transaction)
         try:
-            yield
+            yield self._transaction
         except BaseException:
             self._transaction.rollback()
             raise
--- a/python/lsst/daf/butler/core/datastore.py
+++ b/python/lsst/daf/butler/core/datastore.py
@@ -246,6 +246,50 @@ class Datastore(metaclass=ABCMeta):
         """
         raise NotImplementedError("Must be implemented by subclass")
 
+    def ingest(self, path, ref, formatter=None, transfer=None):
+        """Add an on-disk file with the given `DatasetRef` to the store,
+        possibly transferring it.
+
+        The caller is responsible for ensuring that the given (or predicted)
+        Formatter is consistent with how the file was written; `ingest` will
+        in general silently ignore incorrect formatters (as it cannot
+        efficiently verify their correctness), deferring errors until ``get``
+        is first called on the ingested dataset.
+
+        Datastores are not required to implement this method, but must do so
+        in order to support direct raw data ingest.
+
+        Parameters
+        ----------
+        path : `str`
+            File path, relative to the repository root.
+        ref : `DatasetRef`
+            Reference to the associated Dataset.
+        formatter : `Formatter` (optional)
+            Formatter that should be used to retreive the Dataset.  If not
+            provided, the formatter will be constructed according to
+            Datastore configuration.
+        transfer : str (optional)
+            If not None, must be one of 'move', 'copy', 'hardlink', or
+            'symlink' indicating how to transfer the file.
+            Datastores need not support all options, but must raise
+            NotImplementedError if the passed option is not supported.
+            That includes None, which indicates that the file should be
+            ingested at its current location with no transfer.  If a
+            Datastore does support ingest-without-transfer in general,
+            but the given path is not appropriate, an exception other
+            than NotImplementedError that better describes the problem
+            should be raised.
+
+        Raises
+        ------
+        NotImplementedError
+            Raised if the given transfer mode is not supported.
+        """
+        raise NotImplementedError(
+            "Datastore does not support direct file-based ingest."
+        )
+
     @abstractmethod
     def getUri(self, datasetRef):
         """URI to the Dataset.
--- a/python/lsst/daf/butler/datastores/posixDatastore.py
+++ b/python/lsst/daf/butler/datastores/posixDatastore.py
@@ -421,11 +421,6 @@ class PosixDatastore(Datastore):
         ------
         FileNotFoundError
             Attempt to remove a dataset that does not exist.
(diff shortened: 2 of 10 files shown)


python/lsst/daf/butler/core/datastore.py
```diff
@@ -76,6 +76,24 @@ def registerUndo(self, name, undoFunc, *args, **kwargs):
         """
         self._log.append(self.Event(name, undoFunc, args, kwargs))
 
+    @contextlib.contextmanager
+    def undoAs(self, name, undoFunc, *args, **kwargs):
```
timj: I'm not sure about the name here. What does "undo as?" mean?

python/lsst/daf/butler/core/datastore.py
```diff
@@ -228,6 +246,45 @@ def put(self, inMemoryDataset, datasetRef):
         """
         raise NotImplementedError("Must be implemented by subclass")
 
+    def ingest(self, path, ref, formatter=None, transfer=None):
```
timj, line 249: I know it is marked optional but we now have two methods in the base class that insist on a path to a file or directory. I am torn about this but I'll let it stand for now.

python/lsst/daf/butler/core/datastore.py
```diff
@@ -228,6 +246,45 @@ def put(self, inMemoryDataset, datasetRef):
...
+        formatter : `Formatter` (optional)
+            Formatter that should be used to retreive the Dataset.  If not
+            provided, the formatter will be constructed according to
+            Datastore configuration.
+        transfer : str (optional)
+            If not None, must be one of 'move', 'copy', 'hardlink', or
+            'symlink' indicating how to transfer the file.
+            Datastores need not support all options, but must raise
+            NotImplementedError if the passed option is not supported
+            (including None).
```
timj: Can you be more explicit in saying that the default `None` indicates that the file must be inside the repository already. Or can it also mean that an absolute path to a file is given and it will be referenced directly?

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -292,67 +293,139 @@ def put(self, inMemoryDataset, ref):
...
-        # Associate this dataset with the formatter for later read.
-        fileInfo = StoredFileInfo(formatter, path, ref.datasetType.storageClass)
-        self.addStoredFileInfo(ref, fileInfo)
-
-        # Register all components with same information
-        for compRef in ref.components.values():
-            self.registry.addStorageInfo(compRef, info)
-            self.addStoredFileInfo(compRef, fileInfo)
+        with self.transaction() as transaction:
+            fullPath = os.path.join(self.root, path)
```
timj: Should this check be inside the transaction?

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -292,67 +293,139 @@ def put(self, inMemoryDataset, ref):
         TypeError
             Supplied object and storage class are inconsistent.
         """
+        with self.transaction() as transaction:
```
timj: registry uses a `@transactional` decorator.  Is there a reason we aren't doing something similar here? It would seem to be cleaner and remove lots of extra indenting. On the other hand you'll be using `self._transaction` everywhere rather than `transaction`.

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -292,67 +293,139 @@ def put(self, inMemoryDataset, ref):
...
+            if not os.path.isdir(storageDir):
+                with transaction.undoAs("mkdir", os.rmdir, storageDir):
+                    safeMakeDir(storageDir)
 
-        # Write the file
-        path = formatter.write(inMemoryDataset, FileDescriptor(location,
-                                                               storageClass=storageClass))
+            # Write the file
+            predictedFullPath = os.path.join(self.root, formatter.predictPath(location))
+            with transaction.undoAs("write", os.remove, predictedFullPath):
```
timj: I didn't initially understand why prediction was needed here, but since we aren't guaranteeing a formatter will clean up a partially written file on error then I think I have to agree that we have no choice but to have path prediction.

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -292,67 +293,139 @@ def put(self, inMemoryDataset, ref):
...
+                path = newPath
+                fullPath = newFullPath
+
+            # Create Storage information in the registry
+            checksum = self.computeChecksum(fullPath)
+            stat = os.stat(fullPath)
+            size = stat.st_size
+            info = StorageInfo(self.name, checksum, size)
+            with self.registry.transaction():
+                self.registry.addStorageInfo(ref, info)
```
timj: `addStorageInfo` is already wrapped in a transaction.

tests/test_datastore.py
```diff
@@ -300,17 +299,145 @@ def testNestedTransaction(self):
...
     """PosixDatastore specialization"""
     configFile = os.path.join(TESTDIR, "config/basic/butler.yaml")
     uriScheme = "file:"
+    ingestTransferModes = (None, "copy", "move", "hardlink", "symlink")
+
+    def setUp(self):
+        super().setUp()
+        self.root = tempfile.mkdtemp(dir=TESTDIR)
+        self.config = DatastoreConfig(self.configFile)
+        self.config["root"] = self.root
```
timj, line 396: I really think these should all be using `setConfigRoot` rather than looking inside the config here (it's what `setConfigRoot` is *for*. I'll make the change myself as part of my ticket since I'm rewriting `setConfigRoot` and this will be a good test for it (and with chained datastores added to these tests it will be even more important to be consistent).

timj (approved):
Took me a while to work out what was going on but it looks good in general.

python/lsst/daf/butler/core/datastore.py
TallJimbo: I was trying to indicate that the first argument (name) described the operation you were trying to undo, and that the second was the operation to undo it, but apparently that only made sense in my head.

I was also toying with `undoWith`; do either of those make more sense to you?  
```
with transaction.undoWith("do a thing", undoThing):
    doThing()
```
I think part of the problem is that the name of the operation is the first argument, but the real target of the context manager is `undoFunc`.  I could just swap them; unfortunately we can't really use kwargs on whatever argument isn't first because we need to be able to forward `*args`.

Any thoughts on how to make this better?

python/lsst/daf/butler/core/datastore.py
TallJimbo: Yeah, I've been thinking about this.  From a pure OO perspective, it'd be very natural to have an intermediate (but still abstract) base class `FileDatastore` that knows about formatters and files and ingest.  All concrete `Datastores` that work with formatters and files and can do ingest would then inherit from that instead of directly from `Datastore` itself, and that might include, say, an S3 Datastore that doesn't have files internally, but certainly knows how to work with them.

Anyhow, we'd then be able to define a  bunch of operations that only work when you have a `FileDatastore`. Ingest is the obvious operation that's relevant here, but I wonder if this could even be extended to the problem of whether you can ask for a checksum or size (i.e. "yes, but only if you have a `FileDatastore`").

Unfortunately, I don't think the derived-class-interface approach works with composition (a `ChainedDatastore` can't inherit from `FileDatastore` if and only if it forwards to a `FileDatastore`).  I think that means we can't use the Python type system to express "is a file datastore", but the "file datastore" interface concept may still be one worth preserving.  That's the direction I was trying to go in here, but it's not really fully thought out and I do agree that we should do that thinking before letting paths and formatters spread too much further into generic datastores.

python/lsst/daf/butler/core/datastore.py
TallJimbo: I'll improve the docs.  Right now it does mean that the file must be inside the repository already; trying to support files outside the repository root was a can of worms I did not want to open here.

python/lsst/daf/butler/datastores/posixDatastore.py
TallJimbo: When I first wrote this I didn't realize I could get access to the transaction via `self._transaction`.  Seeing that that is the pattern used in `Registry` is a good enough reason to switch to it here.  Will do.

python/lsst/daf/butler/datastores/posixDatastore.py
TallJimbo: I don't think that it matters, because there are no undo-able operations that happen before it.  Given that I'm going to refactor these to use `@transactional`, it will remain inside the transaction.

python/lsst/daf/butler/core/datastore.py
timj: I think I prefer `undoWith`.

PR #64, opened 2018-08-02
DM-15207: Useful __str__ for Butler

pschella:


python/lsst/daf/butler/butler.py
```diff
@@ -178,6 +178,9 @@ def __reduce__(self):
         """
         return (Butler, (self.config, ))
 
+    def __str__(self):
+        return "Butler(collection='{}', root='{}')".format(self.collection, self.datastore.root)
```
timj: I think I'd prefer to use `self.datastore.name` here rather than introduce `root` as a top-level datastore concept.

tests/test_butler.py
```diff
@@ -262,11 +262,19 @@ class PosixDatastoreButlerTestCase(ButlerTests, lsst.utils.tests.TestCase):
     """PosixDatastore specialization of a butler"""
     configFile = os.path.join(TESTDIR, "config/basic/butler.yaml")
 
+    def testStringification(self):
+        butler = Butler(self.configFile)
+        self.assertEqual(str(butler), "Butler(collection='ingest', root='./butler_test_repository')")
```
timj: I'm not entirely sure I want this sort of hard-wired testing as I add more datastores and butler configs but maybe we have no choice.

timj (changes requested):
I'm not convinced that adding `root` is better than using `name`. I already use `name` to be a description of the datastore.

python/lsst/daf/butler/butler.py
pschella: I originally wanted to use `Butler(collection='{}', datastore='{}', registry='{}').format(collection, str(self.datastore), str(self.registry))` but @RobertLuptonTheGood preferred to just have the root. What does `self.datastore.name` give you?

tests/test_butler.py
pschella: Yeah, I don't really know of a better way here.

pschella: Appart from repeating the internal code of course.

timj: You could do 
```python
self.assertIn("butler_test_repository", str(butler))
```

python/lsst/daf/butler/butler.py
timj: Please let the datastore tell you how it wants to be stringified. I suggest defining a `__str__` for `Datastore` that defaults to returning `self.name`. `root` is not a standard concept for all datastores.

czwa: I've made this change to implement `__str__` for `Datastore`, and a similar one for `Registry`.  I've also added the registry description to the `Butler` string.

tests/test_butler.py
czwa: I've changed the tests to use `self.assertIn()`, with multiple asserts to test for the expected values.

python/lsst/daf/butler/butler.py
```diff
@@ -178,6 +178,10 @@ def __reduce__(self):
         """
         return (Butler, (self.config, ))
 
+    def __str__(self):
+        return "Butler(collection='{}', datastore='{}', registry='{}')".format(
+            self.collection, str(self.datastore), str(self.registry))
```
timj: Are you sure the `str()` are needed here? Won't `format` do that itself?

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -158,6 +158,10 @@ def __init__(self, config, registry=None):
 
         log.debug("Created %s (%s)", self.name, ("ephemeral" if self.isEphemeral else "permanent"))
 
+    def __str__(self):
+        chainName = ", ".join(map(str, self.datastores))
```
timj: I tend to shy away from `map`. I prefer something a bit more pythonic like:
```python
chainName = ", ".join(str(ds) for ds in self.datastores)
```

python/lsst/daf/butler/core/registry.py
```diff
@@ -122,6 +122,9 @@ def __init__(self, registryConfig, schemaConfig=None, create=False):
         self.config = registryConfig
         self._pixelization = None
 
+    def __str__(self):
+        return None
```
timj: Maybe return a string `"None"` instead?

tests/test_butler.py
```diff
@@ -283,19 +283,35 @@ class PosixDatastoreButlerTestCase(ButlerTests, lsst.utils.tests.TestCase):
     configFile = os.path.join(TESTDIR, "config/basic/butler.yaml")
     fullConfigKey = "datastore.formatters"
 
+    def testStringification(self):
```
timj: To fit in more with the style of these subclasses, please define `testStringification` once in the baseclass and define the parameters here. So in this case:
```python
datastoreStr = "datastore='butler_test_repository'"
registryStr = "registry='sqlite:///:memory'"
```

timj (approved):
Looks good, although I have some comments. Can you please add `__str__` to `Datastore`. And since you are in here, please add `Datastore.__repr__` and make it return `self.name` (since that does include the class name).

tests/test_butler.py
czwa: These all make sense, and the branch has been updated with them.

tests/test_butler.py
```diff
@@ -277,24 +277,38 @@ def testMakeRepo(self):
...
+        if self.datastoreStr is not None:
+            self.assertIn(self.datastoreStr, str(butler))
+        if self.registryStr is not None:
+            self.assertIn(self.registryStr, str(butler))
+
 
 class PosixDatastoreButlerTestCase(ButlerTests, lsst.utils.tests.TestCase):
     """PosixDatastore specialization of a butler"""
     configFile = os.path.join(TESTDIR, "config/basic/butler.yaml")
     fullConfigKey = "datastore.formatters"
```
timj, line 292: Remove this blank line to match the others?

PR #66, opened 2018-08-02
DM-15276: Unify DataUnitRegion and DataUnitJoin

pschella:
Only the latter is left.

### DIFF
--- a/config/schema.yaml
+++ b/config/schema.yaml
@@ -667,11 +667,10 @@ schema:
         type: string
         doc: A string value composed only of letters, numbers, and underscores.
 
-  dataUnitRegions:
-    -
-      relates:
-        - Visit
-        - Sensor
+  dataUnitJoins:
+    VisitSensorRegion:
+      lhs: [Visit]
+      rhs: [Sensor]
       spatial: true
       tables:
         VisitSensorRegion:
@@ -728,7 +727,6 @@ schema:
             - Sensor.camera
             - Sensor.sensor
 
-  dataUnitJoins:
     ExposureRangeJoin:
       doc: >
         A calculated join between Datasets identified with an Exposure
--- a/python/lsst/daf/butler/core/dataUnit.py
+++ b/python/lsst/daf/butler/core/dataUnit.py
@@ -149,7 +149,7 @@ class DataUnit:
 
     @property
     def spatial(self):
-        """Is this a spatial `DataUnitRegion`?
+        """Is this a spatial `DataUnitJoin`?
         """
         return self._spatial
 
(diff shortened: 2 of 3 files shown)


python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -195,15 +195,23 @@ class DataUnitJoin:
         an actual `Table` in many cases because joins are often
         materialized as views (and thus are also not present
         in `Registry._schema._metadata`).
+    relates : `tuple` of `DataUnit`
+        The DataUnits in this relationship.
+    spatial : `bool`, optional
+        Is this a spatial `DataUnit`? If so then it either has a ``region``
```
andy-slac, line 201: DataUnit -> DataUnitJoin?

config/schema.yaml
```diff
@@ -667,11 +667,10 @@ schema:
...
 
-  dataUnitRegions:
-    -
-      relates:
-        - Visit
-        - Sensor
+  dataUnitJoins:
+    VisitSensorRegion:
+      lhs: [Visit]
+      rhs: [Sensor]
```
andy-slac, line 673: Just thinking aloud - maybe we should not call them left or right, they are totally interchangeable.

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -235,47 +251,6 @@ def table(self):
         """
         return getattr(self, '_table', None)
 
-
-class DataUnitRegion:
```
andy-slac, line 239: There is small bunch of comments that still refer to `DataUnitRegion` class (they don't show up on this diff), they need cleanup too.

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -345,7 +314,7 @@ def fromConfig(cls, config, builder=None):
         dataUnitRegistry = cls()
         dataUnitRegistry._initDataUnitNames(config['dataUnits'])
         dataUnitRegistry._initDataUnits(config['dataUnits'], builder)
-        dataUnitRegistry._initDataUnitRegions(config['dataUnitRegions'], builder)
+        # dataUnitRegistry._initDataUnitRegions(config['dataUnitRegions'], builder)
```
andy-slac: Should it be now removed completely together with the method itself?

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -541,14 +510,22 @@ def _initDataUnitJoins(self, config, builder):
                     isView = "sql" in tableDescription
             lhs = frozenset((dataUnitJoinDescription.get('lhs', None)))
             rhs = frozenset((dataUnitJoinDescription.get('rhs', None)))
+            dataUnitNames = lhs | rhs
+            relates = frozenset(self[name] for name in dataUnitNames)
             summarizes = dataUnitJoinDescription.get('summarizes', None)
+            spatial = dataUnitJoinDescription.get("spatial", False)
```
andy-slac, line 491: Inconsistent mix of single and double quotes for string literals all over :slightly_smiling_face: PEP8 says "Pick a rule and stick to it", I guess our rule is "whatever goes".

andy-slac (approved):
Looks OK, few minor comments.

config/schema.yaml
pschella: Fair enough. Maybe we should just go back to `relates` for regular joins as well? The only place where it would make a difference is when you want to say `LEFT OUTER JOIN` or something in the SQL that defines the view and direction makes sense?

pschella:
Sorry, I see now that I have been a bit too sloppy with the pair-coding. I think that is what Chris meant when he said we did 2.5 tickets instead of 2 ;) I'll clean things up a bit more as per your comments.

python/lsst/daf/butler/core/dataUnit.py
pschella: I prefer double quotes everywhere and mostly use that. But I don't speak for everyone so perhaps we should do a poll for `daf_butler`...

config/schema.yaml
andy-slac: OUTER JOIN direction is probably context-dependent so I'm not sure that is a good argument. And I'm not advocating changing it on this ticket, just an idea for future improvements.

PR #67, opened 2018-08-03
Revamp equality test for StorageClass and add tests

timj:
* Add proper `__eq__` method that tests the full definition of a StorageClass
* Add extensive tests of equality and factory inserts.
* Add ability to unregister a storage class from the factory

### DIFF
--- a/python/lsst/daf/butler/core/storageClass.py
+++ b/python/lsst/daf/butler/core/storageClass.py
@@ -23,10 +23,13 @@
 
 import builtins
 import itertools
+import logging
 
 from .utils import doImport, Singleton, getFullTypeName
 from .composites import CompositeAssembler
 from .config import ConfigSubset
+
+log = logging.getLogger(__name__)
 
 __all__ = ("StorageClass", "StorageClassFactory", "StorageClassConfig")
 
@@ -66,11 +69,15 @@ class StorageClass:
         if assembler is not None:
             self._assemblerClassName = assembler
             self._assembler = None
-        else:
-            # We set a default assembler so that a class is guaranteed to
-            # support something.
+        elif components is not None:
+            # We set a default assembler for composites so that a class is
+            # guaranteed to support something if it is a composite.
+            log.debug("Setting default assembler for %s", self.name)
             self._assembler = self.defaultAssembler
             self._assemblerClassName = self.defaultAssemblerName
+        else:
+            self._assembler = None
+            self._assemblerClassName = None
         # The types are created on demand and cached
         self._pytype = None
 
--- a/python/lsst/daf/butler/core/storageClass.py
+++ b/python/lsst/daf/butler/core/storageClass.py
@@ -58,7 +58,8 @@ class StorageClass:
         self.name = name
         self._pytypeName = pytype
         if pytype is None:
-            raise ValueError("All StorageClass definitions require a Python type")
+            self._pytypeName = "object"
+            self._pytype = object
         self._components = components if components is not None else {}
         # if the assembler is not None also set it and clear the default
         # assembler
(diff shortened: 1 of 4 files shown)


timj:
@TallJimbo have we decided whether we allow a `StorageClass` to be constructed without any python type defined. This PR defaults to `object`. If we want to raise an exception instead I need to fix the tests that don't define a pytype.

TallJimbo:
I have a slight preference for defaulting to `object`, at least right now.  Sounds like that's less work, too?

timj:
Defaulting to `object` is what this PR currently does so yes, that is less work.

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -63,8 +69,8 @@ def __init__(self, name, pytype=None, components=None, assembler=None):
         else:
             # We set a default assembler so that a class is guaranteed to
             # support something.
-            self._assemblerClassName = None
-            self._assembler = CompositeAssembler
+            self._assembler = self.defaultAssembler
+            self._assemblerClassName = self.defaultAssemblerName
```
TallJimbo: Is this necessary if the StorageClass does not have any components?

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -181,7 +211,7 @@ def __contains__(self, storageClassOrName):
             True if the supplied string is present, or if the supplied
             `StorageClass` is present and identical.
         """
-        if isinstance(storageClassOrName, "str"):
+        if isinstance(storageClassOrName, str):
```
TallJimbo: Pre-existing, but do we need the check for "identical" here?  I agree we need to compare for equality in `registerStorageClass`, after first checking whether the name exists, but that only uses `str` code path here.  And having
```
(storageClass in StorageClassFactory()) != (storageClass.name in StorageClassFactory())
```
might be confusing.

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -271,3 +301,18 @@ def registerStorageClass(self, storageClass):
...
+        ----------
+        storageClassName : `str`
+            Name of storage class to remove.
+
+        Raises
+        ------
+        KeyError
+            The named storage class is not registered.
+        """
+        del self._storageClasses[storageClassName]
```
TallJimbo: This seems to go against the principle that `StorageClassFactory` is append-only.

TallJimbo (approved):
Two questions about what we want to support, but code looks good.

python/lsst/daf/butler/core/storageClass.py
timj: Yes, but for *testing* purposes it's incredibly useful...

python/lsst/daf/butler/core/storageClass.py
timj: This raises the question of what "in" means. Does it just mean "is there something with this name already registered" or does it also mean "is this particular storageclass already registered".  It combines the semantics of a `in` for `dict` (the key is checked) and `in` for `list` or `set` (the value is checked) but the caller gets to decide whether to check the storage class or the storage class name.

python/lsst/daf/butler/core/storageClass.py
timj: with a tweak to butler.py I can make it so that I only default when components are present. I'll make that change and see what happens.

python/lsst/daf/butler/core/storageClass.py
TallJimbo: Sounds like we should just document that intent, then.

python/lsst/daf/butler/core/storageClass.py
TallJimbo: I don't have a terribly strong opinion on this; it just surprised me a bit so I wanted to raise it as a potential issue.

PR #68, opened 2018-08-08
DM-15336: Decouple DataUnitRegistry from Schema

pschella:


config/dataUnitsConfiguration.yaml
```diff
@@ -0,0 +1,290 @@
...
+
+    Label:
+      doc: >
+        An arbitrary string value.  There is no SQL representation or
+        constraint on the values a Label can take.  Multiple labels are
+        represented (by convention) by a comma-separated string.
+      link:
+      - label
+
+  dataUnitJoins:
```
TallJimbo: Now that DataUnits have their own config file, maybe we should just call this section "joins".

config/schema.yaml
```diff
@@ -1,1290 +1,156 @@
...
+          the id of the quantum that produced this dataset, providing access
+          to fine-grained provenance information.
+          may be null for datasets not produced by running a supertask.
+      -
+        name: assembler
+        type: string
+        doc: >
+          fully-qualified name of an importable assembler object that can be
+          used to construct this dataset from its components.  null for
+          datasets that are not virtual composites.
```
TallJimbo: (looking below) I am a bit surprised to see all of the DataUnit link columns explicitly included here, rather than generating them programmatically from the DataUnit definitions.  Do you have any background on why you went with this approach?

tests/test_dataUnit.py
```diff
@@ -24,15 +24,14 @@
 import lsst.utils
 import lsst.utils.tests
 
-from lsst.daf.butler.core.schema import SchemaConfig
-from lsst.daf.butler.core.dataUnit import DataUnit, DataUnitRegistry
+from lsst.daf.butler.core.dataUnit import DataUnit, DataUnitRegistryConfig, DataUnitRegistry
```
TallJimbo: Please just do `from lsst.daf.butler import <those classes>`; I'd like the `core` subpackage and `dataUnit` module to be implementation details.

tests/test_schema.py
```diff
@@ -77,7 +71,7 @@ def testSchemaCreation(self):
     def assertView(self, tableName, tableDescription):
         """Check that a generated view matches its `tableDescription`.
         """
-        table = self.schema.views[tableName]
+        table = self.schema._views[tableName]
```
TallJimbo: I think this should probably be
```
table = self.schema.tables[tableName]`
```
That should return the same object, but do it via the public API.

config/schema.yaml
```diff
@@ -1,1290 +1,156 @@
...
-          to fine-grained provenance information.
-          May be null for Datasets not produced by running a SuperTask.
-
-      foreignKeys:
+          a string value composed only of letters, numbers, and underscores.
+      foreignkeys:
       -
         src: dataset_type_name
-        tgt: DatasetType.dataset_type_name
+        tgt: datasettype.dataset_type_name
```
TallJimbo: I think this capitalization change and all of the similar ones below are incorrect; SQL doesn't care as it's case-insensitive, but this is confusing for humans.  Table names (so all of the first terms in the `tgt` items should be CamelCase and start with a capital letter).

TallJimbo (approved):
Mostly minor comments on the PR.

config/schema.yaml
andy-slac: SQL in general is case-sensitive. Different implementation may have different rules for case-sensitivity (plus there is also SQLAlchemy with its own rules) but we should always be consistent in naming tables.

config/schema.yaml
natelust: @pschella do you remember why this was?

timj: I don't think that code is there any more. Removed in 84970666d4e472dcf586c9a9be6bc12a058bec9f a week or so back. Assemblers are not in the schema any more.

timj: Hmm, okay this is new code but `assembler` should not be here.

TallJimbo: Note that @timj is right (@natelust, you can just remove `assembler` anywhere you see it in the schema), but the original comment was actually on the (not-shown-in-the-summary) lines below, because there wasn't a good diff line on which to attach that comment.

pschella: @TallJimbo, @natelust because it was **much** easier and clearer to just put the link column descriptions here and have no special code. But if deemed desirable it could be auto-generated of course.

tests/test_dataUnit.py
pschella: They should be indeed. But there may be cases where implementation details (instead of API) are unit tested, and in those cases I'd like to not see them exported from `core`.

config/schema.yaml
pschella: Yes, this is probably a mistake.

tests/test_dataUnit.py
TallJimbo: Why?  That just seems like testing multiple orthogonal implementation details (where the code lives, as well as what it does) simultaneously.

timj: If it's available through `lsst.daf.butler` we should use it. `lsst.daf.butler.core.utils` is not exported since it's internal support tools, but those are needed for tests.

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -287,6 +232,12 @@ def regionColumn(self):
         return None
 
 
+class DataUnitRegistryConfig(ConfigSubset):
+    component = "dataUnitsConfig"
```
timj, line 236: I don't really like having "Config" be in the key name. To be consistent with all the other top level configuration files this would be `dataUnitRegistry`. (there is also some use of `dataUnitsY` and other places `dataUnitX`).

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -314,16 +265,13 @@ def fromConfig(cls, config, builder=None):
...
-            When given, create `sqlalchemy.core.Table` entries for every
-            `DataUnit` table.
+            `Registry` schema configuration containing "dataUnits",
+            "joins" entries.
         """
         dataUnitRegistry = cls()
         dataUnitRegistry._initDataUnitNames(config["dataUnits"])
-        dataUnitRegistry._initDataUnits(config["dataUnits"], builder)
-        dataUnitRegistry._initDataUnitJoins(config["dataUnitJoins"], builder)
+        dataUnitRegistry._initDataUnits(config["dataUnits"])
```
timj, line 272: If `config` is a `SchemaConfig` how is it reading `dataUnits` from it?

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -287,6 +232,12 @@ def regionColumn(self):
         return None
 
 
+class DataUnitRegistryConfig(ConfigSubset):
+    component = "dataUnitsConfig"
+    requiredKeys = ("version", "dataUnits")
+    defaultConfigFile = "dataUnitsConfiguration.yaml"
```
timj: Please call this `dataUnitRegistry.yaml` to be consistent. We aren't suffixing any other config files with `Configuration`.

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -424,55 +372,30 @@ def _initDataUnitNames(self, config):
...
                             self._dataUnitNames.connect(dependency, dataUnitName)
 
-    def _initDataUnits(self, config, builder):
+    def _initDataUnits(self, config):
         """Initialize `DataUnit` entries.
 
         Parameters
         ----------
         config : `Config`
             The `dataUnits` component of a `SchemaConfig`.
```
timj: `DataUnitRegistryConfig`?

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -491,24 +414,15 @@ def _initDataUnits(self, config, builder):
                                       if d.name in self._spatialDataUnits))
                 self._dataUnitRegions[key] = dataUnit
 
-    def _initDataUnitJoins(self, config, builder):
+    def _initDataUnitJoins(self, config):
         """Initialize `DataUnit` join entries.
 
         Parameters
         ----------
         config : `SchemaConfig`
```
timj: `DataUnitRegistryConfig`?

python/lsst/daf/butler/core/schema.py
```diff
@@ -56,20 +58,19 @@ def __init__(self, config=None):
             config = SchemaConfig(config)
         self.config = config
         self.builder = SchemaBuilder()
-        self.dataUnits = DataUnitRegistry.fromConfig(config, self.builder)
+        self.dataUnitConfig = DataUnitRegistryConfig(self.config)
```
timj: DataUnitRegistryConfig is not inside a SchemaConfig.

timj (commented):
I am a bit confused with how `DataUnitRegistryConfig` fits into the configuration hierarchy for a butler. If it's top level then `ButlerConfig` needs to be told about it and you may want to add a test to `makeRepo` to ensure that data unit registry overriding happens.

python/lsst/daf/butler/core/dataUnit.py
TallJimbo: @natelust, I think we should probably change this one if you haven't already.

python/lsst/daf/butler/core/dataUnit.py
TallJimbo: Should be fixed now; Nate and I sort-of pair-programmed a bunch of fixes to how `DataUnitRegistries` are initialized, inspired by your slack comment.  To summarize: DataUnit configuration is now a sibling of Schema configuration - they're both nested within Registry, and are known to `ButlerConfig`.

python/lsst/daf/butler/core/dataUnit.py
TallJimbo: @natelust, we should fix this, too.

python/lsst/daf/butler/core/dataUnit.py
TallJimbo: @natelust and this one.

PR #69, opened 2018-08-08
DM-15364: Use double quotes wherever possible in daf_butler

pschella:


timj (approved):
More changes than I expected 😄

PR #70, opened 2018-08-09
DM-15372: Use temp directories for datastores

timj:
Using fixed names causes name collisions when tests run in parallel. Current status is that this fixes the problem, although there's a bit too much code duplication for my liking.

### DIFF
--- a/tests/test_butler.py
+++ b/tests/test_butler.py
@@ -24,7 +24,8 @@
 
 import os
 import unittest
-from tempfile import TemporaryDirectory
+import tempfile
+import shutil
 import pickle
 
 import lsst.utils.tests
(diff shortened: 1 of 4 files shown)


tests/test_butler.py
```diff
@@ -241,15 +257,17 @@ def testMakeRepo(self):
...
-            Butler.makeRepo(root, standalone=True, createRegistry=False,
-                            config=Config(self.configFile))
-            full = Config(os.path.join(root, "butler.yaml"))
-            butler2 = Butler(root, collection="ingest")
+        # Remove the file created in setUp
+        os.unlink(self.tmpConfigFile)
+
+        Butler.makeRepo(self.root, config=Config(self.configFile))
+        limited = Config(self.configFile)
+        print(limited.ppprint())
```
pschella: On purpose?

TallJimbo (approved):
Looks great!  No comments from me, aside from echoing @pschella's question.

PR #71, opened 2018-08-09
DM-15365: Move SqlRegistry API back down into Registry

pschella:


python/lsst/daf/butler/core/registry.py
```diff
@@ -122,6 +124,10 @@ def __init__(self, registryConfig, schemaConfig=None, create=False):
         self.config = registryConfig
         self._pixelization = None
 
+    @contextlib.contextmanager
+    def transaction(self):
+        raise NotImplementedError("Must be implemented by subclass")
```
TallJimbo: Should document what subclasses should implement it to do.

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+        ValueError
+            If dataId is invalid.
+        """
+        raise NotImplementedError("Must be implemented by subclass")
+
+    @abstractmethod
+    @transactional
+    def registerDatasetType(self, datasetType):
+        """
+        Add a new `DatasetType` to the SqlRegistry.
```
TallJimbo: Should document up here (not necessarily in the one-line docstring, but before `Returns`) that it's not an error to register the same `DatasetType` twice.

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+            the component.
+        component : `DatasetRef`
+            A reference to the component dataset.
+        """
+        raise NotImplementedError("Must be implemented by subclass")
+
+    @abstractmethod
+    @transactional
+    def associate(self, collection, refs):
+        r"""Add existing Datasets to a Collection, possibly creating the
```
timj: I don't think the `r` is needed here since I don't see any back slashes.

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+
+        Raises
+        ------
+        ValueError
+            If dataId is invalid.
+        """
+        raise NotImplementedError("Must be implemented by subclass")
+
+    @abstractmethod
+    @transactional
```
timj, line 217: This `@transactional` doesn't do anything right? Since this method is never called anyway and no-one ever does `super()` to call it.  I assume it's more documenting that this method should be transactional in the derived classes?

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+        ------
+        KeyError
+            Requested named DatasetType could not be found in registry.
+        """
+        raise NotImplementedError("Must be implemented by subclass")
+
+    @abstractmethod
+    @transactional
+    def addDataset(self, datasetType, dataId, run, producer=None):
+        """Add a Dataset to a Collection.
```
TallJimbo: I think it'd be better to say "Adds a Dataset entry to the Registry".  In fact, if the `run.collection is None` (which I think is possible but difficult to achieve) it probably should add the Dataset without adding it to any collection.

python/lsst/daf/butler/core/registry.py
timj: Here and elsewhere `SqlRegistry` should be `Registry`.

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+
+        Parameters
+        ----------
+        run : `Run`
+            Instance to add to the `SqlRegistry`.
+
+        Raises
+        ------
+        ValueError
+            If `run` already exists, but is not identical.
```
timj: Use double backticks for `run`.

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+        Returns
+        -------
+        ref : `DatasetRef`
+            A newly-created `DatasetRef` instance.
+
+        Raises
+        ------
+        ValueError
+            If a Dataset with the given `DatasetRef` already exists in the
+            given Collection.
```
TallJimbo: Also raises something if you violate a foreign key constraint by using nonexistent DataUnit values in the data ID.  Should we have a standard exception for that, too?  Or perhaps just document that the data ID values must already have been registered?  (I suppose the Python equivalent of "undefined behavior" is "undefined exception".)

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+        """
+        raise NotImplementedError("Must be implemented by subclass")
+
+    @abstractmethod
+    def getDataset(self, id):
+        """Retrieve an Dataset.
```
TallJimbo: "an" -> "a"; probably say "Dataset entry" so there's no confusion about InMemoryDatasets.

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+
+        Parameters
+        ----------
+        id : `int`
+            The unique identifier for the Dataset.
```
TallJimbo, line 311: I assume this returns a `DatasetRef`?

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+            imported Datasets.
+        """
+        raise NotImplementedError("Must be implemented by subclass")
+
+    @abstractmethod
+    @transactional
+    def transfer(self, src, expr, collection):
+        r"""Transfer contents from a source `SqlRegistry`, limited to those
+        reachable from the Datasets identified by the expression `expr`,
+        into this `SqlRegistry` and collection them with a Collection.
```
timj: > and collection them with a Collection

Needs some work on the grammar.

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+        collection : `str`
+            An additional Collection collection assigned to the newly
+            imported Datasets.
+        """
+        self.import_(src.export(expr), collection)
+
+    @abstractmethod
+    @transactional
+    def subset(self, collection, expr, datasetTypes):
+        r"""Create a new `Collection` by subsetting an existing one.
```
timj: Is there a `Collection` class? (back ticks indicated that should be true).

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+
+        Parameters
+        ----------
+        src : `SqlRegistry`
+            The source `SqlRegistry`.
+        expr : `str`
+            An expression that limits the `DataUnit`\ s and (indirectly)
+            the Datasets transferred.
+        collection : `str`
+            An additional Collection collection assigned to the newly
```
timj: "Collection collection"?

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+            imported Datasets.
+        """
+        self.import_(src.export(expr), collection)
```
timj: This abstract method isn't abstract. It does something.

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+        """Set the assembler to use for a composite dataset.
+
+        Parameters
+        ----------
+        ref : `DatasetRef`
+            Reference to the dataset for which to set the assembler.
+        assembler : `str`
+            Fully qualified name of the assembler.
+        """
+        raise NotImplementedError("Must be implemented by subclass")
```
TallJimbo: I don't think we use this right now; instead we save the assembler in the `Datastore`'s internals as part of its internal StorageClass.  So let's remove this method for now - it's not hard to add it back in if we decide we need it.  You're welcome to also remove it from the schema on this ticket, but if you don't I'll do it on [DM-15210](https://jira.lsstcorp.org/browse/DM-15210).

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -93,6 +93,118 @@ def transaction(self):
...
+        .. todo::
+
+            Insert checks for `storageClass`, `dataUnits` and `template`.
+        """
+        return isinstance(datasetType, DatasetType)
+
+    def _validateDataId(self, datasetType, dataId):
+        """Check if a dataId is valid for a particular `DatasetType`.
+
+        TODO move this function to some other place once DataUnit relations
```
timj: Use same TODO syntax as for `_isValidDatasetType`.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -1175,38 +1158,40 @@ def transfer(self, src, expr, collection):
         """
         self.import_(src.export(expr), collection)
 
-    def makeDatabaseDict(self, table, types, key, value):
-        """Construct a DatabaseDict backed by a table in the same database as
-        this Registry.
+    @transactional
+    def subset(self, collection, expr, datasetTypes):
+        r"""Create a new `Collection` by subsetting an existing one.
```
timj: Again there is no `Collection` class.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -1175,38 +1158,40 @@ def transfer(self, src, expr, collection):
...
-        # but I strongly believe in having signatures that only take what they
-        # really need.
-        config = Config()
-        config["table"] = table
-        return SqlRegistryDatabaseDict(config, types=types, key=key, value=value, registry=self)
+        raise NotImplementedError("Must be implemented by subclass")
+
+    @transactional
+    def merge(self, outputCollection, inputCollections):
+        r"""Create a new Collection from a series of existing ones.
```
timj: Do not need `r"""` here.

timj (approved):
Looks okay. That's a large API surface. I have plenty of minor comments but otherwise it's good to go.

python/lsst/daf/butler/core/registry.py
TallJimbo: `r"..."` string intentional here?

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+    def addExecution(self, execution):
+        """Add a new `Execution` to the `SqlRegistry`.
+
+        Parameters
+        ----------
+        execution : `Execution`
+            Instance to add to the `SqlRegistry`.
+            The given `Execution` must not already be present in the
+            `SqlRegistry`.
+        """
```
TallJimbo: Should document what states of `execution.id` are permitted and how they are (to be) handled.

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+        ----------
+        id : `int`
+            The unique identifier for the Execution.
+        """
+        raise NotImplementedError("Must be implemented by subclass")
+
+    @abstractmethod
+    @transactional
+    def makeRun(self, collection):
+        """Create a new `Run` in the `SqlRegistry` and return it.
```
TallJimbo: Should just be `Registry`.

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+
+        If a run with this collection already exists, return that instead.
```
TallJimbo: With this behavior, we may not need both `makeRun(collection)` and `ensureRun(Run(collection))`; I think the latter can do everything the former can do.

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+            The given `Run` must not already be present in the `SqlRegistry`
+            (or any other).  Therefore its `id` must be `None` and its
+            `collection` must not be associated with any existing `Run`.
+
+        Raises
+        ------
+        ValueError
+            If a run already exists with this collection.
+        """
+        raise NotImplementedError("Must be implemented by subclass")
```
TallJimbo, line 548: I posit that we don't need this as a public API (only `ensureRun`).  Either way, I also think we should relax the requirement that no existing `Run` may have the same `collection`, and simply warn that having multiple `Run`s with the same `collection` will make it impossible to retrieve the `Run` via its `collection`.

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+
+    @abstractmethod
+    def getRun(self, id=None, collection=None):
+        """
+        Get a `Run` corresponding to it's collection or id
```
TallJimbo: `it's` -> `its`.

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+
+        Returns
+        -------
+        run : `Run`
+            The `Run` instance.
+
+        Raises
+        ------
+        ValueError
+            Must supply one of ``collection`` or ``id``.
```
TallJimbo, line 570: Should raise something if `collection` is used but there are multiple `Run`s with that `collection`.

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+            Instance to add to the `SqlRegistry`.
+            The given `Quantum` must not already be present in the
+            `SqlRegistry` (or any other), therefore its:
+
+            - `execution` attribute must be set to an existing `Execution`.
+            - `run` attribute must be set to an existing `Run`.
+            - `predictedInputs` attribute must be fully populated with
+              `DatasetRef`\ s, and its.
+            - `actualInputs` and `outputs` will be ignored.
+        """
```
TallJimbo, line 590: Lots of `SqlRegistry`s in this docstring.

Also, I think it would be better if this automatically added a new `Execution` rather than requiring it to exist.

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+        raise NotImplementedError("Must be implemented by subclass")
+
+    @abstractmethod
+    @transactional
+    def markInputUsed(self, quantum, ref):
+        """Record the given `DatasetRef` as an actual (not just predicted)
+        input of the given `Quantum`.
+
+        This updates both the `SqlRegistry`"s `Quantum` table and the Python
```
TallJimbo: `SqlRegistry`.

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+            Will be updated in this call.
+        ref : `DatasetRef`
+            To set as actually used input.
+
+        Raises
+        ------
+        ValueError
+            If ``ref`` is not already in the predicted inputs list.
+        KeyError
+            If ``ref`` is not a predicted consumer for ``quantum``.
```
TallJimbo: I think this should say `ref` and `quantum` should be reversed here - but then I don't understand how this differs from the `ValueError` condition.

python/lsst/daf/butler/core/registry.py
pschella: Indeed, transactionality is (or is not, depending on what we want) part of the API.

python/lsst/daf/butler/core/registry.py
pschella: An abstract method can have an implementation. E.g. convenient place for a default. Now here this was not the intention of course...

python/lsst/daf/butler/registries/sqlRegistry.py
pschella: There used to be ;) Will fix.

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+        Returns
+        -------
+        graph : `QuantumGraph`
+            Instance (with `units` set to `None`).
+        """
+        raise NotImplementedError("Must be implemented by subclass")
+
+    @abstractmethod
+    def export(self, expr):
+        """Export contents of the `SqlRegistry`, limited to those reachable
```
TallJimbo: `SqlRegistry`

python/lsst/daf/butler/core/registry.py
```diff
@@ -130,4 +136,714 @@ def pixelization(self):
...
+            Containing all rows, from all tables in the `SqlRegistry` that
+            are reachable from the selected Datasets.
+        """
+        raise NotImplementedError("Must be implemented by subclass")
+
+    @abstractmethod
+    @transactional
+    def import_(self, tables, collection):
+        """Import (previously exported) contents into the (possibly empty)
+        `SqlRegistry`.
```
TallJimbo: `SqlRegistry`

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -93,6 +93,118 @@ def transaction(self):
...
+        datasetCollectionTable = self._schema.tables["DatasetCollection"]
+        dataIdExpression = and_((self._schema.dataUnits.links[name] == dataId[name]
+                                 for name in self._schema.dataUnits.getPrimaryKeyNames(
+                                     datasetType.dataUnits)))
+        result = self._connection.execute(select([datasetTable.c.dataset_id]).select_from(
+            datasetTable.join(datasetCollectionTable)).where(and_(
+                datasetTable.c.dataset_type_name == datasetType.name,
+                datasetCollectionTable.c.collection == collection,
+                dataIdExpression))).fetchone()
+        # TODO update unit values and add Run, Quantum and assembler?
```
TallJimbo, line 211: We should probably have some options in the public API controlling how much we do here.  Let's try just having a `full=False` option, where calling with `full=True` would add optional DataUnit link values (e.g. `physical_filter` when `visit` is present), Quantums, and (until we get rid of it) assembler.

TallJimbo (approved):
I've made lots of doc and API-edge-case comments; we may want to spin at least some of the latter into new tickets.

I also wonder if we should just remove some of the APIs that we haven't implemented at all yet, pending a use case that clarifies how they ought to work in detail.

python/lsst/daf/butler/core/registry.py
TallJimbo: In terms of whether derived classes are required to support transactions, I think I'd say that they're required to support the transaction *interface* so one can write sometimes-transaction-safe registry-generic code, but they don't actually have to *be* transactional.  A dummy transaction implementation *might* be something we'd want to put in the base class.

python/lsst/daf/butler/core/registry.py
pschella: Changed to `associate`.

python/lsst/daf/butler/registries/sqlRegistry.py
pschella: Agreed, but probably best done on a new ticket.

python/lsst/daf/butler/core/registry.py
pschella: Probably true, but perhaps best done on a new (mini) ticket?

python/lsst/daf/butler/core/registry.py
pschella: See above.

python/lsst/daf/butler/core/registry.py
pschella: Right, but it doesn't now. Perhaps include that in a cleanup-run ticket?

python/lsst/daf/butler/core/registry.py
pschella: Ah, it actually already does.

PR #72, opened 2018-08-10
DM-15268: fix ci_hsc for gen3-middleware merge

TallJimbo:


timj (approved):
Looks okay (the logging changes shouldn't have to be made but I understand that we haven't yet got other tools to interact well with lsst.log and logging).

PR #73, opened 2018-08-11
DM-15402: Fix race conditions in tests and enhance Config test coverage

timj:
Fix a race condition and whilst doing that we needed to update test coverage of config (there was concern that the tests did not show a failure when they should have).

### DIFF
--- a/python/lsst/daf/butler/core/config.py
+++ b/python/lsst/daf/butler/core/config.py
@@ -85,12 +85,12 @@ class Loader(yaml.CLoader):
         elif isinstance(node, yaml.SequenceNode):
             result = []
             for filename in self.construct_sequence(node):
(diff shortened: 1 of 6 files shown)


TallJimbo (approved):
Looks great!  Glad you didn't take my word in my misdiagnosis of the last race condition I saw.

Only comment I have is that it might be good to put the Datastore test helper in a `tests` subpackage of the main `lsst.daf.butler` package, so it could be used to test hypothetical `Datastores` defined in other packages.

timj:
I'll merge this now and worry about a tests module later. My main reason is that creating `lsst.daf.butler.tests` means I'll have to write much more extensive documentation and ensure that all the assumed state is known.

PR #74, opened 2018-08-15
DM-15447: Fix for registry setDataUnitRegion() method

andy-slac:
DataUnitRegistry uses different keys for different mappings, need to
take care of passing correct set of units to each method. Also added
trivial unit tests for this issue and for DataUnitRegistry, these tests
depend on current schema.yaml definitions.

### DIFF
--- a/python/lsst/daf/butler/registries/sqlRegistry.py
+++ b/python/lsst/daf/butler/registries/sqlRegistry.py
@@ -931,8 +931,7 @@ class SqlRegistry(Registry):
         except IntegrityError as err:
             raise ValueError(str(err))  # TODO this should do an explicit validity check instead
         if region is not None:
-            self.setDataUnitRegion(
-                (dataUnitName,) + tuple(d.name for d in dataUnit.requiredDependencies), v, region)
+            self.setDataUnitRegion((dataUnitName,), v, region)
 
     def findDataUnitEntry(self, dataUnitName, value):
         """Return a `DataUnit` entry corresponding to a `value`.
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -970,7 +969,9 @@ def setDataUnitRegion(self, dataUnitNames, value, region, update=True):
         ----------
         dataUnitNames : sequence
             A sequence of DataUnit names whose instances are jointly associated
-            with a region on the sky.
+            with a region on the sky. This should not include dependencies that
+            are implied, e.g. "Patch" should not include "Tract", but "Sensor"
+            should add "Visit".
```
TallJimbo: Is this "should not" actually a "need not" or a "must not"?  I'm a bit worried that in some contexts it may be difficult to *remove* dependendencies if they are already present.

TallJimbo (approved):
Looks good!  One small question, but thanks for the quick fix.

python/lsst/daf/butler/registries/sqlRegistry.py
andy-slac: With the current implementation I think it means "must not". In general I'm a bit confused about when one should be expected to use one over another (with dependencies vs without). Would it be simpler to require no implicit dependencies and add them internally if they are needed?
Should I change comment to "must not"?

TallJimbo: For now, it's probably best to just change the comment to "must not" and get the bugfix merged.  This is probably best addressed on or after DM-15034.

PR #76, opened 2018-08-21
DM-15471: Control disassembly at butler config level not storage class

timj:
Previously StorageClasses defined how to disassembly, but if we want per-datasetType control over disassembly we have to move them into their own configuration area.

### DIFF
--- a/config/composites.yaml
+++ b/config/composites.yaml
@@ -1,0 +1,17 @@
+composites:
+  # True: concrete composite (composites are disassembled);
+  # False: virtual composite (composites are written monolithically).
+  # default applies to all composites that are not explicitly mentioned
+  # elsewhere.
+  default: False
+  storageClasses:
+    # Use a dict rather than a list to allow easy merging of user config
+    # into default config. Config class has no way of merging lists.
+    ExposureI: False
+    ExposureF: False
+    ExposureCompositeF: True
+    ExposureCompositeF: True
+  datasetTypes:
+    # Same definition as for storage classes, but dataset type explicitly
+    # overrides a storage class definition.
+    calexp: False
--- a/config/composites.yaml
+++ b/config/composites.yaml
@@ -9,7 +9,7 @@ composites:
     # into default config. Config class has no way of merging lists.
     ExposureI: False
     ExposureF: False
-    ExposureCompositeF: True
+    ExposureCompositeI: True
     ExposureCompositeF: True
   datasetTypes:
     # Same definition as for storage classes, but dataset type explicitly
--- a/config/datastores/posixDatastore.yaml
+++ b/config/datastores/posixDatastore.yaml
@@ -35,3 +35,5 @@ datastore:
     Background: lsst.daf.butler.formatters.fitsCatalogFormatter.FitsCatalogFormatter
     Config: lsst.daf.butler.formatters.pexConfigFormatter.PexConfigFormatter
     Packages: lsst.daf.butler.formatters.pickleFormatter.PickleFormatter
+    PropertyList: lsst.daf.butler.formatters.pickleFormatter.PickleFormatter
+    PropertySet: lsst.daf.butler.formatters.pickleFormatter.PickleFormatter
--- a/config/storageClasses.yaml
+++ b/config/storageClasses.yaml
@@ -39,7 +39,7 @@ storageClasses:
     pytype: lsst.daf.base.PropertyList
   Exposure: &Exposure
     pytype: lsst.afw.image.Exposure
-    assembler: lsst.daf.butler.assemblers.exposureAssembler.ExposureAssemblerMonolithic
+    assembler: lsst.daf.butler.assemblers.exposureAssembler.ExposureAssembler
     components:
       image: ImageF
       mask: MaskX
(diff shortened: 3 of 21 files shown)


config/datastores/posixDatastore.yaml
```diff
@@ -35,3 +35,5 @@ datastore:
     Background: lsst.daf.butler.formatters.fitsCatalogFormatter.FitsCatalogFormatter
     Config: lsst.daf.butler.formatters.pexConfigFormatter.PexConfigFormatter
     Packages: lsst.daf.butler.formatters.pickleFormatter.PickleFormatter
+    PropertyList: lsst.daf.butler.formatters.pickleFormatter.PickleFormatter
+    PropertySet: lsst.daf.butler.formatters.pickleFormatter.PickleFormatter
```
TallJimbo, line 39: Probably not an issue for this ticket, but we should have and use a YAML formatter for these.

config/composites.yaml
```diff
@@ -0,0 +1,17 @@
...
+  # default applies to all composites that are not explicitly mentioned
+  # elsewhere.
+  default: False
+  storageClasses:
+    # Use a dict rather than a list to allow easy merging of user config
+    # into default config. Config class has no way of merging lists.
+    ExposureI: False
+    ExposureF: False
+    ExposureCompositeF: True
+    ExposureCompositeF: True
```
TallJimbo: Is it possible to set these options for the `Exposure` "base" StorageClass and have it propagate to "derived" StorageClasses?

python/lsst/daf/butler/core/composites.py
```diff
@@ -0,0 +1,129 @@
...
+            log.debug("%s will not be disassembled (not a composite)", entity)
+            return False
+
+        matchName = "{} (via default)".format(entity)
+
+        # if we were given a str the default is always False
+        if isStr:
+            disassemble = False
+        else:
+            disassemble = self.config["default"]
```
TallJimbo: I don't understand why the default should be different for `str` vs. `DatasetType` or `StorageClass` instances.

config/storageClasses.yaml
```diff
@@ -50,7 +52,7 @@ storageClasses:
       visitInfo: TablePersistableVisitInfo
       apCorrMap: TablePersistableApCorr
       coaddInputs: TablePersistableCoaddInputs
-      metadata: PropertyList
+      metadata: PropertySet
```
TallJimbo: Huh.  Just an FYI: I was under the impression that the thing returned by `Exposure.getMetadata()` was always a `PropertyList`, but it turns out that's because it's implemented that way and pybind11 always downcasts to the most-derived type; the C++ does say `PropertySet`.  I'm inclined to try to get that changed in afw, because I think we really do depend on it being a `PropertyList` in various places, so this will probably get reverted eventually.

TallJimbo (approved):
Looks good!  A few minor comments (on individual commits), but nothing I actually object to.

config/datastores/posixDatastore.yaml
timj: I need to work out what is going on with DM-14503 -- in that ticket YAML support was added to PropertyList by adding it directly to daf_persistence (lsst/daf_persistence#94). Should we add YAML support natively to daf_base itself so that `yaml.dump()` works independently of daf_persistence? Is the issue that we didn't want daf_base to depend on yaml (unlike pickle which is added natively)? @ktlim ?

config/composites.yaml
timj: StorageClass doesn't have a hierarchy in python land so that's not possible. We could change that by changing the way we define storage classes to make them explicitly state a parent rather than using YAML substitution.

TallJimbo: Ok, let's just hope [DM-15238](https://jira.lsstcorp.org/browse/DM-15238) makes this a moot point.

config/datastores/posixDatastore.yaml
TallJimbo: Oh, I hadn't realized the YAML I/O was not in daf_base.  I'd definitely like to see it moved there, though I wouldn't object to having daf_butler depend on daf_persistence temporarily.

config/composites.yaml
andy-slac: Unrelated - there are two `ExposureCompositeF`, should one of them be `ExposureCompositeI`?

python/lsst/daf/butler/core/composites.py
timj: For DatasetType and Storageclass you know for a fact that the requested item is a composite and therefore the default matters. For a `str` you have no idea whether you are being given a composite and returning True for disassembly could be very dangerous.

One option is to drop `str` support completely.

TallJimbo: > One option is to drop str support completely.

If it doesn't make other code uglier, I'm all for it.  Seems like this is a pretty internal API so I'd lean towards simplifying it even if that reduces flexibility.

config/composites.yaml
timj: I don't think true python inheritance of storage classes is all that hard if we need it.

PR #77, opened 2018-08-24
Enhance file template defaulting behavior

timj:
* Allow explicit disabling of default value
* Extra tests

I don't think this interacts with #59 (which I see just merged so I'll rebase).

### DIFF
--- a/tests/test_templates.py
+++ b/tests/test_templates.py
@@ -21,9 +21,13 @@
 
 """Test file name templating."""
 
+import os.path
 import unittest
 
-from lsst.daf.butler import DatasetType, DatasetRef, FileTemplate, StorageClass, Run
+from lsst.daf.butler import DatasetType, DatasetRef, FileTemplates, \
+    FileTemplate, FileTemplatesConfig, StorageClass, Run
+
+TESTDIR = os.path.abspath(os.path.dirname(__file__))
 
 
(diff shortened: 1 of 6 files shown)


tests/test_templates.py
```diff
@@ -105,6 +109,42 @@ def testComponent(self):
...
+        config3 = os.path.join(configRoot, "templates-nodefault2.yaml")
+        templates = FileTemplates(config3)
+        with self.assertRaises(KeyError):
+            templates.getTemplate(ref2.datasetType.name)
+
+        # Try again but specify a default in the constructor
+        default = "{datasetType}/{filter}"
+        templates = FileTemplates(config3, default=default)
+        tmpl = templates.getTemplate(ref2.datasetType.name)
+        self.assertEqual(tmpl.template, default)
```
TallJimbo, line 153: This probably will interact with DM-15189, because a template without `{run}` or `{collection}` is now invalid.

TallJimbo (approved):
One guess at an incompatibility with DM-15189 that you may have already discovered while rebasing.  Otherwise looks good.

tests/test_templates.py
timj: I don't ever expand the template in these tests so it's not a problem. The only conflict was in the import line.

PR #78, opened 2018-08-24
DM-15214: extend pre-flight with output datasets information

andy-slac:
Pre-flight API has changed to support extra information about existing output (and input) datasets, it now returns pre-made `DatasetRefs` instead of raw columns (via new class `PreFlightUnitsRow`). 

Implemented support for specifying multiple input collections and per-dataset collections overrides (new `PreFlightCollections` class).

Pre-flight ginormous query had been extended to support priority-based search of input datasets in multiple collections (see docstring in sqlPreFlight).


python/lsst/daf/butler/core/preFlight.py
```diff
@@ -0,0 +1,144 @@
...
+        self.defaultOutput = defaultOutput
+        self.inputOverrides = inputOverrides or {}
+        self.outputOverrides = outputOverrides or {}
+
+    def getInputCollections(self, datasetTypeName):
+        """Return ordered list of input collections for given dataset type.
+
+        Returns
+        -------
+        `list` of `str`, names of input collections.
```
timj: Returns are meant to be written as:
```
Returns
-------
collections : `list` of `str`
    Names of input collections
```

andy-slac: Do we have to give a name to returned item or can I just write

    Returns
    -------
    `list` of `str`
        Names of input collections

❔

timj: https://developer.lsst.io/python/numpydoc.html?highlight=numpydoc#py-docstring-returns suggests a dummy name is needed.

python/lsst/daf/butler/core/preFlight.py
```diff
@@ -0,0 +1,148 @@
...
+    same `DatasetRef` for output dataset type. It is caller's responsibility
+    to combine multiple `PreFlightUnitsRow` into a suitable structure
+    (e.g. QuantumGraph).
+
+    Attributes
+    ----------
+    dataId : `dict`
+        Maps DataUnit link name to its corresponding value.
+    datasetRefs : `dict`
+        Maps `DatasetType` to its corresponding `DatasetRef`.
```
TallJimbo, line 130: Are the keys actually `DatasetType` instances, or the `str` names associated with them?

python/lsst/daf/butler/core/registry.py
```diff
@@ -700,39 +700,36 @@ def getRegion(self, dataId):
...
+        expression : `str`
+            An expression that limits the `DataUnits <DataUnit>` and
+            (indirectly) the Datasets returned.
         neededDatasetTypes : `list` of `DatasetType`
-            The `list` of `DatasetType`\ s whose instances should be included
-            in the graph and limit its extent.
+            The `list` of `DatasetTypes <DatasetType>` whose DataUnits will
+            be included in the returned column set. Output is limited by the
+            the Datasets of these DatasetTypes which already exist in the
+            registry.
```
TallJimbo: "limited by" -> "limited to"

python/lsst/daf/butler/core/registry.py
```diff
@@ -700,39 +700,36 @@ def getRegion(self, dataId):
...
         futureDatasetTypes : `list` of `DatasetType`
-            The `list` of `DatasetType`\ s whose instances may be added to the
-            graph later, which requires that their `DataUnit` types must be
-            present in the graph.
+            The `list` of `DatasetTypes <DatasetType>` whose DataUnits will
+            be included in the returned column set. It is expected that
+            Datasets for these DatasetTypes do not exist in the registry,
+            but presently this is not checked.
```
TallJimbo, line 727: I think `selectDataUnits` would be more generally useful outside of preflight (e.g. for selecting subsets for export/transfer) if we never actually check for `futureDatasetTypes` not existing in the registry.
In fact, it might be most useful to have this take just take a list of tuples of DataUnits, rather than a list of `DatasetTypes`.  I think that's all preflight uses this for, right?

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -63,44 +65,63 @@ def _filterSummarizes(dataUnitJoins):
...
+def _unitsTopologicalSort(dataUnits):
+    """Return topologically sorted DataUnits.
 
-    Result set generated by query in selectDataUnits() method can include
-    set of regions in each row (encoded as bytes). Due to pixel-based
-    matching some regions may not overlap, this generator method filters
-    rows that have disjoint regions. If result row contains more than two
-    regions (this should not happen with our current schema) then row is
-    filtered if any of two regions are disjoint.
+    Oredering is based on dependencies, units with no dependencies
```
TallJimbo: typo: "Oredering"

Also, FYI, I've added a note to DM-15034 that it should obsolete this code by unifying the topological sorting done both here and in DataUnitRegistry.  Right now I don't see any easy way to combine them.

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -63,44 +65,63 @@ def _filterSummarizes(dataUnitJoins):
...
+
+    Parameters
+    ----------
+    fromClause : `sqlalchemy.FromClause`
+        May be ``None``, in that case ``otherDataUnits`` is expected to be
+        empty and is ignored.
+    dataUnit : `DataUnit`
+        DataUnit to join with ``fromClause``.
+    otherDataUnits : iterable of `DataUnit`
+        DataUnits whose tableshave PKs for ``dataUnit`` table's FK. They all
```
TallJimbo: typo: "tableshave"

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -63,44 +65,63 @@ def _filterSummarizes(dataUnitJoins):
...
+        for otherUnit in otherDataUnits:
+            for name, col in otherUnit.primaryKeyColumns.items():
+                joinOn.append(dataUnit.table.c[name] == col)
+            _LOG.debug("join %s with %s on columns %s", dataUnit.name,
+                       otherUnit.name, list(otherUnit.primaryKeyColumns.keys()))
+        if joinOn:
+            return fromClause.join(dataUnit.table, and_(*joinOn))
+        else:
+            # need explicit cross join here, ugly with SQLAlchemy
+            return fromClause.join(dataUnit.table, literal(True))
```
TallJimbo: Assuming this is what generates those `JOIN <table> ON :param1` lines you mentioned in Jira, it'd probably be good to expand this comment to describe better what the problem is and how this solves it.

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -148,21 +168,30 @@ def selectDataUnits(self, collections, expr, neededDatasetTypes, futureDatasetTy
...
+            Single row is a unique combination of units in a transform.
         """
 
-        # for now only a single collection is supported
-        if len(collections) != 1:
-            raise ValueError("Only single collection is supported by makeDataGraph()")
-        collection = collections[0]
+        # Brief overview of the code below:
+        #  - extract all DataUnits used by all input/output dataset types
+        #  - build a complex SQL query to run agains registry database:
```
TallJimbo: typo: "agains"

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -148,21 +168,30 @@ def selectDataUnits(self, collections, expr, neededDatasetTypes, futureDatasetTy
...
+        #    - first do (natural) join for all tables for all DataUnits
+        #      involved based on their foreing keys
```
TallJimbo: typo: "foreing"

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -230,68 +253,298 @@ def selectDataUnits(self, collections, expr, neededDatasetTypes, futureDatasetTy
...
+                Dataset.DataId
+            FROM Dataset JOIN DatasetCollection
+                ON Dataset.dataset_id = DatasetCollection.dataset_id
+            WHERE Dataset.dataset_type_name = <dsType.name>
+                AND DatasetCollection.collection IN (<collections>)
+
+        (here ``Dataset.DataId`` means ``Dataset.link1, Dataset.link2, etc.``)
+
+        Filtering is complicated, it is simpler to use CTE but not all
+        databases support CTEs so we will have to do with the repeating
```
TallJimbo: What's "CTE" here?

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -230,68 +253,298 @@ def selectDataUnits(self, collections, expr, neededDatasetTypes, futureDatasetTy
...
+                WHERE Dataset.dataset_type_name = <dsType.name>
+                   AND DatasetCollection.collection IN (<collections>)
+                GROUP BY Dataset.DataId) DSG
+            ON DS.colpos = DSG.colpos AND DS.DataId = DSG.DataId
+
+        Parameters
+        ----------
+        dsType : `DatasetType`
+        collections : `PreFlightCollections`
+            Object which provides names of the input/output collections.
```
TallJimbo: I find using "collections" for the name of this argument a bit confusing; that suggests that it would be the same list of strings that's passed to `selectDataUnits`.  I think the best solution is to also come up with a new name for the `PreFlightCollections` class as well: how about calling the class something like `DatasetOriginInfo`, and this argument `originInfo`?

As I side note, I would eventually like to drop "Preflight" from the names of all of your classes in daf_butler, because I think they will eventually be much more generally useful.  But I'm okay with doing that on another ticket, as I think there will also be some refactoring involved, unless you want to do it on this one.

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -230,68 +253,298 @@ def selectDataUnits(self, collections, expr, neededDatasetTypes, futureDatasetTy
...
+        DataIds are all unique (by DataId I mean combination of all link
+        values relevant for this dataset), in that case subquery can be
+        written as:
+
+            SELECT Dataset.dataset_id AS dataset_id, Dataset.link1 AS link1 ...
+            FROM Dataset JOIN DatasetCollection
+                ON Dataset.dataset_id = DatasetCollection.dataset_id
+            WHERE Dataset.dataset_type_name = :dsType_name
+                AND DatasetCollection.collection = :collection_name
+
```
TallJimbo, line 352: Probably good to add a reminder note here that output Dataset subqueries always have this form (if they exist at all).

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -230,68 +253,298 @@ def selectDataUnits(self, collections, expr, neededDatasetTypes, futureDatasetTy
...
+            Dictionary of (DataUnit name, column index), column contains
+            encoded region data
+        dsIdColumns : `dict`
+            Dictionary of (DatasetType, column index), column contains
+            dataset Id, or None if dataset does not exist
+
+        Yields
+        ------
+        row : `PreFlightUnitsRow`
+        """
```
TallJimbo, line 517: This is mostly about the future, but I think we'll eventually want to make this functionality a public API, and that probably means making it into a class in order to make it easier for callers to construct consistent arguments to it.
That's because I think this is the natural place where we could accept custom SQL to drive preflight: while the query you're constructing is quite complex, the form of its outputs really aren't, and it's quite conceivable that someone who wanted to do something complex that's not supported by the query generator could write their own query that generates consistent outputs.  We could then pass the outputs of that custom query through this function and use them in the rest of preflight.
Another option would be to start using custom queries after this function, i.e. make the alternate interface some other iterator that yields `PreFlightUnitsRow`.  That would already be an easier-to-use interface, though it would be nice to provide a general way to filter out non-overlapping regions.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -1045,41 +1045,39 @@ def getRegion(self, dataId):
...
+        collections : `PreFlightCollections`
+            Object which provides names of the input/output collections.
+        expression : `str`
+            An expression that limits the `DataUnits <DataUnit>` and
+            (indirectly) the Datasets returned.
         neededDatasetTypes : `list` of `DatasetType`
-            The `list` of `DatasetType`\ s whose instances should be included
-            in the graph and limit its extent.
+            The `list` of `DatasetTypes <DatasetType>` whose DataUnits will
+            be included in the returned column set. Output is limited by the
```
TallJimbo: "limited to", as above.

TallJimbo (approved):

python/lsst/daf/butler/core/preFlight.py
andy-slac: They are `DatasetType` instances, this is probably not a hard requirement, can switch to names if that is more natural.

python/lsst/daf/butler/core/registry.py
andy-slac: Not checking `futureDatasetTypes` can be added as an option indeed. 

For preflight we still need to associate units with DatasetType names, and the query that we run also uses DatasetType  names to find existing input Datasets, so we need some DatasetType information there, maybe just a  name.

python/lsst/daf/butler/registries/sqlPreFlight.py
andy-slac: I just read this as "table shave" 😃

python/lsst/daf/butler/registries/sqlPreFlight.py
andy-slac: Common Table Expression - special SQL syntax. I'll update docstring.

python/lsst/daf/butler/registries/sqlPreFlight.py
andy-slac: 👍 to DatasetOriginInfo and re-factoring on separate ticket.

python/lsst/daf/butler/core/preFlight.py
TallJimbo: If `DatasetType` instances are more convenient for you, that's fine.  I hadn't realized they were usable as `dict` keys, but it looks like they are.

python/lsst/daf/butler/core/registry.py
TallJimbo: Sounds like it would make the most sense to have another argument that just adds more explicit DataUnits to the query, while leaving this argument as it is.

andy-slac: OK, and I'll leave that for the future tickets :slightly_smiling_face:

PR #79, opened 2018-08-24
DM-15533: Python inheritance for StorageClass

timj:
@TallJimbo here's a proof of concept for StorageClass inheritance. I think if I add support for overriding of components (so ExposureF.components.image can override Exposure.components.image -- could check the child `ImageF` is-a `Image` for example) then storage class definitions suddenly become much much simpler since all the Exposure F/I duplication disappears with the only differences being the actual F and I.

### DIFF
--- a/python/lsst/daf/butler/core/storageClass.py
+++ b/python/lsst/daf/butler/core/storageClass.py
@@ -54,10 +54,22 @@ class StorageClass:
         Fully qualified name of class supporting assembly and disassembly
         of a `pytype` instance.
     """
+    _cls_name = "BaseStorageClass"
+    _cls_components = None
+    _cls_assembler = None
+    _cls_pytype = None
     defaultAssembler = CompositeAssembler
     defaultAssemblerName = getFullTypeName(defaultAssembler)
 
-    def __init__(self, name, pytype=None, components=None, assembler=None):
+    def __init__(self, name=None, pytype=None, components=None, assembler=None):
+        if name is None:
+            name = self._cls_name
+        if pytype is None:
+            pytype = self._cls_pytype
+        if components is None:
+            components = self._cls_components
+        if assembler is None:
+            assembler = self._cls_assembler
         self.name = name
         self._pytypeName = pytype
         if pytype is None:
@@ -146,7 +158,7 @@ class StorageClass:
         if self.name != other.name:
             return False
 
-        if type(self) != type(other):
+        if not isinstance(other, StorageClass):
             return False
 
         # We must compare pytype and assembler by name since we do not want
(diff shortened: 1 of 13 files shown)


python/lsst/daf/butler/core/storageClass.py
```diff
@@ -244,17 +257,22 @@ def addFromConfig(self, config):
...
+        # Children are assumed not to be themselves specifying different
+        # components for now. Ideally this should be recursive.
         composites = {}
         simple = {}
+        children = {}
         for name, info in sconfig.items():
             if "components" in info:
                 composites[name] = info
+            elif "parents" in info:
+                children[name] = info
```
TallJimbo: What do you think about calling this `inherits` instead of `parents` to avoid confusion between compositions and inheritance?

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -244,17 +257,22 @@ def addFromConfig(self, config):
...
             else:
                 simple[name] = info
 
-        for name in itertools.chain(simple, composites):
+        for name in itertools.chain(simple, composites, children):
```
TallJimbo: And maybe `derived` instead of `children`?

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -272,9 +290,38 @@ def addFromConfig(self, config):
...
 
             # Create the new storage class and register it
-            newStorageClass = StorageClass(name, **storageClassKwargs)
+            baseClass = StorageClass
+            if "parent" in info:
+                baseClass = type(self.getStorageClass(info["parent"]))
+
+            newStorageClassType = self.makeNewStorageClass(name, baseClass, **storageClassKwargs)
+            newStorageClass = newStorageClassType()
+            print("Registering {}".format(newStorageClass))
```
TallJimbo: Looks like it should be removed or made a debug log.

TallJimbo (approved):
Looks good; just some quibbles about naming.
Do you want to remove the duplication in `storageClasses.yaml` on the same ticket?

TallJimbo:
Oh, and if it wasn't clear from the :tada:, I'm definitely in favor of making components overridable after inheritance.

timj:
Yes, I'll do the YAML clean up as well. This was a demonstration PR but I'll fix it up and sort out components merging as well.

python/lsst/daf/butler/core/storageClass.py
timj: Ok, although `inheritsFrom` sounds better to me.

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -272,9 +289,56 @@ def addFromConfig(self, config):
...
+    @staticmethod
+    def makeNewStorageClass(name, baseClass, **kwargs):
+        """Create a new Python class as a subclass of `StorageClass`.
+
+        Parameters
+        ----------
+        name : `str`
+            Name to use for this class.
+        baseClass : `type`
+            Base class for this `StorageClass`.
```
TallJimbo: Should this argument have a default, or is this method never used for StorageClasses that do not have a base class?

TallJimbo (approved):
Second (and presumably final) pass; looks good, just one more tiny comment.

python/lsst/daf/butler/core/storageClass.py
timj: I'll add a default. It's questionable whether this should be a public method at all.

PR #80, opened 2018-08-28
DM-15561: Allow templates to be retrieved by DatasetRef and other cleanups

timj:
* Use DatasetRef, DatasetType or StorageClass to retrieve a file template.
* Use same "give me names" code for formatter retrieval.
* Remove storageClass and datasetType distinction from composites configruation.

### DIFF
--- a/config/composites.yaml
+++ b/config/composites.yaml
@@ -7,7 +7,7 @@ composites:
   # Use a dict rather than a list to allow easy merging of user config
   # into default config. Config class has no way of merging lists.
   # Types can be StorageClass names or DatasetType names
-  names:
+  disassembled:
     ExposureI: False
     ExposureF: False
     calexp: False
--- a/config/composites.yaml
+++ b/config/composites.yaml
@@ -4,12 +4,10 @@ composites:
   # default applies to all composites that are not explicitly mentioned
   # elsewhere.
   default: False
-  storageClasses:
-    # Use a dict rather than a list to allow easy merging of user config
-    # into default config. Config class has no way of merging lists.
+  # Use a dict rather than a list to allow easy merging of user config
+  # into default config. Config class has no way of merging lists.
+  # Types can be StorageClass names or DatasetType names
+  names:
     ExposureI: False
     ExposureF: False
-  datasetTypes:
-    # Same definition as for storage classes, but dataset type explicitly
-    # overrides a storage class definition.
     calexp: False
--- a/config/composites.yaml
+++ b/config/composites.yaml
@@ -6,7 +6,8 @@ composites:
   default: False
   # Use a dict rather than a list to allow easy merging of user config
   # into default config. Config class has no way of merging lists.
-  # Types can be StorageClass names or DatasetType names
+  # Types can be StorageClass names or DatasetType names, with DatasetType
+  # name always taking preference over StorageClass name.
   disassembled:
     ExposureI: False
     ExposureF: False
--- a/python/lsst/daf/butler/core/composites.py
+++ b/python/lsst/daf/butler/core/composites.py
@@ -26,7 +26,7 @@ __all__ = ("CompositesConfig", "CompositesMap")
 import logging
 
 from .config import ConfigSubset
-from .datasets import DatasetType
+from .datasets import DatasetType, DatasetRef
 from .storageClass import StorageClass
 
 log = logging.getLogger(__name__)
@@ -34,17 +34,16 @@ log = logging.getLogger(__name__)
 
 class CompositesConfig(ConfigSubset):
     component = "composites"
-    requiredKeys = ("default", "storageClasses")
+    requiredKeys = ("default", "names")
     defaultConfigFile = "composites.yaml"
 
     def validate(self):
         """Validate entries have the correct type."""
         super().validate()
-        for n in ("storageClasses", "datasetTypes"):
-            for k in self[n]:
-                key = f"{n}.{k}"
-                if not isinstance(self[key], bool):
-                    raise ValueError(f"CompositesConfig: Key {key} is not a Boolean")
+        for k in self["names"]:
+            key = f"names.{k}"
+            if not isinstance(self[key], bool):
+                raise ValueError(f"CompositesConfig: Key {key} is not a Boolean")
 
 
 class CompositesMap:
@@ -69,7 +68,7 @@ class CompositesMap:
 
         Parameters
         ----------
-        entity : `StorageClass` or `DatasetType`
+        entity : `StorageClass` or `DatasetType` or `DatasetRef`
             Thing to test against the configuration. The ``name`` property
             is used to determine a match.  A `DatasetType` will first check
             its name, before checking its `StorageClass`.  If there are no
@@ -84,6 +83,8 @@ class CompositesMap:
         components = None
         datasetTypeName = None
         storageClassName = None
+        if isinstance(entity, DatasetRef):
+            entity = entity.datasetType
         if isinstance(entity, DatasetType):
             datasetTypeName = entity.name
             storageClassName = entity.storageClass.name
@@ -102,12 +103,11 @@ class CompositesMap:
         matchName = "{} (via default)".format(entity)
         disassemble = self.config["default"]
 
-        if datasetTypeName is not None and datasetTypeName in self.config["datasetTypes"]:
-            disassemble = self.config[f"datasetTypes.{datasetTypeName}"]
-            matchName = datasetTypeName
-        elif storageClassName is not None and storageClassName in self.config["storageClasses"]:
-            disassemble = self.config[f"storageClasses.{storageClassName}"]
-            matchName = storageClassName
+        for name in (datasetTypeName, storageClassName):
+            if name is not None and name in self.config["names"]:
+                disassemble = self.config[f"names.{name}"]
+                matchName = name
+                break
 
         log.debug("%s will%s be disassembled", matchName, "" if disassemble else " not")
         return disassemble
--- a/python/lsst/daf/butler/core/composites.py
+++ b/python/lsst/daf/butler/core/composites.py
@@ -95,7 +95,7 @@ class CompositesMap:
         matchName = "{} (via default)".format(entity)
         disassemble = self.config["default"]
 
-        for name in (entity.lookupNames()):
+        for name in (entity._lookupNames()):
             if name is not None and name in self.config["disassembled"]:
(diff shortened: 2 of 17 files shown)


config/composites.yaml
```diff
@@ -4,14 +4,12 @@ composites:
   # default applies to all composites that are not explicitly mentioned
   # elsewhere.
   default: False
-  storageClasses:
-    # Use a dict rather than a list to allow easy merging of user config
-    # into default config. Config class has no way of merging lists.
+  # Use a dict rather than a list to allow easy merging of user config
+  # into default config. Config class has no way of merging lists.
+  # Types can be StorageClass names or DatasetType names
```
pschella: Should this be explicit, with one overriding the other?

python/lsst/daf/butler/core/composites.py
```diff
@@ -26,25 +26,24 @@
...
     defaultConfigFile = "composites.yaml"
 
     def validate(self):
         """Validate entries have the correct type."""
         super().validate()
-        for n in ("storageClasses", "datasetTypes"):
-            for k in self[n]:
-                key = f"{n}.{k}"
-                if not isinstance(self[key], bool):
-                    raise ValueError(f"CompositesConfig: Key {key} is not a Boolean")
```
pschella, line 47: Ah, I see it was explicit. But we don't want this anymore?

python/lsst/daf/butler/core/composites.py
```diff
@@ -102,12 +103,11 @@ def doDisassembly(self, entity):
...
-            disassemble = self.config[f"datasetTypes.{datasetTypeName}"]
-            matchName = datasetTypeName
-        elif storageClassName is not None and storageClassName in self.config["storageClasses"]:
-            disassemble = self.config[f"storageClasses.{storageClassName}"]
-            matchName = storageClassName
+        for name in (datasetTypeName, storageClassName):
+            if name is not None and name in self.config["names"]:
+                disassemble = self.config[f"names.{name}"]
+                matchName = name
+                break
```
pschella: So we better hope that there is never a clash between a storageClassName and a datasetTypeName (the latter being something that we can't control and rely on users not to pick incorrectly)?

python/lsst/daf/butler/core/config.py
```diff
@@ -220,7 +220,12 @@ def __initFromYaml(self, stream):
 
     def __getitem__(self, name):
         data = self.data
-        for key in name.split("."):
+        # Override the split for the simple case
+        if name in data:
+            keys = (name,)
```
pschella, line 225: Space after comma?

python/lsst/daf/butler/core/datasets.py
```diff
@@ -148,6 +148,18 @@ def componentTypeName(self, component):
...
+        """Names to use when looking up this datasetType in a configuration.
+
+        The names are returned in order of priority.
+
+        Returns
+        -------
+        names : `tuple` of `str`
+            Tuple of the `DatasetType` name and the `StorageClass` name.
+        """
+        return (self.name, *self.storageClass.lookupNames())
```
pschella: We have multiple lookup names for storageclasses now? Is this to deal with `ImageI`, `ImageF`, etc?

python/lsst/daf/butler/core/datasets.py
```diff
@@ -269,3 +281,15 @@ def detach(self):
...
+        """Names to use when looking up this DatasetRef in a configuration.
+
+        The names are returned in order of priority.
+
+        Returns
+        -------
+        names : `tuple` of `str`
+            Tuple of the `DatasetType` name and the `StorageClass` name.
+        """
+        return self.datasetType.lookupNames()
```
pschella: Why are there multiple names? Why is there a lookup in configuration for a specific `DatasetRef`? If the intent is to lookup something about the `DatasetType` in the configuration i'd rather explicitly have the user do `datasetRef.datasetType.lookupNames()` such that the intention is clear.

python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -79,23 +86,28 @@ def getTemplate(self, datasetTypeName):
         KeyError
             No template could be located for this Dataset type.
         """
+
+        # Get the names to use for lookup
+        names = entity.lookupNames()
```
pschella: I'm still on the fence about this. On the one hand special casing this for `DatasetRef` would be sad, but on the other hand I don't like having `lookupNames` on `datasetRef` if it isn't actually `DatasetRef` specific. Maybe I'm just not liking the name `lookupNames`? Perhaps `configTypeNames`/`typeNamesInConfig`/`typeNames` or something would be better?

python/lsst/daf/butler/core/composites.py
timj: It adds a lot of extra faff in all the routines that look up datasetType and storageClass in config files. They have to always look in two locations. That's not how we did it for formatters (we had a single block of configuration). If we really want to be explicit I should change formatters and templates to be explicit and keep composites as it is. I don't think we ever expect there to be clashes between storage class names and dataset type names. We could ensure that that is true by insisting that all storage class names start with an upper case and all dataset type names start with lower case...

python/lsst/daf/butler/core/composites.py
timj: Yes. We have to decide. At the moment we decided that they can't clash. If we want to ensure they can't clash we need to decide which of the options we want to use (as defined in my previous comment: separate sections in every config, or insist on naming convention).

python/lsst/daf/butler/core/config.py
timj: flake8 did not complain. I think flake8 doesn't like lists without a space at the end but tuples are fine.

python/lsst/daf/butler/core/datasets.py
timj: No, but to retain compatibility with the `lookupNames()` interface it returns a tuple with a single element in it that needs to be unpacked.

python/lsst/daf/butler/core/datasets.py
timj: Doing it like this means I have a single interface for DatasetType, DatasetRef and StorageClass so I call one method on whatever I've been given and I can look it up in the config. Without this I have to have an explicit `isinstance` check in those cases and forward it on myself. Since we want all these look ups to work with all three types it seemed easier and clearer to do it this way. Whilst looking at this I also realized that in theory we can have a disagreement between `DatasetRef.components` and `DatasetRef.datasetType.storageClass.components` and we never check.

python/lsst/daf/butler/core/fileTemplates.py
timj: Correct. It's allowing us to take an object with a compatible interface and let us work out which entry in the config object is relevant. I have no problem with changing the name of this method.

config/composites.yaml
TallJimbo: Per-DatasetType configurations should always override per-StorageClass configurations.

python/lsst/daf/butler/core/composites.py
```diff
@@ -69,7 +66,7 @@ def doDisassembly(self, entity):
 
         Parameters
         ----------
-        entity : `StorageClass` or `DatasetType`
+        entity : `StorageClass` or `DatasetType` or `DatasetRef`
```
TallJimbo, line 69: (not actually this line) I'd be more comfortable just calling this method `isComposite` instead of `doDisassembly`, too - it matches the new and better precedent, and `doDisassembly` to me sounds like something that is *doing* disassembly.

python/lsst/daf/butler/core/datasets.py
TallJimbo: Whoa, I had no idea you could use argument unpacking in a regular tuple definition.

python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -61,13 +61,20 @@ def __init__(self, config, default=None):
...
+
         Parameters
         ----------
-        datasetTypeName : `str`
-            Dataset type name.
+        entity : `DatasetType`, `DatasetRef`, or `StorageClass`
+            Instance to use to look for a corresponding template.
+            A `DatasetType` name or a `StorageClass` name will be used
+            depending on the supplied entity. Priority is given to a
+            `DatasetType` name.
```
TallJimbo, line 77: Just a thought for the future: template lookup by StorageClass is probably less useful in practice than template lookup by "set of DataUnits".  In other words, I probably want my templates for Tract+Patch Datasets to be more similar than my templates for all SourceCatalog Datasets.  But the existing support for optional replacement tokens in templates will hopefully make that a non-issue.

python/lsst/daf/butler/core/fileTemplates.py
TallJimbo: Maybe make it private, since it is an implementation detail of the Butler framework and the classes that use it are effectively "friends"?

tests/config/basic/composites-bad.yaml
```diff
@@ -1,11 +1,10 @@
 composites:
-  storageClasses:
+  names:
```
TallJimbo: What do you think about just dropping this `names` level and letting "default" be a specially-recognized key at the next level?  It just feels like it's a superfluous node in the vast majority of config trees, which will inherit the default from somewhere.  And anyone who wants to name their DatasetType "default" deserves what they get :slightly_smiling_face: .

TallJimbo (approved):

python/lsst/daf/butler/core/composites.py
timj: It's not really saying "is this a composite" it's saying "is this a composite that should be disassembled", so they aren't the same thing. How about `shouldBeDisassembled` ?

tests/config/basic/composites-bad.yaml
timj: I did think about that, but then I worried that it would make it very difficult to add any other configuration items to that config in the future.

timj: How about changing `names` to `disassembly` (then at least the True/False relates to the name)?

TallJimbo: `disassembled`?

python/lsst/daf/butler/core/composites.py
timj: Oops. I forgot to change the name to `shouldBeDisassembled`. I'll sneak it in to another ticket. Sorry.

PR #81, opened 2018-08-29
DM-10302 Change flux->instFlux

parejkoj:


tests/datasetsHelper.py
```diff
@@ -37,8 +37,7 @@ def assertCatalogEqual(self, inputCatalog, outputCatalog):
         inputRecord = inputCatalog[0]
         outputTable = outputCatalog.getTable()
         outputRecord = outputCatalog[0]
-        self.assertEqual(inputTable.getPsfFluxDefinition(), outputTable.getPsfFluxDefinition())
```
r-owen, line 40: Why did you remove this instead of renaming it?

r-owen (approved):
Probably fine, but I did have one question

tests/datasetsHelper.py
parejkoj: There's nothing to rename it to: the `*Definition()` methods went away entirely, and the line immediately following tests that the data is there. Do you have a suggestion for a different test?

PR #82, opened 2018-08-29
DM-15459: make getRegionHolder's inputs more flexible.

TallJimbo:


python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -69,6 +69,18 @@ def __init__(self, name, requiredDependencies, optionalDependencies,
     def __repr__(self):
         return "DataUnit({})".format(self.name)
 
+    def __eq__(self, other):
+        try:
+            return self.name == other.name
+        except AttributeError:
+            return NotImplemented
+
+    def __ne__(self, other):
```
natelust: you should not need to implement this, as the python interpreter will do this automatically if __eq__ exists

natelust (approved):

PR #84, opened 2018-08-31
DM-15614: Allow dynamic hierarchy delimiters for configs

timj:
The delimiter for hierarchies no longer defaults to a `.`. The delimiter is now a class variable and can be overridden by starting the string with the delimiter you wish to use.

### DIFF
--- a/python/lsst/daf/butler/core/config.py
+++ b/python/lsst/daf/butler/core/config.py
@@ -175,7 +175,7 @@ class Config(collections.UserDict):
         return pprint.pformat(self.data, indent=2, width=1)
 
     def __repr__(self):
-        return self.data.__repr__()
+        return f"{type(self).__name__}({self.data!r})"
 
     def __initFromFile(self, path):
         """Load a file from path.
--- a/python/lsst/daf/butler/core/config.py
+++ b/python/lsst/daf/butler/core/config.py
@@ -425,7 +425,7 @@ class Config(collections.UserDict):
         getKeysAsTuples(self.data, keys, None)
         return keys
 
-    def names(self, topLevelOnly=False):
+    def names(self, topLevelOnly=False, delimiter=None):
         """Get the delimited name of all the keys in the hierarchy.
 
         The values returned from this method are guaranteed to be usable
@@ -436,6 +436,10 @@ class Config(collections.UserDict):
         topLevelOnly : `bool`, optional
             If False, the default, a full hierarchy of names is returned.
             If True, only the top level are returned.
+        delimiter : `str`, optional
+            Delimiter to use when forming the keys.  The delimiter must
+            not be present in any of the keys.  If `None` given the delimiter
+            will be automatically provided.
 
         Returns
         -------
@@ -446,6 +450,12 @@ class Config(collections.UserDict):
         -----
         This is different than the built-in method `dict.keys`, which will
         return only the first level keys.
+
+        Raises
+        ------
+        ValueError
+            The externally specified delimiter is unsuitable because a key
+            clashes with the value.
         """
         if topLevelOnly:
             return list(self.keys())
@@ -453,7 +463,39 @@ class Config(collections.UserDict):
         # Get all the tuples of hierarchical keys
         nameTuples = self.nameTuples()
 
-        delimiter = self._D
+        # Form big string for easy check of delimiter clash
+        combined = "".join("".join(str(s) for s in k) for k in nameTuples)
+
+        if delimiter is not None:
+            if delimiter in combined:
+                # For better error message we need to report the clashing key
+                def findMatch(names, searchStr):
+                    for n in names:
+                        for s in n:
+                            if searchStr in s:
+                                return s
+                    return None
+
+                raise ValueError(f"Specified delimiter, {delimiter!r} can not be used.  "
+                                 f"Clashes with key {findMatch(nameTuples, delimiter)!r}")
+        else:
+            # Start with something
+            delimiter = self._D
+            ntries = 0
+            while delimiter in combined:
+                log.debug(f"Trying delimiter '{delimiter}'")
+                ntries += 1
+
+                if ntries > 100:
+                    raise ValueError(f"Unable to determine a delimiter for Config {self}")
+
+                # try another one
+                while True:
+                    delimiter = chr(ord(delimiter)+1)
+                    if not delimiter.isalnum():
+                        break
+
+        log.debug(f"Using delimiter {delimiter!r}")
         strings = [delimiter + delimiter.join(str(s) for s in k) for k in nameTuples]
         return strings
 
--- a/python/lsst/daf/butler/core/config.py
+++ b/python/lsst/daf/butler/core/config.py
@@ -255,8 +255,8 @@ class Config(collections.UserDict):
             if escaped in key:
                 # Replace with a character that won't be in the string
                 temp = "\r"
-                if temp in key or d == "\r":
-                    raise ValueError("Can not use carriage return character in hierarchical key or as"
+                if temp in key or d == temp:
+                    raise ValueError(f"Can not use character {temp!r} in hierarchical key or as"
                                      " delimiter if escaping the delimiter")
                 key = key.replace(escaped, temp)
             hierarchy = key.split(d)
--- a/python/lsst/daf/butler/core/config.py
+++ b/python/lsst/daf/butler/core/config.py
@@ -637,7 +637,7 @@ class ConfigSubset(Config):
         # component.component (since the included files can themselves
         # include the component name)
         if self.component is not None:
-            doubled = "{0}{1}{0}".format(self.component, self.D)
+            doubled = (self.component, ) * 2
             # Must check for double depth first
             if doubled in externalConfig:
                 externalConfig = externalConfig[doubled]
@@ -701,9 +701,9 @@ class ConfigSubset(Config):
 
         if mergeDefaults and containerKey is not None and containerKey in self:
             for idx, subConfig in enumerate(self[containerKey]):
-                self[f"{containerKey}{self.D}{idx}"] = type(self)(other=subConfig, validate=validate,
-                                                                  mergeDefaults=mergeDefaults,
-                                                                  searchPaths=searchPaths)
+                self[(containerKey, idx)] = type(self)(other=subConfig, validate=validate,
+                                                       mergeDefaults=mergeDefaults,
+                                                       searchPaths=searchPaths)
 
         if validate:
             self.validate()
--- a/python/lsst/daf/butler/core/config.py
+++ b/python/lsst/daf/butler/core/config.py
@@ -441,9 +441,9 @@ class Config(collections.UserDict):
             If False, the default, a full hierarchy of names is returned.
             If True, only the top level are returned.
         delimiter : `str`, optional
-            Delimiter to use when forming the keys.  The delimiter must
-            not be present in any of the keys.  If `None` given a delimiter
-            will be automatically provided.
+            Delimiter to use when forming the keys.  If the delimiter is
+            present in any of the keys, it will be escaped in the returned
+            names.  If `None` given a delimiter will be automatically provided.
 
         Returns
         -------
@@ -454,12 +454,6 @@ class Config(collections.UserDict):
         -----
         This is different than the built-in method `dict.keys`, which will
         return only the first level keys.
-
-        Raises
-        ------
-        ValueError
-            The externally specified delimiter is unsuitable because a key
-            clashes with the value.
         """
         if topLevelOnly:
             return list(self.keys())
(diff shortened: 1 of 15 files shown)


python/lsst/daf/butler/core/config.py
```diff
@@ -109,18 +96,32 @@ def extractFile(self, filename):
...
 
-class Config(_ConfigBase):
+class Config(collections.UserDict):
     """Implements a datatype that is used by `Butler` for configuration
     parameters.
 
     It is essentially a `dict` with key/value pairs, including nested dicts
     (as values). In fact, it can be initialized with a `dict`. The only caveat
-    is that keys may **not** contain dots (``.``). This is explained next:
+    is that keys may **not** contain the delimiter. This is explained next:
```
ktlim: I think this is backwards.  You intend to allow any keys whatsoever; it's the accessor who must ensure that the delimiter does not appear in any keys.  Instead, this reads like an admonishment to the key creator not to use an unknown delimiter.

python/lsst/daf/butler/core/config.py
```diff
@@ -109,18 +96,32 @@ def extractFile(self, filename):
...
+    will use ``.`` as a delimiter, but ``foo[":a.b.c"]`` will use ``:``
+    as the delimeter resulting in a single key of ``a.b.c`` being accessed.
+    Using ``foo[":a:b:c"]`` is therefore equivalent to ``foo[".a.b.c"]``.
+    This requires that keys in `Config` instances do not themselves start with
+    non-alphanumeric characters.  If the hierarchy is already available in a
+    `list` or `tuple` it can be provided directly without forming it into a
+    string, such that ``foo["a", "b", "c"]`` is equivalent to
+    ``foo[".a.b.c"]``.  Finally, the delimiter can be escaped if it should
+    not be used in part of the string: ``foo[r".a.b\.c"]`` results in a two
+    element hierarchy of ``a`` and ``b.c``.  For hard-coded strings it is
```
ktlim: Can you escape an initial non-alphanumeric character? Looking at the code, the answer is no. Maybe you should be able to, or else you should make clear that you can't.

python/lsst/daf/butler/core/config.py
```diff
@@ -109,18 +96,32 @@ def extractFile(self, filename):
...
+    is picked up from the first character in that string. For example,
+    ``foo.getValue(".a.b.c")`` is the same as ``foo["a"]["b"]["c"]`` is the
+    same as ``foo[".a.b.c"]``, and either of these syntaxes may be used.
+    If the first character is alphanumeric, no delimiter will be used.
+    ``foo["a.b.c"]`` will be a single key ``a.b.c` whereas ``foo[".a.b.c"]``
+    will use ``.`` as a delimiter, but ``foo[":a.b.c"]`` will use ``:``
+    as the delimeter resulting in a single key of ``a.b.c`` being accessed.
+    Using ``foo[":a:b:c"]`` is therefore equivalent to ``foo[".a.b.c"]``.
+    This requires that keys in `Config` instances do not themselves start with
+    non-alphanumeric characters.  If the hierarchy is already available in a
```
ktlim: You can access non-alphanumeric-starting keys by selecting a different delimiter: `":.a.b.c"` should work, no?

python/lsst/daf/butler/core/config.py
```diff
@@ -140,6 +141,9 @@ class Config(_ConfigBase):
         If `None` is provided an empty `Config` will be created.
     """
 
+    _D = "→"
+    """Default internal delimiter to use for components in the hierarchy"""
```
ktlim: Should say when this default is used.

python/lsst/daf/butler/core/config.py
```diff
@@ -218,13 +225,57 @@ def __initFromYaml(self, stream):
...
+            Hierarchical keys as a `list`.
+        """
+        if isinstance(key, str):
+            if not key[0].isalnum():
+                d = key[0]
+                key = key[1:]
+            else:
+                return [key, ]
+            # Allow escaping of the delimiter: .a.foo\.bar -> a foo.bar
+            escaped = f"\\{d}"
```
ktlim: Maybe `fr"\{d}"`? Or that may be too fancy. What you have is fine.

python/lsst/daf/butler/core/config.py
```diff
@@ -218,13 +225,57 @@ def __initFromYaml(self, stream):
...
+            temp = None
+            if escaped in key:
+                # Replace with a character that won't be in the string
+                temp = "\r"
```
ktlim: This is hacky.  Perhaps you should just use `re.split(fr'(?<=[^\\])\{d}', key)`? (The lookbehind should be safe because there should always be at least one non-delimiter character, but you should probably check for that, too: you wouldn't want to allow a key of `"::a:b\:c"`.)

python/lsst/daf/butler/core/config.py
```diff
@@ -218,13 +225,57 @@ def __initFromYaml(self, stream):
...
+
     def __getitem__(self, name):
         data = self.data
-        # Override the split for the simple case
+        # Override the split for the simple case where there is an exact
+        # match.  This allows `Config.items()` to work since `UserDict`
+        # accesses Config.data directly to obtain the keys and every top
+        # level key should always retrieve the top level values.
         if name in data:
             keys = (name,)
```
ktlim, line 282: Seems kind of wasteful to do another `in` and not just return the value here, but I guess you don't want to repeat the `Config` and `_D` code at the end.

python/lsst/daf/butler/core/config.py
```diff
@@ -238,10 +289,13 @@ def __getitem__(self, name):
...
+            if self._D != Config._D:
+                data._D = self._D
         return data
 
     def __setitem__(self, name, value):
-        keys = name.split(".")
+        keys = self._splitKeys(name)
         last = keys.pop()
         if isinstance(value, Config):
             value = copy.deepcopy(value.data)
```
ktlim: Below this, I'm a little worried about the `setdefault(key, {})`.  What if there's supposed to be a sequence in there?  You'd start getting `"3"` keys instead of `3` indexes.

python/lsst/daf/butler/core/config.py
```diff
@@ -349,32 +406,102 @@ def names(self, topLevelOnly=False):
...
-        def getKeys(d, keys, base):
+        def getKeysAsTuples(d, keys, base):
             if isinstance(d, collections.Sequence):
                 theseKeys = range(len(d))
             else:
                 theseKeys = d.keys()
             for key in theseKeys:
                 val = d[key]
-                levelKey = f"{base}.{key}" if base is not None else key
+                levelKey = base + (key,) if base is not None else (key,)
```
ktlim, line 429: These tuples may include integers, not just strings.  Do you need to stringify the numbers above?

ktlim (commented):
Only got through `nameTuples()` for now.  Still need to look at `names()` and `test_config.py`.

python/lsst/daf/butler/core/config.py
ktlim: Oh, but I just realized you need to replace the quoted delimiters with the actual delimiter afterwards, which my `re.split` doesn't do.  Sorry about that.

python/lsst/daf/butler/core/config.py
```diff
@@ -349,32 +406,102 @@ def names(self, topLevelOnly=False):
...
                 if isinstance(val, (collections.Mapping, collections.Sequence)) and not isinstance(val, str):
-                    getKeys(val, keys, levelKey)
+                    getKeysAsTuples(val, keys, levelKey)
         keys = []
-        getKeys(self.data, keys, None)
+        getKeysAsTuples(self.data, keys, None)
         return keys
 
+    def names(self, topLevelOnly=False, delimiter=None):
+        """Get the delimited name of all the keys in the hierarchy.
```
ktlim: Maybe "a delimited name" instead of "the delimited name" since there isn't one canonical name.

python/lsst/daf/butler/core/config.py
```diff
@@ -349,32 +406,102 @@ def names(self, topLevelOnly=False):
...
+        to access items in the configuration object.
+
+        Parameters
+        ----------
+        topLevelOnly : `bool`, optional
+            If False, the default, a full hierarchy of names is returned.
+            If True, only the top level are returned.
+        delimiter : `str`, optional
+            Delimiter to use when forming the keys.  The delimiter must
+            not be present in any of the keys.  If `None` given the delimiter
```
ktlim: Again, "a delimiter", not "the delimiter".

python/lsst/daf/butler/core/config.py
```diff
@@ -349,32 +406,102 @@ def names(self, topLevelOnly=False):
...
+            will be automatically provided.
+
+        Returns
+        -------
+        names : `list`
```
ktlim: `list of str`?

python/lsst/daf/butler/core/config.py
```diff
@@ -349,32 +406,102 @@ def names(self, topLevelOnly=False):
...
+
+        # Get all the tuples of hierarchical keys
+        nameTuples = self.nameTuples()
+
+        # Form big string for easy check of delimiter clash
+        combined = "".join("".join(str(s) for s in k) for k in nameTuples)
+
+        if delimiter is not None:
+            if delimiter in combined:
+                # For better error message we need to report the clashing key
```
ktlim: I was hoping the backslash-quoting would allow you to use the provided delimiter even in this case, by quoting it.

python/lsst/daf/butler/core/config.py
```diff
@@ -349,32 +406,102 @@ def names(self, topLevelOnly=False):
...
+                def findMatch(names, searchStr):
+                    for n in names:
+                        for s in n:
+                            if searchStr in s:
+                                return s
```
ktlim: Better to return `s` or `n`?

python/lsst/daf/butler/core/config.py
```diff
@@ -349,32 +406,102 @@ def names(self, topLevelOnly=False):
...
+                    return None
+
+                raise ValueError(f"Specified delimiter, {delimiter!r} can not be used.  "
+                                 f"Clashes with key {findMatch(nameTuples, delimiter)!r}")
+        else:
+            # Start with something
+            delimiter = self._D
+            ntries = 0
+            while delimiter in combined:
+                log.debug(f"Trying delimiter '{delimiter}'")
```
ktlim: The message here isn't quite accurate; `delimiter` was already tried and found wanting.

python/lsst/daf/butler/core/config.py
```diff
@@ -586,7 +713,7 @@ def __init__(self, other=None, validate=True, mergeDefaults=True, searchPaths=No
         # component.component (since the included files can themselves
         # include the component name)
         if self.component is not None:
-            doubled = "{0}.{0}".format(self.component)
+            doubled = (self.component, ) * 2
```
ktlim: Less tricky to say `(self.component, self.component)` ?

ktlim (commented):
Still have to check the tests.

tests/test_config.py
```diff
@@ -98,30 +99,53 @@ def testBadConfig(self):
             with self.assertRaises(RuntimeError):
                 Config(badArg)
 
+    def testBasics(self):
+        c = Config({1: 2, 3: 4, "key3": 6})
```
ktlim: These tests seem pretty minimal; can't you assert the actual strings or at least more of them or perhaps do a round-trip of the `repr`?

tests/test_config.py
```diff
@@ -175,23 +204,64 @@ def testHierarchy(self):
...
+        # Try delimited names and tuples
+        for n in itertools.chain(c.names(), c.nameTuples()):
             val = c[n]
             self.assertIsNotNone(val)
             self.assertIn(n, c)
 
+        names = c.names()
+        nameTuples = c.nameTuples()
+        self.assertEqual(len(names), len(nameTuples))
+        self.assertGreater(len(names), 5)
```
ktlim: Why `assertGreater` here?  Don't you know what the length should be?

tests/test_config.py
```diff
@@ -175,23 +204,64 @@ def testHierarchy(self):
...
+        self.assertGreater(len(nameTuples), 5)
+
+        with self.assertRaises(ValueError):
```
ktlim: This would go away if we can do the quoted-delimiter thing.

python/lsst/daf/butler/core/config.py
timj: I don't think I want to support escaping of the delimiter at the start of the key.  We can revisit that if it becomes an issue (I'm not convinced people will want to be escaping delimiters at all).

python/lsst/daf/butler/core/config.py
timj: Yes, it only goes once around the loop anyhow.

python/lsst/daf/butler/core/config.py
timj: `config[".a.b.1"] = 5` works so long as `/a/b` is a sequence.  Adding support for sequences is fairly new and has some rough edges. You can't say `config[".a.b.10.c"] = 4` if `/a/b/` is not a sequence or if `/a/b` is a sequence but only has 5 elements. If we started to want to support that there'd have to be some more deep thinking and we'd have to consider a ConfigSequence object I think. I'd like to defer going down that road until we need to.

python/lsst/daf/butler/core/config.py
timj: I believe there are tests for this and it's not an issue (`names()` does for stringification).

python/lsst/daf/butler/core/config.py
timj: Good point. I had meant to remove that debug message anyhow.

python/lsst/daf/butler/core/config.py
timj: Ah. Yes. You are right. I could escape the delimiter in the keys in this case.

python/lsst/daf/butler/core/config.py
timj: Good point. `n` is more explicit.

tests/test_config.py
timj: I'll do a round trip. I didn't really want to get bogged down in exact stringification form.

tests/test_config.py
timj: Fixed to use Equal.

PR #85, opened 2018-09-07
DM-15424: add limited Registry concept

TallJimbo:
Also includes some cleanups to existing code and configuration; recommend reviewing commit-by-commit.


python/lsst/daf/butler/core/registry.py
```diff
@@ -667,10 +697,13 @@ def findDataUnitEntry(self, dataUnitName, value):
         dataUnitEntry : `dict`
             Dictionary with all `DataUnit` values, or `None` if no matching
             entry is found.
+        NotImplementedError
```
timj: `Raises` section is missing.

python/lsst/daf/butler/core/dataUnit.py
```diff
@@ -264,12 +264,11 @@ def fromConfig(cls, config):
         Parameters
         ----------
         config : `SchemaConfig`
```
timj: Not a `SchemaConfig`

python/lsst/daf/butler/core/schema.py
```diff
@@ -75,17 +77,19 @@ class SchemaBuilder:
     metadata : `sqlalchemy.MetaData`
         The sqlalchemy schema description.
     tables : `dict`
-        All created tables.
-    views : `dict`
-        All created views.
+        A mapping from table or view name to the associated SQLAlchemy object.
```
timj, line 89: I hadn't really noticed before that Schema requires SQLAlchemy even if Registry is implemented in non-SQL manner.

python/lsst/daf/butler/core/registry.py
```diff
@@ -143,6 +158,12 @@ def __init__(self, registryConfig, schemaConfig=None, dataUnitConfig=None, creat
     def __str__(self):
         return "None"
 
+    @property
+    def limited(self):
+        """If True, this Registry does not maintain DataUnit metadata or
+        relationships (`bool`)."""
+        return self.config.get("limited", False)
```
timj, line 165: Now I see why you wanted `get()` to work...

python/lsst/daf/butler/core/registry.py
```diff
@@ -168,7 +189,12 @@ def transaction(self):
 
     @property
     def pixelization(self):
-        """Object that interprets SkyPix DataUnit values (`sphgeom.Pixelization`)."""
+        """Object that interprets SkyPix DataUnit values (`sphgeom.Pixelization`).
```
timj: I think this has to be `lsst.sphgeom.Pixelization`.

python/lsst/daf/butler/core/registry.py
```diff
@@ -168,7 +189,12 @@ def transaction(self):
...
+
+        ``None`` for limited registries.
```
timj: I think single back ticks work for `None` (not that anyone will really be clicking on it to find out what it means).

python/lsst/daf/butler/core/registry.py
```diff
@@ -625,6 +651,7 @@ def markInputUsed(self, quantum, ref):
         raise NotImplementedError("Must be implemented by subclass")
 
     @abstractmethod
+    @disableWhenLimited
     @transactional
```
timj, line 655: This is becoming quite funny now with all these stacked up.

python/lsst/daf/butler/core/registry.py
```diff
@@ -648,10 +675,13 @@ def addDataUnitEntry(self, dataUnitName, values):
         ValueError
             If an entry with the primary-key defined in `values` is already
             present.
+        NotImplementedError
+            If ``self.limited`` is ``True``.
```
timj: Single backticks for `True`.

python/lsst/daf/butler/core/registry.py
```diff
@@ -667,10 +697,16 @@ def findDataUnitEntry(self, dataUnitName, value):
         dataUnitEntry : `dict`
             Dictionary with all `DataUnit` values, or `None` if no matching
             entry is found.
+
+        Raises
+        ------
+        NotImplementedError
+            If ``self.limited`` is ``True``.
```
timj: There must be a way to make this link to the classes `limited` property in the documentation. @jonathansick ?

python/lsst/daf/butler/core/schema.py
```diff
@@ -44,6 +45,9 @@ class Schema:
     config : `SchemaConfig` or `str`, optional
         Load configuration. Defaults will be read if ``config`` is not
         a `SchemaConfig`.
+    limited : `bool`
+        If True, ignore tables, views, and associated foreign keys whose
```
timj: Sometimes `True` is back ticked, here it is not. Be consistent.

python/lsst/daf/butler/core/schema.py
```diff
@@ -19,6 +19,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+
```
timj: Did we need this new line?

python/lsst/daf/butler/core/schema.py
```diff
@@ -44,6 +45,9 @@ class Schema:
     config : `SchemaConfig` or `str`, optional
         Load configuration. Defaults will be read if ``config`` is not
         a `SchemaConfig`.
+    limited : `bool`
+        If True, ignore tables, views, and associated foreign keys whose
+        config descriptions include a 'limited' key set to `False`.
```
timj: I think the quote character is wrong here.

python/lsst/daf/butler/core/schema.py
```diff
@@ -72,6 +76,12 @@ def __init__(self, config=None):
 class SchemaBuilder:
     """Builds a Schema step-by-step.
 
+    Parameters
+    ----------
+    limited : `bool`
+        If True, ignore tables, views, and associated foreign keys whose
+        config descriptions include a 'limited' key set to `False`.
```
timj: Backtick the True (False already is) and fix quote char for `limited`.

timj (approved):
This looks okay to me. Very minor comments.

python/lsst/daf/butler/core/schema.py
TallJimbo: Yeah, very much so.  In fact, while doing some work on DM-15210 it occurred to me that @pschella may have based some of the structure of `schema.yaml` on what SQLAlchemy wants - I've always found our description of foreign keys hard to parse, for instance, but it turns out they're quite convenient to pass to SQLAlchemy.  Anyhow, this bothers me too, but I'm not sure what to do about it and this ticket definitely isn't the place for it.

python/lsst/daf/butler/core/schema.py
TallJimbo: Accident; removed.

python/lsst/daf/butler/core/schema.py
pschella: Schema is indeed quite close to SQLAlchemy, but it doesn't have to actually be required. Overall you need some way for `schema.tables['name']` to give you back something that makes sense to use as a table representation. But it doesn't have to be a SQLAlchemy.Table object per-se.

pschella: The reason why I deliberately stayed pretty close to a one-to-one mapping between the YAML description and SQLAlchemy's object representation was that I expected us to have to use more complicated schema properties at some later stage. I thus assumed SQLAlchemy did what it did for a reason and it would make our lives easier in matching its expressiveness by having a one-to-one mapping as opposed to something that is perhaps a bit easier to read for the simple cases but requires ugly hacks for the complex ones.  This expectation may since have been proven wrong of course ;)

python/lsst/daf/butler/core/registry.py
jonathansick: I'd try

```
If `limited` is `True`.
```

If that doesn't work, I'd do

```
If `RegistryConfig.limited` is `True`.
```

jonathansick: Also, the recommended way of writing the sentence is:

```
Raised if `RegistryConfig.limited` is `True`.
```

https://developer.lsst.io/python/numpydoc.html#py-docstring-raises

TallJimbo: Looks like it works if I just say
```
    If `limited` is `True`.
```

python/lsst/daf/butler/core/schema.py
jonathansick: It should **always** be back ticked in this context.

TallJimbo:
One more commit: a global replace to do single- instead of double-backticks for `True`, `False`, and `None`.

PR #86, opened 2018-09-10
Fix preflight spatial join (DM-15689)

andy-slac:
Trivial one-line fix to use new way to access table data via schema
tables dict. Also removes couple of methods from DataUnitJoin which do
not work.

Added a trivial non-realistic unit test to provide coverage for the code
which was broken.

### DIFF
--- a/tests/test_sqlPreFlight.py
+++ b/tests/test_sqlPreFlight.py
@@ -303,6 +303,40 @@ class SqlPreFlightTestCase(lsst.utils.tests.TestCase):
         rows = list(rows)
         self.assertEqual(len(rows), 0)
 
+    def testPreFlightSpatialMatch(self):
+        """Test involving spacial match using join tables.
+
+        Note that realistic test needs a resonably-defined SkyPix and regions
+        in registry tables which is hard to implement in this simple test.
+        So we do not actually fill registry with any data and all queries will
+        return empty result, but this is still useful for coverage of the code
+        that generates query.
+        """
+        registry = self.registry
+
+        # dataset types
+        registry.makeRun(collection="test")
+        storageClass = StorageClass("testDataset")
+        registry.storageClasses.registerStorageClass(storageClass)
+
+        calexpType = DatasetType(name="CALEXP", dataUnits=("Camera", "Visit", "Sensor"),
+                                 storageClass=storageClass)
+        registry.registerDatasetType(calexpType)
+
+        coaddType = DatasetType(name="deepCoadd_calexp", dataUnits=("SkyMap", "Tract", "Patch",
+                                                                    "AbstractFilter"),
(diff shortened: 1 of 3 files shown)


tests/test_sqlPreFlight.py
```diff
@@ -303,6 +303,40 @@ def testPreFlightSkyMapUnits(self):
...
+                                 storageClass=storageClass)
+        registry.registerDatasetType(calexpType)
+
+        coaddType = DatasetType(name="deepCoadd_calexp", dataUnits=("SkyMap", "Tract", "Patch",
+                                                                    "AbstractFilter"),
+                                storageClass=storageClass)
+        registry.registerDatasetType(coaddType)
+
+        # without data this should run OK but return empty set
+        originInfo = DatasetOriginInfoDef(defaultInputs=["test"], defaultOutput="")
```
natelust: What is the string "test" here, is is something defined elsewhere or some new name? I assume there is a dataset named test located somewhere as this is an input, but it is not clear. If there is a commonly shared dataset, should the name for it be factored out somewhere as a class variable?

tests/test_sqlPreFlight.py
```diff
@@ -303,6 +303,40 @@ def testPreFlightSkyMapUnits(self):
...
+        """
+        registry = self.registry
+
+        # dataset types
+        registry.makeRun(collection="test")
+        storageClass = StorageClass("testDataset")
+        registry.storageClasses.registerStorageClass(storageClass)
+
+        calexpType = DatasetType(name="CALEXP", dataUnits=("Camera", "Visit", "Sensor"),
+                                 storageClass=storageClass)
```
natelust, line 327: Im not an expert on the new system yet, but why is this being set up as a storage class with type Object, rather than the actual Calexp type?, same for the coadd below. Is it that it just dosn't matter?

natelust: Perhaps I am not understanding how the new system works under the hood

tests/test_sqlPreFlight.py
andy-slac: "test" is the name of the collection, same thing as was created at line 318. I will make it clearer by using local variable for both.

tests/test_sqlPreFlight.py
andy-slac: Not sure if I understand question - `storageClass` defines how objects are represented, and it is associated with DatasetType so all Datasets of the same DatasetType use same storageClass. In a sense there is no actual Calexp type, but there is a "calexp" DatasetType with storageClass something like "ExposureF".

In this particular case the actual names do not matter and they are arbitrary, and storageClass is not even used (we don't do any I/O) but it has to be defined for every DatasetType.

PR #87, opened 2018-09-12
DM-15210: technote content generation fixes, minor schema updates

TallJimbo:
This ticket revives the generation of DMTN-073 content from daf_butler YAML source, now using daf_butler classes to do the parsing instead of duplicating that code.  It also makes a few minor updates to that schema content, based on comments from RFC-484.

Note that this PR is currently a diff against DM-15424; I'll rebase and reset the diff to master once that lands.

### DIFF
--- a/config/schema.yaml
+++ b/config/schema.yaml
@@ -9,26 +9,26 @@ schema:
         primary_key: true
         nullable: false
         doc: >
-          a unique autoincrement field used the primary key for dataset.
+          A unique autoincrement field used the primary key for dataset.
       -
         name: dataset_type_name
         type: string
         nullable: false
         doc: >
-          the name of the DatasetType associated with this dataset; a
+          The name of the DatasetType associated with this dataset; a
           reference to the DatasetType table.
       -
         name: run_id
         type: int
         nullable: false
         doc: >
-          the id of the run that produced this dataset, providing access to
+          The id of the run that produced this dataset, providing access to
           coarse provenance information.
       -
         name: quantum_id
         type: int
         doc: >
-          the id of the quantum that produced this dataset, providing access
+          The id of the quantum that produced this dataset, providing access
           to fine-grained provenance information.
           may be null for datasets not produced by running a supertask.
       -
@@ -38,7 +38,7 @@ schema:
         name: abstract_filter
         type: string
         doc: >
-          string name for the abstract filter, frequently a single character.
+          String name for the abstract filter, frequently a single character.
       -
         name: physical_filter
         type: string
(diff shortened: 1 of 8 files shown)


config/schema.yaml
```diff
@@ -9,8 +9,7 @@ schema:
         primary_key: true
         nullable: false
         doc: >
-          a unique autoincrement field used as part of the primary key for
-          dataset.
+          a unique autoincrement field used the primary key for dataset.
```
timj: If docstrings are sentences they should start with capital letter.

config/schema.yaml
```diff
@@ -84,64 +83,60 @@ schema:
...
         src: quantum_id
-        tgt: quantum.execution_id
+        tgt: Quantum.execution_id
+      -
+        src: abstract_filter
+        tgt: AbstractFilter.abstract_filter
       -
         src: camera
         tgt: Camera.camera
-        limited: false
```
timj, line 100: Are these removals an anomaly of the diff?

python/lsst/daf/butler/core/utils.py
```diff
@@ -392,3 +392,11 @@ def getObjectSize(obj, seen=None):
         size += sum([getObjectSize(i, seen) for i in obj])
 
     return size
+
+
+def stripIfNotNone(s):
+    """Strip leading and trailing whitespace if the given object is not None.
+    """
```
timj: Add Parameters and Returns section.

python/lsst/daf/butler/instrument.py
```diff
@@ -96,12 +96,11 @@ def makeExposureEntryFromVisitInfo(dataId, visitInfo, snap=0):
     """
     avg = visitInfo.getDate()
```
timj, line 97: The docstring above should say `lsst.afw.image.VisitInfo` and `lsst.afw.image.Exposure`

python/lsst/daf/butler/instrument.py
```diff
@@ -96,12 +96,11 @@ def makeExposureEntryFromVisitInfo(dataId, visitInfo, snap=0):
...
     begin = DateTime(int(avg.nsecs(DateTime.TAI) - 0.5E9*visitInfo.getExposureTime()), DateTime.TAI)
+    end = DateTime(int(avg.nsecs(DateTime.TAI) + 0.5E9*visitInfo.getExposureTime()), DateTime.TAI)
```
timj, line 99: This is an undocumented dependency on `daf_base`. I'm a bit worried about this code relying on afw and daf_base being inside butler.

Is this going to be replaced by obs_metadata?

timj (approved):
Looks okay in general.

config/schema.yaml
TallJimbo: No, they became unnecessary, because I've made it so Foreign Keys are automatically removed from a limited schema if either of their tables are.

python/lsst/daf/butler/instrument.py
TallJimbo: It'll be replaced by obs_metadata and may be moved to obs_base even before that happens.

python/lsst/daf/butler/instrument.py
TallJimbo: :+1: for `lsst.afw.image.VisitInfo`, but this is a daf_butler Exposure DataUnit, not an `lsst.afw.image.Exposure`...I need to follow up on that rename.

PR #88, opened 2018-09-13
DM-15721: Refactor config hierarchy access and add __delitem__

timj:
@TallJimbo noted that `__delitem__` was missing. Rather than copy the hierarchy access code to a new method this PR factors it out so it is shared for multiple dunder methods and `__delitem__` is then implemented on top of that.

### DIFF
--- a/python/lsst/daf/butler/core/config.py
+++ b/python/lsst/daf/butler/core/config.py
@@ -272,57 +272,29 @@ class Config(collections.UserDict):
         else:
             return list(key)
 
-    def __getitem__(self, name):
-        data = self.data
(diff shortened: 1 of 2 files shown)


TallJimbo (approved):
Everything looks good, and it's a clear improvement.

I've pushed two more commits of my own:
 - more unit tests (I'm afraid I just ended up overwriting your latest ones, as mine were a bit more complete);
 - documentation for the edge case we just discussed on Slack.

timj:
Thanks for the extra tests. I've also removed UserDict completely. Running Jenkins now.

PR #89, opened 2018-09-25
DM-15850: Always import default storage classes

timj:


python/lsst/daf/butler/core/storageClass.py
```diff
@@ -342,6 +342,22 @@ def __init__(self, config=None):
...
+            Summary of the factory status.
+        """
+        sep = "\n"
+        return f"""Number of registered StorageClasses: {len(self._storageClasses)}
+
+StorageClasses
+--------------
+{sep.join(f"{s}: {self._storageClasses[s]}" for s in self._storageClasses)}
+"""
+
```
natelust, line 359: This would be nicer if the text was in a similar formatting, so maybe:
```
        returnString = f"""
            Number of registered StorageClasses: {len(self._storageClasses)}
            StorageClasses
            --------------
            {sep.join(f"{s}: {self._storageClasses[s]}" for s in self._storageClasses)}
             """
        return textwrap.dedent(returnString.lstrip("\n"))   
```

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -335,9 +335,29 @@ def __init__(self, config=None):
         self._storageClasses = {}
         self._configs = []
 
+        # Always seed with the default config
+        c = StorageClassConfig()
```
natelust: In general I think single letter variable names are not good, even in simple cases. If this was named something like defaultConfiguration, there would be no need for the comment. Alternatively, there is little point in naming the output with a single character to just consume it. This could be `self.addFromConfig(StorageClassConfig())` with the comment left above.

natelust (approved):
Some comments to consider, but I am  fine with the solution to this problem

python/lsst/daf/butler/core/storageClass.py
timj: I don't really like all the dedent shenanigans and I don't have a problem with the left justification so I'd prefer to leave it as is.

python/lsst/daf/butler/core/storageClass.py
timj: I'll remove the variable. `defaultConfiguration` seems like a very long variable name for such a short-lived variable.

PR #90, opened 2018-09-26
DM-15760: Change first/last validity from exposure ID to datetime

timj:


config/schema.yaml
```diff
@@ -53,16 +53,16 @@ schema:
         type: int
       -
         name: valid_first
-        type: int
+        type: datetime
         doc: >
-          First exposure identifier included in the range (inclusive).  may
-          be zero to indicate an open interval.
+          Date of first exposure included in the range (inclusive).  May
+          be far in the past to indicate an open interval.
```
TallJimbo: Should we say TAI explicitly here to avoid any possibility of confusion in the future?

And this is *mostly* off-topic, but it got me thinking: do you think we should by using MJDs for all of our datetimes in the Registry to avoid the possibility of people using database datetime UDFs that are inappropriate for our time systems?

TallJimbo (approved):

config/schema.yaml
timj: TAI MJD might be safer (and we don't need millisecond accuracy). Shall I make that change on this ticket?

TallJimbo: You're welcome to make it on this ticket as far as I'm concerned, but note that that will involve some code changes, *maybe* including some that extend beyond daf_butler, as we actually are populating the `datetime_begin` (etc) fields.

Just curious: what kind of accuracy would that give us?

timj: Single 64bit float using MJD is about 1 micro second accuracy I think. Integer MJD day and 64-bit float for fractional day seems to be 0.02 ns.

PR #91, opened 2018-09-26
DM-15871: Move doImport out of daf_butler.

TallJimbo:


python/lsst/daf/butler/butler.py
```diff
@@ -27,7 +27,8 @@
 import contextlib
 import logging
 
-from .core.utils import doImport, transactional
+from lsst.utils import doImport
```
timj, line 30: You probably should move the lsst.utils imports before the butler imports.

timj (approved):
Only comment is to fix the order of all the imports so that lsst.utils is above the butler internal imports.

python/lsst/daf/butler/butler.py
TallJimbo: It is already, isn't it?

timj: 😄 I think there might be one that isn't later on... Sorry.

TallJimbo:
@timj I've pushed a broader cleanup to the import order and style (since I found some pre-existing things).  Please make sure you don't hate it.

timj (approved):
Looks good to me.

PR #92, opened 2018-09-28
DM-15256: Include calibration repositories in Gen2->Gen3 conversion

czwa:
The Gen2->Gen3 conversion code ignored calibration repositories
entirely.  This fixes the issue, by having the gen2convert/walker.py
script directly query calibration repos as they are discovered.  As
calibRegistry.sqlite3 contains no mapper information, this is derived
by searching parent directories to find a valid mapper to use.
Matching DM-15760, this uses datetime objects for the
valid_first/valid_last entries in the Dataset table.

Information about the MaskedImageF was added to the storage
class/datastore yaml files to allow the FLAT to be handled properly.

Additional changes to support this in obs_subaru.

### DIFF
--- a/python/lsst/daf/butler/gen2convert/walker.py
+++ b/python/lsst/daf/butler/gen2convert/walker.py
@@ -22,6 +22,8 @@
 import os
 import pickle
 import yaml
+import sqlite3
+import datetime
 
 # register YAML loader for repositoryCfg.yaml files.
 import lsst.daf.persistence.repositoryCfg   # noqa F401
@@ -61,6 +63,7 @@ class ConversionWalker:
         self._skyMaps = dict()
         self._skyMapRoots = dict()
         self._visitInfo = dict()
+        self._calibDict = dict()
 
     def scanAll(self):
         """Recursively inspect and scan Gen2 data repositories.
@@ -101,11 +104,24 @@ class ConversionWalker:
         MapperClass = None
         parentPaths = []
         repoCfgPath = os.path.join(root, "repositoryCfg.yaml")
+        calibPath = os.path.join(root, "calibRegistry.sqlite3")
+        isCalib = False
         if os.path.exists(repoCfgPath):
             with open(repoCfgPath, "r") as f:
                 repoCfg = yaml.load(f)
             parentPaths = [parent.root for parent in repoCfg.parents]
             MapperClass = repoCfg.mapper
+        elif os.path.exists(calibPath):
+            # Slurp all the entries into _calibDict
+            self.readCalibInfo(calibPath=calibPath)
+            calibRoot = os.path.dirname(root)
+            mapperFilePath = os.path.join(calibRoot, "_mapper")
+            MapperClass = None
+            if os.path.exists(mapperFilePath):
+                with open(mapperFilePath, "r") as f:
+                    mapperClassPath = f.read().strip()
+                    MapperClass = doImport(mapperClassPath)
+            isCalib = True
         else:
             parentLinkPath = os.path.join(root, "_parent")
             if os.path.exists(parentLinkPath):
@@ -118,7 +134,7 @@ class ConversionWalker:
 
         # This directory has no _mapper, no _parent, and no repositoryCfg.yaml.
         # It probably just isn't a repository root.
-        if not (parentPaths or MapperClass):
+        if not (parentPaths or MapperClass or isCalib):
             log.debug("%s: not a data repository.", root)
             return None
 
@@ -188,7 +204,7 @@ class ConversionWalker:
                            if not self.tryRoot(os.path.join(dirPath, dirName))]
             relative = dirPath[len(repo.root) + 1:]
             for fileName in fileNames:
-                if fileName in ("registry.sqlite3", "_mapper", "repositoryCfg.yaml"):
+                if fileName in ("registry.sqlite3", "_mapper", "repositoryCfg.yaml", "calibRegistry.sqlite3"):
                     continue
                 filePath = os.path.join(relative, fileName)
                 dataset = extractor(filePath)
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/gen2convert/walker.py
```diff
@@ -101,11 +104,49 @@ def tryRoot(self, root):
...
+                        "valid_last": datetime.datetime.strptime(row["validEnd"], "%Y-%m-%d"),
+                    }
+                else:
+                    self._calibDict[(row["type"], row["calibDate"], row["ccd"])] = {
+                        "filter": row["filter"],
+                        "valid_first": datetime.datetime.strptime(row["validStart"], "%Y-%m-%d"),
+                        "valid_last": datetime.datetime.strptime(row["validEnd"], "%Y-%m-%d"),
+                    }
+
+            calibConn.close()
```
TallJimbo: You should be able to write this as:
```
with sqlite3.connect(calibPath) as calibConn:
    ...
```
to avoid the explicit `close` (and make sure the `close` happens in the presence of exceptions).  I was a bit surprised to find that that doesn't seem to be in the Python sqlite3 module documentation, but it definitely works.

python/lsst/daf/butler/gen2convert/walker.py
```diff
@@ -101,11 +104,49 @@ def tryRoot(self, root):
...
+
+            calibRoot = os.path.dirname(root)
+            parentPaths.append(calibRoot)
+            mapperFilePath = os.path.join(calibRoot, "_mapper")
+            if os.path.exists(mapperFilePath):
+                with open(mapperFilePath, "r") as f:
+                    mapperClassPath = f.read().strip()
+                MapperClass = doImport(mapperClassPath)
+            print("In calib: %s %s %d" % (calibPath, mapperFilePath, len(self._calibDict)))
+            isCalib = True
```
TallJimbo, line 124: It would be good to factor out all of this extraction of calibration registry content into a separate function or method, just for clarity (i.e. even if it's only called here, it will make the rest of this method easier to read).

I should add that I also don't know whether the schema of the HSC calibration registries is the same one used in the calibration registries for other cameras. I don't think we should try to support all other cameras on this ticket, but it might be worthwhile to take a look at this from the perspective of maybe having to inject some camera-customization hooks here.

python/lsst/daf/butler/gen2convert/walker.py
```diff
@@ -118,7 +159,7 @@ def tryRoot(self, root):
 
         # This directory has no _mapper, no _parent, and no repositoryCfg.yaml.
         # It probably just isn't a repository root.
-        if not (parentPaths or MapperClass):
+        if not ((parentPaths or MapperClass) or isCalib):
```
TallJimbo: The inner parentheses here do not need to and probably should not exist.

python/lsst/daf/butler/gen2convert/walker.py
```diff
@@ -199,6 +240,16 @@ def scanRepo(self, repo):
...
                     repo.datasets[dataset.datasetType.name][filePath] = dataset
 
+                    try:
+                        calibRow = self._calibDict[(
+                            dataset.datasetType.name, dataset.dataId["calibDate"], dataset.dataId["ccd"])]
+                        if calibRow["filter"] is None or calibRow["filter"] == dataset.dataId["filter"]:
+                            dataset.dataId["valid_first"] = calibRow["valid_first"]
+                            dataset.dataId["valid_last"] = calibRow["valid_last"]
+                            print(calibRow)
+                    except KeyError:
```
TallJimbo: Instead of using `KeyError` to do control flow for the common case where the dataset is not a calibration product, I'd recommend using something like:
```
calibRow = self._calibDict.get(..., None)
if calibRow is not None:
    ...
```
That's just a slight preference, and possibly one driven by experience with languages that approach exceptions differently from Python.  Feel free to push back if you disagree.

python/lsst/daf/butler/gen2convert/walker.py
```diff
@@ -199,6 +240,16 @@ def scanRepo(self, repo):
                     log.debug("%s: found %s with %s", repo.root, dataset.datasetType.name, dataset.dataId)
                     repo.datasets[dataset.datasetType.name][filePath] = dataset
 
+                    try:
+                        calibRow = self._calibDict[(
+                            dataset.datasetType.name, dataset.dataId["calibDate"], dataset.dataId["ccd"])]
+                        if calibRow["filter"] is None or calibRow["filter"] == dataset.dataId["filter"]:
+                            dataset.dataId["valid_first"] = calibRow["valid_first"]
+                            dataset.dataId["valid_last"] = calibRow["valid_last"]
+                            print(calibRow)
```
TallJimbo: This `print` looks like a debugging relic.  It'd be nice to keep it, but via `log.debug` and a more specific message so we only get it when we turn the verbosity up.

Also, I *think* we might want to have an `else` clause here that does a `log.warn`, because it seems like it might be a problem if we find the dataset in the calib registry but don't understand its filter.  Is that right?

TallJimbo (approved):
Approach looks good, but I have some style/structure comments.

I'm actually amazed it was this straightforward to add (and that it didn't require any changes to the `ConversionWriter` class as well).

python/lsst/daf/butler/gen2convert/walker.py
czwa: This check is designed to allow datasets to either depend on the filter (flat/sky) or not (dark/bias) as needed.  This was a debug statement, so I've made it a log.debug() call.

python/lsst/daf/butler/gen2convert/walker.py
czwa: I'm fine with this.  Using the exception was something I wasn't totally happy with, so a default value works nicely.

python/lsst/daf/butler/gen2convert/walker.py
czwa: Done.  I have a parentheses problem.

python/lsst/daf/butler/gen2convert/walker.py
czwa: Done.

python/lsst/daf/butler/gen2convert/walker.py
czwa: I've added the additional method to handle fetching information from the gen2 calibration registry.

Checking the registries that I can find for other cameras (DECam and comCam), it appears that the set ("dark","flat","fringe","sky","bias","defect") contains all tables and calibrations currently available.  I have modified this section to generate the calibDict query based on the tables that actually exist in the registry.

PR #95, opened 2018-10-05
DM-15914: Use obs_metadata for unit extraction

timj:


python/lsst/daf/butler/instrument.py
```diff
@@ -64,74 +62,72 @@ def _addCamera(self, registry):
         registry.addDataUnitEntry("Camera", {"camera": self.camera})
 
     def _addPhysicalFilters(self, registry):
+        print("REGISTERING PHYSICAL FILTER")
```
TallJimbo: I'm guessing you didn't intend to leave these in (more below).

TallJimbo (approved):

python/lsst/daf/butler/instrument.py
timj: There are plenty of prints scattered about. I'll remove them before merging.

timj: They are mostly there so I could work out what logic was being triggered when.

PR #96, opened 2018-10-09
DM-15686: Implement constraint check in associate method

andy-slac:
Current schema lacks actual constraints for collections, this commit
adds code to verify that added rows do not violate constraints. It can
only work reliably in single-user environment, we have to move
constraints to schema definition for more reliable validation.

### DIFF
--- a/python/lsst/daf/butler/registries/sqlRegistry.py
+++ b/python/lsst/daf/butler/registries/sqlRegistry.py
@@ -174,6 +174,49 @@ class SqlRegistry(Registry):
         config["table"] = table
         return SqlRegistryDatabaseDict(config, types=types, key=key, value=value, registry=self)
 
+    def _findDatasetId(self, collection, datasetType, dataId):
+        """Lookup a dataset ID.
+
+        This can be used to obtain a ``dataset_id`` that permits the dataset
+        to be read from a `Datastore`.
+
+        Parameters
+        ----------
+        collection : `str`
+            Identifies the collection to search.
+        datasetType : `DatasetType`
+            The `DatasetType`.
+        dataId : `dict`
+            A `dict` of `DataUnit` link name, value pairs that label the
+            `DatasetRef` within a collection.
+
+        Returns
+        -------
+        dataset_id : `int` or `None`
+            ``dataset_id`` value, or `None` if no matching Dataset was found.
+
+        Raises
+        ------
+        ValueError
+            If dataId is invalid.
+        """
+        self._validateDataId(datasetType, dataId)
+        datasetTable = self._schema.tables["Dataset"]
+        datasetCollectionTable = self._schema.tables["DatasetCollection"]
+        dataIdExpression = and_((self._schema.datasetTable.c[name] == dataId[name]
+                                 for name in self._dataUnits.getPrimaryKeyNames(
+                                     datasetType.dataUnits)))
+        result = self._connection.execute(select([datasetTable.c.dataset_id]).select_from(
+            datasetTable.join(datasetCollectionTable)).where(and_(
+                datasetTable.c.dataset_type_name == datasetType.name,
+                datasetCollectionTable.c.collection == collection,
+                dataIdExpression))).fetchone()
+        # TODO update unit values and add Run, Quantum and assembler?
+        if result is not None:
+            return result.dataset_id
+        else:
+            return None
+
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -482,17 +509,93 @@ def associate(self, collection, refs):
         """Add existing Datasets to a collection, possibly creating the
         collection in the process.
 
+        If a DatasetRef with the same exact `dataset_id`` is alredy in a
```
TallJimbo: typo: `alredy`

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -482,17 +509,93 @@ def associate(self, collection, refs):
...
+        ------
+        ValueError
+            If a Dataset with the given `DatasetRef` already exists in the
+            given collection.
         """
-        datasetCollectionTable = self._schema.tables["DatasetCollection"]
-        self._connection.execute(datasetCollectionTable.insert(),
-                                 [{"dataset_id": ref.id, "collection": collection} for ref in refs])
+        # A collection cannot contain more than one Dataset with the same
+        # DatasetRef. Our SQL schema does not enforce this constaint yet so
```
TallJimbo: typo: `constaint`

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -482,17 +509,93 @@ def associate(self, collection, refs):
...
+            # full scan of a collection to compare DatasetRef units
+            datasetTable = self._schema.tables["Dataset"]
+            datasetCollectionTable = self._schema.tables["DatasetCollection"]
+            query = datasetTable.select()
+            query = query.where(and_(datasetTable.c.dataset_id == datasetCollectionTable.c.dataset_id,
+                                     datasetCollectionTable.c.collection == collection))
+            result = self._connection.execute(query)
+            for row in result:
+                # skip DatasetRefs that are already there
+                refs = [ref for ref in refs if not _matchRef(row, ref)]
```
TallJimbo, line 595: This is a lot of list construction and re-construction.  Would it be better to pop `refs` from a `set` or `dict`?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -482,17 +509,93 @@ def associate(self, collection, refs):
...
+            ------
+            ValueError
+                If DatasetRef unit values match row data but their IDs differ.
+            """
+            if row.dataset_id == ref.id:
+                return True
+
+            columns = self._dataUnits.getPrimaryKeyNames(ref.datasetType.dataUnits)
+            dataId = ref.dataId
+            if all(row[col] == dataId[col] for col in columns):
```
TallJimbo, line 570: I think the columns compared here also needs to include `dataset_type_name`.

tests/test_sqlRegistry.py
```diff
@@ -50,6 +50,15 @@ class RegistryTests(metaclass=ABCMeta):
     def makeRegistry(self):
         raise NotImplementedError()
 
+    def asserRowCount(self, registry, table, count, where=None):
```
TallJimbo: typo: `asser`

TallJimbo (approved):
There may be a bug in that the uniqueness check doesn't seem to always account for different DatasetTypes.  Other than that, looks good, except for some typos.

python/lsst/daf/butler/registries/sqlRegistry.py
andy-slac: I did think about how to do it more naturally but not hard enough. Main issue is that our DataId is not hashable (it's a map). I could potentially define a key function for DataId but extracting the same key from database row would be somewhat complicated (need to know which columns are relevant and which are not). So given that this code would have to disappear anyways when we implement database constraints I decided for simplest and inefficient approach.

python/lsst/daf/butler/registries/sqlRegistry.py
andy-slac: I knew something was missing :grin: I'll try to make a unit test for this.

andy-slac: Added unit test and fixed my mistake.

PR #99, opened 2018-10-22
DM-16077: Improve pickling support for butler classes.

andy-slac:
Also added unit tests to verify that pickling works for classes and
cases that we need.


timj (approved):
Looks okay to me.

PR #100, opened 2018-10-23
DM-15537: Rename sensor->detector and camera->instrument.

TallJimbo:


timj (approved):

config/dataUnits.yaml
```diff
@@ -81,38 +82,40 @@ dataUnits:
...
-        SQL representation; even a definition table is not necessary, given that
-        the allowable values and the associated spatial regions are best
-        computed on-the-fly.  SkyPix units are preferred to SkyMap (i.e. Tract-
-        Patch) units for Datasets without any overlap regions (e.g. sharded
-        reference catalogs).
+        HEALPix; we will select and support just one, but which is TBD). Has
+        no SQL representation; even a definition table is not necessary, given
+        that the allowable values and the associated spatial regions are best
+        computed on-the-fly.  SkyPix units are preferred to SkyMap (i.e.
+        Tract- Patch) units for Datasets without any overlap regions (e.g.
```
andy-slac: Space after dash

config/dataUnits.yaml
```diff
@@ -81,38 +82,40 @@ dataUnits:
...
-        and custom indexes, which are much harder to implement across multiple
-        RDMSs.  As a result, we will (at least initially) try to define just a
-        single level for all SkyPix values.  Our preliminary guess is that this
-        level should have pixels be approximately (within a factor of ~4) of the
-        size of a single Sensor on the sky.
-      spatial: true
+        level of the standard pixelization: if all SkyPix values are at a
+        single level, they can be indexed using standard B-Trees and compared
+        with simple equality comparison.  In contrast, comparing SkyPix values
+        at different levels requires pixelization- specific bit-shifting
```
andy-slac: Space after dash here too

andy-slac (commented):
Couple of minor comments.

One side note - we should probably try to test the code with something other than sqlite at some point, I'm worried about case sensitivity for database names.

TallJimbo:
@andy-slac, two thoughts on case sensitivity:
 - I'm hoping that some of my data ID refactoring I have planned for Nov.-Dec. will result in us not using the same names for fields and tables quite so much, which may help a bit.
 - Once we're slightly more stable I'm going to try to interest the HSC team at NAOJ in starting to try using Gen3 Butler, and they'll certainly want to do everything with PostgreSQL.  And of course there's an Oracle implementation coming at NCSA at some point.
I don't know of any other databases we could easily integrate into unit or integration tests, though.  Do you?

andy-slac:
For unit tests I don't think there is any other viable option besides sqlite, anything else would require running a server.
My worry about case sensitivity is mostly for hand-written SELECTs which do not quote table/column names, the rules for unquoted identifiers depend on the backend (just as the rules for quoting).

PR #104, opened 2018-11-30
DM-16648: new classes for DataUnits (now Dimensions)

TallJimbo:
This is one piece of DM-15034, but as it's a pretty big redesign I wanted to get some early feedback.

This is hence a PR against tickets/DM-15034, not master, though those are currently the same.  I'll merge this into that branch when it's reviewed, and then merge other sub-task tickets to that branch until all of DM-15034 is done.  That will include making all of the other Registry and Preflight code use these classes and removing the old ones, as well as adding some new classes to represent Data IDs in at least some contexts.

### DIFF
--- a/config/dimensions.yaml
+++ b/config/dimensions.yaml
@@ -1,0 +1,275 @@
+dimensions:
+  version: 0
+  units:
+    Instrument:
+      doc: >
+        An entity that produces observations.  An Instrument defines a set of
+        PhysicalFilters and Detectors and a numbering system for the Exposures
+        and Visits that represent observations with it.
+      link:
+      - instrument
+
+    AbstractFilter:
+      doc: >
+        A filter that is not associated with a particular Instrument.  An
+        abstract filter can be used to relate similar physical filters, and
+        is typically the filter associated with coadds.
+      link:
+      - abstract_filter
+
+    PhysicalFilter:
+      dependencies:
+        required:
+        - Instrument
+        optional:
+        - AbstractFilter
+      doc: >
+        A filter associated with a particular Instrument.  PhysicalFilters are
+        used to identify datasets that can only be associated with a single
+        observation.
+      link:
+      - physical_filter
+
+    Detector:
+      dependencies:
+        required:
+        - Instrument
+      doc: >
+        A detector associated with a particular Instrument (not an observation
+        of that detector; that requires specifying an exposure or visit as
+        well).
+      link:
+      - detector
+
+    Visit:
+      dependencies:
+        required:
+        - Instrument
+        optional:
+        - PhysicalFilter
+      doc: >
+        A sequence of observations processed together, comprised of one or
+        more Exposures from the same Instrument with the same pointing and
+        PhysicalFilter.
+        The Visit table contains metadata that is both meaningful only for
+        science Exposures and the same for all Exposures in a Visit.
+      link:
+      - visit
+      hasRegion: true
+
+    Exposure:
+      dependencies:
+        required:
+        - Instrument
+        optional:
+        - Visit
+        - PhysicalFilter
+      doc: >
+        An observation associated with a particular instrument.  All direct
+        observations are identified with an Exposure, but derived datasets
+        that may be based on more than one Exposure (e.g. multiple snaps) are
+        typically identified with Visits instead, even for instruments that
+        don't have multiple Exposures per Visit.  As a result, Instruments
+        that don't have multiple Exposures per Visit will typically have Visit
+        entries that are essentially duplicates of their Exposure entries.
+
+        The Exposure table contains metadata entries that are relevant for
+        calibration Exposures, and does not duplicate entries in Visit that
+        would be the same for all Exposures within a Visit.
+      link:
+      - exposure
+
+    ExposureRange:
+      dependencies:
+        required:
+        - Instrument
+      doc: >
+        An inclusive range of Exposure dates that may be open in either
+        direction, typically used to identify master calibration products.
+        There is no SQL table associated with ExposureRanges; there is no
+        additional information associated with an ExposureRange besides the
+        instrument, valid_first, and valid_last fields already present in
+        Dataset.
+      link:
+      - valid_first
+      - valid_last
+
+    SkyPix:
+      doc: >
+        A pixel in a hierarchical decomposition of the sky (e.g. HTM, Q3C, or
+        HEALPix; we will select and support just one, but which is TBD). Has
+        no SQL representation; even a definition table is not necessary, given
+        that the allowable values and the associated hasRegion regions are best
+        computed on-the-fly.  SkyPix units are preferred to SkyMap (i.e.
+        Tract-Patch) units for Datasets without any overlap regions (e.g.
+        sharded reference catalogs).
+
+        There are also considerable advantages to standardizing on just one
+        level of the standard pixelization: if all SkyPix values are at a
+        single level, they can be indexed using standard B-Trees and compared
+        with simple equality comparison.  In contrast, comparing SkyPix values
+        at different levels requires pixelization-specific bit-shifting
+        operations and custom indexes, which are much harder to implement
+        across multiple RDMSs.  As a result, we will (at least initially) try
+        to define just a single level for all SkyPix values.  Our preliminary
+        guess is that this level should have pixels be approximately (within a
+        factor of ~4) of the size of a single Detector on the sky.
+      hasRegion:
+        true
+      link:
+      - skypix
+
+    SkyMap:
+      doc: >
+        A set of Tracts and Patches that subdivide the sky into rectangular
+        regions with simple projections and intentional overlaps.
+      link:
+      - skymap
+
+    Tract:
+      dependencies:
+        required:
+        - SkyMap
+      doc: >
+        A large rectangular region mapped to the sky with a single map
+        projection, associated with a particular SkyMap.
+      link:
+      - tract
+      hasRegion: true
+
+    Patch:
+      dependencies:
+        required:
+        - SkyMap
+        - Tract
+      doc: >
+        A rectangular region within a Tract.
+      link:
+      - patch
+      hasRegion: true
+
+    Label:
+      doc: >
+        An arbitrary string value.  There is no SQL representation or
+        constraint on the values a Label can take.  Multiple labels are
+        represented (by convention) by a comma-separated string.
+      link:
+      - label
+
+  joins:
+
+    VisitDetectorRegion:
+      doc: >
+        A many-to-many join table that provides region information for
+        Visit-Detector combinations.
+      lhs: [Visit]
+      rhs: [Detector]
+      hasRegion: true
+
+    VisitDetectorSkyPixJoin:
+      doc: >
+        A spatial join table that relates Visit+Detector to SkyPix, also used
+        as an intermediate when relating Visit+Detector to SkyMap DataUnits.
+
+        Entries are expected to be calculated outside the database and
+        added/updated whenever VisitDetectorRegion is.
+
+      lhs: [Visit, Detector]
+      rhs: [SkyPix]
+
+    VisitSkyPixJoin:
+      doc: >
+        A spatial join table that relates Visit to SkyPix, also used as an
+        intermediate when relating Visit to SkyMap DataUnits.
+
+        This can and probably should be implemented as a trivial view on
+        VisitDetectorSkyPixJoin.
+
+      lhs: [Visit]
+      rhs: [SkyPix]
+      summarizes: [VisitDetectorSkyPixJoin]
+
+    PatchSkyPixJoin:
+      doc: >
+        A spatial join table that relates Patch to SkyPix, also used as an
+        intermediate when relating Patch to Instrument DataUnits.
+
+        Entries are expected to be calculated outside the database and
+        added along with the Patch itself.
+
+      lhs: [Patch]
+      rhs: [SkyPix]
+
+    TractSkyPixJoin:
+      doc: >
+        A spatial join table that relates Tract to SkyPix, also used as an
+        intermediate when relating Tract to Instrument DataUnits.
+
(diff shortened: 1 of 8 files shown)


config/dimensions.yaml
```diff
@@ -0,0 +1,260 @@
...
+      doc: >
+        A many-to-many join table that provides region information for
+        Visit-Detector combinations.
+      lhs: [Visit]
+      rhs: [Detector]
+      hasRegion: true
+
+    VisitDetectorSkyPixJoin:
+      doc: >
+        A hasRegion join table that relates Visit+Detector to SkyPix, also used
```
andy-slac: Not very clear from the doc itself what "hasRegion" join table means here.

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -0,0 +1,194 @@
...
+
+        Returns
+        -------
+        dependencies : `DimensionSet` of `DimensionUnit`
+        """
+        return self._allDependencies if optional else self._requiredDependencies
+
+
+class DimensionUnit(Dimension):
+    r"""A discrete unit of data used to organize `Dataset`\s and associate
```
andy-slac: I think we had trouble with flake8 and raw docstrings in the past, to avoid backslashes references can be written as `` `Datasets <Dataset>` ``

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -0,0 +1,194 @@
...
+            while the primary keys of optional dependencies are not.
+
+        Returns
+        -------
+        dependencies : `DimensionSet` of `DimensionUnit`
+        """
+        return self._allDependencies if optional else self._requiredDependencies
+
+
+class DimensionUnit(Dimension):
```
andy-slac: `DimensionUnit` confuses me somewhat (not sure about other people). Would it be clearer to just call this class `Dimension` (and base class something else)?

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -0,0 +1,194 @@
...
+        self._primaryKey = frozenset(primaryKey)
+
+    def __eq__(self, other):
+        try:
+            return self.universe is other.universe and self.name == other.name
+        except AttributeError:
+            return NotImplemented
+
+    def __hash__(self):
+        return hash(tuple(self.name))
```
andy-slac: `tuple(self.name, )` or `hash(self.name)`?

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -0,0 +1,194 @@
...
+    @property
+    def doc(self):
+        """Documentation for this dimension (`str`).
+        """
+        return self._doc
+
+    @property
+    def primaryKey(self):
+        """The names of fields that uniquely identify this dimension in a
+        data ID dict(`frozenset` of `str`).
```
andy-slac: Space before paren.

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -0,0 +1,194 @@
...
+        self._rhs = DimensionUnitSet(universe, rhs, optional=False)
+        # self._summarizes initialized later in DimensionMeta.load.
+
+    @property
+    def lhs(self):
+        r"""The `DimensionUnits`\s on the left hand side of the join
+        (`DimensionUnitSet`).
+
+        Left vs. right is completely arbitrary; the terminology simply provides
+        an easy way to distiguish between the two sides.
```
andy-slac, line 169: Should we drop lhs/rhs entirely if they are not useful and just do `def joins(): return (lhs, rhs)`?

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -0,0 +1,194 @@
...
+        """
+        return self._rhs
+
+    @property
+    def summarizes(self):
+        r"""A set of other `DimensionJoin`\s that provide more fine-grained
+        relationships this one (`DimensionJoinSet`).
```
andy-slac: _than_ this one?

python/lsst/daf/butler/core/dimensions/sets.py
```diff
@@ -0,0 +1,402 @@
...
+    which would then cause it to be added back into the result set.
+
+    The constructor and `union` and `intersection` methods for
+    `DimensionUnitSet` take an extra ``optional`` argument that controls
+    whether optional dependencies of all elements are also included in
+    the returned set.  This defaults to `False`.
+    """
+
+    def __repr__(self):
+        return f"DimensionUnitSet({str(self)})"
```
andy-slac: Do you need `str` here, I think it's implied?

python/lsst/daf/butler/core/dimensions/sets.py
```diff
@@ -0,0 +1,402 @@
...
+
+    def _populate(self, names, optional=False):
+        expanded = set()
+
+        # recursive local function to add dependencies of dependencies
+        def visit(unit):
```
andy-slac: I'd probably use generator instead of closure for this.

python/lsst/daf/butler/core/dimensions/sets.py
```diff
@@ -0,0 +1,402 @@
...
+    """
+
+    def __repr__(self):
+        return f"DimensionJoinSet({str(self)})"
+
+    def _populate(self, names):
+        for dimension in self.universe._joins:
+            if dimension.name in names:
+                self._elements[dimension.name] = dimension
+                names.remove(dimension.name)
```
andy-slac: can `names` be a `frozenset`?

andy-slac (commented):
I only looked at few modules and they look OK (but there are many new things so it needs some guesswork), few comments for now. I'll try to look at the rest later today if I can find time.

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -0,0 +1,194 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from ..utils import PrivateConstructorMeta
+from .sets import DimensionUnitSet
+
+__all__ = ("Dimension", "DimensionUnit", "DimensionJoin")
```
andy-slac: This is supposed to be before imports or it may not work as expected (same for other files).

python/lsst/daf/butler/core/dimensions/sets.py
```diff
@@ -0,0 +1,402 @@
...
+from abc import abstractmethod
+from collections.abc import Set
+from collections import OrderedDict
+
+__all__ = ("DimensionSet", "DimensionUnitSet", "DimensionJoinSet", "UniverseMismatchError")
+
+
+class UniverseMismatchError(Exception):
+    """Exception thrown when trying to relate dimension objects from one
+    universe with those from another.
```
andy-slac: So `daf_butler` is ready to be deployed across multiverse now? 😃

python/lsst/daf/butler/core/dimensions/graph.py
```diff
@@ -0,0 +1,362 @@
...
+    `DimensionGraph` is conceptually just the combination of a
+    `DimensionUnitSet` (accessible via the `units` property) and a
+    `DimensionJoinSet` (accessible via the `joins` method) that are
+    guaranteed to have the following properties:
+
+     - `units` always contains all dependencies (both required and optional)
+     of all of its elements.
+
+     - `joins()` returns all joins whose dependencies are present in `units`
+     (though by default it may filter out some of these; see the `join`
```
andy-slac: `` `join` `` -> `` `joins` ``

python/lsst/daf/butler/core/dimensions/graph.py
```diff
@@ -0,0 +1,362 @@
...
+    not implement the `collections.abc.Set` interface, but it does provide
+    its set comparison, intersection, and union operations.
+    Set difference and symmetric difference are not supported, as these are in
+    conflict with the requirement a graph always includes all of the
+    dependenices of all of its dimensions.
+
+    `DimensionGraph`\s can only be constructed by loading definitions from
+    config via the `load` method.  This returns a special "universe"
+    `DimensionGraph` that defines the maximal `DimensionUnit` and
+    `DimensionJoin` sets.  Dimension objects from different objects are not
```
andy-slac: "different objects" means different graphs?

python/lsst/daf/butler/core/dimensions/graph.py
```diff
@@ -0,0 +1,362 @@
...
+        for joinName, subconfig in config["joins"].items():
+            joinsToDo[joinName] = frozenset(subconfig["lhs"] + subconfig["rhs"])
+
+        while unitsToDo:
+
+            # Create and add to universe any DimensionUnits whose
+            # dependencies are already in the universe.
+            namesOfUnblockedUnits = [name for name, deps in unitsToDo.items()
+                                     if deps.issubset(universe._units.names)]
+            namesOfUnblockedUnits.sort()  # break topological ties with lexographical sort
```
andy-slac: lexographical -> lexicographical  (below too)

python/lsst/daf/butler/core/dimensions/graph.py
```diff
@@ -0,0 +1,362 @@
...
+                join = DimensionJoin._construct(universe, name=joinName, config=config["joins"][joinName])
+                universe._joins._elements[joinName] = join
+                del joinsToDo[joinName]
+
+        if joinsToDo:
+            raise ValueError("Orphaned join(s) found: [{}]".format(", ".join(joinsToDo.keys())))
+
+        # The 'summarizes' field of DimensionJoin needs to be populated in the
+        # opposite order (DimensionJoins that relate more DimensionUnits
+        # summarize those that relate fewer), so we put off setting those
```
andy-slac: I think it's the opposite, fewer Dimensions should summarize more Dimensions?

python/lsst/daf/butler/core/dimensions/graph.py
```diff
@@ -0,0 +1,362 @@
...
+                continue
+            unitNames.add(unit.name)
+            unitNames |= unit.dependencies().names
+        # Make a set of the names of all requested joins.
+        joinNames = set()
+        for join in joins:
+            if not isinstance(join, DimensionJoin):
+                join = self._joins[join]
+            else:
+                UniverseMismatchError.check(self, unit)
```
andy-slac: `unit` -> `join`?

andy-slac (approved):
Few more (silly) comments. Looks fine, don't have time to look at unit test though, presume it's fine too.

config/dimensions.yaml
TallJimbo: Unintentional, overzealous find-and-replace from "spatial" to "hasRegion"; will fix.

python/lsst/daf/butler/core/dimensions/elements.py
TallJimbo: I think I once had a similar conversation with @timj and @jonathansick that landed on the opposite conclusion, with raw docstrings preferred.  I hope one of them can chime in with a clarification.

python/lsst/daf/butler/core/dimensions/elements.py
TallJimbo: Agreed.

python/lsst/daf/butler/core/dimensions/elements.py
TallJimbo: Good catch, should have been the latter.

python/lsst/daf/butler/core/dimensions/elements.py
TallJimbo: Possibly; I'd like to try to go through the transition to using these classes first to verify that we don't need them first.

python/lsst/daf/butler/core/dimensions/sets.py
TallJimbo: I wasn't sure it was `str` that was implied rather than `repr`, which would have been bad.  But it looks like you're right.

python/lsst/daf/butler/core/dimensions/sets.py
TallJimbo: When you get a chance, could you elaborate?  I'm not quite sure what you're proposing.

python/lsst/daf/butler/core/dimensions/sets.py
TallJimbo: Apparently not; I'll transform to `set` at the beginning of this function both to ensure it is one and avoid argument-mutating side-effects.

python/lsst/daf/butler/core/dimensions/graph.py
TallJimbo: Different universes; you can create one universe graph and call `subgraph` to make a new graph that is interoperable with the first.

python/lsst/daf/butler/core/dimensions/sets.py
andy-slac: I mean something like:


            def visit(unit):
                yield unit.name
                for dep in unit.dependencies(optional=optional):
                    yield from visit(dep)

            expaned = set(visit(unit))

python/lsst/daf/butler/core/dimensions/elements.py
TallJimbo: I gave this a try, and it doesn't quite work; the names for lots of other attributes and methods make more sense if the name for the base concept is really a first-class name.

Would `elements` instead of `units` be any better?  Any other ideas?

TallJimbo:
While this initial review was very helpful, the design has moved on quite a bit, and the integration with the rest of daf_butler is now complete.  I'm just going to close this PR and open a new one for the entirety of DM-15034.

PR #105, opened 2018-12-10
DM-15034: new classes for DataUnits (now Dimensions) and Data IDs

TallJimbo:
This is a complete overhaul of the code that implements the DataUnit system, including a rename of the concept to Dimension.  That made for some pretty significant `Registry` API changes as well.

The goals are:
 - move a lot of the dimension-manipulation logic in preflight into the Dimension system itself;
 - remove little inconsistencies that had cropped up in Butler and Registry APIs, generally by making them all more flexible and providing common standardization code for similar APIs.
 - add a `DataId` class to standardize and validate user-provided data ID dictionaries and kwargs, so downstream code can expect things in standard form.  This also unblocks several future tickets.


python/lsst/daf/butler/butler.py
```diff
@@ -200,23 +200,74 @@ def transaction(self):
...
-            An identifier with `DataUnit` names and values.
-            When `None` a `DatasetRef` should be supplied as the second
-            argument.
+        dataId : `dict` or `DataId`
+            A `dict` of `Dimension` link name, value pairs that label the
+            `DatasetRef` within a Collection. When `None`, a `DatasetRef`
+            should be provided as the second argument.
         producer : `Quantum`, optional
             The producer.
+        kwds : `dict`
```
jonathansick: I'd do It something like

```
kwds
   Additional keyword arguments used to construct or augment the `DataId`. See `TODO` parameters.
```

Then make `TODO` the API reference to whatever the `self.registry.addDataset` method is. This way the user will actually be able to figure out what these keyword arguments are.

python/lsst/daf/butler/butler.py
```diff
@@ -200,23 +200,74 @@ def transaction(self):
             with self.datastore.transaction():
                 yield
 
+    def _standardizeArgs(self, datasetRefOrType, dataId=None, **kwds):
+        """Standardize the arguments passed to several Butler APIs.
+
+        Butler APIs that conceptually need a DatasetRef also allow passing a
```
jonathansick: I'd make these two paragraphs a Notes section and move it to the bottom. A pattern I've noticed from our docstrings is a general abuse of the [Extended Summary](
https://developer.lsst.io/python/numpydoc.html#extended-summary) section. When I say the extended summary can be "A few sentences" I do mean it :) Having heavy summaries pushes the Parameters down the page by variable amounts and makes docs harder for readers to scan.

python/lsst/daf/butler/core/config.py
```diff
@@ -126,6 +126,7 @@ class Config(collections.abc.MutableMapping):
     Note that adding a multi-level key implicitly creates any nesting levels
     that do not exist, but removing multi-level keys does not automatically
     remove empty nesting levels.  As a result::
+
         >>> c = Config()
```
jonathansick, line 130: The `::` and `>>>` syntaxes are (I believe) mutually exclusive. I'd do:

```rst
As a result:

>>> c = Config()
```

and so on.

python/lsst/daf/butler/core/dimensions/dataId.py
```diff
@@ -0,0 +1,351 @@
...
+from collections.abc import Mapping
+from .graph import DimensionGraph
+from .elements import Dimension
+
+
+class DataId(Mapping):
+    r"""A dict-like identifier for data usable across multiple collections
+    and `DatasetType`\s.
+
+    They keys of a `DataId` correspond to the "link" columns of one or more
```
jonathansick: Please move all this content to the [Notes](
https://developer.lsst.io/python/numpydoc.html#notes) section.

python/lsst/daf/butler/core/dimensions/dataId.py
```diff
@@ -0,0 +1,351 @@
...
+
+import itertools
+from collections.abc import Mapping
+from .graph import DimensionGraph
+from .elements import Dimension
+
+
+class DataId(Mapping):
+    r"""A dict-like identifier for data usable across multiple collections
+    and `DatasetType`\s.
```
jonathansick, line 32: To do plurals of API references, it is true that you need to use `\`. In my experience I've needed to add a space, like this (which gets eaten on render):

```rst
`DatasetType`\ s.
```

Another option, which often reads better is to never make class names plural, and instead add a plural noun. For example:

```rst
... multiple collections and `DatasetType` instances.
```

This is something I've learned from the [Google Style Guide](
https://developers.google.com/style/api-reference-comments#documentation_basics) which I'll be presenting to DM soon as an RFC.

python/lsst/daf/butler/core/dimensions/dataId.py
```diff
@@ -0,0 +1,351 @@
...
+        Spatial region on the sky associated with this combination of
+        dimension entries.
+    entries : `dict`, optional
+        A nested dictionary of additional metadata column values associated
+        with these dimensions, with `DimensionElement` instances or `str`
+        names as the outer keys, `str` column names as inner keys, and
+        column values as innter dictionary values.
+    extra : `dict`, optional
+        Additional key-value pairs to update ``dataId`` with.
+    kwds : `dict`, optional
```
jonathansick, line 61:
```rst
kwds
    Additional keyword arguments passed to `TODO`.
```

Where `TODO` is the API object, or objects, that receive the keyword arguments.

Note that there's no type information on the `kwds` parameter either. To do the user, kwds isn't a `dict`.

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -0,0 +1,278 @@
...
+    `DimensionElement` has exactly two subclasses: `Dimension` and
+    `DimensionJoin`.
+
+    `DimensionElement` objects are not directly constructable; they can only
+    be obtained (possibly indirectly) from a special "universe"
+    `DimensionGraph` loaded from configuration.
+    """
+
+    #
+    # Constructor is private, so its docs are just comments.
```
jonathansick, line 39: Interesting approach.

python/lsst/daf/butler/core/dimensions/graph.py
```diff
@@ -0,0 +1,473 @@
...
+    component = "dimensions"
+    requiredKeys = ("version", "units", "joins")
+    defaultConfigFile = "dimensions.yaml"
+
+
+class DimensionGraph:
+    r"""A automatically-expanded collection of both `Dimension`\s and
+    `DimensionJoin`\s.
+
+    A `DimensionGraph` behaves much like a `DimensionSet` containing only
```
jonathansick, line 57: Move these paragraphs to a [Notes](
https://developer.lsst.io/python/numpydoc.html#notes) section.

python/lsst/daf/butler/core/dimensions/graph.py
```diff
@@ -0,0 +1,473 @@
...
+        """
+        return DimensionGraph(self.universe, self._dimensions.union(*others), optional=optional)
+
+    def intersection(self, *others, optional=False):
+        """Return a new graph containing all elements that are in both  ``self``
+        and all of the other given graphs.
+
+        Parameters
+        ----------
+       *others : iterable over `Dimension` or `str`.
```
jonathansick, line 352:
```rst
*others : `Dimension` or `str`
    Other sets whose elements may be included in the result.
```

Remember that `others` is an iterable only internally. To a user, they pass `Dimension` or `str` instances as positional arguments.

python/lsst/daf/butler/core/dimensions/sets.py
```diff
@@ -0,0 +1,558 @@
...
+    except AttributeError:
+        pass
+    return DimensionNameSet(element.name if isinstance(element, DimensionElement) else element
+                            for element in container)
+
+
+class DimensionSet(Set):
+    r"""A custom set/dict hybrid class for collections of `DimensionElement`\s.
+
+    `DimensionSet` objects implement the full (immutable)
```
jonathansick: Move these paragraphs to a [Notes](
https://developer.lsst.io/python/numpydoc.html#notes) section.

python/lsst/daf/butler/core/dimensions/sets.py
```diff
@@ -0,0 +1,558 @@
...
+        elif len(t) == 0:
+            return default
+        return t[0]
+
+
+class DimensionNameSet:
+    r"""An incomplete, name-only stand-in for `DimensionSet` or
+    `DimensionGraph`.
+
+    Because true `DimensionSet`\s and `DimensionGraph`\s cannot be constructed
```
jonathansick, line 408: Move these paragraphs to a [Notes](
https://developer.lsst.io/python/numpydoc.html#notes) section.

jonathansick (commented):
I haven't covered all of the docstrings, but I've made some line comments about specific examples of the different docstring issues you've asked about. To summarize:

- Use the [Notes](
https://developer.lsst.io/python/numpydoc.html#notes) section for those multi-paragraph discussions. This prevents the Parameters section from getting pushed down the page, making things harder to scan.

- Plurals generally need a space after the backslash, `DataId`\ s. But I'm starting to like the Google approach to this, which is to never pluralize class names, and instead add a plural noun after the class name. "`DataId` instances".

- For keyword arguments, leave out type information since the user doesn't actually pass a `dict`. Make sure to include the API name in the description that receives these keyword arguments.

- For extra positional arguments, describe the type of individual positional arguments. The fact that it becomes a collection is an internal detail.

Thanks!

python/lsst/daf/butler/butler.py
```diff
@@ -200,23 +200,74 @@ def transaction(self):
...
+        and a `DataId` or `dict`.
+
+        Standardization of `dict` vs `DataId` is best handled by passing the
+        returned ``dataId`` (and ``kwds``) to `Registry` APIs, which are
+        generally similarly flexible.
+
+        Parameters
+        ----------
+        datasetRefOrType : `DatasetRef`, `DatasetType`, or `str`
+            When `DatasetRef` the `dataId` should be `None`.
```
andy-slac, line 209: ` ``dataId`` `?

python/lsst/daf/butler/core/dimensions/dataId.py
```diff
@@ -0,0 +1,351 @@
...
+from .graph import DimensionGraph
+from .elements import Dimension
+
+
+class DataId(Mapping):
+    r"""A dict-like identifier for data usable across multiple collections
+    and `DatasetType`\s.
+
+    They keys of a `DataId` correspond to the "link" columns of one or more
+    `Dimension`\s, while values identify particular rows in the tables or view
```
andy-slac: Should `view` be plural too?

python/lsst/daf/butler/core/dimensions/dataId.py
```diff
@@ -0,0 +1,351 @@
...
+        or ``dimension``, or when dimensions are inferred from the provided
+        link keys.
+    region : `lsst.sphgeom.ConvexPolygon`, optional
+        Spatial region on the sky associated with this combination of
+        dimension entries.
+    entries : `dict`, optional
+        A nested dictionary of additional metadata column values associated
+        with these dimensions, with `DimensionElement` instances or `str`
+        names as the outer keys, `str` column names as inner keys, and
+        column values as innter dictionary values.
```
andy-slac: Typo in `innter`

python/lsst/daf/butler/core/dimensions/dataId.py
```diff
@@ -0,0 +1,351 @@
...
+
+        if dimensions is None:
+            if dimension is None:
+                if universe is None:
+                    raise ValueError(f"Cannot infer dimensions without universe.")
+                allLinks = dict(dataId)
+                if extra is not None:
+                    allLinks.update(extra)
+                allLinks.update(kwds)
+                dimensions = universe.extract(dim for dim in universe if dim.link.issubset(allLinks))
```
andy-slac, line 139: This branch does not set `dimension`, is this intended (branch below tries to deduce it from `dimensions`)?

python/lsst/daf/butler/core/dimensions/dataId.py
```diff
@@ -0,0 +1,351 @@
...
+                self.region = self.entries[holder].get("region", None)
+
+    @property
+    def dimensions(self):
+        """The dimensions this `DataId` identifies (`DimensionGraph`).
+        """
+        return self._dimensions
+
+    @property
+    def dimension(self):
```
andy-slac: General comment - `dimension/dimensions` look almost indistinguishable, maybe use something different for one of them to avoid confusion (e.g. `leafDimension`)?

python/lsst/daf/butler/core/dimensions/dataId.py
```diff
@@ -0,0 +1,351 @@
...
+
+    def __eq__(self, other):
+        try:
+            return self._links == other._links
+        except AttributeError:
+            # also compare equal to regular dicts with the same keys and values
+            return self._links == other
+
+    def __hash__(self):
+        return hash(tuple(self._links.items()))
```
andy-slac: Does tuple have to be ordered for hashing?

python/lsst/daf/butler/core/dimensions/dataId.py
```diff
@@ -0,0 +1,351 @@
...
+
+    def fields(self, element, region=True, metadata=True):
+        """Return the entries for a particular `DimensionElement`.
+
+        Parameters
+        ----------
+        dimension : `DimensionElement` or `str`
```
andy-slac: dimension -> element?

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -0,0 +1,278 @@
...
+    #     associated with it.
+    # hasRegion : `bool`
+    #     Whether entries for this dimension are associated with a region
+    #     on the sky.
+    # link : iterable of `str`
+    #     The names of primary key fields used by this element but not any of
+    #     its dependencies.
+    # required : iterable of `str`
+    #     The names of the `Dimension`\ s whose primary keys are a subset of
+    #     this element's primary key.
```
andy-slac, line 56: More of a nitpick - this comment seems to imply  that schema is implemented in a certain way (with composite PKs), though there may be other ways to define PKs. I think this is also true for all uses of "primary key" below, and I'm not proposing to change that now.

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -0,0 +1,278 @@
...
+    # optional : iterable of `str`
+    #     The names of the `Dimension`\ s whose primary keys are specified by
+    #     foreign keys in this element.
+    # doc : `str`
+    #     Documentation string for this element.
+    #
+    def __init__(self, universe, name, hasRegion, link, required, optional, doc):
+        from .sets import DimensionSet
```
andy-slac, line 64: Is it here because of circular imports?

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -0,0 +1,278 @@
...
+            return self._optionalDependencies
+        raise ValueError("At least one of 'required' and 'optional' must be True.")
+
+    def graph(self, optional=False):
+        """Return the minimal `DimensionGraph` that contains ``self``.
+
+        Parameters
+        ----------
+        optional : `bool`
+            If `True` (default), include optional as well as required
```
andy-slac: Default is `False`

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -0,0 +1,278 @@
...
+    def __init__(self, universe, name, config):
+        from .sets import DimensionSet
+
+        lhs = list(config["lhs"])
+        rhs = list(config["rhs"])
+        super().__init__(universe, name, hasRegion=config.get("hasRegion", False), link=(),
+                         required=lhs + rhs, optional=(), doc=config["doc"])
+        self._lhs = DimensionSet(universe, lhs, optional=False)
+        self._rhs = DimensionSet(universe, rhs, optional=False)
+        # self._summarizes initialized later in DimensionMeta.fromConfig.
```
andy-slac: DimensionMeta -> DimensionGraph

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -0,0 +1,278 @@
...
+
+    @property
+    def lhs(self):
+        r"""The `Dimension`\s on the left hand side of the join
+        (`DimensionSet`).
+
+        Left vs. right is completely arbitrary; the terminology simply provides
+        an easy way to distiguish between the two sides.
```
andy-slac: distiguish misses 'n', same in next method.

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -0,0 +1,278 @@
...
+        """
+        return self._rhs
+
+    @property
+    def summarizes(self):
+        r"""A set of other `DimensionJoin`\s that provide more fine-grained
+        relationships than this one (`DimensionJoinSet`).
+
+        When a join "summarizes" another, it means the table for that join
+        could (at least conceptually) be defined as a aggregate view on the
```
andy-slac: an aggregate

python/lsst/daf/butler/core/dimensions/graph.py
```diff
@@ -0,0 +1,473 @@
...
+
+
+class DimensionConfig(ConfigSubset):
+    component = "dimensions"
+    requiredKeys = ("version", "units", "joins")
+    defaultConfigFile = "dimensions.yaml"
+
+
+class DimensionGraph:
+    r"""A automatically-expanded collection of both `Dimension`\s and
```
andy-slac: An

config/dimensions.yaml
```diff
@@ -1,6 +1,6 @@
-dataUnits:
+dimensions:
   version: 0
-  dimensions:
+  units:
```
andy-slac: Is `units` the right name here? Maybe call top-level key "graph" or "universe", and keep this as "dimensions"?

python/lsst/daf/butler/core/dimensions/graph.py
```diff
@@ -0,0 +1,473 @@
...
+        """The names of all fields that uniquely identify these dimensions in
+        a data ID dict (`frozenset` of `str`).
+        """
+        return self._dimensions.links
+
+    def __hash__(self):
+        return hash(self._dimensions)
+
+    def __eq__(self, other):
+        return self._dimensions == other
```
andy-slac, line 291: Does `DimensionSet` know how to compare itself with `DimensionGraph`?

python/lsst/daf/butler/core/dimensions/graph.py
```diff
@@ -0,0 +1,473 @@
...
+        return NotImplemented
+
+    def extract(self, elements, optional=False):
+        r"""Return a new graph containing the given elements.
+
+        Parameters
+        ----------
+        elements : iterable of `DimensionElement` or `str`
+            `Dimension`\s, `DimensionJoin`\s, or names thereof.  These must be
+            in ``self`` or ``self.joins()``.
```
andy-slac, line 379: There is a also support for `DimensionGraph` in the code 🙂

andy-slac (commented):
I went through maybe half of the diffs, will continue tomorrow, but here are the comments that I made so far.

python/lsst/daf/butler/core/dimensions/sets.py
```diff
@@ -0,0 +1,558 @@
...
+__all__ = ("DimensionSet", "DimensionNameSet", "toNameSet")
+
+from collections.abc import Set
+from collections import OrderedDict
+from .elements import DimensionElement
+
+
+def toNameSet(container):
+    """Transform an iterable of `DimensionElement` or names thereof into a
+    `DimensionNameSet`.
```
andy-slac, line 32: It also seems to accept an object with the `names` attribute, not just iterable?

python/lsst/daf/butler/core/dimensions/sets.py
```diff
@@ -0,0 +1,558 @@
...
+    immutable and hashable, and also provide named-method versions of most
+    operators.  Unlike Python sets, they are deterministically sorted.
+
+    `DimensionSet` comparison operators and named-method relation operations
+    accept other set-like objects and iterables containing either
+    `DimensionElement` instances or their string names; because
+    `DimensionElement`\s cannot be directly constructed, APIs that accept them
+    should generally accept a name as an alternative when the transformation
+    to a `DimensionElement` can be done internally.  Operators that return
+    new sets (`|`, `&`, `^`, and `^`) do require `DimensionSet` operands on
```
andy-slac: `^`, and `^`?

python/lsst/daf/butler/core/dimensions/sets.py
```diff
@@ -0,0 +1,558 @@
...
+    be passed to any of these. The `names` attribute can also be used to
+    obtain a `set`-like object containing those names.
+
+    `DimensionSet` instances cannot be constructed directly; they can only be
+    obtained (possibly indirectly) from a special "universe" `DimensionGraph`
+    loaded from configuration.
+
+    Parameters
+    ----------
+    univsere : `DimensionGraph`
```
andy-slac: universe

python/lsst/daf/butler/core/dimensions/sets.py
```diff
@@ -0,0 +1,558 @@
...
+                    elem = self._universe.elements[elem]
+                yield (elem.name, elem)
+                if expand:
+                    yield from toPairs(elem.dependencies(optional=optional))
+
+        names = dict(toPairs(elements))
+        if names:
+            for candidate in self._universe.elements:
+                if names.pop(candidate.name, None):
+                    self._elements[candidate.name] = candidate
```
andy-slac: So the values in the `names` dict are not used at all? You could build a set of names instead of dict for small optimization.

python/lsst/daf/butler/core/dimensions/sets.py
```diff
@@ -0,0 +1,558 @@
...
+        return "{{{}}}".format(", ".join(self._elements.keys()))
+
+    def __repr__(self):
+        return f"DimensionSet({self})"
+
+    def __contains__(self, key):
+        try:
+            return key.name in self._elements
+        except AttributeError:
+            return key in self._elements
```
andy-slac: I'm not fond of using exception handling for regular control flow. Something like:

    key = getattr(key, "name", key)
    return key in self._elements

would avoid raising/catching exceptions. Same for other methods below.

python/lsst/daf/butler/core/dimensions/sets.py
```diff
@@ -0,0 +1,558 @@
...
+            return self._elements.get(key.name, default)
+        except AttributeError:
+            return self._elements.get(key, default)
+
+    @property
+    def names(self):
+        """The names of all elements (`set`-like, immutable).
+
+        The order of the names is consistent with the iteration order of the
+        `DimensionSet` itself.
```
andy-slac: Is there a guarantee in OrderedDict that iteration over result of `keys()` happens in the same order as `iter()`?

python/lsst/daf/butler/core/dimensions/sets.py
andy-slac: I guess that critical part here is that insertion must happen in the same order as order of dimensions in the universe? Can you add a comment to this loop with a warning to never change the order?

python/lsst/daf/butler/core/dimensions/sets.py
```diff
@@ -0,0 +1,558 @@
...
+
+    Parameters
+    ----------
+    names : iterable of `str`
+        The names of elements to conceptually include in the set.
+    """
+
+    def __init__(self, names):
+        if not isinstance(names, Set):
+            names = frozenset(names)
```
andy-slac, line 436: Maybe do this unconditionally? If someone passes `set` as an argument `__hash__` will be broken.

python/lsst/daf/butler/core/dimensions/sets.py
```diff
@@ -0,0 +1,558 @@
...
+
+    def __xor__(self, other):
+        if isinstance(other, DimensionNameSet):
+            return self.symmetric_difference(other)
+        return NotImplemented
+
+    def __sub__(self, other):
+        if isinstance(other, DimensionNameSet):
+            return self.difference(other)
+        return NotImplemented
```
andy-slac: General comment to this module - there is plenty of code duplication between DimensionSet and DimensionNameSet. Could DimensionSet be just a DimensionNameSet plus universe?

python/lsst/daf/butler/core/registry.py
```diff
@@ -798,3 +822,142 @@ def selectDataUnits(self, originInfo, expression, neededDatasetTypes, futureData
...
+            # By the time we get here 'dimension' and 'dataId.dimension'
+            # correspond to the same thing, but:
+            #  - 'dimension' might be a string or a `DimensionElement`
+            #  - 'dataId.dimension' is definitely a `Dimension`
+            #  - 'dataId.dimension' may not be None even if 'dimension' is
+            if dimension is not None and not isinstance(metadata, Mapping):
+                # If a single dimension was passed explicitly, permit 'metadata' to
+                # be a sequence corresponding to just that dimension.
+                result = self._queryMetadata(dataId.dimension, dataId, metadata)
+                DataId.entries[dataId.dimension].update(result)
```
andy-slac: `DataId` should be `dataId`?

python/lsst/daf/butler/core/registry.py
```diff
@@ -798,3 +822,142 @@ def selectDataUnits(self, originInfo, expression, neededDatasetTypes, futureData
...
+            else:
+                for element, columns in metadata.items():
+                    subDataId = DataId(dataId, dimension=element)
+                    result = self._queryMetadata(element, subDataId, columns)
+                    DataId.entries[element].update(result)
```
andy-slac: Here too

python/lsst/daf/butler/core/registry.py
```diff
@@ -798,3 +822,142 @@ def selectDataUnits(self, originInfo, expression, neededDatasetTypes, futureData
...
+        ------
+        LookupError
+            Raised if no entry for the given data ID exists.
+        NotImplementedError
+            Raised if `limited` is `True`.
+        """
+        if dimension.name == "SkyPix":
+            # SkyPix is special; we always obtain those regions from
+            # self.pixelization
+            return self.pixelization.pixel(dataId["skypix"])
```
andy-slac, line 928: skypix/SkyPix magical names, maybe define constants for them or do something fancy in configuration (in future ticket)?

python/lsst/daf/butler/core/utils.py
```diff
@@ -357,3 +252,58 @@ def stripIfNotNone(s):
...
+        class Hidden(metaclass=PrivateConstructorMeta):
+
+            def __init__(self, a, b):
+                self.a = a
+                self.b = b
+
+    This doesn't work:
+
+        >>> instance = Hidden(a=1, b="two")
+        NotImplementedError: Foo objects cannot be constructed directly.
```
andy-slac: NotImplementedError -> TypeError
Foo -> Hidden

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -19,74 +19,23 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-__all__ = ("SqlPreFlight")
+__all__ = ("SqlPreFlight",)
```
andy-slac, line 22: I wonder how did that work before 🤔 And we probably don't need `__all__` here at all.

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -95,33 +44,33 @@ class SqlPreFlight:
...
+    dimensions : `DimensionGraph`
+        "Universe" graph containing all dimensions and joins.
     connection : `sqlalchmey.Connection`
         Connection to use for database access.
     """
-    def __init__(self, schema, dataUnits, connection):
+    def __init__(self, schema, dimensions, connection):
         self._schema = schema
-        self._dataUnits = dataUnits
+        self._dimensions = dimensions
```
andy-slac: May be better to call it _universe if this is supposed to be whole thing?

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -130,180 +79,161 @@ def _joinOnForeignKey(self, fromClause, dataUnit, otherDataUnits):
...
+            return self._schema.tables[dimension.name]
         else:
             joinOn = []
-            for otherUnit in otherDataUnits:
-                primaryKeyColumns = {name: self._schema.tables[otherUnit.name].c[name]
-                                     for name in otherUnit.primaryKey}
+            for otherDimension in otherDimensions:
+                primaryKeyColumns = {name: self._schema.tables[otherDimension.name].c[name]
+                                     for name in otherDimension.primaryKey
+                                     if name in self._schema.tables[otherDimension.name].c}
```
andy-slac: This `if` is new, is there a case with new code when PK is not in table schema?

python/lsst/daf/butler/core/dimensions/dataId.py
TallJimbo: I had seen the form with the space as well, but found that Sphinx renders things just fine without it; does that change your opinion at all?

As for adding "instances", I think that's a good idea when it adds clarity, but adding it just to avoid a markup issue seems wrong; from the reader's perspective it's just noise-words that don't add any real information.

python/lsst/daf/butler/core/dimensions/graph.py
TallJimbo: That's what the type here reflects, actually; each additional argument is itself an iterable.

Side question: should I use `others` or `*others` here?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -57,8 +56,6 @@ class SqlRegistry(Registry):
         Load configuration
     schemaConfig : `SchemaConfig` or `str`
         Definition of the schema to use.
-    dataUnitConfig : `DataUnitConfig` or `str`
-        Definition of the DataUnitRegistry to use.
```
andy-slac, line 61: doc for `dimensionConfig`?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -365,55 +324,27 @@ def getDatasetType(self, name):
             raise KeyError("Could not find entry for datasetType {}".format(name))
 
         storageClass = self.storageClasses.getStorageClass(result["storage_class"])
-        # Get DataUnits (if any) from DatasetTypeUnits table
-        result = self._connection.execute(select([datasetTypeUnitsTable.c.unit_name]).where(
-            datasetTypeUnitsTable.c.dataset_type_name == name)).fetchall()
-        dataUnits = (r[0] for r in result) if result else ()
+        # Get Dinmensions (if any) from DatasetTypeDimensions table
```
andy-slac: Dinmensions

python/lsst/daf/butler/core/dimensions/dataId.py
TallJimbo: Good catch; we should permit `dimension` to be inferred even if `dimensions` is inferred from links.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -989,235 +922,113 @@ def markInputUsed(self, quantum, ref):
...
-                                               neededDatasetTypes,
-                                               futureDatasetTypes)
+    def _queryMetadata(self, element, dataId, columns):
+        # Docstring inherited from Registry._queryMetadata.
+        table = self._schema.tables[element.name]
+        cols = [table.c[col] for col in columns]
+        row = self._connection.execute(
+            select(cols)
+            .where(
+                and_(table.c[name] == dataId[name] for name, value in dataId.items())
```
andy-slac: `table.c[name] == value`?

tests/test_dimensions.py
```diff
@@ -0,0 +1,327 @@
...
+    def checkGraphInvariants(self, graph):
+        """Run tests on DimensionGraph that should pass for any instance.
+        """
+        self.checkSetInvariants(graph.asSet)
+        self.assertLessEqual(graph, self.universe)
+        for dim in graph:
+            self.assertIn(dim, graph)
+            self.assertIn(dim.name, graph)
+            self.assertEqual(graph[dim.name], dim)
+            self.assertEqual(graph.get(dim.name), dim)
```
andy-slac: Should these two tests be `assertIs`?

andy-slac (approved):
Looks OK, few minor comments.

python/lsst/daf/butler/core/dimensions/dataId.py
TallJimbo: Thinking about it a bit more, I think the concept of having a leaf dimension probably belongs on `DimensionGraph` itself, and it may make sense to generalize to multiple leaf dimensions.  I'll give that a try.

python/lsst/daf/butler/core/dimensions/dataId.py
TallJimbo: Good point; probably, so `frozenset` would be better here.

python/lsst/daf/butler/core/dimensions/elements.py
TallJimbo: Yes, though I think we (perhaps unfortunately) actually have baked in that assumption to a certain degree - at least at the level of that these "link" columns must be unique identifiers and used frequently as foreign keys, and once you've gone that far I'd be surprised if there was much value to defining primary keys separately.  That said, it would be good to try to move away from the rather database-specific term `primary key` in favor of `links` or something else as much as possible.

python/lsst/daf/butler/core/dimensions/elements.py
TallJimbo: Yes.

config/dimensions.yaml
TallJimbo: I really struggled with this.  It really ought to be "dimensions" for both the outer (as the name of the system) and the inner key (as the name of the central concept in that system), but that confuses some of our "guess what level this config is at" logic.  Without "Dimension", "graph" or "universe" doesn't make sense at the top level, I think.

Maybe I should capitalize the second-level "Dimensions", and also change "joins" to "DimensionJoins", to exactly match the classes, and rely on capitalization to avoid confusing the config code?

python/lsst/daf/butler/core/dimensions/graph.py
TallJimbo: Yes; it calls `toNameSet` on the other operand, and hence it's comparable with anything that accepts.

python/lsst/daf/butler/core/dimensions/graph.py
TallJimbo: Yes, but `DimensionGraph` *is* an iterable of `DimensionElement`.

python/lsst/daf/butler/core/dimensions/sets.py
TallJimbo: Should have been `^` and `-`.

config/dimensions.yaml
andy-slac: Call top level `DimensionGraph`/`DimensionUniverse`? Case-only difference in names is generally considered questionable style.

python/lsst/daf/butler/core/dimensions/sets.py
TallJimbo: Yes, it's all about the insertion order, and of course this is only a reasonable approach to that because we know the set of Dimensions is going to be small and hence making everything (at least) O(N) in the number of dimensions in universe is not a problem.  I'll add a comment covering all of that.

python/lsst/daf/butler/core/dimensions/sets.py
TallJimbo: I certainly strongly agree with that principle in C++; in Python I also prefer to avoid exceptions in control-flow but don't think it's worth much effort to avoid them (i.e. if the alternative involves a big `if` or a lot of otherwise-unnecessary variables).  But your alternative in this case is clearly better; will do.

python/lsst/daf/butler/core/dimensions/sets.py
TallJimbo: Yes (at least that's strongly implied by the examples in the Python docs, if not clearly stated in the text).

python/lsst/daf/butler/core/dimensions/sets.py
TallJimbo: The conditional here is to avoid transforming a `OrderedDict` keys-view (which is a `Set`) into a `frozenset`, as that transformation discards the ordering.  But maybe I should also check that's not a `MutableSet` (which `OrderedDict` keys-view is not).  I'll do that and add some comments.

python/lsst/daf/butler/core/dimensions/sets.py
TallJimbo: I think there would need to be a third (abstract) base class for both, because they don't actually share any state attributes.  I'll at least take a look at that and see if it improves things; overall I'm not *too* concerned about duplication here because it's mostly duplication of trivial things, not tricky logic.

python/lsst/daf/butler/core/registry.py
TallJimbo: I've added a comment to DM-13990 to handle this on that ticket.

python/lsst/daf/butler/registries/sqlPreFlight.py
TallJimbo: I think we must have just never used `from sqlPreFlight import *` anywhere, so it was never used.  In any case, now that it's fixed I think it's probably better to have it than not, if just to help Sphinx out (I think it relies on `__all__` at some level to know what should be included in public APIs).

python/lsst/daf/butler/registries/sqlPreFlight.py
TallJimbo: Fine with me - I mostly used `dimensions` in the Registry context where "dimension" isn't implied, and "universe" only in the "dimensions" subpackage.  But unlike Registries, Preflight actually has a prominent "dimensions" local variable that *isn't* the universe, so universe might be clearer here.

config/dimensions.yaml
TallJimbo: I'm going to look into merging the contents of `units` and `joins` into `elements`, and see if that's at least no worse in terms of constructing the universe `DimensionGraph`.  It feels a little ugly to make that kind of change to just work around a naming problem, but given that `Dimension` and `DimensionJoin` now have more in common, it might be better overall (and if it isn't I'll go back to thinking more about names).

python/lsst/daf/butler/registries/sqlPreFlight.py
TallJimbo: I've removed it.  In a previous version `.primaryKey` was being expanded to include links from optional dependencies, but I've now fixed that.

python/lsst/daf/butler/core/dimensions/sets.py
TallJimbo: I've added an abstract base class that defines all of the comparison operators and standard set-creation operations in terms of an abstract `.names` property; that works quite well.

python/lsst/daf/butler/core/dimensions/dataId.py
TallJimbo: I've added a `.leaves` property to `DimensionGraph`, and that (slightly) simplifies the `DataId` and probably helps with some `SqlRegistry` code, too.

python/lsst/daf/butler/core/dimensions/dataId.py
czwa: `The keys`?

config/dimensions.yaml
TallJimbo: Success!  Moving everything under an `elements` heading does actually simplify the graph-building logic, so it's worth keeping, and this naming issue is no more.

PR #106, opened 2018-12-13
DM-15503: fix raw data and cameraGeom in gen2->gen3 conversion

TallJimbo:


config/gen2convert.yaml
```diff
@@ -37,18 +37,21 @@ collections:
     -
       pattern: "^(.+)/rerun/"
       repl: "shared/"
+    -
+      pattern: "^(.+)/CALIB"
+      repl : "calib"
```
czwa, line 42: I'm running tests to see if this is the problem I think it is, but in DM-16467, I've had to change the collection of calibration data to match the input data to get things to work.  I don't know if this is still the case now (hence the testing), but has there been a discussion of this: datasets that should be members of all collections?

config/gen2convert.yaml
```diff
@@ -37,18 +37,21 @@ collections:
...
     ref_cat: ref/{name}
     ref_cat_config: ref/{name}
-    transmission_filter: transmission/{instrument}
-    transmission_sensor: transmission/{instrument}
-    transmission_optics: transmission/{instrument}
-    transmission_atmosphere: transmission/{instrument}
+    transmission_filter: calib
+    transmission_sensor: calib
+    transmission_optics: calib
+    transmission_atmosphere: calib
```
czwa, line 54: This is only needed for the gen2 conversion, correct?  Specifically, once fully switched to gen3, all of this information will be in the butler repository, and will not need yaml files to special case things?

python/lsst/daf/butler/gen2convert/writer.py
```diff
@@ -198,14 +201,14 @@ def run(self, registry, datastore):
         (we merely check that the needed Instruments, Detectors, and PhysicalFilters
```
czwa: This statement is no longer true, I think.

czwa (approved):
I don't think my comments reveal major issues.  The calib collection issue has already existed, so this isn't making it worse in any way.

config/gen2convert.yaml
TallJimbo: The vision was instead to permit multiple input collections, which would be treated as a priority-ordered search path.  It looks like that's doable with `--input collection1,collection2`, though I'm not completely certain it's been implemented.  Have you tried that?  Or is there a usability problem with that approach that you've identified?

config/gen2convert.yaml
TallJimbo: Correct; the idea is that eventually obs_ packages will provide some kind of script that dumps these datasets into the right collection in a butler repository, including validity ranges for all of them.  I don't know exactly what those scripts (or their source data) will look like yet.

config/gen2convert.yaml
czwa: I haven't tried using the comma separated version, but if that's the plan, then I'm happy to use that.  I need to rebuild my test environment to confirm this will work with the ISR pipelineTask.

PR #111, opened 2018-12-18
DM-16869: Fix DimensionSetBase hash method

andy-slac:
Hash has to return the same value for the same set of names so we need
to convert set of names to the same hashable type (frozenset).


TallJimbo (approved):
Looks good, and sorry for the trouble.

PR #113, opened 2018-12-27
DM-15675: improve data ID expansion and filename templates

TallJimbo:


python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -347,6 +347,22 @@ def addDataset(self, datasetType, dataId, run, producer=None, recursive=False, *
...
+        for link in dataId.dimensions.optionals.links:
+            if link in links:
+                continue
+            for dim in dataId.dimensions.withLink(link):
+                if dim not in dataId.dimensions:
+                    continue
+                try:
+                    links[link] = dataId.entries[dim][link]
+                except KeyError:
+                    raise LookupError(f"'{link}' not provided in data ID, but Registry is limited.")
```
natelust: This could be a more helpful error message

python/lsst/daf/butler/core/dimensions/dataId.py
```diff
@@ -205,7 +206,7 @@ def __init__(self, dataId=None, *, dimensions=None, dimension=None, universe=Non
             for element, subdict in entries.items():
                 self.entries[element].update(subdict)
 
-        missing = self.dimensions.links - self._links.keys()
+        missing = self.dimensions.links() - self._links.keys()
```
natelust: I think it is a slight shame that links is a method on dimensions in one case and _links is a dict on self, but I am not sure I have something much better to suggests, except perhaps longer names?

natelust (approved):

python/lsst/daf/butler/registries/sqlRegistry.py
TallJimbo: This particular error message was removed in a later commit on this branch, but I've improved the error message that occurs for the same condition there.

python/lsst/daf/butler/core/dimensions/dataId.py
TallJimbo: `self.dimensions.links()` is also a `frozenset` while `self._links` is a `dict`.  I'll rename the latter to `_linkValues`.

PR #114, opened 2019-01-03
DM-17048: Fix handling of ref_cat in gen2convert.

TallJimbo:


python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -222,15 +258,20 @@ def selectDimensions(self, originInfo, expression, neededDatasetTypes, futureDat
                 regionHolders.append(regionHolder)
 
                 # We also have to include regions from each side of the join
-                # into resultset so that we can filter-out non-overlapping
+                # into result set so that we can filter-out non-overlapping
                 # regions.
-                regionColumnIndices[regionHolder.name] = len(selectColumns)
-                regionColumn = self.registry._schema.tables[regionHolder.name].c.region
-                selectColumns.append(regionColumn)
+                if regionHolder.name not in regionColumnIndices:
```
natelust, line 268: Why is this now in a conditional block?

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -489,6 +530,10 @@ def _convertResultRows(self, rowIter, dimensions, linkColumnIndices, regionColum
             # filtered if any of two regions are disjoint.
             disjoint = False
             regions = {holder: Region.decode(row[col]) for holder, col in regionColumnIndices.items()}
+            if "skypix" in linkColumnIndices:
```
natelust, line 542: I know this is something we worked on together but you may want to add a comment talking about why this is needed and or mention a ticket that might fix this.

natelust (approved):

python/lsst/daf/butler/registries/sqlPreFlight.py
TallJimbo: To account for the possibility that the same dimension appears multiple times in this loop, as a part of multiple different join tables (e.g. visit-skypix and visit-tract).  Prior to `MeasureMergedCoaddSourcesTask`, we'd only ever had one spatial join table in a preflight query, and that hid this bug.  Will add comment to that effect.

python/lsst/daf/butler/registries/sqlPreFlight.py
TallJimbo: Will add a comment.  While it might be nice to try to unify the special-casing of SkyPix, some special-casing is always going to have to be there.

PR #116, opened 2019-01-04
DM-14821: mechanisms for packing (some) data IDs into integers

TallJimbo:


config/schema.yaml
```diff
@@ -450,6 +450,24 @@ schema:
...
           Globally unique string indentifier for this Instrument.
+      -
+        name: visit_max
+        type: int
+        doc: >
+          Maximum value for the 'visit' field for Visits associated with
+          this instrument (exclusive).
+      -
+        name: exposure_max
+        type: int
```
natelust, line 461: How are you sure there is a max like this (and above)? Is it based on the fact that some instruments use 16,32,64,etc bit numbers for identifiers? If so what if an instrument uses say 128, and python int is 64? Or is this int actually a hint to sql and nothing more, and they just have a "number" field. What if a telescope does not use numbers to increment, like some sort of lexographic numbering based on some datestring?

python/lsst/daf/butler/core/dimensions/dataId.py
```diff
@@ -19,12 +19,155 @@
...
+                self._keys = keys
+            elif isinstance(keys, DimensionGraph):
+                self._keys = keys.toSet()
+            else:
+                self._keys = DimensionSet(universe, keys)
+        else:
+            self._keys = DimensionSet(universe, byName.keys())
+
+        if where is None:
+            where = lambda x: True  # noqa
```
natelust: Maybe use _ in the lambda to emphasize that this is never going to be used

python/lsst/daf/butler/core/dimensions/dataId.py
```diff
@@ -19,12 +19,155 @@
...
+    def __len__(self):
+        return len(self._keys)
+
+    def __eq__(self, other):
+        try:
+            return self._dict == other._dict
+        except AttributeError:
+            return NotImplemented
+
+    def updateValues(self, other):
```
natelust, line 146: So I think it would be good to talk about interior v exterior mutability at some point. If someone reads the documentation for the class, they may take this class as "immutable" but this method implies that it is mutable (though I know it is just interior mutability) however these are not normally python concepts, and someone might be surprised by calling items, getting a list of the items, and then later find that that list does not match what is in the DimensionsKeyDict. I dont' know that we want to disallow this behavior, but it can be surprising, so we should think if this is the ideal structure, or what kind of documentation on the promises this class makes needs to be written.

python/lsst/daf/butler/core/registry.py
```diff
@@ -888,13 +890,11 @@ def expandDataId(self, dataId=None, *, dimension=None, metadata=None, region=Fal
                 # If a single dimension was passed explicitly, permit
                 # 'metadata' to be a sequence corresponding to just that
                 # dimension (by normalizing it into a dict-of-sets now).
```
natelust: I think this comment needs updated

python/lsst/daf/butler/core/dataIdPacker.py
```diff
@@ -0,0 +1,464 @@
...
+       as well as additional keyword arguments to pass to its constructor.
+
+     - Regular class construction sequence is invoked, passing the dimensions
+       from the configuration, a `DataId` expanded to include the metadata
+       identified in the first stage, and the additional keyword arguments
+       obtained from the first stage.
+    """
+
+    # Override __new__ so subclasses aren't forced to call super().__init__
+    # and have to support at least the standard signature.
```
natelust: This message could be clearer

python/lsst/daf/butler/core/dataIdPacker.py
```diff
@@ -0,0 +1,464 @@
...
+        Once the packer class is imported, we can retreive the dictionary of
+        metadata entries it needs for instantiation, and hence ensure those
+        entries are always loaded whenever data IDs are expanded.
+        """
+        # Standardize dimension input
+        dimensions = toNameSet(dimensions)
+        matched = []
+        for given, loaders in self._loaders.items():
+            if given.issubset(dimensions):
+                for name, loader in loaders.items():
```
natelust: it is confusing to have loaders be part of a tuple on items in the dictionary named _loaders

natelust (changes requested):
We talked about moving the packer object lookups onto the registry, and have two different calls to either pick if the given dataId is already equal to the required, and one to just get the packer.

I would like to see that work before this merges.

python/lsst/daf/butler/core/dimensions/dataId.py
timj: I thought we frowned upon lambda in this context and preferred:
```python
def where():
    return True
```
?

timj: PS `noqa` is meant to include the code that you are ignoring.

config/schema.yaml
timj: The metadata translator constructs the exposure_id as an integer regardless of how the instrument is defining it in the headers. Python integers can be any size.

python/lsst/daf/butler/core/dimensions/dataId.py
TallJimbo: Agree with `_`.

I generally disagree with PEP8 on local `def` being better than assigning a lambda to a variable (I don't know of any other language with that preference).  Obviously it's a bad idea to pick a fight with PEP8 in general, but relying on a local `def` to shadow a kwarg seemed like a particularly bad idea.  I'll certainly look up the code for `noqa`, though.

config/schema.yaml
TallJimbo: I'm basically reasoning from the other direction: I need an `exposure_max` or something equivalent to implement the ID packers we need to support (Gen2 uses a number of bits, which I think just obfuscates the fact that it defines a maximum exposure number).  So I'm essentially putting it on the `Instrument` subclass (and hence the obs package) to define something and stick to it.  Given the size of even just an int32 and the number of exposures that can practically taken with real instruments of the sort we're trying to support, I don't think that should be a problem.

python/lsst/daf/butler/core/dimensions/dataId.py
TallJimbo: I had written a brief blurb on this in the `Notes` section of the class doc; I'll expand that.  It does turn out to be a very convenient pattern for the dimension system, in which there are frequently contexts in which the set of dimensions is fixed, and we already have an immutable `DimensionSet` class to use as the return value for `keys()`.

TallJimbo:
@natelust, I've made the changes we discussed, and all of the line comments I didn't respond to have become obsolete as a result.  Please take another look when you get a chance.

natelust (approved):

PR #117, opened 2019-01-15
DM-16227: Move collection integrity constraint into DB and fix transactions

TallJimbo:


python/lsst/daf/butler/core/dimensions/dataId.py
```diff
@@ -511,6 +512,26 @@ def __eq__(self, other):
...
+            argument to update the hash.
+        """
+        for k, v in self.items():
+            message.update(k.encode('utf8'))
+            if isinstance(v, int):
+                message.update(v.to_bytes(64, byteorder='little'))
+            elif isinstance(v, str):
+                message.update(v.encode('utf8'))
+            elif isinstance(v, datetime.datetime):
+                message.update(v.isoformat().encode('utf8'))
```
andy-slac: Does `isoformat` make unambiguous representation (I'm thinking about timezones)? Would it be easier to just take number of seconds in UTC?

python/lsst/daf/butler/core/dimensions/dataId.py
```diff
@@ -511,6 +512,26 @@ def __eq__(self, other):
...
+        """Add this data ID to a secure hash.
+
+        Parameters
+        ----------
+        `message` : `hashlib` message instance
+            Object with an ``update`` method that takes a single `bytes`
+            argument to update the hash.
+        """
+        for k, v in self.items():
+            message.update(k.encode('utf8'))
```
andy-slac: `utf-8` is the default for `encode()`, maybe just drop it rather than repeat everywhere?

python/lsst/daf/butler/core/registry.py
```diff
@@ -409,16 +409,27 @@ def attachComponent(self, name, parent, component):
     @abstractmethod
     @transactional
     def associate(self, collection, refs):
-        """Add existing Datasets to a collection, possibly creating the
-        collection in the process.
+        """Add existing Datasets to a collection, implicitly creating the
+        collection if it does not already exist.
+
+        If a DatasetRef with the same exact `dataset_id`` is already in a
```
andy-slac: One more backtick on the left

python/lsst/daf/butler/core/schema.py
```diff
@@ -19,17 +19,43 @@
...
+        length = 4*ceil(nbytes/3)
+        TypeDecorator.__init__(self, *args, length=length, **kwds)
+        self.nbytes = nbytes
+
+    def process_bind_param(self, value, dialect):
+        if value is None:
+            return None
+        if not isinstance(value, bytes):
+            raise TypeError(f"Base64Bytes fields require 'bytes' values; got {value}")
+        return b64encode(value).decode('utf8')
```
andy-slac: Nitpick - base64 is guaranteed to make `ascii` bytes

config/schema.yaml
```diff
@@ -13,6 +13,7 @@ schema:
       -
         name: dataset_type_name
         type: string
+        length: 128
```
andy-slac, line 16: For future extension I'd probably suggest doing something like:

    type: string
    typeKwargs: {length: 128}

but with just two of these new keywords this maybe an overkill. Can wait until we add third 🙂

python/lsst/daf/butler/core/utils.py
```diff
@@ -172,18 +172,12 @@ def __call__(cls):  # noqa N805
 def transactional(func):
     """Decorator that wraps a method and makes it transactional.
 
-    This also adds an optional ``transactional`` keyword argument to the
-    decorated function, which can be set to `False` to disable method-level
-    transactions.  This can be useful when the method is being called within
-    a higher-level transaction block but exceptions it raises will be
-    caught before they propagate through the higher-level context manager.
+    This depends on the class also defining a `transactional` method
```
andy-slac: transactional -> transaction?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -370,23 +416,23 @@ def addDataset(self, datasetType, dataId, run, producer=None, recursive=False, *
...
                                                                        run_id=run.id,
+                                                                       dataset_ref_hash=datasetRef.hash,
                                                                        quantum_id=None,
                                                                        **links))
-        datasetRef = DatasetRef(datasetType=datasetType, dataId=dataId, id=result.inserted_primary_key[0],
-                                run=run)
+        datasetRef._id = result.inserted_primary_key[0]
         # A dataset is always associated with its Run collection
-        self.associate(run.collection, [datasetRef, ], transactional=False)
+        self.associate(run.collection, [datasetRef, ])
```
andy-slac, line 426: That comma looks suspiciously lonely like if something is supposed to follow it but was deleted accidentally 🙂

python/lsst/daf/butler/registries/sqliteRegistry.py
```diff
@@ -78,3 +89,24 @@ def __init__(self, registryConfig, schemaConfig, dimensionConfig, create=False):
...
+            'INSERT INTO DatasetCollection ("dataset_id", "dataset_ref_hash", "collection") '
+            'VALUES (:dataset_id, :dataset_ref_hash, :collection) '
+            'ON CONFLICT ("dataset_id", "collection") DO NOTHING'
+        )
+        try:
+            self._connection.execute(query,
+                                     [{"dataset_id": ref.id, "dataset_ref_hash": ref.hash,
+                                       "collection": collection} for ref in refs])
+        except IntegrityError as err:
+            raise ValueError(str(err))
```
andy-slac: That `raise` would cause the `During handling of the above exception, another exception occurred:` which I find confusing. It may be better to do something like:

    raise ValueError("INSERT failed") from err

to generate `The above exception was the direct cause of the following exception:` message which I think reflects better what is going on.

config/schema.yaml
```diff
@@ -31,16 +32,27 @@ schema:
           The id of the quantum that produced this dataset, providing access
           to fine-grained provenance information.
           may be null for datasets not produced by running a supertask.
+      -
+        name: dataset_ref_hash
+        type: hash
+        nbytes: 32
+        nullable: false
+        doc: >
+          Secture hash of the data ID (i.e. dimension link values) and
```
andy-slac: Secure

config/schema.yaml
```diff
@@ -495,19 +533,22 @@ schema:
...
         name: instrument
         type: string
+        length: 8
         primary_key: true
         doc: >
           Name of the Instrument with which this filter is associated.
       -
         name: physical_filter
         type: string
+        length: 8
```
andy-slac: Is 8 characters enough for everyone?

python/lsst/daf/butler/core/registry.py
```diff
@@ -409,16 +409,27 @@ def attachComponent(self, name, parent, component):
...
+        collection nothing is changed. If a `DatasetRef` with the same
+        `DatasetType1` and dimension values but with different ``dataset_id``
```
andy-slac, line 417: DatasetType*1*?

python/lsst/daf/butler/core/schema.py
```diff
@@ -19,17 +19,43 @@
...
 
+class Base64Bytes(TypeDecorator):
+    """A SQLAlchemy custom type that maps Python `bytes` to a base64-encoded
+    `sqlalchemy.String`.
+    """
+
+    impl = String
+
+    def __init__(self, nbytes, *args, **kwds):
+        length = 4*ceil(nbytes/3)
```
andy-slac, line 44: Should it divide by 4?

python/lsst/daf/butler/core/schema.py
```diff
@@ -19,17 +19,43 @@
...
+        TypeDecorator.__init__(self, *args, length=length, **kwds)
+        self.nbytes = nbytes
+
+    def process_bind_param(self, value, dialect):
+        if value is None:
+            return None
+        if not isinstance(value, bytes):
+            raise TypeError(f"Base64Bytes fields require 'bytes' values; got {value}")
```
andy-slac: Maybe `type(value)`?

andy-slac (approved):

python/lsst/daf/butler/core/dimensions/dataId.py
TallJimbo: I *think* it does, but I'm not certain, and I didn't see an easy way to get number of seconds in UTC from Python `datetime`.  In any case, I expect this all to get fixed in [DM-15890](https://jira.lsstcorp.org/browse/DM-15890), which will switch all of the datetimes we use in the database to TAI MJD to avoid these kinds of problems entirely.

config/schema.yaml
TallJimbo: I'd prefer to just get the schema definition out of YAML ([DM-17154](https://jira.lsstcorp.org/browse/DM-17154)) and into direct SQLAlchemy Python code; we're on the road to trying to map the full complexity of SQLAlchemy schema definition to a home-brew YAML format, and I think we need to get off that road, especially before getting serious about an Oracle implementation.

python/lsst/daf/butler/registries/sqlRegistry.py
TallJimbo: Just arguably weird style.  The statement is what it's supposed to be.

config/schema.yaml
TallJimbo: Most filter names are single characters, and physical filters may contain instrument prefixes or version suffixes, that'll still be pretty small. That said, maybe the size of the instrument field + 8 would be better (i.e. 16).

python/lsst/daf/butler/core/schema.py
TallJimbo: I don't think so: https://stackoverflow.com/questions/13378815/base64-length-calculation

python/lsst/daf/butler/core/schema.py
TallJimbo: I'll include both the value and its type.

python/lsst/daf/butler/core/schema.py
andy-slac: Ok, I was confused by what `nbytes` mean - I though it is stored data size. Never mind.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -523,81 +523,39 @@ def associate(self, collection, refs):
...
+                                           "collection": collection} for ref in refs])
+            elif row.dataset_id != ref.id:
+                # A different Dataset with this DatasetType and Data ID already
+                # exist in this collection.
+                raise ValueError(
+                    "A dataset of type {} with id: {} already exists in collection {}".format(
+                        ref.datasetType, ref.dataId, collection
+                    )
+                )
+            # If the same Dataset is already in this collection, do nothing.
```
andy-slac, line 559: I suppose this (select+check+insert) will work OK for SQLite because SQLite locks the whole database for the duration of transaction. For other backends which support concurrent transactions I think there is a race condition - if two clients do `select` at the same instant they both get empty set and will try to insert (one of them will fail).
I think more or less portable strategy for this kind of update is to just try to insert+commit and analyse DuplicateKey error that could happen on commit.
Some backends support `select ... for update` for row-level locking (and index gap locking) but it also needs special care so I would not recommend that as portable fix.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -523,81 +523,39 @@ def associate(self, collection, refs):
...
+            row = self._connection.execute(query, hash=ref.hash).fetchone()
+            if row is None:
+                # No Dataset with this DatasetType and Data ID in collection;
+                # insert it now.
+                self._connection.execute(datasetCollectionTable.insert(),
+                                         [{"dataset_id": ref.id, "dataset_ref_hash": ref.hash,
+                                           "collection": collection} for ref in refs])
+            elif row.dataset_id != ref.id:
+                # A different Dataset with this DatasetType and Data ID already
+                # exist in this collection.
```
andy-slac: exists

andy-slac (approved):
I think latest `associate()` update should be OK for sqlite, for non-sqlite backends we probably need different strategy.

python/lsst/daf/butler/registries/sqlRegistry.py
TallJimbo: Interesting - I think this indicates a fundamental misunderstanding by me on what transactions guarantee in general, which means I need to spend some time thinking about what those guarantees actually are.

I don't think "analyzing the `DuplicateKey` error" approach can really be done portably, either, because  we don't have any guarantee about what kind of information the error message might contain (and we need to know which constraint failed), so ultimately I think this (and several other `Registry` methods) need to be left pure-abstract in `SqlRegistry` so they must be implemented by DB-specific derived classes.

As I said on Jira, I think it's best not to do that now because (for now) a solution that's subject to race conditions is better than no solution at all.  I'll be sure to time the future changes carefully with NCSA to make sure we don't have a regression in Oracle support at an awkward time.

andy-slac: By analyzing I meant that you would need to select from database at that point, after failure. Relying on exception content is not possible of course.

PR #118, opened 2019-01-18
Convert isrTask to valid pipelineTask.

czwa:
Add additional storage classes needed for ISR products.  Fix
pickleFormatter issue that prevented python2 pickle products from
being read.

### DIFF
--- a/config/storageClasses.yaml
+++ b/config/storageClasses.yaml
@@ -122,3 +122,7 @@ storageClasses:
     pytype: lsst.pex.config.Config
   Packages:
     pytype: lsst.base.Packages
+  NumpyArray:
+    pytype: numpy.ndarray
+  Thumbnail:
+    pytype: numpy.ndarray


python/lsst/daf/butler/formatters/pickleFormatter.py
```diff
@@ -55,7 +55,7 @@ def _readFile(self, path, pytype=None):
         """
         try:
             with open(path, "rb") as fd:
-                data = pickle.load(fd)
+                data = pickle.load(fd, encoding='latin1')
```
timj: I thought this was only needed to read old pickle files written in python 2?

czwa: The obs_subaru brighter fatter kernel is a python2 pickle.

timj: I'd rather we fixed that.

timj: If you can't do that I'd rather we had a separate formatter for reading legacy files.

SimonKrughoff: Is it possible to pass an optional encoding argument?

SimonKrughoff (approved):
I agree with Tim that it's a little uncomfortable to force the encoding in the reader.  Can we just read in and re-pickle the BF kernel?

Other than that, LGTM.

python/lsst/daf/butler/formatters/pickleFormatter.py
TallJimbo: Does this argument actually break reading of Python 3 pickles?

timj: The formatter is only designed to read in pickles that were written by the formatter. It's not really meant for reading in items created in other places. Did this pickle file get added to the repository via the back door?

The documentation suggests that encoding is only used if a python2 pickle file is found. I'd prefer a separate formatter explicitly for these legacy files.

czwa: I've removed the formatter change, in favor of having an external tool to read and rewrite the obs_subaru bfKernel for this test.  As this file needs to be copied and manually inserted into the registry for Gen3 tests anyway, this isn't too much hassle, and avoids this issue for now.

PR #119, opened 2019-01-22
DM-17300: add configuration, bugfixes to support brightObjectMask dataset

TallJimbo:
First commit here uses the "lying" approach of Gen2, in which we pretend the `ObjectMaskCatalog` is saved as FITS because it has a `readFits` method.

Second and third commits stop that lying by using a new formatter defined in the pipe_tasks branch for this ticket.


timj (approved):
Looks okay although I'm a bit worried that I added that type checking for a reason 😄

PR #120, opened 2019-01-26
DM-17489: improvements for human-calibrated master calibrations

TallJimbo:


config/datastores/posixDatastore.yaml
```diff
@@ -5,7 +5,10 @@ datastore:
     table: PosixDatastoreRecords
   create: true
   templates:
-    default: "{collection}/{datasetType}/{skymap:?}/{tract:?}/{patch:?}/{label:?}/{abstract_filter:?}/{instrument:?}/{physical_filter:?}/{visit:?}/{exposure:?}/{detector:?}"
+    # valid_first and valid_last here are YYYYMMDD; we assume we'll switch to
+    # MJD (DM-15890) before we need more than day resolution, since that's all
+    # Gen2 has.
+    default: "{collection}/{datasetType}/{skymap:?}/{tract:?}/{patch:?}/{label:?}/{abstract_filter:?}/{instrument:?}/{physical_filter:?}/{valid_first:%Y%m%d?}-{valid_last:%Y%m%d?}/{visit:?}/{exposure:?}/{detector:?}"
```
SimonKrughoff, line 11: I'm a little confused by this change.  It seems like a generic PosixDatastore should not have a validity range.  Obviously, this is needed for calibrations, but most data products will not have concrete validity ranges (I don't think).  Are the two treated separately, or is it just expected that reduced datasets have infinite validity ranges?

SimonKrughoff (approved):
Looks good.  I'm don't see how datasets without validity ranges are handled, but I'm sure you'll explain :).

config/datastores/posixDatastore.yaml
timj: The `?` at the end of the format indicates that the item is optional. If they validity range is not present then it won't be used.

SimonKrughoff: Ah.  Cool.  Is that also part of the f-string niceness?  If so, that is amazingly convenient.
I interpreted the `?` as meaning the variable would be formatted, but the exact format was not known ahead of time.

timj: No. It's implemented in the `FileTemplates` class at https://github.com/lsst/daf_butler/blob/master/python/lsst/daf/butler/core/fileTemplates.py#L115 -- I use the standard format-string parser though so you can combine standard formatting with my extensions.

PR #123, opened 2019-01-30
DM-17495: attempt to avoid deadlocks with try-insert and immediate transactions

TallJimbo:


python/lsst/daf/butler/core/datasets.py
```diff
@@ -227,6 +227,30 @@ def __deepcopy__(self, memo):
...
+        ------
+        ValueError
+            Raised if the DatasetType is invalid, either because one or more
+            dimensions in ``self.dimensions`` is not in ``universe``, or the
+            storage class name is not recognized.
+        """
+        if not isinstance(self._dimensions, DimensionGraph):
+            self._dimensions = universe.extract(self._dimensions)
+        try:
+            self.storageClass
```
natelust, line 254: This seems a bit weird, should it be a KeyError? Is that because of the internals of self? I would expect and attribute error or something

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -192,62 +193,101 @@ def makeDatabaseDict(self, table, types, key, value):
...
-            ``dataset_id`` value, or `None` if no matching Dataset was found.
-
-        Raises
-        ------
-        ValueError
-            If dataId is invalid.
+        ref : `DatasetRef`.
+            A new `DatasetRef` instance.
         """
+        if datasetType is None:
```
natelust, line 216: Why would you not want to always use this information? Aka why have the parameter? What if they conflict?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -278,88 +318,54 @@ def query(self, sql, **params):
...
+        datasetType.normalize(universe=self.dimensions)
+        # If the DatasetType is already in the cache, we assume it's already in
+        # the DB (note that we don't actually provide a way to remove them from
+        # the DB).
+        existingDatasetType = self._datasetTypes.get(datasetType.name, None)
+        # If it's not in the cache, try to insert it.
+        if existingDatasetType is None:
+            try:
+                self._connection.execute(
+                    self._schema.tables["DatasetType"].insert().values(
```
natelust, line 330: Would it be good to have a begin around this as it is an insert

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -278,88 +318,54 @@ def query(self, sql, **params):
...
-                                   compStorageClass)
-            self.registerDatasetType(compType)
-        return True
+                # If adding the DatasetType record itself succeeded, add its
+                # dimensions (if any).  We don't guard this in a try block
+                # because a problem with this insert means the database
+                # content must be corrupted.
+                if datasetType.dimensions:
+                    self._connection.execute(
+                        self._schema.tables["DatasetTypeDimensions"].insert(),
```
natelust, line 347: Is this a virtual table that populates all the Dimension tables? Also does this insert into Dataset table because of foreign keys?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -278,88 +318,54 @@ def query(self, sql, **params):
...
+                for compName, compStorageClass in datasetType.storageClass.components.items():
+                    compType = DatasetType(datasetType.componentTypeName(compName),
+                                           dimensions=datasetType.dimensions,
+                                           storageClass=compStorageClass)
+                    self.registerDatasetType(compType)
+                # Inserts succeeded, nothing left to do here.
+                return True
+        # A DatasetType with this name exists, check if is equal
+        if datasetType == existingDatasetType:
+            return False
```
natelust, line 362: Returning false here becuase it already exists is a bit confusing for a downstream user who may be using true/false as a signal if something will be in the database. Maybe it would be good to add another exception for already in database? I know normally people look at exceptions as errors, but in a lot of python they are used as normal control flow.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -397,22 +405,9 @@ def addDataset(self, datasetType, dataId, run, producer=None, recursive=False, *
...
-        # in the presence of concurrent calls to addDataset.
-        # Then again, it is undoubtedly not the only place where
-        # this problem occurs. Needs some serious thought.
-        if self._findDatasetId(run.collection, datasetType, dataId) is not None:
-            raise ValueError("A dataset of type {} with id: {} already exists in collection {}".format(
-                datasetType, dataId, run.collection))
-
+        # Add the Dataset table entry itself.  Note that this will get rolled
+        # back if the subsequent call to associate raises, which is what we
+        # want.
```
natelust, line 409: is this true? this method does not need to be in a transaction?

natelust (approved):

python/lsst/daf/butler/core/datasets.py
TallJimbo: It's because it's looking up the storage class instance from a global registry, given the storage class name; I'm just accessing this property to make it do that.
I'm not sure it's the right exception class, but it's not obviously wrong, either, and I'd rather not change it on this ticket.  I will add a comment explaining what is going on.

python/lsst/daf/butler/registries/sqlRegistry.py
TallJimbo: Some callers have an already-validated `DataseType` instance, and in that case, we want to just use it to avoid a database query (every little bit helps).  Other callers don't, and then we have no choice but to run the query.  We are just assuming that they don't conflict (i.e. that's a precondition for this function), and that's one reason this ought to stay a private function.  I'll document that more clearly.

python/lsst/daf/butler/registries/sqlRegistry.py
TallJimbo: There is already a `begin` from the `transactional` decorator.

python/lsst/daf/butler/registries/sqlRegistry.py
TallJimbo: This is a table that just says which dimensions are used for a particular DatasetType (i.e. for "calexp" there would be entries with the strings "Instrument", "Visit", and "Detector").  The only foreign key in play here requires that the insert to DatasetType happen before associated inserts into DatasetTypeDimensions, but that's all.

python/lsst/daf/butler/registries/sqlRegistry.py
andy-slac: Just a note - in "true" databases which use different transaction isolation modes this conflict could happen during commit phase, not necessarily at insert.

python/lsst/daf/butler/registries/sqlRegistry.py
TallJimbo: I expect the return value to usually be ignored.  This is probably a case where the intent would be clearer f the docs from `Registry` were still copied here, but of course that has its own problems.  I'll add an extra note to the docs clarifying that a non-exception return always guarantees that the Registry definition is consistent with the given one.

python/lsst/daf/butler/registries/sqlRegistry.py
TallJimbo: It's in a transaction due to the `transactional` decorator (and, in practice, usually an outer `with self.registry.transaction()` block in `Butler.put`, too).

The neat thing about these nested transactions is they only roll back the context manager scopes through which the exception propagates; if you catch (and do not re-throw) the exception, only some of the transaction will be rolled back.  (In SQLite, these are impemented using SAVEPOINT, not BEGIN, though there is an outermost BEGIN).

In `Butler.put` that means this one would propagate all the way up, rolling back the `Datastore` transaction and removing any files already written, too.  In raw ingest, we actually use that to catch and ignore trying to re-ingest the same raw data twice (subject to configuration).

python/lsst/daf/butler/registries/sqlRegistry.py
TallJimbo: Ah, good point - and I assume we also can't assume that in a nested transaction the innermost commit will be enough to trigger the possible violation, either.

I think that means we really do need to some kind of UPSERT working for this method and `associate` eventually, then.

andy-slac: Well, even for upsert I think concurrency conflicts may be detected only on commit, again depending on transaction isolation and locking model used by particular database.

TallJimbo: Yes; my hope was that with an UPSERT we can tell the database directly how to resolve the conflict based on which of two constraints was violated, and then we don't care when that conflict resolution happens.  I'm not certain that's possible in general, but I think we've demonstrated that it can work with both SQLite and Oracle, even if (in the former case, at least) we're not sure if it's documented, guaranteed behavior or just luck that the implementation happens to do what we want.

cs2018ncsa: Oracle does not wait for commits to detect and resolve deadlocks. I don't really know how to see the SQL that is causing this without being able to run it in Oracle but is it possible to write these transactions to all lock resources in the same order? 

Still trying to figure out if @andy-slac considers Oracle a "true" database. :)

andy-slac: Oracle is definitely much truer database than SQLite 😉

PR #124, opened 2019-01-31
DM-17496: query size and data ID expansion optimizations

TallJimbo:
This change reduces the number of table clauses in the big `selectDimensions` query used in preflight, getting it below SQLite's maximum number of joins (64) in our tests (and hopefully all realistic cases).

~The remaining commits attempt to speed up some of the data ID manipulations we apply to each result row of that query, which turn out to be much slower than the query itself.  In my profiling, that brought down time spent in preflight for ci_hsc from about 12 minutes to about 5 minutes.  That's still much too slow, but my remaining plans for speedups involve some combination of more significant refactoring and lifting code to C++, and I don't think either belongs on this ticket.~

### DIFF
--- a/python/lsst/daf/butler/registries/sqlPreFlight.py
+++ b/python/lsst/daf/butler/registries/sqlPreFlight.py
@@ -628,7 +628,8 @@ class SqlPreFlight:
                     )
                     datasetRefs[dsType] = ref if ref is not None else DatasetRef(dsType, dsDataId, id=None)
                 else:
-                    datasetRefs[dsType] = self.registry.getDataset(id=row[col])
+                    datasetRefs[dsType] = self.registry.getDataset(id=row[col], datasetType=dsType,
+                                                                   dataId=dsDataId)
 
             count += 1
             yield PreFlightDimensionsRow(dataId, datasetRefs)
--- a/python/lsst/daf/butler/registries/sqlPreFlight.py
+++ b/python/lsst/daf/butler/registries/sqlPreFlight.py
@@ -30,6 +30,12 @@ from lsst.sphgeom.relationship import DISJOINT
 from lsst.daf.butler import DatasetRef, PreFlightDimensionsRow, DimensionJoin, DataId
 
 
+class DATASET_ID_DEFERRED:  # noqa N801
+    """Type used as a sentinal value to indicate when a dataset ID was not
+    queried for, as opposed to not found."""
+    pass
+
+
 _LOG = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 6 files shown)


andy-slac (approved):
Looks good. I'd probably keep `_buildDatasetSubquery` unchanged for now if there is a chance we could still use it for something else, and just pass `False` for its `isOutput` argument. There is a separate issue with output dataset_id, I'll comment on JIRA.

TallJimbo:
For anyone following along at home, this branch has at least temporarily become something of an integration branch for various bugfixes that have come up in preparing for the demo, not all of them related to the original ticket goals.  We'll clean that up when we have some breathing room (possibly after the demo).

TallJimbo:
Good news: the 5 minutes reported above seems to have included at *lot* of profiler overhead; I just timed preflight for a slightly larger graph (added forced photometry tasks, relative to the above), and it came in at 1m30s.

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -30,6 +30,12 @@
 from lsst.daf.butler import DatasetRef, PreFlightDimensionsRow, DimensionJoin, DataId
 
 
+class DATASET_ID_DEFERRED:  # noqa N801
+    """Type used as a sentinal value to indicates when a dataset ID was not
```
andy-slac: indicates -> indicate?

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -44,12 +50,39 @@ class SqlPreFlight:
...
+    neededDatasetTypes : `list` of `DatasetType`
+        The `list` of `DatasetTypes <DatasetType>` whose Dimensions will
+        be included in the returned column set. Output is limited to the
+        the Datasets of these DatasetTypes which already exist in the
+        registry.
+    futureDatasetTypes : `list` of `DatasetType`
+        The `list` of `DatasetTypes <DatasetType>` whose Dimensions will
+        be included in the returned column set. It is expected that
+        Datasets for these DatasetTypes do not exist in the registry,
+        but presently this is not checked.
```
andy-slac: This second sentence probably does not make much sense any more, that check is done in `GraphBuilder`, this class is not supposed to do that check.

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -279,15 +297,19 @@ def findSkyPixSubstitute():
...
-                     [(dsType, True) for dsType in futureDatasetTypes]
+        allDsTypes = [(dsType, False) for dsType in self.neededDatasetTypes] + \
+                     [(dsType, True) for dsType in self.futureDatasetTypes]
         for dsType, isOutput in allDsTypes:
 
             _LOG.debug("joining %s dataset type: %s",
                        "output" if isOutput else "input", dsType.name)
 
+            if isOutput and self.deferOutputIdQueries:
+                dsIdColumns[dsType] = DATASET_ID_DEFERRED
```
andy-slac, line 304: Maybe add `_LOG.debug` here to get some diagnostics about this shortcut

tests/test_sqlPreFlight.py
```diff
@@ -126,33 +125,33 @@ def testPreFlightInstrumentUnits(self):
...
-                             DataId(row.dataId, dimensions=packer2.dimensions.required))
-            self.assertNotEqual(packer1.pack(row.dataId), packer2.pack(row.dataId))
-        self.assertCountEqual(set(row.dataId["exposure"] for row in rows),
-                              (100, 101, 110, 111))
-        self.assertCountEqual(set(row.dataId["visit"] for row in rows), (10, 11))
-        self.assertCountEqual(set(row.dataId["detector"] for row in rows), (1, 2, 3))
+        for deferOutputIdQueries in (True, False):
+            with self.subTest(deferOutputIdQueries=deferOutputIdQueries):
+                preFlight = SqlPreFlight(self.registry, neededDatasetTypes=[rawType],
+                                         futureDatasetTypes=[calexpType], originInfo=originInfo)
```
andy-slac: Do you want to pass `deferOutputIdQueries` as argument to `SqlPreFlight` too?

andy-slac (approved):
Looks OK, few minor comments.

tests/test_sqlPreFlight.py
TallJimbo: Yes, oops.  Looks like I wasn't testing what I thought I was testing, but happily it does still pass with it fixed.

PR #126, opened 2019-02-02
DM-17611: performance optimizations for data ID manipulations

TallJimbo:


TallJimbo:
This change was reviewed on DM-17496; I'm just moving it here to separate it from unrelated ongoing work on that ticket.

PR #129, opened 2019-02-05
DM-17720: Improve user expression handling in preflight

andy-slac:
User expression is now parsed by preflight and then translated into
sqlalchemy clause expression. It now also supports using link names
instead of table.column syntax in expressions.

### DIFF
--- a/python/lsst/daf/butler/registries/sqlPreFlight.py
+++ b/python/lsst/daf/butler/registries/sqlPreFlight.py
@@ -27,8 +27,8 @@ from sqlalchemy.sql import select, and_, or_, not_, functions, literal, case, be
 
 from lsst.sphgeom import Region
 from lsst.sphgeom.relationship import DISJOINT
-from .. import DatasetRef, PreFlightDimensionsRow, DimensionJoin, DataId, exprParser
-from ..exprParser import ParserYacc
+from .. import DatasetRef, PreFlightDimensionsRow, DimensionJoin, DataId
+from ..exprParser import ParserYacc, TreeVisitor
 
 
 class DATASET_ID_DEFERRED:  # noqa N801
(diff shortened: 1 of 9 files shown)


python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -513,6 +519,88 @@ def _columns(selectable, names):
...
 
+    def _buildUserExpression(self, expression, dimensions):
+        """Generate SQLAlchemy clause from user expression.
+
+        Parameters
+        ----------
+        expression : `exprParser.Node`
+            Parsed user expression tree.
+        dimensions : `DimensionGraph`
+            All Dimensions included in this query.
```
TallJimbo: How hard would it be to make this all dimensions known to the registry (which is now its own class, `DimensionUniverse`, by the way), and then expand the query to include dimensions referenced in the expression but not the dataset types?

I could imagine it being useful to be able to run single-epoch processing on all visits that overlap a tract, for example.

python/lsst/daf/butler/registries/sqlPreFlight.py
```diff
@@ -513,6 +519,88 @@ def _columns(selectable, names):
...
+            Expression to be used in WHERE clause.
+
+        Raises
+        ------
+        ValueError
+            Raised for unexpected tree element, e.g. unknown link name.
+        Exception
+            Any exception raised by SQLAlchemy is propagated to caller.
+        """
+        if isinstance(expression, (exprParser.StringLiteral, exprParser.NumericLiteral)):
```
TallJimbo: Would it be possible to transform this conditional block into delegation to a polymorphic method on the `exprParser` classes that returns a `ClauseElement`?

All of these `isinstance` checks make me think of a classic OO antipattern, but I could also interpret this as the kind of match on algebraic sum types that would be quite natural in a language like Rust or OCaml that supported them, and in that case I suppose I don't have an objection to trying to fake that in Python.

TallJimbo (approved):
Looks good!  I think this is fine as is (and a very nice improvement in functionality), but I had a couple of comments that might conceivably make it better.

python/lsst/daf/butler/registries/sqlPreFlight.py
andy-slac: It should be possible in principle but it's probably non very trivial. Make a new ticket if you want that extra feature :slightly_smiling_face:

python/lsst/daf/butler/registries/sqlPreFlight.py
andy-slac: I thought about adding that facility to exprParser tree classes but decided that I want to keep that code free of extra dependencies like sqlalchemy - if we want to support few additional tree transformations it would not be right design to do all of that inside tree itself. You are right that this is a common problem and some languages have better support for implementing this sort of double-dispatch code. I'll see if I can add "better" visitor patter support for those classes if it does not make things too complicated.

andy-slac: @TallJimbo, I have re-implemented double dispatch using visitor pattern, could you check latest commit d11f7e4 to see if that smells better now.

TallJimbo (approved):
Looks good, nice work!

PR #133, opened 2019-02-18
DM-14459: Check for file overwrite

timj:


python/lsst/daf/butler/core/datasets.py
```diff
@@ -331,21 +331,22 @@ def dataId(self):
...
     def producer(self):
-        """The `Quantum` instance that produced (or will produce) the
-        Dataset.
+        """The `~lsst.daf.butler.Quantum` instance that produced (or will
+        produce) the Dataset.
 
-        Read-only; update via `Registry.addDataset()`,
-        `QuantumGraph.addDataset()`, or `Butler.put()`.
+        Read-only; update via `~lsst.daf.butler.Registry.addDataset()`,
+        `Quantum.addPredictedInput()`, or `lsst.daf.butler.Butler.put()`.
```
TallJimbo: Now that I'm seeing your Slack question in context, I think this should be `Quantum.addOutput` instead.

TallJimbo (approved):

PR #134, opened 2019-02-20
DM-17025: Template handling enhancements

timj:
* Add support for instrument overrides in formatter and template selection.

### DIFF
--- a/python/lsst/daf/butler/core/configSupport.py
+++ b/python/lsst/daf/butler/core/configSupport.py
@@ -100,17 +100,19 @@ class LookupKey:
 
     @property
     def name(self):
-        """Primary name string to use as lookup."""
+        """Primary name string to use as lookup. (`str`)"""
         return self._name
 
     @property
     def dimensions(self):
-        """Dimensions associated with lookup."""
+        """Dimensions associated with lookup.
+        (`DimensionGraph` or `DimensionNameSet`)"""
         return self._dimensions
 
     @property
     def dataId(self):
-        """Set of key/value tuples that are important for dataId lookup."""
+        """Set of key/value tuples that are important for dataId lookup.
+        (`frozenset`)"""
         return self._dataId
 
     def __hash__(self):
@@ -164,7 +166,7 @@ def normalizeLookupKeys(toUpdate, universe):
     toUpdate : `dict` with keys of `LookupKey`
         Dictionary to update.  The values are reassigned to normalized
         versions of the keys.  Keys are ignored that are not `LookupKey`.
-    universe : `DimensionGraph`
+    universe : `DimensionUniverse`
         The set of all known dimensions. If `None`, returns without
         action.
 
@@ -177,7 +179,8 @@ def normalizeLookupKeys(toUpdate, universe):
     Raises
     ------
     ValueError
-        A key exists where a dimension is not part of the ``universe``.
+        Raised if a key exists where a dimension is not part of
+        the ``universe``.
     """
     if universe is None:
         return
@@ -213,6 +216,38 @@ def processLookupConfigs(config):
         A `dict` with keys constructed from the configuration keys and values
         being simple strings.  It is assumed the caller will convert the
         values to the required form.
+
+    Notes
+    -----
+    The configuration is a mapping where the keys correspond to names
+    that can refer to dataset type or storage class names, or can use a
+    special syntax to refer to dimensions or dataId values.
+
+    Dimensions are indicated by using dimension names separated by a ``+``.
+    If a single dimension is specified this is also supported so long as
+    a ``+`` is found.  Dimensions are normalized before use such that if
+    ``PhysicalFilter+Visit`` is defined, then an implicit ``Instrument``
+    will automatically be added.
+
+    DataID overrides can be specified using the form: ``field<value>`` to
+    indicate a subhierarchy.  All keys within that new hierarchy will take
+    precedence over equivalent values in the root hierarchy.
+
+    Currently only a single dataId field can be specified for a key.
+    For example with a config such as:
+
+    .. code::
+
+       something:
+         calexp: value1
+         instrument<HSC>:
+           calexp: value2
+
+    Requesting the match for ``calexp`` would return ``value1`` unless
+    a `DatasetRef` is used with a dataId containing the key ``instrument``
+    and value ``HSC``.
+
+    The values of the mapping are stored as strings.
     """
     contents = {}
     for name, value in config.items():
--- a/python/lsst/daf/butler/core/configSupport.py
+++ b/python/lsst/daf/butler/core/configSupport.py
@@ -21,12 +21,12 @@
 
 """Support for configuration snippets"""
 
-__all__ = ("LookupKey", "processLookupConfigs")
+__all__ = ("LookupKey", "processLookupConfigs", "normalizeLookupKeys")
 
 import logging
 import re
 from collections.abc import Mapping
-from .dimensions import DimensionNameSet
+from .dimensions import DimensionNameSet, DimensionGraph
 
 log = logging.getLogger(__name__)
 
@@ -42,7 +42,7 @@ class LookupKey:
         represents dimensions (via ``dim1+dim2+dim3`` syntax) the name
         is converted to a `DimensionNameSet` and stored in ``dimensions``
         property.
-    dimensions : `DimensionNameSet`, optional
+    dimensions : `DimensionNameSet` or `DimensionGraph`, optional
         Dimensions that are relevant for lookup. Should not be specified
         if ``name`` is also specified.
     dataId : `dict`, optional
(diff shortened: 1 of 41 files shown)


python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -145,6 +145,18 @@ def __init__(self, template):
...
+        if not isinstance(other, FileTemplate):
+            return False
+
+        return self.template == other.template
+
+    def __str__(self):
+        return self.template
+
+    def __repr__(self):
+        return f"{self.__class__.__name__}(\"{self.template}\")"
```
TallJimbo: Seems like it'd be cleaner to use single quotes for the outer string.

python/lsst/daf/butler/core/formatter.py
```diff
@@ -140,6 +141,30 @@ class FormatterFactory:
...
+        instrument-specific section can be defined to override more general
+        formatter specifications.  This is represented in YAML using a
+        key of form ``instrument<name>`` which can then define formatters
+        that will be returned if a `DatasetRef` contains a matching instrument
+        name in the data ID.
+        """
+        for name, f in config.items():
+            if isinstance(f, Mapping):
+                for subName, subF in f.items():
+                    self.registerFormatter(f"{name}{subName}", subF)
```
TallJimbo: This suggests the docstring for `registerFormatter`'s first argument should be updated.

python/lsst/daf/butler/core/configSupport.py
```diff
@@ -0,0 +1,191 @@
...
+        return False
+
+    @property
+    def name(self):
+        """Primary name string to use as lookup."""
+        return self._name
+
+    @property
+    def dimensions(self):
+        """Dimensions associated with lookup."""
```
TallJimbo: Should document the type here (and other properties), especially if it's something that could either be a `DimensionNameSet` or a `DimensionGraph`, depending on whether rendezvous with `Registry.dimensions` has happened yet (or at least that's how I thought this work from Slack conversations yet; haven't gotten to that part of the code yet).

python/lsst/daf/butler/core/datasets.py
```diff
@@ -330,6 +330,12 @@ class DatasetRef:
 
     def __init__(self, datasetType, dataId, id=None, run=None, hash=None, components=None):
         assert isinstance(datasetType, DatasetType)
+
+        # Check the universes match if a DataId is provided
+        if isinstance(dataId, DataId):
+            if dataId.dimensions().universe != datasetType.dimensions.universe:
+                raise ValueError(f"Dimension universe mismatch for {dataId} and {datasetType}")
```
TallJimbo: I think it's safe to assume that the universes match - the only way they wouldn't be is if you've got two Registry instances from different dimension configurations in play, and even then I don't know that it's a serious problem if they don't match.

What is important is that the dimensions themselves do match, i.e. `dataId.dimensions() == datasetType.dimensions` - but this comparison can only safely be done after `datasetType.dimensions` has been "upgraded" to a `DimensionGraph` (`dataId.dimensions()` is always guaranteed to be a `DimensionGraph`).

python/lsst/daf/butler/core/configSupport.py
```diff
@@ -153,6 +156,46 @@ def clone(self, name=None, dimensions=None, dataId=None):
...
+
+    # Get the keys because we are going to change them
+    allKeys = list(toUpdate.keys())
+
+    for k in allKeys:
+        if not isinstance(k, LookupKey):
+            continue
+        if k.dimensions is not None and not isinstance(k.dimensions, DimensionGraph):
+            newDimensions = universe.extract(k.dimensions)
+            print(f"Olddim: {k.dimensions} New: {newDimensions}")
```
TallJimbo: Looks like a debugging relic.

python/lsst/daf/butler/core/configSupport.py
```diff
@@ -153,6 +156,46 @@ def clone(self, name=None, dimensions=None, dataId=None):
...
 
+def normalizeLookupKeys(toUpdate, universe):
+    """Normalize dimensions used in keys of supplied dict.
+
+    Parameters
+    ----------
+    toUpdate : `dict` with keys of `LookupKey`
+        Dictionary to update.  The values are reassigned to normalized
+        versions of the keys.  Keys are ignored that are not `LookupKey`.
+    universe : `DimensionGraph`
```
TallJimbo: Type should be `DimensionUniverse` (a special subclass of `DimensionGraph` for the universe).

python/lsst/daf/butler/core/formatter.py
```diff
@@ -141,6 +141,31 @@ class FormatterFactory:
     def __init__(self):
         self._mappingFactory = MappingFactory(Formatter)
 
+    def normalizeDimensions(self, universe):
+        """Normalize formatter lookups that use dimensions.
+
+        Parameters
+        ----------
+        universe : `DimensionGraph`
```
TallJimbo: Type should be `DimensionUniverse`.

python/lsst/daf/butler/core/formatter.py
```diff
@@ -141,6 +141,31 @@ class FormatterFactory:
...
+        dimensions, rewrites those keys to use a verified set of
+        dimensions.
+
+        Returns without action if the formatter keys have already been
+        normalized.
+
+        Raises
+        ------
+        ValueError
+            A key exists where a dimension is not part of the ``universe``.
```
TallJimbo: I thought we were supposed to start these sentences with something like "Raised if ...", but I could be mistaken.

python/lsst/daf/butler/core/mappingFactory.py
```diff
@@ -44,15 +44,45 @@ class MappingFactory:
     """
 
     def __init__(self, refType):
+        self.normalized = False
         self._registry = {}
         self.refType = refType
 
+    def normalizeRegistryDimensions(self, universe):
+        """Normalize dimensions used in registry keys to the supplied universe.
+        Parameters
```
TallJimbo, line 54: Missing newline.  Also `DimensionUniverse` below.

python/lsst/daf/butler/core/mappingFactory.py
```diff
@@ -44,15 +44,45 @@ class MappingFactory:
...
+
+        self.normalized = True
+
     def getFromRegistry(self, *targetClasses):
         """Get a new instance of the object stored in the registry.
 
         Parameters
         ----------
-        *targetClasses : `str` or objects supporting ``name`` attribute
+        *targetClasses : `LookupKey`, ``str` or objects with ``name`` attribute
```
TallJimbo: Extra backtick for `str`.

python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -78,6 +87,33 @@ def __init__(self, config, default=None):
...
+    def validateTemplate(self, *entities):
+        """Retrieve the template associated with each dataset type and
+        validate the dimensions against the template.
+
+        Parameters
+        ----------
+        *entities : `DatasetType`, `DatasetRef`, or `StorageClass`
+            Entities to validate against the templates.  True validation
+            requires that the supplied entities have access to dimension
+            graphs.
```
TallJimbo: Is "untrue validation" something that users should not do, because it will e.g. generate false positives, or something that might just not validate some things?

TallJimbo (approved):
Some minor comments.  Overall looks great!

python/lsst/daf/butler/core/formatter.py
timj: I think you must have been looking at each commit separately. This was from the first implementation that got instrument override working. The approach concatenated strings. The later commits supported dimensions and significantly improved the instrument overrides.

python/lsst/daf/butler/core/fileTemplates.py
timj: I mean that if you validate a template against a StorageClass or DatasetType that does not have a universe then you can't really do much. I will rephrase.

python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -236,3 +378,48 @@ def format(self, ref):
...
+
+        Notes
+        -----
+        If a `StorageClass` is supplied then validation can not be done
+        against Dimensions and it will always succeed.  If a `DatasetType`
+        or `DatasetRef` is provided without dimensions validation will
+        succeed.
+        """
+
+        # If we do not have dimensions defined then all we can do is shrug
```
timj, line 433: I think I should also check that one of `run` or `collection` is present in the template regardless of the supplied entity. Maybe I should do that on construction of a `FileTemplate` rather than during this validation.

python/lsst/daf/butler/core/formatter.py
timj: Ah, you are right that the docstring is now wrong for the current implementation using `LookupKey`.

python/lsst/daf/butler/core/formatter.py
timj: You are right. I never noticed that rules before.

PR #135, opened 2019-02-21
DM-16539: overhaul calibration dimensions and refactor selectDimensions implementations

TallJimbo:


config/schema.yaml
```diff
@@ -927,6 +929,105 @@ schema:
...
+        name: valid_first
+        type: datetime
+        doc: >
+          TAI timestamp of first exposure included in the range (inclusive).
+          May be far in the past to indicate an open interval.
+      -
+        name: valid_last
+        type: datetime
+        doc: >
+          TAI timestamp of last exposure included in the range (inclusive).
```
andy-slac, line 964: I think usual case for intervals/ranges is to make second timepoint exclusive, it makes some operations on ranges simpler (e.g. easier to do `[t0, t1), [t1, t2)` than `[t0, t1-something], [t1, t2]`). Though if you are only interested in exposure timepoints (i.e. you have reasonable/predictable gaps) inclusive can work fine too.

config/schema.yaml
```diff
@@ -927,6 +929,105 @@ schema:
...
+          May be far in the future to indicate an open interval.
+      foreignKeys:
+      -
+        src: instrument
+        tgt: Instrument.instrument
+
+    ExposureCalibrationLabelJoin:
```
andy-slac, line 971: Is this view used by anything, I can't find it on this PR except in unit test?

python/lsst/daf/butler/core/registry.py
```diff
@@ -888,14 +888,14 @@ def setDimensionRegion(self, dataId=None, *, update=True, region=None, **kwds):
...
+    def selectMultipleDatasetTypes(self, originInfo, expression=None, required=(), optional=(),
+                                   prerequisite=(), perDatasetTypeDimensions=(), expandDataIds=True):
         """Evaluate a filter expression and lists of
         `DatasetTypes <DatasetType>` and return a set of data unit values.
 
         Returned set consists of combinations of units participating in data
-        transformation from ``neededDatasetTypes`` to ``futureDatasetTypes``,
-        restricted by existing data and filter expression.
+        transformation from ``requiredDatasetTypes`` to ``optional
+        DatasetTypes``, restricted by existing data and filter expression.
```
andy-slac: ` ``required`` to ``optional`` DatasetTypes`?

python/lsst/daf/butler/core/registry.py
```diff
@@ -904,28 +904,40 @@ def selectDimensions(self, originInfo, expression, neededDatasetTypes, futureDat
...
-        futureDatasetTypes : `list` of `DatasetType` or `str`
+        optional : iterable of `DatasetType` or `str`
             The `list` of `DatasetTypes <DatasetType>` whose Dimensions will
-            be included in the returned column set. It is expected that
-            Datasets for these DatasetTypes do not exist in the registry,
-            but presently this is not checked.
+            be included in the returned column set. Datasets of these types
+            may or may not existin the registry.
+        prerequisite : iterable of `DatasetType` or `str`
+            `DatasetType`s that should not constrain the query results, but
```
andy-slac: `DatasetType`s -> `DatasetTypes <DatasetType>`?

or just drop backticks entirely.

python/lsst/daf/butler/core/registry.py
```diff
@@ -904,28 +904,40 @@ def selectDimensions(self, originInfo, expression, neededDatasetTypes, futureDat
...
+            must be present for all result rows.  These are included with
+            a LEFT OUTER JOIN, but the results are checked for NULL.  Unlike
+            regular inputs, prerequisite inputs lookups may be deferred
+            (see the documentaiton ``defer`` argument).
```
andy-slac: typo in `documentaiton`.
Which `defer` argument?

python/lsst/daf/butler/core/registry.py
```diff
@@ -904,28 +904,40 @@ def selectDimensions(self, originInfo, expression, neededDatasetTypes, futureDat
...
+            Any `DatasetType`'s that are present in both required and
+            prerequisite are considered ``prerequisite``.
```
andy-slac: backticks for `required` and first `prerequisite`?

python/lsst/daf/butler/sql/multipleDatasetQueryBuilder.py
```diff
@@ -0,0 +1,373 @@
...
+        this method directly.
+
+        Parameters
+        ----------
+        datasetType : `DatasetType`
+            Object representing the type of dataset to query for.
+        collections : `list` of `str`
+            String names of the collections in which to search for the dataset,
+            ordered from the first to be searched to the last to be searched.
+        optional : `bool`
```
andy-slac: update for `necessity`

python/lsst/daf/butler/sql/multipleDatasetQueryBuilder.py
```diff
@@ -0,0 +1,373 @@
...
+            If `True`, a dataset of this type does not need to be present to
+            generate a query result row; it should be added with a
+            LEFT OUTER JOIN.
+        defer : `bool`
+            If `True`, defer querying for the IDs for this dataset until
+            processing the main query results.  Must be `False` unless
+            ``optional`` is `True`.  Note that this does nothing unless
```
andy-slac: `necessity` here too.

python/lsst/daf/butler/sql/queryBuilder.py
```diff
@@ -0,0 +1,535 @@
...
+            self._whereClause = sqlExpression
+        else:
+            self._whereClause = op(self.whereClause, sqlExpression)
+
+    def whereParsedExpression(self, expression, op=and_):
+        """Add a SQL expression to the query's WHERE clause.
+
+        Parameters
+        ----------
+        sqlExpression
```
andy-slac, line 266: `expression : str`?

python/lsst/daf/butler/sql/queryBuilder.py
```diff
@@ -0,0 +1,535 @@
...
+        """Convert a query result row to the type appropriate for this
+        `QueryBuilder`.
+
+        This method is a customization point for `execute` that must be
+        implemented by subclasses.  Other code should generally not need
+        to call it directly.
+
+        Parameters
+        ----------
+        managed :  ResultColumnsManager.ManagedRow`
```
andy-slac: backtick is missing

python/lsst/daf/butler/sql/resultColumnsManager.py
```diff
@@ -0,0 +1,294 @@
...
+        ----------
+        selectable : `sqlalchemy.FromClause`
+            SQLAlchemy object representing a part of the query from which
+            a column can be selected.
+        link : `str`
+            String name of a dimension link column.  If this link is
+            already in the selected columns, this method does nothing.
+        datasetType : `DatasetType`, optional
+            If not `None`, this link corresponds to a "per-DatasetType"
+            dimension in a querym for this `DatasetType`.
```
andy-slac: querym

python/lsst/daf/butler/sql/resultColumnsManager.py
```diff
@@ -0,0 +1,294 @@
...
+        """
+        column = selectable.c[link]
+        if datasetType is None:
+            if link in self._indicesForDimensionLinks:
+                return
+            self._indicesForDimensionLinks[link] = len(self._columns)
+        else:
+            if link in self._indicesForPerDatasetTypeDimensionLinks:
```
andy-slac: `link in self._indicesForPerDatasetTypeDimensionLinks[datasetType]`?

python/lsst/daf/butler/sql/resultColumnsManager.py
```diff
@@ -0,0 +1,294 @@
...
+                    return True
+            return False
+
+        def makeDataId(self, *, datasetType=None, expandDataId=True, **kwds):
+            """Construct a `DataId` from the result row.
+
+            Parameters
+            ----------
+            datasetType : `DatasetType`, optional
+
```
andy-slac: Extra blank line

python/lsst/daf/butler/sql/singleDatasetQueryBuilder.py
```diff
@@ -0,0 +1,366 @@
...
+        return result
+
+    def convertResultRow(self, managed, *, expandDataId=True):
+        """Convert a result row for this query to a `DatasetRef`.
+
+        Parameters
+        ----------
+        managed : `ResultsColumnsManager.ManagedRow`
+            Intermediate result row object to convert.
+        expandDataId : `True`
```
andy-slac: `bool`?

andy-slac (commented):
Great job untangling preflight mess! I did a quick scrolling through it and I need to look at it more but here is a small set of comments that I have so far.

python/lsst/daf/butler/sql/queryBuilder.py
```diff
@@ -0,0 +1,535 @@
...
+class _ClauseVisitor(TreeVisitor):
+    """Implement TreeVisitor to convert user expression into SQLAlchemy
+    clause.
+
+    Parameters
+    ----------
+    tables : `dict`
+        Mapping of table names to `sqlalchemy.Table` instances.
+    dimensions : `DimensionGraph`
+        All Dimensions included in the query.
```
andy-slac: Parameters section needs an update

python/lsst/daf/butler/sql/queryBuilder.py
```diff
@@ -0,0 +1,535 @@
...
+            selectable = self.queryBuilder.findSelectableByName(table)
+            if selectable is None:
+                raise ValueError(f"No table or clause with name '{table}' in this query.")
+        else:
+            column = table
+            selectable = self.queryBuilder.findSelectableForLink(column)
+            if selectable is None:
+                raise ValueError(f"No table or clause providing link '{column}' in this query.")
+        try:
+            return selectable.c[column]
```
andy-slac: I think I have to blame mostly myself for this, but after reading our code I strongly prefer `columns` to single-letter `c`. No need to change it now, maybe just try to use it more in the future.

python/lsst/daf/butler/sql/queryBuilder.py
```diff
@@ -0,0 +1,535 @@
...
+            and the SELECT clause managed by `resultColumns`.
+        """
+        query = self.resultColumns.selectFrom(self.fromClause)
+        if self._whereClause is not None:
+            query = query.where(self._whereClause)
+        if whereSql is not None:
+            query = query.where(whereSql)
+        _LOG.debug("building query: %s",
+                   query.compile(bind=self.registry._connection.engine,
+                                 compile_kwargs={"literal_binds": True}))
```
andy-slac: Maybe wrap it into `if _LOG.isEnabledFor(logging.DEBUG):` to avoid calling `query.compile` if it's not needed.

python/lsst/daf/butler/sql/queryBuilder.py
```diff
@@ -0,0 +1,535 @@
...
+        Returns
+        -------
+        result
+            A single object of the type returned by `convertResultRow`, or
+            `None` if the query generated no results.
+
+        Notes
+        -----
+        A query row that include disjoint regions is filtered out, resulting
+        in `None` being returned.
```
andy-slac, line 379: If this is filtered by region overlap then I would assume that query is defined in a way to return just one "true" row  (with regions overlapping) but there may be other "false" rows (without overlap). With that assumption I think you cannot do `fetchone` but instead it should iterate and return first row that passes filter.

python/lsst/daf/butler/sql/singleDatasetQueryBuilder.py
```diff
@@ -0,0 +1,366 @@
...
+        fromClause = datasetTable.join(datasetCollectionTable,
+                                       datasetTable.c.dataset_id == datasetCollectionTable.c.dataset_id)
+        whereClause = and_(datasetTable.c.dataset_type_name == datasetType.name,
+                           datasetCollectionTable.c.collection == collection)
+        return cls(registry, fromClause=fromClause, whereClause=whereClause, datasetType=datasetType,
+                   selectableForDataset=datasetTable, addResultColumns=addResultColumns)
+
+    @classmethod
+    def fromCollections(cls, registry, datasetType, collections, addResultColumns=True):
+        """Construct a builder that searches a single collection for the
```
andy-slac: single collection -> multiple collections

python/lsst/daf/butler/sql/singleDatasetQueryBuilder.py
```diff
@@ -0,0 +1,366 @@
...
+            ON (DS.colpos = DSG.colpos
+                    AND
+                DS.link1 = DSG.link1
+                    AND
+                ...
+                    AND
+                DS.linkN = DSG.linkN)
+        """
+        if len(collections) == 1:
+            return cls.fromSingleCollection(registry, datasetType, collections[0])
```
andy-slac: Do you want to pass `addResultColumns` argument too?

python/lsst/daf/butler/sql/multipleDatasetQueryBuilder.py
```diff
@@ -0,0 +1,373 @@
...
+
+    Notes
+    -----
+    Logically, an instance of this class represents a single "path" connecting
+    a set of `DatasetRef`s which exist or may exist for a given set of
+    `DatasetType`s based on the `Dimension` relational algebra.
+
+    A multiple-dataset query returns a sequence of `MultipleDatasetQueryRow`
+    instances; each instance will have a unique `DataId`, but the
+    `DatasetRef`s in `datasetRefs` are not necessarily unique.  For example,
```
andy-slac: Double backticks around `datasetRefs`.

python/lsst/daf/butler/sql/multipleDatasetQueryBuilder.py
```diff
@@ -0,0 +1,373 @@
...
+            # Do the deferred dataset query, by running essentially the same
+            # SQL we would have used as a subquery before, but with the
+            # JOIN..ON expression now used as a WHERE expression with explicit
+            # values from the row's data ID.
+            expr = []
+            for link in data.joinLinks:
+                selectable = data.builder.findSelectableForLink(link)
+                column = selectable.c[link]
+                expr.append(column == dataId[link])
+            ref = data.builder.executeOne(whereSql=and_(*expr))
```
andy-slac, line 363: One possible optimization that I was thinking about when doing deferred queries is to reduce number of queries by writing queries to return multiple datasetRefs (e.g. all datasets in output collection if that is not too large) and doing in-memory join (or lookup). I have a feeling that doing individual queries here will be the main bottleneck for large graphs, if that is true we may consider re-implementing this piece of code.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -41,7 +41,8 @@
 from ..core.config import Config
 from ..core.dimensions import DataId
 from .sqlRegistryDatabaseDict import SqlRegistryDatabaseDict
-from .sqlPreFlight import SqlPreFlight
+from ..sql import MultipleDatasetQueryBuilder
+
 
 __all__ = ("SqlRegistryConfig", "SqlRegistry")
```
andy-slac: Should be moved above all imports

python/lsst/daf/butler/instrument.py
```diff
@@ -183,3 +184,35 @@ def updateVisitEntryFromObsInfo(dataId, obsInfo):
         exposure_time=obsInfo.exposure_time.to_value("s"),
     )
     return dataId
+
+
+def addUnboundedCalibrationLabel(registry, instrumentName):
```
andy-slac, line 189: Don't you want it added to `__all__`?

andy-slac (approved):
Looks OK (though non-trivial for sure). Few minor comments.

config/schema.yaml
TallJimbo: For this ticket I'd like to keep it inclusive to retain the previous logic (or, at least, the old as-documented logic - thinking more about the changes to the unit tests I had to make, I suppose this may actually be a change to the behavior to bring it in line with the documentation).

Anyhow, I think I did have some concrete use-case reasons for why I thought an inclusive upper bound would be better here, and I think it did involve the fact that there are always gap between exposures.  But I'm not sure I can reconstruct all of that thinking right now, and I'm happy to revisit it in the future.

config/schema.yaml
TallJimbo: Yes, it definitely is, and having it just appear in the unit tests makes sense; this should now behave pretty much just like any of the spatial join views (which also should appear by name only in the tests), but is computed from temporal overlap rather than spatial overlap.

python/lsst/daf/butler/core/registry.py
TallJimbo: Removed; I had copied this from the docs for `MultipleDatasetQueryBuilder`, but the `defer` option there is controlled by configuration in `SqlRegistry`, not passed directly by the user, so I've instead just said that deferral is something that `Registry` implementations may or may not do.

python/lsst/daf/butler/sql/queryBuilder.py
TallJimbo: I've switched to `columns` for all of the code in the new `sql` package.  I'll leave the rest of the `SqlRegistry` code as-is for now.

python/lsst/daf/butler/sql/multipleDatasetQueryBuilder.py
TallJimbo: The other code path that does include these queries in the big one instead of deferring them is still alive, and should Just Work for databases other than SQLite (where we need to defer to avoid the 64-term join limit), and I'm hoping we can avoid scaling problems with SQLite by just not using SQLite at scale.

Right now we also always do other follow-up queries to obtain component datasets for composites, of course, but that's a separate issue.

PR #136, opened 2019-03-04
DM-18181: Add butler configuration validation tool

timj:
Adds validation command line tool.

cc/ @MichelleGower

### DIFF
--- a/bin.src/validateButlerConfiguration.py
+++ b/bin.src/validateButlerConfiguration.py
@@ -36,6 +36,10 @@ if __name__ == "__main__":
                         help="Collection to refer to in this repository.")
     parser.add_argument("--quiet", "-q", action="store_true",
                         help="Do not report individual failures.")
+    parser.add_argument("--datasettype", "-d", action="append", type=str,
+                        help="Specific DatasetType to validate")
+    parser.add_argument("--ignore", "-i", action="append", type=str,
+                        help="DatasetType(s) to ignore for validation (e.g., 'raw')")
 
     args = parser.parse_args()
     # The collection does not matter for validation but if a run is specified
@@ -45,7 +49,8 @@ if __name__ == "__main__":
         logFailures = True
         if args.quiet:
             logFailures = False
-        butler.validateConfiguration(logFailures=logFailures)
+        butler.validateConfiguration(logFailures=logFailures, datasetTypeNames=args.datasettype,
+                                     ignore=args.ignore)
     except ValidationError:
         sys.exit(1)
     else:
--- a/bin.src/validateButlerConfiguration.py
+++ b/bin.src/validateButlerConfiguration.py
@@ -22,6 +22,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import argparse
+import sys
 
 from lsst.daf.butler import Butler, ValidationError
 
@@ -33,14 +34,19 @@ if __name__ == "__main__":
                         help="Filesystem path for an existing Butler repository.")
     parser.add_argument("--collection", "-c", default="validate", type=str,
                         help="Collection to refer to in this repository.")
+    parser.add_argument("--quiet", "-q", action="store_true",
+                        help="Do not report individual failures.")
 
     args = parser.parse_args()
     # The collection does not matter for validation but if a run is specified
     # in the configuration then it must be consistent with this collection
     butler = Butler(config=args.root, collection=args.collection)
     try:
-        butler.validateConfiguration(logFailures=True)
+        logFailures = True
+        if args.quiet:
+            logFailures = False
+        butler.validateConfiguration(logFailures=logFailures)
     except ValidationError:
-        pass
+        sys.exit(1)
     else:
         print("No problems encountered with configuration.")
--- a/bin.src/validateButlerConfiguration.py
+++ b/bin.src/validateButlerConfiguration.py
@@ -27,6 +27,28 @@ import sys
 from lsst.daf.butler import Butler, ValidationError
 
 
+def processCommas(arg):
+    """Given a list that might contain strings with commas, return expanded
+    list.
+
+    Parameters
+    ----------
+    arg : iterable of `str`
+        Values read from command line.
+
+    Returns
+    -------
+    expanded : `list` of `str`
+        List where any items with commas are expanded into multiple entries.
+    """
+    expanded = []
+    if arg is None:
+        return expanded
+    for item in arg:
+        expanded.extend(item.split(","))
+    return expanded
+
+
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="Validate the configuration files for a "
                                      "Gen3 Butler repository.")
@@ -37,20 +59,26 @@ if __name__ == "__main__":
     parser.add_argument("--quiet", "-q", action="store_true",
                         help="Do not report individual failures.")
     parser.add_argument("--datasettype", "-d", action="append", type=str,
-                        help="Specific DatasetType to validate")
+                        help="Specific DatasetType(s) to validate (can be comma-separated)")
     parser.add_argument("--ignore", "-i", action="append", type=str,
-                        help="DatasetType(s) to ignore for validation (e.g., 'raw')")
+                        help="DatasetType(s) to ignore for validation (can be comma-separated)")
 
     args = parser.parse_args()
     # The collection does not matter for validation but if a run is specified
     # in the configuration then it must be consistent with this collection
     butler = Butler(config=args.root, collection=args.collection)
+
+    logFailures = True
+    if args.quiet:
+        logFailures = False
+
+    # Process any commas in dataset type or ignore list
+    ignore = processCommas(args.ignore)
+    datasetTypes = processCommas(args.datasettype)
+
     try:
-        logFailures = True
-        if args.quiet:
-            logFailures = False
-        butler.validateConfiguration(logFailures=logFailures, datasetTypeNames=args.datasettype,
-                                     ignore=args.ignore)
+        butler.validateConfiguration(logFailures=logFailures, datasetTypeNames=datasetTypes,
+                                     ignore=ignore)
     except ValidationError:
         sys.exit(1)
     else:
--- a/bin.src/validateButlerConfiguration.py
+++ b/bin.src/validateButlerConfiguration.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python
+
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import argparse
+
+from lsst.daf.butler import Butler, ValidationError
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Validate the configuration files for a "
+                                     "Gen3 Butler repository.")
+    parser.add_argument("root",
+                        help="Filesystem path for an existing Butler repository.")
+    parser.add_argument("--collection", "-c", default="validate", type=str,
+                        help="Collection to refer to in this repository.")
+
+    args = parser.parse_args()
+    # The collection does not matter for validation but if a run is specified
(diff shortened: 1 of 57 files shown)


bin.src/validateButlerConfiguration.py
```diff
@@ -0,0 +1,46 @@
...
+                        help="Filesystem path for an existing Butler repository.")
+    parser.add_argument("--collection", "-c", default="validate", type=str,
+                        help="Collection to refer to in this repository.")
+
+    args = parser.parse_args()
+    # The collection does not matter for validation but if a run is specified
+    # in the configuration then it must be consistent with this collection
+    butler = Butler(config=args.root, collection=args.collection)
+    try:
+        butler.validateConfiguration(logFailures=True)
```
andy-slac: Maybe make `logFailures` configurable via command line arguments? I'm thinking about use case when this can be used just to verify that things are OK (in some other script maybe). `pass` below can be replaced with `sys.exit(1)` to make useful like that.

python/lsst/daf/butler/butler.py
```diff
@@ -39,13 +39,18 @@
 from .core.butlerConfig import ButlerConfig
 from .core.composites import CompositesMap
 from .core.dimensions import DataId
+from .core.exceptions import ValidationError
 
-
-__all__ = ("Butler",)
+__all__ = ("Butler", "ButlerValidationError")
```
andy-slac: `__all__` should be above all imports.

python/lsst/daf/butler/butler.py
```diff
@@ -546,3 +551,42 @@ def remove(self, datasetRefOrType, dataId=None, *, delete=True, remember=True, *
...
+        Raises
+        ------
+        ButlerValidationError
+            Raised if there is some inconsistency with how this Butler
+            is configured.
+        """
+        entities = list(self.registry.getAllDatasetTypes())
+
+        # Find all the registered instruments
+        instrument = []
```
andy-slac: `instruments`?

python/lsst/daf/butler/core/datastore.py
```diff
@@ -365,3 +372,27 @@ def transfer(self, inputDatastore, datasetRef):
             Reference to the required Dataset.
         """
         raise NotImplementedError("Must be implemented by subclass")
+
+    @abstractmethod
+    def validateConfiguration(self, *entities, logFailures=False):
```
andy-slac: What is the reason for using `*entities`, can't you just say that `entities` is a list or iterable?

python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -90,29 +91,49 @@ def __init__(self, config, default=None):
...
+            except FileTemplateValidationError as e:
+                failed.append(e)
+                if logFailures:
+                    log.fatal("Template failure: %s", str(e))
+
+        if failed:
+            if len(failed) == 1:
+                msg = str(failed[0])
+            else:
+                msg = f"{len(failed)} template validation failures: {'; '.join([str(f) for f in failed])}"
```
andy-slac: To my taste it's a bit too much for f-string expression, I'd probably move that `join()` into a regular code. And if `failed` contained strings instead of exceptions then you would not need list comprehension here.

python/lsst/daf/butler/core/registry.py
```diff
@@ -349,6 +360,17 @@ def getDatasetType(self, name):
...
         raise NotImplementedError("Must be implemented by subclass")
 
+    @abstractmethod
+    def getAllDatasetTypes(self):
+        r"""Get every registered `DatasetType`.
+
+        Returns
+        -------
+        types : `tuple` of `DatasetType`
+            All the `DatasetType`\ s.
```
andy-slac: I'd drop backticks here (just write DatasetTypes), there is plenty of links already to `DatasetType` in this docstring.

python/lsst/daf/butler/core/datastore.py
```diff
@@ -365,3 +372,27 @@ def transfer(self, inputDatastore, datasetRef):
...
+        """Validate some of the configuration for this datastore.
```
andy-slac, line 378: What do you mean by "some of the configuration"? Parts of configuration related to entities? Or is it entirely up to a subclass?

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -456,3 +457,40 @@ def transfer(self, inputDatastore, ref):
         assert inputDatastore is not self  # unless we want it for renames?
         inMemoryDataset = inputDatastore.get(ref)
         return [datastore.put(inMemoryDataset, ref) for datastore in self.datastores]
+
+    def validateConfiguration(self, *entities, logFailures=False):
+        """Validate some of the configuration for this datastore.
```
andy-slac, line 462: I think we starting to use more of

        # Docstring inherited from Datastore.validateConfiguration

style of comments instead of copying whole docstrings from base classes?

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -560,6 +558,53 @@ def transfer(self, inputDatastore, ref):
...
+                formatterFailed.append(e)
+                if logFailures:
+                    log.fatal("Formatter failure: %s", str(e))
+
+        if templateFailed or formatterFailed:
+            messages = []
+            if templateFailed:
+                messages.append(str(templateFailed))
+            if formatterFailed:
+                messages.append(",".join(formatterFailed))
```
andy-slac, line 605: Do you need to wrap each item in `formatterFailed` with `str()`?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -267,6 +267,14 @@ def _makeDatasetRefFromRow(self, row, datasetType=None, dataId=None):
         return DatasetRef(datasetType=datasetType, dataId=dataId, id=row["dataset_id"], run=run,
                           hash=datasetRefHash, components=components)
 
+    def getAllCollections(self):
+        # Docstring inherited from Registry.getAllCollections
+        datasetCollectionTable = self._schema.tables["DatasetCollection"]
+        result = self._connection.execute(select([datasetCollectionTable.c.collection]).distinct()).fetchall()
```
andy-slac, line 273: I started to dislike that single-letter `c` attribute, but it's OK to keep it for consistency with the rest of the code in this file.

python/lsst/daf/butler/core/registry.py
```diff
@@ -349,6 +360,17 @@ def getDatasetType(self, name):
         """
         raise NotImplementedError("Must be implemented by subclass")
 
+    @abstractmethod
+    def getAllDatasetTypes(self):
+        r"""Get every registered `DatasetType`.
+
+        Returns
+        -------
+        types : `tuple` of `DatasetType`
```
andy-slac: Why is it a tuple? (I'm probably an old school guy where tuple means a "record", and list means a sequence).

tests/test_sqlRegistry.py
```diff
@@ -449,6 +455,20 @@ def testDatasetUnit(self):
...
+        self.assertTrue(len(instruments), 1)
+        self.assertIn(dimensionValue["instrument"], instruments)
+
+        # Add a new instrument
+        dimensionValue3 = {"instrument": "DummyCam2", "visit_max": 10, "exposure_max": 10, "detector_max": 2}
+        registry.addDimensionEntry(dimensionName, dimensionValue3)
+
+        instrumentEntries = registry.findDimensionEntries(dimensionName)
+        instruments = {e["instrument"] for e in instrumentEntries}
+        self.assertEqual(instruments, set(["DummyCam", "DummyCam2"]))
```
andy-slac: You do `{item, item}` in other places but `set([item, item])` here?

andy-slac (approved):
Looks OK, few minor comments.

python/lsst/daf/butler/butler.py
timj: Ok. (even though I didn't put it there to begin with)

python/lsst/daf/butler/core/datastore.py
timj: I wanted to be able to say `datastore.validateConfiguration(ref)` without having to put the one DatasetRef inside a list.  I can change it if you think it's cleaner that way.

python/lsst/daf/butler/datastores/chainedDatastore.py
timj: Agreed, but here the notes section is different because it explains that each datastore is queried in turn.

python/lsst/daf/butler/core/datastore.py
timj: It's entirely up to the subclass and I'm only checking that templates and formatters are consistent. Even that is only done by PosixDatastore. InMemoryDatastore doesn't check anything at the moment.  The code isn't looking at every configuration item.

python/lsst/daf/butler/core/registry.py
timj: I don't really like returning mutable object for something that is meant to reflect a fixed state. I think of tuple as a read-only list.

python/lsst/daf/butler/core/datastore.py
andy-slac: If you want to accept both single item and a set of items in the same method then I think passing it via single argument as an iterable is a cleaner way. Alternatively you can accept both and wrap single items into a list internally (checking argument type) but I find that sort of interface a bit fragile, especially when item type is not fixed. It's your call but I'd prefer iterable here if it does not make API more difficult to use.

python/lsst/daf/butler/datastores/chainedDatastore.py
andy-slac: I wish we could extend inherited docstrings. Or could we do it with something like

    See `BaseClass.method` for description of parameters and return types.

    Notes
    -------
    Some additional info

?

python/lsst/daf/butler/core/registry.py
andy-slac: Could you return `frozenset` instead?

timj: Yes. Not sure why I don't.

python/lsst/daf/butler/core/registry.py
timj: I've tweaked the wording.

python/lsst/daf/butler/butler.py
```diff
@@ -546,3 +551,141 @@ def remove(self, datasetRefOrType, dataId=None, *, delete=True, remember=True, *
...
+                    if logFailures:
+                        log.fatal(f"Key '{key}' has unknown instrument")
+                    failedDataId.add(key)
+
+        messages = []
+
+        if datastoreErrorStr:
+            messages.append(datastoreErrorStr)
+
+        for failed, msg in ((failedNames, "Keys with out corresponding DatasetType or StorageClass entry: "),
```
andy-slac: with out or without?

python/lsst/daf/butler/core/configSupport.py
```diff
@@ -75,15 +75,23 @@ def __init__(self, name=None, dimensions=None, dataId=None):
             self._dimensions = dimensions
 
         # The dataId is converted to a frozenset of key/value
-        # tuples.
+        # tuples so that it is not mutable
         if dataId is not None:
             self._dataId = frozenset((k, v) for k, v in dataId.items())
```
andy-slac: Is comprehension needed here, I'd think `frozenset(dataId.items())` should do the same?

python/lsst/daf/butler/core/mappingFactory.py
```diff
@@ -110,8 +140,35 @@ def getFromRegistry(self, *targetClasses):
...
+        Returns
+        -------
+        instance : `object`
+            Instance of class stored in registry associated with the first
+            matching target class.
+
+        Raises
+        ------
+        KeyError
+            None of the supplied target classes match an item in the registry.
```
andy-slac: "Raised if"

python/lsst/daf/butler/butler.py
```diff
@@ -546,3 +551,141 @@ def remove(self, datasetRefOrType, dataId=None, *, delete=True, remember=True, *
...
+
+                # See if it is a StorageClass or a DatasetType
+                if key.name in self.storageClasses:
+                    pass
+                else:
+                    try:
+                        datasetType = self.registry.getDatasetType(key.name)
+                    except KeyError:
+                        if logFailures:
+                            log.fatal(f"Key '{key}' does not correspond to a DatasetType or StorageClass")
```
andy-slac: Does not flake8 warn that you are supposed to use `%`-style formatting with logging? Or is it confused by f-strings?

python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -80,42 +84,93 @@ def __init__(self, config, default=None):
...
+        failed = []
         for entity in entities:
-            template = self.getTemplate(entity)
-            template.validateTemplate(entity)
+            try:
+                matchKey, template = self.getTemplateWithMatch(entity)
+            except KeyError as e:
+                failed.append(e)
+                if logFailures:
+                    log.fatal("%s", str(e))
```
andy-slac: `str` is superfluous here (and few other `log.fatal` cases too)

andy-slac (approved):
Looks OK (but I'm under jetlag+coffee so things look mostly blurry 🙂). Few trivial comments below.

PR #137, opened 2019-03-13
DM-18438: Add ability to specify field lengths for DatabaseDict and use that

timj:
PosixDatastore now defines the lengths for PosixDatastoreRecords

### DIFF
--- a/tests/test_sqlDatabaseDict.py
+++ b/tests/test_sqlDatabaseDict.py
@@ -66,6 +66,8 @@ class SqlDatabaseDictTestCase(unittest.TestCase):
         d[1] = data[0]
         self.assertEqual(len(d), 1)
         self.assertEqual(d[1], data[0])
+        del d[1]
+        self.assertEqual(len(d), 0)
 
     def testKeyInValue(self):
         """Test that the key is not permitted to be part of the value."""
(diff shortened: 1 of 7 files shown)


timj:
@natelust I decided to pass in a separate dict with the lengths. I could change it so that the length could be passed in through the `types` directly if you preferred. ie:

```python
        types = {"path": (str, 256), "formatter": (str, 128), "storage_class": (str, 64),
                 "file_size": int, "checksum": (str, 64), "dataset_id": int}
```

tests/test_sqlDatabaseDict.py
```diff
@@ -83,6 +85,24 @@ def testKeyNotInValue(self):
         d = self.registry.makeDatabaseDict(table="TestTable", key=self.key, types=self.types, value=value)
         self.checkDatabaseDict(d, data)
 
+    def testLengths(self):
+        """Test that when a length is specified that it is actually used."""
+        value = namedtuple("TestValue", ["y", "z"])
+        data = {
+            0: value(y="zero", z=0.0),
+            1: value(y="this is too long", z=0.1),
```
natelust: My only comment is that it might be nice to name these keys "passing" and "failing" or something like that rather than generic 0 and 1 magic numbers, but its in test code, and I dont feel very strongly about it.

natelust (approved):

PR #138, opened 2019-03-19
Tickets/DM-18367 Get butler working with Oracle

natelust:


python/lsst/daf/butler/core/registry.py
```diff
@@ -189,6 +189,12 @@ def __init__(self, registryConfig, schemaConfig=None, dimensionConfig=None, crea
         for packerFactory in self._dataIdPackerFactories.values():
             packerFactory.updateFieldsToGet(self._fieldsToAlwaysGet)
 
+    @abstractmethod
+    def shutdown(self):
```
andy-slac: Should this also be called from registry `__del__` method in addition or instead of `Butler.__del__`? I think I prefer less explicit way of resource control (mostly because it's not always clear who or when should call this method), relying on object life time should simplify things. Can we get away completely without this method and instead adding necessary code to corresponding classes' `__del__`?

python/lsst/daf/butler/core/schema.py
```diff
@@ -318,14 +324,27 @@ def makeColumn(self, columnDescription):
                 typeKwargs[opt] = value
         if typeKwargs:
             columnType = columnType(**typeKwargs)
-        args = (columnName, columnType)
+        args.append(columnType)
+        # check if autoinc is in the description and if it is, check if it is
+        # set to true
+        if "autoinc" in description:
+            autoinc = description.pop("autoinc")
```
andy-slac: You can replace two above lines with equivalent and shorter:

    autoinc = description.pop("autoinc", False)

python/lsst/daf/butler/core/schema.py
```diff
@@ -318,14 +324,27 @@ def makeColumn(self, columnDescription):
...
+            if autoinc:
+                # Generate a sequence to use for auto incrementing for
+                # databases that do not support it naively, this will be
```
andy-slac: naively -> natively?

python/lsst/daf/butler/core/views.py
```diff
@@ -41,8 +41,23 @@ def __init__(self, name):
 
 @compiler.compiles(CreateView)
 def compileCreateView(element, compiler, **kw):
-    return "CREATE VIEW IF NOT EXISTS %s AS %s" % (element.name,
-                                                   compiler.sql_compiler.process(element.selectable))
+    # This is special cased as oracle does not support IF NOT EXISTS,
+    # but should be kept until the consequences of using CREATE OR REPLACE
+    # would be on users who are actively using a view when someone else
+    # attempts to create or replace it.
+    if "oracle" in compiler.__class__.__module__:
```
andy-slac: Relying on module name may not be super-portable, if you want to define some oracle-specific rules then it may be more reliable to pass dialect name to `compiles`:

    @compiler.compiles(CreateView, "oracle")
    def compileCreateView(element, compiler, **kw):

(https://docs.sqlalchemy.org/en/latest/core/compiler.html?highlight=compiler#dialect-specific-compilation-rules)
And I think the order of registration may matter, you may need to define generic case first and then dialect-specific like in the above link example.

python/lsst/daf/butler/gen2convert/writer.py
```diff
@@ -405,14 +405,14 @@ def insertObservationRegions(self, registry, datastore):
...
+               "    FROM Dataset Wcs "
+               "        INNER JOIN DatasetCollection WcsCollection "
                "            ON (Wcs.dataset_id = WcsCollection.dataset_id) "
-               "        INNER JOIN Dataset AS Metadata "
+               "        INNER JOIN Dataset Metadata "
                "            ON (Wcs.instrument = Metadata.instrument "
                "                AND Wcs.visit = Metadata.visit "
                "                AND Wcs.detector = Metadata.detector) "
-               "        INNER JOIN DatasetCollection AS MetadataCollection "
+               "        INNER JOIN DatasetCollection MetadataCollection "
```
andy-slac, line 415: General comment - I think we should not be writing queries as text but rely more on sqlalchemy methods to avoid backend-specific things in text. That would need different API for `registry.query`, and it's certainly not for this ticket, just something to think about for future development.

python/lsst/daf/butler/registries/oracleRegistry.py
```diff
@@ -0,0 +1,116 @@
...
+    BEGIN
+        SELECT count(view_name) into nCount FROM user_views where view_name = name;
+        IF(ncount <= 0)
+        THEN
+            execute immediate stmt;
+        END IF;
+    END view_exists_guard;
+    """
+    with closing(dbapiConnection.cursor()) as cursor:
+        cursor.execute(ifExistFunction)
```
andy-slac: Hmm, I'm not sure this is going to work well for all cases. Creating procedure in Oracle requires certain privilege and we may want to have read-only accounts that don't have any update privileges. I think it would be better to make this procedure at the same time (and in the same schema) as all regular registry tables and views. Another alternative is to try to implement above `compileCreateView` without relying on this procedure (but I do not know how to do it).

python/lsst/daf/butler/registries/oracleRegistry.py
```diff
@@ -0,0 +1,116 @@
...
+from sqlalchemy import create_engine
+from sqlalchemy.pool import NullPool
+
+
+from lsst.daf.butler.core.config import Config
+from lsst.daf.butler.core.registry import RegistryConfig
+
+from .sqlRegistry import SqlRegistry, SqlRegistryConfig
+
+__all__ = ("OracleRegistry", )
```
andy-slac: `__all__` should be above all imports.

python/lsst/daf/butler/registries/oracleRegistry.py
```diff
@@ -0,0 +1,116 @@
...
+                return text
+            elif text.startswith('"') and text.endswith('"'):
+                return text
+            else:
+                return text.replace(name, f'"{name}"')
+
+        def _oracleExecute(connection, cursor, statement, parameters, context, executemany):
+            for name in tables:
+                if name in statement:
+                    statement = re.sub(f"\"?'?{name}\\b'?\"?", lambda x: _ignoreQuote(name, x), statement)
```
andy-slac, line 107: I'm not very sure what the purpose of this code is? Does it try to fix some hand-written queries with incorrectly quoted or unquoted table names? Can you add a comment? This really looks like something that we should not be doing (hack is probably a better name for it) and it does not feel very robust. If the problem is with our hand-written queries then we should fix those.

python/lsst/daf/butler/registries/oracleRegistry.py
```diff
@@ -0,0 +1,116 @@
...
+            return statement, parameters
+        engine = create_engine(self.config["db"], poolclass=NullPool)
```
andy-slac, line 109: `NullPool` could have significant negative impact on Oracle performance, we should be ready to reconsider that at some point.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -452,6 +456,9 @@ def addDataset(self, datasetType, dataId, run, producer=None, recursive=False, *
                                                                        quantum_id=None,
                                                                        **links))
         datasetRef._id = result.inserted_primary_key[0]
+        # If the result is reported as a list of a number, unpack the list
+        if isinstance(datasetRef._id, list):
+            datasetRef._id = datasetRef._id[0]
```
andy-slac, line 461: Do you need this for Oracle? This seems to contradict sqlalchemy documentation: https://docs.sqlalchemy.org/en/latest/core/connections.html?highlight=inserted_primary_key#sqlalchemy.engine.ResultProxy.inserted_primary_key
(we should probably report it if Oracle backend violates that)

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -626,12 +632,22 @@ def removeDatasetLocation(self, datastoreName, ref):
     def addExecution(self, execution):
         # Docstring inherited from Registry.addExecution
         executionTable = self._schema.tables["Execution"]
-        result = self._connection.execute(executionTable.insert().values(execution_id=execution.id,
-                                                                         start_time=execution.startTime,
-                                                                         end_time=execution.endTime,
-                                                                         host=execution.host))
+        kwargs = {}
+        # Only pass in the execution_id to the insert statement if it is not
+        # None Otherwise some databases attempt to instert a null and fail. The
```
andy-slac: instert -> insert.
Dot or comma before Otherwise?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -672,8 +688,9 @@ def addRun(self, run):
         # TODO: this check is probably undesirable, as we may want to have
         # multiple Runs output to the same collection.  Fixing this requires
         # (at least) modifying getRun() accordingly.
-        selection = select([exists().where(runTable.c.collection == run.collection)])
-        if self._connection.execute(selection).scalar():
+        selection = select([func.count(1)]).select_from(runTable).where(runTable.c.collection ==
```
andy-slac: Just `count()` (which is `COUNT(*)`) is probably more common pattern for row counting.

andy-slac (approved):
Looks OK in general but some things feel questionable to me, see my comments/questions below.

python/lsst/daf/butler/core/registry.py
natelust: The short answer is no, I tried that first as I agree that is better. Unfortunately (I think because something somewhere still has references, or perhaps the order in which they are cleaned up) this still causes an exception to be raised when sqlalchemy does its own object clean up. The only way I could get it to not raise was explicitly calling this in whatever owned the registry.

natelust: It may be possible to reason and trace through the destructor orders, but I would prefer to get that done in another ticket, as it may turn into a decent chunk of work itself.

python/lsst/daf/butler/core/schema.py
natelust: :+1: I think that was left over from when I was doing something else in the block, or I am just being silly who knows. Thanks for the tip will change

python/lsst/daf/butler/gen2convert/writer.py
natelust: I agree

python/lsst/daf/butler/core/registry.py
timj: Python doesn't guarantee that `__del__` will be called so it's effectively useless. The usual recommendation is to provide `close`/`shutdown` methods and context managers.

python/lsst/daf/butler/registries/oracleRegistry.py
natelust: I will add a comment to the code, but basically the oracle database as it is currently setup does not support our case sensitive names of our tables correctly. This means all table names need to be quoted (in certain contexts). This rewrites all statements to quote the tables appropriately. Hopefully long term this will not be necessary as they are working to remove this issue. I will say as much in the comment.

python/lsst/daf/butler/registries/oracleRegistry.py
natelust: I followed the recommendation in the sqlRegistry file. I think the logic is that each task has its own butler instance, and thus its own connection, and as there is only ever one statement executed at a time in a task that NullPool was the most appropriate

python/lsst/daf/butler/registries/sqlRegistry.py
natelust: Yes, what is returned is a list on an element and an exception is raised

python/lsst/daf/butler/core/schema.py
natelust: Fun side note, my spell checker does not recognize that as a word.

python/lsst/daf/butler/core/views.py
natelust: Thanks for the info, that is much nicer.

python/lsst/daf/butler/registries/oracleRegistry.py
natelust: I was initialy thinking we may need this type of functionality in more places, but it turns out that we only needed it for views. I then was just over thinking it and changed the name to reflect it was for views, really I should have thought to go put it only in the Views construction phase which I did now.  We can always use that as a template to reconstruct "IF NOT EXISTS" behavior elsewhere if it ends up being needed.

python/lsst/daf/butler/butler.py
```diff
@@ -197,6 +197,21 @@ def __init__(self, config=None, collection=None, run=None):
...
+        # garbage collected. Python does not guarantee that this method
+        # will be called, or that it will be called in the proper order
+        # so if possible the user should explicitly call shutdown. This
+        # exists only to make an attempt to shut things down as a last
+        # resort if at all possible.
+        self.shutdown()
+
+    def shutdown(self):
+        """This method should be called to properly close any resources the
+        butler may have open
```
andy-slac: I'd like documentation to be extended a bit, in particular with information about how usable this instance is after calling `shutdown()` on it (or `close()` if you decide to rename it). I think it's probably safest to say that object is unusable, then just mention explicitly that  clients should not call any other methods after that.

python/lsst/daf/butler/registries/oracleRegistry.py
andy-slac: If it's not only for hand-written queries but for general case then I'm not happy that we are fighting SQLAlchemy rules. Would be easier to just use SQLAlchemy consistently everywhere than trying to workaround it. I'm OK for it for now but it should be revisited and discussed at some point (I'm not sure I understand all issues here).

python/lsst/daf/butler/registries/sqlRegistry.py
andy-slac: I'd like to spend a little bit of time on understanding that when I'm back, OK to have this for now.

andy-slac (approved):
Looks good with all changes, same comment for renaming `shutdown` as for `ctrl_mpexec`.

python/lsst/daf/butler/core/views.py
```diff
@@ -45,6 +45,29 @@ def compileCreateView(element, compiler, **kw):
...
+def compileCreateView(element, compiler, **kw):
+    # This is special cased as oracle does not support IF NOT EXISTS,
+    # but should be kept until the consequences of using CREATE OR REPLACE
+    # would be on users who are actively using a view when someone else
+    # attempts to create or replace it.
+    return """
+DECLARE
+    nCount NUMBER
+BEGIN
+    SELECT count(view_name) into nCount FROM user_views where view_name = {};
```
TallJimbo: Inconsistent capitalization of keywords here - probably should be `COUNT`, `INTO`, `WHERE`.

TallJimbo (approved):

PR #139, opened 2019-03-19
DM-10156: Replace all uses of Calib with PhotoCalib

parejkoj:


natelust (approved):

python/lsst/daf/butler/assemblers/exposureAssembler.py
```diff
@@ -211,7 +211,7 @@ def assemble(self, components):
 
         # Set other components
         exposure.setPsf(components.pop("psf", None))
-        exposure.setCalib(components.pop("calib", None))
+        exposure.setPhotoCalib(components.pop("photoCalib", None))
```
TallJimbo, line 214: Random side question: is `setCalib` sticking around, too? (I assume it's at least deprecated, but I had guessed that particular setter and it's associated getter wouldn't even be changed.)

parejkoj: Yes, `getCalib`/`setCalib` exist as deprecated interfaces in `Exposure` and `ExposureInfo`, replaced by `getPhotoCalib`/`setPhotoCalib`.

PR #140, opened 2019-03-21
DM-18356: Modify default template to make filenames unique

hsinfang:
Alternatively we can change this only in the BPS layer ([gen3_pegasus](https://github.com/lsst-dm/gen3_pegasus) ) and not here in daf_butler.

### DIFF
--- a/config/datastores/posixDatastore.yaml
+++ b/config/datastores/posixDatastore.yaml
@@ -8,7 +8,7 @@ datastore:
     # valid_first and valid_last here are YYYYMMDD; we assume we'll switch to
     # MJD (DM-15890) before we need more than day resolution, since that's all
     # Gen2 has.
-    default: "{collection}/{datasetType}.{component:?}/{skymap:?}/{tract:?}/{patch:?}/{label:?}/{abstract_filter:?}/{instrument:?}/{physical_filter:?}/{calibration_label:?}/{visit:?}/{exposure:?}/{detector:?}"
+    default: "{collection}/{datasetType}.{component:?}/{tract:?}/{patch:?}/{label:?}/{abstract_filter:?}/{physical_filter:?}/{visit:?}/{datasetType}_{component:?}_{tract:?}_{patch:?}_{label:?}_{abstract_filter:?}_{physical_filter:?}_{calibration_label:?}_{visit:?}_{exposure:?}_{detector:?}_{instrument:?}_{skymap:?}_{run}"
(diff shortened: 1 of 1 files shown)


timj (approved):
I'm happy with making the default template usable. It was only added as a proof of concept to start with. I think I'd prefer removing all those useless directories though and retaining the first two layers (so collection and run and datasetType.component) part.

hsinfang:
Sounds good. I got rid of some layers of directories, leaving those I think useful so that the numbers of files in one directory aren't getting too large too soon.

config/datastores/posixDatastore.yaml
```diff
@@ -8,7 +8,7 @@ datastore:
     # valid_first and valid_last here are YYYYMMDD; we assume we'll switch to
     # MJD (DM-15890) before we need more than day resolution, since that's all
     # Gen2 has.
-    default: "{collection}/{datasetType}.{component:?}/{skymap:?}/{tract:?}/{patch:?}/{label:?}/{abstract_filter:?}/{instrument:?}/{physical_filter:?}/{calibration_label:?}/{visit:?}/{exposure:?}/{detector:?}"
+    default: "{collection}/{datasetType}.{component:?}/{tract:?}/{patch:?}/{label:?}/{abstract_filter:?}/{physical_filter:?}/{visit:?}/{instrument:?}{run}_{skymap:?}_{datasetType}_{component:?}_{tract:?}_{patch:?}_{label:?}_{abstract_filter:?}_{physical_filter:?}_{calibration_label:?}_{visit:?}_{exposure:?}_{detector:?}"
```
timj: Do you want instrument and run to run into each other?

hsinfang: I actually did it on purpose but not for any good/real reasons (run is always a number).  Happy to have underscore there too.

hsinfang: Some datatypes don't have `instrument` set; we don't mind having some filenames starting with `_` right?

timj: Good point.  Maybe leave the underscore out then.  Not entirely sure what the correct fix is. Possibly remove strings between optional and mandatory if there are no `[a-zA-Z]` characters...

hsinfang: Another take...  Filenames are now led by `datasetType`; probably better aesthetically speaking although it's only a personal preference.

PR #141, opened 2019-03-25
DM-18577: Support butler relocation

timj:
* Allow a datastore/registry root to be specified relative to the location of the `butler.yaml` file.
* Allow the name of a posix datastore to be specified in the config file.

### DIFF
--- a/config/datastores/posixDatastore.yaml
+++ b/config/datastores/posixDatastore.yaml
@@ -1,6 +1,6 @@
 datastore:
   cls: lsst.daf.butler.datastores.posixDatastore.PosixDatastore
-  root: default_butler_repository
+  root: <root>/datastore
   records:
     table: PosixDatastoreRecords
   create: true
--- a/config/datastores/posixDatastore.yaml
+++ b/config/datastores/posixDatastore.yaml
@@ -1,6 +1,6 @@
 datastore:
   cls: lsst.daf.butler.datastores.posixDatastore.PosixDatastore
-  root: <root>/datastore
+  root: <butlerRoot>/datastore
   records:
     table: PosixDatastoreRecords
   create: true
--- a/config/datastores/posixDatastore.yaml
+++ b/config/datastores/posixDatastore.yaml
@@ -8,7 +8,7 @@ datastore:
     # valid_first and valid_last here are YYYYMMDD; we assume we'll switch to
     # MJD (DM-15890) before we need more than day resolution, since that's all
     # Gen2 has.
-    default: "{collection}/{datasetType}.{component:?}/{tract:?}/{patch:?}/{label:?}/{abstract_filter:?}/{physical_filter:?}/{visit:?}/{datasetType}_{component:?}_{tract:?}_{patch:?}_{label:?}_{abstract_filter:?}_{physical_filter:?}_{calibration_label:?}_{visit:?}_{exposure:?}_{detector:?}_{instrument:?}_{skymap:?}_{run}"
+    default: "{collection}/{datasetType}.{component:?}/{tract:?}/{patch:?}/{label:?}/{abstract_filter:?}/{physical_filter:?}/{visit:?}/{datasetType}_{component:?}_{tract:?}_{patch:?}_{label:?}_{abstract_filter:?}_{physical_filter:?}_{calibration_label:?}_{visit:?}_{exposure:?}_{detector:?}_{instrument:?}_{skymap:?}_{skypix:?}_{run}"
   formatters:
     TablePersistable: lsst.daf.butler.formatters.fitsCatalogFormatter.FitsCatalogFormatter
     TablePersistableWcs: lsst.daf.butler.formatters.fitsCatalogFormatter.FitsCatalogFormatter
--- a/python/lsst/daf/butler/butler.py
+++ b/python/lsst/daf/butler/butler.py
@@ -43,6 +43,7 @@ from .core.composites import CompositesMap
 from .core.dimensions import DataId
 from .core.exceptions import ValidationError
 from .core.repoRelocation import ROOT_TAG
+from .core.safeFileIo import safeMakeDir
 
 log = logging.getLogger(__name__)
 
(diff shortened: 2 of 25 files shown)


config/datastores/posixDatastore.yaml
```diff
@@ -1,6 +1,6 @@
 datastore:
   cls: lsst.daf.butler.datastores.posixDatastore.PosixDatastore
-  root: default_butler_repository
+  root: <root>/datastore
```
TallJimbo: Why `<root>`, rather than something more familiar like `{root}`, `$root`, or `%(root)s`?

python/lsst/daf/butler/butler.py
```diff
@@ -147,22 +148,22 @@ def makeRepo(root, config=None, standalone=False, createRegistry=True):
         config = Config(config)
         full = ButlerConfig(config)  # this applies defaults
         datastoreClass = doImport(full["datastore", "cls"])
-        datastoreClass.setConfigRoot(root, config, full)
+        datastoreClass.setConfigRoot(f"{ROOT_TAG}/datastore", config, full)
         registryClass = doImport(full["registry", "cls"])
-        registryClass.setConfigRoot(root, config, full)
+        registryClass.setConfigRoot(ROOT_TAG, config, full)
```
TallJimbo: I may have just not gotten far enough into the changes yet, but why the asymmetry between datastore and registry here?

python/lsst/daf/butler/core/butlerConfig.py
```diff
@@ -58,6 +58,8 @@ class ButlerConfig(Config):
 
     def __init__(self, other=None):
 
+        self.configDir = None
+
```
TallJimbo, line 69: Maybe move this down to just above the block that overrides it, to keep the logic that initializes this attribute in one place?

python/lsst/daf/butler/core/repoRelocation.py
```diff
@@ -0,0 +1,56 @@
...
+    necessary).
+
+    Parameters
+    ----------
+    configRoot : `str`
+        Directory root location as specified in a configuration file.
+    butlerRoot : `str`
+        Butler root directory.  Absolute path is inserted into the
+        ``configRoot`` where the ``<root>`` string is found.  If `None`
+        the current working directory is used for the root.
```
TallJimbo: I don't suppose you can use an f-string as a docstring, to insert the literal "<root>" here via `{ROOT_TAG}`?

python/lsst/daf/butler/core/repoRelocation.py
```diff
@@ -0,0 +1,56 @@
...
+import os.path
+
+# The special string to be used in configuration files to indicate that
+# the butler root location should be used.
+ROOT_TAG = "<root>"
+
+
+def replaceRoot(configRoot, butlerRoot):
+    """Update a configuration root with the butler root location (if
+    necessary).
```
TallJimbo: Could you elaborate on what "if necessary" means here?

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -106,14 +107,17 @@ def setConfigRoot(cls, root, config, full):
...
 
-    def __init__(self, config, registry):
+    def __init__(self, config, registry, butlerRoot=None):
         super().__init__(config, registry)
         if "root" not in self.config:
             raise ValueError("No root directory specified in configuration")
-        self.root = self.config["root"]
+
+        # Support repository relocation in config
+        self.root = replaceRoot(self.config["root"], butlerRoot)
```
TallJimbo, line 123: Just for my own edification, this expands the root in the in-memory config instance into an absolute path with no `<root>` placeholder, but never affects on-disk configuration, right?

TallJimbo (approved):
A few minor (mostly clarification) comments.  My only big-picture concern (which is also a minor one) is that - if I've understood correctly - sometimes the root in a `Config` instance is expanded, and sometimes it isn't, and that means knowing what to expect in one depends on some state that's hard to track.  If it's not too hard, I'd prefer for there to be a different attribute (maybe in the `Config`, maybe in the things that use one) that's always expanded so the one in the `Config` can always be unexpanded.  But if there's no good place to put that attribute, I could certainly believe that suggestion isn't worth the cost.

Mind updating the commit messages for the two YAML commits to make it clear that the unsafe loading is in gen2convert?  Right now, looking at just the messages, one would get the impression that one commit reverts the other.

config/datastores/posixDatastore.yaml
timj: I wanted something that looked a bit like a HTML tag that stood out as obviously not being a file system indicator, but I'm not completely hung up on it so `{root}` would work (on the basis that it looks like the templates lower down).

python/lsst/daf/butler/butler.py
timj: datastores write many files and where they end up and what they look like is entirely at the whims of the templates. Registry has a single file to writing it to a registry directory didn't seem worthwhile. Defaulting to a `datastore` subdirectory seems cleaner to me as a default (albeit causing pain in ci_hsc ingest).

python/lsst/daf/butler/core/repoRelocation.py
timj: Ok. I mean "only if `<root>` is in the configuration string". If it's not there then nothing will change.

python/lsst/daf/butler/datastores/posixDatastore.py
timj: Correct.

timj:
@TallJimbo I don't understand your main comment about being uncertain about whether something will be expanded or not. If you don't use `<root>` in your config file then there is no change from the current behavior.  If you use `<root>` then in all standard cases it will be replaced either with the value of `root` at the top of the butler config, or the directory containing the butler.yaml file you have asked to read.  The only time you may be unclear as to what is going on is if your butler is created with a `ButlerConfig` that was created without the use of a YAML file. In that case current working directory will be used for the root (and I could issue a warning for that rare case).

TallJimbo (approved):
Looks good!

python/lsst/daf/butler/core/repoRelocation.py
timj: No. Instead I documented the constant and refer to it so that at least in the sphinx version of the docs you can see the correct value.

config/datastores/posixDatastore.yaml
timj: As discussed on Slack, if I switch to `{root}` you need to quote the string in the YAML file. That seems more trouble than it's worth so I am sticking with the HTML-style tag.

python/lsst/daf/butler/core/butlerConfig.py
timj: Turned out that I have to leave it there because there was a bug in the original version of this code that meant that configDir was lost if a ButlerConfig was seeded from another ButlerConfig.

python/lsst/daf/butler/butler.py
timj: The asymmetry has been removed.

PR #142, opened 2019-04-02
DM-18839 Remove explicit registry close in butler

natelust:


TallJimbo (approved):
Code looks good, but commit message seems incorrect.

PR #144, opened 2019-04-04
DM-15374: Add Permissions/Constraints support code

timj:


python/lsst/daf/butler/core/configSupport.py
```diff
@@ -284,3 +288,48 @@ def processLookupConfigs(config):
             contents[lookup] = value
 
     return contents
+
+
+def processLookupConfigList(config):
+    """Process sections of configuration relating to lookups by dataset type
+    name, storage class name, dataId components or dimensions.
```
TallJimbo: What does "dataId components" mean?  I'd have thought it meant "dimensions", but that's also in the list.

python/lsst/daf/butler/core/configSupport.py
```diff
@@ -284,3 +288,48 @@ def processLookupConfigs(config):
...
+            for dataIdLookUp, subKeys in name.items():
+                kv = DATAID_RE.match(dataIdLookUp)
+                if kv:
+                    dataIdKey = kv.group(1)
+                    dataIdValue = kv.group(2)
+                    for subKey in subKeys:
+                        lookup = LookupKey(name=subKey, dataId={dataIdKey: dataIdValue})
+                        contents.add(lookup)
+                else:
+                    log.warning("Hierarchical key '%s' not in form 'key<value>', ignoring", name)
```
TallJimbo: I think I'd make this a hard error.

python/lsst/daf/butler/core/permissions.py
```diff
@@ -0,0 +1,124 @@
...
+from .config import Config
+from .configSupport import LookupKey, processLookupConfigList
+from .exceptions import ValidationError
+
+log = logging.getLogger(__name__)
+
+
+class PermissionsValidationError(ValidationError):
+    """Exception thrown when a permissions list is not consistent with the
+    associated `DatasetType`."""
```
TallJimbo: It's not obvious to me how this can occur, but maybe it's just the wording. Can you give an example?

python/lsst/daf/butler/core/permissions.py
```diff
@@ -0,0 +1,124 @@
...
+    pass
+
+
+class PermissionsConfig(Config):
+    """Configuration information for `Permissions`"""
+    pass
+
+
+class Permissions:
+    """Control whether a `DatasetType` is allowed to be handled.
```
TallJimbo: I'd add "by a Datastore" here, as that context is not obvious if you don't already know it.

Also, maybe "Dataset" instead of "DatasetType", because that's actually the general case (for dimension-based resolutions).

python/lsst/daf/butler/core/permissions.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+
+        # Get the names to use for lookup
+        names = set(entity._lookupNames())
+
+        # Test if this entity is explicitly mentioned for accept/reject
+        isExplicitlyAccepted = bool(names & self._accept)
+
+        if isExplicitlyAccepted:
+            return True
```
TallJimbo: Should we wait to see if it's also explicitly rejected, first (and raise)?  Or is that impossible given the syntax and/or other code?

python/lsst/daf/butler/core/permissions.py
```diff
@@ -0,0 +1,124 @@
...
+
+        isExplicitlyRejected = bool(names & self._reject)
+
+        if isExplicitlyRejected:
+            return False
+
+        # Now look for wildcard match -- we have to also check for dataId
+        # overrides
```
TallJimbo: Does the earlier short-circuiting mean that you can't accept or reject at the DatasetType level and then reverse that at the dimension level?  I don't actually know that we need that, but I think it's how I would have expected things to work (depending on the form of the config, which I haven't internalized, and we might want to document that if it *doesn't* work.

tests/test_permissions.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def testWildcardReject(self):
+        # Reject everything
+        config = PermissionsConfig({"reject": ["all"]})
+        permissions = Permissions(config)
+
+        self.assertFalse(permissions.hasPermission(self.calexpA))
+        self.assertFalse(permissions.hasPermission(self.pviA))
+
+        print("TESTING instrument reject")
```
TallJimbo: Looks like a debug relic.

tests/test_permissions.py
```diff
@@ -0,0 +1,180 @@
...
+        self.assertTrue(permissions.hasPermission(self.pviA))
+
+        # Accept everything explicitly
+        config = PermissionsConfig({"accept": ["all"]})
+        permissions = Permissions(config)
+
+        self.assertTrue(permissions.hasPermission(self.calexpA))
+        self.assertTrue(permissions.hasPermission(self.pviA))
+
+        print("TESTING instrument accept")
```
TallJimbo: Another debug relic.

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -153,6 +153,26 @@ def __init__(self, config, registry=None, butlerRoot=None):
                 break
         self.isEphemeral = isEphemeral
 
+        # And read the permissions list
+        permissionsConfig = self.config["permissions"] if "permissions" in self.config else None
```
TallJimbo: Would `permissionsConfig = self.config.get("permissions")` work?

tests/test_datastore.py
```diff
@@ -543,5 +549,49 @@ class ChainedDatastoreMemoryPermissionsTestCase(InMemoryDatastorePermissionsTest
...
+        sc2 = self.storageClassFactory.getStorageClass("StructuredDataJson")
+        dimensions = frozenset(("Visit", "PhysicalFilter", "Instrument"))
+        dataId1 = {"visit": 52, "physical_filter": "V", "instrument": "DummyCamComp"}
+        dataId2 = {"visit": 52, "physical_filter": "V", "instrument": "HSC"}
+
+        for datasetTypeName, dataId, sc, accepted in (("metric", dataId1, sc1, (False, True, False)),
+                                                      ("metric2", dataId1, sc1, (False, False, False)),
+                                                      ("metric2", dataId2, sc1, (True, False, False)),
+                                                      ("metric33", dataId2, sc2, (True, True, False)),
+                                                      ("metric2", dataId1, sc2, (False, True, False))):
```
TallJimbo: Recommend adding a `with self.subTest(...)` here to identify loop iterations (yeah, that doesn't work in pytest yet, but I hope it will someday, and in the meantime it's still useful for debugging when running tests directly with Python).

TallJimbo (approved):
Only very minor comments on the PR.

My only big one is that I'm still a little uncomfortable with the "permissions" terminology, and wouldn't mind throwing this problem to #dm-naming-things to see if we can get any better ideas.

python/lsst/daf/butler/core/configSupport.py
timj: I mean the actual value of a dimension as opposed to the set of dimension names.

python/lsst/daf/butler/core/permissions.py
timj: It's generic enough that it's not datastore-specific. You can use the same logic for anything involving a DatasetRef and knowing whether you are allowed to do something with it.

python/lsst/daf/butler/core/permissions.py
timj: This logic works with all the test cases I could come up with. If you have a suggestion for an edge case I'm happy to add it and will see whether it works.

python/lsst/daf/butler/core/permissions.py
timj: I will take another look at the dimensions side of things. If you accept a specific datasettype but have explicit reject on the dimensions then you will accept it. I think that if you accept dimensions but reject a datasettype then the datasettype should be rejected but I will see if I have a test for that.

python/lsst/daf/butler/datastores/chainedDatastore.py
timj: Probably. I always forget about `get()`.

python/lsst/daf/butler/core/permissions.py
timj: Currently only if you ask to accept all and reject all. I've tried to reword.

python/lsst/daf/butler/core/permissions.py
timj: As discussed on slack, we can already encode "accept only `raw` from `LATISS`" so I'm resolving this conversation.

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -350,19 +358,34 @@ def ingest(self, path, ref, formatter=None, transfer=None):
...
-        if counter == len(self.datastores):
-            raise NotImplementedError("Ingest not implemented by any of the chained datastores")
+                notImplementedCounter += 1
+            except DatasetTypeNotSupportedError:
+                notAcceptedCounter += 1
+
+        if (notAcceptedCounter + notImplementedCounter) == len(self.datastores):
+            log.debug("Not accepted counter: %d; Not implemented counter: %d", notAcceptedCounter,
+                      notImplementedCounter)
+            print(f"NotAccepted: {notAcceptedCounter}; NotImplemented: {notImplementedCounter}")
```
TallJimbo: Debug relic?

PR #145, opened 2019-04-09
DM-18629: Add ability to select which schema should be used by Gen3 Butler

MichelleGower:


python/lsst/daf/butler/registries/oracleRegistry.py
```diff
@@ -22,9 +22,12 @@
 __all__ = ("OracleRegistry", )
 
 import re
+import __main__
```
timj, line 25: I think this import is sufficiently non-standard that it's worth adding a comment that you need this to be able to associate the connection with a particular executable script.

python/lsst/daf/butler/registries/oracleRegistry.py
```diff
@@ -67,10 +70,20 @@ def setConfigRoot(cls, root, config, full):
...
+        url = engine.url.make_url(registryConfig[".db.url"])
+        if url.get_driver_name() != "cx_oracle":
+            raise ValueError("Unsupported driver(%s).  cx_oracle is only driver supported by OracleRegistry" %
+                             url.get_driver_name())
+
         self.schemaConfig = schemaConfig
         super().__init__(registryConfig, schemaConfig, dimensionConfig, create,
                          butlerRoot=butlerRoot)
 
+        if ".db.schema" in self.config:
```
timj, line 84: I see that the documentation for `OracleRegistry` is wrong in that it doesn't document all the parameters that are accepted by the `__init__` method. Regardless, can you please document somewhere what db fields Oracle uses?

python/lsst/daf/butler/registries/oracleRegistry.py
```diff
@@ -107,6 +120,28 @@ def _oracleExecute(connection, cursor, statement, parameters, context, executema
...
+            """ This function sets some Oracle values for each new pool
+            connection (e.g., module, schema).  Add connect listener to
+            Pool even if pool_size=1 for automatic reconnects.
+            """
+            # get executable name to save as DBMS_APPLICATION_INFO module
+            try:
+                modname = __main__.__file__
+            except AttributeError:
+                modname = "Unknown"
+            modname = os.path.basename(modname)[:48]  # column limited to 48 chars
```
timj, line 134: Do you need to drop path information?  Can't you store the final 48 characters regardless so that you get some idea of where the executable came from?
```
modname = modname[-48:]
```
?

timj (approved):
Looks good. Minor comments.

python/lsst/daf/butler/registries/oracleRegistry.py
andy-slac: How do you specify Oracle URL? Can this `db.schema` be extracted from URL itself?

MichelleGower: There does not currently appear to be any SQLAlchemy/cx_Oracle way to specify schema on the URL.   We could put the schema on the URL string and pre-parse it out ourselves before giving it to the create_engine.   As discussed in the Gen3 meeting, we are trying to move towards separating common server information from user information to have a central config + smaller user config.   The dict approach will allow us to use existing config combining methods.

andy-slac: OK, sounds reasonable. I was worried a bit that we add backend-specific information to a generic configuration but schemas are supported in Postgres too so this option could be more generic that just Oracle. One thing I'm still not very happy about is a sort of backdoor-ish way of communicating schema name to the `Metadata` (via `self._schema.metadata.schema`). It is probably cleaner to pass `schema=` argument to `Metadata` constructor if we could.

python/lsst/daf/butler/registries/oracleRegistry.py
MichelleGower: Mostly it was just an assumption that this extra information could be determined via other means (like software provenance) and didn't want to require use of wildcards in searches.
   
Example final 48 characters: 
- _butler/17.0.1-12-gd4f779f/bin/makeButlerRepo.py
- ipe_tasks/17.0.1-8-g40ca8d54/bin/ingestImages.py
- x64/ctrl_mpexec/17.0.1-4-g65ad0af+2/bin/pipetask

I checked with the DB admins.   There are no technical reasons to not store full 48 characters (space not an issue, no indexing to be affected by use of a wildcard for beginning of string, etc).   Will change code to always store last 48 characters.

python/lsst/daf/butler/registries/oracleRegistry.py
MichelleGower: The call to the Metadata constructor is buried farther down in code.  oracleRegistry.__init__ -> sqlRegistry.__init__ -> sqlRegistry._createSchema -> Schema.__init__ (doesn't get passed db config) -> SchemaBuilder.__init__ -> Metadata.

I could change the Schema codes to also get the db config.   The code could just assume that if db.schema is defined it should set it and let SQLAlchemy die if DB type doesn't like schema (so there wouldn't have to be dialect checks or Oracle specific schema.py).  This just seemed to be more invasive to the core code for the current ticket.

MichelleGower: Andy and I had a longer Slack discussion in which he said:  

>  I think it should be better this way, if Metadata is instantiated with incorrect (empty) schema name first there is a chance that that schema name is used for some operations before you change it to a correct one and that can cause issues.

python/lsst/daf/butler/registries/oracleRegistry.py
```diff
@@ -107,6 +120,28 @@ def _oracleExecute(connection, cursor, statement, parameters, context, executema
...
+
+            # get the connection
+            dbapi = kw['dbapi_connection']
+
+            # cx_Oracle specific means to set current schema and module
+            dbapi.module = modname
+            if ".db.schema" in self.config:
+                dbapi.current_schema = self.config[".db.schema"]
+        event.listen(Pool, "connect", _connCreator, named=True)
```
natelust, line 143: I was thinking about this, and the reason we needed this instead of just using the one in metadata is because of the views creation right? That is already special cased for each dialect. Looking at that code a bit, if we add metadata that is already passed into the class to the CreateView class here: https://github.com/lsst/daf_butler/blob/master/python/lsst/daf/butler/core/views.py#L109, then in the oracle sql string we can format it before executing it to supply the schema from the metadata in front of the table names, i.e. f"...{element.metadata.schema}.user_views  ..." .  This should remove the need for this block of code.

timj:
@MichelleGower is this pull request going anywhere?

python/lsst/daf/butler/registries/oracleRegistry.py
```diff
@@ -67,10 +70,20 @@ def setConfigRoot(cls, root, config, full):
     def __init__(self, registryConfig, schemaConfig, dimensionConfig, create=False,
                  butlerRoot=None):
         registryConfig = SqlRegistryConfig(registryConfig)
+
+        # currently code in class specific to cx_oracle driver
+        url = engine.url.make_url(registryConfig[".db.url"])
```
timj, line 75: The preference is to write the config access as : `["db", "url"]` rather than using a delimiter.

timj:
I think #191 may make this PR irrelevant.

timj:
@MichelleGower should we close this PR and the associated Jira ticket? Did #191 remove the need for this change?

MichelleGower:
Code changes are way out of date.   Closing pull request.

PR #146, opened 2019-04-19
Tickets/DM-19387 vectorize adding dimension entries in registry

natelust:


python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -830,6 +836,55 @@ def addDimensionEntry(self, dimension, dataId=None, entry=None, **kwds):
             self.setDimensionRegion(dataId)
         return dataId
 
+    @disableWhenLimited
+    @transactional
+    def addDimensionEntryList(self, dimension, dataIdList, entry=None, **kwds):
```
timj, line 841: Is there a downside to `addDimensionEntry` calling this itself so you don't get the duplicated code?  Also it seems odd to me that the list version has significantly more in it than the one-at-a-time version.

python/lsst/daf/butler/core/schema.py
```diff
@@ -126,7 +144,7 @@ class SchemaBuilder:
         The names of all entries in ``tables`` that are actually implemented as
         views.
     """
-    VALID_COLUMN_TYPES = {"string": String, "int": Integer, "float": Float, "region": LargeBinary,
+    VALID_COLUMN_TYPES = {"string": String, "int": Integer, "float": Float, "region": Base64Region,
                           "bool": Boolean, "blob": LargeBinary, "datetime": DateTime,
```
czwa, line 147: Is this backwards compatible, or do old butler repos need to be reconstructed?  I'm not clear how the `impl` is interpreted.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -70,6 +70,12 @@ class SqlRegistry(Registry):
     """Path to configuration defaults. Relative to $DAF_BUTLER_DIR/config or
     absolute path. Can be None if no defaults specified.
     """
+    defaultChunkSize = 999
+    """This sets the chunksize that vectorized operations will be broken into.
+    This is needed as some dbapis cannot handle very long lists of records to
+    process, so the complete list must be chunked into smaller lists and
+    handled one at a time
+    """
```
czwa, line 78: Is there a reason for 999?  Is this the largest minimum size for the dbapis we use?

python/lsst/daf/butler/registries/sqlRegistry.py
czwa: I agree.  This also duplicates some of the `setDimensionRegion` code, so perhaps that needs a `List` version as well to handle the `skypix`/region handling?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -830,6 +836,55 @@ def addDimensionEntry(self, dimension, dataId=None, entry=None, **kwds):
...
+    @transactional
+    def addDimensionEntryList(self, dimension, dataIdList, entry=None, **kwds):
+        # Docstring inherited from Registry.addDimensionEntryList.
+        dataIdList = [DataId(dataId, dimension=dimension, universe=self.dimensions, **kwds) for dataId in
+                      dataIdList]
+        # The given dimension should be the only leaf dimension of the graph,
+        # and this should ensure it's a true `Dimension`, not a `str` name.
+        # all of dataIds should share a common dimension, so use the first
+        dimension, = dataIdList[0].dimensions().leaves
+        if entry is not None:
```
czwa, line 849: Does there need to be a check that all entries in `dataIdList` have the same set of `dimensions`?  The case I'm worried about is based on the DM-19271 skymap change: if all `tract` and `patch` entries are contained in a single list (due to a think-o or something similar), does this fail noticeably?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -830,6 +836,55 @@ def addDimensionEntry(self, dimension, dataId=None, entry=None, **kwds):
...
+            for i, chunk in enumerate(chunks):
+                self._connection.execute(table.insert().values([dataId.fields(dimension, region=True) for
+                                                                dataId in chunk]))
+
+        except IntegrityError:
+            # TODO check for conflict, not just existence.
+            raise ConflictingDefinitionError(f"Existing definition for {dimension.name} entry in {chunk}.")
+        # Insert all of the cunked up skypix entries into the skypix join table
+        for i, chunk in enumerate(skypixParamChunks):
+            self._connection.execute(self._schema.tables[skypixJoin.name].insert(), skypixParams)
```
czwa, line 885: This doesn't raise here or in `setDimensionRegion`, but I'm not clear why.  Is attempting to set the region in a repo that doesn't have a skymap/skypix over that region not an error?

czwa (approved):
I think we discussed these issues in person and came to the conclusion that a future full-vectorization will solve any major concerns I had.

python/lsst/daf/butler/core/schema.py
natelust: Old repos will need recreated

python/lsst/daf/butler/registries/sqlRegistry.py
natelust: This is the largest size that sqlite will handle. I made it a class level variable so any registry subclasses could override it to whatever they want.

python/lsst/daf/butler/registries/sqlRegistry.py
natelust: That would fail noticeably

python/lsst/daf/butler/registries/sqlRegistry.py
natelust: The handling of setting dimension regions is a tiny bit different when calling the two different functions. In the future there will be a ticket to vectorize many of the other registry calls, in which case I agree that the single entry could differ to the vectorized version.

PR #147, opened 2019-04-23
Tentative implementation for S3 backed Datastore and RDS backed Registry

DinoBektesevic:
Supports a local Registry and an S3 backed DataStore or a RDS database service
Registry and S3 backed datastore. 

Tests are not yet completely mocked! ~~The tests for the S3 object storage are likely going to appear end of this week or early next week.~~ Testing RDS seems mission-impossible. 

Currently tests run against an pre-existing ~~bucket and an~~ RDS service. ~~If people are interested, I made the bucket public and I can grant individual IAMs account-based access to the bucket.~~ I can also share the password, username and white-list an IP address of people interested at looking at the RDS PostgreSQL Registry. As long as its only for running the tests of course, downloads and uploads are being charged for.

In real life use, the credentials for the bucket are expected to exist in `~/.aws/credentials` and `~/.rds/credentials` files. Bucket creation configuration (region, type...) should live in `~/.aws/config`. Both pass `test_butler` and both should pass `test_butlerFits` (S3Datastore+SQLite Registry tested, RDS should follow suite but no test implemented). 

The code is a bit messy - my apologies; but it is meant to start a discussion on the topic of what should the final implementation look like. Comments welcomed. Answers to many questions left as code comments especially welcomed.

Edit Apr. 23rd 2019, 13PDT: I figured out the S3 mocking mistake and added it in.


config/datastores/s3Datastore.yaml
```diff
@@ -0,0 +1,46 @@
...
+  root: <butlerRoot>/datastore
+  records:
+    table: S3DatastoreRecords
+  create: true
+  templates:
+    # valid_first and valid_last here are YYYYMMDD; we assume we'll switch to
+    # MJD (DM-15890) before we need more than day resolution, since that's all
+    # Gen2 has.
+    default: "{collection}/{datasetType}.{component:?}/{tract:?}/{patch:?}/{label:?}/{abstract_filter:?}/{physical_filter:?}/{visit:?}/{datasetType}_{component:?}_{tract:?}_{patch:?}_{label:?}_{abstract_filter:?}_{physical_filter:?}_{calibration_label:?}_{visit:?}_{exposure:?}_{detector:?}_{instrument:?}_{skymap:?}_{skypix:?}_{run}"
+  formatters:
```
timj, line 12: Duplicating all the formatters is clearly not sustainable. I think we probably need to change the `ConfigSubset.defaultConfigFile` to return multiple names so that we can merge your s3Datastore configuration with a posixDatastore one (or maybe pull the formatters out into a separate file in the root config and use `!include formatters.yaml`).

config/schema.yaml
```diff
@@ -44,7 +44,7 @@ schema:
       -
         name: instrument
         type: string
-        length: 8
+        length: 16
```
timj, line 47: These changes are on master so a rebase will remove them.

python/lsst/daf/butler/butler.py
```diff
@@ -211,6 +300,23 @@ def __init__(self, config=None, collection=None, run=None):
             if self.run is None:
                 self.run = self.registry.makeRun(runCollection)
 
+    def __del__(self):
```
timj, line 303: This and `close()` are no longer required. Please remove (not entirely sure why they are here if you've rebased).

python/lsst/daf/butler/butler.py
```diff
@@ -338,6 +444,7 @@ def put(self, obj, datasetRefOrType, dataId=None, producer=None, **kwds):
             # This is an entity without a disassembler.
             self.datastore.put(obj, ref)
 
+
```
timj, line 447: Remove this extra line

python/lsst/daf/butler/core/datasets.py
```diff
@@ -347,7 +347,6 @@ class DatasetRef:
 
     def __init__(self, datasetType, dataId, id=None, run=None, hash=None, components=None):
         assert isinstance(datasetType, DatasetType)
-
```
timj, line 350: why was this line removed?

python/lsst/daf/butler/core/location.py
```diff
@@ -61,7 +61,6 @@ def pathInStore(self):
 
     def updateExtension(self, ext):
         """Update the file extension associated with this `Location`.
-
```
timj, line 64: put this line back and all the others that were removed.

python/lsst/daf/butler/core/location.py
```diff
@@ -114,15 +110,171 @@ def fromUri(self, uri):
...
         path : `str`
             A standard POSIX path, relative to the `Datastore` root.
-
         Returns
         location : `Location`
             The equivalent `Location`.
         """
         uri = urllib.parse.urljoin("file://", path)
         return self.fromUri(uri)
+
```
timj, line 123: Two lines not three (does flake8 pass?)

python/lsst/daf/butler/core/butlerConfig.py
```diff
@@ -75,8 +90,21 @@ def __init__(self, other=None, searchPaths=None):
...
+        if isinstance(other, str):
+            scheme, root, relpath = utils.parsePath2Uri(other)
+            if scheme == 'file://':
+                if  os.path.isdir(other):
+                    other = os.path.join(other, "butler.yaml")
+            elif scheme == 's3://' and utils.bucketExists(other):
+                if NO_BOTO:
+                    raise ModuleNotFoundError(('boto3 not found.'
+                                               'Are you sure it is installed?'))
+                # I wonder if it would be a good idea to add a 'load' and 'loadFromfile' to
```
timj, line 102: I was expecting this logic to be in `Config` where it sees if it has been given a file.

python/lsst/daf/butler/core/butlerConfig.py
```diff
@@ -75,8 +90,21 @@ def __init__(self, other=None, searchPaths=None):
...
-        if isinstance(other, str) and os.path.isdir(other):
-            other = os.path.join(other, "butler.yaml")
+        if isinstance(other, str):
+            scheme, root, relpath = utils.parsePath2Uri(other)
+            if scheme == 'file://':
+                if  os.path.isdir(other):
+                    other = os.path.join(other, "butler.yaml")
+            elif scheme == 's3://' and utils.bucketExists(other):
+                if NO_BOTO:
+                    raise ModuleNotFoundError(('boto3 not found.'
```
timj, line 100: I think this should be `raise RuntimeError("Configuration file from S3 not supported")`

python/lsst/daf/butler/formatters/fileFormatter.py
```diff
@@ -150,9 +142,80 @@ def read(self, fileDescriptor, component=None):
...
+        data = self._assembleDataset(data, fileDescriptor, component)
+
+        return data
+
+    def fromBytes(self, pickledDataset, fileDescriptor, component=None):
+        """Read data from a bytestring.
+
+        Parameters
+        ----------
+        pickledDataset : `str`
```
timj, line 191: If this is bytes shouldn't it be `bytes` and not `str`?

python/lsst/daf/butler/formatters/fileFormatter.py
```diff
@@ -150,9 +142,80 @@ def read(self, fileDescriptor, component=None):
...
+        if data is None:
+            raise ValueError("Unable to read data with URI {}".format(fileDescriptor.location.uri))
+
+        # Assemble the requested dataset and potentially return only its component
+        # coercing it to its appropriate ptype
+        data = self._assembleDataset(data, fileDescriptor, component)
+
+        return data
+
+    def fromBytes(self, pickledDataset, fileDescriptor, component=None):
```
timj, line 186: fileDescriptor is mentioned here but is not documented in parameters.

python/lsst/daf/butler/formatters/fileFormatter.py
```diff
@@ -150,9 +142,80 @@ def read(self, fileDescriptor, component=None):
...
+
+        Parameters
+        ----------
+        pickledDataset : `str`
+            Bytes object to unserialize
+        component : `str`, optional
+            Component to read from the file. Only used if the `StorageClass`
+            for reading differed from the `StorageClass` used to write the
+            file.
+        pytype : 'class'
```
timj, line 197: This is not an argument to this method.

python/lsst/daf/butler/formatters/fileFormatter.py
```diff
@@ -173,11 +236,27 @@ def write(self, inMemoryDataset, fileDescriptor):
...
+        """Write a Python object to a bytestring.
+
+        Parameters
+        ----------
+        inMemoryDataset : `object`
+            The Python object to store.
+
+        Returns
+        -------
+        pickedDataset : `str`
```
timj, line 253: It should be returning bytes. `pickledDataset` is not the right name. There is no requirement that pickling be involved.

python/lsst/daf/butler/formatters/fileFormatter.py
```diff
@@ -173,11 +236,27 @@ def write(self, inMemoryDataset, fileDescriptor):
...
+            The bytestring representing the pickled dataset.
```
timj, line 254: It's not a bytestring it's some bytes, and they aren't pickled.

python/lsst/daf/butler/formatters/pickleFormatter.py
```diff
@@ -78,3 +78,46 @@ def _writeFile(self, inMemoryDataset, fileDescriptor):
         """
         with open(fileDescriptor.location.path, "wb") as fd:
             pickle.dump(inMemoryDataset, fd, protocol=-1)
+
+    def _fromBytes(self, pickledDataset, pytype=None):
+        """Read the bytes object as a python object.
+
+        Parameters
+        ----------
+        pickledDataset : `str`
```
timj, line 87: `bytes` not `str`

python/lsst/daf/butler/formatters/yamlFormatter.py
```diff
@@ -61,6 +62,38 @@ def _readFile(self, path, pytype=None):
 
         return data
 
+    def _fromBytes(self, pickledDataset, pytype):
+        """Read the bytes object as a python object.
+
+        Parameters
+        ----------
+        pickledDataset : `str`
```
timj, line 70: not pickled. Should be `bytes`.

python/lsst/daf/butler/formatters/yamlFormatter.py
```diff
@@ -61,6 +62,38 @@ def _readFile(self, path, pytype=None):
...
+            Not used by this implementation.
+
+        Returns
+        -------
+        data : `object`
+            Either data as Python object read from the pickled string, or None
+            if the string could not be read.
+        """
+        try:
+            data = pickle.loads(pickledDataset)
```
timj, line 82: This is the yamlFormatter so why is it using pickle?

python/lsst/daf/butler/formatters/yamlFormatter.py
```diff
@@ -84,6 +117,26 @@ def _writeFile(self, inMemoryDataset, fileDescriptor):
...
+        -------
+        data : `str`
+            Bytes object representing the pickled object.
+
+        Raises
+        ------
+        Exception
+            The object could not be pickled.
+        """
+        return pickle.dumps(inMemoryDataset, protocol=-1)
```
timj, line 138: YAML formatter should be returning a yaml string encoded to bytes.

python/lsst/daf/butler/registries/oracleRegistry.py
```diff
@@ -32,7 +32,6 @@
 
 from .sqlRegistry import SqlRegistry, SqlRegistryConfig
 
-
```
timj, line 35: This line has to be there so put it back.

python/lsst/daf/butler/registries/postgresqlRegistry.py
```diff
@@ -0,0 +1,96 @@
...
+        Load configuration
+    """
+
+    @classmethod
+    def setConfigRoot(cls, root, config, full):
+        """Set filesystem dependent config options for this Registry.
+        There are no such options currently.
+        """
+        # must override the Registry skypix class and level
+        super().setConfigRoot(root, config, full)
```
timj, line 35: You don't need to override if all you are doing is calling super().

python/lsst/daf/butler/registries/postgresqlRegistry.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def __init__(self, registryConfig, schemaConfig, dimensionConfig, create=False,
+                 butlerRoot=None):
+        registryConfig = SqlRegistryConfig(registryConfig)
+        self.schemaConfig = schemaConfig
+        super().__init__(registryConfig, schemaConfig, dimensionConfig, create,
+                         butlerRoot=butlerRoot)
+
+    def _createEngine(self):
```
timj, line 44: These hacks with the quoting will hopefully go away very soon with #150

python/lsst/daf/butler/registries/postgresqlRegistry.py
```diff
@@ -0,0 +1,96 @@
...
+        # then delete whats not neccessary.
+        constr = self.config['db']
+
+        # if both username and pass are in the db string - all is well
+        # if username and pass are a nick that exists in a local conf file
+        # read username and password from there
+        parsed = urlparse.urlparse(constr)
+        if parsed.username and (parsed.password is None):
+            localconf = configparser.ConfigParser()
+            localconf.read(os.path.expanduser('~/.rds/credentials'))
```
timj, line 88: there is nothing special about rds here so I don't think we should be using rds in the directory path. @MichelleGower how are you handling passwords for DB access?

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -432,16 +432,20 @@ def getDatasetType(self, name):
...
             datasetType.normalize(universe=self.dimensions)
 
         # Make a full DataId up front, so we don't do multiple times
         # in calls below.  Note that calling DataId with a full DataId
-        # is basically a no-op.
-        dataId = DataId(dataId, dimensions=datasetType.dimensions, universe=self.dimensions, **kwds)
+        # is basically a no-op
+        dataId = DataId(dataId, dimensions=datasetType.dimensions,
+                        universe=self.dimensions, **kwds)
+#       [TMPFIX?] I think because I have dimensionless datasettype this instantiation
```
timj, line 445: Remove this and fix up the lines above so that they don't change in this diff.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -561,7 +565,6 @@ def associate(self, collection, refs):
         # implementation with special "UPSERT" or "MERGE" syntax.  This
         # implementation is only concurrency-safe for databases that implement
         # transactions with database- or table-wide locks (e.g. SQLite).
-
```
timj, line 564: Put this line back.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -432,16 +432,20 @@ def getDatasetType(self, name):
     @transactional
     def addDataset(self, datasetType, dataId, run, producer=None, recursive=False, **kwds):
         # Docstring inherited from Registry.addDataset
-
```
timj, line 435: Probably better to put this line back so that it's clear that nothing is changing in this file.

tests/config/basic/butler-s3rds.yaml
```diff
@@ -0,0 +1,5 @@
+run: ingest
+datastore: !include s3Datastore.yaml
+registry: !include rdsRegistry.yaml
```
timj, line 3: For testing we should use sqlite and not even consider rds/postgres.

tests/config/basic/rdsRegistry.yaml
```diff
@@ -0,0 +1,7 @@
+# Default with a postgreSQL registry
```
timj, line 1: Do not include this file.

tests/config/basic/s3Datastore.yaml
```diff
@@ -0,0 +1,34 @@
+datastore:
+  cls: lsst.daf.butler.datastores.s3Datastore.S3Datastore
+  root: ./butler_test_repository
+  templates:
```
timj, line 4: Removing all this duplication is something I need to worry about.

tests/config/basic/butler-s3store.yaml
```diff
@@ -0,0 +1,5 @@
+run: ingest
+registry: !include sqlRegistry.yaml
```
timj, line 2: If this is so you can run in memory with sqlite can you rename it to something like sqliteMemoryRegistry.yaml?

tests/test_butler.py
```diff
@@ -27,6 +27,12 @@
 import tempfile
 import shutil
 import pickle
+import string
+import random
+
+import boto3
```
timj, line 33: These three imports have to be protected so that tests can be skipped if they are missing.

tests/test_butler.py
```diff
@@ -384,9 +390,10 @@ def testMakeRepo(self):
         self.assertEqual(butler2.registry.getAllCollections(), collections1)
 
     def testStringification(self):
+        # is root declared in the self.configFile, which one exactly is this?
         butler = Butler(self.tmpConfigFile)
-        butlerStr = str(butler)
 
+        butlerStr = str(butler)
```
timj, line 396: Revert this so that there is no change.

tests/test_butler.py
```diff
@@ -384,9 +390,10 @@ def testMakeRepo(self):
         self.assertEqual(butler2.registry.getAllCollections(), collections1)
 
     def testStringification(self):
+        # is root declared in the self.configFile, which one exactly is this?
```
timj, line 393: Remove this line. You have to declare what the string form of your datastore looks like so you can check to see that S3 (or whatever) turns up when the butler object is stringified.

tests/test_butler.py
```diff
@@ -541,6 +548,320 @@ def tearDown(self):
         if self.root is not None and os.path.exists(self.root):
             shutil.rmtree(self.root, ignore_errors=True)
 
+@mock_s3
```
timj, line 551: You will probably have to define a no-op `mock_s3` if moto can't be loaded.

tests/test_butler.py
```diff
@@ -541,6 +548,320 @@ def tearDown(self):
...
+    def genRoot(self):
+        """Returns a random string of len 20 to serve as a root
+        name for the temporary bucket repo.
+        """
+        rndstr = ''.join(
+            random.choice(string.ascii_uppercase + string.digits) for _ in range(20)
+        )
+        return rndstr
+
+    def setUp(self):
```
timj, line 579: Add a decorator along the line of
```
@unittest.skipIf(not boto3, "Warning: S3 boto is not installed")
```

python/lsst/daf/butler/formatters/pickleFormatter.py
```diff
@@ -78,3 +78,46 @@ def _writeFile(self, inMemoryDataset, fileDescriptor):
         """
         with open(fileDescriptor.location.path, "wb") as fd:
             pickle.dump(inMemoryDataset, fd, protocol=-1)
+
+    def _fromBytes(self, pickledDataset, pytype=None):
```
timj, line 82: My intent was for _fromBytes to be called from _readFile and for _writeFile to call _toBytes, so that we did not get duplication of calls to the serialization code.

tests/test_butler.py
```diff
@@ -541,6 +548,320 @@ def tearDown(self):
...
+        # BUTLER_ROOT_TAG
+        rooturi = f's3://{self.bucketName}/{self.root}'
+        self.datastoreStr = f"datastore={self.root}"
+        self.datastoreName = [f"S3Datastore@{rooturi}"]
+        Butler.makeRepo(rooturi, config=Config(self.configFile))
+        self.tmpConfigFile = os.path.join(rooturi, "butler.yaml")
+
+    def tearDown(self):
+        # clean up the test bucket, note that there are no directories in S3
+        # so there is no recursive Key deleteion functionality either.
```
timj, line 602: deleteion -> deletion

tests/test_butler.py
```diff
@@ -541,6 +548,320 @@ def tearDown(self):
...
+    Very practical for testing the S3Datastore quickly (~6x as fast as RDS setup/teardown).
+    I need to have a closer look at why is that - is it only because there are many DB
+    being simultaneously oppened on a 20GB 1vcpu machine or something else.
+    """
+    configFile = os.path.join(TESTDIR, "config/basic/butler-s3store.yaml")
+    fullConfigKey = None
+    validationCanFail = True
+
+    bucketName = 'bucketname'
+    permRoot = 'root/'
```
timj, line 566: Please document these attributes.

tests/test_butler.py
```diff
@@ -541,6 +548,320 @@ def tearDown(self):
         if self.root is not None and os.path.exists(self.root):
             shutil.rmtree(self.root, ignore_errors=True)
 
+@mock_s3
+class S3DatastoreButlerTestCase(ButlerTests, unittest.TestCase):
+    """S3Datastore specialization of a butler. Realistically this is a
+    S3 storage DataStore + a local SqlRegistry (could be an interesting usecase, but
+    currently absolutely useless as local registry is in memory.)
+
+    Very practical for testing the S3Datastore quickly (~6x as fast as RDS setup/teardown).
```
timj, line 557: Remove this comment since it's not describing this test, but describing a separate investigation.

tests/test_butler.py
```diff
@@ -541,6 +548,320 @@ def tearDown(self):
...
+    validationCanFail = True
+
+    bucketName = 'bucketname'
+    permRoot = 'root/'
+
+    registryStr = f"registry='sqlite:///:memory:'"
+
+    def genRoot(self):
+        """Returns a random string of len 20 to serve as a root
+        name for the temporary bucket repo.
```
timj, line 572: Might be worth noting that this is the S3 equivalent of the `tempfile.mkdtemp` calls for the posix datastore.

tests/test_butler.py
```diff
@@ -541,6 +548,320 @@ def tearDown(self):
...
+        """
+        rndstr = ''.join(
+            random.choice(string.ascii_uppercase + string.digits) for _ in range(20)
+        )
+        return rndstr
+
+    def setUp(self):
+        # I finally understand, inside MOTO's virtual AWS acc nothing yet exists
```
timj, line 580: This comment needs to be updated.

tests/test_butler.py
```diff
@@ -541,6 +548,320 @@ def tearDown(self):
...
+        # we need to recreate everything
+        s3 = boto3.resource('s3')
+        s3.create_bucket(Bucket=self.bucketName)
+
+        # create new repo root dir - analog to mkdir
+        if self.useTempRoot:
```
timj, line 586: I don't think you need to check for the case of `useTempRoot == False`.

tests/config/basic/s3Datastore.yaml
```diff
@@ -0,0 +1,34 @@
+datastore:
+  cls: lsst.daf.butler.datastores.s3Datastore.S3Datastore
+  root: ./butler_test_repository
```
timj, line 3: You need to consider what this should really say.

tests/test_butler.py
```diff
@@ -541,6 +548,320 @@ def tearDown(self):
...
+            self.root = self.genRoot()+'/'
+        else:
+            self.root = self.permRoot
+
+        # datastoreStr and Name need to be created here, otherwise how can we tell the
+        # name of the randomly created root, I'm not sure how to work this out with the magical
+        # BUTLER_ROOT_TAG
+        rooturi = f's3://{self.bucketName}/{self.root}'
```
timj, line 594: `BUTLER_ROOT_TAG` would be used if you wanted a scheme whereby you wanted to define the location of your datastore within S3 relative to the location of the butler.yaml file stored in S3. I think you need to consider a scheme where you do support `<butlerRoot>` in S3Datastore so that your default root in config files looks like:
```
root: s3://<butlerRoot>/
```
(whether the `s3://` prefix should be mandatory is something I haven't thought much about).

tests/test_butler.py
```diff
@@ -541,6 +548,320 @@ def tearDown(self):
...
+
+        # Check the template based on dimensions. This one is a bad template
+        with self.assertRaises(FileTemplateValidationError):
+            butler.datastore.templates.validateTemplates([ref])
+
+        with self.assertRaises(FileExistsError):
+            butler.put(metric, "metric3", dataId3)
+
+@mock_s3
+class S3RdsButlerTestCase(S3DatastoreButlerTestCase):
```
timj, line 681: I suggest we drop this from the pull request as it's not part of the S3 functionality.

python/lsst/daf/butler/butler.py
```diff
@@ -143,28 +242,18 @@ def makeRepo(root, config=None, standalone=False, createRegistry=True):
...
-        registryClass = doImport(full["registry", "cls"])
-        registryClass.setConfigRoot(BUTLER_ROOT_TAG, config, full)
-        if standalone:
-            config.merge(full)
-        config.dumpToFile(os.path.join(root, "butler.yaml"))
-        # Create Registry and populate tables
-        registryClass.fromConfig(config, create=createRegistry, butlerRoot=root)
+        
+        # this shouldn't probably be a string check but something cleverer-er
+        scheme, rootpath, relpath = parsePath2Uri(root)
```
timj, line 247: since rootpath and relpath are not being used it would be simpler to say:
```python
if scheme.startswith("root://"):
    # S3 version
else:
    # File version
```

python/lsst/daf/butler/butler.py
```diff
@@ -53,6 +61,97 @@ class ButlerValidationError(ValidationError):
...
+
+
+def _makeBucketRepo(root, config=None, standalone=False, createRegistry=True):
+    """Create an empty data repository by adding a butler.yaml config
+    to a repository root directory in a S3 Bucket.
+    """
+    if NO_BOTO:
+        raise ModuleNotFoundError("Could not find boto3. Are you sure it is installed?")
+
+    if isinstance(config, (ButlerConfig, ConfigSubset)):
```
timj, line 110: This test has already been applied so no need to duplicate it here.

python/lsst/daf/butler/butler.py
```diff
@@ -53,6 +61,97 @@ class ButlerValidationError(ValidationError):
...
+    # do not propagate an explicit root from the config file
+    #if "root" in config:
+    #    del config["root"]
+    full = ButlerConfig(config)  # this applies defaults
+
+    datastoreClass = doImport(full["datastore", "cls"])
+    datastoreClass.setConfigRoot(root, config, full)
+    registryClass = doImport(full["registry", "cls"])
+    # this if here is a monkeypatch for the bucket datastore, local repo tests
+    # I don't really understand the BUTLER_ROOT_TAG thing
```
timj, line 134: It forces the newly created configuration file to be relative to the butler.yaml file that is being created. This is a reasonable default.  Later on repoRelocation runs to replace `<butlerRoot>` with the actual root calculated from the location of the butler.yaml.  This requires that you have recorded the location of the config file in the `ButlerConfig`.

python/lsst/daf/butler/butler.py
```diff
@@ -53,6 +61,97 @@ class ButlerValidationError(ValidationError):
     pass
 
 
+# I didn't want to clog up the butler with more classmethods or a long makeRepo with ifs
```
timj, line 64: There is too much code duplication here.  The S3 stuff involves the serialization of the config and the pushing it to the relevant S3 location. I think that should be moved to `Config` so that you can switch on S3 when deciding whether to write the config file in the file system and calling `config.dumpToFile(os.path.join(root, "butler.yaml"))` or instead calling `config.storeInS3()`.

python/lsst/daf/butler/core/location.py
```diff
@@ -114,15 +110,171 @@ def fromUri(self, uri):
...
+class S3Location:
+    """Identifies a location in the `Datastore`.
+    TODO: This will have broken functionality for extensions etc.
+    """
+
+    __slots__ = ("_scheme", "_bucket", "_datastoreRoot", "_relpath")
+
+    def __init__(self, scheme, bucket, datastoreRoot, relpath, **kwargs):
+        # no risks, maximal sanitation
+        self._scheme = scheme+'://' if scheme[-3:]!='://' else scheme
```
timj, line 135: This is not flake8 compliant. Needs some spacing.

python/lsst/daf/butler/core/location.py
```diff
@@ -114,15 +110,171 @@ def fromUri(self, uri):
...
+        self._bucket = bucket.strip('/') + '/'
+        self._datastoreRoot = datastoreRoot.strip('/') + '/'
+        self._relpath = relpath.lstrip('/')
+
+    def __str__(self):
+        return self.scheme + os.path.join(self._bucket, self._datastoreRoot, self._relpath)
```
timj, line 141: This could forward to `uri` property as is done in `Location`.

python/lsst/daf/butler/core/location.py
```diff
@@ -114,15 +110,171 @@ def fromUri(self, uri):
...
             The equivalent `Location`.
         """
         uri = urllib.parse.urljoin("file://", path)
         return self.fromUri(uri)
+
+
+
+class S3Location:
+    """Identifies a location in the `Datastore`.
+    TODO: This will have broken functionality for extensions etc.
```
timj, line 128: Please explain what you mean by broken for extensions.

python/lsst/daf/butler/core/location.py
```diff
@@ -114,15 +110,171 @@ def fromUri(self, uri):
...
+        return os.path.join(self._datastoreRoot, self._relpath)
+
+    @property
+    def pathInStore(self):
+        """Path corresponding to location relative to `Datastore` root.
+        """
+        return self._relpath
+
+    def updateExtension(self, ext):
+        """Update the file extension associated with this `Location`.
```
timj, line 173: `S3Location` not `Location` (this is not a subclass).

python/lsst/daf/butler/core/location.py
```diff
@@ -114,15 +110,171 @@ def fromUri(self, uri):
...
+            return
+        path, _ = os.path.splitext(self._relpath)
+
+        # Ensure that we have a leading "." on file extension (and we do not
+        # try to modify the empty string)
+        if ext and not ext.startswith("."):
+            ext = "." + ext
+
+        self._relpath = path + ext
+
```
timj, line 191: Too many blank lines (should be 2)

python/lsst/daf/butler/core/location.py
```diff
@@ -114,15 +110,171 @@ def fromUri(self, uri):
...
+
+
+class S3LocationFactory:
+    """Factory for `Location` instances.
```
timj, line 195: Factor for S3Location.

python/lsst/daf/butler/core/location.py
```diff
@@ -114,15 +110,171 @@ def fromUri(self, uri):
...
+        ----------
+        datastoreRoot : `str`
+            Root location of the `Datastore` in the filesystem.
+        """
+        # no chances, maximal sanitation
+        self._bucket = bucket.strip('/')
+        self._datastoreRoot = datastoreRoot.strip('/')
+
+    def fromUri(self, uri):
+        """Factory function to create a `Location` from a URI.
```
timj, line 210: S3Location

python/lsst/daf/butler/core/location.py
```diff
@@ -114,15 +110,171 @@ def fromUri(self, uri):
...
+        uri : `str`
+            A valid Universal Resource Identifier.
+        Returns
+        location : `Location`
+            The equivalent `Location`.
+        """
+        if uri is None or not isinstance(uri, str):
+            raise ValueError("URI must be a string and not {}".format(uri))
+
+        # uri should alwazs contain scheme, bucket, root, and relpath
```
timj, line 222: "alwazs"

python/lsst/daf/butler/core/location.py
```diff
@@ -114,15 +110,171 @@ def fromUri(self, uri):
...
+            A standard POSIX path, relative to the `Datastore` root.
+        Returns
+        location : `Location`
+            The equivalent `Location`.
+        """
+        if os.path.isabs(path):
+            raise ValueError(('A path whose absolute location is in an S3 bucket '
+                             'can not have an absolute path: {}').format(path))
+
+        # sometimes people give relative path with the datastoreRoot
```
timj, line 247: Who is "people" here?  This comment concerns me since it implies that something is going wrong with S3Datastore or that something is going wrong with PosixDatastore `LocationFactory.fromPath` and I haven't realized yet.  Why are you getting calls to `fromPath` with paths that include the bucket name?

python/lsst/daf/butler/core/location.py
```diff
@@ -114,15 +110,171 @@ def fromUri(self, uri):
...
+        # and sometimes they give the pathInStore type path. Sometimes the bucket
+        # is provided as well. This can duplicate the datastoreRoot and bucket
+        # which we remove
+        dirs = path.lstrip('/').split('/')
```
timj, line 251: Why does `os.path.split` not work?  Are you worried that these are URL paths and not posix paths as such and therefore it is possible that `/` is not the separator?  Also, this variable will contain filename as well as directories.

python/lsst/daf/butler/core/location.py
```diff
@@ -114,15 +110,171 @@ def fromUri(self, uri):
...
+
+        return S3Location(scheme, bucketname, root, relpath)
+
+    def fromPath(self, path):
+        """Factory function to create a `Location` from a POSIX path.
+        Parameters
+        ----------
+        path : `str`
+            A standard POSIX path, relative to the `Datastore` root.
+        Returns
```
timj, line 239: Blank line before Returns and `------` after. (I think the original version has the same problem).

python/lsst/daf/butler/core/location.py
```diff
@@ -114,15 +110,171 @@ def fromUri(self, uri):
...
+            return self.fromUri('s3://' + os.path.join(self._bucket, self._datastoreRoot, path))
+        elif nEqual == 1:
+            # path contains the root dir or bucket
+            return self.fromUri('s3://' + os.path.join(self._bucket, path))
+        elif nEqual >1:
+            # path probably contains multiple datastoreRoots and/or buckets, or is completely broken
+            return self.fromUri('s3://'+ os.path.join(self._bucket, self._datastoreRoot, path))
+        else:
+            # an act of god
+            raise OSError("Don't understand path at all: {}".format(path))
```
timj, line 280: I think it's impossible to get to this else unless nEqual is not >= 0.  Since it starts at 0 and is only ever added to, I don't think this error message can ever be hit.

python/lsst/daf/butler/core/location.py
```diff
@@ -114,15 +110,171 @@ def fromUri(self, uri):
...
+        # and sometimes they give the pathInStore type path. Sometimes the bucket
+        # is provided as well. This can duplicate the datastoreRoot and bucket
+        # which we remove
+        dirs = path.lstrip('/').split('/')
+        nEqual = 0
+        # we shouldn't remove a sub-dir in the bucket that can have the of the same name
+        # so we require that bucketname/rootdir order is respected. A nicer way of writing
+        # this would be nice
+        for d in dirs:
+            if (d==self._bucket):
```
timj, line 257: If bucket can only ever be first in the list how about:
```python
if dirs[0] == self._bucket:
    dirs.pop(0)
```

python/lsst/daf/butler/core/location.py
```diff
@@ -114,15 +110,171 @@ def fromUri(self, uri):
...
+                nEqual += 1
+            else:
+                break
+
+        for d in dirs[nEqual:]:
```
timj, line 262:
```python
while dirs[0] == self._datastoreRoot:
    dirs.pop()
```
?
then you only need first fromUri call below don't you?

python/lsst/daf/butler/core/utils.py
```diff
@@ -23,12 +23,118 @@
            "getFullTypeName", "getInstanceOf", "Singleton", "transactional",
            "getObjectSize", "stripIfNotNone", "PrivateConstructorMeta")
 
+import os
 import sys
 import functools
+import boto3
```
timj, line 29: This import needs protecting since boto3 may not be available.

python/lsst/daf/butler/core/utils.py
```diff
@@ -23,12 +23,118 @@
            "getFullTypeName", "getInstanceOf", "Singleton", "transactional",
            "getObjectSize", "stripIfNotNone", "PrivateConstructorMeta")
```
timj, line 24: Add new functions to `__all__`.

tests/test_butler.py
```diff
@@ -541,6 +548,320 @@ def tearDown(self):
...
+                raise
+        else:
+            # key exists, remove it
+            bucket = s3.Bucket(self.bucketName)
+            bucket.objects.filter(Prefix=self.root).delete()
+
+        bucket = s3.Bucket(self.bucketName)
+        bucket.delete()
+
+    def testPutTemplates(self):
```
timj, line 622: This test is testing template expansion so it's probably okay to remove it here since the template infrastructure is being tested in PosixDatastore.  There's too much code duplication for my liking and I think that we would be unlucky if this test failed and PosixDatastore version passed (assuming the other put tests were working).

python/lsst/daf/butler/core/utils.py
```diff
@@ -23,12 +23,118 @@
...
 
 from lsst.utils import doImport
 
 
+def s3CheckFileExists(client, bucket, filepath):
+    """Checks that the file exists and if it does returns its size.
+
+    Parameters
+    ----------
+    client : 'boto3.client'
```
timj, line 41: I think these are the wrong quotes. Should be backticks.

python/lsst/daf/butler/core/utils.py
```diff
@@ -23,12 +23,118 @@
...
+        S3 Client object to query.
+    bucket : 'str'
+        Name of the bucket in which to look.
+    filepath : 'str'
+        Path to file.
+    """
```
timj, line 47: Please document the Returns.

python/lsst/daf/butler/core/utils.py
```diff
@@ -23,12 +23,118 @@
...
+    # https://github.com/boto/botocore/issues/1248
+    # https://github.com/boto/boto3/issues/1128
+    # https://stackoverflow.com/questions/33842944/check-if-a-key-exists-in-a-bucket-in-s3-using-boto3
+
+    # this has maxkeys kwarg, limited to 1000
+    response = client.list_objects_v2(
+        Bucket=bucket,
+        Prefix=filepath
+    )
+    # Hopefully multiple identical files will never exist?
```
timj, line 58: datastore.put is meant to check that something isn't already in the datastore using that name. (the template system is meant to be generating unique names).

python/lsst/daf/butler/core/utils.py
```diff
@@ -23,12 +23,118 @@
...
+    # if not then this returns list len 1 if found, this is charged for - worth it?
+    matches = [x for x in response.get('Contents', []) if x["Key"] == filepath]
+    if len(matches) == 1:
+        return (True, matches[0]['Size'])
+    else:
+        return (False, -1)
+
+
+# it would be great if this can be better
```
timj, line 67: Maybe remove this comment and add a section in `Notes` in docstring to explain the limitations?

python/lsst/daf/butler/core/utils.py
```diff
@@ -23,12 +23,118 @@
...
+def parsePath2Uri(path):
```
timj, line 68: The docstring is confusing given the name of the function.   It is documented to take a URI to parse but the argument is called a path. It's not returning a URI at all.  It's parsing the supplied string, that can be a URI or a path to a file.

python/lsst/daf/butler/core/utils.py
```diff
@@ -23,12 +23,118 @@
...
+    """If the path is a local filesystem path constructs elements of a URI.
```
timj, line 69: One sentence for function overview, then a blank line.

Something like:

"Parse a path or URI into component parts."

and then explain in a notes section or the URI parameter definition what the details are.

python/lsst/daf/butler/core/utils.py
```diff
@@ -23,12 +23,118 @@
...
+        URI to parse.
+
+    Returns
+    -------
+    scheme : 'str'
+        Either 'file://' or 's3://'.
+    root : 'str'
+        S3 Bucket name or Posix-like path to the top of the relative path
+    relpath : 'str'
+        Posix-like path relative to roothpath.
```
timj, line 84: Posix -> POSIX.
roothpath -> rootpath.

python/lsst/daf/butler/core/utils.py
```diff
@@ -23,12 +23,118 @@
...
 import sys
 import functools
+import boto3
+import urllib
+from urllib.parse import urlparse
 
 from lsst.utils import doImport
 
 
+def s3CheckFileExists(client, bucket, filepath):
```
timj, line 36: If the S3 routines are only used by S3Datastore then they should be moved there. If they are being used elsewhere then I think they should be in their own `s3utils.py` file.

python/lsst/daf/butler/core/utils.py
```diff
@@ -23,12 +23,118 @@
...
+    """
+    parsed = urlparse(path)
+    # if the parsed path is the supplied one - filesystem
+    if parsed.path == path:
+        scheme = 'file://'
+        if os.path.isabs(path):
+            root = '/'
+            relpath = parsed.path.lstrip('/')
+        else:
+            root = os.path.abspath(path).split(path)[0]
```
timj, line 94: This doesn't work in the general case (e.g., if path starts with `../`). You need something like:
```python
abspath = os.path.abspath(path)
cwd = os.path.abspath(os.path.curdir)
root = os.path.commonprefix([cwd, abspath])
relpath = abspath[len(root):]
```

python/lsst/daf/butler/core/utils.py
timj: This function needs some tests added for it.

python/lsst/daf/butler/core/utils.py
```diff
@@ -23,12 +23,118 @@
...
+        scheme = 's3://'
+        # this ends up being the bucketname
+        root = parsed.netloc
+        relpath = parsed.path.lstrip('/')
+    else:
+        raise urllib.error.URLError(f'Can not parse path: {path}')
+
+    return scheme, root, relpath
+
+def bucketExists(uri):
```
timj, line 106: The name is fine if it's in `s3utils` and not general utils package.

python/lsst/daf/butler/core/utils.py
```diff
@@ -23,12 +23,118 @@
...
+    scheme, root, relpath = parsePath2Uri(uri)
+
+    try:
+        client.get_bucket_location(Bucket=root)
+        # bucket exists, all is well
+        return True
+    except client.exceptions.NoSuchBucket:
+        return False
+
+#        # creating new buckets, generalizations hard?
```
timj, line 130: Do not include commented out code.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+
+import os
+import shutil
+import hashlib
+import logging
+from collections import namedtuple
+from urllib.parse import urlparse
+import tempfile
+
+import boto3
```
timj, line 34: This import is fine since we only import s3Datastore if we are using an s3Datastore (it's the other boto3 imports that are the ones that need protecting).

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+
+        # Name ourselves either using an explicit name or a name
+        # derived from the (unexpanded) root
+        if "name" in self.config:
+            self.name = self.config["name"]
+        else:
+            self.name = "S3Datastore@{}".format(self.config["root"])
+
+        # Support repository relocation in config
+        self.root = replaceRoot(self.config["root"], butlerRoot)
```
timj, line 130: This is the magic you need to sort out if you are enabling repo relocation in S3Datastore.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+
+        parsed = urlparse(self.root)
+        self.s3locationFactory = S3LocationFactory(parsed.netloc, parsed.path)
```
timj, line 133: Since you aren't using the other type of location factory, I think this attribute should be called `locationFactory` (although I see later that you are using internal API for this class so maybe that is not a good idea).

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+        # getting the bucket name from S3Location or somewhere else. After checking if
+        # the named bucket doesn't exist, ideally there would be a create flag(?) and a
+        # bucket could be created through DataStore for the user and no online clicking
+        # would be required - but generalizations are hard? Lot of options when opening.
+        self.client = boto3.client('s3')
+        self.bucket = self.s3locationFactory._bucket
+        try:
+            self.client.get_bucket_location(Bucket=self.bucket)
+            # bucket exsists - all is well
+            pass
```
timj, line 150: This `pass` can be removed.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+        # we don't need the self.bucket anywhere, EXCEPT in the test suite so I left it in
+        # but it should probably be removed at some point sooner or later, preferentially
+        # getting the bucket name from S3Location or somewhere else. After checking if
+        # the named bucket doesn't exist, ideally there would be a create flag(?) and a
+        # bucket could be created through DataStore for the user and no online clicking
+        # would be required - but generalizations are hard? Lot of options when opening.
+        self.client = boto3.client('s3')
+        self.bucket = self.s3locationFactory._bucket
+        try:
+            self.client.get_bucket_location(Bucket=self.bucket)
```
timj, line 148: Probably want a comment here to indicate that the even though it's called `get_bucket_location` we are not interested in the location itself.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+            client.create_bucket(ACL='authenticated-read',
+                                 Bucket=self.bucket,
+                                 LocationConstraint = {'LocationConstraint':'us-west-2'}
+            )
+
+        self.formatterFactory = FormatterFactory()
+        self.storageClassFactory = StorageClassFactory()
+
+        # Now associate formatters with storage classes
+        self.formatterFactory.registerFormatters(self.config["formatters"])
```
timj, line 161: The API for this has recently been simplified to:
```python
        self.formatterFactory.registerFormatters(self.config["formatters"],
                                                 universe=self.registry.dimensions)
```

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+        self.formatterFactory.normalizeDimensions(self.registry.dimensions)
+
+        # Read the file naming templates
+        self.templates = FileTemplates(self.config["templates"])
+        self.templates.normalizeDimensions(self.registry.dimensions)
```
timj, line 166: Similarly these two lines are now one:
```python
        self.templates = FileTemplates(self.config["templates"],
                                       universe=self.registry.dimensions)
```

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+
+        # Name ourselves
+        self.name = "S3Datastore@{}".format(self.root)
```
timj, line 169: You have already named yourself above.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+        # instantiation. By default it will intialize to boto3.Session('DEFAULT')
+        # profile in ~/.aws/credentials. This needs to be better. Realistically also
+        # we don't need the self.bucket anywhere, EXCEPT in the test suite so I left it in
+        # but it should probably be removed at some point sooner or later, preferentially
+        # getting the bucket name from S3Location or somewhere else. After checking if
+        # the named bucket doesn't exist, ideally there would be a create flag(?) and a
+        # bucket could be created through DataStore for the user and no online clicking
+        # would be required - but generalizations are hard? Lot of options when opening.
+        self.client = boto3.client('s3')
+        self.bucket = self.s3locationFactory._bucket
```
timj, line 146: Why is that not using `bucket` property?

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+                raise ValueError("Failure from formatter for Dataset {}: {}".format(ref.id, e))
+
+            return result, fileDescriptor
+
+        def downloadFile():
+            """Downloads a dataset from an object storage into a temporary file from which
+            it is read as an appropriate python object.
+            """
+            with tempfile.TemporaryDirectory() as tmpRootDir:
+                datasetName = location.path.split('/')[-1]
```
timj, line 347: Splitting a path manually on `/` seems risky to me. `os.path.basename(location.path)` seems safer.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+                if err['Error']['Code'] == '404':
+                    raise FileNotFoundError('Could not find dataset {}'.format(location.uri)) from err
+
+            pickledDataset = response['Body'].read()
+            fileDescriptor = FileDescriptor(location, readStorageClass=readStorageClass,
+                                            storageClass=writeStorageClass, parameters=parameters)
+            try:
+                result = formatter.fromBytes(pickledDataset, fileDescriptor, component=component)
+            except Exception as e:
+                raise ValueError("Failure from formatter for Dataset {}: {}".format(ref.id, e))
```
timj, line 338: I think f-strings would be clearer, and consider saying `from e` at the end here so that it is clear you are augmenting the error message and not raising an entirely new exception.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+            """Downloads a dataset from an object storage as a bytestring and returns it as an
+            appropriate python object.
+            """
+            try:
+                response = self.client.get_object(Bucket=location.bucket, Key=location.path)
+            except self.client.exceptions.ClientError as err:
+                if err['Error']['Code'] == '404':
+                    raise FileNotFoundError('Could not find dataset {}'.format(location.uri)) from err
+
+            pickledDataset = response['Body'].read()
```
timj, line 332: Not necessarily pickled. I think `serializedDataset` might be clearer name.

python/lsst/daf/butler/formatters/yamlFormatter.py
```diff
@@ -61,6 +62,38 @@ def _readFile(self, path, pytype=None):
...
+            # its either my mis-use of yaml or intended behaviour but yaml wants
+            # to return a dictionary sometimes and an actual object other times.
+            # Later FileFormatter assembles and checkes if only one of objects
+            # components was requested. Pickle, however, always gets back the object.
+            # I didn't want to figure out what is involved in the two different cases
+            # so I return all my yamls as dicts and let the assembler figure it out
+            pass
+        return data
+
     def _writeFile(self, inMemoryDataset, fileDescriptor):
```
timj, line 97: `_writeFile` should probably be reimplemented in terms of `_toBytes` so that things are internally consistent. Possibly `_readFile`/`_fromBytes` as well.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+        # is the accepted style here.
+        def downloadBytes():
+            """Downloads a dataset from an object storage as a bytestring and returns it as an
+            appropriate python object.
+            """
+            try:
+                response = self.client.get_object(Bucket=location.bucket, Key=location.path)
+            except self.client.exceptions.ClientError as err:
+                if err['Error']['Code'] == '404':
+                    raise FileNotFoundError('Could not find dataset {}'.format(location.uri)) from err
```
timj, line 330: I think you need to reraise `err` if it's not a 404 don't you? Otherwise a failure will occur in the next line.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+                    fileDescriptor = FileDescriptor(tmpLoc, readStorageClass=readStorageClass,
+                                                    storageClass=writeStorageClass, parameters=parameters)
+                    try:
+                        result = formatter.read(fileDescriptor, component=component)
+                    except Exception as e:
+                        raise ValueError("Failure from formatter for Dataset {}: {}".format(ref.id, e))
+
+            return result, fileDescriptor
+
+        if hasattr(formatter, '_fromBytes'):
```
timj, line 366: `_fromBytes` is an implementation detail of the file formatter implementation so checking this internal attribute limits the applicability of the bytes code path. 

I was envisaging that unimplemented bytes formatters would raise `NotImplementedError` but clearly that's too late in the game since you need to know whether you are calling an API to pull bytes directly into memory or calling an API to write the bytes first to a file (which presumably is done in a memory efficient manner).  I think this means that we may well have to stop trying to be clever and instead use a class property along the lines of `Formatter.canUseBytes` that is checked here instead.  That's probably preferable to making `_fromBytes` and `_toBytes` part of the public API and insisting they are set to `None` when not implemented.  For the proof of concept the current approach will work.

Regardless, the `Formatter` abstract base class needs to have `fromBytes` and `toBytes` added to it so that formatters that are not inheriting from FileFormatter can still work with S3 and be optimized for it. They should not be abstract but should throw `NotImplementedError` in the base class.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+
+                # removing the tmpdir not neccesary? Just a rollback
+                with self.transaction() as transaction:
+                    self.client.download_file(location.bucket, location.path, tmpLoc.path)
+                    fileDescriptor = FileDescriptor(tmpLoc, readStorageClass=readStorageClass,
+                                                    storageClass=writeStorageClass, parameters=parameters)
+                    try:
+                        result = formatter.read(fileDescriptor, component=component)
+                    except Exception as e:
+                        raise ValueError("Failure from formatter for Dataset {}: {}".format(ref.id, e))
```
timj, line 362: f-strings preferred I think (which I know is not in the original posixDatastore) and `from e` at the end.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+        def downloadFile():
+            """Downloads a dataset from an object storage into a temporary file from which
+            it is read as an appropriate python object.
+            """
+            with tempfile.TemporaryDirectory() as tmpRootDir:
+                datasetName = location.path.split('/')[-1]
+                tmpLoc = Location(tmpRootDir, datasetName)
+
+                if os.path.exists(tmpLoc.path):
+                    raise FileExistsError(f"Cannot write file for ref {ref} as "
```
timj, line 351: This test is presumably unnecessary since you have asked for a temporary directory to be created. `predictedFullPath` is not relevant here.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+            except Exception as e:
+                raise ValueError("Failure from formatter for Dataset {}: {}".format(ref.id, e))
+
+            return result, fileDescriptor
+
+        def downloadFile():
+            """Downloads a dataset from an object storage into a temporary file from which
+            it is read as an appropriate python object.
+            """
+            with tempfile.TemporaryDirectory() as tmpRootDir:
```
timj, line 346: Are you using a temporary directory rather than temporary file name in order to make it easier to debug (since the file name will be recognizable still).

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+                datasetName = location.path.split('/')[-1]
+                tmpLoc = Location(tmpRootDir, datasetName)
+
+                if os.path.exists(tmpLoc.path):
+                    raise FileExistsError(f"Cannot write file for ref {ref} as "
+                                          f"output file {predictedFullPath} already exists")
+
+                # removing the tmpdir not neccesary? Just a rollback
+                with self.transaction() as transaction:
```
timj, line 355: why a transaction here?

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+        """
+        datasetType = ref.datasetType
+        storageClass = datasetType.storageClass
+
+        # Sanity check
+        if not isinstance(inMemoryDataset, storageClass.pytype):
+            raise TypeError("Inconsistency between supplied object ({}) "
+                            "and storage class type ({})".format(type(inMemoryDataset),
+                                                                 storageClass.pytype))
+
```
timj, line 408: The new constraints checking code has to be added here to support per-datastore control of whether to accept specific datasets.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+            raise DatasetTypeNotSupportedError(f"Unable to find formatter for {ref}") from e
+
+        location.updateExtension(formatter.extension)
+        if s3CheckFileExists(self.client, location.bucket, location.path)[0]:
+            raise FileExistsError(f"Cannot write file for ref {ref} as "
+                                  f"output file {location.uri} already exists")
+
+        # unlike get, the existance of the bucket has been verified at init and ButlerConfig
+        # downloads - therefore not a lot of error handling is needed?
+        if hasattr(formatter, '_toBytes'):
```
timj, line 430: In this particular case, since we are doing the `toBytes` call early, it's entirely reasonable to put it in a `try` block and call `toBytes` and catch `NotImplementedError` before reverting to file formatter (the transaction would be above them both).

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+            with self.transaction() as transaction:
+                fileDescriptor = FileDescriptor(location, storageClass=storageClass)
+                pickledDataset = formatter.toBytes(inMemoryDataset,  fileDescriptor)
+                self.client.put_object(Bucket=location.bucket, Key=location.path, Body=pickledDataset)
+            log.debug("Wrote file to %s", location.uri)
```
timj, line 435: Maybe say "wrote dataset directly to %s" to indicate that you didn't go through a temporary file.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+        if s3CheckFileExists(self.client, location.bucket, location.path)[0]:
+            raise FileExistsError(f"Cannot write file for ref {ref} as "
+                                  f"output file {location.uri} already exists")
+
+        # unlike get, the existance of the bucket has been verified at init and ButlerConfig
+        # downloads - therefore not a lot of error handling is needed?
+        if hasattr(formatter, '_toBytes'):
+            with self.transaction() as transaction:
+                fileDescriptor = FileDescriptor(location, storageClass=storageClass)
+                pickledDataset = formatter.toBytes(inMemoryDataset,  fileDescriptor)
```
timj, line 433: Not pickled.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+                self.client.put_object(Bucket=location.bucket, Key=location.path, Body=pickledDataset)
+            log.debug("Wrote file to %s", location.uri)
+        else:
+            # we can't change location attrs, so we need to make up a new one that we clean up afterwards.
+            with tempfile.TemporaryDirectory() as tmpRootDir:
+                datasetName = location.path.split('/')[-1]
```
timj, line 439: Why not `os.path.basename`?  I do agree that in general URIs are not paths but in this case we are retrieving the path component of the URI so it doesn't seem unreasonable to use basename.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+                tmpLoc = Location(tmpRootDir, datasetName)
+                # this should never happen
+                if os.path.exists(tmpLoc.path):
+                    raise FileExistsError(f"Cannot write file for ref {ref} as "
+                                          f"output file {tmpLoc.uri} already exists")
+                with self.transaction() as transaction:
+                    tmpPath = formatter.write(inMemoryDataset, FileDescriptor(tmpLoc, storageClass=storageClass))
+                    assert tmpLoc.path == os.path.join(tmpRootDir, tmpPath)
```
timj, line 447: Is this a debug check?

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+            Raised if the file at ``path`` does not exist.
+        FileExistsError
+            Raised if ``transfer is not None`` but a file already exists at the
+            location computed from the template.
+        """
+        if formatter is None:
+            formatter = self.formatterFactory.getFormatter(ref)
+
+        # if I understood the point of ingest correctly - put is for putting python objects in store,
+        # ingest is for putting file-like stuff in store. That makes an undefined and 4 defined cases:
```
timj, line 498: Yes. We do need to ingest raw data (which may not have a fully compatible python in-memory equivalent) and some calibration files directly into a datastore without having to go through a Python object.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+                # someone wants to ingest a local file, but not transfer it to object storage
+                abspath = os.path.join(root, relpath)
+                raise RuntimeError(("'{}' is not inside repository root '{}'. Ingesting local data to"
+                                    " S3Datastore without upload to S3 not allowed.").format(abspath, self.root))
+            if scheme == 's3://':
+                # if both transfer is None and scheme s3, file was already uploaded in put
+                # rootdir is sometimes sent as s3://bucketname/root and sometimes s3://bucketname/root/
+                # There is no equivalent of os.isdir/os.isfile for the S3 - problem for parsing.
+                # The string comparisons will fail since one of them will have the '/' and the other not.
+                #I hope I just don't know how to use urllib or something because this is crap.
```
timj, line 525: The problem seems to be that firstly, there is no standard python class for manipulating the path component of a URI (`furl.Path` seems to try).  In this case URI parsers really do want to treat `root` as a file and `root/` as a directory so they insist that they are different.  I think you have to understand why you are getting the inconsistency that requires these manipulations.  Why can't you normalize `self.root` in the constructor so that it always has the trailing slash to indicate that it's a directory?

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+            absolute.
+        ref : `DatasetRef`
+            Reference to the associated Dataset.
+        formatter : `Formatter` (optional)
+            Formatter that should be used to retreive the Dataset.  If not
+            provided, the formatter will be constructed according to
+            Datastore configuration.
+        transfer : str (optional)
+            If not None, must be one of 'move', 'copy', 'hardlink', or
+            'symlink' indicating how to transfer the file.  The new
```
timj, line 478: This needs fixing to remove hardlink and symlink. I'm not sure that `upload` is needed as such since that is implied by `move` or `copy`.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+                # rootdir is sometimes sent as s3://bucketname/root and sometimes s3://bucketname/root/
+                # There is no equivalent of os.isdir/os.isfile for the S3 - problem for parsing.
+                # The string comparisons will fail since one of them will have the '/' and the other not.
+                #I hope I just don't know how to use urllib or something because this is crap.
+                rootSchema, bucketname, rootDir = parsePath2Uri(self.root)
+                topDir = relpath.split('/')[0] + '/'
+                rootDir = rootDir + '/' if rootDir[-1] != '/' else rootDir
+                if (bucketname != root) or (rootDir != topDir):
+                    raise RuntimeError("'{}' is not inside repository root '{}'".format(path, self.root))
+        elif transfer == 'upload' or transfer == 'copy':
```
timj, line 531: `move` should be allowed. It requires the file to be deleted after upload.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+            if scheme == 'file://':
+                # reuploads not allowed?
+                if s3CheckFileExists(self.client, root, relpath)[0]:
+                    raise FileExistsError("File '{}' already exists".format(path))
+
+                template = self.templates.getTemplate(ref)
+                location = self.s3locationFactory.fromPath(template.format(ref))
+                location.updateExtension(formatter.extension)
+
+                with self.transaction() as transaction:
```
timj, line 541: `transaction` is not used.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+                    self.client.upload_file(path, location.bucket, location.path)
+            if scheme == 's3://':
+                # ingesting is done from another bucket - not tested
+                if s3CheckFileExists(self.client, root, relpath)[0]:
+                    fullpath = os.path.join(root, relpath)
+                    raise FileExistsError("File '{}' already exists".format(schema+fullpath))
+
+                copySrc = {'Bucket': root, 'Key': relpath}
+                self.client.copy(copySrc, self.bucket, relpath)
```
timj, line 550: Should this be a transaction (I admit I don't know what happens to the S3 state if the copy fails).

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+        else:
+            raise NotImplementedError("Transfer type '{}' not supported.".format(transfer))
+
+        path = os.path.join(self.bucket, relpath)
+        location = self.s3locationFactory.fromPath(path)
+
+        # Create Storage information in the registry
+        # TODO: fix the checksum calcs already
+        #checksum = self.computeChecksum(fullPath)
```
timj, line 559: Checksum can work for the file you have lying around but it's too late to work out it out for the bytes that you uploaded in `put`.  Presumably we could calculate the checksum of the bytes and pass it in as an optional argument to `ingest`, only dealing with files here.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+
+        # the file should exist on the bucket by now
+        exists, size = s3CheckFileExists(self.client, location.bucket, relpath)
```
timj, line 562: Is the size of the S3 object different to the size of the file that was uploaded? Or since we are asking for the size from S3 can we check that it's the same as the size of the thing we ingested? Does that matter?

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+        # Get file metadata and internal metadata
+
+        try:
+            storedFileInfo = self.getStoredFileInfo(ref)
+        except KeyError:
+            raise FileNotFoundError("Requested dataset ({}) does not exist".format(ref))
+        location = self.s3locationFactory.fromPath(storedFileInfo.path)
+        if not s3CheckFileExists(self.client, location.bucket, location.path ): # os.path.exists(location.path):
+            raise FileNotFoundError("No such file: {0}".format(location.uri))
+        self.client.delete_object(Bucket=location.bucket, Key=location.path)
```
timj, line 654: Is it more efficient to call the delete_object directly and catch the exception rather than check and then delete?  Attempting something and catching is usually thought of as being the safe approach if race conditions are possible (something else trying to delete it just after you've deleted it).

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+        -------
+        hexdigest : `str`
+            Hex digest of the file.
+        """
+        if algorithm not in hashlib.algorithms_guaranteed:
+            raise NameError("The specified algorithm '{}' is not supported by hashlib".format(algorithm))
+
+        hasher = hashlib.new(algorithm)
+
+        tmpdir = "/home/dinob/uni/lsstspark/simple_repo/s3_repo"
```
timj, line 766: This doesn't seem to be something we should be committing.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,781 @@
...
+        scheme, root, relpath = parsePath2Uri(filename)
+        dirpath, name = os.path.split(relpath)
+        potentialloc = os.path.join(tmpdir, name)
+        if os.path.exists(potentialloc):
+            filename = potentialloc
+        else:
+            client = boto3.client('s3')
+            client.download_file(root, relpath, potentialloc)
```
timj, line 774: Does S3 have an API for you to grab bytes as they come from the server without having to store bytes all in memory? ie can you do "give me N bytes at a time" so that you don't need to download to a temp file. Where is this writing the file?

timj (changes requested):
@DinoBektesevic thanks for the pull request.  I have many comments I'm afraid.  Some of these are general problems that I probably need to fix later (such as the code duplication between S3Datastore and PosixDatastore) but most are issues that do need to be fixed before we can merge.

python/lsst/daf/butler/registries/postgresqlRegistry.py
DinoBektesevic: I believe this question was raised in Slack where the response was something down the lines of "Oracle id tokens" but now I can't find it. Ergo the configparse work-around until someone tells me how to do it correctly.

tests/test_butler.py
DinoBektesevic: String and random are builtins I believe, do I have to skip them too?

python/lsst/daf/butler/datastores/s3Datastore.py
DinoBektesevic: The sizes should be the same. I need the size to create the `StoredFileInfo` two lines down I believe. This is where checksum creation is missing. The file-info is then added to the registry couple of lines down in `self.addStoredFileInfo(ref, fileInfo)`. File-size is later compared to stored file size on retrieval.  So I'm not quite sure what I'm being asked here. We do want at least minimal consistency checks, correct?

python/lsst/daf/butler/datastores/s3Datastore.py
DinoBektesevic: This is all kinds of my-mistakes here. I thought I removed everything from the checksum experimentations but here we are. There is a `StreamingBody` object in boto3 I can try and utilize for this purpose.

tests/test_butler.py
timj: No. Sorry for the confusion. It's boto3 that needs to be protected.

python/lsst/daf/butler/datastores/s3Datastore.py
timj: Consistency checks are good. It's probably a good idea to compare the size from S3 with the size of the thing you uploaded.

PR #148, opened 2019-04-23
DM-19373: Add lsst.meas.algorithms.Defects to StorageClass and formatter

timj:
Uses the normal FITS catalog formatter since writeFits and
readFits work properly.


timj:
Depends on lsst/meas_algorithms#161

czwa (approved):

PR #149, opened 2019-04-25
DM-19485: Change instrument length in schema from 8 to 16

timj:


timj:
Or should I use 32?

TallJimbo (approved):
Either 16 or 32 is fine with me.

PR #150, opened 2019-04-26
DM-17663: make table names lowercase

TallJimbo:


python/lsst/daf/butler/sql/singleDatasetQueryBuilder.py
```diff
@@ -236,14 +236,14 @@ def _columns(selectable, names):
...
 
-        datasetTable = registry._schema.tables["Dataset"]
-        datasetCollectionTable = registry._schema.tables["DatasetCollection"]
+        datasetTable = registry._schema.tables["dataset"]
+        datasetCollectionTable = registry._schema.tables["dataset_collection"]
 
         # full set of link names for this DatasetType
         links = list(datasetType.dimensions.links())
 
         # Starting point for both subqueries below: a join of Dataset to
```
timj: Probably should be "dataset" here...

tests/test_queryBuilder.py
```diff
@@ -11,7 +11,7 @@
 # the Free Software Foundation, either version 3 of the License, or
 # (at your option) any later version.
 #
-# This program is distributed in the hope that it will be useful,
+# This program is distributed in the hope that it will be useful,`
```
timj: Not sure what happened here but the back tick should be removed.

tests/test_queryBuilder.py
```diff
@@ -59,34 +59,34 @@ def testDatasetOriginInfoDef(self):
         self.assertEqual(originInfo.getOutputCollection("ds2"), "out2")
 
     def testInstrumentDimensions(self):
-        """Test involving only Instrument dimensions, no joins to SkyMap"""
+        """Test involving only instrument dimensions, no joins to skymap"""
```
timj, line 62: SkyMap -> skymap and not sky_map seems anomalous but I have no problem with skymap.

timj (approved):

tests/test_queryBuilder.py
TallJimbo: Yeah, it's a little inconsistent, but I think there's a bit of precedent for it and people will find this a bit less jarring.

(More generally, I recently came to the sad conclusion that even within pure camelCase, our brains just refuse to make the same word boundaries in different contexts - we almost always write `BBox` when the first "B" is capitalized, but *always* write `bbox` instead of `bBox` when the first "b" is lowercase, for example.)

timj: and then we have the whole camelCase insistence that "LSST" become "Lsst"... 😄

PR #157, opened 2019-05-08
DM-19671: Allow setConfigRoot to retain input values

timj:
Adjust setConfigRoot such that it will optionally not override values in the supplied config.  This is the common use case from makeRepo where an external config would like to set a value for root, but that value was always being overridden.

Update makeRepo to disable overwrite.

### DIFF
--- a/python/lsst/daf/butler/core/config.py
+++ b/python/lsst/daf/butler/core/config.py
@@ -731,13 +731,13 @@ class Config(collections.abc.MutableMapping):
             self.dump(f)
 
     @staticmethod
-    def overrideParameters(configType, config, full, toUpdate=None, toCopy=None, overwrite=True):
-        """Generic helper function for overriding specific config parameters
+    def updateParameters(configType, config, full, toUpdate=None, toCopy=None, overwrite=True):
+        """Generic helper function for updating specific config parameters.
 
         Allows for named parameters to be set to new values in bulk, and
         for other values to be set by copying from a reference config.
 
-        Assumes that the supplied config is compatible with ```configType``
+        Assumes that the supplied config is compatible with ``configType``
         and will attach the updated values to the supplied config by
         looking for the related component key.  It is assumed that
         ``config`` and ``full`` are from the same part of the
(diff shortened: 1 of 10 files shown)


python/lsst/daf/butler/core/config.py
```diff
@@ -769,6 +769,9 @@ def overrideParameters(configType, config, full, toUpdate=None, toCopy=None):
         toCopy : `tuple`, optional
             `tuple` of keys whose values should be copied from ``full``
             into ``config``.
+        overwrite : `bool`, optional
```
andy-slac, line 772: I think it's a bit confusing for method named `overrideParameters` to take an argument which says "don't override".  Maybe rename it to `updateParameters` if it's not too much work? (Or maybe it's only confusing to me, I have not looked at this code much?)

python/lsst/daf/butler/core/registry.py
```diff
@@ -115,9 +115,19 @@ def setConfigRoot(cls, root, config, full):
             Repository-specific options that should not be obtained
             from defaults when Butler instances are constructed
             should be copied from `full` to `Config`.
```
andy-slac: It's not your change but can be fixed if you wish - ` ``full`` and ``config`` `?

python/lsst/daf/butler/core/config.py
```diff
@@ -769,6 +769,9 @@ def overrideParameters(configType, config, full, toUpdate=None, toCopy=None):
...
+            If `False`, do not modify a value in ``config`` if the value
+            already exists.  Default is always to overwrite.
```
andy-slac, line 774: "the key already exists" may be more correct, but I'm not going to insist.

andy-slac (approved):
Looks OK to me, few minor comments. Docstrings for `setConfigRoot` seem to literally repeat in few classes, not sure we can do anything about it.

python/lsst/daf/butler/core/config.py
timj: Yes, I was worrying about that too.  updateParameters sounds better.

PR #159, opened 2019-05-17
Add support for S3 Datastore

DinoBektesevic:
As was suggested to me by @r-owen and @timj the draft PR for `S3Datastore` and `PostgreSqlRegistry` together was too large and clunky to be reasonable. Draft PR https://github.com/lsst/daf_butler/pull/147 is now closed and substituted by two PRs, separating `S3Datastore` and `PostgreSqlRegistry`.

### Changes
Based on review comments from the Draft-PR the following was changed compared to the draft PR. 

* Reworked how to/from/Bytes method work
    - Added `toBytes` and `fromBytes` functionality to `JsonFormatter`
    - Changed the `toFile` and `fromFile` methods to internally use the `to/from/Bytes` on all formatters that support it
    - `S3Datastore` now downloads all files as bytes and attempts to read them. If it fails it stores them in a temporary file and then attempts to read them from there. 
* Refactored formatters section in datastore YAML config files into their own file that is included into datastore config files. 
* Removed anything RDS related out of this PR.
* `Config` class was made aware of S3 and the code in `butlerConfig` and `Butler.makeRepo` that was used as a work-around to that was removed. 
* `Butler.makeRepo` was re-factored back to (almost) the way it was and is now pretty again
* Removed the `Butler.computeChecksum` as I never made it to work.
* Moved all S3-related utilities from `daf.butler.core.utils` to `daf.butler.core.s3utils`
* Added tests for `s3utils`.
* Fixed up the de-duplication hacks in `S3Location` by fixing `parsePathToUriElements`
* General housekeeping (better names, formatters are now their own yaml file included where needed, removed old unnecessary comments, fixed docstrings, added more docstrings, new default and test config files for S3, better style for `boto3` imports protection, mocks and skips for tests when `moto` or `boto3` is missing...)

### Unresolved/Other

* `<butlerRoot>` tag is still not supported for the current `S3Datastore` as, from what I can see, the main effect of the butler relocation code is that `sqlite:///:memory:` registry location is replaced by `'sqlite:///<butlerRoot>/gen3.sqlite3` which just does not make practical sense when working with `S3Datastore`. We can not talk to sqlite DB in a Bucket, while, for `S3Datastore` tests at least, it does make sense to have an ephemeral Registry. This makes sense for cases when the `db` string in the registry YAML config can be specified directly and won't be changed, so I'm deferring this issue for later.
[...]

PR #160, opened 2019-05-17
Test fix of Location and LocationFactory.

DinoBektesevic:
If I run the following code

```
import lsst.daf.butler as dafButler

LocationFactory = dafButler.LocationFactory('datastoreRoot')

absloc = locationFactory.fromPath('/absolute/relative/file.ext')
relloc = locationFactory.fromPath('relative/file.ext')

print('path         ', absloc.path)
print('pathInStore  ', absloc.pathInStore)
print('uri          ', absloc.uri)
print('str          ', absloc)

print()
print('path         ', relloc.path)
print('pathInStore  ', relloc.pathInStore)
print('uri          ', relloc.uri)
print('str          ', relloc)

absloc = locationFactory.fromUri('file:///absolute/relative/file.ext')
relloc = locationFactory.fromUri('file://relative/file.ext')

print('\n\n')
print('path         ', absloc.path)
print('pathInStore  ', absloc.pathInStore)
print('uri          ', absloc.uri)
print('str          ', absloc)

print()
print('path         ', relloc.path)
print('pathInStore  ', relloc.pathInStore)
print('uri          ', relloc.uri)
print('str          ', relloc)
```

I get the following outputs
```
path          datastoreRoot/absolute/relative/file.ext
pathInStore   absolute/relative/file.ext
uri           file:///absolute/relative/file.ext
str           file:///absolute/relative/file.ext

path          datastoreRoot/relative/file.ext
pathInStore   relative/file.ext
uri           file:///relative/file.ext
str           file:///relative/file.ext
```
and
```
path          datastoreRoot/absolute/relative/file.ext
pathInStore   absolute/relative/file.ext
uri           file:///absolute/relative/file.ext
str           file:///absolute/relative/file.ext

path          datastoreRoot/file.ext
pathInStore   file.ext
uri           file://relative/file.ext
str           file://relative/file.ext
```

The problem is assuming that `file://relative/file.ext` is a valid URI. It seems to me from the behaviour of `urllib` that this is not correct. The following are correct `file://` URI's:
```
file://localhost/absolute/relative/file.ext
file:///absolute/relative/file.ext
```
The following URIs seem to be "incorrect"
```
file://relative/file.ext
file://file.ext`
```
This is the reason why `pathInStore` is returned differently depending on whether the `Location` was created by using `fromUri` or `fromPath`. The `relative` is interpreted as `localhost` instead.

This PR is a proposition of a solution to this that enables the `Location` to treat `file://relative` and `file:///absolute/relative...` paths differently by assuming that the `datastoreRoot` provided to the `LocationFactory` is either an absolute path or a path relative to current directory and that the `Location` path is then relative to that. I do not know if this is desired behaviour or not but it seems to follow the logic seen elsewhere in `daf_butler`. Naturally, checks for consistent root locations when dealing with absolute paths are now mandatory. 

Comments welcome.

### DIFF
--- a/python/lsst/daf/butler/core/location.py
+++ b/python/lsst/daf/butler/core/location.py
@@ -51,7 +51,7 @@ class Location:
 
         This path includes the root of the `Datastore`.
         """
-        return  self._uri.path
+        return self._uri.path
 
     @property
     def pathInStore(self):
@@ -92,8 +92,8 @@ class LocationFactory:
         Parameters
         ----------
         datastoreRoot : `str`
-            Root location of the `Datastore` in the filesystem. Assumed relative
-            to current directory unless absolute.
+            Root location of the `Datastore` in the filesystem. Assumed
+            relative to current directory unless absolute.
         """
         self._datastoreRoot = os.path.abspath(os.path.expanduser(datastoreRoot))
 
@@ -114,10 +114,18 @@ class LocationFactory:
 
         parsed = urllib.parse.urlparse(uri)
         if parsed.scheme == 'file' and parsed.netloc:
-            # should we try and guess if we are working with path intended to be
-            # relative to self._datastoreRoot or just treat it as an error?
-            path = os.path.join(self._datastoreRoot, parsed.netloc, parsed.path.lstrip(os.sep))
-            uri = urllib.parse.urlunparse((parsed.scheme, '', path, '', '', ''))
+            # should we try and guess if we are working with path intended to
+            # be relative to self._datastoreRoot or just treat it as an error?
+            # When the path is just a filename relative to curdir it gets
+            # registered as a netloc, then joining root, netloc and path
+            # produces a '/' at the end we don't want but can't remove because
+            # perhaps it's a dir?
+            if not parsed.path:
+                path = os.path.join(self._datastoreRoot, parsed.netloc)
+            else:
+                path = os.path.join(self._datastoreRoot, parsed.netloc, parsed.path.lstrip(os.sep))
+            # scheme, netloc, path, query, fragment, user, pass, host, port
+            uri = urllib.parse.urlunparse((parsed.scheme, '', path, '', '', parsed.fragment))
             parsed = urllib.parse.urlparse(uri)
         if os.path.commonprefix((parsed.path, self._datastoreRoot)) != self._datastoreRoot:
             raise ValueError((f'URI {uri} does not share the same datastore root '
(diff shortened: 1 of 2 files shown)


timj:
I have a general comment on path manipulations in URLs. It seems like Python recommends that when you are manipulating the path from a URL you explicitly use `posixpath`.  From the docs:

>  Some of this can actually be useful on non-Posix systems too, e.g. for manipulation of the pathname component of URLs.

So for maximum portability we have to be careful and use `os.path` (and `os.sep`) when we are dealing with filenames in a local datastore, and `posixpath` (and `posixpath.sep`) when dealing with URLs.

DinoBektesevic:
I changed all URI path manipulation to explicitly use `posixpath` module. @ktlim told me about this but I forgot. 

I should also note that this then requires me to fix the `S3Location` in PR https://github.com/lsst/daf_butler/pull/159 to only ever explicitly use POSIX paths as it could be confusing if `S3Location.pathInStore` returns a Windows-style path but `S3Location.uri` the correct `s3://` uri. I don't believe that this should be done here as `Location` might refer to a local path that would then be OS dependent? I am known to make mistakes, however.

EDIT: I ran ahead of myself again. `Location.fromPath`, as is, won't change `/` for `\` when it receives a different style path. Should I add a `replace` there to make sure the path uses `\` in there for good measure?

timj:
I've looked around the [web a bit](https://tools.ietf.org/html/rfc3986#section-3.3) and relative URIs for files are meant to be written as `file:relative/path.ext`:

```python
>>> urllib.parse.urlparse("file:relative/path.ext")
ParseResult(scheme='file', netloc='', path='relative/path.ext', params='', query='', fragment='')
```
The problem is that when urllib puts that back together it doesn't round trip:
```python
>>> urllib.parse.urlparse("file:relative/path.ext").geturl()
'file:///relative/path.ext'
>>> urllib.parse.urlparse(urllib.parse.urlparse("file:///relative/path.ext").geturl())
ParseResult(scheme='file', netloc='', path='/relative/path.ext', params='', query='', fragment='')
```
indicating that `urllib` is not fit for purpose. I wonder if we are meant to use a different url manipulation library.

timj:
I've looked at the furl package and the uri package and they both fail to handle relative path URIs. I've filed issues as marrow/uri#11 and gruns/furl#115

timj:
@DinoBektesevic Reading this PR and #159 I think I need to play with this branch a bit. I worry that there are lots of hacks hidden around the place (and in #159) stemming from less than consistent thinking inside `Location` and `LocationFactory` (in particular the completely wrong line at the end using `file://` with a relative path).

timj:
I am closing this PR in favor of #167

PR #161, opened 2019-05-17
Add support for PostgreSQL Registry

DinoBektesevic:
As was suggested to me by @r-owen and @timj the draft PR for S3Datastore and PostgreSqlRegistry together was too large and clunky to be reasonable. That Draft PR #147 is now closed. Since all the comments I received about the `PostgreSqlRegistry` implementation were down the line "please remove it" I am opening a new draft PR request separate to that of `S3Datastore`. 

For the most part the inclusion of a `PostgreSqlRegistry` seems equivalent/analogous to `OracleRegistry` with the exception of authorization. Specifically there are 2 authorizations that need to happen, one is to the AWS service, in the form of security policies and groups (f.e. if your IP is not white-listed outside source you will be denied access to the DB, or if your IAM  is not permitted access etc..) and the second one is authorizing to the database itself. The latter is currently done by reading the credentials from `~/.rds/credentials` file and then substituting the correct username and password according to the nick given in the connection string. 

A `PosixDatastore` with an RDS backed `PostgresSqlRegistry` test was left in because I have noticed certain slowness when running the tests that I have not seen for Sqlite registry. I am willing to make arrangements for access to people that are willing to try it out. Additionally, it outlines the neccessary events that would have to occur if actual `PostgreSqlRegistry` would be tested (I hope). 

Comments and suggestions welcomed.


tests/test_butler.py
```diff
@@ -542,5 +544,163 @@ def tearDown(self):
...
+    configFile = os.path.join(TESTDIR, "config/basic/butler-rds.yaml")
+    fullConfigKey = ".datastore.formatters"
+    validationCanFail = True
+
+    # there are too many important parameters in the db string that need to
+    # carry over to here, so instead of making people copy-paste it on every change
+    # read it from config and re-format it to something useful. Caveat: the RDS name can,
+    # but does not have to be, the same as the dbname, we want to avoid replacing anything
+    # except the dbname that's why we manually add the name at the end instead of replace().
+    config =  Config(configFile)
```
timj: If you don't care about the non-registry bits, I think it would be clearer to use `config = RegistryConfig(configFile, mergeDefaults=False)` here -- that would resolve the double registry key problem below.

python/lsst/daf/butler/registries/postgresqlRegistry.py
```diff
@@ -0,0 +1,69 @@
...
+from sqlalchemy import event
+from sqlalchemy import create_engine
+
+from lsst.daf.butler.core.config import Config
+from lsst.daf.butler.core.registry import RegistryConfig
+
+from .oracleRegistry import OracleRegistry
+from .sqlRegistry import SqlRegistryConfig
+
+class PostgreSqlRegistry(OracleRegistry):
```
timj: I think it might be confusing to continue to inherit from `OracleRegistry` since two of the methods are replaced and the only remaining one is `setConfigRoot`.  I think it would be better to have an explicit `setConfigRoot` here instead of inheriting from Oracle but maybe @TallJimbo has an opinion. I imagine it depends on whether `deferDatasetIdQueries` should always be copied for Oracle and Postgres but not sqlite.

python/lsst/daf/butler/registries/postgresqlRegistry.py
```diff
@@ -0,0 +1,69 @@
...
+    def _createEngine(self):
+        constr = self.config['db']
+
+        # if both username and pass are in the db string - all is well
+        # if username and pass are a nick that exists in a local conf file
+        # read username and password from there
+        parsed = urlparse.urlparse(constr)
+        if parsed.username and (parsed.password is None):
+            localconf = configparser.ConfigParser()
+            localconf.read(os.path.expanduser('~/.rds/credentials'))
```
timj: @ktlim probably has an opinion on the contents of this credentials file but I'm fine with this for now.  We have historically been using `~/.lsst/db-auth.paf` for database access but I don't know if this is something we wish to continue with.  For example should the credentials be in `~/.rds/` and `~/.oracle` or be in a single file inside a `~/.lsst` or `~/.butler` directory?  Also, why are postgres credentials in an `.rds` directory? If this is RDS specific then why is this class named after postgres?

cc/ @MichelleGower

tests/test_butler.py
```diff
@@ -542,5 +544,163 @@ def tearDown(self):
             shutil.rmtree(self.root, ignore_errors=True)
 
 
+class RDSRegistryButlerTestCase(ButlerTests, unittest.TestCase):
+    """PosixDatastore using an RDS PostgreSQL Registry specialization of a butler"""
+    configFile = os.path.join(TESTDIR, "config/basic/butler-rds.yaml")
+    fullConfigKey = ".datastore.formatters"
+    validationCanFail = True
+
+    # there are too many important parameters in the db string that need to
```
timj: This should be placed in setUpClass class method for clarity.

tests/test_butler.py
```diff
@@ -542,5 +544,163 @@ def tearDown(self):
...
+
+        # First, replace the default connection string with a db that we know will always exist.
+        # The name will always be last, but its easier to just replace it than to deconstruct and
+        # then reconstruct the whole string
+        defaultName = self.connectionStr.split('/')[-1]
+        #constr = self.connectionStr.replace(defaultName, 'postgres')
+        constr = self.connectionStr
+
+        # Second, get the local connection credentials from ~/.rds
+        parsed = urlparse.urlparse(constr)
```
timj: This code and the code below it is the same code found in PostgresSqlRegistry.  This strongly suggests that the code for reading credentials out of external files should be in one place and not copied here. The postgres code should know where its own credentials file is and there should be a helper method in the Registry classes. If we used the same approach for oracle that would be even better.

tests/test_butler.py
```diff
@@ -542,5 +544,163 @@ def tearDown(self):
...
+        from sqlalchemy import create_engine
+        import urllib.parse as urlparse
+        import configparser
+
+        # First, replace the default connection string with a db that we know will always exist.
+        defaultName = self.connectionStr.split('/')[-1]
+        constr = self.connectionStr.replace(defaultName, 'postgres')
+
+        # Second, get the local connection credentials from ~/.rds
+        parsed = urlparse.urlparse(constr)
```
timj: Again lots of duplicated code.

timj (changes requested):
Thank you for rearranging things such that it's clear what's needed for postgres support.  My main comment is that I don't think we are ready to accept the test code. daf_butler tests do not yet require access to a remote database in order to run and I see no code here that skips the tests if RDS is not available. We do not test with Oracle in daf_butler either.  I think this means that we need separate integration tests that can be set up to run with different database backends but that should not be included here.

I'm happy to merge the first non-test code once we understand how credentials in external text files should be handled.

python/lsst/daf/butler/registries/postgresqlRegistry.py
DinoBektesevic: Except for the two review comments all others relate to `test_butler` code. As I mentioned in the PR - that code exists only as a quick way for anyone else that perhaps wants to take the `PostgreSqlRegistry` for a spin and because I thought it was a shame to completely "forget" about the steps I needed to take in test set-up. Considering the comment:

> ... happy to merge the first non-test code once we understand how credentials in external text files should be handled ...

I wonder wouldn't it be better to just totally remove the piece of test-code in a new commit where I sort out the authorization issues? This would leave the code in history (at least on my end) and simultaneously get rid of most of the complaints in this Draft-PR.

timj:
Removing the test code in a separate commit is fine with me, although it might be worth deferring that removal until we have sorted out the code duplication so that you can still test what you have whilst you are developing.

python/lsst/daf/butler/registries/postgresqlRegistry.py
brianv0: You probably want an environment variable that you can set which points to the config file more than you want to put it in a well defined place. This sucks from a dev/prod database perspective if wanting to run code against both instances from the same home filesystem because you're going to be doing a bunch of symlink gymnastics. You can still have a default, which should probably be under `~/.lsst`, as it's not a Postgres config file (e.g. `.pgpass` file)

A postgres-specific solution would be to use postgres norms via env vars or a .pgpass file:
https://www.postgresql.org/docs/9.3/libpq-envars.html

DinoBektesevic: I read up on all of the suggestions made here and on the Slack channel yesterday. To summarize, the following suggestions were made:

* Postgres env vars 
  * not recommended by postgres manual as they say on some systems anyone can see any env variable.
* `.paf`, `.conf` file in `~/.rds`, `~/.postgres`, `~/.lsst` or `~/.butler` directory
  * the way its done now
  * I have a feeling (from @ktlim and @timj) that its `.paf` file using something built in the stack like the [`DBAuth` code](https://github.com/lsst/daf_persistence/blob/1e7c341a1a884cf06fc74a946a126c1f72f56360/src/DbAuth.cc) would be most preferred
  * but I would still like some clarification on where exactly should I put this: `~/.lsst` or `~/.butler`? `~/.rds`or `~/.postgres` seem to be not as popular
* `.pgpass` file
  * I can set the `PGPASSFILE` env var to point to `~/.lsst` or `~/.butler` directories so it works well with the file (above) approach
  * the format of entries in the file is given as `hostname:port:database:username:password` so it makes no sense to keep the current connection string format intact. I could change the current connection string to `postgresql://user:pass@hostname/dbname` and read the port and the password for the matching username, hostname and dbname from `.pgpass` file. This is also good as it simplifies the connection string.
  * but at some point if I want to connect to the DB using SQLAlchemy I still have to go there, read the file as plaintext and make substitutions in the connection string. So on face value it doesn't seem any different than the other file based approaches. 
  * The difference is only visible if from terminal one invokes a clean `:~$ psql` command as that is when postgres will natively go there and read the file if no user and pass env vars are set. If there's only one entry in the file it will connect to that directly otherwise `--username` `--host` and `--dbname` have to be given.
  * I'm not counting the env var as a win because that's easy to implement and depends on exactly the same questions as `.pgpass` does: when and where is the variable exported and to which location should it point to?
* vault + confd 
  * This requires a running `vault` server, which is simple enough I guess. I'm just confused about the general approach I guess: should there be one central vault, is everybody supposed to run their own, where is the vault in the AWS then, how do tests work... A lot of things I know not a lot about. 
  * Then the templatefiles need to be created by `confd`, which I guess could be a part of a startup script somewhere (where and when?)
  * but then again, every time I want to connect using SQLAlchemy I would have to run something like:
    ```
    ROOT_TOKEN=$(vault read -field id auth/token/lookup-self)

    confd -onetime -backend vault -node http://127.0.0.1:8200 \
          -auth-type token -auth-token $ROOT_TOKEN
    ```
    and would again have to read the files and make string substitutions to SQLAlchemy as was the case in all the options so far. Nobody can read the values in the files, even if they could access them, unless they luck out and look at them post token-substitutions? I guess it does get me more flexibility, especially if I run in deamon mode that will update the files if the back-end password changes. 

The dot files are plain text, but they are not public in the sense that they need to be in `600` or `400` permission set. All of the answers seem to boil down to reading the dotfile and making plain string substitutions. It was my understanding that the main problem was reading/keeping usernames and passwords floating about in plain text around the code. So I guess I'm confused where the security comes in to play in any of the suggestions, short of relying on OS privileges? (or maybe this is "the security", I don't know a lot about this stuff so someone needs to point it out to me)

The passwords are read prior to creating the connection string and passed on to SQLAlchemy as soon as possible (which then hides them again). This was the reasoning behind duplicating all the read-password code. If I write a function that will read and return username and password isn't that more dangerous than just doing it in place like it is? This is what, I though, I needed to avoid. 

I think I prefer the `.pgpass` file with `PGPASSFILE`, by default, pointing to one of the  `~/.lsst` or `~/.butler` directories by which is read by the LSST's `DBAuth` code approach the best. It gets me everything regular dotfile approach does but also integrates well with native `psql` commands. While we're at it, we can change the location of the AWS credentials file to be in the same default location as the rest of them by setting `AWS_SHARED_CREDENTIALS_FILE`. Any thoughts, comments and answers very welcomed! (especially, I think, input from @brianv0 might be very illuminating to me)

Thank you all for your help!

timj: I'm fine with a postgres file in a ~/.lsst directory. It's really important that you don't duplicate all the code for handling the password reading in multiple places. The more places you have it the more places where a mistake can be made. There is nothing wrong I can see with using function calls to obtain information and then inserting the answers into your connection string.  You should check that the file has the correct restricted permissions.

python/lsst/daf/butler/registries/postgresqlRegistry.py
DinoBektesevic: I have continued to inherit from Oracle only 1 method is re-implemented (the `create_engine` bit) Do you still want me to inherit `SqlRegistry` directly?

python/lsst/daf/butler/registries/postgresqlRegistry.py
DinoBektesevic: Most of the work is focused on this comment. I have implemented a `ConnectionStringBuilder` class in `lsst.daf.butler.core.registry`. It constructs the connection string as generally as possible from the given config and `db-auth.paf` file. All tests pass (including the `RDSButlerTestCase` which was updated in previous commit but removed from the latest one). 

I think I might have over-extended the level of changes you requested but I wanted to avoid repeat of the experience with URI handling from before. I have followed the [SQLAlchemy manual](https://docs.sqlalchemy.org/en/13/core/engines.html) that cites [RFC-1738](https://www.ietf.org/rfc/rfc1738.txt) and all the default values are set equal to those [SQLAlchemy Dialect part](https://docs.sqlalchemy.org/en/13/dialects/index.html?highlight=dialects) of manual states. This should hopefully enable the construction of maximally general connection strings so all cases should be covered.

I was not 100% sure how Oracle connection string has been supplied but I suspect that they use conection strings without user and pass and then leave Oracle to work out the security. So I left that as an keyword option in `ConnectionStringBuilder.fromConfig`. 

I have left questions in form of comments in the `ConnectionStringBuilder` class that I can remove once I hear if you like this or not.

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
     defaultConfigFile = "registry.yaml"
 
 
+class ConnectionStringBuilder:
```
andy-slac: I am not sure that this module is the right place for a class like this. `Registry` is a pure abstract thing and it is not supposed to care about things like database or connection strings. The connection string that is built here is SQLAlchemy-specific, so better place for it is probably `registries.sqlRegistry` (or even better a separate module in `registries` package).

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
...
+    """Builds a connection string by parsing (in order of precedence):
+    I)   RegistryConfig for explicitly defined keys,
```
andy-slac: I understand this is a new addition to the configuration? Why do we need explicit keys, I thought that URL-style connection string is already flexible enough for anything that might be needed. I think having more than one way to specify things adds unnecessary complexity, would be better to keep things simple.

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
...
+    II)  'db' key in RegistryConfig and
+    III) and ~/.lsst/db-auth.paf file.
```
andy-slac: It may be better to re-format this list using markdown syntax so that it's rendered naturally in generated documentation.
There is an `and` on previous line and this line too.

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
...
+
+    When not stated explicitly, or included in the 'db' string, dialect and
+    driver are inferred from targeted registry class defaults. The returned
+    connection string has the following format:
+
+        `dialect+driver://(username:password@?)(host?)/(dbname?)(:port?))`
```
andy-slac: Port number should follow host, not database name. Parentheses are not balanced.

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
...
+
+    Where host, port, username and password are not returned for sqlite
+    dialects, making dbname effectively the path to the DB. Username and
+    password are optional in other cases, such as when credentials hadling
```
andy-slac: hadling -> handling?

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
...
+    are be left to the DB itself.
```
andy-slac: to be left?

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
...
+        #
+        #
+        regConf = RegistryConfig(regConf)
+        explicitKwargs, conStrKwargs = dict(), dict()
+        for key in cls.requiredKeys:
+            explicitKwargs[key] = regConf.get(key)
+            conStrKwargs[key] = None
+
+        # ensure host is only the hostname and not mistaken for the 'db' key as
+        # there is no additional parsing for host unlike 'db'.
```
andy-slac: This looks like a very specific failure, I wonder if it is common enough to justify this test. There may be more generic way to test for correctness of the parameters, though I'd probably leave all checking to database driver itself.

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
...
+        if explicitKwargs['host'] is not None:
+            parsed = urlparse.urlparse(explicitKwargs['host'])
+            if parsed.scheme or parsed.netloc or ':' in parsed.path:
```
andy-slac: I think this may break if `host` is specified as IPv6 colon-separated address (e.g. `fe80::1ff:fe23:4567:890a`)

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
...
+                raise ValueError(("Invalid 'host' key, 'host' should only be "
+                                  "the hostname. For URLs/URIs use 'db' key. "
+                                  f"Received host: {explicitKwargs['host']}. "))
+
+        #
+        #    Parse the keywords from the connection string.
+        #
+        parsed = urlparse.urlparse(regConf['db'])
```
andy-slac: I'd suggest to use `sqlalchemy.engine.url.make_url` method instead if you expect `db` to be in SQLAchemy format, that code already knows everyhting about SQLAlchemy specific features (dialect+driver) so that you can avoid parsing it yourself.

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
...
+
+            # when dialects match use it, when they don't 'db' is an URI, URL
+            # (handled later) or inherited from default (handled now).
+            # Everything else is resolved according to urllib.
+            if registryCls.dialect == parsed.scheme:
+                conStrKwargs['dialect'] = parsed.scheme
+            elif registryCls.dialect != parsed.scheme:
+                conStrKwargs['dialect'] = registryCls.dialect
+                conStrKwargs['driver'] = registryCls.driver
+            else:
```
andy-slac: Is this `else` ever used, I think above if/elif tests are complementary?

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
...
+                conStrKwargs['dialect'] = parsed.scheme
+                conStrKwargs['dbname'] = parsed.path
+
+            conStrKwargs['dbname'] = parsed.path
+            conStrKwargs['port'] = parsed.port
+
+        # scheme-less URLs (f.e. host.com(:port?)(/dbname?)) or its sqlite so
+        # it's safe to ignore it (all cases handled by first or last section)
```
andy-slac: Do we need to support this scheme-less case?

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
...
+            if '+' in parsed.scheme:
+                conStrKwargs['dialect'], conStrKwargs['driver'] = parsed.scheme.split('+')
+            else:
+                conStrKwargs['dialect'] = parsed.scheme
+                conStrKwargs['driver'] = registryCls.driver
+            conStrKwargs['host'] = parsed.hostname
+            conStrKwargs['port'] = parsed.port
+            conStrKwargs['dbname'] = parsed.path
+
+        # are explicit plain text passwords in config files allowed?
```
andy-slac: I think config object does not have to be read from file, it can potentially be constructed or updated from other sources.

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
...
+        mergedKwargs = dict()
+        for key, val in explicitKwargs.items():
+            mergedKwargs[key] = val
+            if val is None:
+                mergedKwargs[key] = conStrKwargs[key]
+
+        return mergedKwargs
+
+    @classmethod
+    def fromConfig(cls, config, withPass=True):
```
andy-slac: Docstring would be nice here too.

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
...
+        # special case of sqlite where user and pass are not really supported
+        # am I safe in assuming I can skip user/pass for sqlite?
+        # https://www.sqlite.org/src/doc/trunk/ext/userauth/user-auth.txt
+        if conKwargs['dialect'] == 'sqlite':
+            constr = '{dialect}+{driver}://{dbname}'
+            withPass = False
+            validationKeys.remove('host')
+            validationKeys.remove('port')
+
+        if withPass:
```
andy-slac: If username/password are already specified in config you probably want to skip searching for it in DbAuth? As it was mentioned already DbAuth store path is hard-coded so it may be problematic for some use cases, it would be better to make it configurable but I do not know if it's possible with DbAuth.

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
...
+        conKwargs = cls._dictFromRegistryConf(config)
+        validationKeys = list(cls.requiredKeys)
+        constr = "{dialect}+{driver}://{username}:{password}@{host}:{port}{dbname}"
+
+        if not withPass:
+            constr = "{dialect}+{driver}://{host}:{port}{dbname}"
+        # special case of sqlite where user and pass are not really supported
+        # am I safe in assuming I can skip user/pass for sqlite?
+        # https://www.sqlite.org/src/doc/trunk/ext/userauth/user-auth.txt
+        if conKwargs['dialect'] == 'sqlite':
```
andy-slac: I think this special case is self-inflicted problem. If we could just stick with URLs only, lot's of the complexity above and this special case would not need to exist.

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
...
+            if val is None:
+                mergedKwargs[key] = conStrKwargs[key]
+
+        return mergedKwargs
+
+    @classmethod
+    def fromConfig(cls, config, withPass=True):
+        conKwargs = cls._dictFromRegistryConf(config)
+        validationKeys = list(cls.requiredKeys)
+        constr = "{dialect}+{driver}://{username}:{password}@{host}:{port}{dbname}"
```
andy-slac: I think port number should be optional in configuration, and I believe `conKwargs['port']` will be `None` in that case, would it be rendered as `None` on output too?

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
...
+
+        # to preserve absolute VS relative paths in URIs the leading '/' have
+        # not been altered. If 'dbname' has not got a leading '/' at this point
+        # however, the returned URI would not be valid. This can happen when
+        # dbname is given explicitly without the leading '/'. This is the one
+        # only ugly hack in this class.
+        if not conKwargs['dbname'].startswith('/'):
+            conKwargs['dbname'] = '/' + conKwargs['dbname']
+
+        return constr.format(**conKwargs)
```
andy-slac: TO me all of the above looks overly complicated (maybe for the reasons that I do not know). If I would do it I'd probably stick to a simpler approach:
- only allow URL (`db`) in configuration, no specific overrides for host/port/etc.
- parse URL with `sqlalchemy.engine.url.make_url` which returns `sqlalchemy.engine.url.URL` instance
- if username/password are not specified in URL then get host/port from it and use those to get info from `DbAuth`, make new URL which includes those
- return `sqlalchemy.engine.url.URL` from this method instead of string, pass that directly to `create_engine()`

python/lsst/daf/butler/registries/postgresqlRegistry.py
andy-slac: Inheriting from `OracleRegistry` does not look right here. In true OOP terms sub-classing means extension of functionality but subclass should still be usable wherever base class is accepted (famous Liskov Substitution Principle). I do not think you would want to use `PostgreSqlRegistry` class to access registry in Oracle. I'd indeed inherited from `SqlRegistry`, if there is some significant common functionality that can be shared between Oracle and Postgres then it can go into a separate base (or mix-in) class.

python/lsst/daf/butler/registries/oracleRegistry.py
```diff
@@ -38,6 +38,8 @@ class OracleRegistry(SqlRegistry):
     config : `SqlRegistryConfig` or `str`
         Load configuration
     """
+    dialect = 'oracle'
+    driver = 'cx_oracle'
```
andy-slac: You don't want to reimplement `_createEngine()` in this class to use `ConnectionStringBuilder`?

tests/test_connectionString.py
```diff
@@ -0,0 +1,82 @@
...
+        """Tests that relative and absolute paths are preserved."""
+        regConf = RegistryConfig(os.path.join(self.configDir, 'conf2.yaml'))
+
+        regConf['host'] = 'some.host.com:1234'
+        with self.assertRaises(ValueError):
+            ConnectionStringBuilder.fromConfig(regConf)
+
+        regConf['host'] = 'some.host.com:1234/dbname'
+        with self.assertRaises(ValueError):
+            ConnectionStringBuilder.fromConfig(regConf)
```
andy-slac: Can you add test case when port number is not specified in configuration (for non-sqlite case)?

andy-slac (changes requested):
By @timj invite I looked at the code and left a bunch of comments. I tried to get some context for the proposed changes but could not gather much (is there a ticket for this in JIRA?) One thing that I do not quite like is an addition of overrides for URL pieces in registry config, I think this adds too much complexity to the code with unknown (to me) benefits. Could we still stick with URL-only connection configuration? As @brianv0 mentioned in one of the comments `DbAuth` use is problematic with its hard-coded location of config file, it would be better to have more flexibility there (I think ticket for `daf_persistence` is the right thing to do). And I think that inheritance for new registry class needs to be changed.

python/lsst/daf/butler/core/registry.py
```diff
@@ -76,6 +78,185 @@ class RegistryConfig(ConfigSubset):
...
+        #    Parse the keywords from the connection string.
+        #
+        parsed = urlparse.urlparse(regConf['db'])
+        registryCls = doImport(regConf['cls'])
+
+        # URLs with scheme (f.e. dialect(+driver?)://(netloc?)/(path?)(:port?))
+        if parsed.scheme and not parsed.netloc:
+            # driver is infered from target registry class when not explicit
+            if '+' in parsed.scheme:
+                conStrKwargs['dialect'], conStrKwargs['driver'] = parsed.scheme.split('+')
```
andy-slac: This will crash if there is more than one `+` in scheme. I'd suggest `partition()` instead of `split()`.

python/lsst/daf/butler/core/registry.py
DinoBektesevic: I tried, but then because I cast whatever `config` is into a `RegistryConfig` I tend to end up with circular import errors. This happens in the new implementation too. I'll work on it tomorrow, and it will get moved out to lets say `connectionString.py`.

python/lsst/daf/butler/core/registry.py
DinoBektesevic: I didn't know there's a strict requirement that the `db` string be an URL and it was not obvious that this is the case from conversations or other PRs. I wanted more flexibility. The current changes are more in line with what you said.

python/lsst/daf/butler/core/registry.py
DinoBektesevic: Self inflicted damage. Not important anymore.

python/lsst/daf/butler/core/registry.py
DinoBektesevic: @ktlim seemed to be very intent on DBAuth usage and recommended that I use it, so I did. Other options are discussed [above](https://github.com/lsst/daf_butler/pull/161#discussion_r288731646) and I keep hearing advice and conflicting opinions on authorization and credentials issues without a clear cut answer to what is it that I should do. 

The conclusion was that insofar this approach is good enough. Certainly if you have a different concrete suggestions as what I should do/use to read credentials and where they should be stored I'd be happy to implement that.

python/lsst/daf/butler/core/registry.py
DinoBektesevic: Partially, `DBAuth` matches on `host` and `port` so in some cases `port` becomes mandatory. I don't believe in the previous case it would have been rendered as "None" but as "". This was used in cases where `dbname` was the path to a file as then it would, usually, get shifted leftwards so file-targeting DB connection strings would work too. 

In any case, not important in the reworked version. Added a test that expects the correct result when no port, no username and no password are provided for good measure.

python/lsst/daf/butler/core/registry.py
DinoBektesevic: There was a lot of talk about how authentication should work, especially in aws poc group and on slack, but not a lot of concrete plans were laid down. I was just trying to be as flexible as possible based on just about everything I could scrape together ([1](https://github.com/lsst/daf_butler/pull/161#discussion_r288274111), [2](https://github.com/lsst/daf_butler/pull/147#discussion_r279095140), [3](https://lsstc.slack.com/archives/C2JPT1KB7/p1559159192163100?thread_ts=1559159029.160100&cid=C2JPT1KB7), [4](https://lsstc.slack.com/archives/C2JPT1KB7/p1559158810158900), [5](https://lsstc.slack.com/archives/C2JPT1KB7/p1557249828219600), [6](https://github.com/lsst/daf_butler/pull/145/files/1aa25b75d400d976b7f97a360db06e16ea3264ce), [7](https://lsstc.slack.com/archives/C2JPT1KB7/p1559158810158900) and many conversations with @ktlim, @timj and in the AWS POC meetings). Additionally I did not know about `sqlalchemy.engine.url.URL` and freshly came off of `ButlerURI` bandwagon so I might have overdid it doing it from `urllib`. Completely my bad. 

The new implementation is much shorter and based on `sqlalchemy.engine.url.make_url` and follows more of your suggestions. I still left certain options in to add explicit keys to `RegistryConfig` (or config in general) just because the nature of `make_url` makes it pop up naturally.

python/lsst/daf/butler/registries/oracleRegistry.py
DinoBektesevic: After all the conversations it still was not quite clear to me how does oracle do it (f.e. [3](https://lsstc.slack.com/archives/C2JPT1KB7/p1557249828219600)) so finally [4](https://github.com/lsst/daf_butler/pull/145/files/1aa25b75d400d976b7f97a360db06e16ea3264ce) made me not want to touch `OracleRegistry` until I actually am sure I won't annoy someobody.

I added the `ConnectionStringBuilder` here too now.

tests/test_connectionString.py
DinoBektesevic: I can do that but I'm not sure what the usecase is. Will username and password be required in this non-sqlite case? How would I use the `DBAuth` then? Does this still apply?

timj:
@ktlim has agreed to look at this pull request.

python/lsst/daf/butler/core/registry.py
DinoBektesevic: Fixed. Moved into its own file now.

DinoBektesevic:
At PCW 2019 DM Hack session it was agreed upon, between @ktlim, @andy-slac and me, that correct registry class should be resolved based on the dialect part of the connection string. 

TO DO: resolve correct Registry class to use directly from connection string.

timj:
My main concern with deriving the class from the connection string is that we still allow for the mapping from connection string to class to be a configuration option.

cc/ @TallJimbo

TallJimbo:
> My main concern with deriving the class from the connection string is that we still allow for the mapping from connection string to class to be a configuration option.

Was the second clause of that sentence supposed to be slightly different?  What else would the connection string be other than a configuration option?  (Sorry if that's a silly question; I've thus far avoided getting involved in the connection/authentication configuration discussions.)

DinoBektesevic:
> What else would the connection string be other than a configuration option? (Sorry if that's a silly question; I've thus far avoided getting involved in the connection/authentication configuration discussions.)

Right now the config looks like:

```
registry:
  cls: lsst.daf.butler.registries.sqliteRegistry.SqliteRegistry
  db: 'sqlite:///:memory:'
```

the idea is that we kick the `cls` out and infer the correct `cls` from the `db` string, which is possible, but not quite how configs are currently used in the stack. (If I understood what was said to me correctly). Namely in `Butler.makeRepo`:

```
datastoreClass = doImport(full["datastore", "cls"])
datastoreClass.setConfigRoot(BUTLER_ROOT_TAG, config, full, overwrite=forceConfigRoot)
registryClass = doImport(full["registry", "cls"])
registryClass.setConfigRoot(BUTLER_ROOT_TAG, config, full, overwrite=forceConfigRoot)
```

timj:
@TallJimbo my concern is that I'm not sure I  want python code to be the place where something says "if this looks like a sqlite connection string then import the sqlite registry". Maybe everyone else does want that? It means that no-one could specify their own override sqlite registry implementation but maybe we don't want that?  Deriving the class from the connection string does have the advantage that you aren't ever going to specify the oracle registry with a postgres connection string. 

I was suggesting we have an extra yaml configuration section something like:

```
registryMap:
  sqlite: lsst.daf.butler.registry.sqliteRegistry
  postgresql: lsst.daf.butler...
  oracle: lsst.daf.butler...
```
Then in theory someone could experiment with different classes. It does mean that the connection string parser can only support a limited set of keys in registryMap.

TallJimbo:
That works for me.  Another approach that I think would provide the same flexibility would be to retain `cls`, but make it optional and look in the connection string only if it is not provided.

timj:
@TallJimbo allowing `cls` for overrides sounds good to me, but still needs the `registryMap` concept to exist in the YAML if you want to avoid hard-coding the mapping in the python.

TallJimbo:
I was thinking it'd be fine to hard-code the mapping if you had a way to override it, but I guess that's not quite as flexible (presumably the need to change this means there are some databases that can't use `cls` for some reason?).

DinoBektesevic:
>  (presumably the need to change this means there are some databases that can't use `cls` for some reason?).

Any class importable with `doImport` would work. I thought that it was to avoid information duplication such as pointing to a wrong class while using a wrong connection string. Just in case I got it wrong, could @ktlim or @andy-slac elaborate on the reasons further? 

EDIT: is this really the best place for this discussion or should it move to slack?

andy-slac:
The concern is indeed to avoid duplication and keep things consistent. URL string should be enough to identify everything that code needs, I don't think we'll have different implementations for the same backend (ideally we would need only one SQL implementation with SQLAlchemy handling all backend peculiarities). A agree that for developers it may be needed to override the class name for particular backend. I think Tim's map is probably cleanest way to do that (and `cls` override would not be needed if we have a map), what worries me is that configuration becomes more and more involved and if defaults are specifies in the config file and not in the code then it effectively means that we depend on configuration in the same way as the code. I'd probably like more if we had defaults in the code (maybe as a default configuration rather than hard-coded `if ... else ...`). So if defaults are in the code then having `cls` override may be easier to manage than a map. Sorry if this is not super-decisive, I don't know an answer that can satisfy all possible use cases.

TallJimbo:
Sorry to derail this a bit by asking everyone to catch me up on the motivation.

I certainly don't have terribly strong opinions about this, but I think the simplest thing that does everything we might want is @timj's addendum (adding a mapping from connection substring to registry class to the config).  I think we can revisit whether we need even that later; I may have a bit more clarity in the next month or so on how much we will actually be able to collapse the Registry subclasses down by relying more on SQLAlchemy and avoiding database-specific syntax in our use of it.

DinoBektesevic:
So I added automatic resolution of correct Registry class based on the dialect found in the connection string. This is still an early commit, waiting for the dbAuth to get finalized before connection string works as agreed upon, but it could be useful to open this matter for early discussion. 

Specifically, is there a better way to read the `registryMap.yaml` config file? Should the `registryMap` even be a key in `registryMap.yaml` config file?

timj:
Regarding registryMap. How about we put it in config/registry.yaml. Replacing:
```yaml
registry:
  cls: lsst.daf.butler.registries.sqliteRegistry.SqliteRegistry
  ...
```
with
```yaml
registry:
   clsMap: 
      sqlite: lsst.daf.butler.registries.sqliteRegistry.SqliteRegistry
      oracle: lsst.daf.butler.registries.sqliteRegistry.OracleRegistry
```

DinoBektesevic:
@timj Sorry, got distracted and forgot about this response. I'm not sure how that would work in the global scheme of things since the configs in the `daf_butler/config` are considered `defaults` so when no `registry` is explicitly given in the "custom" `butler.yaml` then the registry is inherited from there.

See the `full` config in `Butler.makeRepo`. I don't know if that's what you intended, or the intention was to just add `clsMap` key to the default `registry.yaml` config?

timj:
Yes, exactly, the clsMap would be inherited into all the configs. That's what we want isn't it? We agreed yesterday that if you put an explicit `registry.cls` in then that would be used instead of getting it from the connection string. Also, people could define their own `clsMap` and it would override the defaults.

DinoBektesevic:
I tried my best but then I got very confused again. Couple of things I'm not clear about (even after sitting on it for two days):

1) Technically I didn't need the `ConnectionString` class and could have reverted to having `ConnectionStringBuilder.fromConfig` be a class method taking a `DbAuth` instance. It wouldn't shorten the code, or resolve the following issue, because I would have to instantiate a new `DbAuth` object everytime anyhow. 
2) I don't know how I should have made the `ConnectionStringFactory` default `envVar`   and `path` kwargs correctly. I followed the `repoRelocation.py` example, but I wanted to make that customizable through the methods. Because `Butler.makeRepo` needs to resolve the class firstly (to make sure the correct `setConfigRoot` is called) to do that I would need to change `Butler.makeRepo`, `Butler.__init__` and `Registry.fromConfig` to take in either a `ConnectionStringFactory` or a `path` and `envVar` at least. To get rid of creating the factory in different registry implementations of `create_engine` I could have the `ConnectionStringFactory` as an attribute of `Registry` which both seems reasonable but I didn't want to change too much. 
3) I had to kick out setting `cls` keys out of the `Registry.setConfigRoot` methods since they don't have to necessarily exist any more. I don't know how mad or unmad that can make people.
4) I really am not sure how to test if any of the touched methods (`makeRepo`, `fromConfig`, butlers `__init__` etc actually work in other cases except the sqlite default one because any created registry will try to connect to it but for postgres and oracle they might not actually exist. 

I also think I made a bit of a jumble with the history having merged ktl's work before it got to master. I do hope none of this is very bad.

python/lsst/daf/butler/butler.py
```diff
@@ -203,8 +204,24 @@ def makeRepo(root, config=None, standalone=False, createRegistry=True, searchPat
...
         datastoreClass = doImport(full["datastore", "cls"])
         datastoreClass.setConfigRoot(BUTLER_ROOT_TAG, config, full, overwrite=forceConfigRoot)
-        registryClass = doImport(full["registry", "cls"])
+
+        # Use the explicitly set "cls" key in given config, determine the
+        # correct registry class from connection url or use the default cls
+        # This needs to be done explicitly here because Registry.setConfigRoot
+        # could differ for different Registries.
+        conStrFactory = ConnectionStringFactory()
+        if config.get(".registry.cls") is not None:
```
timj: We should not be duplicating this logic here and lower down. There should be a single place where code works out the registry class.  How about adding all this as a class method to `Registry`?
```python
registryClass = Registry.determine_class(config)
```
?

python/lsst/daf/butler/butler.py
```diff
@@ -203,8 +204,24 @@ def makeRepo(root, config=None, standalone=False, createRegistry=True, searchPat
...
+            registryClass = doImport(config["registry", "cls"])
+        elif config.get(".registry.cls") is None and config.get(".registry.db") is not None:
```
timj: Here and elsewhere use `config.get("registry", "cls")` rather than using the `.` delimiter.

python/lsst/daf/butler/core/config.py
```diff
@@ -794,7 +794,7 @@ def dumpToUri(self, uri, updateFile=True, defaultFileName="butler.yaml"):
         if isinstance(uri, str):
             uri = ButlerURI(uri)
 
-        if uri.scheme == "file":
+        if uri.scheme == "file" or not uri.scheme:
```
timj: I think if you rebase this will disappear?

python/lsst/daf/butler/core/registry.py
```diff
@@ -182,8 +183,22 @@ def fromConfig(registryConfig, schemaConfig=None, dimensionConfig=None, create=F
             else:
                 raise ValueError("Incompatible Registry configuration: {}".format(registryConfig))
 
-        cls = doImport(registryConfig["cls"])
-        return cls(registryConfig, schemaConfig, dimensionConfig, create=create, butlerRoot=butlerRoot)
+        # If `cls` has been explicitly given in the config, use targeted class,
+        # otherwise infer from the `db` string
+        if registryConfig.get("cls") is not None:
```
timj: This will change to use the new class method.

python/lsst/daf/butler/registries/oracleRegistry.py
```diff
@@ -79,4 +79,6 @@ def __init__(self, registryConfig, schemaConfig, dimensionConfig, create=False,
                          butlerRoot=butlerRoot)
 
     def _createEngine(self):
-        return create_engine(self.config["db"], pool_size=1)
+        conStrFactory = ConnectionStringFactory()
+        conStr = conStrFactory.fromConfig(self.config)
```
timj: Can we have this use the Factory internally so this becomes one line as a Class method rather than 2? These same 2 or 3 lines are used in multiple places.

python/lsst/daf/butler/registries/oracleRegistry.py
```diff
@@ -69,7 +69,7 @@ def setConfigRoot(cls, root, config, full, overwrite=True):
         """
         super().setConfigRoot(root, config, full, overwrite=overwrite)
         Config.updateParameters(RegistryConfig, config, full,
-                                toCopy=("cls",), overwrite=overwrite)
```
timj, line 72: Should this now be copying across `db`?

timj (commented):
I've made some suggestions on the non-database code.

python/lsst/daf/butler/core/connectionString.py
```diff
@@ -0,0 +1,168 @@
...
+        """Returns connection string arguments as a dictionary."""
+        return self._conStr.translate_connect_args()
+
+
+class ConnectionStringFactory:
+    """Factory for `ConnectionString` instances.
+
+    Parameters
+    ----------
+    path : `str` or None, optional
```
ktlim: I don't think you need this generality.  These parameters were meant to allow the application to configure DbAuth.  The application itself doesn't need to be configured.  You can remove all these parameters and just use the module defaults.

python/lsst/daf/butler/core/connectionString.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def __init__(self, path=None, envVar=None, authList=None):
+        if path is None:
+            path = DB_AUTH_PATH
+        if envVar is None:
+            envVar = DB_AUTH_ENVVAR
+        self._dbAuth = DbAuth(path, envVar, authList)
+
+    def __str__(self):
+        return f"{self.__class__.__name__}@{self._dbAuth.authList}"
```
ktlim: This is dangerous; we don't want to accidentally expose the user's auth information in a log message or print output.  I'd remove it.

python/lsst/daf/butler/core/connectionString.py
```diff
@@ -0,0 +1,168 @@
...
+    def fromConfig(self, config):
+        """Parses a Config 'db' string and other keywords and returns an
+        `sqlalchemy.engine.url.URL` object containing the components of a URL
+        used to connect to a database.
+
+        Parameters
+        ----------
+        config : `Config` or `RegistryConfig`
+            Config to parse for connection string components.
+        dbAuth : `DbAuth` or None
```
ktlim: There's no `dbAuth` parameter here (and I don't think there needs to be one).

python/lsst/daf/butler/core/connectionString.py
```diff
@@ -0,0 +1,168 @@
...
+            function.
+        """
+        regConf = RegistryConfig(config)
+        conStr = url.make_url(regConf['db'])
+
+        for key in self.keys:
+            if getattr(conStr, key) is None:
+                setattr(conStr, key, regConf.get(key))
+
+        # when the databsase is a file, host will be None and the path
```
ktlim: When the database is a file, there are no passwords, and you don't need to call DbAuth at all.

python/lsst/daf/butler/core/connectionString.py
```diff
@@ -0,0 +1,168 @@
...
+class ConnectionString(url.URL):
+    """Builds a connection string of the format:
+
+        dialect+driver://user:password@host:port/database
+
+    by parsing (in order of precedence):
+
+    1. 'db' key in RegistryConfig,
+    2. RegistryConfig for explicitly defined keys
+    3. and DbAuth instance with which it was created.
```
ktlim: I don't think this is doing what it says.  In fact, I don't think it is or should be doing much more than `sqlalchemy.engine.URL`.  If you really want to add `checkDialectDriverExist()`, that's OK.

python/lsst/daf/butler/core/connectionString.py
```diff
@@ -0,0 +1,96 @@
...
+    connection string, the 'db' key in the registry configuration.
+    Username, password, host, port or database can be specified as keys in the
+    config explicitly. If username or password are missing a matching DB is
+    found in the credentials file pointed to by `DB_AUTH_ENVVAR` or
+    `DB_AUTH_PATH` values.
+    """
+
+    keys = ('username', 'password', 'host', 'port', 'database')
+
+    def fromConfig(self, registryConfig):
```
timj: I think this should be a class method. `self` is not used other than to access the `keys`.

python/lsst/daf/butler/butler.py
DinoBektesevic: Registry now has a `getRegistryDialect` and a `getRegistryClass` that will do the parsing and `doImport` to return the correct subclass. Is this more along the lines of what you imagined?

python/lsst/daf/butler/core/registry.py
```diff
@@ -129,6 +130,26 @@ def setConfigRoot(cls, root, config, full, overwrite=True):
...
+
+        Returns
+        -------
+        dialect : `str`
+            Dialect found in the connection string.
+        """
+        # this import can not live at the top due to circular import issue
+        from .connectionString import ConnectionStringFactory
+        conStrFactory = ConnectionStringFactory()
+        conStr = conStrFactory.fromConfig(registryConfig)
```
timj: No need for a class so replace with
```python
constr = ConnectionStringFactory.fromConfig(registryConfig)
```

python/lsst/daf/butler/core/registry.py
```diff
@@ -202,6 +254,15 @@ def __init__(self, registryConfig, schemaConfig=None, dimensionConfig=None, crea
     def __str__(self):
         return "None"
 
+    @property
+    def connectionString(self):
+        """Return the connection string to the underlying database
+        (`sqlalchemy.engine.url.URL`).
+        """
+        from .connectionString import ConnectionStringFactory
```
timj: Add comment about import

python/lsst/daf/butler/butler.py
DinoBektesevic: I ran into a bit of an issue with `onfig.get("registry", "cls")` returning `"cls"` and had to do `onfig.get(("registry", "cls"))` if the config is not a `RegistryConfig`. Is that acceptable?

python/lsst/daf/butler/core/registry.py
```diff
@@ -202,6 +254,15 @@ def __init__(self, registryConfig, schemaConfig=None, dimensionConfig=None, crea
...
+        conStrFac = ConnectionStringFactory()
+        return conStrFac.fromConfig(self.config)
```
timj:
```python
return ConnectionStringFactory.fromConfig(self.config)
```

python/lsst/daf/butler/registries/oracleRegistry.py
DinoBektesevic: `Registry.connectionString` property now does this. I wasn't quite sure where to put it, or what to make it and property in Registry made it the "cleanest" (i.e. minimizes the number of lines in registries).  Is this acceptable?

python/lsst/daf/butler/core/connectionString.py
DinoBektesevic: Application now not configurable. Tests changed to reflect this - the module `DB_AUTH_PATH` and `DB_AUTH_ENVVAR` vars are now being retargeted.

python/lsst/daf/butler/core/connectionString.py
DinoBektesevic: When username or password are given in the `db` connection string or as explicit keys in the config the `DbAuth` is not instantiated. Is that ok?

python/lsst/daf/butler/core/connectionString.py
DinoBektesevic: Removed. `ConnectionStringFactory` is now a `sqlalchemy.engine.url.URL` factory effectively. Can the naming stay or would `URLBuilder` or `URLFactory`, or an even better suggestion, be more appropriate?

timj (approved):
Much clearer. I think something is odd with your rebase though because at least one of those K-T commits in your history is a whitespace fix that seems unrelated to the commit message.

python/lsst/daf/butler/core/connectionString.py
DinoBektesevic: Yes, I should have remembered to review that myself before committing. 

You are fine with the `getDialect`, `getRegistryClass` and `Registry.connectionString` being static, class and property? And the `any` test in `Butler.makeRepo`?

python/lsst/daf/butler/butler.py
timj: Yes, sorry. Tuple is needed for `get()`.

python/lsst/daf/butler/core/registry.py
```diff
@@ -182,8 +203,39 @@ def fromConfig(registryConfig, schemaConfig=None, dimensionConfig=None, create=F
...
+        from a list of aliases found under `clsMap` key of the registry config.
+
+        Parameters
+        ----------
+        registryConfig : `ButlerConfig`, `RegistryConfig`, `Config` or `str`
+            Registry configuration
+
+        Returns
+        -------
+        registry : `Registry` (subclass)
```
timj: I think it returns a `type` of `Registry` not an instance of `Registry`.

DinoBektesevic: A tongue twister but better?

DinoBektesevic:
I changed the rebase so the odd changes don't appear anymore. I'm not sure if this was the right way of doing it but nothing else seems broken?

python/lsst/daf/butler/core/registry.py
timj: Maybe say "class of type `Registry`" ?

DinoBektesevic: Sorted, 

> class of type `Registry`

it is.

python/lsst/daf/butler/core/connectionString.py
```diff
@@ -0,0 +1,97 @@
...
+
+        for key in cls.keys:
+            if getattr(conStr, key) is None:
+                setattr(conStr, key, regConf.get(key))
+
+        # when the databsase is a file, host will be None and the path
+        # becomes the database, but we are not treating 'localhost' explicitly
+        host = conStr.host
+        if conStr.host is None:
+            host = str(conStr.database)
```
ktlim: I still dislike this.  If there is no host, we shouldn't set one, especially not to the database name.  I'd just `return conStr` here.

python/lsst/daf/butler/core/connectionString.py
ktlim: See below.

python/lsst/daf/butler/core/registry.py
```diff
@@ -129,6 +130,25 @@ def setConfigRoot(cls, root, config, full, overwrite=True):
         Config.updateParameters(RegistryConfig, config, full,
                                 toCopy=(("skypix", "cls"), ("skypix", "level")), overwrite=overwrite)
 
+    @staticmethod
+    def getDialect(registryConfig):
```
timj: I'm not completely sold on this being a static method of Registry.  I think it would be better as a normal instance method on `RegistryConfig`.

python/lsst/daf/butler/core/registry.py
```diff
@@ -202,6 +253,15 @@ def __init__(self, registryConfig, schemaConfig=None, dimensionConfig=None, crea
     def __str__(self):
         return "None"
 
+    @property
+    def connectionString(self):
+        """Return the connection string to the underlying database
+        (`sqlalchemy.engine.url.URL`).
+        """
+        # this import can not live at the top due to circular import issue
```
ktlim: I guess moving `RegistryConfig` into its own module would break this circular import?  You might consider that.

python/lsst/daf/butler/core/registry.py
```diff
@@ -182,8 +202,39 @@ def fromConfig(registryConfig, schemaConfig=None, dimensionConfig=None, create=F
...
 
-        cls = doImport(registryConfig["cls"])
-        return cls(registryConfig, schemaConfig, dimensionConfig, create=create, butlerRoot=butlerRoot)
+        cls = Registry.getRegistryClass(registryConfig)
+
+        return cls(registryConfig, schemaConfig, dimensionConfig, create=create,
+                   butlerRoot=butlerRoot)
+
+    @classmethod
+    def getRegistryClass(cls, registryConfig):
```
timj: Reading the code here and seeing that it doesn't really need `cls`, and thinking of my comment on `getDialect` I am tending towards this also being an instance method of `RegistryConfig`.

tests/test_connectionString.py
```diff
@@ -0,0 +1,70 @@
...
+from lsst.daf.butler.core.connectionString import ConnectionStringFactory
+
+TESTDIR = os.path.abspath(os.path.dirname(__file__))
+
+
+class ConnectionStringBuilderTestCase(unittest.TestCase):
+    """Tests for ConnectionStringBuilder."""
+    configDir = os.path.join(TESTDIR, "config/basic/connectionStringConfs/")
+    configFiles = os.listdir(configDir)
+    credentialsFile = os.path.join(TESTDIR, "testDbAuth.paf")
```
ktlim: Not really a `.paf` file; rename to `.yaml`.

tests/testDbAuth.paf
```diff
@@ -0,0 +1,17 @@
...
+    url: 'sqlite://lsst-db.ncsa.illinois.edu:3306/*',
+    username: boris,
+    password: natasha,
+  },
+  {
+    url: 'oracle://lsst-db.ncsa.illinois.edu:3306/*',
+    username: nottest,
+    password: notglobular.test,
+  }
+]
```
ktlim: Should have a newline at the end of the last line.

ktlim (approved):
Looks reasonable to me.

python/lsst/daf/butler/core/registry.py
DinoBektesevic: Moved to `RegistryConfig` as an instance method.

timj:
@andy-slac you have a "change requested" review on this PR. I think the changes have been done so would you like to update your review?

python/lsst/daf/butler/core/registry.py
DinoBektesevic: I don't think that's the case. The problem is that I'd like for `ConnectionStringFactory` to "downgrade/upgrade" any type of config to a `RegistryConfig`. That's why I'm recasting any config or string sent to `ConnectionStringFactory.fromConfig` into a `RegistryConfig`. 

Meanwhile I am using the `ConnectionStringFactory` in the `RegistryConfig`s methods. This wouldn't be the case if the methods are left in the `Registry`, but since we moved them to `RegistryConfig` the problem is solvable only if I make `ConnectionStringFactory.fromConfig` into a `ConnectionStringFactory.fromRegistryConfig` but I think that's very impractical for the cost of a single import line in a single classmethod. 

I went forward and separated the `RegistryConfig` into a new module anyhow because there's some "meat" on it now so I thought it looked neater. I can bring it back though if you don't think it's big enough for that or if It would just be better if they stay bundled like `Datastore` and `DatastoreConfig` (question for @timj as well).

python/lsst/daf/butler/core/registry.py
DinoBektesevic: Moved to `RegistryConfig` as an instance method. It does work better.

python/lsst/daf/butler/core/connectionString.py
DinoBektesevic: This is a relic from before when I used to let the `db` string be just the host or a path and the `cls` key to set the dialect. If the `db` was a path then it would become the database name, but if it was an url, the host. Since I was treating the host as the target location previously I had to make that switch. Since we kicked out `cls` and I discovered URL (thanks to Andy) I believe it's not needed now at all so I removed it. 

This affects the username and password setting issue I asked you about. I don't want to return the connection string here because there are single file databases that support username and passwords (LiteDB, RaptorDB, sqlite has an extension that supports usernames and passwords). I don't know/think that they would be ever used but it's been made clear to me that I should try to be as general as possible (but also that it's ok not to do it for sqlite). 

So I left it as is, but I take it, from the early `return`, that you're not fine with not using DbAuth when username+password are given in db connection string or as explicit keys in the config file for single file dbs?

python/lsst/daf/butler/core/registry.py
timj: Separate is fine now that there's real code in there.

andy-slac (approved):

andy-slac:
I looked at it quickly again, looks better indeed. Approved.

DinoBektesevic:
Thank you all for the reviews and all the help! Sorry if I was a bother.

PR #162, opened 2019-05-20
DM-19818: Fix and improve tests for butler pickling.

TallJimbo:


python/lsst/daf/butler/butler.py
```diff
@@ -208,15 +208,15 @@ def __init__(self, config=None, butler=None, collection=None, run=None, searchPa
             if config is not None or searchPaths is not None:
                 raise TypeError("Cannot pass config or searchPaths arguments with butler argument.")
             # save arguments for pickling
-            self._args = (butler._args[0], collection, run)
+            self._args = (butler._args[0], None, collection, run)
```
timj: This seems fragile. Is it at all possible to use kwargs here?

TallJimbo: Yeah, I thought about actually forcing kwargs for all arguments after the first one, but `__reduce__` doesn't support kwargs, and I don't know of something similar that does.  Do you?

natelust: Talked about this in person, and it seems like we should save the arguments by what is passed into the constructor exactly by looking up the values from the defined locals, such that a newly constructed butler will be exactly the same

TallJimbo: Improved as per @natelust's suggestion, though I'd still love to be able to use kwargs.  It looks like we *might* be able to use kwargs if we pickled via `__get_newargs_ex__` instead of `__reduce__`, but it seems like that's not recommended unless you actually define `__new__`, and I'm not sure if it behaves the same way in other, subtler, respects.  I'm inclined to open a ticket to look into that in the future instead of tackling it now, as I have other higher-priority things to work on once the bug itself is squashed.

natelust (approved):

PR #164, opened 2019-05-24
DM-19851: Use window function for multi-collection query

andy-slac:
This replaces a query which used JOIN to find best collection for a
dataset with a query that uses window function for the same purpose. It
should reduce the number of passes over table data and improve
performance.

### DIFF
--- a/python/lsst/daf/butler/sql/singleDatasetQueryBuilder.py
+++ b/python/lsst/daf/butler/sql/singleDatasetQueryBuilder.py
@@ -22,7 +22,8 @@
 __all__ = ("SingleDatasetQueryBuilder",)
 
 import logging
-from sqlalchemy.sql import select, and_, func, functions, case
+from sqlalchemy.sql import select, and_, func, functions, case, literal
+from sqlalchemy import String
 from ..core import DimensionJoin
 from .queryBuilder import QueryBuilder
 
@@ -241,7 +242,9 @@ class SingleDatasetQueryBuilder(QueryBuilder):
         # MIN window function, plus SUBSTR
         nDigits = 3
         collFmt = f"{{:0{nDigits}d}}{{}}"
-        collCase = case({coll: collFmt.format(pos, coll) for pos, coll in enumerate(collections)},
+        # literal() for key is needed to avoid ORA-12704 error
+        collCase = case({literal(coll, String): collFmt.format(pos, coll)
+                         for pos, coll in enumerate(collections)},
                         value=datasetCollectionTable.columns.collection)
         collMin = functions.min(collCase).over(partition_by=_columns(datasetTable, links))
         firstColl = func.substr(collMin, nDigits+1).label("first_collection")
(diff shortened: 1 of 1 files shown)


python/lsst/daf/butler/sql/singleDatasetQueryBuilder.py
```diff
@@ -242,38 +224,31 @@ def _columns(selectable, names):
...
-        # now join these two
-        joinsOn = [groupSubq.columns.collorder == combined.columns.collorder] + \
-                  [groupSubq.columns[colName] == combined.columns[colName] for colName in links]
-
-        return cls(registry, fromClause=combined.join(groupSubq, and_(*joinsOn)),
-                   datasetType=datasetType, selectableForDataset=combined, addResultColumns=addResultColumns)
+        # CASE clause that transforms collection name to prefixed name, plus
+        # MIN window function, plus SUBSTR
+        nDigits = 3
+        collFmt = f"{{:0{nDigits}d}}{{}}"
```
TallJimbo, line 244: This is a really minor issue, of course, but it looks like the docs have a ":" between the order and the collection name, while the actual code does not.  Seems best to keep them consistent, just for clarity.

python/lsst/daf/butler/sql/singleDatasetQueryBuilder.py
```diff
@@ -242,38 +224,31 @@ def _columns(selectable, names):
...
+        collCase = case({coll: collFmt.format(pos, coll) for pos, coll in enumerate(collections)},
+                        value=datasetCollectionTable.columns.collection)
+        collMin = functions.min(collCase).over(partition_by=_columns(datasetTable, links))
+        firstColl = func.substr(collMin, nDigits+1).label("first_collection")
+
+        columns = _columns(datasetTable, ["dataset_id"] + links) + \
+            _columns(datasetCollectionTable, ["collection"]) + [firstColl]
+
+        subq = select(columns).select_from(subJoin).where(subWhere)
+        whereClause = subq.columns.collection == subq.columns.first_collection
```
TallJimbo, line 258: I don't see any CTE here.  Is SQLAlchemy auto-detecting that it should use that syntax, or did you decide not to use it in the end?

TallJimbo (approved):
Code looks fine, though I have a question about what happened to the CTE, and I have not yet tested on Oracle.  Will do that soon and ping you when I have (hopefully later tonight).

python/lsst/daf/butler/sql/singleDatasetQueryBuilder.py
andy-slac: There is no CTE in code indeed, with how it is structured now I don't think we can use CTE as we pass FromCaluse and WhereClause separately to class constructor. And in this particular case there is not much benefit in using CTE vs. subquery, but I still wanted it mentioned in docstring because Chris says it's his favorite as it makes things easier to optimize in Oracle. If it's confusing it may be better not to mention it at all.

TallJimbo: Probably best to just summarize this discussion in the docs - we *could* use CTE, but would have to restructure the code, and have not (yet) done so.

timj:
What's the status of this pull request?

TallJimbo:
I think it's unlikely to be merged anytime soon, but I'd like to keep the branch around in case we want to lift code from it in the future.

PR #166, opened 2019-05-30
DM-18739: Make gen3.py skip defects

SimonKrughoff:


r-owen (approved):
Looks good (I think; I'm not really familiar with the reasoning behind not converting these files)

timj:
I think because in gen2 to gen3 conversion @SimonKrughoff is suggesting that we recreate the defects from scratch rather than migrating them from gen2 repo.

SimonKrughoff:
That's exactly right.

PR #167, opened 2019-05-30
DM-19916: Rewrite URI handling within Location and LocationFactory

timj:
Previously we were a bit loose with our definition of URI.  This rewrite changes the API to Location such that it now takes a URI ParseResult as the root of the datastore and a relative path reflecting the position in the datastore.  This means that we no longer have to deal with strange relative path URIs that are non-standard for files.

This change removes fromUri from the LocationFactory API since it makes no sense to define a root URI for the datastore and then allow another URI to be provided.  No butler code called fromUri.  It also changes the Location constructor but only LocationFactory was calling it.

There are some shenanigans with trying to deal with the difference between a posixpath and an os.path but given that I am testing this on a posix system there might be some inconsistencies still.

Tests have been added. All tests pass.

### DIFF
--- a/python/lsst/daf/butler/core/location.py
+++ b/python/lsst/daf/butler/core/location.py
@@ -19,13 +19,14 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-__all__ = ("Location", "LocationFactory")
+__all__ = ("Location", "LocationFactory", "ButlerURI")
 
 import os
 import os.path
 import urllib
 import posixpath
 from pathlib import Path
+import copy
 
 if os.sep == posixpath.sep:
     IS_POSIX = True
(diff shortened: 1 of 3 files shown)


timj:
cc/ @DinoBektesevic 

This supersedes #160

python/lsst/daf/butler/core/location.py
```diff
@@ -19,45 +19,329 @@
...
 import os.path
 import urllib
+import posixpath
+from pathlib import Path
+import copy
+
+if os.sep == posixpath.sep:
+    IS_POSIX = True
+else:
+    IS_POSIX = False
```
andy-slac: Maybe shorten to `IS_POSIX = os.sep == posixpath.sep`? Or `IS_POSIX = os.path is posixpath`?

python/lsst/daf/butler/core/location.py
```diff
@@ -19,45 +19,329 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-__all__ = ("Location", "LocationFactory")
+__all__ = ("Location", "LocationFactory", "ButlerURI")
 
 import os
 import os.path
```
andy-slac, line 25: No need to import `os.path` separately (though some people disagree).

python/lsst/daf/butler/core/location.py
```diff
@@ -19,45 +19,329 @@
...
+        Path using POSIX path separator
+    """
+    if IS_POSIX:
+        return path
+
+    # Split on OS path separator but have to restore root directory
+    # Untested on Windows
+    ospaths = path.split(os.sep)
+    if ospaths[0] == "":
+        ospaths[0] = Path(path).root
```
andy-slac: I am confused by this but it's probably just me - on non-Posix system `root` will be local root, not Posix root? It looks like you can return a combination of both local and Posix separators here.

python/lsst/daf/butler/core/location.py
```diff
@@ -19,45 +19,329 @@
...
+    ospath : `str`
+        Path using OS path separator
+    """
+    if IS_POSIX:
+        return path
+
+    # Have to restore the root directory after splitting
+    posixpaths = path.split(posixpath.sep)
+    if posixpaths[0] == "":
+        posixpaths[0] = posixpath.sep
```
andy-slac: `os.sep`?

python/lsst/daf/butler/core/location.py
```diff
@@ -19,45 +19,329 @@
...
+    Returns
+    -------
+    ospath : `str`
+        Path using OS path separator
+    """
+    if IS_POSIX:
+        return path
+
+    # Have to restore the root directory after splitting
+    posixpaths = path.split(posixpath.sep)
```
andy-slac: Could `path` be empty?

python/lsst/daf/butler/core/location.py
```diff
@@ -19,45 +19,329 @@
...
+    Parameters
+    ----------
+    uri : `str` or `urllib.parse.ParseResult`
+        URI in string form.  Can be scheme-less if referring to a local
+        filesystem path.
+    root : `str`, optional
+        When fixing up a relative path in a `file` scheme or if scheme-less,
+        use this as the root. Must be absolute.  If `None` the current
+        working directory will be used.
+    forceAbsolute : `bool`
```
andy-slac: `optional`?

python/lsst/daf/butler/core/location.py
```diff
@@ -19,45 +19,329 @@
...
+        If `True`, scheme-less relative URI will be converted to an absolute
+        path using a `file` scheme. If `False` scheme-less URI will remain
+        scheme-less and will not be updated to `file` or absolute path.
+    """
+
+    def __init__(self, uri, root=None, forceAbsolute=True):
+        if isinstance(uri, str):
+            parsed = urllib.parse.urlparse(uri)
+        elif isinstance(uri, urllib.parse.ParseResult):
+            parsed = copy.copy(uri)
```
andy-slac, line 122: What if it is something else? I think code below will crash with `NameError`.

python/lsst/daf/butler/core/location.py
```diff
@@ -19,45 +19,329 @@
...
+
+        parsed = self._fixupFileUri(parsed, root=root, forceAbsolute=forceAbsolute)
+        self._uri = parsed
+
+    @property
+    def scheme(self):
```
andy-slac, line 130: No need to document properties?

python/lsst/daf/butler/core/location.py
```diff
@@ -19,45 +19,329 @@
...
+        instance.
+
+        Returns
+        -------
+        new : `ButlerURI`
+            New `ButlerURI` object with updated values.
+        """
+        return self.__class__(self._uri._replace(**kwargs))
+
+    def replaceFile(self, newfile):
```
andy-slac: Some potential for confusion here - `replace()` returns new copy, `replaceFile()` updates object in place. Maybe call it `updateFile`?

python/lsst/daf/butler/core/location.py
```diff
@@ -19,45 +19,329 @@
...
+        -----
+        Updates the URI in place.
+        """
+        if self.scheme:
+            # POSIX
+            pathclass = posixpath
+        else:
+            pathclass = os.path
+
+        print(f"Processing path {self.path}")
```
andy-slac: Should it be here? :printer:

python/lsst/daf/butler/core/location.py
```diff
@@ -19,45 +19,329 @@
...
+    if posixpaths[0] == "":
+        posixpaths[0] = posixpath.sep
+    return os.path.join(*posixpaths)
+
+
+class ButlerURI:
+    """Convenience wrapper around URI parsers.
+
+    Provides access to URI components and can convert file
+    paths into absolute path URIs. Schemeless URIs are treated as if
```
andy-slac: `Scheme-less` for consistency with all other uses?

python/lsst/daf/butler/core/location.py
```diff
@@ -19,45 +19,329 @@
...
+        ----------
+        parsed : `~urllib.parse.ParseResult`
+            The result from parsing a URI using `urllib.parse`.
+        root : `str`, optional
+            Path to use as root when converting relative to absolute.
+            If `None`, it will be the current working directory. This
+            is a local file system path, not a URI.
+        forceAbsolute : `bool`
+            If `True`, scheme-less relative URI will be converted to an
+            absolute path using a `file` scheme. If `False` scheme-less URI
```
andy-slac: double backticks around `file`? Also in class docstring.

tests/test_location.py
```diff
@@ -0,0 +1,126 @@
...
+        print(f"Factory created: {factory}")
+
+        pathInStore = "relative/path/file.ext"
+        loc1 = factory.fromPath(pathInStore)
+
+        self.assertEqual(loc1.path, posixpath.join("/butler/datastore", pathInStore))
+        self.assertEqual(loc1.pathInStore, pathInStore)
+        self.assertTrue(loc1.uri.startswith("https://"))
+        self.assertTrue(loc1.uri.endswith("file.ext"))
+        loc1.updateExtension("fits")
```
andy-slac, line 125: No assert for this?

andy-slac (approved):
Looks OK, though OS vs POSIX handling feels confusing in places (not sure how it can be improved). Few comments below.

python/lsst/daf/butler/core/location.py
timj: I generally prefer to be explicit. Python doesn't guarantee that importing a top level will import all sublevels (even if that indeed happens for `os`).

python/lsst/daf/butler/core/location.py
timj: I'm revisiting this. I had forgotten that `split` does not split the path into all components.

timj:
I've cleaned up the os2posix/posix2os code a bit, making use of PurePath  and PurePosixPath. I think it might be a bit more robust now (with the caveat that it never gets called because I'm not on Windows).

python/lsst/daf/butler/core/location.py
timj: Yes. This was wrong. Fixed now.

python/lsst/daf/butler/core/location.py
timj: Hopefully now fixed.

python/lsst/daf/butler/core/location.py
timj: Empty as in giving a relative path of ""?

andy-slac: I do not know, I guess it depends on caller, my question was what happens in this code if path is empty (maybe accidentally).

timj: With the new implementation if you pass in the empty string you get back the empty string (which seems correct to me).

timj:
@DinoBektesevic I've now merged this PR so feel free to rebase #159 on top of master.  Hopefully this will let you remove a lot of URI mangling from your PR.

PR #168, opened 2019-06-11
DM-20054: Normalize dimensions up-front to avoid incorrect comparisons

TallJimbo:


python/lsst/daf/butler/core/dimensions/sets.py
```diff
@@ -402,57 +208,153 @@ def findIf(self, predicate, default=None):
...
-    `DimensionGraph`.
+    def __str__(self):
+        return "{{{}}}".format(", ".join(self._elements.keys()))
 
-    Parameters
-    ----------
-    names : iterable of `str`
-        The names of elements to conceptually include in the set.
+    def __hash__(self):
+        return hash(tuple(self._elements.keys()))
```
andy-slac, line 220: Don't you need `sorted()` somewhere here?

tests/config/basic/posixDatastore.yaml
```diff
@@ -26,7 +26,6 @@ datastore:
     ThingOne: lsst.daf.butler.formatters.yamlFormatter.YamlFormatter
     datasetType.component: lsst.daf.butler.formatters.yamlFormatter.YamlFormatter
     pvi: lsst.daf.butler.formatters.pickleFormatter.PickleFormatter
-    visit+physical_filter+instrument: lsst.daf.butler.formatters.yamlFormatter.YamlFormatter
```
timj, line 29: Why did this line get removed?

timj (approved):

tests/config/basic/posixDatastore.yaml
TallJimbo: The first half of the story is that once I started using real, expanded dimensions in tests, this override started getting picked up in places where the test code didn't expect it, causing problems.  The one I remember was `DatastoreTests.testBasicPutGet`, which started failing because this override always beat out the `StorageClass`-level setting that switched from YAML to JSON to pickle, causing an attempt to write the same YAML file three times.

The second half is why I didn't replace it with something else: after this problem, I rolled back to before I changed any of the dimensions in the tests and removed this line, hoping that would identify which test depended on it so I could find some other way to make it work.  And nothing failed, so I concluded that it wasn't actually doing anything.

But now that I'm thinking about it again, I'm worried that I might have been fooled by the sconsUtils logic that only re-runs failed tests unless they all succeeded last time, and I should go repeat that experiment before merging.

timj: The main driver for the test was to show that dimensions can be specified in formatters and will override storage classes, and also that an instrument specific override of dimensions work. The dimensions in the test don't have to be those particular ones but it is a bit trickier now that they have to be _real_ dimensions.

TallJimbo: Ok, I re-did the experiment, with the same result: if I go back to master, and just remove this line from this supporting config, none of the tests fail.  And I don't see any tests in either `test_datastore.py` or `test_formatters.py` that seem to be about what you just described.  Is there somewhere else I should be looking for this test?

timj: Sounds like I wasn't testing things properly 👎

TallJimbo: Ok, I'm gonna go ahead and merge this (after addressing other comments), and let you follow up with more tests later at your discretion.  FWIW, the "abstract_filter" and especially "label" dimensions are particularly useful for tests if you don't want the complexity of dimension relationships.

python/lsst/daf/butler/core/dimensions/sets.py
TallJimbo: Happily, no.  Now that there's no unnormalized `DimensionNameSet` anymore, the ordering of the keys is deterministic (topological and then lexicographically sorted).

andy-slac: Maybe add a comment here so that I don't ask the same question again in the next review?

PR #169, opened 2019-06-27
DM-19988: support code (and removal of dead code) for QuantumGraph generation rewrite

TallJimbo:


config/schema.yaml
```diff
@@ -1030,6 +1030,66 @@ schema:
...
+        primary_key: true
+        nullable: false
+        doc: label used to identify a group of master calibration products.
+      -
+        name: visit
+        type: int
+        primary_key: true
+        nullable: false
+        doc: >
+          Unique (with instrument) integer identifier for an visit.
```
andy-slac: a visit

python/lsst/daf/butler/core/quantum.py
```diff
@@ -21,49 +21,88 @@
...
+        this Quantum.  If not provided, ``taskClass`` must be.
+    taskClass : `type`, optional
+        The Task class that executed or will execute this Quantum.  If not
+        provided, ``taskName`` must be.  Overrides ``taskName`` if both are
+        provided.
+    dataId : `DataId`, optional
+        The dimension values that identify this `Quantum`.
+    run : `Run`, optional
         The Run this Quantum is a part of.
+    initInputs : collection of `DatasetRef`
```
andy-slac: `optional`?

python/lsst/daf/butler/core/quantum.py
```diff
@@ -21,49 +21,88 @@
...
-        self._task = task
+    __slots__ = ("_taskName", "_taskClass", "_dataId", "_run",
+                 "_initInputs", "_predictedInputs", "_actualInputs", "_outputs")
+
+    def __init__(self, *, taskName=None, taskClass=None, dataId=None, run=None,
+                 initInputs=None, predictedInputs=None, actualInputs=None, outputs=None,
+                 **kwargs):
+        super().__init__(**kwargs)
+        if taskClass is not None:
+            taskName = f"{taskClass.__module__}.{taskClass.__name__}"
```
andy-slac, line 79: Everyone seem to prefer f-strings for simple string operations, I still like old fashioned

    taskClass.__module__ + "." + taskClass.__name__

python/lsst/daf/butler/core/quantum.py
```diff
@@ -21,49 +21,88 @@
...
+        self._taskClass = taskClass
         self._run = run
-        self._predictedInputs = {}
-        self._actualInputs = {}
-        self._outputs = {}
+        self._dataId = dataId
+        if initInputs is None:
+            initInputs = {}
+        elif not hasattr(initInputs, "keys"):
+            initInputs = {ref.datasetType: ref for ref in initInputs}
```
andy-slac, line 87: Could it be that there is more than one `ref` in `initInputs` with the same dataset type?

python/lsst/daf/butler/core/quantum.py
```diff
@@ -21,49 +21,88 @@
...
+        self._initInputs = NamedKeyDict(initInputs)
+        self._predictedInputs = NamedKeyDict(predictedInputs if predictedInputs is not None else {})
+        self._actualInputs = NamedKeyDict(actualInputs if actualInputs is not None else {})
+        self._outputs = NamedKeyDict(outputs if outputs is not None else {})
```
andy-slac: Some serious typing could be saved if `NamedKeyDict` constructor accepted `None`. Or can't you use old pattern `outputs or {}`?

andy-slac (approved):
Looks OK, few minor comments.

python/lsst/daf/butler/core/quantum.py
TallJimbo: Yeah, I do like the f-string version better.  In this case there isn't much of a difference, but I like using the same pattern for simple and more complex cases that seem to be part of the same family of patterns, and I think this qualifies.

python/lsst/daf/butler/core/quantum.py
TallJimbo: No, the `PipelineTask` declarations already assume all `initInput` and `initOutput` dataset types are scalar (and they have to be because they have empty data IDs).

python/lsst/daf/butler/core/quantum.py
TallJimbo: I don't think `None` makes sense for `NamedKeyDict` in general, so I'd prefer not to move the handling there.  And I once asked about a variant of the `outputs or {}` pattern on Slack/`#software-def`, and it was not popular (people considered it a Perl pattern that was a Python antipattern).  But I've come up with another solution: it's safe to pass an empty tuple `NamedKeyDict`'s constructor, so I'll just use that as the default for these arguments.

python/lsst/daf/butler/sql/queryBuilder.py
```diff
@@ -260,12 +260,12 @@ def whereSqlExpression(self, sqlExpression, op=and_):
...
 
     def whereParsedExpression(self, expression, op=and_):
-        """Add a SQL expression to the query's WHERE clause.
+        """Add a parsed dimension expression to the query's WHERE clause.
 
         Parameters
         ----------
-        sqlExpression
-            SQLAlchemy boolean column expression.
+        expression : str
```
andy-slac: backticks around str?

andy-slac (approved):
Looks OK, one minor comment

PR #173, opened 2019-07-26
DM-16297: Add StorageClasses and formatter declarations for CameraGeom.

timj:


TallJimbo (commented):
If you ever have the time and inclination to chop off the rest of the "TablePersistable" prefixes, please go ahead.  I've also come to the conclusion that they're not useful.

timj:
Ok, I think I can do that on another ticket. I realize that this particular change requires fixes in ip_isr and obs_subaru so I'll do those.  Changing `TablePersistableX` to `X` should be straightforward.

timj:
Actually, I think it makes sense to merge the TablePersistable removal first and then do this ticket.

czwa (approved):

PR #175, opened 2019-07-26
DM-20763: Add initial support for Gen3 Butler to obs_decam

parejkoj:


python/lsst/daf/butler/instrument.py
```diff
@@ -47,6 +49,21 @@ class Instrument(metaclass=ABCMeta):
...
+    """Paths to config files to read for specific Tasks.
+
+    The paths in this list should contain files of the form `task.py`, for
+    each of the Tasks that requires special configuration.
+    """
+
+    filterDefinitions = None
+    """`~lsst.obs.base.FilterDefinitions`: The definitions of the filters used
+    by this instruments.
+    """
```
TallJimbo: What do you think about making this an `abstractmethod` getter or property?  I think that better captures the intent that this is something subclasses must override, and we might appreciate the extra flexibility it would provide for programmatically constructing the list of filters someday.

python/lsst/daf/butler/instrument.py
```diff
@@ -47,6 +49,21 @@ class Instrument(metaclass=ABCMeta):
...
+
+    def __init__(self, *args, **kwargs):
+        self.filterDefinitions.defineFilters()
```
TallJimbo, line 68: I don't think this helps in general; there is no guarantee that the `Instrument` class will be instantiated at any time after it is used to initialize a data repository.  I don't think we can do anything about this problem that isn't a workaround for the singleton in afw, so I don't have any great ideas, but given that it's a workaround I'd rather not establish it as an expected side-effect of `Instrument` construction.

python/lsst/daf/butler/instrument.py
```diff
@@ -58,13 +75,40 @@ def __init_subclass__(cls, **kwargs):
...
     @abstractmethod
     def register(self, registry):
         """Insert instrument, physical_filter, and detector entries into a
         `Registry`.
         """
         raise NotImplementedError()
 
+    def _registerFilters(self, registry, dataId):
+        """Register the physical and abstract filter Dimension relationships.
+
```
TallJimbo, line 100: Why is this marked private?  Is it conceptually protected, as it's a utility function for subclasses to call?

timj (approved):

python/lsst/daf/butler/instrument.py
```diff
@@ -58,13 +75,40 @@ def __init_subclass__(cls, **kwargs):
...
+        Parameters
+        ----------
+        registry : `lsst.daf.butler.core.Registry`
+            The registry to add dimensions to.
+        dataId : `dict`
+            The dataId to register the filters with.
```
TallJimbo: I think the `dataId` argument always has to be `{"instrument": self.getName()}`.  Might as well just do that instead of taking it as an argument.  Or, equivalently:
```
registry.addDimensionEntry(
    "physical_filter",
    instrument=self.getName(),
    physical_filter=filter.physical_filter,
    abstract_filter=filter.abstract_filter
)
```

python/lsst/daf/butler/instrument_tests.py
```diff
@@ -0,0 +1,95 @@
...
+These are not tests themselves, but can be subclassed (plus unittest.TestCase)
+to get a functional test of an Instrument.
+"""
+
+import abc
+import dataclasses
+import os
+
+from lsst.daf.butler.core import Registry
+from lsst.daf.butler.core.butlerConfig import ButlerConfig
```
TallJimbo: Better to import these directly from `lsst.daf.butler`; the existence of `core` and any modules below that is considered an implementation detail (and only symbols lifted to package scope from it are considered public).

python/lsst/daf/butler/instrument_tests.py
```diff
@@ -0,0 +1,95 @@
...
+    """The name of the Camera this instrument describes."""
+
+    nDetectors: int
+    """The number of detectors in the Camera."""
+
+    firstDetectorName: str
+    """The name of the first detector in the Camera."""
+
+    physical_filters: {str}
+    """The names of the physical filters that are registered."""
```
TallJimbo: Listing *all* physical filters here strikes me as more of a fragile don't-repeat-yourself violation than a useful test. Maybe spot-checking that a *few* physical filters exist (and are associated with the right abstract filter) would be better?

tests/test_instrument.py
```diff
@@ -33,10 +33,21 @@
 
 class DummyCam(Instrument):
 
+    class FilterDefinitions:
+        """Stopgap until Instrument is moved into obs_base and we can use the
+        real FilterDefinitions there.
+        """
+        def defineFilters(self):
+            pass
+    filterDefinitions = FilterDefinitions()
```
TallJimbo, line 42: It looks like the `register` implementation below doesn't call `_registerFilters` or otherwise use `FilterDefinitions`.  Do you need this shim?

TallJimbo (approved):
Only minor comments.

python/lsst/daf/butler/instrument.py
parejkoj: That's what I did in `fitsRawFormatterBase`. I think one of those should just delegate to the other, but I'm not sure what the correct dependency order is there.

python/lsst/daf/butler/instrument.py
parejkoj: It's there to be called by `register`. I've made a comment to that effect in the docstring.

python/lsst/daf/butler/instrument.py
parejkoj: I think I'm going to need some advice on how to manage this: I suspect this is the cause of my current ci_hsc failures (the afw `Filters` not being defined at the right time).

TallJimbo: I think it would work to put this in the constructor for the HSC raw data formatter, behind some kind of guard that checks that it's only ever done once.  Or at least that's a place where it should go; there may be other places (also behind the same call-once guard).

Now that I think about it, it may make sense to put that guard in `FilterDefinitionCollection.defineFilters`, and make it so the behavior is that:
 - calling `defineFilters` multiple times on the same `FilterDefinitionCollection` is a silent no-op;
 - calling `defineFilters` on different instances of `FilterDefinitionCollection` is an error.  That's unfortunate, but I don't think we can expect trying to put multiple instruments' filters into the afw singleton to lead to anything but pain.

parejkoj: Why not put that `defineFilters` call into `fitsRawFormatterBase.__init__`? Otherwise, it would have to go into every child class's `__int__`. I think we need to have some guarantee that the filters are defined *somewhere*, and enforcing it in a parent class seems the most reasonable to me.

timj: This is all going to get very difficult once we start having gen 3 repositories with multiple instruments in them.

TallJimbo: > Why not put that defineFilters call into `fitsRawFormatterBase.__init__`?

:+1:

I still think that'll be necessary but not sufficient, as it doesn't solve the problem with having filters defined before non-`raw` `Exposures` are read.  I don't know how to deal with that (or how we've avoided dealing with it) so far.  Maybe the `Exposure` read code is more robust against undefined filters than I'm guessing.

> This is all going to get very difficult once we start having gen 3 repositories with multiple instruments in them.

Not just difficult.  I consider multi-instrument repositories effectively blocked on the removal of the `Filter` singleton in afw.

PR #176, opened 2019-07-31
DM-20842: Change Formatter API

timj:
The motivation for this is to simplify formatters so that they know they can store state in their instance without fear that a Datastore is going to reuse the Formatter for some other location.
    
* This changes the read() and write() methods since they no longer need FileDescriptor parameter.
* All the formatters need to be tweaked.
* Datastore needs changing to ensure that FileDescriptor is created a bit earlier.
* getFormatter gets an extra argument which can be None for formatters that will not be using read/write at all.

@TallJimbo , @parejkoj does this look okay to you?

### DIFF
--- a/python/lsst/daf/butler/core/formatter.py
+++ b/python/lsst/daf/butler/core/formatter.py
@@ -27,6 +27,7 @@ import logging
 from .configSupport import processLookupConfigs
 from .mappingFactory import MappingFactory
 from .utils import getFullTypeName
+from .fileDescriptor import FileDescriptor
 
 log = logging.getLogger(__name__)
 
@@ -34,6 +35,12 @@ log = logging.getLogger(__name__)
 class Formatter(metaclass=ABCMeta):
     """Interface for reading and writing Datasets with a particular
     `StorageClass`.
+
+    Parameters
+    ----------
+    fileDescriptor : `FileDescriptor`
+        Identifies the file to read or write, and the associated storage
+        classes and parameter information.
     """
 
     unsupportedParameters = frozenset()
@@ -42,6 +49,15 @@ class Formatter(metaclass=ABCMeta):
     are supported.
     """
 
+    def __init__(self, fileDescriptor):
+        if fileDescriptor is not None and not isinstance(fileDescriptor, FileDescriptor):
+            raise TypeError("File descriptor must be a FileDescriptor")
+        self._fileDescriptor = fileDescriptor
+
+    @property
+    def fileDescriptor(self):
+        return self._fileDescriptor
+
     @classmethod
     def name(cls):
         """Returns the fully qualified name of the formatter.
@@ -49,14 +65,11 @@ class Formatter(metaclass=ABCMeta):
         return getFullTypeName(cls)
 
     @abstractmethod
-    def read(self, fileDescriptor, component=None):
+    def read(self, component=None):
         """Read a Dataset.
 
         Parameters
         ----------
-        fileDescriptor : `FileDescriptor`
-            Identifies the file to read, type to read it into and parameters
-            to be used for reading.
         component : `str`, optional
             Component to read from the file. Only used if the `StorageClass`
             for reading differed from the `StorageClass` used to write the
@@ -70,15 +83,13 @@ class Formatter(metaclass=ABCMeta):
         raise NotImplementedError("Type does not support reading")
 
     @abstractmethod
-    def write(self, inMemoryDataset, fileDescriptor):
+    def write(self, inMemoryDataset):
         """Write a Dataset.
 
         Parameters
         ----------
         inMemoryDataset : `InMemoryDataset`
             The Dataset to store.
-        fileDescriptor : `FileDescriptor`
-            Identifies the file to write.
 
         Returns
         -------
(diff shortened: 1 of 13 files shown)


python/lsst/daf/butler/core/formatter.py
```diff
@@ -27,13 +27,20 @@
...
 log = logging.getLogger(__name__)
 
 
 class Formatter(metaclass=ABCMeta):
     """Interface for reading and writing Datasets with a particular
     `StorageClass`.
+
+    Parameters
+    ----------
+    fileDescriptor : `FileDescriptor`
```
parejkoj: I'm not sure, but I think you still have to give the full namespace in order for this to become a link.

tests/test_formatter.py
```diff
@@ -45,15 +45,15 @@ def testRegistry(self):
         formatterTypeName = "lsst.daf.butler.formatters.fitsCatalogFormatter.FitsCatalogFormatter"
         storageClassName = "Image"
         self.factory.registerFormatter(storageClassName, formatterTypeName)
-        f = self.factory.getFormatter(storageClassName)
+        f = self.factory.getFormatter(storageClassName, None)
```
parejkoj: Are all of these `None`s really necessary here? Could you make wherever that is landing take a `None` default arg?

parejkoj (commented):
Just two comments, otherwise this looks fine.

python/lsst/daf/butler/core/formatter.py
timj: I was under the impression that classes in the local package didn't need to be fully qualified.

tests/test_formatter.py
timj: I decided to force FileDescriptor to take an argument.  This at least made it obvious where I needed to put an argument during testing. It's more explicit this way that you are saying you know that the formatter will never be used to read or write anything.

python/lsst/daf/butler/core/formatter.py
```diff
@@ -42,21 +49,27 @@ class Formatter(metaclass=ABCMeta):
     are supported.
     """
 
+    def __init__(self, fileDescriptor):
+        if fileDescriptor is not None and not isinstance(fileDescriptor, FileDescriptor):
+            raise TypeError("File descriptor must be a FileDescriptor")
+        self._fileDescriptor = fileDescriptor
```
TallJimbo, line 56: Seems like this might as well be a public attribute.

TallJimbo (approved):
:+1: I'm looking forward to seeing how this improves the concrete raw formatters; I think they'll really benefit from this change.

python/lsst/daf/butler/core/formatter.py
timj: No. It's a read-only property. There's a getter further down.

timj: Meaning I want to guarantee that people can't change the file location once the object is constructed.

timj:
Related pull requests in lsst/obs_subaru#217 and lsst/obs_base#165

I think I'm ready for a proper review. I rejigged things in the way suggested by @TallJimbo to allow for a `Formatter.getFormatterClass` method which makes the logic in datastore.ingest more explicit.

python/lsst/daf/butler/core/formatter.py
```diff
@@ -87,17 +99,38 @@ def write(self, inMemoryDataset, fileDescriptor):
...
-    def predictPath(self, location):
+    def predictPathFromLocation(cls, location):
         """Return the path that would be returned by write, without actually
         writing.
 
+        Parameters
+        ----------
         location : `Location`
-            The location to simulate writing to.
+            Location of file for which path prediction is required.
```
TallJimbo, line 118: I gather this is a root filename, and what's returned is the full thing with the extension (or some generalization thereof)?  This is mostly preexisting, but it's not currently clear from the docs what the distinction between input and output is.

python/lsst/daf/butler/core/storedFileInfo.py
```diff
@@ -50,9 +52,13 @@ class StoredFileInfo:
...
     def __init__(self, formatter, path, storageClass, checksum=None, size=None):
-        assert isinstance(formatter, str) or isinstance(formatter, Formatter)
-        if isinstance(formatter, Formatter):
+        if isinstance(formatter, str):
+            pass
+        elif isinstance(formatter, Formatter) or (inspect.isclass(formatter) and
+                                                  issubclass(formatter, Formatter)):
             formatter = formatter.name()
+        else:
+            raise ValueError(f"Supplied formatter '{formatter}' is not supported")
```
TallJimbo: What does "not supported" mean in this context?  That it's not actually a `Formatter`?

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
```diff
@@ -31,8 +31,18 @@ class FitsExposureFormatter(Formatter):
...
     """
     extension = ".fits"
+    _metadata = None
 
-    def readImageComponent(self, fileDescriptor, component):
+    @property
+    def metadata(self):
+        """The metadata read from this file. It will be stripped as
+        components are extracted from it.
+        """
```
TallJimbo, line 41: Docstring should include the return type.

TallJimbo (approved):

python/lsst/daf/butler/core/formatter.py
```diff
@@ -197,7 +234,57 @@ def getLookupKeys(self):
         """
         return self._mappingFactory.getLookupKeys()
 
-    def getFormatterWithMatch(self, entity):
+    def getFormatterClassWithMatch(self, entity):
+        """Get a the matching formatter class along with the matching registry
```
parejkoj: get a the?

python/lsst/daf/butler/core/formatter.py
```diff
@@ -197,7 +234,57 @@ def getLookupKeys(self):
...
+        key.
+
+        Parameters
+        ----------
+        entity : `DatasetRef`, `DatasetType` or `StorageClass`, or `str`
```
parejkoj: Either missing an `or`, or has one to many. Same in `getFormatterClass` below.

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
parejkoj: Almost: https://developer.lsst.io/python/numpydoc.html#documenting-class-properties

parejkoj (approved):
Couple of docstring comments, otherwise looks good.

python/lsst/daf/butler/core/formatter.py
timj: This method doesn't care about input or output. It just wants a `Location` object which refers to a place in the datastore -- this method generally applies the right file suffix and then returns the filename from the Location.

PR #178, opened 2019-08-05
DM-20915: Add str(Formatter) and force Formatter to require an arg for constructor

timj:
There is no longer any gain in allowing an optional FileDescriptor argument so remove that option.

### DIFF
--- a/python/lsst/daf/butler/core/formatter.py
+++ b/python/lsst/daf/butler/core/formatter.py
@@ -55,6 +55,9 @@ class Formatter(metaclass=ABCMeta):
             raise TypeError("File descriptor must be a FileDescriptor")
         self._fileDescriptor = fileDescriptor
 
+    def __str__(self):
+        return f"{self.name()}@{self.fileDescriptor.location.path}"
+
     @property
     def fileDescriptor(self):
         """FileDescriptor associated with this formatter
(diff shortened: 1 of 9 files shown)


python/lsst/daf/butler/core/formatter.py
```diff
@@ -50,11 +50,14 @@ class Formatter(metaclass=ABCMeta):
...
-    def __init__(self, fileDescriptor=None):
-        if fileDescriptor is not None and not isinstance(fileDescriptor, FileDescriptor):
+    def __init__(self, fileDescriptor):
+        if not isinstance(fileDescriptor, FileDescriptor):
             raise TypeError("File descriptor must be a FileDescriptor")
         self._fileDescriptor = fileDescriptor
 
+    def __str__(self):
+        return f"{self.name()}@{self.fileDescriptor.location.path}"
+
```
TallJimbo, line 60: Seems like it'd be straightforward to add the ideal kind of `__repr__`, too.

TallJimbo (approved):

python/lsst/daf/butler/core/formatter.py
timj: Yes, it was straightforward here but required I added repr() to FileDescriptor and Location and also that I cleaned up repr() for StorageClass to make it less chatty.  You now get a repr that looks something like:
```
lsst.daf.butler.formatters.jsonFormatter.JsonFormatter(FileDescriptor(Location('/Users/timj/work/lsstsw3/build/daf_butler/a', 'B'), StorageClass('xxx'), readStorageClass=StorageClass('yyy')))
```
@TallJimbo you might want to take a look at all the new code now.

TallJimbo: Looks good.  I didn't know about `{!r}`, and I'm glad that I now do.

PR #179, opened 2019-08-06
DM-13361: Add support for S3 backed Datastore

DinoBektesevic:
As per @timj request I have closed PR https://github.com/lsst/daf_butler/pull/159 and have reopened this PR instead to allow Jenkins tests to run. 

The mentioned PR is a continuation from draft PR https://github.com/lsst/daf_butler/pull/147 and adds support for S3 backed datastore.


python/lsst/daf/butler/core/location.py
```diff
@@ -141,6 +141,26 @@ def path(self):
         """The path component of the URI."""
         return self._uri.path
 
+    @property
+    def ospath(self):
+        """Path component of the URI localized to current OS."""
+        if self.scheme == 's3':
```
timj: We should probably only have ospath return something if we have file or scheme-less. If scheme-less it should return the path directly and if `file` it should return posix2os.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,736 @@
...
+        else:
+            raise NotImplementedError(f"Scheme type {srcUri.scheme} not supported.")
+
+        # Transfer is generaly None when put calls ingest. In that case file is
+        # uploaded in put, or already in proper location, so source location
+        # must be inside repository. In other cases, created target location
+        # must be inside root and source file must be deleted when 'move'd.
+        if transfer is None:
+            rootUri = ButlerURI(self.root)
+            if srcUri.scheme == "file":
```
timj: This should be saying if scheme is anything except s3 then it can't possibly be transfer=None because the dataset can't be in the s3 datastore.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -0,0 +1,736 @@
...
+                                   "Ingesting local data to S3Datastore without upload "
+                                   "to S3 is not allowed.")
+            elif srcUri.scheme == "s3":
+                if not srcUri.path.startswith(rootUri.path):
+                    raise RuntimeError(f"'{srcUri}' is not inside repository root '{rootUri}'.")
+            p = pathlib.PurePosixPath(srcUri.relativeToPathRoot)
+            pathInStore = str(p.relative_to(rootUri.relativeToPathRoot))
+            tgtLocation = self.locationFactory.fromPath(pathInStore)
+        elif transfer == "move" or transfer == "copy":
+            if srcUri.scheme == "file":
```
timj: This should also support scheme-less (local files). Also put an else at the end as a sanity check.

timj (approved):
Some minor things to fix but this looks great. Thanks for doing all the work.

DinoBektesevic:
@timj fixed the requests, Jenkins tests please.

PR #180, opened 2019-08-15
Retrieve DB authentication data.

ktlim:


python/lsst/daf/butler/core/dbAuth.py
```diff
@@ -0,0 +1,227 @@
...
+            # Check for IP v6
+            if matchHost.startswith("["):
+                match = re.search(r"^(\[[0-9a-f:]+\]):(\d+)$", matchHost)
+                if match:
+                    matchHost, matchPort = match.groups()
+                # Prep for fnmatch
+                matchHost = re.sub(r"^\[", "[[]", matchHost)
+            elif ":" in matchHost:
+                matchHost, matchPort = matchHost.split(":")
+            if matchPort is not None and (port is None or port != matchPort):
```
DinoBektesevic: I'm not sure if this will work if port is an `int` because `matchPort` is always an `str`. I know the documentation states that `port` must be `str` but would it be possible to make a `str()` cast here and relax that requirement?

ktlim: What do you get out of SQLAlchemy?

DinoBektesevic: I think sqlalchemy would return me an `str` while `urlib.parse.urlparse("postgresql://test:testpass@host:1234/dbname")` would return an `int` and explicitly defined keys in the butler yaml config would depend whether or not the given port was wrapped in `""` or not

ktlim: Changed to enable numeric ports.

python/lsst/daf/butler/core/dbAuth.py
```diff
@@ -0,0 +1,216 @@
...
+
+        try:
+            with open(secretPath) as secretFile:
+                self.authList = yaml.safe_load(secretFile)
+        except Exception as exc:
+            raise DbAuthError(
+                "Unable to load DbAuth configuration file: " +
+                secretPath) from exc
+
+    def getAuth(self, drivername, username, host, port, database):
```
andy-slac, line 85: Could some parameters made default, I'd prefer not to provide port number if I do not know it?

andy-slac: Another idea - could this method have an overload which takes url string and returns new string with username and password in it, I think it would be even more usable.

ktlim: The idea was that the arguments are the components of the SQLAlchemy URL object.  Making parameters defaultable would I think have to make the order different than in the URL and less memorable.

ktlim: (And you don't have to provide the port number; None or "" are fine.)

If I were going to take a string URL, I'd make that a separate method.  That would be easy enough to do if it's useful.

Returning a re-composed URL from a new URL-taking method shouldn't be too hard.  It's not possible with the current method, as all parts of the URL aren't available.

Is anyone going to use such a new method to make it worthwhile writing the tests?

andy-slac: If this is purely sqlalchemy-oriented then it may be more natural to accept `URL` instance and return modified `URL`. I was thinking about other possible use cases (e.g. PPDB with Cassandra) which do not use sqlalchemy at all but could still mimic URL-style connection string, but given that this code is in `daf_butler` it is probably not very reusable for other clients anyways.

timj: We could in theory move it into a dbutils package if needs be.

tests/data/dbAuth/good_test.yaml
```diff
@@ -0,0 +1,40 @@
+[
+{
+        url: "postgresql://user@host.example.com:5432/my_database",
```
timj: I don't like this YAML form because of all the white space and extra brackets everywhere.  Can't we use:
```yaml
- password: test1
  url: postgresql://user@host.example.com:5432/my_database
- password: test2
  url: postgresql://user@host.example.com/my_database
- password: test3
  url: postgresql://host.example.com:5432/my_database
  username: user
- password: test4
  url: postgresql://host.example.com/my_database
  username: user
- password: test5
  url: postgresql://*.example.com/my_*
  username: user
- password: test6
  url: postgresql://*.example.com/
  username: user
- password: test7
  url: postgresql://*.example.com
  username: user
- password: test8
  url: postgresql://*
  username: user
```
?

python/lsst/daf/butler/core/dbAuth.py
ktlim: While it is SQLAlchemy-oriented, I didn't want it to actually depend on SQLAlchemy.  I was hoping that others could reuse it in the future, but it wasn't clear that we have other immediate clients.

tests/data/dbAuth/good_test.yaml
ktlim: Sure.  Although password-first seems strange.

timj: Yes, I didn't notice. That's the order that came out when I dumped the list that I had read from the initial yaml. It sorted the keys.

timj (approved):

andy-slac (approved):
I would like it to be more generic but for that  I guess we'll have to wait until we get some other non-sqlalchemy clients. Looks OK to me as it satisfies our immediate need.

andy-slac (approved):
new `getUrl` method looks OK. SQLAlchemy URL scheme part can have optional driver name in it (`getAuth` is confusing a bit because it call schema a driver while in sqlalchemy scheme is a dialect+driver). DO we support `schema+driver`, or do we want to support it? Something like `postgresql+psycopg2://andy@localhost/ppdb` should probably match `postgresql://andy@localhost/ppdb` in the DbAuth file?

ktlim:
I was thinking of allowing wildcards in the scheme/driver to handle this, but parsing then becomes problematic so I went with the simpler exact string match.  Since it's only a convenience, I'd prefer to leave it out for now.

python/lsst/daf/butler/core/dbAuth.py
```diff
@@ -212,5 +212,47 @@ def getAuth(self, drivername, username, host, port, database):
...
+        Raises
+        ------
+        DbAuthError
+            Raised if the input is missing elements, an authorization
+            dictionary is missing elements, the authorization file is
+            misconfigured, or no matching authorization is found.
+
+        See also
+        --------
+        `getAuth`
```
andy-slac, line 241: Somehow sphinx does not like this line:
```
Exception occurred:
  File "/home/salnikov/.local/lib/python3.7/site-packages/numpydoc/docscrape.py", line 255, in parse_item_name
    raise ParseError("%s is not a item name" % text)
numpydoc.docscrape.ParseError: `getAuth` is not a item name in 'Fill in a username and password in a database connection URL.\n\nThis function parses the URL and calls `getAuth`.\n\nParameters\n----------\nurl : `str`\n    Database connection URL.\n\nReturns\n-------\nurl : `str`\n    Database connection URL with username and password.\n\nRaises\n------\nDbAuthError\n    Raised if the input is missing elements, an authorization\n    dictionary is missing elements, the authorization file is\n    misconfigured, or no matching authorization is found.\n\nSee also\n--------\n`getAuth`'
The full traceback has been saved in /tmp/sphinx-err-djoayeel.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
```
It stops crashing if I change this line to something like

    Method `getAuth`

but then there is no `getAuth` link in the output.

andy-slac: Apparently backticks should not be used in `see also` section, stripping backticks stops crashing and link is generated OK.

PR #183, opened 2019-08-19
DM-20953: Add support for range selection to gen3 expression parser

andy-slac:
This adds range literal support inside IN() expressions and adds
translation of that syntax to sqlalchemy. With narrow range it just adds
all its elements to IN() list, if list becomes long (longer than 1000
items) then additional BETWEEN ... AND MOD() are generated.

### DIFF
--- a/python/lsst/daf/butler/exprParser/exprTree.py
+++ b/python/lsst/daf/butler/exprParser/exprTree.py
@@ -214,10 +214,13 @@ class RangeLiteral(Node):
         Start value of a range.
     stop : `int`
         End value of a range, inclusive, same or higher than ``start``.
-    stride : `int` or `None`
-        Stride value, must be positive, can be `None`.
+    stride : `int` or `None`, optional
+        Stride value, must be positive, can be `None` which means that stride
+        was not specified. Consumers are supposed to treat `None` the same way
+        as stride=1 but for some consumers it may be useful to know that
+        stride was missing from literal.
     """
-    def __init__(self, start, stop, stride=1):
+    def __init__(self, start, stop, stride=None):
         self.start = start
         self.stop = stop
         self.stride = stride
@@ -227,9 +230,7 @@ class RangeLiteral(Node):
         return visitor.visitRangeLiteral(self.start, self.stop, self.stride, self)
 
     def __str__(self):
-        res = f"{self.start}..{self.stop}"
-        if self.stride is not None:
-            res += f":{self.stride}"
+        res = f"{self.start}..{self.stop}" + (f":{self.stride}" if self.stride else "")
         return res
 
 
--- a/python/lsst/daf/butler/exprParser/exprTree.py
+++ b/python/lsst/daf/butler/exprParser/exprTree.py
@@ -201,6 +201,38 @@ class Identifier(Node):
         return "{name}".format(**vars(self))
 
 
+class RangeLiteral(Node):
+    """Node representing range literal appearing in `IN` list.
+
+    Range literal defines a range of integer numbers with start and
+    end of the range (with inclusive end) and optional stride value
+    (default is 1).
+
+    Attributes
+    ----------
+    start : `int`
+        Start value of a range.
+    stop : `int`
+        End value of a range, inclusive, same or higher than ``start``.
+    stride : `int` or `None`
+        Stride value, must be positive, can be `None`.
+    """
+    def __init__(self, start, stop, stride=1):
+        self.start = start
+        self.stop = stop
+        self.stride = stride
+
+    def visit(self, visitor):
+        # Docstring inherited from Node.visit
+        return visitor.visitRangeLiteral(self.start, self.stop, self.stride, self)
+
+    def __str__(self):
+        res = f"{self.start}..{self.stop}"
+        if self.stride is not None:
+            res += f":{self.stride}"
+        return res
+
+
 class IsIn(Node):
     """Node representing IN or NOT IN expression.
 
--- a/python/lsst/daf/butler/exprParser/parserLex.py
+++ b/python/lsst/daf/butler/exprParser/parserLex.py
@@ -38,7 +38,7 @@ from .ply import lex
 #  Local non-exported definitions --
 # ----------------------------------
 
-_RE_RANGE = re.compile(r"(-?\d+)\s*\.\.\s*(-?\d+)(\s*:\s*(\d+))?")
+_RE_RANGE = r"(?P<start>-?\d+)\s*\.\.\s*(?P<stop>-?\d+)(\s*:\s*(?P<stride>[1-9]\d*))?"
 """Regular expression to match range literal in the form NUM..NUM[:NUM],
 this must match t_RANGE_LITERAL docstring.
 """
@@ -58,7 +58,7 @@ class ParserLexError(Exception):
     remain : str
         Remaining non-parsed part of the expression
     pos : int
-        Current parsing posistion, offset from beginning of expression in
+        Current parsing position, offset from beginning of expression in
         characters
     lineno : int
         Current line number in the expression
@@ -154,15 +154,13 @@ class ParserLex:
         t.value = t.value[1:-1]
         return t
 
-    # range literal in format N..M[:S], spaces allowed
-    # docstring must match _RE_RANGE regexp
+    # range literal in format N..M[:S], spaces allowed, see _RE_RANGE
+    @lex.TOKEN(_RE_RANGE)
     def t_RANGE_LITERAL(self, t):
-        r"""(-?\d+)\s*\.\.\s*(-?\d+)(\s*:\s*(\d+))?
-        """
-        match = _RE_RANGE.match(t.value)
-        start = int(match.group(1))
-        stop = int(match.group(2))
-        stride = match.group(4)
+        match = re.match(_RE_RANGE, t.value)
+        start = int(match.group("start"))
+        stop = int(match.group("stop"))
+        stride = match.group("stride")
         if stride is not None:
             stride = int(stride)
         t.value = (start, stop, stride)
--- a/python/lsst/daf/butler/exprParser/parserLex.py
+++ b/python/lsst/daf/butler/exprParser/parserLex.py
@@ -38,6 +38,11 @@ from .ply import lex
 #  Local non-exported definitions --
 # ----------------------------------
 
+_RE_RANGE = re.compile(r"(-?\d+)\s*\.\.\s*(-?\d+)(\s*:\s*(\d+))?")
+"""Regular expression to match range literal in the form NUM..NUM[:NUM],
+this must match t_RANGE_LITERAL docstring.
+"""
+
 # ------------------------
 #  Exported definitions --
 # ------------------------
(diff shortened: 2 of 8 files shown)


python/lsst/daf/butler/exprParser/exprTree.py
```diff
@@ -201,6 +201,38 @@ def __str__(self):
...
+    end of the range (with inclusive end) and optional stride value
+    (default is 1).
+
+    Attributes
+    ----------
+    start : `int`
+        Start value of a range.
+    stop : `int`
+        End value of a range, inclusive, same or higher than ``start``.
+    stride : `int` or `None`
```
natelust: I'm not sure I agree None is a useful value for stride, especially as the constructor already has a default value. From the perspective of someone reading the code and or documentation, I am not sure how to interpret what None would mean in a stride. I'm not opposed to it if you had something specific in mind, but if so can you document what it would mean to have a None stride?

python/lsst/daf/butler/exprParser/exprTree.py
```diff
@@ -201,6 +201,38 @@ def __str__(self):
...
+        self.stride = stride
+
+    def visit(self, visitor):
+        # Docstring inherited from Node.visit
+        return visitor.visitRangeLiteral(self.start, self.stop, self.stride, self)
+
+    def __str__(self):
+        res = f"{self.start}..{self.stop}"
+        if self.stride is not None:
+            res += f":{self.stride}"
```
natelust: If you keep None as an option maybe write
`res = f"{self.start}..{self.stop}{f':{self.stride}' if self.stride else ''}`
to avoid boiler plate and or string addition

python/lsst/daf/butler/exprParser/parserLex.py
```diff
@@ -38,6 +38,11 @@
 #  Local non-exported definitions --
 # ----------------------------------
 
+_RE_RANGE = re.compile(r"(-?\d+)\s*\.\.\s*(-?\d+)(\s*:\s*(\d+))?")
```
natelust: do we really want to support an arbitrary number of spaces between the elements? (In practice it probably will never be a problem, just trying to cover bases)

python/lsst/daf/butler/exprParser/parserLex.py
```diff
@@ -148,6 +154,20 @@ def t_STRING_LITERAL(self, t):
         t.value = t.value[1:-1]
         return t
 
+    # range literal in format N..M[:S], spaces allowed
+    # docstring must match _RE_RANGE regexp
+    def t_RANGE_LITERAL(self, t):
+        r"""(-?\d+)\s*\.\.\s*(-?\d+)(\s*:\s*(\d+))?
```
natelust: If this must match, would it be a good idea to stick it in with string formatting so that it for sure stays the same?

python/lsst/daf/butler/exprParser/parserLex.py
```diff
@@ -148,6 +154,20 @@ def t_STRING_LITERAL(self, t):
...
+        """
+        match = _RE_RANGE.match(t.value)
+        start = int(match.group(1))
+        stop = int(match.group(2))
+        stride = match.group(4)
```
natelust: I see that this is match group 4, I presume so you can have a group to make optional. Would it make sense to make that a non-capturing group so that these lines could just be 1,2,3?

python/lsst/daf/butler/exprParser/parserYacc.py
```diff
@@ -245,11 +245,22 @@ def p_literal_num(self, p):
...
+
     def p_literal_str(self, p):
         """ literal : STRING_LITERAL
         """
         p[0] = StringLiteral(p[1])
 
+    def p_literal_range(self, p):
+        """ literal : RANGE_LITERAL
+        """
+        p[0] = RangeLiteral(*p[1])
```
natelust: So this may be outside the scope of this ticket as it relates to past work, but magic unpacking of tuples/lists? by indexing makes me nervous for maintainability. Could you at least add either in the doc string or a code comment explaining what the structure of p is expected to be, so if it changes for some reason in the future, someone has some incite into debugging?

python/lsst/daf/butler/exprParser/exprTree.py
natelust: What would it mean if stride was zero? Should we test and disallow it?

python/lsst/daf/butler/sql/queryBuilder.py
```diff
@@ -68,16 +68,24 @@ def __init__(self, queryBuilder):
...
         # Docstring inherited from TreeVisitor.visitStringLiteral
         return literal(value)
 
+    def visitRangeLiteral(self, start, stop, stride, node):
+        # Docstring inherited from TreeVisitor.visitRangeLiteral
+
+        # just return a triple and let parent clauses handle it
+        if stride is None:
+            stride = 1
+        return (start, stop, stride)
```
natelust: If you keep None as an options maybe just
`return (start, stop, stride or 1)`

python/lsst/daf/butler/sql/queryBuilder.py
```diff
@@ -113,10 +121,57 @@ def visitBinaryOp(self, operator, lhs, rhs, node):
...
+        for start, stop, stride in ranges_remain:
+            # X BETWEEN START AND STOP AND MOD(X, STRIDE) = MOD(START, STRIDE)
+            expr = lhs.between(start, stop)
+            if stride != 1:
+                expr = and_(expr, (lhs % stride) == (start % stride))
+            clauses.append(expr)
+
         if not_in:
-            return lhs.notin_(values)
+            clauses = [not_(expr) for expr in clauses]
```
natelust: it seems a shame to need to have a loop over the len(ranges_remain) twice in the case of not_in just to add the not expression. Consider putting the if not_ also in the above loop, something like:
`clauses.append(not_(expr) if not_in else expr)`
Or of course the converse Boolean logic

natelust (approved):
There are a few things that I think need clarified before this gets merged, but I am confident that they can be discussed and hammered out.

python/lsst/daf/butler/exprParser/exprTree.py
andy-slac: Indeed I need to be more consistent about stride=1 vs None. I want to keep None for case when stride was not specified just in case some consumers of the parsed expression want to know exact syntax of the original expression (for whatever reason). Substitution of None to 1 should happen on the consumer side.I'll update code for consistency and also add stride=0 check, that can be done at parsing stage, there is no useful case for stride=0.

python/lsst/daf/butler/exprParser/exprTree.py
andy-slac: I think this is less readable, too much f-string nesting. I'll do something like:

    res = f"{self.start}..{self.stop}" + f":{self.stride}" if self.stride else ""

python/lsst/daf/butler/exprParser/parserLex.py
andy-slac: Well, I want to allow one space as I think some people could find it more readable. From a typical parser perspective one space and more than one space is the same thing, the only difference is between allowing no space or any number of spaces (TABS/newlines).

python/lsst/daf/butler/exprParser/parserLex.py
andy-slac: I really wanted to have this regex in the docstring so I can see it right here. I agree that there should not be a repetition, I'll do something about it.

python/lsst/daf/butler/exprParser/parserLex.py
andy-slac: I should probably switch to named groups, counting parens is indeed tedious.

python/lsst/daf/butler/exprParser/parserYacc.py
andy-slac: I'll add a comment here, but this is really thing that repeats for every method. `p` is a list of parser symbols, symbols correspond to the rule being processed (as it appears in a grammar in docstring). `p[0]` is the result (`literal` in this case) `p[1]` is the value of `RANGE_LITERAL` token. The value of `RANGE_LITERAL` is what lexer assigns to it, in this case I know that lexer assigns three-tuple of (start, stop, stride).
I know this is hard to read but I do not expect many people reading this code, and those who want to understand what's going on should also read PLY documentation first.

python/lsst/daf/butler/sql/queryBuilder.py
andy-slac: I replaced it with

        expr = or_(*clauses)
        if not_in:
            expr = not_(expr)

it make slightly different expression structure but it should be logically equivalent indeed.

natelust: I'm not sure I understand the change you made. The change I was thinking of was having two if not_in blocks, one that is the returns, and the other inside where you are creating the expr for the first time. I think that would produce exactly the same thing you had, but with one less loop over all the strides.

andy-slac: Oh, I see, you do not like two loops there? I guess this is how I first thought of it and it ended up in implementation. I don't see big issue with iterating twice over ranges, should not be any overhead compared to constructing all expressions. Still I can merge them into this if you prefer:

        literals, ranges = [], []
        for item in values:
            if isinstance(item, tuple):
                ranges.append(item)
            else:
                literals.append(item)

        clauses = []
        for start, stop, stride in ranges:
            count = (stop - start + 1) // stride
            if len(literals) + count > max_in_items:
                # X BETWEEN START AND STOP AND MOD(X, STRIDE) = MOD(START, STRIDE)
                expr = lhs.between(start, stop)
                if stride != 1:
                    expr = and_(expr, (lhs % stride) == (start % stride))
                clauses.append(expr)
            else:
                # add all values to literal list, stop is inclusive
                literals += [literal(value) for value in range(start, stop+1, stride)]
        
        if literals:
            # add IN() in front of BETWEENs
            clauses.insert(0, lhs.in_(literals))

        expr = or_(*clauses)
        if not_in:
            expr = not_(expr)

        return expr

I prefer to handle `not_in` in one place, makes it clearer what is going on (at least for myself)

andy-slac:
Committed all fixes and rebased to latest master.

andy-slac:
@natelust, I think I answered all your concern (hopefully), is it OK to merge it?

natelust (commented):
My only remaining concern that I may have just missed you address is a check if someone puts a zero for a stride

andy-slac:
That is done now at parser/lexer level: https://github.com/lsst/daf_butler/blob/8644b458ead164428b13fdc8c73b0338bd821c9e/python/lsst/daf/butler/exprParser/parserLex.py#L41
Regex will only match non-zero stride and will give a syntax error for 0 or negative number. I tried to find a better way to produce reasonable message at parser level but it gets complicated. Alternatively I could allow 0 in lexer and move this check to queryBuilder but that is also non-ideal as it does not have enough context to point to exact location of a region in expression (and I prefer not to delay handling of errors until very late). Error reporting from parser may not be perfect now but we can look at it later and try to improve it if people complain about it.

natelust:
Yep I did miss that then, I was not expecting it handled that way but I really like it. Good to go

PR #184, opened 2019-08-19
DM 21015: Fix AWS mock credentials.

DinoBektesevic:
When trying to run the S3 datastore tests, with `boto3` and `moto` installed, they fail unless some values are present in a `~/.aws/credentials` file. 

Tests now check whether the file or if the environmental variables are set and if not sets them. The values are unset at the end of the tests.

### DIFF
--- a/python/lsst/daf/butler/core/s3utils.py
+++ b/python/lsst/daf/butler/core/s3utils.py
@@ -19,7 +19,10 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-__all__ = ("s3CheckFileExists", "bucketExists")
+__all__ = ("s3CheckFileExists", "bucketExists", "setAwsEnvCredentials",
+           "unsetAwsEnvCredentials")
+
+import os
 
 try:
     import boto3
@@ -126,3 +129,38 @@ def bucketExists(bucketName, client=None):
         return True
     except s3.exceptions.NoSuchBucket:
         return False
+
+
+def setAwsEnvCredentials(accessKeyId='dummyAccessKeyId', secretAccessKey="dummySecretAccessKey"):
+    """Set AWS credentials environmental variables AWS_ACCESS_KEY_ID and
+    AWS_SECRET_ACCESS_KEY.
+
+    Parameters
+    ----------
+    accessKeyId : `str`
+        Value given to AWS_ACCESS_KEY_ID environmental variable. Defaults to
+        'dummyAccessKeyId'
+    secretAccessKey : `str`
+        Value given to AWS_SECRET_ACCESS_KEY environmental variable. Defaults
+        to 'dummySecretAccessKey'
+
+    Returns
+    -------
+    setEnvCredentials : `bool`
+        True when environmental variables were set, False otherwise.
+    """
+    if "AWS_ACCESS_KEY_ID" not in os.environ or "AWS_SECRET_ACCESS_KEY" not in os.environ:
+        os.environ["AWS_ACCESS_KEY_ID"] = accessKeyId
+        os.environ["AWS_SECRET_ACCESS_KEY"] = secretAccessKey
+        return True
+    return False
+
+
+def unsetAwsEnvCredentials():
+    """Unsets AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environmental
+    variables.
+    """
+    if "AWS_ACCESS_KEY_ID" in os.environ:
+        del os.environ["AWS_ACCESS_KEY_ID"]
+    if "AWS_SECRET_ACCESS_KEY" in os.environ:
+        del os.environ["AWS_SECRET_ACCESS_KEY"]
--- a/python/lsst/daf/butler/core/s3utils.py
+++ b/python/lsst/daf/butler/core/s3utils.py
@@ -148,6 +148,11 @@ def setAwsEnvCredentials(accessKeyId='dummyAccessKeyId', secretAccessKey="dummyS
     -------
     setEnvCredentials : `bool`
         True when environmental variables were set, False otherwise.
+
+    Notes
+    -----
+    If either AWS_ACCESS_KEY_ID or AWS_SECRET_ACCESS_KEY are not set, both
+    values are overwritten.
     """
     if "AWS_ACCESS_KEY_ID" not in os.environ or "AWS_SECRET_ACCESS_KEY" not in os.environ:
         os.environ["AWS_ACCESS_KEY_ID"] = accessKeyId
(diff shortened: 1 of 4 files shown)


tests/test_butler.py
```diff
@@ -619,6 +619,13 @@ def setUp(self):
         uri = ButlerURI(config[".datastore.datastore.root"])
         self.bucketName = uri.netloc
 
+        # set up some fake credentials if they do not exist
+        if not os.path.exists("~/.aws/credentials"):
```
timj: Is this because sometimes you want to be able to use real credentials in the test?

tests/test_butler.py
```diff
@@ -619,6 +619,13 @@ def setUp(self):
...
+            if "AWS_ACCESS_KEY_ID" not in os.environ:
```
timj: Similarly here? You want to allow people to use their own credentials? Meaning that somehow these tests can run without a mock?

tests/test_butler.py
```diff
@@ -619,6 +619,13 @@ def setUp(self):
...
+                os.environ["AWS_ACCESS_KEY_ID"] = "dummyAccessKeyId"
+            if "AWS_SECRET_ACCESS_KEY" not in os.environ:
```
timj: If only one of these environment variables has been set is that a reasonable scenario? Does it make sense to have one be real and one be dummy?  If it's all or nothing then I suggest you check for existence of both env vars and don't set dummy values if both are already set.

tests/test_butler.py
```diff
@@ -649,6 +656,12 @@ def tearDown(self):
         bucket = s3.Bucket(self.bucketName)
         bucket.delete()
 
+        # unset any potentially set dummy credentials
+        keys = ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"]
+        for key in keys:
+            if key in os.environ and "dummy" in os.environ[key]:
```
timj: I'm not a fan of guess work. I'd prefer the first setUp call to have set a property to indicate that dummy credentials are being used and then here do something like:
```python
if self.usedDummyCredentials:
    for key in ("AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"):
        del os.environ[key]
```

tests/test_butler.py
```diff
@@ -649,6 +656,12 @@ def tearDown(self):
         bucket = s3.Bucket(self.bucketName)
         bucket.delete()
 
+        # unset any potentially set dummy credentials
+        keys = ["AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"]
```
timj: No need for this to be a list. Use a tuple.

timj (commented):
I have two main comments:

1. It looks to me like this entire test case is mocked out so why all the work to pass through real credentials?
2. If we are trying to pass through real credentials then there is too much duplicated code here that should be moved to a single location: s3utils seems plausible.

tests/test_butler.py
DinoBektesevic: Yes. But I guess if I comment out mock decorators and set the env vars myself is also good enough. Removed.

tests/test_butler.py
DinoBektesevic: I don't believe so. Conditional changed.

tests/test_butler.py
DinoBektesevic: `self.usedDummyCredentials` added.

tests/test_butler.py
DinoBektesevic: For my personal future reference, why does it make a difference here? I understand tuples are immutable, but this is digested immediately.

timj: I think it's always better to be explicit about mutability. You are providing a fixed list of environment variables that will never change so use a tuple. It's no extra typing to do so.

tests/test_butler.py
```diff
@@ -619,6 +619,13 @@ def setUp(self):
         uri = ButlerURI(config[".datastore.datastore.root"])
         self.bucketName = uri.netloc
 
+        # set up some fake credentials if they do not exist
+        self.usedDummyCredentials = False
```
timj: Please move this code to s3utils and do something like:
```python
self.usedDummyCredentials = setTestCredentials()
```
then in `tearDown`:
```python
if self.usedDummyCredentials:
    clearTestCredentials()
```
where `clearTestCredentials` and `setTestCredentials` are in s3utils. Granted these are only used for tests so you could have a local `s3testUtils.py` in the `tests/` directory instead if you prefer.

timj: Actually, `clearTestCredentials` should probably be called `clearEnvCredentials` since it ensures that there are no environment variable credentials and isn't specific to testing.

tests/test_butler.py
DinoBektesevic: This was changed yesterday already, but is also mute point with the newest commit. Will keep that in mind for the future however.

tests/test_butler.py
DinoBektesevic: Moved setting and unsetting of AWS credentials as environmental variables into s3utils.

python/lsst/daf/butler/core/s3utils.py
```diff
@@ -126,3 +129,38 @@ def bucketExists(bucketName, client=None):
...
+    secretAccessKey : `str`
+        Value given to AWS_SECRET_ACCESS_KEY environmental variable. Defaults
+        to 'dummySecretAccessKey'
+
+    Returns
+    -------
+    setEnvCredentials : `bool`
+        True when environmental variables were set, False otherwise.
+    """
+    if "AWS_ACCESS_KEY_ID" not in os.environ or "AWS_SECRET_ACCESS_KEY" not in os.environ:
```
timj, line 157: Maybe add a Notes section to the docs saying that you overwrite both values if any one value is not set.

tests/test_butler.py
```diff
@@ -649,6 +654,9 @@ def tearDown(self):
         bucket = s3.Bucket(self.bucketName)
         bucket.delete()
 
+        # unset any potentially set dummy credentials
+        unsetAwsEnvCredentials()
```
timj: I think you want an `if self.usingDummyCredentials` statement here (like you have done in the other tests).

timj (approved):
Looks good. One minor clarification and one tweak needed before merging.

python/lsst/daf/butler/core/s3utils.py
DinoBektesevic: Added Notes section.

tests/test_butler.py
DinoBektesevic: You are right, my bad. Thanks for the quick review.

PR #186, opened 2019-08-20
DM-21024: Edit DbAuth docstring syntax

jonathansick:
This follows on from changes in #180. I've checked that the docs successfully build locally.

- The change in "See also" is necessary since otherwise it generates a breaking error in the Sphinx build.
- The change from quotes to code literals helps avoid issues with accidentally starting an inline emphasis using the glob "*". Escaping the asterisk could also have worked.


ktlim (approved):
Sorry about that.  Looks good.

PR #187, opened 2019-08-21
DM-21009: Refactor Posix and S3 datastores to reduce code duplication

timj:
Also uses dataclasses rather than named tuple for database dict.

### DIFF
--- a/python/lsst/daf/butler/core/databaseDict.py
+++ b/python/lsst/daf/butler/core/databaseDict.py
@@ -19,16 +19,53 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-__all__ = ("DatabaseDict",)
+__all__ = ("DatabaseDict", "DatabaseDictRecordBase")
 
-from collections.abc import MutableMapping
+from dataclasses import fields, asdict
+from collections.abc import MutableMapping, Sequence
 
 from lsst.utils import doImport
 
 
+class DatabaseDictRecordBase(Sequence):
+    """A base class to be used to define a record to be stored in a
+    `DatabaseDict`.
+
+    Expected to be subclassed with a dataclass defining the fields and
+    types to be stored in the registry, along with a specification of
+    lengths of string fields.
+    """
+    # make a dataclass named tuple-like by accessing fields by index
+    def __getitem__(self, i):
+        return getattr(self, fields(self)[i].name)
+
+    def __len__(self):
+        return len(fields(self))
+
+    lengths = {}
+    """Lengths of string fields (optional)."""
+
+    key_type = int
+    """Type to use for key field."""
+
+    @classmethod
+    def fields(cls):
+        """Emulate the namedtuple._fields class attribute"""
+        return tuple(f.name for f in fields(cls))
+
+    @classmethod
+    def types(cls):
+        """Return a dict indexed by name and with the python type as the
+        value."""
+        return {f.name: f.type for f in fields(cls)}
+
+    def _asdict(self):
+        return asdict(self)
+
+
 class DatabaseDict(MutableMapping):
     """An abstract base class for dict-like objects with a specific key type
-    and namedtuple values, backed by a database.
+    and data class values, backed by a database.
 
     DatabaseDict subclasses must implement the abstract ``__getitem__``,
     ``__setitem__``, ``__delitem__`, ``__iter__``, and ``__len__`` abstract
@@ -43,27 +80,21 @@ class DatabaseDict(MutableMapping):
     ----------
     config : `Config`
         Configuration used to identify and construct a subclass.
-    types : `dict`
-        A dictionary mapping `str` field names to type objects, containing
-        all fields to be held in the database.
     key : `str`
         The name of the field to be used as the dictionary key.  Must not be
         present in ``value._fields``.
     value : `type`
         The type used for the dictionary's values, typically a
-        `namedtuple`. Must have a ``_fields`` class attribute that is a
-        tuple of field names (i.e., as defined by
-        `~collections.namedtuple`); these field names must also appear
-        in the ``types`` arg, and a ``_make`` attribute to construct it
-        from a sequence of values (again, as defined by
-        `~collections.namedtuple`).
-    lengths : `dict`, optional
-        Specific lengths of string fields.  Defaults will be used if not
-        specified.
+        `DatabaseDictRecordBase`.  Must have a ``fields`` class method
+        that is a tuple of field names; these field names must also appear
+        in the return value of the ``types()`` class method, and it must be
+        possible to construct it from a sequence of values. Lengths of string
+        fields must be obtainable as a `dict` from using the ``lengths``
+        property.
     """
 
     @staticmethod
-    def fromConfig(config, types, key, value, lengths=None, registry=None):
+    def fromConfig(config, key, value, registry=None):
         """Create a `DatabaseDict` subclass instance from `config`.
 
         If ``config`` contains a class ``cls`` key, this will be assumed to
@@ -76,23 +107,17 @@ class DatabaseDict(MutableMapping):
         ----------
         config : `Config`
             Configuration used to identify and construct a subclass.
-        types : `dict`
-            A dictionary mapping `str` field names to Python type objects,
-            containing all fields to be held in the database.
         key : `str`
             The name of the field to be used as the dictionary key.  Must not
-            be present in ``value._fields``.
+            be present in ``value.fields()``.
         value : `type`
             The type used for the dictionary's values, typically a
-            `namedtuple`. Must have a ``_fields`` class attribute that is a
-            tuple of field names (i.e., as defined by
-            `~collections.namedtuple`); these field names must also appear
-            in the ``types`` arg, and a ``_make`` attribute to construct it
-            from a sequence of values (again, as defined by
-            `~collections.namedtuple`).
-        lengths : `dict`, optional
-            Specific lengths of string fields.  Defaults will be used if not
-            specified.
+            `DatabaseDictRecordBase`.  Must have a ``fields`` class method
+            that is a tuple of field names; these field names must also appear
+            in the return value of the ``types()`` class method, and it must be
+            possible to construct it from a sequence of values. Lengths of
+            string fields must be obtainable as a `dict` from using the
+            ``lengths`` property.
         registry : `Registry`, optional
             A registry instance from which a `DatabaseDict` subclass can be
             obtained.  Ignored if ``config["cls"]`` exists; may be None if
@@ -105,14 +130,14 @@ class DatabaseDict(MutableMapping):
         """
         if "cls" in config:
             cls = doImport(config["cls"])
-            return cls(config=config, types=types, key=key, value=value, lengths=lengths)
+            return cls(config=config, key=key, value=value)
         else:
             table = config["table"]
             if registry is None:
                 raise ValueError("Either config['cls'] or registry must be provided.")
-            return registry.makeDatabaseDict(table, types=types, key=key, value=value, lengths=lengths)
+            return registry.makeDatabaseDict(table, key=key, value=value)
 
-    def __init__(self, config, types, key, value, lengths=None):
+    def __init__(self, config, key, value):
         # This constructor is currently defined just to clearly document the
         # interface subclasses should conform to.
         pass
--- a/python/lsst/daf/butler/core/databaseDict.py
+++ b/python/lsst/daf/butler/core/databaseDict.py
@@ -21,10 +21,9 @@
 
 __all__ = ("DatabaseDict", "DatabaseDictRecordBase")
 
-import typing
 from dataclasses import fields, asdict
-from collections.abc import MutableMapping, Sequence
-from typing import Dict, Type, Any, ClassVar, Optional
+from collections.abc import MutableMapping
+from typing import Dict, Type, Any, ClassVar, Optional, Sequence
 
 from lsst.utils import doImport
 from .config import Config
@@ -38,7 +37,12 @@ class DatabaseDictRecordBase(Sequence):
     Expected to be subclassed with a dataclass defining the fields and
     types to be stored in the registry, along with a specification of
     lengths of string fields.
+
+    The fields themselves can be retrieved by index and the number
+    of fields defined in the class can be queried.
     """
+    __slots__ = ()
+
     # make a dataclass named tuple-like by accessing fields by index
     def __getitem__(self, i):
         return getattr(self, fields(self)[i].name)
@@ -53,7 +57,7 @@ class DatabaseDictRecordBase(Sequence):
     """Type to use for key field."""
 
     @classmethod
-    def fields(cls) -> typing.Sequence[str]:
+    def fields(cls) -> Sequence[str]:
         """Emulate the namedtuple._fields class attribute"""
         return tuple(f.name for f in fields(cls))
 
--- a/python/lsst/daf/butler/core/databaseDict.py
+++ b/python/lsst/daf/butler/core/databaseDict.py
@@ -21,10 +21,14 @@
 
 __all__ = ("DatabaseDict", "DatabaseDictRecordBase")
 
+import typing
 from dataclasses import fields, asdict
 from collections.abc import MutableMapping, Sequence
+from typing import Dict, Type, Any, ClassVar, Optional
 
 from lsst.utils import doImport
+from .config import Config
+from .registry import Registry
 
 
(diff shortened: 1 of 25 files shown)


python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -0,0 +1,565 @@
...
+        Factory for creating locations relative to the datastore
+        root.
+    formatterFactory : `FormatterFactory`
+        Factory for creating instances of formatters.
+    storageClassFactory : `StorageClassFactory`
+        Factory for creating storage class instances from name.
+    templates : `FileTemplates`
+        File templates that can be used by this `Datastore`.
+    name : `str`
+        Label associated with this Datastore.
```
TallJimbo: I didn't think Sphinx actually knew how to handle an `Attributes` section like this.  I've (recently) been adding class attributes (with type annotations with no default values) that I expect to shadow with instance attributes, just for the purposes of hanging docs off them (e.g. https://github.com/lsst/daf_butler/blob/2dcc2143d8cb69677398048342e1a046ab0bd77d/python/lsst/daf/butler/core/dimensions/elements.py#L212).  I think that's what @jonathansick recommended to me at one point, and now that I've gotten in the habit of doing it with type annotations (so that the attributes of regular classes are documented the same way as the attributes of dataclasses), it's really growing on me.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -0,0 +1,565 @@
...
+
+    Parameters
+    ----------
+    config : `DatastoreConfig` or `str`
+        Configuration.
```
TallJimbo: Where a `str` means the path to some configuration?

TallJimbo: You're also missing some parameters that are present in the actual `__init__` signature.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -0,0 +1,565 @@
...
+            If `False`, do not modify a value in ``config`` if the value
+            already exists.  Default is always to overwrite with the provided
+            ``root``.
+
+        Notes
+        -----
+        If a keyword is explicitly defined in the supplied ``config`` it
+        will not be overridden by this method if ``overwrite`` is `False`.
+        This allows explicit values set in external configs to be retained.
+        """
```
TallJimbo, line 169: Should/could this docstring be inherited from the base `Datastore` instead?

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -0,0 +1,565 @@
...
+            The `Formatter` to use to deserialize the dataset.
+        info : `StoredFileInfo`
+            Stored information about this file and its formatter.
+        assemblerParams : `dict`
+            Parameters to use for post-processing the retrieved
+            dataset.
+        component : `str`
+            The component to be retrieved (can be `None`).
+        readStorageClass : `StorageClass`
+            The `StorageClass` of the dataset being read.
```
TallJimbo: This is a lot of items to return in a tuple. A `dict`, a `SimpleNamespace`, or a new `dataclass` instance might be better, at least if this is going to be called in more than a couple of places.

TallJimbo (approved):
Looks great; only minor comments.

I wish I could knock out refactors this fast  :slightly_smiling_face: .

python/lsst/daf/butler/datastores/fileLikeDatastore.py
timj: This is not new code as such because it all came from PosixDatastore. I'm happy to tweak it though.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
timj: I was torn. They grew organically during the refactor and of course it's a bit of a red flag when you have to split the line that is calling this method before the equals sign because there are so many variables coming back. I was shying away from a dataclass or some struct thing because it seemed like a lot of overhead for what is an internal method, but I'll take another look at it.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
timj: Yes. Probably. I'm not sure sphinx is picking those up yet. Again this code is a move and not new.

timj:
@TallJimbo I've done your fixes which led to me also refactoring in memory datastore and adding a bunch of type annotations. I think these sufficiently to warrant a quick bonus review if you wouldn't mind.

python/lsst/daf/butler/core/datastore.py
```diff
@@ -47,6 +48,14 @@ class DatastoreValidationError(ValidationError):
     pass
 
 
+@dataclass(frozen=True)
+class Event:
+    name: str
+    undoFunc: typing.Callable
+    args: tuple
+    kwargs: dict
```
TallJimbo, line 60: I'd recommend using `__slots__` here, because there may be a lot of these.  (It's a real shame there's no way for `dataclass` to do that for you, but I've looked into it - apparently the decorator fires too late for anything it does in that department to matter).

python/lsst/daf/butler/core/datastore.py
```diff
@@ -63,19 +65,17 @@ class DatastoreTransaction:
...
-
-    Attributes
-    ----------
-    parent : `DatastoreTransaction`
-        The parent transaction.
     """
-    Event = Event
+    Event: ClassVar[Type] = Event
+
+    parent: Optional['DatastoreTransaction']
```
TallJimbo, line 73: At the risk of sending you down another rabbit hole, if you do `from __future__ import annotations`, you don't need to use a string here.  As a heads-up, if you do that, you'll need to bump the Travis Python version to 3.7.

python/lsst/daf/butler/core/datastore.py
```diff
@@ -63,19 +65,17 @@ class DatastoreTransaction:
...
+    """The parent transaction. (`DatastoreTransaction`, optional)"""
 
     def __init__(self, parent=None):
         self.parent = parent
         self._log = []
 
-    def registerUndo(self, name, undoFunc, *args, **kwargs):
+    def registerUndo(self, name: str, undoFunc: Callable, *args: Any, **kwargs: Any) -> None:
```
TallJimbo, line 80: I have been leaving off the return type annotation for functions that always return `None`, but I don't have a strong opinion on that.  Should we establish a convention one way or the other?

python/lsst/daf/butler/core/datastore.py
```diff
@@ -92,7 +92,7 @@ def registerUndo(self, name, undoFunc, *args, **kwargs):
         self._log.append(self.Event(name, undoFunc, args, kwargs))
 
     @contextlib.contextmanager
-    def undoWith(self, name, undoFunc, *args, **kwargs):
+    def undoWith(self, name: str, undoFunc: Callable, *args: Any, **kwargs: Any) -> Generator:
```
TallJimbo, line 97: Hmm.  I think this is technically correct, but I've been using the return type of the decorated function (i.e. `typing.ContextManager`) in this case, because I find it more useful for human readers of the code.  But I suppose if we ever did get to the point of running mypy, it'd have to be like this, i.e. the return type of the pre-decorated function?

python/lsst/daf/butler/datastores/inMemoryDatastore.py
```diff
@@ -91,6 +87,15 @@ class InMemoryDatastore(Datastore):
     """A new datastore is created every time and datasets disappear when
     the process shuts down."""
 
+    storageClassFactory: StorageClassFactory
+    """Factory for creating storage class instances from name."""
+
+    datasets: Dict[int, object]
```
TallJimbo: You've used `Any` instead of `object` elsewhere (as have I), though I think they are indeed equivalent here.

python/lsst/daf/butler/core/databaseDict.py
```diff
@@ -42,24 +46,24 @@ def __getitem__(self, i):
...
+    lengths: ClassVar[Dict[str, int]] = {}
     """Lengths of string fields (optional)."""
 
-    key_type = int
+    key_type: ClassVar[Type] = int
     """Type to use for key field."""
 
     @classmethod
-    def fields(cls):
+    def fields(cls) -> typing.Sequence[str]:
```
TallJimbo: -> just `Sequence`, since it's imported directly?

python/lsst/daf/butler/core/formatter.py
```diff
@@ -289,7 +288,7 @@ def getLookupKeys(self):
         """
         return self._mappingFactory.getLookupKeys()
 
-    def getFormatterClassWithMatch(self, entity):
+    def getFormatterClassWithMatch(self, entity) -> Tuple[LookupKey, Type]:
```
TallJimbo: A `typing.Union`-based alias for the `entity` arguments would be pretty nice, I think.

python/lsst/daf/butler/core/formatter.py
```diff
@@ -397,7 +396,7 @@ def getFormatter(self, entity, *args, **kwargs):
         _, formatter = self.getFormatterWithMatch(entity, *args, **kwargs)
         return formatter
 
-    def registerFormatter(self, type_, formatter):
+    def registerFormatter(self, type_, formatter) -> None:
```
TallJimbo: Forget the annotations on the arguments?

python/lsst/daf/butler/core/storedFileInfo.py
```diff
@@ -19,87 +19,51 @@
...
-        used to read the file if a component is being requested from
-        a concrete composite.
-
-    See Also
-    --------
-    StorageInfo
+    This is an empty base class.  Datastore implementations are expected to
+    write their own subclasses.
     """
+    pass
```
TallJimbo: Might want to do `__slots__ = ()` here so derived class that have a reason to be compact can be.

python/lsst/daf/butler/core/storedFileInfo.py
```diff
@@ -19,87 +19,51 @@
...
+    def __post_init__(self):
+        # This modification prevents the data class from being frozen.
+        if isinstance(self.formatter, str):
+            # We trust that this string refers to a Formatter
+            pass
+        elif isinstance(self.formatter, Formatter) or \
+                (inspect.isclass(self.formatter) and issubclass(self.formatter, Formatter)):
+            self.formatter = self.formatter.name()
+        else:
+            raise ValueError(f"Supplied formatter '{self.formatter}' is not a Formatter")
```
TallJimbo: `TypeError` instead?

python/lsst/daf/butler/core/storedFileInfo.py
```diff
@@ -19,87 +19,51 @@
...
-    def __repr__(self):
-        return f'{type(self).__qualname__}(path="{self.path}", formatter="{self.formatter}"' \
-            f' size={self.size}, checksum="{self.checksum}", storageClass="{self.storageClass.name}")'
+    def __post_init__(self):
+        # This modification prevents the data class from being frozen.
+        if isinstance(self.formatter, str):
+            # We trust that this string refers to a Formatter
+            pass
+        elif isinstance(self.formatter, Formatter) or \
+                (inspect.isclass(self.formatter) and issubclass(self.formatter, Formatter)):
```
TallJimbo, line 68: Why do we need the `inspect` check?  Does `issubclass` raise instead of returning `False` when given non-types?

python/lsst/daf/butler/core/storedFileInfo.py
```diff
@@ -19,87 +19,51 @@
...
-        """
-        return self._size
+    file_size: int
+    """Size of the serialized dataset in bytes."""
 
-    def __repr__(self):
-        return f'{type(self).__qualname__}(path="{self.path}", formatter="{self.formatter}"' \
-            f' size={self.size}, checksum="{self.checksum}", storageClass="{self.storageClass.name}")'
+    def __post_init__(self):
+        # This modification prevents the data class from being frozen.
```
TallJimbo, line 63: You may be interested in https://github.com/lsst/daf_butler/commit/19981afea055fc72ffefa1c985f6be0d0f913a6f (feel free to cherry-pick to this branch if you are).

python/lsst/daf/butler/datastores/genericDatastore.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+class GenericBaseDatastore(Datastore):
+    """Methods useful for most implementations of a `Datastore`.
+
+    Should always be sub-classed since key abstract methods are missing.
+    """
+
+    records: MutableMapping
+    """Place to store internal records about datasets."""
```
TallJimbo, line 41: If this is only used in the default implementation of `removeStoredItemInfo`, I think it might be cleaner overall to leave that abstract (like `addStoredItemInfo`) and let subclasses define `records` or not; it might avoid some code duplication today, but I don't think `DatabaseDict` will survive adding vectorized inserts of datasets, and that I think that means we shouldn't expect the records for `InMemoryDatastore` to continue to look like the records for other `Datastores`.

TallJimbo (approved):
A few more comments, once again, none of them major.

We should talk sometime about the future of `DatabaseDict` and the relationship between `Registry` and `Datastore`.  I don't really have a clear plan yet, but I think some fairly significant changes will be needed to add vectorized inserts.

python/lsst/daf/butler/core/datastore.py
timj: It had not even crossed my mind that dataclasses did not use slots. I would have put money on them being specifically designed to be efficient.

I see much discussion at ericvsmith/dataclasses#28

python/lsst/daf/butler/core/datastore.py
timj: I was following the examples in the mypy documentation. I think I would tend towards an explicit `None` indicating that we don't return anything.

python/lsst/daf/butler/core/datastore.py
timj: I used Generator here because I ran mypy and it told me I was doing it wrong.

python/lsst/daf/butler/datastores/inMemoryDatastore.py
timj: I was hoping that `object` would imply "must be an instance of something" but I see that Python thinks that pretty much everything is an object so it doesn't really matter.

python/lsst/daf/butler/core/databaseDict.py
timj: `Sequence` is already `collections.abc.Sequence` so I was assuming that's not the same as `typing.Sequence`.

timj: Do we want to always use the short imports for `typing` and be explicit with `collections.abc` when using inheritance?  I think that might be clearer since type annotations will be used a lot more than `collections.abc` in the code.

python/lsst/daf/butler/core/formatter.py
timj: Yes. I was definitely going to do that at some point but that seemed to be a distraction too far for the moment. 😄

python/lsst/daf/butler/core/storedFileInfo.py
timj: Yes. Amazingly enough `issubclass` raises if it's given anything other than a class, rather than returning `False` (or checking whether the instance is equivalent).

python/lsst/daf/butler/datastores/genericDatastore.py
timj: Are you saying that at some point there is going to be a `Datastore.putMany()` that will process multiple in-memory datasets?

genericDatastore now has implementations for add/get/removeStoredItemInfo so it's not just `add` that is using records in this file. If dict-like record store disappears from FileLikeDatastore then we can move them around at that point I think.

TallJimbo: I don't have a use case for `putMany`, but we definitely have a use case for a vectorized version of `Datastore.ingest`.

python/lsst/daf/butler/core/storedFileInfo.py
timj: I don't think that helps. It is no different in behavior to me doing `@dataclass(frozen=True)` and it still prevents me from updating the `formatter`.

python/lsst/daf/butler/core/databaseDict.py
timj: I've changed it to stop using `collections.abc`

PR #188, opened 2019-08-22
DM-17023: overhaul of Registry dimensions and query systems

TallJimbo:


TallJimbo:
@timj, @andy-slac: this ticket isn't fully ready for review, because I haven't tried to update packages downstream of daf_butler yet (and that's where a lot of our test coverage lives), but I think additional daf_butler changes are likely to be either pure additions (i.e. in gen2convert) or small (I bet I made some small mistakes in the new query logic).  So it may make sense for you to start in on this now.  I was thinking:
 - @timj, could you look at all commits _except_ the big one ("Rewrite dimensions and sql...")?
 - @andy-slac, could you look at that big one?

Both of you may want to glance at the other changes for context, and I'd recommend you both start by taking a look at ``docs/dimensions.rst`` (from the big commit), but hopefully splitting it up like this will still make it a bit easier.

andy-slac:
@TallJimbo, I can't find `dimensions.rst` in commits (but I see a link to it from `index.rst`). Did you forget to add it to git index?

TallJimbo:
That was exactly what happened; fixed.

doc/lsst.daf.butler/dimensions.rst
```diff
@@ -0,0 +1,83 @@
...
+The dimensions schema also has some tables that do not map directly to `Dimension` instances.
+Some of these provide extra metadata fields for combinations of dimensions, and are represented by the `DimensionElement` class in Python (this is also the base class of the `Dimension` class, and provides much of its functionality).
+Others represent the overlaps between spatial dimensions, and are discussed in :ref:`lsst.daf.butler-dimensions_spatial_and_temporal`.
+
+.. _lsst.daf.butler-dimensions_relationships_and_containers:
+
+Dimension Relationships and Containers
+--------------------------------------
+
+Dimensions have can have relationships, and in fact these relationships are used almost exclusively to define relationships between datasets, which have no direct relationships between them.
```
andy-slac: One `have` too many.

doc/lsst.daf.butler/dimensions.rst
```diff
@@ -0,0 +1,83 @@
...
+`DimensionGraph` also guarantees a deterministic and topological sort order for its elements.
+
+Because `Dimension` instances have a `~Dimension.name` attribute, we typically
+use `NamedValueSet` and `NamedKeyDict` as containers when immutability is needed or the guarantees of `DimensionGraph`.
+This allows the string names of dimensions to be used as well in most places where `Dimension` instances are expected.
+
+The complete set of all compatible dimensions is held by a special subclass of `DimensionGraph`, `DimensionUniverse`.
+A dimension universe is constructed from configuration, and is responsible for constructing all `Dimension` and `DimensionElement` instances; within a universe, there is exactly one `Dimension` instance that is always used to represent a particular dimension.
+
+`DimensionUniverse` instances themselves are held in a global map keyed by the version number in the configuration used to construction, so they behave somewhat like singletons.
```
andy-slac: to construction -> for construction?
Can you expand a little bit on version numbers, how are they managed, is there a default or most recent version? Or is this covered somewhere in configuration docs?

doc/lsst.daf.butler/dimensions.rst
```diff
@@ -0,0 +1,83 @@
...
+
+.. _lsst.daf.butler-dimensions_data_ids:
+
+Data IDs
+--------
+
+The most common way butler users encounter dimensions is as the keys in a *data ID*, a dictionary that maps dimensions to their primary key values (see :ref:`lsst.daf.butler-dimensions_data_ids`).
```
andy-slac: This `:ref:` references this same section, is it intended?

doc/lsst.daf.butler/dimensions.rst
```diff
@@ -0,0 +1,83 @@
...
+It is also possible to associate a region or timespan with a combination of dimensions (such as the region for a visit and a detector), by defining a `DimensionElement` for that combination.
+
+One kind of spatial dimension is special: a `SkyPixDimension` represents a complete pixelization of the sky, as defined by an `lsst.sphgeom.Pixelization` object.
+These are typically hierarchical pixelizations, like :py:class:`Hierarchical Triangular Mesh (HTM) <lsst.sphgeom.HtmPixelization>`), :py:class:`Q3C <lsst.sphgeom.Q3cPixelization>`, or HEALPix (which has no `lsst.sphgeom` implementation currently), but a skypix dimension encodes both the pixelization scheme and a level, defining a unique mapping from points on the sky to integer IDs.
+By convention, the name of a skypix dimension starts with a short, lowercase name for the pixelization scheme followed by the integer level (e.g. "htm7").
+
+An efficient database representation of temporal relationships is straightforward: these are overlaps of 1-d intervals, we can use regular (i.e. B-tree) indexes to join directly on overlap expressions of intervals expressed as pairs of columns.
```
andy-slac: In general interval indexing is not a trivial problem (it is the same issue as 2D spatial indexing). Some clever optimizations may be possible in case of non-overlapping intervals but they need special queries.

doc/lsst.daf.butler/dimensions.rst
```diff
@@ -0,0 +1,83 @@
...
+
+ - If dimension ``A`` is a *required* dependency of dimension ``B``, ``A``'s primary key value must be provided in order to uniquely identify ``B``, and ``A`` has a one-to-many relationship with ``B``.
+   For example, the detector dimension has a required dependency on the instrument dimension, and hence one uses both an instrument name and a detector ID (or detector name) to fully identify a detector.
+   When both dimensions are associated with database tables, a required dependency involves having the dependency's primary key fields both foreign key fields and part of a compound primary key in the dependent table.
+
+ - If dimension ``C`` is an *implied* dependency of dimension ``D``, a value for ``D`` implies a value for ``C``, and ``C`` has a one-to-many relationship with ``D``.
+   For example, the visit dimension has an implied dependency on the physical filter dimension, because a visit is observed through exactly one filter and hence each visit ID determines a filter name.
+   When both dimensions are associated with database tables, an implied dependency involves having a foreign key field in the dependent table that is not a primary key in the dependent table.
```
andy-slac: not a primary key -> not a part of a primary key?

andy-slac (commented):
Few small comments for `dimensions.rst`

python/lsst/daf/butler/core/datasets.py
```diff
@@ -380,8 +375,8 @@ def datasetType(self):
 
     @property
     def dataId(self):
-        """A `dict` of `Dimension` link fields that labels the Dataset
-        within a Collection (`dict` or `DataId`).
+        """A mapping of `Dimension` vales that labels the Dataset
```
andy-slac: vales?

python/lsst/daf/butler/core/dimensions/config.py
```diff
@@ -0,0 +1,129 @@
...
+    Parameters
+    ----------
+    config : `Config`
+        The subset of a `DimensionConfig` that corresponds to the "skypix" key.
+
+    Returns
+    -------
+    dimensions: `dict`
+        A dictionary mapping `str` names to partially-constructed
+        `SkyPixDimension` instances; the called (i.e. a `DimensionUniverse`)
```
andy-slac: called -> caller?

python/lsst/daf/butler/core/dimensions/config.py
```diff
@@ -0,0 +1,129 @@
...
+    for name in skyPixNames:
+        subconfig = config[name]
+        pixelizationClass = doImport(subconfig["class"])
+        level = subconfig.get("level", None)
+        if level is not None:
+            pixelization = pixelizationClass(level)
+        else:
+            pixelization = pixelizationClass()
+        dimensions[name] = SkyPixDimension(name, pixelization)
+    return dimensions, dimensions[config["common"]]
```
andy-slac: Maybe add a protection in case `config["common"]` is not in keys (just to generate more user-friendly exception)?

python/lsst/daf/butler/core/dimensions/config.py
```diff
@@ -0,0 +1,129 @@
...
+    ----------
+    config : `Config`
+        The subset of a `DimensionConfig` that corresponds to the "elements"
+        key.
+
+    Returns
+    -------
+    dimensions : `dict`
+        A dictionary mapping `str` names to partially-constructed
+        `DimensionElement` instances; the called (i.e. a `DimensionUniverse`)
```
andy-slac: called

python/lsst/daf/butler/core/dimensions/coordinate.py
```diff
@@ -0,0 +1,260 @@
...
+
+    __slots__ = ("graph",)
+
+    def __new__(cls, graph: DimensionGraph, values: Tuple[Any, ...]):
+        self = super().__new__(cls, graph._requiredIndices, values)
+        self.graph = graph
+        return self
+
+    @staticmethod
+    def standardize(mapping: Optional[Mapping[str, Any]], *,
```
andy-slac: Don't you need default (`None`) value for this parameter if it is optional?

python/lsst/daf/butler/core/dimensions/coordinate.py
```diff
@@ -0,0 +1,260 @@
...
+        try:
+            # Optimized code path for DataCoordinate comparisons.
+            return self.graph == other.graph and self.values() == other.values()
+        except AttributeError:
+            # Also support comparison with informal data ID dictionaries that
+            # map dimension name to value.
+            return self.byName() == other
+
+    def __str__(self):
+        return f"{dict(self.byName())}"
```
andy-slac: `byName()` already makes a dict.

python/lsst/daf/butler/core/dimensions/coordinate.py
```diff
@@ -0,0 +1,260 @@
...
+            Callable that accepts a single `bytes` argument to update
+            the hash; usually the ``update`` method of an instance from
+            the ``hashlib`` module.
+        """
+        for k, v in self.items():
+            update(k.name.encode("utf8"))
+            if isinstance(v, int):
+                update(v.to_bytes(64, "big", signed=False))
+            elif isinstance(v, str):
+                update(v.encode("utf8"))
```
andy-slac, line 204: No other types of dimensions could ever exist besides int and str? Maybe throw an exception for an unexpected type here?

python/lsst/daf/butler/core/dimensions/coordinate.py
```diff
@@ -0,0 +1,260 @@
...
+            `True` if all keys that are in in both ``other`` and ``self``
+            are associated with the same values, and `False` otherwise.
+            `True` if there are no keys in common.
+        """
+        d = getattr(other, "full", other)
+        return all(self[k] == d[k] for k in (self.keys() & d.keys()))
+
+    def subset(self, graph: DimensionGraph) -> DataCoordinate:
+        """Return a new `DataCoordinate` that identifies a subset of the
+        dimensions identified by ``self``.
```
andy-slac: I'm confused with what here identifies what. Could you rephrase?

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -19,283 +19,357 @@
...
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-__all__ = ("DimensionElement", "Dimension", "DimensionJoin")
+from __future__ import annotations
+
+__all__ = ["DimensionElement", "Dimension", "SkyPixDimension"]
+
+from typing import Optional, Iterable, AbstractSet, TYPE_CHECKING
+
+from sqlalchemy import Integer
```
andy-slac, line 28: Hmm, I though we wanted `core` layer to be sqlalchemy-free? Or is it too late for that already?

TallJimbo:
I've just pushed a fixup for the big commit that fixes problems I apparently missed because `scons` wasn't re-running all of the tests, despite how many times I re-ran it.  I'm still not sure what the root cause was, but `scons --clean` fixed it.

Anyhow, I'm happy to rebase and squash if that makes things easier, but I didn't want to do that on a PR that might have people looking at it without asking first.

andy-slac:
OK to rebase/squash, it does not touch anything that I commented on so far.

python/lsst/daf/butler/core/schema.py
```diff
@@ -83,317 +116,341 @@ def process_result_value(self, value, dialect):
...
+            raise SchemaValidationError(f"Both length and nbytes provided for field '{self.name}'.")
+        self.primaryKey = config.get("primaryKey", self.primaryKey)
+        self.autoincrement = config.get("autoinc", self.autoincrement)
+        self.nullable = config.get("nullable", False if self.primaryKey else self.nullable)
+        self.doc = stripIfNotNone(config.get("doc", None))
+        return self
+
+    def getSizedColumnType(self):
+        """Return a sized version of the column type, utilizing either (or
+        neither) of ``self.length`` and ``self.nbytes``.
```
timj, line 207: Docstring missing Returns section.

python/lsst/daf/butler/core/schema.py
```diff
@@ -83,317 +116,341 @@ def process_result_value(self, value, dialect):
...
-
-        Returns
-        -------
-        included : `bool`
-            Whether the table or view should be included in the schema.
+        tableName : `str`
+            Name of the logical table to which this column belongs.
+        schema : `Schema`
+            Object represening the full schema.  May be modified in-place.
         """
```
timj, line 234: Docstring missing Returns section.

python/lsst/daf/butler/core/schema.py
```diff
@@ -83,317 +116,341 @@ def process_result_value(self, value, dialect):
...
+        args = [self.name, self.getSizedColumnType()]
+        if self.autoincrement:
+            # Generate a sequence to use for auto incrementing for databases
+            # that do not support it natively.  This will be ignored by
+            # sqlalchemy for databases that do support it.
+            args.append(Sequence(f"{tableName}_{self.name}_seq", metadata=schema.metadata))
+        return Column(*args, nullable=self.nullable, primary_key=self.primaryKey, comment=self.doc)
+
+
+@dataclass
```
timj, line 244: Can this one be marked frozen?

python/lsst/daf/butler/core/schema.py
```diff
@@ -83,317 +116,341 @@ def process_result_value(self, value, dialect):
...
-
-    def normalizeForeignKeyConstraint(self, constraintDescription):
-        """Convert configuration for a ForeignKeyConstraint to standard form
-        and return the target table.
+        return cls(
+            fields=NamedValueSet(FieldSpec.fromConfig(c) for c in config["columns"]),
+            unique={tuple(u) for u in config.get("unique", ())},
+            foreignKeys=[ForeignKeySpec.fromConfig(c) for c in config.get("foreignKeys", ())],
+            sql=config.get("sql"),
+            doc=stripIfNotNone(config.get("doc", None)),
```
timj: `None` is already the default.

python/lsst/daf/butler/core/schema.py
```diff
@@ -83,317 +116,341 @@ def process_result_value(self, value, dialect):
...
-            - foreign_key, link to other table
-            - length, length of the field
-            - nbytes, length of decoded string (only for ``type=='hash'``)
-            - doc, docstring
+        tableName : `str`
+            Name of the logical table to which this constraint belongs
+            (the table for source columns).
+        schema : `Schema`
+            Object represening the full schema.  May be modified in-place.
         """
```
timj, line 307: No Returns section in docstring.

python/lsst/daf/butler/core/schema.py
```diff
@@ -83,317 +116,341 @@ def process_result_value(self, value, dialect):
...
         ----------
-        constraintDescription : `dict`
-            Description of the ForeignKeyConstraint to be created.
-            Should always contain:
+        tableName : `str`
+            Name of the logical table.
+        schema : `Schema`
+            Object represening the full schema.  Will be modified in-place.
+
+        Notes
```
timj, line 395: Missing a Returns section.

Does this method create an actual table in the database or construct a SQL alchemy representation that later will be used to construct the tables?

python/lsst/daf/butler/core/schema.py
```diff
@@ -83,317 +116,341 @@ def process_result_value(self, value, dialect):
...
+    """
+    def __init__(self, spec=None):
+        if spec is None:
+            config = SchemaConfig()
+            spec = config.toSpec()
+        self.metadata = MetaData()
+        self.views = set()
+        self.tables = dict()
+        for tableName, tableSpec in spec.items():
+            tableSpec.toSqlAlchemy(tableName, self)
```
timj, line 483: `toSqlAlchemy` has return values but we are ignoring them here. Given the docstring for `toSqlAlchemy` this behavior confused me.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -124,8 +124,13 @@ def _createSchema(self, schemaConfig):
         in the database - it is called even when an existing database is used
         in order to construct the SQLAlchemy representation of the expected
         schema.
+
+        Parameters
+        ----------
+        spec : `dict` mapping `str` to `TableSpec`
+            Specification of the logical tables to be created.
         """
```
timj, line 159: Missing Returns docstring

tests/test_schema.py
```diff
@@ -27,7 +27,7 @@
 from sqlalchemy import create_engine, MetaData
 
 from lsst.daf.butler.core.utils import iterable
-from lsst.daf.butler.core.schema import SchemaConfig, Schema, Table, Column, SchemaBuilder
+from lsst.daf.butler.core.schema import SchemaConfig, Schema, Table, Column, VALID_COLUMN_TYPES
```
timj: May want to use non-core imports for the public bits.

tests/test_schema.py
```diff
@@ -112,20 +112,20 @@ def assertForeignKeyConstraints(self, table, constraintsDescription):
...
-            if tgtTable in self.schema.views:
+            table = constraint["table"]
+            source = tuple(sorted(iterable(constraint["source"])))
+            target = tuple(f"{table}.{column}" for column in sorted(iterable(constraint["target"])))
+            if table in self.schema.views:
                 continue
-            self.assertIn(src, tableConstraints)
-            self.assertEqual(tableConstraints[src], tgt)
+            self.assertIn(source, tableConstraints)
+            self.assertEqual(tableConstraints[source], target)
```
timj: subTest? (which you might do later on)

python/lsst/daf/butler/core/utils.py
```diff
@@ -648,3 +648,55 @@ def freeze(self):
         """
         if not isinstance(self._dict, MappingProxyType):
             self._dict = MappingProxyType(self._dict)
+
+
+class IndexedTupleDict(Mapping[K, V]):
+    """An immutable mapping that combines a tuple of values with a (possibly
+    shared) mapping from key to tuple index.
```
timj, line 669: We don't care that there can be more values in the tuple than defined in the mapping? We also don't check that the integer value in Mapping does not exceed bounds of tuple?

timj (approved):
Minor comments on the part I was asked to approve (thankfully not the big bit) 😄

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -19,283 +19,357 @@
...
+        For internal use by `DimensionUniverse` only.
+        """
+        # Attach set self.universe and add self to universe attributes;
+        # let subclasses override which attributes by calling a separate
+        # method.
+        self._attachToUniverse(universe)
+        # Expand direct dependencies into recursive dependencies.
+        expanded = set(self._directDependencyNames)
+        for name in self._directDependencyNames:
+            expanded.update(universe[name]._recursiveDependencyNames)
```
andy-slac, line 121: These three lines look like they want to be a `universe` method.

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -19,283 +19,357 @@
...
+        # Define self.implied, a public, sorted version of
+        # self._impliedDependencyNames.
+        self.implied = NamedValueSet(universe.sorted(self._impliedDependencyNames))
+        self.implied.freeze()
+        # Attach a DimensionGraph that provides the public API for getting
+        # at requirements.  Again delegate to subclasses.
+        self._attachGraph()
+        # Create and attach a DimensionRecord subclass to hold values of this
+        # dimension type.
+        from .records import _subclassDimensionRecord
```
andy-slac: Should this import be at top level or is it cyclic thing too?

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -19,283 +19,357 @@
...
-    r"""A join that relates two or more `Dimension`\s.
-
-    `DimensionJoin`\s usually map to many-to-many join tables or views that
-    relate `Dimension` tables.
-
-    `DimensionJoin` objects are not directly constructable; they can only be
-    obtained from a `DimensionGraph`.
+    metadata: NamedValueSet[FieldSpec]
+    """Additional metadata fields included in this element's table
+    (`NamedValueSet` of `FieldSpec`).
```
andy-slac, line 248: I think we are supposed to write it now as

    `NamedValueSet` [`FieldSpec`]

https://developer.lsst.io/python/numpydoc.html#py-docstring-parameter-types-sequences

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -19,283 +19,357 @@
...
+    RecordClass: type
+    """The `DimensionRecord` subclass used to hold records for this element
+    (`type`).
 
-    @property
-    def rhs(self):
-        r"""The `Dimension`\s on the right hand side of the join
-        (`DimensionSet`).
+    Because `DimensionRecord` subclasses are generated dynamically, this type
+    cannot be imported directly and hence canonly be obtained from this
```
andy-slac, line 256: canonly

python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -19,283 +19,357 @@
...
-        could (at least conceptually) be defined as an aggregate view on the
-        summarized join table. For example, "tract_skypix_join" summarizes
-        "patch_skypix_join", because the set of skypix rows associated with a
-        tract row is just the set of skypix rows associated with all patches
-        associated with that tract.  Or, in SQL:
+@immutable
+class Dimension(DimensionElement):
+    """A named data-organization concept that can be used as a key in a data
+    ID.
+
```
andy-slac, line 275: Can you say couple of words here in which way does it extend  `DimensionElement`?

python/lsst/daf/butler/core/dimensions/graph.py
```diff
@@ -19,493 +19,448 @@
...
+    `DimensionGraph` should be used instead of other collections in any context
+    where a collection of dimensions is required and a `DimensionUniverse` is
+    available.
 
-    def __init__(self, universe, elements, dimensions, joins):
-        self._universe = universe
-        self._elements = elements
-        self._dimensions = dimensions
-        self._joins = joins
+    While `DimensionUniverse` inherits from `DimensionGraph`, it should
```
andy-slac, line 103: Which class `it` refers to?

python/lsst/daf/butler/core/dimensions/graph.py
```diff
@@ -19,493 +19,448 @@
...
+                names: Optional[Iterable[str]] = None,
+                conform: bool = True) -> DimensionGraph:
+        if names is None:
+            if dimensions is None:
+                names = ()
+            else:
+                try:
+                    names = set(dimensions.names)
+                except AttributeError:
+                    names = set(d.name for d in dimensions)
```
andy-slac, line 118: Can this line be used instead of `names = set(dimensions.names)` too? Or is there a case when `dimensions.names` can mean something else than `d.name for d in dimensions`? Parameter description and attribute say it should be iterable so it's a bit confusing to see `dimensions.names` above.

python/lsst/daf/butler/core/dimensions/graph.py
```diff
@@ -19,493 +19,448 @@
...
+        """Construct a new graph containing only dimensions in all of the
+        operands.
         """
-        return self.universe.extract(self._dimensions.union(*others), implied=implied)
+        return self.intersection(other)
 
-    def intersection(self, *others, implied=False):
-        """Return a new graph containing all elements that are in both ``self``
-        and all of the other given graphs.
+    def getSpatial(self, *, independent: bool = True,
```
andy-slac, line 351: `Optional[bool]`?

python/lsst/daf/butler/core/dimensions/packer.py
```diff
@@ -0,0 +1,204 @@
...
+    # present just to hold the docstrings for those instance attributes.
+
+    fixed: ExpandedDataCoordinate
+    """The dimensions provided to to the packer at construction
+    (`ExpandedDataCoordinate`)
+
+    The packed ID values are only unique and reversible with these
+    dimensions held fixed.
+    """
+
```
andy-slac, line 160: No docstring for `self.dimensions`?

python/lsst/daf/butler/core/dimensions/records.py
```diff
@@ -0,0 +1,364 @@
...
+    because it does not have overridable methods.
+
+    Record classes have attributes that correspond exactly to the fields in the
+    related database table, a few additional methods inherited from the
+    `DimensionRecord` base class, and two additional injected attributes:
+
+     - ``definition`` is a class attribute that holds the `DimensionElement`;
+
+     - ``timespan`` is a property that returns a `Timespan`, present only
+       on record classes that correspond to temporal elements.
```
andy-slac, line 95: Does it mean that `timespan` column cannot be in table schema?

python/lsst/daf/butler/core/dimensions/records.py
```diff
@@ -0,0 +1,364 @@
...
+        `NotImplemented` if there multiple spatial dimensions identified; in
+        the future this will be replaced with the intersection.
+    timespan : `Timespan`, optional
+        Timespan associated with this data ID, or `None` if there are no
+        temporal dimensions.
+    conform : `bool`, optional
+        If `True` (default) adapt arguments from arbitrary mappings to
+        `NamedKeyDict` and check that all expected key-value pairs are
+        present.  `False` is primarily for internal use, or at least
+        callers that can independently guarantee correctness.
```
andy-slac: Looks like if `conform` is False then `full`, `region`, and `timespan` are completely ignored?
Also if `conform` is False then `full` and `records` are expected to be `IndexedTupleDict`, not `collections.abc.Mapping`?

python/lsst/daf/butler/core/dimensions/records.py
```diff
@@ -0,0 +1,364 @@
...
+            dimension configuration).
+        returnMaxBits : `bool`, optional
+            If `True` (`False` is default), return the maximum number of
+            nonzero bits in the returned integer across all data IDs.
+
+        Returns
+        -------
+        packed : `int`
+            Integer ID.  This ID is unique only across data IDs that have
+            the same values for the packer's "fixed" dimensions.
```
andy-slac: `Returns` should match `pack()` method, it is missing `maxBits`.

python/lsst/daf/butler/core/dimensions/records.py
```diff
@@ -0,0 +1,364 @@
...
+        )
+
+    def __getnewargs_ex__(self) -> Tuple(tuple, dict):
+        return (
+            (self.graph, self.values()),
+            dict(
+                records=self.records,
+                full=self.full,
+                region=self.region,
+                timespan=self.timespan,
```
andy-slac: Do you also want `conform=False` here?

python/lsst/daf/butler/core/dimensions/records.py
```diff
@@ -0,0 +1,364 @@
...
+            )
+        )
+
+    # Class attributes below are shadowed by instance attributes, and are
+    # present just to hold the docstrings for those instance attributes.
+
+    full: IndexedTupleDict[Dimension, Any]
+    """Dictionary mapping dimensions to their primary key values for all
+    dimensions in the graph, not just required ones (`NamedKeyDict`).
```
andy-slac: NamedKeyDict -> IndexedTupleDict?

python/lsst/daf/butler/core/dimensions/records.py
```diff
@@ -0,0 +1,364 @@
...
+    """
+
+    records: IndexedTupleDict[DimensionElement, DimensionRecord]
+    """Dictionary mapping `DimensionElement` to the associated
+    `DimensionRecord` (`NamedKeyDict`).
```
andy-slac: Same here

python/lsst/daf/butler/core/dimensions/schema.py
```diff
@@ -0,0 +1,195 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = ("makeElementTableSpec", "makeOverlapTableSpec", "Timespan", "REGION_FIELD_SPEC",
```
andy-slac: Should `Timespan` be here? 
`makeForeignKeySpec` and `addDimensionForeignKey` are not exported, are they truly internal?

python/lsst/daf/butler/core/dimensions/storage.py
```diff
@@ -0,0 +1,522 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = ("setupDimensionStorage", "DimensionRecordStorage",
+           "DatabaseDimensionRecordStorage", "CachingDimensionRecordStorage")
```
andy-slac: No ChainedDimensionRecordStorage or SkyPixDimensionRecordStorage?

python/lsst/daf/butler/core/dimensions/storage.py
```diff
@@ -0,0 +1,522 @@
...
+        The logical table for the element.  May be a select query instead
+        of a true table.
+    commonSkyPixOvelapTable : `sqlalchemy.sql.FromClause`, optional
+        The logical table for the overlap table with the dimension universe's
+        common skypix dimension.
+    """
+
+    def __init__(self, connection: Connection, element: DimensionElement, *,
+                 elementTable: FromClause,
+                 commonSkyPixOverlapTable: Optional[FromClause]):
```
andy-slac: Does it need default value as optional?

python/lsst/daf/butler/core/dimensions/storage.py
```diff
@@ -0,0 +1,522 @@
...
+        # Docstring inherited from DimensionRecordStorage.element.
+        return self._element
+
+    def clearCaches(self):
+        # Docstring inherited from DimensionRecordStorage.clearCaches.
+        pass
+
+    def matches(self, dataId: Optional[DataId] = None) -> bool:
+        # Docstring inherited from DimensionRecordStorage.matches.
+        return True
```
andy-slac, line 245: Does it mean that any DataId can match?

python/lsst/daf/butler/core/dimensions/storage.py
```diff
@@ -0,0 +1,522 @@
...
+        self._connection.execute(self._elementTable.insert(), *elementRows)
+        if self.element.spatial and commonSkyPixRows:
+            self._connection.execute(self._commonSkyPixOverlapTable.insert(), *commonSkyPixRows)
+
+    def fetch(self, dataId: DataCoordinate) -> Optional[DimensionRecord]:
+        # Docstring inherited from DimensionRecordStorage.fetch.
+        RecordClass = self.element.RecordClass
+        # I don't know how expensive it is to construct the query below, and
+        # hence how much gain there might be to caching it, so I'm going to
+        # wait for it to appear as a hotspot in a profile before trying that.
```
andy-slac, line 289: I think it's mostly determined by how frequently it's called, if it's somewhere in a tight loop then you'll need to optimize if (or do multi-row select in one query).

python/lsst/daf/butler/core/dimensions/universe.py
```diff
@@ -0,0 +1,237 @@
...
+
+    For internal use only.
+    """
+
+    def __new__(cls, config: Optional[Config] = None) -> DimensionUniverse:
+        # Create the universe instance and add core attributes.
+        # We don't want any of what DimensionGraph.__new__ does, so we just go
+        # straight to object.__new__.  The C++ side of my brain is offended by
+        # this, but I think it's the right approach in Python, where we don't
+        # have the option of having multiple constructors with different roles.
```
andy-slac, line 82: hmm 🤔

python/lsst/daf/butler/core/dimensions/universe.py
```diff
@@ -0,0 +1,237 @@
...
+        for name, subconfig in config.get("packers", {}).items():
+            self._packers[name] = DimensionPackerFactory.fromConfig(universe=self, config=subconfig)
+
+        # Use the version number from the config as a key in the singleton
+        # dict containing all instances; that will let us transfer dimension
+        # objects between processes using pickle without actually going
+        # through real initialization, as long as a universe with the same
+        # version has already been constructed in the receiving process.
+        self._version = config["version"]
+        cls._instances[self._version] = self
```
andy-slac, line 131: Maybe check `_instances` fro existing version number before creating new universe? Though this of course depends on how the whole thing is constructed, maybe it's not needed?

python/lsst/daf/butler/core/dimensions/universe.py
```diff
@@ -0,0 +1,237 @@
...
+        return self
+
+    def __repr__(self) -> str:
+        return f"DimensionUniverse({self})"
+
+    def extract(self, iterable: Iterable[Union[Dimension, str]]) -> DimensionGraph:
+        """Construct a `DimensionGraph` from a possibly-heterogenous iterable
+        of `Dimension` instances and string names thereof.
+        """
```
andy-slac, line 155: Returns and Parameters are missing from docstring

python/lsst/daf/butler/core/queries/builder.py
```diff
@@ -0,0 +1,400 @@
...
+        automatically call it if the `DimensionElement` has been identified as
+        one that must be included.
+
+        Parameters
+        ----------
+        element : `DimensionElement`
+            Element for which the relationship should be added.  The element
+            must be associated with a database table (see
+            `DimensionElement.hasTable`) and must be
+            `~Dimensionelement.spatial`.
```
andy-slac: You can just say "must be spatial" or "must have `~Dimensionelement.spatial` set to `True`"

python/lsst/daf/butler/core/queries/builder.py
```diff
@@ -0,0 +1,400 @@
...
+        assert element in self.summary.spatial,\
+            f"{element}'s spatial information is not relevant for this query."
+        table = self._dimensionStorage[element].getCommonSkyPixOverlapTable(self.summary.dataId)
+        assert table is not None, f"No relationship found from the common skypix to {element}."
+        dimensions = NamedValueSet(element.graph.required)
+        dimensions.add(self.summary.universe.commonSkyPix)
+        self.joinTable(table, dimensions)
+
+    def joinDataset(self, datasetType: DatasetType, collections: CollectionsExpression, *,
+                    isResult: bool = True, addRank: bool = False):
```
andy-slac, line 138: `Optional[bool]` ?

python/lsst/daf/butler/core/queries/builder.py
```diff
@@ -0,0 +1,400 @@
...
+            The type of datasets to search for.
+        collections : sequence of `str` or `Like`, or ``...``
+            An expression describing the collections in which to search for
+            the datasets.  ``...`` indicates that all collections should be
+            searched.
+        isResult : `bool`, optional
+            If `True` (default), include the ``dataset_id`` column in the
+            result columns of the query, allowing complete `DatasetRef`
+            instances to be produced from the query results for this dataset
+            type..  If `False`, the existence of datasets of this type is used
```
andy-slac: Two dots.

andy-slac (commented):
I'm still not finished but my brain is freezing so I just want to checkpoint all my comments so far and finish it later. I hope not too much is left :)

python/lsst/daf/butler/core/queries/structs.py
```diff
@@ -0,0 +1,420 @@
...
+        combinations in any visit that overlaps that tract, not just the
+        visit+detector combinations that directly overlap the tract.
+    """
+    def __init__(self, requested: DimensionGraph, *,
+                 dataId: Optional[ExpandedDataCoordinate] = None,
+                 expression: Optional[Union[str, QueryWhereExpression]] = None,
+                 given: Optional[DimensionGraph] = None,
+                 entire: Optional[NamedValueSet[DimensionElement]] = None):
+        self.requested = requested
+        self.dataId = dataId if dataId is not None else ExpandedDataCoordinate(self.universe.empty, ())
```
andy-slac: `self.universe` depends on `self.requested` being set so there is an implicit initialization ordering requirements here which feels a bit fragile. It may be better not to use `self.universe` in constructor, but be explicit and use `requested.universe` instead. (Another `self.universe` is below)

python/lsst/daf/butler/core/queries/structs.py
```diff
@@ -0,0 +1,420 @@
...
+        # - it isn't also given at query construction or execution time.
+        result = self.requested.getSpatial(prefer=self.entire) - self.given.elements
+        if len(result) == 1:
+            # There's no spatial join, but there might be a WHERE filter based
+            # on a given region.
+            if self.given.spatial:
+                # We can only perform those filters against SkyPix dimensions,
+                # so if what we have isn't one, add the common SkyPix dimension
+                # to the query; the element we have will be joined to that.
+                element, = result
```
andy-slac, line 250: That comma induces momentary coma in reviewer's brain 🙂 Can't you just do `element = result[0]` instead?

python/lsst/daf/butler/core/queries/structs.py
```diff
@@ -0,0 +1,420 @@
...
+                if not isinstance(element, SkyPixDimension):
+                    result.add(self.universe.commonSkyPix)
+            else:
+                # There is no spatial join or filter in this query.  Even
+                # if this element might be associated with spatial
+                # information, we don't need it for this query.
+                return NamedValueSet()
+        elif len(result) > 1:
+            # There's a spatial join.  Those require the common SkyPIx
```
andy-slac: `SkyPIx`

python/lsst/daf/butler/core/queries/structs.py
```diff
@@ -0,0 +1,420 @@
...
+        if len(result) == 1 and not self.given.getTemporal():
+            # No temporal join or filter.  Even if this element might be
+            # associated with temporal information, we don't need it for this
+            # query.
+            return NamedValueSet()
+        return result
+
+    @property
+    def mustHaveKeysJoined(self) -> DimensionGraph:
+        """Dimensions whose primary keys must be included in the FROM clause
```
andy-slac: I'm not sure I understand this, FROM clause includes tables or sub-queries, how do primary keys enter here?

python/lsst/daf/butler/core/queries/query.py
```diff
@@ -0,0 +1,229 @@
...
+
+from .structs import QuerySummary, QueryColumns, QueryParameters
+
+
+class Query:
+    """A wrapper for a SQLAlchemy query that knows how to re-bind parameters
+    and transform result rows into data IDs and dataset references.
+
+    A `Query` should almost always be constructed directly by a call to
+    `QueryBuilder` finish; direct construction will make it difficult to be
```
andy-slac: `QueryBuilder.finish`?

python/lsst/daf/butler/core/queries/query.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def predicate(self, region: Optional[Region] = None) -> Callable[[RowProxy], bool]:
+        """Return a callable that can perform extra Python-side filtering of
+        query results.
+
+        To get the expected results from a query, the returned predicate *must*
+        be used to ignore rows for which it returns `False`; this permits the
+        `QueryBuilder` implementation to move logic from the database to Python
+        without changing the public interface.
+
```
andy-slac, line 89: Maybe add couple of words to explain how region-based filtering works, code below may not be obvious for everyone.
Also do you need to return predicate, can't you just define a method in this class for that? You seem to have other methods that work on RowProxies, why this one is different?

python/lsst/daf/butler/core/registry.py
```diff
@@ -728,378 +713,150 @@ def getRun(self, id=None, collection=None):
...
+        where : `str`, optional
+            A string expression similar to a SQL WHERE clause.  May involve
+            any column of a dimension table or (as a shortcut for the primary
+            key column of a dimension table) dimension name.
+        deduplicate : `bool`, optional
+            If `True` (`False` is default), for each result data ID, only
+            yield one `DatasetRef` of each `DatasetType`, from the first
+            collection in which a dataset of that dataset type appears
+            (according to the order of ``collections`` passed in).  Cannot be
+            used if ``collections``.
```
andy-slac: if collections?

python/lsst/daf/butler/core/queries/expressions.py
```diff
@@ -0,0 +1,298 @@
...
+        if literals:
+            # add IN() in front of BETWEENs
+            clauses.insert(0, lhs.in_(literals))
+
+        expr = or_(*clauses)
+        if not_in:
+            return lhs.notin_(values)
+        else:
+            return lhs.in_(values)
+            expr = not_(expr)
```
andy-slac, line 282: These 5 lines above do not look right, probably due to manual merge with other ticket. I think this should look like:

        expr = or_(*clauses)
        if not_in:
            expr = not_(expr)

Here it is on master: https://github.com/lsst/daf_butler/blob/master/python/lsst/daf/butler/sql/queryBuilder.py#L171-L173

andy-slac (commented):
Sorry, did not manage to complete it today, other things interfered :) Just few comments today, should be able to finish the rest tomorrow.

python/lsst/daf/butler/core/queries/datasets.py
```diff
@@ -0,0 +1,287 @@
...
+                 tables: Mapping[str, FromClause]):
+        self._connection = connection
+        self._universe = universe
+        self._datasetTypeTable = tables["dataset_type"]
+        self._datasetTypeDimensionsTable = tables["dataset_type_dimensions"]
+        self._datasetTable = tables["dataset"]
+        self._datasetCollectionTable = tables["dataset_collection"]
+
+    def fetchDatasetTypes(self, datasetType: DatasetTypeExpression = ..., *,
+                          collections: CollectionsExpression = ...,
```
andy-slac, line 139: Should it be `Optional` too?

python/lsst/daf/butler/core/queries/datasets.py
```diff
@@ -0,0 +1,287 @@
...
+        )
+        whereTerms = []
+        if datasetType is ...:
+            # "..." means no restriction on the dataset types; get all
+            # of them.
+            pass
+        elif isinstance(datasetType, str):
+            whereTerms.append(self._datasetTypeTable.columns.dataset_type_name == datasetType)
+        elif isinstance(datasetType, Like):
+            whereTerms.append(self._datasetTypeTable.columns.dataset_type_name == datasetType)
```
andy-slac: Should it be 

    self._datasetTypeTable.columns.dataset_type_name.like(datasetType.pattern)

instead?

python/lsst/daf/butler/core/queries/datasets.py
```diff
@@ -0,0 +1,287 @@
...
+            dimensionNames.add(dimensionName)
+        return [DatasetType(datasetTypeName,
+                            dimensions=DimensionGraph(self._universe, names=dimensionNames),
+                            storageClass=storageClassName)
+                for datasetTypeName, (storageClassName, dimensionNames) in grouped.items()]
+
+    def getDatasetSubquery(self, datasetType: DatasetType, *,
+                           collections: CollectionsExpression,
+                           dataId: Optional[ExpandedDataCoordinate] = None,
+                           isResult: bool = True,
```
andy-slac, line 214: `Optional[bool]`?

python/lsst/daf/butler/instrument.py
```diff
@@ -108,106 +109,103 @@ def applyConfigOverrides(self, name, config):
...
-
-
-def updateExposureEntryFromObsInfo(dataId, obsInfo):
-    """Construct an exposure Dimension entry from
+        # Docstring inherited from DimensionPacker.unpack
+        return DataCoordinate.standardize(
+            {
+                "instrument": self._instrumentName,
+                "detector": packedId % self._detectorMax,
+                self._observationName: packedId // self._detectorMax
```
andy-slac: I usually prefer `divmod` to two separate expressions.

andy-slac (approved):
I think I looked at all files in this large commit, but only very cursory at tests/*. It looks OK and I'm both impressed and overwhelmed by the scale of it (I thought only Qserv guys would do such huge PRs, apparently I'm mistaken 🙂). Few general observations in addition to the diff comments:
- there is a bunch of imports in the executable code, I guess they are there because of cyclic dependencies, it may be better to try to avoid those if possible
- type annotations almost everywhere. I'm not sure they make things more readable, though maybe I'm just not used to it. General feeling is that trying to make Python to look more like static-typing-language just makes it more confusing. Or maybe we just need better tools that could use those annotations (or hide them).
- ellipsis looks fancy but `None` would probably work as well, I think None as special value is used much more frequently in Python and LSST code than ellipsis.
- more importantly, after seeing all this new code I think I'm sort of lost in it, I see there is a lot of thought went into the design but I'm still missing a big picture in my head. It would be extremely helpful to have some sort of document describing high-level design, maybe with few diagrams.

doc/lsst.daf.butler/dimensions.rst
TallJimbo: I haven't tried at all to come up with conventions or guidelines for the version numbers; all we have right now is code that blindly uses the version provided in the configuration as a key (which happens to be `zero`, since that's what's in the default configs).  And it's fine right now for us to not bother with meaningful version numbers, because we're not promising any kind of backwards compatibility yet, and the only reason we'll need versions (at least in the context described above) is so you can have different `Registry` instances with different dimensions alive in the same Python process, presumably because you're trying to utilize an older repository and a new one together for some reason.

doc/lsst.daf.butler/dimensions.rst
TallJimbo: Good catch - it's a sentence that used to be in the overview section, and doesn't need the no-longer-forward reference anymore.

doc/lsst.daf.butler/dimensions.rst
TallJimbo: I'll add a sentence noting that the approach to interval indexing is still fairly naive.

python/lsst/daf/butler/core/datasets.py
TallJimbo: -> "primary key values"

python/lsst/daf/butler/core/dimensions/records.py
TallJimbo: Correct.  Instead, there are always two columns in the table schema (begin and end) for the timespan.

python/lsst/daf/butler/core/dimensions/records.py
TallJimbo: Actually, `full`, `region`, and `timespan` are ignored if `conform=True`, but I suspect that's what you meant. I'll document that.

And while `IndexedTupleDict` implements `collections.abc.Mapping`, so the types aren't really incorrect, I'll emphasize more strongly that `conform=False` is really for internal use.

python/lsst/daf/butler/core/dimensions/schema.py
TallJimbo: `Timespan` should not be there.  I suppose `makeForeignKeySpec` and `addDimensionForeignKey` might as well be - like all of the others, they are internal to the `dimensions` subpackage but not this module.

python/lsst/daf/butler/core/dimensions/storage.py
TallJimbo: Fixed.

python/lsst/daf/butler/core/dimensions/storage.py
TallJimbo: Formally, no - in type annotations, `Optional[T]` is exactly equivalent to `Union[T, None]`, but it probably does make sense to add the default here anyway.

python/lsst/daf/butler/core/dimensions/storage.py
TallJimbo: Yes.  This is really a hook for future functionality that hasn't been fully implemented yet (just prototyped a bit).

python/lsst/daf/butler/core/dimensions/storage.py
TallJimbo: There's a trade-off here between multi-row select, caching, and fetching duplicate values from the database, so at some level structuring fetch as a single-record operation is a guess that caching and avoiding fetching duplicates is more valuable for performance than multi-row select.  I think there's some indirect evidence for that being true, at least the preflight queries, from the last round of optimizing we did there.

python/lsst/daf/butler/core/dimensions/universe.py
TallJimbo: I thought I had logic to do that (check for existing instance with that version number) already, but I guess it got lost.  I'll add it back in.

python/lsst/daf/butler/core/queries/builder.py
TallJimbo: No, actually - this is an unfortunate difference between `typing.Optional` (which means "may be `None`") and the less-formal usage of `<type>, optional` in docstrings as specified by the dev guide (which means "need not be passed by caller").


This might be something @jonathansick might be interested in thinking about, if he hasn't already, though I don't think there's an easy answer: both meanings are useful and precisely defined, and are arguably each the most useful meaning for their own slightly-different contexts.

python/lsst/daf/butler/core/queries/datasets.py
TallJimbo: (discussed in `builder.py` comment thread - I'll just :-1: future cases)

python/lsst/daf/butler/core/queries/query.py
TallJimbo: :+1: on docs.

By returning a new callable, we can do some work up-front instead of per-row.  We're not really using that much right now, but I'd like to keep the interface this way in case that becomes valuable in the future.

python/lsst/daf/butler/core/queries/structs.py
TallJimbo: I can't do `result[0]` because it's a `set`, not a sequence.  I'm open to other ideas.

python/lsst/daf/butler/core/queries/structs.py
TallJimbo: I guess I mean, "dimensions whose primary keys must be used in `JOIN ON` clauses".  Does that make more sense?

python/lsst/daf/butler/core/registry.py
TallJimbo: Should have been `Cannot be used if any element in ``collections`` is an expression.`

python/lsst/daf/butler/core/dimensions/storage.py
andy-slac: Hmm, I would expect multi-row select (with `multi >> 1`) to produce bigger benefit simply because of fewer round-trips to database. But indeed we need it to be measured first to understand if it's an issue.

python/lsst/daf/butler/core/queries/structs.py
andy-slac: I see. Both `list(result)[0]` and `next(iter(result))` look worse than comma in my opinion. Maybe `[element] = result` or `(element,) = result` to make it more obvious than just comma?

python/lsst/daf/butler/core/queries/structs.py
andy-slac: Yeah, that is probably better.

python/lsst/daf/butler/core/schema.py
TallJimbo: Well, tests pass if I do mark it `frozen`, at least.  I'm not certain it's a good idea here, because I *have* found use cases for mutating `TableSpec` and `FieldSpec` (usually some kind of "make a modified copy" operation), and because the lifetimes and sharing of of these objects is generally limited.

python/lsst/daf/butler/core/schema.py
TallJimbo: Added `Returns` section and clarifying docs: this only constructs the SQLAlchemy representation.

python/lsst/daf/butler/core/schema.py
TallJimbo: On my next big refactor ticket I'd like to just remove this `Schema` class and then start using the return values; now that it isn't responsible for actually translating `Config` to SQLAlchemy anymore, `Schema` is not much more than a dictionary.  I've kept it around for now because switching to a plain `dict` would require more changes to `SqlRegistry` that are otherwise unrelated to this ticket.  In the meantime, I'll add a code comment.

python/lsst/daf/butler/core/utils.py
TallJimbo: We do care, but in all of the use cases for this class it's easy and much more efficient to guarantee those invariants outside of the class.  Sounds like I should document those expectations, and maybe `assert` on them here.

TallJimbo:
I've now rebased on master, fixed some conflicts, and addressed all diff comments (at least initially; I'll go back and do another sweep later for active conversations).  I've also added a few commits to reflect ongoing work updating downstream packages.  I don't think it makes sense for anyone to look again at this yet; I'll be careful to keep "Address review" commits and "updates for downstream packages" separate from each other and from what has already been reviewed, and plan to do squashing only after final review is complete.

Replies to @andy-slac 's more general comments below:
> there is a bunch of imports in the executable code, I guess they are there because of cyclic dependencies, it may be better to try to avoid those if possible

I've managed to untangle a lot of this on 6627d7b10170214f1dbd22d45e2e02c0ea767c66, moving local imports to module scope and moving some module-scope imports into blocks protected by `if TYPE_CHECKING:`, which I've come to think of as type-annotation-python's version to forward declarations.  That also involved moving `ExpandedDataCoordinate` to the `coordinate` module, which is a nice bonus. Many of the circular relationships between classes remain - given the umbrella role of `DimensionUniverse`, I think they're very hard to avoid - but at least the modules not longer have circular dependencies.

> type annotations almost everywhere. I'm not sure they make things more readable, though maybe I'm just not used to it. General feeling is that trying to make Python to look more like static-typing-language just makes it more confusing. Or maybe we just need better tools that could use those annotations (or hide them).

I suspect @timj has some plans for tooling that would use them (I am aware of the existence of such tools, but no more than that).  I actually like them just because I find that they make the code *more* readable - I hate looking at regular Python signatures and not really knowing what I am expected to pass in and what I should expect to get back.  And I find that with the `typing` module you can really declare those things precisely in a way that just isn't natural in documentation.

> ellipsis looks fancy but None would probably work as well, I think None as special value is used much more frequently in Python and LSST code than ellipsis.

The first version of this used `None`, but I ran into two problems:
- in many cases I also wanted `None` to be available to mean "argument not provided; do the default", which is not always the same as "match anything";
- `None` coerces to boolean `False` and more generally conceptually means "nothing"; it seemed confusing it use it as a value that instead means "everything".

If I was syntactically allowed to use `*` that would be even better than `...`.  I think the real choice is between `...` and the special string `"*"`, and I like the idea of being able to use a true language token rather than just a special string.  But I could be convinced otherwise.

> more importantly, after seeing all this new code I think I'm sort of lost in it, I see there is a lot of thought went into the design but I'm still missing a big picture in my head. It would be extremely helpful to have some sort of document describing high-level design, maybe with few diagrams.

I'm sure you're right, but I was hoping the `dimensions.rst` page would play at least some of this role.  Could you say anything more about what it was lacking (not enough detail? Too much focus on usage rather than design? Do I need a similar page for the `queries` subpackage?).

timj:
Regarding type annotations, the hope is to be able to use them instead of explicit types in the numpydoc strings. I believe there is a sphinx extension to do that but I'm not sure if we've turned it on yet. mypy is the tool I've been using that does static type checking: it works okay but some work is needed on dependent packages to make it work accurately.

andy-slac:
I have no better arguments for `None` vs. ellipsis, so I'm OK with yuor preferences.

> I'm sure you're right, but I was hoping the dimensions.rst page would play at least some of this role. Could you say anything more about what it was lacking (not enough detail? Too much focus on usage rather than design? Do I need a similar page for the queries subpackage?).

`dimensions.rst` does explain things about basic concepts and their corresponding classes, what is harder to guess from there is the design of the rest of software responsible for persisting all these things to database. There are many classes and hierarchies of classes involved and it's not always obvious how do they interact with each other. Maybe a separate page with a design description mentioning the roles of each or most important classes could help people to understand it? Ideally some UML diagrams could be added too but even a short description of most common use case would be fine too. Of course if you think that design is not stable yet and things are going to change then it can wait for better time.

TallJimbo:
@andy-slac, could you please look at:

- https://github.com/lsst/daf_butler/pull/188/commits/f4816710abe18f2bdb5132371b1e5005ae3a177f
- https://github.com/lsst/daf_butler/pull/188/commits/b8b6b52ed660f6fcc7709c47dfef862990239c7d
- https://github.com/lsst/daf_butler/pull/188/commits/b81f717733994c7d1af25a100cf2529d036c46a6
- https://github.com/lsst/daf_butler/pull/188/commits/ee3722e9323b4d433e76aff9ed789dcc1ac3720b
 - https://github.com/lsst/daf_butler/pull/188/commits/8bc9dbaf3d37d8befa7880a9235ebc1cfefb33ba
- https://github.com/lsst/daf_butler/pull/188/commits/0331ffc18227faff2c23795c9e487763cacde2dc
- https://github.com/lsst/daf_butler/pull/188/commits/926b08034210e854554bd89c13785ba459a53af8
- https://github.com/lsst/daf_butler/pull/188/commits/8f007f052b0ce7ad6dd3a350b76713513f61562d

@timj, could you please look at:
- https://github.com/lsst/daf_butler/pull/188/commits/dc3ce9a66b8296becfdbee7dbbc996e3d6f1fd64
- https://github.com/lsst/daf_butler/pull/188/commits/6f72d73be87349af5ba567b252bde977d652dd73
- https://github.com/lsst/daf_butler/pull/188/commits/03ad1c5b4fffc0f51fe7f18087bfef466dff1e70
- https://github.com/lsst/daf_butler/pull/188/commits/7b7d22fdc61bc56084573f9b12eb6a8feef2ae02
- https://github.com/lsst/daf_butler/pull/188/commits/75c63f5a292c83a615eabced7aea7a210dffbd71
- https://github.com/lsst/daf_butler/pull/188/commits/d1eed0e3e2d37795159bb7c21bb1c66c3961c966
- https://github.com/lsst/daf_butler/pull/188/commits/526eb93ec410ba4b8be16eeb9f526fbe345fdd21

andy-slac:
I looked at the commits from my list and it looks good to me.

timj (approved):
Looks okay although I have to say that the complexity of the export/import support code is blowing my mind a little. I don't see any tests for any of it in daf_butler though. Is it not possible to export something from the test butler and then import it into a different collection?

TallJimbo:
I've added a simple test for export/import to `test_butler.py`, and removed the `StructuredDataRepo(Import|Export)Backend` classes in favor of just having the YAML implementations inherit directly from the root ABC.

PR #190, opened 2019-09-06
DM-21210: On transaction abort, PostgreSQL ignores commands in the entire transaction block.

DinoBektesevic:
When an SQL query fails in a transaction, PostgreSQL invalidates the entire transaction. Any subsequent SQL statement will then fail with 

```
sqlalchemy.exc.InternalError: (psycopg2.InternalError) current transaction is aborted, commands ignored until end of transaction block
```

statement. In `SqlRegistry.associate` method an insert statement is expected to fail sometimes, but a rollback is never issued to fix the transaction, leading to this error being raised on any following query, even when correct.

Minor fix to `PostgreSqlRegistry.setConfigRoot` by adding a `toCopy` parameter that I somehow forgot in https://github.com/lsst/daf_butler/pull/189.

I put the transaction in the `SqlRegistry` but I can move it explicitly to `PostgreSqlRegistry`. I don't know if that's desired due to code duplication or not.

### DIFF
--- a/python/lsst/daf/butler/registries/postgresqlRegistry.py
+++ b/python/lsst/daf/butler/registries/postgresqlRegistry.py
@@ -68,7 +68,7 @@ class PostgreSqlRegistry(SqlRegistry):
         """
         super().setConfigRoot(root, config, full, overwrite=overwrite)
         Config.updateParameters(RegistryConfig, config, full,
-                                overwrite=overwrite)
+                                toCopy=("db",), overwrite=overwrite)
 
     def __init__(self, registryConfig, schemaConfig, dimensionConfig, create=False,
                  butlerRoot=None):
--- a/python/lsst/daf/butler/registries/sqlRegistry.py
+++ b/python/lsst/daf/butler/registries/sqlRegistry.py
@@ -562,8 +562,9 @@ class SqlRegistry(Registry):
                 raise AmbiguousDatasetError(f"Cannot associate dataset {ref} without ID.")
 
             try:
-                self._connection.execute(insertQuery, {"dataset_id": ref.id, "dataset_ref_hash": ref.hash,
-                                                       "collection": collection})
+                with self.transaction():
+                    self._connection.execute(insertQuery, {"dataset_id": ref.id, "dataset_ref_hash": ref.hash,
+                                                           "collection": collection})
             except IntegrityError as exc:
                 # Did we clash with a completely duplicate entry (because this
                 # dataset is already in this collection)?  Or is there already


andy-slac:
DM-# in PR title looks wrong.

DinoBektesevic:
Thanks, I'm not at all sure how that happened (it seems to happen often with me though).

andy-slac (approved):
I _think_ this is OK for now and it may even work with Oracle and other backends (you can ask @TallJimbo to verify Oracle). Maybe even better way is not to use nested transactions but either do row-level locking or use special conflict resolution syntax which is backend-specific (comment above in that method suggests latter).
Bigger problem is the transaction management in general, what you are fixing now is a manifestation of a wider class of issues with transaction management and isolation. We will need another look at the whole thing at some point.

DinoBektesevic:
@andy-slac Some things you said are confusing me. I do not understand what locking and concurrency have to do with this, I can achieve the same behaviour in an interactive shell and the problem occurs for `pipetask -j 1` as well. 

I believe that all that transaction there does is change:

```
BEGIN;
INSERT INTO test_table VALUES (1, 1);
INSERT INTO test_table VALUES (2, 2);
INSERT INTO test_table VALUES (1, 1);
    ERROR:  duplicate key value violates unique constraint "test_table_pkey"
    DETAIL:  Key (id)=(1) already exists.
SELECT * FROM test_table;
    ERROR:  current transaction is aborted, commands ignored until end of transaction block
COMMIT;
```

which will issue a full rollback and all will be lost, into 

```
BEGIN;
INSERT INTO test_table VALUES (1, 1);
INSERT INTO test_table VALUES (2, 2);
SAVEPOINT in_between_every_command_actually;
INSERT INTO test_table VALUES (1, 1);
    ERROR:  duplicate key value violates unique constraint "test_table_pkey"
    DETAIL:  Key (id)=(1) already exists.
/* SELECT statement in here would not work */
ROLLBACK TO SAVEPOINT in_between_every_command_actually;
SELECT * FROM test_table;
/* will work*/
COMMIT;
```
in which case 1 and 2 will not be lost. I tested this out interactively via `psql` (`set ON_ERROR_ROLLBACK off` for the realistic experience) and I can open a new connection in a new terminal and do as many inserts as I want without any locking issues. If I try and insert `(1,1)` or `(2,2)` in Terminal 2 however, I better have set a savepoint, otherwise I will have lost the entire transaction again (due to the same invalidate transaction ingoring commands in the rest of transaction block issue).

As far as the second point - I was aware of `ON CONFLICT` statement, but I can't see how to use that to achieve the same behaviour that exists currently. I can target the unique constraint on a table by inspecting `datasetCollectionTable.__table__.constraints` but the statement needs to look something like:
```
INSERT INTO .... VALUES (.......)
ON CONFLICT ON CONSTRAINT dataset_id_pkey DO NOTHING
WHERE thing_inserted_to_compare==table_thing_to_compare AND ....
```
and I'm just not that nimble. I believe then the error won't be silenced, so the same issue will exist, but we would not have needed the second query, so all would be well. Help welcomed if you want to take that approval back.

TallJimbo (approved):
I think @DinoBektesevic and @andy-slac have already said everything I would say here, and accurately characterized the situation: this is a good stopgap solution that should work for other DBs as well as PostgreSQL because it reduces our reliance on non-standard behavior by SQLite and Oracle.  It doesn't address our bigger problem with transactions, and plans are already in the works to do that (mostly via the custom ON CONFLICT / MERGE syntax discussed here and on DM-21201).

andy-slac:
I agree with Jim, this is fine for now, OK to merge. Just want to point out that even with conflict resolution in place we can get an error when committing transactions (and this could be backend-specific thing too).

PR #191, opened 2019-09-09
DM-21222: Stop forcing existence of credentials in connection strings.

DinoBektesevic:
Not forcing username and password in connection strings allows other mechanisms, such as Oracle Wallet, to insert the credentials. 

It was difficult to untangle `getAuth` into sub-functionality so `exists` is implemented in terms of `getAuth`. Is that ok?

Would this work?

### DIFF
--- a/python/lsst/daf/butler/core/connectionString.py
+++ b/python/lsst/daf/butler/core/connectionString.py
@@ -22,7 +22,7 @@
 __all__ = ("DB_AUTH_ENVVAR", "DB_AUTH_PATH", "ConnectionStringFactory")
 
 from sqlalchemy.engine import url
-from lsst.daf.butler.core.dbAuth import DbAuth, DbAuthError
+from lsst.daf.butler.core.dbAuth import DbAuth, DbAuthError, DbAuthPermissionsError
 
 DB_AUTH_ENVVAR = "LSST_DB_AUTH"
 """Default name of the environmental variable that will be used to locate DB
@@ -68,8 +68,7 @@ class ConnectionStringFactory:
         Raises
         ------
         DBAuthError
-            If the credentials file has incorrect permissions, doesn't exist at
-            the given location or is formatted incorrectly.
+            If the credentials file has incorrect permissions.
         """
         # this import can not live on the top because of circular import issue
         from lsst.daf.butler.core.registryConfig import RegistryConfig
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/core/connectionString.py
```diff
@@ -80,12 +80,20 @@ def fromConfig(cls, registryConfig):
...
+        # the credentials to exist, always re-raise only the case where
+        # credentials# file exists but is incorrect permissions, for safety
+        try:
             dbAuth = DbAuth(DB_AUTH_PATH, DB_AUTH_ENVVAR)
-            auth = dbAuth.getAuth(conStr.drivername, conStr.username, conStr.host,
-                                  conStr.port, conStr.database)
-            conStr.username = auth[0]
-            conStr.password = auth[1]
+        except DbAuthError as e:
+            if 'permissions' in e.args[0]:
```
timj: Don't rely on parsing the error message in an attempt to work out what type of problem this was. What you need to do is raise a different exception for permissions problem. You want to be able to code this as something like:
```python
try:
    dbAuth = DbAuth(...)
except DBAuthFileNotFoundError:
   pass
else:
   ...
```
and have the constructor throw `DbAuthPermissionsError` that would not be caught. They could both inherit from `DbAuthError` but you'd only be ignoring the lack of db auth file rather than ignoring everything and trying to parse the error.

DinoBektesevic: Yeah, I was pretty sure this would get a comment but wasn't sure whether a `DbAuthPermissionsError` is wanted ot should I just raise a regular python `PermissionError` (which isn't there exactly for this thing, but matched the closest to what I wanted).

ktlim:
I was just thinking of removing most of the error `raise`s, allowing `username` and `password` to be returned as empty strings (or `None`).

Reading through the file twice (for `exists` and then `getAuth`) seems undesirable.

python/lsst/daf/butler/core/connectionString.py
timj: DbAuthPermissionError could inherit from PermissionError and from DbAuthError depending on how you want people to be catching things.

DinoBektesevic:
Ok. redid it the way @ktlim wanted. 

Username and password are allowed to be `None.` Instead of avoiding `sqlite` specifically, we avoid any situation where `host` is not specified explicitly. `ConnectionStringFactory` now believes the correctness of given connection string optimistically.

brianv0 (approved):
Looks fine, any PR where you end up with less LOC is usually good 👍

DinoBektesevic:
@ktlim You are probably quite busy this week, but could you take a look and comment on the changes. Is this more like what you had in mind?

Someone should also run Jenkins on this because I don't have the privileges too.

timj (approved):
Looks okay to me. I'll do a Jenkins run. I don't think we need to wait for @ktlim to also approve.

PR #192, opened 2019-09-13
DM-21301: pipetask --register-dataset-types doesn't work with Postgres registry on AWS

hsinfang:


python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -347,7 +347,13 @@ def registerDatasetType(self, datasetType):
...
+        # If it's not in the cache, check the DB
+        if existingDatasetType is None:
+            try:
+                existingDatasetType = self.getDatasetType(datasetType.name)
+            except KeyError:
+                existingDatasetType = None
+        # If it's not in the cache nor the DB, try to insert it.
         if existingDatasetType is None:
             try:
                 self._connection.execute(
```
DinoBektesevic: I think that following the logic in https://github.com/lsst/daf_butler/pull/190 would be better. Wrapping this in `transaction` will work until a better mechanism is put in place. Do: 

```
try:
    with self.transaction():
        self._connection.execute(
            self._schema.tables.... )
except IntegrityError as exc:
    existingDatasetType = self.getDatasetType(datasetType.name)
```

instead and keep the original code comments?

I did a quick scan through the file for other `IntegrityError` checks but found that the error is either re-raised or insert is not protected on purpose. This would be the last vestige of the problem then?

hsinfang: Indeed this is the same issue as https://github.com/lsst/daf_butler/pull/190

My alternative approach here is to avoid the conflict by checking before the insert attempt, rather than always trying and then catching the conflict.  I can imagine in some other scenarios it would have to be the latter. Here we have options. 

I do see the benefits of handling this the same way as https://github.com/lsst/daf_butler/pull/190   and let [DM-21201](https://jira.lsstcorp.org/browse/DM-21201) solve the bigger problem later. 

Before I change this to the nested transactions approach, @TallJimbo  @andy-slac  do you have opinions one way or the other? Any idea which approach is more efficient?

andy-slac: I'm OK with nested transactions for now, better to use uniform approach for the same issue.

TallJimbo: :+1: for nested transactions.  I believe that approach is at least sometimes (always?) robust to parallel insertions, while querying for the item in advance definitely isn't.

TallJimbo (approved):

hsinfang:
I made a mistake in the branch name. I'll close this PR and open a new one with the correct name.

hsinfang:
The new PR with the corrected branch name is https://github.com/lsst/daf_butler/pull/193

PR #195, opened 2019-09-30
DM-21454: Include all column names in foreign key names.

TallJimbo:
This is the easiest way to ensure unique foreign key names.

### DIFF
--- a/python/lsst/daf/butler/core/schema.py
+++ b/python/lsst/daf/butler/core/schema.py
@@ -197,7 +197,7 @@ class FieldSpec:
         if self.length is not None and self.nbytes is not None:
             raise SchemaValidationError(f"Both length and nbytes provided for field '{self.name}'.")
         self.primaryKey = config.get("primaryKey", self.primaryKey)
-        self.autoincrement = config.get("autoinc", self.autoincrement)
+        self.autoincrement = config.get("autoincrement", self.autoincrement)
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/core/schema.py
```diff
@@ -305,9 +305,10 @@ def toSqlAlchemy(self, tableName: str, schema: Schema) -> ForeignKeyConstraint:
         constraint : `sqlalchemy.ForeignKeyConstraint`
             SQLAlchemy version of the foreign key constraint.
         """
+        name = "_".join(["fkey", tableName, self.table] + list(self.target) + list(self.source))
```
andy-slac, line 308: I think this has a potential to exceed name size limits for some backends.

TallJimbo: I'm a bit worried about that, too, but @MichelleGower hasn't reported any problems in her tests yet, and I don't think we can guarantee a unique name without including all of these terms, even if they aren't usually necessary.  Given that this isn't really a useful-visible name (except maybe in some pretty low-level debugging contexts), I think the next step is probably trying to hash these terms into something with fixed size, and I'd rather not do that on this ticket.

MichelleGower:
With these changes and a newer version of SQLAlchemy, I was able to run ci_hsc_gen3 with Oracle.

PR #196, opened 2019-10-04
DM-21201: Research cross-database approach to inserts with custom conflict resolution

andy-slac:
Implement conflict resolution for inserts. This is done as dialect-specific query generation with a caveat that it needs more testing. The unit tests that I added only test that generated queries look reasonable but it does not run those queries (except for SQLite for which we have complete test).
Packaging of this is a separate question, I implemented it now as a new (internal) method in SqlRegistry class but there may be other ways depending on how it will be used. Any suggestions are welcome.

### DIFF
--- a/python/lsst/daf/butler/registries/sqlRegistry.py
+++ b/python/lsst/daf/butler/registries/sqlRegistry.py
@@ -29,7 +29,7 @@ from typing import Union, Iterable, Optional, Mapping, Iterator
 
 from sqlalchemy import create_engine, text, func
 from sqlalchemy.pool import NullPool
-from sqlalchemy.sql import select, and_, bindparam, union
+from sqlalchemy.sql import select, and_, union
 from sqlalchemy.exc import IntegrityError, SADeprecationWarning
 
 from ..core.utils import transactional, NamedKeyDict
@@ -576,42 +576,29 @@ class SqlRegistry(Registry):
         self._connection.execute(datasetCompositionTable.insert().values(**values))
         parent._components[name] = component
 
-    @transactional
     def associate(self, collection, refs):
         # Docstring inherited from Registry.associate.
 
-        # Most SqlRegistry subclass implementations should replace this
-        # implementation with special "UPSERT" or "MERGE" syntax.  This
-        # implementation is only concurrency-safe for databases that implement
-        # transactions with database- or table-wide locks (e.g. SQLite).
+        def records(refs):
+            """Generate records to insert into database.
+
+            Parameters
+            ----------
+            refs : iterable of `DatasetRef`
+                An iterable of `DatasetRef` instances.
+            """
+            for ref in refs:
+                if ref.id is None:
+                    raise AmbiguousDatasetError(f"Cannot associate dataset {ref} without ID.")
+                yield {"dataset_id": ref.id, "dataset_ref_hash": ref.hash, "collection": collection}
+                yield from records(ref.components.values())
 
         datasetCollectionTable = self._schema.tables["dataset_collection"]
-        insertQuery = datasetCollectionTable.insert()
-        checkQuery = select([datasetCollectionTable.c.dataset_id], whereclause=and_(
-            datasetCollectionTable.c.collection == collection,
-            datasetCollectionTable.c.dataset_ref_hash == bindparam("hash")))
-
-        for ref in refs:
-            if ref.id is None:
-                raise AmbiguousDatasetError(f"Cannot associate dataset {ref} without ID.")
-
-            try:
-                with self.transaction():
-                    self._connection.execute(insertQuery, {"dataset_id": ref.id, "dataset_ref_hash": ref.hash,
-                                                           "collection": collection})
-            except IntegrityError as exc:
-                # Did we clash with a completely duplicate entry (because this
-                # dataset is already in this collection)?  Or is there already
-                # a different dataset with the same DatasetType and data ID in
-                # this collection?  Only the latter is an error.
-                row = self._connection.execute(checkQuery, hash=ref.hash).fetchone()
-                if row.dataset_id != ref.id:
-                    raise ConflictingDefinitionError(
-                        "A dataset of type {} with id: {} already exists in collection {}".format(
-                            ref.datasetType, ref.dataId, collection
-                        )
-                    ) from exc
-            self.associate(collection, ref.components.values())
+        values = list(records(refs))
+        try:
+            self._insert(datasetCollectionTable, values, onConflict="ignore")
+        except IntegrityError as exc:
+            raise ConflictingDefinitionError(f"A dataset already exists in collection {collection}") from exc
 
     @transactional
     def disassociate(self, collection, refs):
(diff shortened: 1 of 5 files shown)


python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -954,3 +954,83 @@ def queryDatasets(self, datasetType: DatasetTypeExpression, *,
...
+        Even with conflict resolution options it is possible that inserts
+        will generate conflicts due to concurrency and implementation details
+        of transaction isolation. When it happens the only reasonable course
+        of action is to restart transaction and repeat the whole operation.
+        Conflicts can also appear due to violation of other non-PK constraints
+        and it is not possible to distinguish those. To avoid infinite looping
+        on non-PK constraint violations this method only performs few retries.
+
+        This method needs to handle transactions itself, do not call it if you
+        are already in a transaction.
```
TallJimbo, line 977: Could we relax (perhaps optionally) this restriction so that it does a savepoint instead, and hence can be used inside existing transactions (with some performance penalty, probably)?  I think that's just a matter of doing `self._connection.begin_nested()` instead of `self._connection.begin()`.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -954,3 +954,83 @@ def queryDatasets(self, datasetType: DatasetTypeExpression, *,
...
+                query = table.insert()
+                self._connection.execute(query, values)
+            return
+
+        # When doing non-aborting conflict resolution the COMMIT could
+        # potentially fail, in that case we want to restart transaction and
+        # re-run the whole thing again, but not forever.
+        retries = 0
+        query = self._makeInsertWithConflict(table, onConflict=onConflict)
+        while True:
```
TallJimbo, line 998: Seems like a `for` loop with a break would be a cleaner way of limiting the number of retries.

python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -954,3 +954,83 @@ def queryDatasets(self, datasetType: DatasetTypeExpression, *,
...
+            except IntegrityError:
+                # There error could be due to PK conflict or other unique key
+                # conflict, there is no way to identify exact reason, so we
+                # re-try several times.
+                if retries > retryLimit:
+                    # stop trying, looks like we can't win
+                    raise
+                retries += 1
+
+    def _makeInsertWithConflict(self, table, onConflict):
```
TallJimbo, line 1014: Might as well make this an `abstractmethod`, I think.

TallJimbo (approved):
Looks great!

Dispatching this from a protected `Registry` method is as good as anything else right now.  I might move it to a free function dispatched on the SQLAlchemy dialect in the future, so I can have some helper classes for `Registry` that don't need to have back pointer to the `Registry` instance, but I'm by no means certain I'll do that.

The one open question I have is about your statement from Jira about the multiple-constraint (`dataset_collection`) case:

> After reading a bit more on PostgreSQL syntax it is clear that ON CONFLICT DO UPDATE can only work with a single constraint and it is not possible to specify more than one constraint.
>
> I think for our use case this may be OK, in dataset_collections table insert we could just depend on PK uniqueness to achive the same behavior as we have now:
>
> - When inserting we could do "ON CONFLICT (dataset_id, collection) DO NOTHING"
> - When same (dataset_id, collection) is already there it means that we already OK, there is no need to update dataset_ref_hash (it is the same by construction)
> - If there a record with different (dataset_id, collection) but the same (dataset_ref_hash, collection) then it will cause an error (as it is not handled by "ON CONFLICT")

The added complication is that for this table, when the PK constraint in this table is violated, the non-PK constraint is always also violated - so we have to rely on the "IGNORE" behavior for the PK constraint having higher precedence than the implied "ABORT" logic for the non-PK constraint.  I tested that with SQLite in the past and got different results for different versions, and the documentation doesn't say what behavior is expected (though at least the newer version has the behavior we want).  I [asked about this on Stack Overflow](https://stackoverflow.com/questions/54261493/sqlite-unique-conflict-resolution-ordering
), and got no responses.

Given that the SQLite syntax was added with the intent of mimicking PostgreSQL (at least that's what the docs suggest), I'd feel a lot better about this if we could verify that it is the behavior PostgreSQL is documented to have - without that, I might still be inclined to switch to using REPLACE even if that changes the current behavior, because at least then it'd be reliable behavior.  Have you tested this case, with that added complication, on PostgreSQL and/or SQLite?

python/lsst/daf/butler/registries/sqlRegistry.py
andy-slac: This is a tricky part - the reason why it has to be done in a completely separate transaction is that in some cases (with concurrent updates and depending on backend) integrity errors could only be detected on COMMIT. Nested transactions do not do COMMIT, only top-level transactions do. So if we allow this transaction to be nested into another one then we lose logic for catching COMMIT-time errors. This also applies to other cases in `SqlRegistry` where we catch `IntegrityError` inside `@transactional` methods - depending on circumstances this exception could happen in the decorator itself and not in the try/except code.

python/lsst/daf/butler/registries/sqlRegistry.py
andy-slac: I think that re-throwing logic would be less clear then, with this code it's more obvious that loop stops either on success (`break`) or it raises an exception.

andy-slac:
Regarding multiple constraints - I believe that in PostgreSQL description (https://www.postgresql.org/docs/9.5/sql-insert.html#SQL-ON-CONFLICT) it is clear that INSERT does not happen if constraint in ON CONFLICT is violated, and that I think implies that no other constraint is ever triggered. SQLite should follow the same logic if they implement ON CONFLICT after PostgreSQL, the unit test should catch case when SQLite behaves differently.

python/lsst/daf/butler/registries/sqlRegistry.py
TallJimbo: Gotcha.  But in that case I think we need to move the begin/commit/retry logic out of this function (and probably into a context manager), even if that means the scope of what we need to retry increases - we do need higher-level transactions so logical operations that involve multiple tables can't die in the middle and leave things in an inconsistent state.

It might be possible to move the retry logic into a decorator, but we'd have to be very careful with how we used it.  This will be tricky regardless, and require a lot more care.  But I'm hoping that if we can define our operations such that operations that can recover from a conflict always use special ignore or replace syntax (and don't care which of any concurrent writes "win" in either case).  In that case, we shouldn't get failures on COMMIT anyway (right?), and of course operations that can't recover from conflicts don't need to retry.  Does that make sense?

andy-slac: Even with special conflict resolution syntax it could be possible that COMMIT generates an error in case of concurrent updates - this is all very much implementation-dependent and can also vary with different transaction isolation policies.
I think we indeed can move retry logic into a reusable method, I'd probably prefer context manager over decorator to have more explicit control of what needs to be retried and what recovery options should be taken.

PR #197, opened 2019-10-07
DM-21451: Remove DatabaseDict and vectorize Datastore/Butler ingest

TallJimbo:


python/lsst/daf/butler/core/registry.py
```diff
@@ -197,32 +197,76 @@ def transaction(self):
...
 
     @abstractmethod
-    def makeDatabaseDict(self, table, key, value):
-        """Construct a `DatabaseDict` backed by a table in the same database as
-        this Registry.
+    def registerOpaqueTable(self, name: str, spec: TableSpec):
+        """Add an opaque (to the `Registry`) table for use by a `Datastore` or
+        other data repository client.
+
+        Opaque table record can be added via `insertOpaqueData`, retreived via
```
timj: Typo: retreived

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -305,97 +324,53 @@ def put(self, inMemoryDataset, ref):
...
-            # Raise rather than use boolean return value.
-            raise DatasetTypeNotSupportedError(f"Dataset {ref} has been rejected by this datastore via"
-                                               " configuration.")
-
-        if transfer is None:
-            raise NotImplementedError("ChainedDatastore does not support transfer=None")
-
-        # A "move" is sometimes a "copy"
-        moveIsCopy = False
-        if transfer == "move" and os.path.isabs(path):
```
timj, line 365: You decided against keeping this logic?

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -305,97 +324,53 @@ def put(self, inMemoryDataset, ref):
...
+                log.debug("Datastore %s skipping ingest via configuration for ref %s", name, dataset.ref)
+                return False
+            else:
+                return True
 
-        notImplementedCounter = 0
-        notAcceptedCounter = 0
+        # Filter down to just datasets the chained datastore's own
+        # configuration accepts.
+        okForParent: List[FileDataset] = [dataset for dataset in datasets
```
timj, line 341: If `okForParent` is empty shouldn't we raise DatasetTypeNotSupportedError ?

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -305,97 +324,53 @@ def put(self, inMemoryDataset, ref):
...
-                          datastore.name, ref)
-                notAcceptedCounter += 1
-                continue
-
-            dstransfer = transfer
-            # Each child datastore must copy the file for a move operation
-            if moveIsCopy:
-                dstransfer = "copy"
+            if constraints is not None:
+                okForChild: List[FileDataset] = [dataset for dataset in okForParent
```
timj, line 353: I'm a bit concerned that we've lost any visibility into whether any of the supplied datasets were ingested or not. Don't we worry if a dataset was rejected by all the child datastores because of constraints? As written it seems we worry about transfer modes but not constraints. If I ask 10 datasets to be ingested and only 9 are accepted how do I tell that? Previously ingest worked if a dataset was accepted by at least one datastore (and then we had the issue of deciding what to do if it was only accepted by an ephemeral datastore).

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -202,51 +215,41 @@ def __init__(self, config, registry, butlerRoot=None):
...
+                     storage_class=info.storageClass.name,
+                     checksum=info.checksum, file_size=info.file_size)
+            )
+        self.registry.insertOpaqueData(self._tableName, *records)
+
+    def getStoredItemInfo(self, ref):
+        # Docstring inherited from GenericBaseDatastore
+        records = list(self.registry.fetchOpaqueData(self._tableName, dataset_id=ref.id))
+        if len(records) == 0:
+            raise KeyError("Unable to retrieve formatter associated with Dataset {}".format(ref.id))
```
timj: I don't think that error message should be referring to a formatter.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -365,27 +368,107 @@ def _prepare_for_put(self, inMemoryDataset, ref):
...
-        fileInfo = StoredFileInfo(formatter, path, ref.datasetType.storageClass,
-                                  file_size=size, checksum=checksum)
-        self._register_dataset(ref, fileInfo)
+        raise NotImplementedError("Must be implemented by subclasses.")
+
+    def _prepIngest(self, *datasets: FileDataset, transfer: Optional[str] = None) -> _IngestPrepData:
+        # Docstring inherited from Datastore._prepIngest.
+        filtered = []
+        for dataset in datasets:
+            if not self.constraints.isAcceptable(dataset.ref):
```
timj, line 452: Again how do we know that some of the datasets were not ingested?

python/lsst/daf/butler/datastores/genericDatastore.py
```diff
@@ -121,29 +82,33 @@ def removeStoredItemInfo(self, ref):
...
         """
-        self.registry.addDatasetLocation(ref, self.name)
+        expandedRefs = list(refs)
+        expandedItemInfos = list(itemInfos)
 
-        # TODO: this is only transactional if the DatabaseDict uses
-        #       self.registry internally.  Probably need to add
-        #       transactions to DatabaseDict to do better than that.
-        self.addStoredItemInfo(ref, itemInfo)
+        for ref, itemInfo in zip(refs, itemInfos):
```
timj: I don't think zip is going to complain if the number of refs is not equal to the number of infos. Would it be safer to accept a sequence of tuples instead?

timj (approved):
This looks good in general. My main concern is that I'm not entirely clear we have a story for cases where some files are not ingested due to constraints. Previously this triggered DatasetTypeNotSupporterdError but I can't see where that happens now (maybe it does and I've missed it).

python/lsst/daf/butler/datastores/chainedDatastore.py
TallJimbo: Yeah.  At first I was just going to move it around, but it was a lot messier after the split into two methods, and it wasn't clear to me we had a real need for it.

python/lsst/daf/butler/datastores/chainedDatastore.py
TallJimbo: Raising that exception (in ingest) is now done exclusively by `Datastore.ingest`, which will check that the set of datasets actually ingested matches the set of datasets that the caller asked to be ingested and raise if that isn't true.  That's because we need `_prepIngest` to still return the list of datasets it *can* ingest even when it can't ingest all of them, and it can't do that if it raises.

python/lsst/daf/butler/datastores/chainedDatastore.py
TallJimbo: I think that's all still true (see reply at https://github.com/lsst/daf_butler/pull/197#discussion_r338674106).  Unlike the constraint-based rejections, transfer-mode rejections mean that all datasets would be rejected by the nested datastore, so we don't have to return anything and we can still use the raise-and-catch logic that was in place before.

python/lsst/daf/butler/datastores/chainedDatastore.py
timj: I see that now. Thanks.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
TallJimbo: I've just moved it here from https://github.com/lsst/daf_butler/pull/197/commits/797a3dfe93f6a1435f2a0ec5201fc1b6535271ab#diff-4b3b7a0376839e94233cd3a2bddd157bL112, but I think I agree.  `"Unable to retrieve location associated with Dataset {}"` instead?

python/lsst/daf/butler/datastores/fileLikeDatastore.py
TallJimbo: (I think this is already answered above)

PR #198, opened 2019-10-14
DM-21748: Fixes for Oracle MERGE implementation

andy-slac:
The _Merge class has to inherit Executable so that Connection can
execute it, the order of base classes for _Merge is important. Call to
bindparam() in generated MERGE needs a `type_` argument to correctly
interpret parameter data using all registered type decorators.

### DIFF
--- a/python/lsst/daf/butler/registries/oracleRegistry.py
+++ b/python/lsst/daf/butler/registries/oracleRegistry.py
@@ -24,6 +24,7 @@ __all__ = ("OracleRegistry", )
 from sqlalchemy import create_engine
 from sqlalchemy.ext import compiler
 from sqlalchemy.sql import ClauseElement, and_, bindparam, select
+from sqlalchemy.sql.expression import Executable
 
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/registries/oracleRegistry.py
```diff
@@ -48,7 +49,7 @@ def _merge(merge, compiler, **kw):
     pkColumns = [col.name for col in table.primary_key]
     nonPkColumns = [col for col in allColumns if col not in pkColumns]
 
-    selectColumns = [bindparam(col).label(col) for col in allColumns]
+    selectColumns = [bindparam(col.name, type_=col.type).label(col.name) for col in table.columns]
```
MichelleGower, line 54: For future code readers, a comment about why the type_ is needed would probably be helpful.

MichelleGower (approved):
Code changes match discussion.  Changes fixed failures when running ci_hsc_gen3 with Oracle.

PR #202, opened 2019-11-08
Temporary change to allow gen3 fringes to work.

czwa:


TallJimbo:
Is this still needed after DM-22068, or is this now just being added as a nice-to-have?

czwa:
It is not needed after DM-22068, so I left it as a "nice to have."  I'm fine dropping it if not having a use case is a concern.

python/lsst/daf/butler/core/deferredDatasetHandle.py
```diff
@@ -112,3 +116,37 @@ def get(self, parameters: typing.Union[None, dict] = None, **kwargs: dict) -> ty
...
+        if self.parameters is not None:
+            mergedParameters = self.parameters.copy()
+            if parameters is not None:
+                mergedParameters.update(parameters)
+        elif parameters is not None:
+            mergedParameters = parameters
+        else:
+            mergedParameters = {}
+
+        return self._getUri(mergedParameters)
```
natelust: I don't think it is a good idea to have all the logic for merging parameters duplicated different places, I think if you want to keep this function, that both of these get factored out into a common private method that these both use.

As a side note I have a ticket open to allow fetching one component of a composite dataset given this handle. We should talk about how that may intersect with your ticket.

natelust (approved):
Approve with the caveat that the duplication should be removed first.

python/lsst/daf/butler/core/deferredDatasetHandle.py
```diff
@@ -112,3 +116,37 @@ def get(self, parameters: typing.Union[None, dict] = None, **kwargs: dict) -> ty
...
+
+    def getUri(self, parameters: typing.Union[None, dict] = None, **kwargs: dict) -> typing.Any:
+        """ Retrieves the dataset URI pointed to by this handle
+
+        This handle may be used multiple times, possibly with different
+        parameters.
+
+        Parameters
+        ----------
+        parameters : `dict` or None
```
timj: I mentioned this on the commit itself but not in the PR, but getUri doesn't take parameters so I'm confused by the use of the term here. (also the docstring talks about `get`).

czwa:
I'm going to close this PR.  The initial change is no longer needed, and I'd rather not complicate the ticket goal with extra butler code.

PR #204, opened 2019-11-13
DM-22205: Better exception handling for single str input.

leannep:


python/lsst/daf/butler/registries/sqlRegistry.py
```diff
@@ -831,7 +831,9 @@ def queryDimensions(self, dimensions: Iterable[Union[Dimension, str]], *,
                         where: Optional[str] = None,
                         expand: bool = True,
                         **kwds) -> Iterator[DataCoordinate]:
-        # Docstring inherited from Registry.queryDimensions.
+        # Docstring inherited from Registry.queryDimensions
+        if isinstance(dimensions, str):
```
timj: @TallJimbo since you can take an iterable of `str` maybe the better option here is to use `lsst.daf.butler.core.utils.iterable` on the argument?

natelust (approved):
Good to go after addressing Tim's comment

jonathansick:
I think we want to remove the `.idea` directory from the commits?

timj:
Yes, add them to .gitignore so this won't happen again.

timj:
Can we squah the idea removal commit with the commit that added them so that the history doesn't show they ever existed?

PR #205, opened 2019-11-14
DM-22073: Add support for writing matplotlib figures

leeskelvin:


tests/test_matplotlibFormatter.py
```diff
@@ -0,0 +1,73 @@
...
+
+"""Tests for MatplotlibFormatter.
+"""
+
+import unittest
+import tempfile
+import os
+import shutil
+
+import matplotlib
```
timj: Please protect the matplotlib import and skip the tests if matplotlib is not available.  daf_butler should not require matplotlib.

python/lsst/daf/butler/formatters/matplotlibFormatter.py
```diff
@@ -0,0 +1,43 @@
...
+__all__ = ("MatplotlibFormatter",)
+
+from lsst.daf.butler.formatters.fileFormatter import FileFormatter
+
+
+class MatplotlibFormatter(FileFormatter):
+    """Interface for writing matplotlib figures.
+    """
+
+    extension = '.png'
```
timj: Please use double quotes in daf_butler to be consistent with the rest of the code base.

tests/test_matplotlibFormatter.py
```diff
@@ -0,0 +1,73 @@
...
+        if os.path.exists(self.root):
+            shutil.rmtree(self.root, ignore_errors=True)
+
+    def testMatplotlibFormatter(self):
+        butler = Butler(self.root, run='testrun')
+        datasetType = DatasetType('test_plot', [], 'Plot',
+                                  universe=butler.registry.dimensions)
+        butler.registry.registerDatasetType(datasetType)
+        pyplot.imshow(np.random.randn(3, 4))
+        butler.put(pyplot.gcf(), datasetType)
```
timj: @TallJimbo does this work without a DataId because the dataset type has no dimensions?

tests/test_matplotlibFormatter.py
```diff
@@ -0,0 +1,73 @@
...
+        with tempfile.NamedTemporaryFile(suffix='.png') as file:
+            pyplot.gcf().savefig(file.name)
+            # predicting the filename path based on test run
+            self.assertTrue(
+                filecmp.cmp(
+                    os.path.join(self.root, 'testrun', datasetType.name,
```
timj: I think I would prefer it if this path was constructed from calling `butler.getUri` since then you aren't relying on knowing how the file name was constructed within the datastore.

tests/test_matplotlibFormatter.py
```diff
@@ -0,0 +1,73 @@
...
+                                 f'{datasetType.name}_{butler.run.id}.png'),
+                    file.name,
+                    shallow=True
+                )
+            )
```
timj, line 75: Please add calls to `butler.datasetExists` and `butler.remove` so that we can check that these PNG images are being treated like a normal dataset. Also try to do a `butler.get` and check that it fails with `with self.assertRaises`.

timj (approved):
This looks good although I do have some comments for improvements to the tests.

It is a bit annoying that `butler.get` can't work but it is important that `butler.getUri` is tested.

tests/test_matplotlibFormatter.py
TallJimbo: Yes, exactly.

tests/test_matplotlibFormatter.py
TallJimbo: That's what we tried first, but annoyingly `filecmp` apparently can't handle `file://` prefixes, and it occurred to me that *actual* usage would look more like this, but perhaps with users explicitly defining the template so they'd know where to look in general, rather than asking the butler where to look each time.

Do you have an incantation handy for how to get that `file://` prefix off gracefully?

TallJimbo:
@leeskelvin, I'm going to let you take the lead on addressing these comments, but please find me for more pair-programming if you want help with anything.

tests/test_matplotlibFormatter.py
timj: You can either use `ButlerURI` or you can go for the standard option of:
```python
parsed = urllib.parse.urlparse(uri)
```
and then use `parsed.path`.

tests/test_matplotlibFormatter.py
timj: I think that since this code could be seen as providing example usage that we should not be relying on the super special case here of datasetType and datasetRef being interchangeable. Please receive the `ref` from the `butler.put` here and then use it later in the `getUri`  and later butler methods below.  This all works because there are no dimensions but we don't want to give the impression that it's going to work more generally. 

```python
ref = butler.put(...)
parsed = ...(butler.getUri(ref))
...
butler.datasetExists(ref)
```
etc.

tests/test_matplotlibFormatter.py
gpdf: Is this resolved?  I agree that it's important.

gpdf: (It _looks_ resolved, given the new `try` block for the import of `matplotlib`, but I haven't been able to test it in practice.)

TallJimbo: It's hard for any of us to test it, I think, because we all use conda enviroments that always need to have matplotlib for other reasons.  But I'm pretty confident what was merged will work.

PR #207, opened 2019-11-21
DM-21900: Allow additional dimension information to be used in file templates

timj:
detector.name_in_raft and visit.name now work.

### DIFF
--- a/python/lsst/daf/butler/core/fileTemplates.py
+++ b/python/lsst/daf/butler/core/fileTemplates.py
@@ -294,6 +294,11 @@ class FileTemplate:
     At least one or both of `run` or `collection` must be provided to ensure
     unique filenames.
 
+    More detailed information can be requested from dimensions by using a dot
+    notation, so ``visit.name`` would use the name of the visit and
+    ``detector.name_in_raft`` would use the name of the detector within the
+    raft.
+
     The mini-language is extended to understand a "?" in the format
     specification. This indicates that a field is optional. If that
     Dimension is missing the field, along with the text before the field,
@@ -331,7 +336,7 @@ class FileTemplate:
     def __repr__(self):
         return f'{self.__class__.__name__}("{self.template}")'
 
-    def fields(self, optionals=False, specials=False):
+    def fields(self, optionals=False, specials=False, subfields=False):
         """Return the field names used in this template.
 
         Parameters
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -443,6 +457,14 @@ def format(self, ref):
             if field_name in ("run", "collection"):
                 usedRunOrCollection = True
 
+            # Check for request for additional information from the dataId
+            if "." in field_name:
+                primary, secondary = field_name.split(".")
+                if primary in extras:
+                    element = extras[primary]
+                    if hasattr(element, secondary):
+                        fields[field_name] = getattr(element, secondary)
```
TallJimbo: What happens if the template has a dotted name that doesn't exist?  From this it looks like we just don't do anything with it, but I'm probably not seeing the big picture.

tests/test_butler.py
```diff
@@ -191,6 +191,7 @@ def runPutGetTest(self, storageClass, datasetTypeName):
             with self.subTest(args=args):
                 ref = butler.put(metric, *args)
                 self.assertIsInstance(ref, DatasetRef)
+                print(f"ref={ref.dataId.full}")
```
TallJimbo: Did you intend to leave this in?

TallJimbo (approved):
Looks good; glad it wasn't difficult.  Just one question about how we handle bad template placeholders.

python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -443,6 +457,14 @@ def format(self, ref):
             if field_name in ("run", "collection"):
                 usedRunOrCollection = True
 
+            # Check for request for additional information from the dataId
+            if "." in field_name:
+                primary, secondary = field_name.split(".")
+                if primary in extras:
+                    element = extras[primary]
```
TallJimbo: Probably better to call this local variable "record"; "element" would refer to the keys in this dict, but "primary" is also totally reasonable for that since it reflects where the key is coming from in the template.

tests/test_butler.py
timj: No. Sorry.

python/lsst/daf/butler/core/fileTemplates.py
timj: A few lines later you get an exception because `field_name` won't be in `fields`. It will say something like "detector.name is not defined" or something.

PR #208, opened 2019-11-25
DM-22364: Allow makeRepo returned Config to be used to create Butler

timj:


TallJimbo (approved):
I can't say I fully understand what's going on inside `Config` here, except for the fact that it looks like copy construction was previously skipping a key attribute.  I'm a little concerned that we apparently treat named config files differently from in-memory config objects (in that we remember their filenames and that's apparently important), and would appreciate an explanation of that, but certainly don't think it should get in the way of what's an obvious improvement.

python/lsst/daf/butler/core/repoRelocation.py
```diff
@@ -45,16 +45,27 @@ def replaceRoot(configRoot, butlerRoot):
...
-    if butlerRoot is None:
-        butlerRoot = os.path.curdir
+    # Do nothing if there is nothing to be done
+    if BUTLER_ROOT_TAG not in configRoot:
+        return configRoot
+
+    # None or empty string indicate a problem
+    if not butlerRoot:
+        raise ValueError(f"Required to replace {BUTLER_ROOT_TAG} in '{configRoot}' "
+                         "but a replacement has not been defined")
```
TallJimbo, line 69: Can you think of any way a common high-level operation would get to this error?  If so, we may want to beef up the error message with a guess about what might have gone wrong, but I'm guessing the other changes on this ticket mean it's hard to imagine how this might happen, and that's fine too.

TallJimbo (approved):

PR #210, opened 2019-12-05
DM-21448: Clean up DatasetRef comparisons and immutability

TallJimbo:


python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -0,0 +1,259 @@
...
+        self.datasetType = datasetType
+        if conform:
+            self.dataId = DataCoordinate.standardize(dataId, graph=datasetType.dimensions)
+        else:
+            self.dataId = dataId
+        self._components = dict()
+        if components is not None:
+            self._components.update(components)
+        self.run = run
+        if hash is not None:
```
timj, line 121: It seems that `None` means that `_hash` will not be set at all and later on the hash will be calculated. If there is a reason that `None` wouldn't work here rather than `hasattr` in `hash()` can we add a comment about it?

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -0,0 +1,259 @@
...
+            self._hash = hash
+        return self
+
+    def __eq__(self, other: DatasetRef):
+        return (self.datasetType, self.dataId, self.id) == (other.datasetType, other.dataId, other.id)
```
timj: If `other` is not a `DatasetRef` should we return `False` or should we raise?  I always tended to think that `__eq__` should return `False` if they aren't equal.

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def __hash__(self) -> int:
+        return hash(self.datasetType, self.dataId, self.id)
+
+    def __repr__(self):
+        return f"DatasetRef({self.datasetType}, {self.dataId}, id={self.id}, run={self.run})"
```
timj: What about components?

TallJimbo:
Just added another commit because Jenkins failed, uncovering (in ci_hsc_gen3) some more loophole-based mutating.

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -0,0 +1,278 @@
...
+        and the given ID and run.
+
+        Parameters
+        ----------
+        id : `int`
+            The unique integer identifier assigned when the dataset is created.
+        run : `Run`
+            The run this dataset was associated with when it was created.
+        components : `dict`, optional
+            A dictionary mapping component name to a `DatasetRef` for that
```
timj, line 200: I think this should also mention that the components supplied here overwrite any existing components but also merged with the existing ones. If this is a resolved datasetRef should we check that all the components have resolved dataset refs? Should we check they have the same run? What if there are some unresolved components and this call only overwrites some of them with resolved versions?

tests/test_datasets.py
```diff
@@ -196,38 +203,37 @@ class DatasetRefTestCase(unittest.TestCase):
 
     def setUp(self):
```
timj, line 205: Are there any tests of components in DatasetRef?

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -0,0 +1,278 @@
...
+    id : `int`, optional
+        The unique integer identifier assigned when the dataset is created.
+    run : `Run`, optional
+        The run this dataset was associated with when it was created.
+    hash : `bytes`, optional
+        A hash of the dataset type and data ID.  Should only be provided if
+        copying from another `DatasetRef` with the same dataset type and data
+        ID.
+    components : `dict`, optional
+        A dictionary mapping component name to a `DatasetRef` for that
```
timj, line 58: We don't enforce that the names of these components match the allowed names for the associated StorageClass. I also don't think we check that `run` in the components matches the parent or whether some components have IDs but others don't.

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -0,0 +1,278 @@
...
+        """
+        newComponents = self._components.copy()
+        if components:
+            newComponents.update(components)
+        return DatasetRef(datasetType=self.datasetType, dataId=self.dataId,
+                          id=id, run=run, hash=self.hash, components=newComponents, conform=False)
+
+    def unresolved(self) -> DatasetRef:
+        """Return a new `DatasetRef` with the same data ID and dataset type,
+        but no ID, run, or components.
```
timj, line 221: Why are components dropped rather than including them in their unresolved form?

timj (approved):
This looks good and useful. My main comment is that the role of components seems somewhat under-defined in terms of whether they are meant to be unresolved if the parent is unresolved and whether it's allowed to have components listed in the DatasetRef that are not present in the StorageClass.

python/lsst/daf/butler/core/datasets/ref.py
TallJimbo: There is a reason - the `immutable` decorator works by permitting all attributes to be set only once.  I'll add the comment.

python/lsst/daf/butler/core/datasets/ref.py
TallJimbo: Thanks for the reminder.  I think the right thing to do is actually to return `NotImplemented` to give `other.__eq__` a chance.

python/lsst/daf/butler/core/datasets/ref.py
TallJimbo: I'm just maintaining the old behavior here, which was to only include them in `__str__` for for brevity.  Happy to go either way if you have a preference; I don't.

python/lsst/daf/butler/core/datasets/ref.py
TallJimbo: So, I'm pretty sure we have concrete uses for:
 - resolved `DatasetRefs` with components
 - resolved `DatasetRefs` without components
 - unresolved `DatasetRefs` without components

where by "without components" I mean that the `DatasetRef` doesn't know about any components it might have, even if they exist, because it's just a wasteful to do anything with them because we know they won't be used.

The dataset-insertion interplay between `Registry` and `Datastore` is also a lot simpler if we allow the components to be updated in-place in a `DatasetRef` (and I think that's okay w.r.t. immutability, since components aren't used in `__eq__`/`__hash__`).

Anyhow, with all that in mind, how about this:
 - We set `components` to `None` on an unresolved `DatasetRef`.  We raise in `__init__` if components are passed without `id` and `run`, and merge in `resolved()` only if re-resolving an already resolved `DatasetRef`.
 - We check that all component `DatasetRefs` are resolved, and that names are valid according to the storage class.

tests/test_datasets.py
TallJimbo: There are some indirect ones over in `test_sqlRegistry.py`.

python/lsst/daf/butler/core/datasets/ref.py
TallJimbo: :+1: to checking the component names against the storage class.  I don't think the run has to be the same between the parent and components in general (specifically when components are written as separate datasets and then connected into a virtual composite later).  We don't do that in practice right now, and I'm not sure it all works, but I don't want to throw up any road blocks.

python/lsst/daf/butler/core/datasets/ref.py
TallJimbo: See other discussion on components - I don't think we ever want the components on an unresolved `DatasetRef`, and I didn't want to waste effort making them.

TallJimbo:
@timj , I've acted on all of your comments, but I'm not sure I've actually resolved all of your concerns.  Any further thoughts?

python/lsst/daf/butler/core/datasets/ref.py
timj: In some sense `repr` is more relevant for reporting them. I did try to compare th new version with the original to see what was previously there. If we've survived without components turning up in `repr` this long it's probably fine to leave them out.

python/lsst/daf/butler/core/datasets/ref.py
timj: Sounds good to me.

tests/test_datasets.py
timj: Can we have an explicit test here? Especially given the changes to components handling discussed above.

python/lsst/daf/butler/core/datasets/ref.py
timj: Sounds good to me.

timj:
@TallJimbo sorry for the delay. I've made some minor follow up comments but I mostly agree with your plan.

python/lsst/daf/butler/core/datasets/ref.py
TallJimbo: I've rewritten them to make `repr` the long one and `str` even more concise.  I think `exec(repr(datasetRef) == datasetRef` would now be true *if* `StorageClass` similarly round-tripped, but I think that's out of scope for this ticket.

tests/test_datasets.py
TallJimbo: Done, and thanks for pushing for this; at first I'd thought the tests even caught a bug in the new code, but instead they caught a bug in the new test code (trying to pass a `DatasetType` instead of a `DatasetRef` as a component) that was worth adding a special check for to yield a better error message.

PR #212, opened 2019-12-09
DM-22508: add Database class, concrete implementations, and tests

TallJimbo:
The code here is not used at all in Butler yet; it's the beginning of a new backend for Registry, and hence there's some temporary code duplication with existing Registry code.

The testing strategies for PostgreSQL and Oracle merit some special attention:

- For PostgreSQL, we use the `testing.postgresql` module (available via `pip`)  to spin up a local userspace database server to test against.  We do that in `setUpClass` to avoid doing it repeatedly, and then create one or more schemas within that database _per test function_.  We delegate cleaning that all up in `tearDownClass` to `testing.postgresql`.  Tests are skipped if `testing.postgresql` cannot be imported, and we assume that the presence of `testing.postgresql` indicates that the actual PostgreSQL server binaries are in the PATH, because `testing.postgresql` can't ever do anything unless that's true, even though it's not taken care of by `pip` (the `postgresql` conda package does take care of it, however).

- For Oracle, we look for a `DAF_BUTLER_ORACLE_TEST_URI` environment variable that provides the connection string to pass to SQLAlchemy.  If that's set, we make a single connection to the database server (in `setUpClass`) and use (as @MichelleGower has advocated) a randomly-generated prefix for all tables we create in that schema to simulate a virtual schema.  There's custom code in `tearDownClass` that then drops all tables that were created, using a Python DBAPI rewrite of the `clobber.sql` script at https://confluence.lsstcorp.org/display/DM/Oracle+with+Gen3.

While this basically works, it's worth pointing out the disadvantages of the Oracle approach; while the PostgreSQL is obviously much harder to pull off with any proprietary database engine, it'd be really nice to set up something more like that for Oracle, even if it's only something we can interact with via some CI server.  In particular:
 - SQLAlchemy basically takes care of real schemas for us, and we sort of have to fight against it to use table prefixes to make those work like a virtual schema.  The need to use prefixes actually adds more complexity to the `OracleDatabase` specialization than anything else unique to Oracle.
 - By not using real schemas in the Oracle tests, we don't actually get test coverage for trying to use real schemas in Oracle, which we of course really want to do in the future.
 - While I'm pretty sure test cleanup usually happens correctly when tests fail, I'd guess there are myriad ways to interrupt the tests such that the cleanup is at best incomplete.  And that of course means one has to go in and clean things up manually.
 - Running tests against a local server is lot faster - or at least I assume that's the reason for the pretty huge gap between the ~1.6s PostgreSQL tests and Oracle tests, which I've seen take anything between 20s and 3 minutes (it might also be something silly I've done, or overhead in Oracle in creating tables - which tests naturally do a lot more of than production use would).

### DIFF
--- a/python/lsst/daf/butler/registry/databases/oracle.py
+++ b/python/lsst/daf/butler/registry/databases/oracle.py
@@ -27,7 +27,7 @@ from typing import Optional
 
 import sqlalchemy
 
-from ..interfaces import ConnectionStruct, Database, ReadOnlyDatabaseError
+from ..interfaces import Database, ReadOnlyDatabaseError
 from .. import ddl
 from ..nameShrinker import NameShrinker
 
@@ -93,7 +93,7 @@ class OracleDatabase(Database):
 
     Parameters
     ----------
-    cs : `ConnectionStruct`
+    connection : `sqlalchemy.engine.Connection`
         An existing connection created by a previous call to `connect`.
     origin : `int`
         An integer ID that should be used as the default for any datasets,
@@ -112,36 +112,36 @@ class OracleDatabase(Database):
         tables.
     """
 
-    def __init__(self, *, cs: ConnectionStruct, origin: int, namespace: Optional[str] = None,
-                 writeable: bool = True, prefix: Optional[str] = None):
+    def __init__(self, *, connection: sqlalchemy.engine.Connection, origin: int,
+                 namespace: Optional[str] = None, writeable: bool = True, prefix: Optional[str] = None):
         # Get the schema that was included/implicit in the URI we used to
         # connect.
-        dbapi = cs.engine.raw_connection()
+        dbapi = connection.engine.raw_connection()
         namespace = dbapi.current_schema
-        super().__init__(cs=cs, origin=origin, namespace=namespace)
+        super().__init__(connection=connection, origin=origin, namespace=namespace)
         self._writeable = writeable
         self.dsn = dbapi.dsn
         self.prefix = prefix
-        self._shrinker = NameShrinker(cs.engine.dialect.max_identifier_length)
+        self._shrinker = NameShrinker(connection.engine.dialect.max_identifier_length)
 
     @classmethod
-    def connect(cls, uri: str, *, writeable: bool = True) -> ConnectionStruct:
-        cs = ConnectionStruct(sqlalchemy.engine.create_engine(uri, pool_size=1))
+    def connect(cls, uri: str, *, writeable: bool = True) -> sqlalchemy.engine.Connection:
+        connection = sqlalchemy.engine.create_engine(uri, pool_size=1).connect()
         # Work around SQLAlchemy assuming that the Oracle limit on identifier
         # names is even shorter than it is after 12.2.
-        oracle_ver = cs.engine.dialect._get_server_version_info(cs.connection)
+        oracle_ver = connection.engine.dialect._get_server_version_info(connection)
         if oracle_ver < (12, 2):
             raise RuntimeError("Oracle server version >= 12.2 required.")
-        cs.engine.dialect.max_identifier_length = 128
-        return cs
+        connection.engine.dialect.max_identifier_length = 128
+        return connection
 
     @classmethod
-    def fromConnectionStruct(cls, cs: ConnectionStruct, *, origin: int, namespace: Optional[str] = None,
-                             writeable: bool = True) -> Database:
+    def fromConnection(cls, connection: sqlalchemy.engine.Connection, *, origin: int,
+                       namespace: Optional[str] = None, writeable: bool = True) -> Database:
         # TODO: is there anything we can do to make the connection read-only if
         # writeable is False?  If not (and at present) we just rely on the
         # Python code being well-behaved and not *trying* to make changes.
-        return cls(cs=cs, origin=origin, writeable=writeable, namespace=namespace)
+        return cls(connection=connection, origin=origin, writeable=writeable, namespace=namespace)
 
     def isWriteable(self) -> bool:
         return self._writeable
@@ -180,7 +180,7 @@ class OracleDatabase(Database):
     def replace(self, table: sqlalchemy.schema.Table, *rows: dict):
         if not self.isWriteable():
             raise ReadOnlyDatabaseError(f"Attempt to replace into read-only database '{self}'.")
-        self._cs.connection.execute(_Merge(table), *rows)
+        self._connection.execute(_Merge(table), *rows)
 
     prefix: Optional[str]
     """A prefix included in all table names to simulate a database namespace
--- a/python/lsst/daf/butler/registry/databases/oracle.py
+++ b/python/lsst/daf/butler/registry/databases/oracle.py
@@ -22,6 +22,7 @@ from __future__ import annotations
 
 __all__ = ["OracleDatabase"]
 
+from contextlib import closing, contextmanager
 import copy
 from typing import Optional
 
@@ -138,10 +139,15 @@ class OracleDatabase(Database):
     @classmethod
     def fromConnection(cls, connection: sqlalchemy.engine.Connection, *, origin: int,
                        namespace: Optional[str] = None, writeable: bool = True) -> Database:
-        # TODO: is there anything we can do to make the connection read-only if
-        # writeable is False?  If not (and at present) we just rely on the
-        # Python code being well-behaved and not *trying* to make changes.
         return cls(connection=connection, origin=origin, writeable=writeable, namespace=namespace)
+
+    @contextmanager
+    def transaction(self, *, interrupting: bool = False) -> None:
+        with super().transaction(interrupting=interrupting):
+            if not self.isWriteable():
+                with closing(self._connection.connection.cursor()) as cursor:
+                    cursor.execute("SET TRANSACTION READ ONLY")
+            yield
 
     def isWriteable(self) -> bool:
         return self._writeable
--- a/python/lsst/daf/butler/registry/databases/oracle.py
+++ b/python/lsst/daf/butler/registry/databases/oracle.py
@@ -27,7 +27,7 @@ from typing import Optional
 
 import sqlalchemy
 
-from ..interfaces import Database, ReadOnlyDatabaseError
+from ..interfaces import ConnectionStruct, Database, ReadOnlyDatabaseError
 from .. import ddl
 from ..nameShrinker import NameShrinker
 
@@ -93,61 +93,55 @@ class OracleDatabase(Database):
 
     Parameters
     ----------
+    cs : `ConnectionStruct`
+        An existing connection created by a previous call to `connect`.
     origin : `int`
         An integer ID that should be used as the default for any datasets,
         quanta, or other entities that use a (autoincrement, origin) compound
         primary key.
-    uri : `str`, optional
-        Database connection string.  If not provided, ``engine`` must be.
-    engine : `sqlalchemy.engine.Engine`, optional
-        SQLAlchemy engine instance to use directly.
     namespace : `str`, optional
-        Namespace (schema) for all tables used by this database.  May be,
-        `None` which will use the schema implied by ``uri`` or ``engine``
-        (which may be `None` for no schema).
+        The namespace (schema) this database is associated with.  If `None`,
+        the default schema for the connection is used (which may be `None`).
+    writeable : `bool`, optional
+        If `True`, allow write operations on the database, including
+        ``CREATE TABLE``.
     prefix : `str`, optional
         Prefix to add to all table names, effectively defining a virtual
         schema that can coexist with others within the same actual database
         schema.  This prefix must not be used in the un-prefixed names of
         tables.
-    writeable : `bool`, optional
-        If `True` (default) allow writes to the database.
     """
 
-    def __init__(self, *, origin: int,
-                 uri: Optional[str] = None,
-                 engine: Optional[sqlalchemy.engine.Engine] = None,
-                 namespace: Optional[str] = None,
-                 prefix: Optional[str] = None,
-                 writeable: bool = True):
-        if engine is None:
-            engine = sqlalchemy.engine.create_engine(uri, pool_size=1)
-        connection = engine.connect()
-        # Work around SQLAlchemy assuming that the Oracle limit on identifier
-        # names is even shorter than it is after 12.2.
-        oracle_ver = engine.dialect._get_server_version_info(connection)
-        if oracle_ver < (12, 2):
-            raise RuntimeError("Oracle server version >= 12.2 required.")
-        engine.dialect.max_identifier_length = 128
+    def __init__(self, *, cs: ConnectionStruct, origin: int, namespace: Optional[str] = None,
+                 writeable: bool = True, prefix: Optional[str] = None):
         # Get the schema that was included/implicit in the URI we used to
         # connect.
-        dbapi = engine.raw_connection()
-        if namespace is None:
-            namespace = dbapi.current_schema
-        else:
-            if dbapi.current_schema != namespace:
-                # TODO: could we instead add the schema to the URI before
-                # trying to connect?  Or is an Oracle connection without a
-                # schema not something that can exist?
-                raise RuntimeError("'namespace' and 'uri' arguments specify different schemas.")
-        # TODO: is there anything we can do to make the connection read-only if
-        # writeable is False?  If not (and at present) we just rely on the
-        # Python code being well-behaved and not *trying* to make changes.
-        super().__init__(origin=origin, engine=engine, namespace=namespace, connection=connection)
+        dbapi = cs.engine.raw_connection()
+        namespace = dbapi.current_schema
+        super().__init__(cs=cs, origin=origin, namespace=namespace)
         self._writeable = writeable
         self.dsn = dbapi.dsn
         self.prefix = prefix
-        self._shrinker = NameShrinker(engine.dialect.max_identifier_length)
+        self._shrinker = NameShrinker(cs.engine.dialect.max_identifier_length)
+
+    @classmethod
+    def connect(cls, uri: str, *, writeable: bool = True) -> ConnectionStruct:
+        cs = ConnectionStruct(sqlalchemy.engine.create_engine(uri, pool_size=1))
+        # Work around SQLAlchemy assuming that the Oracle limit on identifier
+        # names is even shorter than it is after 12.2.
+        oracle_ver = cs.engine.dialect._get_server_version_info(cs.connection)
+        if oracle_ver < (12, 2):
+            raise RuntimeError("Oracle server version >= 12.2 required.")
+        cs.engine.dialect.max_identifier_length = 128
+        return cs
+
+    @classmethod
+    def fromConnectionStruct(cls, cs: ConnectionStruct, *, origin: int, namespace: Optional[str] = None,
+                             writeable: bool = True) -> Database:
+        # TODO: is there anything we can do to make the connection read-only if
+        # writeable is False?  If not (and at present) we just rely on the
+        # Python code being well-behaved and not *trying* to make changes.
+        return cls(cs=cs, origin=origin, writeable=writeable, namespace=namespace)
 
     def isWriteable(self) -> bool:
         return self._writeable
@@ -186,7 +180,7 @@ class OracleDatabase(Database):
     def replace(self, table: sqlalchemy.schema.Table, *rows: dict):
         if not self.isWriteable():
             raise ReadOnlyDatabaseError(f"Attempt to replace into read-only database '{self}'.")
-        self._connection.execute(_Merge(table), *rows)
+        self._cs.connection.execute(_Merge(table), *rows)
 
     prefix: Optional[str]
     """A prefix included in all table names to simulate a database namespace
--- a/python/lsst/daf/butler/registry/databases/oracle.py
+++ b/python/lsst/daf/butler/registry/databases/oracle.py
@@ -1,0 +1,198 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = ["OracleDatabase"]
+
+import copy
+from typing import Optional
+
+import sqlalchemy
+
+from ..interfaces import Database, ReadOnlyDatabaseError
+from .. import ddl
+from ..nameShrinker import NameShrinker
+
+
+class _Merge(sqlalchemy.sql.expression.Executable, sqlalchemy.sql.ClauseElement):
+    """A SQLAlchemy query that compiles to a MERGE invocation that is the
+    equivalent of PostgreSQL and SQLite's INSERT ... ON CONFLICT REPLACE on the
+    primary key constraint for the table.
+    """
+
+    def __init__(self, table):
+        super().__init__()
+        self.table = table
+
+
+@sqlalchemy.ext.compiler.compiles(_Merge, "oracle")
+def _merge(merge, compiler, **kw):
+    """Generate MERGE query for inserting or updating records.
+    """
+    table = merge.table
+    preparer = compiler.preparer
+
+    allColumns = [col.name for col in table.columns]
+    pkColumns = [col.name for col in table.primary_key]
+    nonPkColumns = [col for col in allColumns if col not in pkColumns]
+
+    # To properly support type decorators defined in core/schema.py we need
+    # to pass column type to `bindparam`.
+    selectColumns = [sqlalchemy.sql.bindparam(col.name, type_=col.type).label(col.name)
+                     for col in table.columns]
+    selectClause = sqlalchemy.sql.select(selectColumns)
+
+    tableAlias = table.alias("t")
+    tableAliasText = compiler.process(tableAlias, asfrom=True, **kw)
+    selectAlias = selectClause.alias("d")
+    selectAliasText = compiler.process(selectAlias, asfrom=True, **kw)
+
+    condition = sqlalchemy.sql.and_(
+        *[tableAlias.columns[col] == selectAlias.columns[col] for col in pkColumns]
+    )
+    conditionText = compiler.process(condition, **kw)
+
+    query = f"MERGE INTO {tableAliasText}" \
+            f"\nUSING {selectAliasText}" \
+            f"\nON ({conditionText})"
+    updates = []
+    for col in nonPkColumns:
+        src = compiler.process(selectAlias.columns[col], **kw)
+        dst = compiler.process(tableAlias.columns[col], **kw)
+        updates.append(f"{dst} = {src}")
+    updates = ", ".join(updates)
+    query += f"\nWHEN MATCHED THEN UPDATE SET {updates}"
+
+    insertColumns = ", ".join([preparer.format_column(col) for col in table.columns])
+    insertValues = ", ".join([compiler.process(selectAlias.columns[col], **kw) for col in allColumns])
+
+    query += f"\nWHEN NOT MATCHED THEN INSERT ({insertColumns}) VALUES ({insertValues})"
+    return query
+
+
+class OracleDatabase(Database):
+    """An implementation of the `Database` interface for Oracle.
+
+    Parameters
+    ----------
+    origin : `int`
+        An integer ID that should be used as the default for any datasets,
+        quanta, or other entities that use a (autoincrement, origin) compound
+        primary key.
+    uri : `str`, optional
+        Database connection string.  If not provided, ``engine`` must be.
+    engine : `sqlalchemy.engine.Engine`, optional
+        SQLAlchemy engine instance to use directly.
+    namespace : `str`, optional
+        Namespace (schema) for all tables used by this database.  May be,
+        `None` which will use the schema implied by ``uri`` or ``engine``
+        (which may be `None` for no schema).
+    prefix : `str`, optional
+        Prefix to add to all table names, effectively defining a virtual
+        schema that can coexist with others within the same actual database
+        schema.  This prefix must not be used in the un-prefixed names of
+        tables.
+    writeable : `bool`, optional
+        If `True` (default) allow writes to the database.
+    """
+
+    def __init__(self, *, origin: int,
+                 uri: Optional[str] = None,
+                 engine: Optional[sqlalchemy.engine.Engine] = None,
+                 namespace: Optional[str] = None,
+                 prefix: Optional[str] = None,
+                 writeable: bool = True):
+        if engine is None:
+            engine = sqlalchemy.engine.create_engine(uri, pool_size=1)
+        connection = engine.connect()
+        # Work around SQLAlchemy assuming that the Oracle limit on identifier
+        # names is even shorter than it is after 12.2.
+        oracle_ver = engine.dialect._get_server_version_info(connection)
+        if oracle_ver < (12, 2):
+            raise RuntimeError("Oracle server version >= 12.2 required.")
+        engine.dialect.max_identifier_length = 128
+        # Get the schema that was included/implicit in the URI we used to
+        # connect.
+        dbapi = engine.raw_connection()
+        if namespace is None:
+            namespace = dbapi.current_schema
+        else:
+            if dbapi.current_schema != namespace:
+                # TODO: could we instead add the schema to the URI before
+                # trying to connect?  Or is an Oracle connection without a
+                # schema not something that can exist?
+                raise RuntimeError("'namespace' and 'uri' arguments specify different schemas.")
+        # TODO: is there anything we can do to make the connection read-only if
+        # writeable is False?  If not (and at present) we just rely on the
+        # Python code being well-behaved and not *trying* to make changes.
+        super().__init__(origin=origin, engine=engine, namespace=namespace, connection=connection)
+        self._writeable = writeable
+        self.dsn = dbapi.dsn
+        self.prefix = prefix
+        self._shrinker = NameShrinker(engine.dialect.max_identifier_length)
+
+    def isWriteable(self) -> bool:
+        return self._writeable
+
+    def __str__(self) -> str:
+        if self.namespace is None:
+            name = self.dsn
+        else:
+            name = f"{self.dsn:self.namespace}"
+        return f"Oracle@{name}"
+
+    def shrinkDatabaseEntityName(self, original: str) -> str:
+        return self._shrinker.shrink(original)
+
+    def expandDatabaseEntityName(self, shrunk: str) -> str:
+        return self._shrinker.expand(shrunk)
+
+    def _convertForeignKeySpec(self, table: str, spec: ddl.ForeignKeySpec, metadata: sqlalchemy.MetaData,
+                               **kwds) -> sqlalchemy.schema.ForeignKeyConstraint:
+        if self.prefix is not None:
+            spec = copy.copy(spec)
+            spec.table = self.prefix + spec.table
+        return super()._convertForeignKeySpec(table, spec, metadata, **kwds)
+
+    def _convertTableSpec(self, name: str, spec: ddl.TableSpec, metadata: sqlalchemy.MetaData,
+                          **kwds) -> sqlalchemy.schema.Table:
+        if self.prefix is not None and not name.startswith(self.prefix):
+            name = self.prefix + name
+        return super()._convertTableSpec(name, spec, metadata, **kwds)
+
+    def getExistingTable(self, name: str, spec: ddl.TableSpec) -> Optional[sqlalchemy.schema.Table]:
+        if self.prefix is not None and not name.startswith(self.prefix):
+            name = self.prefix + name
+        return super().getExistingTable(name, spec)
+
+    def replace(self, table: sqlalchemy.schema.Table, *rows: dict):
+        if not self.isWriteable():
+            raise ReadOnlyDatabaseError(f"Attempt to replace into read-only database '{self}'.")
+        self._connection.execute(_Merge(table), *rows)
+
+    prefix: Optional[str]
+    """A prefix included in all table names to simulate a database namespace
+    (`str` or `None`).
+    """
+
+    dsn: str
+    """The TNS entry of the database this instance is connected to (`str`).
+    """
--- a/python/lsst/daf/butler/registry/databases/postgresql.py
+++ b/python/lsst/daf/butler/registry/databases/postgresql.py
@@ -26,7 +26,7 @@ from typing import Optional
 
 import sqlalchemy
 
-from ..interfaces import Database, ReadOnlyDatabaseError
+from ..interfaces import ConnectionStruct, Database, ReadOnlyDatabaseError
 from ..nameShrinker import NameShrinker
 
 
@@ -35,17 +35,18 @@ class PostgresqlDatabase(Database):
 
     Parameters
     ----------
+    cs : `ConnectionStruct`
+        An existing connection created by a previous call to `connect`.
     origin : `int`
         An integer ID that should be used as the default for any datasets,
         quanta, or other entities that use a (autoincrement, origin) compound
         primary key.
-    uri : `str`, optional
-        Database connection string.
     namespace : `str`, optional
-        Namespace (schema) for all tables used by this database.  May be `None`
-        only if a schema is included in ``uri``.
+        The namespace (schema) this database is associated with.  If `None`,
+        the default schema for the connection is used (which may be `None`).
     writeable : `bool`, optional
-        If `True` (default) allow writes to the database.
+        If `True`, allow write operations on the database, including
+        ``CREATE TABLE``.
 
     Notes
     -----
@@ -55,26 +56,36 @@ class PostgresqlDatabase(Database):
     PostgreSQL server is installed and can be run locally in userspace.
     """
 
-    def __init__(self, *, origin: int, uri: str, namespace: Optional[str] = None, writeable: bool = True):
-        engine = sqlalchemy.engine.create_engine(uri, pool_size=1)
-        dbapi = engine.raw_connection()
+    def __init__(self, *, cs: ConnectionStruct, origin: int, namespace: Optional[str] = None,
+                 writeable: bool = True):
+        super().__init__(origin=origin, cs=cs, namespace=namespace)
+        dbapi = cs.engine.raw_connection()
         try:
             dsn = dbapi.get_dsn_parameters()
         except (AttributeError, KeyError) as err:
             raise RuntimeError(f"Only the psycopg2 driver for PostgreSQL is supported.") from err
         if namespace is None:
-            namespace = dsn.get("schema")
-            if namespace is None:
-                raise RuntimeError("Namespace (schema) must be provided as an argument or included in URI.")
-        else:
-            if dsn.get("schema", namespace) != namespace:
-                raise RuntimeError("'namespace' and 'uri' arguments specify different schemas.")
-        super().__init__(origin=origin, engine=engine, namespace=namespace)
-        if not writeable:
-            dbapi.set_session(readonly=True)
-        self.dbname = dsn["dbname"]
+            namespace = dsn["schema"]
+        self.namespace = namespace
+        self.dbname = dsn.get("dbname")
         self._writeable = writeable
-        self._shrinker = NameShrinker(engine.dialect.max_identifier_length)
+        self._shrinker = NameShrinker(cs.engine.dialect.max_identifier_length)
+
+    @classmethod
+    def connect(cls, uri: str, *, writeable: bool = True) -> ConnectionStruct:
+        cs = ConnectionStruct(sqlalchemy.engine.create_engine(uri, pool_size=1))
+        if not writeable:
+            dbapi = cs.engine.raw_connection()
+            try:
+                dbapi.set_session(readonly=True)
+            except AttributeError as err:
+                raise RuntimeError(f"Only the psycopg2 driver for PostgreSQL is supported.") from err
+        return cs
+
+    @classmethod
+    def fromConnectionStruct(cls, cs: ConnectionStruct, *, origin: int, namespace: Optional[str] = None,
+                             writeable: bool = True) -> Database:
+        return cls(cs=cs, origin=origin, namespace=namespace, writeable=writeable)
 
     def isWriteable(self) -> bool:
         return self._writeable
@@ -102,4 +113,4 @@ class PostgresqlDatabase(Database):
                 for column in table.columns
                 if column.name not in table.primary_key}
         query = query.on_conflict_do_update(constraint=table.primary_key, set_=data)
-        self._connection.execute(query, *rows)
+        self._cs.connection.execute(query, *rows)
--- a/python/lsst/daf/butler/registry/databases/postgresql.py
+++ b/python/lsst/daf/butler/registry/databases/postgresql.py
@@ -1,0 +1,81 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = ["PostgresqlDatabase"]
+
+from typing import Optional
+
+import sqlalchemy
+
+from ..interfaces import Database
+
+
+class PostgresqlDatabase(Database):
+    """An implementation of the `Database` interface for PostgreSQL.
+
+    Parameters
+    ----------
+    origin : `int`
+        An integer ID that should be used as the default for any datasets,
+        quanta, or other entities that use a (autoincrement, origin) compound
+        primary key.
+    uri : `str`, optional
+        Database connection string.
+    namespace : `str`, optional
+        Namespace (schema) for all tables used by this database.  May be `None`
+        only if a schema is included in ``uri``.
+    writeable : `bool`, optional
+        If `True` (default) allow writes to the database.
+
+    Notes
+    -----
+    This currently requires the psycopg2 driver to be used as the backend for
+    SQLAlchemy.  Running the tests for this class requires the
+    ``testing.postgresql`` be installed, which we assume indicates that a
+    PostgreSQL server is installed and can be run locally in userspace.
+    """
+
+    def __init__(self, *, origin: int, uri: str, namespace: Optional[str] = None, writeable: bool = True):
+        engine = sqlalchemy.engine.create_engine(uri, pool_size=1)
+        dbapi = engine.raw_connection()
+        try:
+            dsn = dbapi.get_dsn_parameters()
+        except (AttributeError, KeyError) as err:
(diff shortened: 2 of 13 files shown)


python/lsst/daf/butler/registry/databases/sqlite.py
```diff
@@ -0,0 +1,217 @@
...
+    on the primary key constraint for the table.
+    """
+    pass
+
+
+@sqlalchemy.ext.compiler.compiles(_Replace, "sqlite")
+def _replace(insert, compiler, **kw):
+    """Generate an INSERT ... ON CONFLICT REPLACE query.
+
+    Unlike the PostgreSQL case, sqlalchemy does not provide special support for
```
DinoBektesevic: Is "Unlike the PostgreSQL case" neccessary?

python/lsst/daf/butler/registry/databases/sqlite.py
```diff
@@ -0,0 +1,217 @@
...
+    return result
+
+
+_AUTOINCR_TABLE_SPEC = ddl.TableSpec(
+    fields=[ddl.FieldSpec(name="id", dtype=sqlalchemy.Integer, primaryKey=True)]
+)
+
+
+@dataclass
+class _AutoincrementWorkaround:
```
DinoBektesevic: Rename to "_AutoincrementCompoundKey"? I understand this is, partially, internal and is used only once but would it still make code that uses it easier to read?

python/lsst/daf/butler/registry/databases/sqlite.py
```diff
@@ -0,0 +1,217 @@
...
+    def replace(self, table: sqlalchemy.schema.Table, *rows: dict):
+        if not self.isWriteable():
+            raise ReadOnlyDatabaseError(f"Attempt to replace into read-only database '{self}'.")
+        if table.name in self._autoincr:
+            raise NotImplementedError(
+                "replace does not support compound primary keys with autoincrement fields."
+            )
+        self._connection.execute(_Replace(table), *rows)
+
+    filename: Optional[str]
```
DinoBektesevic, line 351: I'm confused, if this is a class attribute shouldn't it be on top under class docstring?

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+)
+
+import sqlalchemy
+
+from .. import ddl
+
+
+def _checkExistingTableDefinition(name: str, spec: ddl.TableSpec, inspection: Dict[str, Any]):
+    """Test that the definition of a table in a `ddl.TableSpec` and from
+    database introspection are the consistent.
```
DinoBektesevic: "the" is extra?

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+
+        Notes
+        -----
+        The default implementation should be sufficient for most derived
+        classes.
+        """
+        # TODO: should we guard against non-SELECT queries here?
+        return self._connection.execute(sql, *args, **kwds)
+
+    origin: int
```
DinoBektesevic, line 1001: I went looking through the DM Style-guide and couldn't find a place that mentions positioning of class attributes so I assume this is ok too. I am confused though, is this a new thing everyone should be doing? I thought they should all go to the top?

tests/test_postgresql.py
```diff
@@ -0,0 +1,130 @@
...
+        cls.server.stop()
+
+    def setUp(self):
+        cls = type(self)
+        cls.testCounter += 1
+        self.dbCounter = 1
+
+    def makeEmptyDatabase(self, origin: int = 0) -> PostgresqlDatabase:
+        namespace = f"namespace_{self.testCounter}_{self.dbCounter}"
+        self.dbCounter += 1
```
DinoBektesevic: Pytest parallel tests run in separate processes, so the counter won't be shared but it won't be unique either. It was my impression from reading the code that namespace needs to be separate to avoid conflicts because this is all executing in the same DB. 

Did I miss-understand? If not, should this be replaced by something that gives unique counters?

python/lsst/daf/butler/registry/tests/database.py
```diff
@@ -0,0 +1,387 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+
```
andy-slac, line 22: Probably needs `__all__` here?

python/lsst/daf/butler/registry/nameShrinker.py
```diff
@@ -0,0 +1,70 @@
...
+
+__all__ = ["NameShrinker"]
+
+import hashlib
+
+
+class NameShrinker:
+    """A utility class for `Database` implementations that need a nontrivial
+    implementation of `Database.shrinkDatabaseEntityName` and
+    `Database.expandDatabaseEntityName`.
```
andy-slac, line 31: If all users of this class are implementation classes in `database` package maybe this module should be there too?

python/lsst/daf/butler/registry/ddl.py
```diff
@@ -0,0 +1,191 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = ("TableSpec", "FieldSpec", "ForeignKeySpec", "Base64Bytes", "Base64Region")
+
+from base64 import b64encode, b64decode
+from math import ceil
+from dataclasses import dataclass
+from typing import Optional, Tuple, Set
+import typing
```
andy-slac: Looks like this import is used for `typing.Sequence`, I'm not sure why `Sequence` is treated differently from `Tuple` or `Set` imported above, can you do all of them in the same style?

python/lsst/daf/butler/registry/ddl.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def process_bind_param(self, value, dialect):
+        # 'value' is native `bytes`.  We want to encode that to base64 `bytes`
+        # and then ASCII `str`, because `str` is what SQLAlchemy expects for
+        # String fields.
+        if value is None:
+            return None
+        if not isinstance(value, bytes):
+            raise TypeError(
+                f"Base64Bytes fields require 'bytes' values; got {value} with type {type(value)}"
```
andy-slac: Put quotes around `{value}` in case it's empty or very long?

python/lsst/daf/butler/registry/ddl.py
```diff
@@ -0,0 +1,191 @@
...
+
+    If not `None` (which indicates that a constraint violation exception should
+    be raised), should be either "SET NULL" or "CASCADE".
+    """
+
+
+@dataclass
+class TableSpec:
+    """A struct-like class used to define a logical `Registry` table (which may
+    be implemented in the database as a query on any number of tables).
```
andy-slac: Not sure how to read this right, does "query" mean a regular SQL view here or does it mean that we can represent schema of the regular query with this structure?
If it can be used for schema of arbitrary query then one potential issue is that column names in the query result may not be unique (this obviously depends on particular backend and a query) and using NamedValueSet for fields can be an issue. I do not know if this is an issue for registry schema, just a general comment.

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+        Raised if the definitions are inconsistent.
+    """
+    columnNames = [c["name"] for c in inspection]
+    if spec.fields.names != set(columnNames):
+        raise RuntimeError(f"Table '{name}' exists but is defined differently in the database; "
+                           f"specification has columns {list(spec.fields.names)}, while the "
+                           f"table in the database has {columnNames}.")
+
+
+class TransactionInterruption(RuntimeError):
```
andy-slac: I think exception classes should end with "Error". Same for `SynchronizationConflict` below.

python/lsst/daf/butler/registry/interfaces/database.py
andy-slac: extraneous `the`.

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+    Raises
+    ------
+    RuntimeError
+        Raised if the definitions are inconsistent.
+    """
+    columnNames = [c["name"] for c in inspection]
+    if spec.fields.names != set(columnNames):
+        raise RuntimeError(f"Table '{name}' exists but is defined differently in the database; "
+                           f"specification has columns {list(spec.fields.names)}, while the "
+                           f"table in the database has {columnNames}.")
```
andy-slac: Do you plan checking for types in the future?

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+            _checkExistingTableDefinition(name, spec, self._inspector.get_columns(name,
+                                                                                  schema=self._db.namespace))
+        table = self._db._convertTableSpec(name, spec, self._db._metadata)
+        for foreignKeySpec in spec.foreignKeys:
+            self._foreignKeys.append(
+                (table, self._db._convertForeignKeySpec(name, foreignKeySpec, self._db._metadata))
+            )
+        return table
+
+    def addTableTuple(self, specs: Tuple[ddl.TableSpec]) -> Tuple[sqlalchemy.schema.Table]:
```
andy-slac: Should it be `NamedTuple` instead of `Tuple`?

andy-slac (commented):
I only started looking into it, but cannot finish today. Here are the comments that I have so far, will continue tomorrow.

DinoBektesevic (commented):
I don't see anything obviously wrong but I have only started walking through the tests this afternoon. I've noticed Andy pushing his review and I've noticed we overlap in places I'm adding questions I had when I read through it. 

Thanks for simultaneously fixing the name length limitations for PostgreSQL, that helps me a lot!

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+        """Return a context manager that represents a transaction.
+
+        Parameters
+        ----------
+        savepoint : `bool`
+            If `True`, issue a ``SAVEPOINT`` command that allows this
+            transaction to be rolled back or committed independent of any
+            outer transaction.  If `False` (default), transactions may still
+            be nested, but nothing will be committed until the outermost
+            transaction context ends.
```
andy-slac: It's phrased a bit confusingly, with SAVEPONT commit does not actually happen until outermost (regular) transaction is committed. Main difference between those is in rollback, with SAVEPOINT rollback means "rollback to savepoint", while with nested regular transaction rollback aborts outermost transaction.

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+
+        Raises
+        ------
+        TransactionInterruption
+            Raised if ``interrupting`` is `True` and a transaction is already
+            active.
+        """
+        if interrupting and self._transactions:
+            raise TransactionInterruption("Logic error in transaction nesting: an operation that would "
+                                          "the active transation context has been requested.")
```
andy-slac: I think there is a missing word in this message.

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+        This is passed as the ``schema`` argument when constructing a
+        `sqlalchemy.schema.MetaData` instance.  We use ``namespace`` instead to
+        avoid confusion between "schema means namespace" and "schema means
+        table definitions".
+    engine: `sqlalchemy.engine.Engine`.
+        SQLAlchemy engine object.  May be shared with other `Database`
+        instances.
+    connection : `sqlalchemy.engine.Connection`, optional
+        SQLAlchemy connection object.  May be shared with other `Database`
+        instances.  Constructed from ``engine`` if not provided.
```
andy-slac: Do we really need connections shared between Databases? With shared connection some logic in this class may be broken, e.g. `_transactions` may be empty but transaction may actually be in progress if started by another Database instance.

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+            self._transactions.pop()
+
+    @contextmanager
+    def declareStaticTables(self, *, create: bool) -> StaticTablesContext:
+        """Return a context manager in which the database's static DDL schema
+        can be declared.
+
+        Parameters
+        ----------
+        create : `bool` If `True`, attempt to create all tables at the end of
```
andy-slac: Need a line break after type, same in Returns section

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+        if create and not self.isWriteable():
+            raise ReadOnlyDatabaseError(f"Cannot create tables in read-only database {self}.")
+        self._metadata = sqlalchemy.MetaData(schema=self.namespace)
+        try:
+            context = StaticTablesContext(self)
+            yield context
+            for table, foreignKey in context._foreignKeys:
+                table.append_constraint(foreignKey)
+            if create:
+                if create and self.namespace is not None:
```
andy-slac: `create` is tested on the line above already

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+                    if self.namespace not in context._inspector.get_schema_names():
+                        self._engine.execute(sqlalchemy.schema.CreateSchema(self.namespace))
+                self._metadata.create_all(self._engine)
+        except BaseException:
+            self._metadata.drop_all(self._engine)
```
andy-slac: This looks dangerous, should we only drop tables that were created by previous call to `create_all` (not sure if we can find which ones existed before this call)?

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+    @abstractmethod
+    def __str__(self) -> str:
+        """Return a human-readable identifier for this `Database`, including
+        any namespace or schema that identifies its names within a `Registry`.
+        """
+        raise NotImplementedError()
+
+    def shrinkDatabaseEntityName(self, original: str) -> str:
+        """Return a version of the given name that fits within this database
+        engine's length limits for table, constraint, indexe, and sequence
```
andy-slac: `indexe`

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+        Parameters
+        ----------
+        table : `str`
+            Name of the table this column is being added to.
+        spec : `FieldSpec`
+            Specification for the field to be added.
+        metadata : `sqlalchemy.MetaData`
+            SQLAlchemy representation of the DDL schema this field's table is
+            being added to.
+        kwds
```
andy-slac: I have no strong opinion on this but our doc examples show this should be `**kwds`: https://developer.lsst.io/python/numpydoc.html#parameters

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+        )
+
+    def _convertTableSpec(self, name: str, spec: ddl.TableSpec, metadata: sqlalchemy.MetaData,
+                          **kwds) -> sqlalchemy.schema.Table:
+        """Convert a `TableSpec` to a `sqlalchemy.schema.Table`.
+
+        Parameters
+        ----------
+        spec : `TableSpec`
+            Specification for the foreign key to be added.
```
andy-slac, line 501: foreign key?

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+            SQLAlchemy representation of the table.
+
+        Raises
+        ------
+        TransactionInterruption
+            Raised if a transaction is active when this method is called.
+        ReadOnlyDatabaseError
+            Raised if `isWriteable` returns `False`, and the table does not
+            already exist.
+        RuntimeError
```
andy-slac: Given that TransactionInterruption and ReadOnlyDatabaseError inherit from RuntimeError there is a potential for confusion if someone write code like this:

    try:
        db.ensureTableExists(...)
    catch RuntimeError:
        log.error("Where did you get this spec, it's inconsistent with database schema!")
    catch TransactionInterruption
        log.error("What are you doing, database is already in transaction!")

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+                    # Need to perform check() for this branch inside the
+                    # transaction, so we roll back an insert that didn't do
+                    # what we expected.  That limits the extent to which we
+                    # can reduce duplication between this block and the other
+                    # ones that perform similar logic.
+                    n, bad, result = check()
+                    if n < 1:
+                        raise RuntimeError("Insertion in sync did not seem to affect table.  This is a bug.")
+                    elif n > 2:
+                        raise RuntimeError("Keys passed to sync do not comprise a unique constraint.")
```
andy-slac: Maybe give a list of key columns and table name to simplify debugging?

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+                                           "possible if the same table is being updated by a concurrent "
+                                           "process that isn't using sync.")
+                # No exceptions, so it looks like we inserted the requested row
+                # successfully.
+                inserted = True
+            except sqlalchemy.exc.IntegrityError as err:
+                # Most likely cause is that an equivalent row already exists,
+                # but it could also be some other constraint.  Query for the
+                # row we think we matched to resolve that question.
+                n, bad, result = check()
```
andy-slac, line 772: Do you need `self.transaction()` around this line?

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+
+        May be used inside transaction contexts, so implementations may not
+        perform operations that interrupt transactions.
+        """
+        if not self.isWriteable():
+            raise ReadOnlyDatabaseError(f"Attempt to insert into read-only database '{self}'.")
+        if not returnIds:
+            self._connection.execute(table.insert(), *rows)
+        else:
+            return [self._connection.execute(table.insert(), row).inserted_primary_key[0] for row in rows]
```
andy-slac: Maybe move `table.insert()` out of the loop for small optimization.

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+    def replace(self, table: sqlalchemy.schema.Table, *rows: dict):
+        """Insert one or more rows into a table, replacing any existing rows
+        for which insertion of a new row would violate the primary key
+        constraint.
+
+        Parameters
+        ----------
+        table : `sqlalchemy.schema.Table`
+            Table rows should be inserted into.
+        rows
```
andy-slac: `*rows` maybe?

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+        raise NotImplementedError()
+
+    def delete(self, table: sqlalchemy.schema.Table, columns: Iterable[str], *rows: dict) -> int:
+        """Delete one or more rows from a table.
+
+        Parameters
+        ----------
+        table : `sqlalchemy.schema.Table`
+            Table that rows should be deleted from.
+        keys: `~collections.abc.Iterable` of `str`
```
andy-slac: keys -> columns

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+            The names of columns that will be used to constrain the rows to
+            be deleted; these will be combined via ``AND`` to form the
+            ``WHERE`` clause of the delete query.
+        rows
+            Positional arguments are the keys of rows to be deleted, asIterable
```
andy-slac: I think this needs some rephrasing.

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+        Return
```
andy-slac: Needs an empty line before it.

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+        ------
+        count : `int`
+            Number of rows deleted.
+
+        Raises
+        ------
+        ReadOnlyDatabaseError
+            Raised if `isWriteable` returns `False` when this method is called.
```
andy-slac, line 912: Double backticks for `isWriteable`

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+            Table containing the rows to be updated.
+        keys : `dict` [`str`, `str`]
+            A mapping from the names of columns that will be used to search for
+            existing rows to the keys that will hold these values in ``rows``
+            dictionaries.  Note that these may not be the same due to
+            SQLAlchemy limitations.
+        rows
+            Positional arguments are the rows to be updated, as dictionaries
+            mapping column name to value.  The keys in all dictionaries must
+            be the combination of the names in ``keys`` and ``values``.
```
andy-slac: There are no `values` parameter, I think you may need it.

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,927 @@
...
+
+        The default implementation should be sufficient for most derived
+        classes.
+        """
+        if not self.isWriteable():
+            raise ReadOnlyDatabaseError(f"Attempt to update read-only database '{self}'.")
+        sql = table.update().where(
+            sqlalchemy.sql.and_(*[table.columns[k] == sqlalchemy.sql.bindparam(v) for k, v in keys.items()])
+        )
+        return self._connection.execute(sql, *rows).rowcount
```
andy-slac, line 972: I think something is missing here, UPDATE needs `.values()` method too in addition to `.where()`.

andy-slac (commented):
Few more comments for `Database` class, will continue working on other modules.

python/lsst/daf/butler/registry/tests/database.py
```diff
@@ -0,0 +1,387 @@
...
+    def testDynamicTables(self):
+        """Tests for `Database.ensureTableExists` and
+        `Database.getExistingTable`.
+        """
+        # Need to start with the static schema.
+        newDatabase = self.makeEmptyDatabase()
+        with newDatabase.declareStaticTables(create=True) as context:
+            context.addTableTuple(STATIC_TABLE_SPECS)
+        # Try to ensure the dyamic table exists in a read-only version of that
```
andy-slac, line 147: `dyamic`

python/lsst/daf/butler/registry/tests/database.py
```diff
@@ -0,0 +1,387 @@
...
+    def testSchemaSeparation(self):
+        """Test that creating two different `Database` instances allows us
+        to create different tables with the same name in each.
+        """
+        db1 = self.makeEmptyDatabase(origin=1)
+        with db1.declareStaticTables(create=True) as context:
+            tables = context.addTableTuple(STATIC_TABLE_SPECS)
+        self.checkStaticSchema(tables)
+
+        db2 = self.makeEmptyDatabase(origin=2)
```
andy-slac, line 196: Should db1 and db2 use different `namespace` parameter rather than `origin`?

python/lsst/daf/butler/registry/tests/database.py
```diff
@@ -0,0 +1,387 @@
...
+        self.assertEqual([dict(r) for r in db.query(tables.a.select()).fetchall()], [row])
+        # Insert multiple autoincrement rows but do not try to get the IDs
+        # back immediately.
+        db.insert(tables.b, {"name": "b1", "value": 10}, {"name": "b2", "value": 20})
+        results = [dict(r) for r in db.query(tables.b.select().order_by("id")).fetchall()]
+        self.assertEqual(len(results), 2)
+        for row in results:
+            self.assertIn(row["name"], ("b1", "b2"))
+            self.assertIsInstance(row["id"], int)
+        self.assertEqual(results[1]["id"], results[0]["id"] + 1)
```
andy-slac: I'm not sure this is guaranteed to be always true, it can break for Oracle RAC for example.

python/lsst/daf/butler/registry/nameShrinker.py
```diff
@@ -0,0 +1,70 @@
...
+        return shrunk
+
+    def expand(self, shrunk: str) -> str:
+        """Return the original name that was passed to a previous call to
+        `shrink`.
+
+        If the given name was not passed to `shrink` or was not modified by
+        it, it is returned unmodified.
+        """
+        return self._names.get(shrunk, shrunk)
```
andy-slac, line 70: I guess functionality of this method depends on client calling `shrink()` first. Can this requirement be enforced/checked to avoid nasty surprises?

python/lsst/daf/butler/registry/databases/sqlite.py
```diff
@@ -0,0 +1,217 @@
...
+    def _convertFieldSpec(self, table: str, spec: ddl.FieldSpec, metadata: sqlalchemy.MetaData,
+                          **kwds) -> sqlalchemy.schema.Column:
+        if spec.autoincrement:
+            if not spec.primaryKey:
+                raise RuntimeError(f"Autoincrement field {table}.{spec.name} that is not a "
+                                   f"primary key is not supported.")
+            if spec.dtype != sqlalchemy.Integer:
+                # SQLite's autoincrement is really limited; it only works if
+                # the column type is exactly "INTEGER".  But it also doesn't
+                # doesn't care about the distinctions between different integer
```
andy-slac: doesn't doesn't

python/lsst/daf/butler/registry/databases/sqlite.py
```diff
@@ -0,0 +1,217 @@
...
+            # insert into this single-column table and use those IDs.
+            if not rows:
+                return [] if returnIds else None
+            with self.transaction():
+                newRows = []
+                ids = []
+                for row in rows:
+                    newRow = row.copy()
+                    id = self._connection.execute(autoincr.table.insert()).inserted_primary_key[0]
+                    newRow[autoincr.column] = id
```
andy-slac: What if client provides an explicit value for auto-increment column?

python/lsst/daf/butler/registry/databases/oracle.py
```diff
@@ -0,0 +1,198 @@
...
+        An integer ID that should be used as the default for any datasets,
+        quanta, or other entities that use a (autoincrement, origin) compound
+        primary key.
+    uri : `str`, optional
+        Database connection string.  If not provided, ``engine`` must be.
+    engine : `sqlalchemy.engine.Engine`, optional
+        SQLAlchemy engine instance to use directly.
+    namespace : `str`, optional
+        Namespace (schema) for all tables used by this database.  May be,
+        `None` which will use the schema implied by ``uri`` or ``enginre``
```
andy-slac: `enginre`

tests/test_sqlite.py
```diff
@@ -0,0 +1,65 @@
...
+        self.root = tempfile.mkdtemp(dir=TESTDIR)
+        self.counter = 1
+
+    def tearDown(self):
+        if self.root is not None and os.path.exists(self.root):
+            shutil.rmtree(self.root, ignore_errors=True)
+
+    def makeEmptyDatabase(self, origin: int = 0) -> SqliteDatabase:
+        filename = os.path.join(self.root, f"db-{self.counter}.sqlite3")
+        self.counter += 1
```
andy-slac: I agree with Dino that counter-based unique name may be an issue, can you use some random names for that (tempfile.mkstemp)?

tests/test_sqlite.py
```diff
@@ -0,0 +1,65 @@
...
+import os.path
+import shutil
+import stat
+import tempfile
+import unittest
+
+from lsst.daf.butler.registry.databases.sqlite import SqliteDatabase
+from lsst.daf.butler.registry.tests import DatabaseTests
+
+TESTDIR = os.path.abspath(os.path.dirname(__file__))
```
andy-slac, line 36: I'd probably make this directory name unique too, with some random stuff in it.

tests/test_oracle.py
```diff
@@ -0,0 +1,146 @@
...
+        cursor.close()
+
+    def setUp(self):
+        cls = type(self)
+        cls._testCounter += 1
+        self._dbCounter = 1
+
+    def makeEmptyDatabase(self, origin: int = 0) -> OracleDatabase:
+        prefix = f"test_{self._token}_{self._testCounter}_{self._dbCounter}_"
+        self._dbCounter += 1
```
andy-slac: Same here, I think there is a chance for collisions depending on how pytest runs. Maybe some random string would work better here?

andy-slac (approved):
Looks good in general, small bunch of comments, most of them minor. There is a lot of new code and there is a chance I missed something, you can blame me for that later 🙂

python/lsst/daf/butler/registry/nameShrinker.py
TallJimbo: My intent was to make all of the files in the `databases` package single-class modules that define a concrete `Database` subclass.  Sounds like maybe I should add a README or something to that directory so that's not just in my head, at least.

python/lsst/daf/butler/registry/ddl.py
TallJimbo: Will do.  It's historical; over in the prototype version (where this is still in `daf/butler/core/schema.py`) we also do `from sqlalchemy.schema import Sequence`, but I figured just importing sqlalchemy without lifting its symbols was usually more readable here.

python/lsst/daf/butler/registry/ddl.py
TallJimbo: The idea was that this class is often used to communicate to `Database` and other classes what columns their users expect to be able to query.  The `Database` (or whatever is given the `TableSpec`) is then free to implement that query interface via a true table, a view, or even just Python/SQLAlchemy object that returns a query that could be used as a subquery (essentially a view implemented in Python instead in the database).

Since that's really about how this class is used by other classes, maybe I should just remove most of that discussion from these docs.

python/lsst/daf/butler/registry/interfaces/database.py
TallJimbo: No plans right now.  One issue I know of is that SQLite actually requires us to change `BigInteger` and `SmallInteger` to just `Integer` in order for autoincrement to work, but then it doesn't actually care how big the numbers are that we put in those fields, so checking for exact types there would not work.  Even for databases with more strict typing, many database types will map to the same Python type, so checking at this point naively might break some otherwise-backwards-compatible migrations.  And the real goal here is to catch cases where the Python code is out of sync with what's in the database, and I think that's unlikely to just involve just type changes.

python/lsst/daf/butler/registry/interfaces/database.py
TallJimbo: Ideally, yes, this would be `NamedTuple[ddl.TableSpec, ...]` (I just realized I forgot the `...` above).  But `typing.NamedTuple` apparently wasn't intended to be used for this ([the docs](https://docs.python.org/3/library/typing.html#typing.NamedTuple) only show usage in defining a new `NamedTuple` class), and it doesn't work in mypy.  I guess that's because `collections.namedtuple` is just a type factory, not a base class or even a metaclass.

I'll add a note about this to the docs.

python/lsst/daf/butler/registry/interfaces/database.py
TallJimbo: The idea was that these would be different `Database` instances in the same multi-layer `Registry`, each pointing to a different schema in the same server.  I think there sharing a connection is desirable.  It's definitely not okay for different `Butler` instances to share a connection.

python/lsst/daf/butler/registry/interfaces/database.py
TallJimbo: I think this already does what we want - it only drops tables known to the `MetaData` object, and that can only be the tables we've created since the start of the `with db.declareStaticTables` context, since we create the `MetaData` object at the start of that.

python/lsst/daf/butler/registry/interfaces/database.py
TallJimbo: I [pinged slack](https://lsstc.slack.com/archives/C2JPL8U78/p1576168318008800) about what kinds of exceptions to make these.  I think I'm going to replace `TransactionInterruption` with just using `assert`, and I think I might generalize `SynchronizationConflict` to `DatabaseConflictError`, since from the user perspective inconsistent row data and inconsistent DDL for a dynamic table aren't meaningfully different.

python/lsst/daf/butler/registry/interfaces/database.py
TallJimbo: :man_shrugging: I've been putting them here just because I've been sold on the idea that it makes sense for `__init__` to go at the top so it immediately follows the class docstring, and then I like keeping other methods with it.  I had assumed there was no convention or strong precedent one way or the other (especially for these, which are formally class attributes but really exist just as a place to hang instance attribute documentation).

python/lsst/daf/butler/registry/databases/sqlite.py
TallJimbo: (see previous comment thread on this subject)

tests/test_sqlite.py
TallJimbo: A random part is added below.

python/lsst/daf/butler/registry/nameShrinker.py
TallJimbo: Only if we also remembered the names of any already-short-enough named that was passed to `shrink`.  This method is really only intended to help with debugging error messages at a pretty low level, and at least right now I don't think it's worth it.

tests/test_postgresql.py
TallJimbo: I'm pretty sure pytest would also have to run `setUpClass` in each process or `cls.server` (etc) won't exist sometimes and the tests wouldn't run at all (unless it's pickling the `TestCase`, I guess - I hope it isn't).  In any case, it's easy to change to something random here, and I'll just do that anyway.

python/lsst/daf/butler/registry/interfaces/database.py
TallJimbo: I think it works fine as is; it's a method on the same class, not a parameter.  I will check later when `Database` is used by `Registry` and gets added to the Sphinx build (I don't think that makes sense quite yet).

python/lsst/daf/butler/registry/interfaces/database.py
TallJimbo: As you probably guessed, I used to have a `values` parameter.  I removed it because I didn't need it, and couldn't even easily use it - SQLAlchemy automatically creates implicit `bindparams` with the names of all of the columns in the `VALUES` clause for which keys exist in the dictionaries you pass later.  I wish it didn't because then I could be confident that logic was only happening once, instead of once per row, but there doesn't seem to be a way around it.

Anyhow, given that, I think the right fix here is to change the docs to say,
> The keys in the dictionaries are either the names of columns to be updated or values in the ``keys`` argument.

(and maybe I should rename the `keys` parameter).

python/lsst/daf/butler/registry/interfaces/database.py
TallJimbo: I think `update().values()` (with no args) is equivalent to just `update()`.  Same with `insert()` and `insert().values()`.

python/lsst/daf/butler/registry/databases/sqlite.py
TallJimbo: I think it helps explain why this code is here (it and the code is copied from something @andy-slac wrote earlier), but maybe it's better as a code comment than a docstring.

python/lsst/daf/butler/registry/interfaces/database.py
TallJimbo: I didn't think so, but you might know better than I do - `check` only executes a `SELECT` query, and I didn't think there was any point to putting that in a transaction (except maybe in SQLite when using super-aggressive locking, but I don't want to rely on that going forward, and I don't think we need to).

python/lsst/daf/butler/registry/interfaces/database.py
TallJimbo: Line shouldn't have even been affected by this commit; must have hit some bad keystrokes in my editor.

python/lsst/daf/butler/registry/databases/sqlite.py
TallJimbo: I'll go with `_AutoincrementCompoundKeyWorkaround`, because I don't want to imply that it *is* the compound key.

python/lsst/daf/butler/registry/databases/sqlite.py
TallJimbo: Glad you caught this.  I was planning to go in and add extra logic for this but forgot (that's why the check for no rows is there).  To handle this properly, we need to place some restrictions on which this workaround applies (which is okay, given what we need):
 - compound primary keys that include an autoincrement value must have exactly one other field in the primary key called "origin"
 - we can only generate autoincrement values when `origin == self.origin`.

There's quite a bit more new code involved in this change now, and I'd appreciate it if you could take another look. It's currently on new commit, but I'll squash it later.

python/lsst/daf/butler/registry/tests/database.py
TallJimbo: I'll weaken it to just `self.assertGreater(results[1]["id"], results[0]["id"])`.  Is that safe?

TallJimbo (commented):
I think I've addressed all review comments (and rebased).  Lots of replies should be appearing simultaneously.

python/lsst/daf/butler/registry/interfaces/database.py
DinoBektesevic: I was asking just in case I make a PR again.... Thanks.

python/lsst/daf/butler/registry/interfaces/database.py
andy-slac: Still, if the same connection instance is shared between two databases then there is a potential that state of one Database instance (`_transactions` list) may not reflect actual state of the connection which was modified by other Database instance. Maybe instead of tracking transactions in `_transactions` list you could just check whether connection is already in transaction?

python/lsst/daf/butler/registry/interfaces/database.py
andy-slac: Hmm, I though that metadata `create_all` just ignores re-existing tables. So if there were some table in the database and in metadata, `drop_all` would remove those even if they existed before. I guess this is probably not an issue as we either should have all tables in a database or none.

python/lsst/daf/butler/registry/interfaces/database.py
andy-slac: Well, this code should execute outside any transaction (code above uses interrupting=True), and if you don't manage transactions yourself then it is probably sqlalchemy auto-commit feature that handles transactions for you (https://docs.sqlalchemy.org/en/13/core/connections.html#understanding-autocommit). I think it's better to manage transactions explicitly, especially if you tracking active transactions in your class.

python/lsst/daf/butler/registry/interfaces/database.py
andy-slac: Sorry, my confusion, I somehow thought `isWriteable` is a parameter to a method, please ignore.

python/lsst/daf/butler/registry/interfaces/database.py
andy-slac: I did not know that implicit bindparam, thanks for explainig!

python/lsst/daf/butler/registry/tests/database.py
andy-slac: I think it's safer and should work OK with current set of backends that we use.

python/lsst/daf/butler/registry/databases/sqlite.py
andy-slac: Will look at it later today or tomorrow.

andy-slac: It looks OK to me though I think there is a potential for confusion here. I myself do not understand very well how autoincrement is supposed to work in case of composite key, maybe it's a general problem and not an issue with our implementation. As an example suppose we have a table with `PK=(auto, origin)` and we insert a row `(origin=1)` first which generates `auto=1`. Next we insert explicit `(auto=1, origin=2)` which I think is a valid operation and should work in general, but this will cause issues with our implementation because autoincr table already has `id=1`. I do not know how to solve this, maybe removing a row from autoincr table before insert (or checking that id is not there, which should be cheaper) would work?

TallJimbo: The code on the new commit (59e7189b05239c067e5cdf32465de1dfde9c2867) tries to handle exactly this case, by inserting into the autoincr table only when `origin == self.origin`.  That means your example will work if and only if `self.origin==1`, but we prohibit the caller from inserting e.g. just (`origin=2)` without an explicit `auto` value.  And that limitation is okay for all of the use cases I have identified.

Of course, that's for SQLite, where this is all manually implemented.  I don't know what the behavior is supposed to be for other databases, but PostgreSQL and Oracle pass the tests I have so far, and those (now) cover this case.  My guess is that they will just skip generating some `auto` values that would have been safe, but that's fine.

python/lsst/daf/butler/registry/tests/database.py
TallJimbo: Each call to `makeEmptyDatabase` is guaranteed to use a new (randomly-generated) `namespace` value.  So using a different `origin` here really doesn't do anything, but it does reflect how we would do things in practice.

python/lsst/daf/butler/registry/interfaces/database.py
TallJimbo: Hmm, that's a good point.  I still think that this probably would be easy to avoid in the future `Registry` code that I expect to be the only place shared connections are used (other than unit tests), but this does feel like a problem in the interface if we have to rely on code outside of `Database` behaving a certain way in order for its behavior to be correct.  I'll see if I can guarantee that the `_transactions` list is shared between `Databases` if and only if the `_connection` is.

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -0,0 +1,1034 @@
...
+        -------
+        db : `Database`
+            A new `Database` instance.
+        """
+        return cls.fromConnectionStruct(cls.connect(uri, writeable=writeable), origin=origin,
+                                        writeable=writeable)
+
+    @classmethod
+    @abstractmethod
+    def connect(cls, uri: str, *, writeable: bool = True) -> ConnectionStruct:
```
DinoBektesevic: You guys move too fast, while I'm writing a review 5 new commits are made :). 

It used to be that one would instantiate a Registry from a RegistryConfig. It used to be that it was the job of RegistryConfig's [connectionString](https://github.com/lsst/daf_butler/blob/master/python/lsst/daf/butler/core/registryConfig.py#L68) property to resolve the full connection URI from the YAML file, supplementing usernames, ports, password etc. from the `~/.lsst/db-auth.yaml` file as, if at all, necessary based on the db string provided in `butler.yaml`. 

Would it be too much to ask that `connect` optionally accepts a direct URI or an `RegistryConfig` instance? That way one can still depend on the `connectionString` property to resolve, or not, users and passwords. 

I understand, from the [architecture design doc](https://confluence.lsstcorp.org/display/DM/Architectural+Prototype+for+the+New+Gen3+Registry), that `Registry` has to go but it wasn't clear to me if that's the case for `RegistryConfig` also.

TallJimbo: Heh, you're moving too fast, commenting on my commits while I'm still amending them :smiley: .

Anyhow, I was mostly planning to tackle this in a future ticket (when I make `Registry` start to use `Database`), but I am certainly aiming to keep the high-level behavior the same.  I think I'd lean towards leaving these APIs as-is, but use `RegistryConfig` to generate the final connection string to pass to them.  It'd also be responsible for inferring the `Database` subclass (instead of the `Registry` subclass) from what's in the config.  Does that sound like it would work?

python/lsst/daf/butler/registry/interfaces/database.py
TallJimbo: I've pushed one more big commit, which adds some factory `classmethods` and a new `dataclass` that ensures the engine, connection and list of transactions is passed together.  It's probably still possible to misuse, but I think it's now a lot harder to *accidentally* misuse, and the checks for whether a transaction is active should now be safe.

python/lsst/daf/butler/registry/interfaces/database.py
DinoBektesevic: That sounds perfect, in fact that sounds to work out exactly like it is right now. The correct registry class is inferred from the `RegistryConfig.getRegistryClass` based on  the mapping provided in `config/registry.yaml`(for future reference).

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -135,6 +137,36 @@ def addTableTuple(self, specs: Tuple[ddl.TableSpec, ...]) -> Tuple[sqlalchemy.sc
...
+    server.
+    """
+
+    connection: sqlalchemy.engine.Connection
+    """A SQLAlchemy object representing the connection itself.
+    """
+
+    transactions: List[sqlalchemy.engine.Transaction]
+    """A list of active transactions for this connection; an empty list
+    indicates that no transaction is active.
```
andy-slac: I'm looking at how `transactions` is used and the only pattern that I see is checking for non-empty list to decide whether any transaction is in progress. If this is the only thing that you need from this list then I'd probably vote for removing it entirely and using `Connection.in_transaction()` for that test. 
Even with all current fixes I think that logic for handling `transactions` list is not completely water-tight, see comment below.

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -208,23 +332,23 @@ def transaction(self, *, savepoint: bool = False, interrupting: bool = False) ->
...
         else:
-            trans = self._connection.begin()
-        self._transactions.append(trans)
+            trans = self._cs.connection.begin()
+        self._cs.transactions.append(trans)
         try:
             yield
             trans.commit()
         except BaseException:
             trans.rollback()
```
andy-slac, line 314: If non-savepoint nested transaction is aborted then it aborts actual database transaction (meaning _all_ nested transactions are aborted too), so you also need to clear `self._cs.transactions`, not just `pop()` latest one.

python/lsst/daf/butler/registry/databases/oracle.py
```diff
@@ -93,61 +93,55 @@ class OracleDatabase(Database):
...
-                raise RuntimeError("'namespace' and 'uri' arguments specify different schemas.")
+        cs.engine.dialect.max_identifier_length = 128
+        return cs
+
+    @classmethod
+    def fromConnectionStruct(cls, cs: ConnectionStruct, *, origin: int, namespace: Optional[str] = None,
+                             writeable: bool = True) -> Database:
         # TODO: is there anything we can do to make the connection read-only if
         # writeable is False?  If not (and at present) we just rely on the
         # Python code being well-behaved and not *trying* to make changes.
```
andy-slac: Oracle has `SET TRANSACTION READ ONLY` syntax (https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/SET-TRANSACTION.html), I think this needs to be run as a first command in a transaction. It not only disables update commands but sets what is called "transaction-level read consistency" which means that all SELECTs view the data from the same snapshot (made at a time of transaction start).

 Postgres supports the same `SET TRANSACTION READ ONLY` (https://www.postgresql.org/docs/12/sql-set-transaction.html) but I could not find what exactly it does for that option.

python/lsst/daf/butler/registry/databases/postgresql.py
```diff
@@ -55,26 +56,36 @@ class PostgresqlDatabase(Database):
...
-        self._shrinker = NameShrinker(engine.dialect.max_identifier_length)
+        self._shrinker = NameShrinker(cs.engine.dialect.max_identifier_length)
+
+    @classmethod
+    def connect(cls, uri: str, *, writeable: bool = True) -> ConnectionStruct:
+        cs = ConnectionStruct(sqlalchemy.engine.create_engine(uri, pool_size=1))
+        if not writeable:
+            dbapi = cs.engine.raw_connection()
+            try:
+                dbapi.set_session(readonly=True)
```
andy-slac: I wonder if this works 100% reliably. I understand that connections are cached in a pool, but it may also happen that pool can decide to drop and re-establish a connection under some circumstances, and new connection will not have `readonly` option set. It may be better to pass a `creator` argument to `create_engine` or use `SET TRANSACTION READ ONLY` when transaction is started.

python/lsst/daf/butler/registry/databases/sqlite.py
```diff
@@ -106,34 +107,133 @@ class SqliteDatabase(Database):
...
-            target += '?mode=ro&uri=true'
+    def __init__(self, *, cs: ConnectionStruct, origin: int, namespace: Optional[str] = None,
+                 writeable: bool = True):
+        super().__init__(origin=origin, cs=cs, namespace=namespace)
+        # Get the filename from a call to 'PRAGMA database_list'.
+        with closing(cs.connection.connection.cursor()) as cursor:
+            dbList = list(cursor.execute("PRAGMA database_list").fetchall())
+        if len(dbList) == 0:
+            raise RuntimeError("No database in connection.")
+        for _, dbname, filename in dbList:
```
andy-slac, line 141: I think you can simplify conditional logic below if you do `namespace = "main" if namespace is None else namespace`

andy-slac (approved):
Few minor (or not so minor) comments for this commit.

python/lsst/daf/butler/registry/interfaces/database.py
andy-slac: I think larger issue here is that even when you are in a savepoint transaction the underlying engine can decide to abort top-level transaction for any reason (e.g. Postgres aborts transactions for any error). In that case you somehow need to know what is the actual state of a backend transaction is and keep your `transactions` list consistent.

TallJimbo: Hmm, good point.  And I think that means that non-nested transactions are just fundamentally incompatible with Python context managers, at least in terms of yielding non-surprising behavior.  I think I'm just going to remove the `savepoint` argument and just go back to always using `begin_nested` for now (as we do in `SqlRegistry`).  Hopefully just using bulk operations will still keep the number of actual SAVEPOINT commands low, but if not I think we need to look into some other way of expressing transactions in Python.

python/lsst/daf/butler/registry/databases/oracle.py
TallJimbo: I'll try to find SQLAlchemy event hooks where we can put those.

python/lsst/daf/butler/registry/interfaces/database.py
TallJimbo: Hopefully switching to `Connection.in_transaction()` will take care of that.  If not, I think we'll have to see how the problem manifests to be able to do something about it.

python/lsst/daf/butler/registry/databases/oracle.py
TallJimbo: So, the event hooks exist, but there's no good way to get the information about whether the DB is read-only into them.  So instead I've just emitted that `SET TRANSACTION READ ONLY` statement at the beginning of transactions we start ourselves.  That by no means covers everything, but this is really just an extra guard against bad code, not the first line of defense.

TallJimbo (commented):
@andy-slac, up for one more round?  I'm now just using `Connection.in_transaction` (wish I'd realized that existed earlier), which I think addresses your biggest concern about as well as we can.  Combined with another change I discovered I needed to make on DM-22609 downstream of this, I realized I could just replace `ConnectionStruct` with directly using the SQLAlchemy `Connection` instance, since we don't need the engine either.  Six new small commits, some of which I'll squash later.

python/lsst/daf/butler/registry/interfaces/database.py
```diff
@@ -330,9 +326,11 @@ def transaction(self, *, savepoint: bool = False, interrupting: bool = False) ->
             "Logic error in transaction nesting: an operation that would "
             "interrupt the active transaction context has been requested."
         )
-        if savepoint:
+        if self._cs.connection.in_transaction():
```
andy-slac: I guess this means we treat all nested transactions as savepoints. Savepoints have bigger overhead compared to non-savepoint "nested" sqlalchemy transactions, if we have a lot of nested transaction in registry this may cause performance issues.

TallJimbo: Yup, understood.  As this is already a problem in the current `SqlRegistry`, I'm just punting the problem to a later ticket (I have some other mitigations in mind) as it's too tricky to solve quickly on this one.

andy-slac:
I checked recent commits, everything looks OK and waits to be battle-tested 🙂

DinoBektesevic (approved):

PR #213, opened 2019-12-10
DM-22481: Fix deletion of file on formatter failure

timj:
Previously if the Formatter wrote a partial file that file would not be cleaned up.

### DIFF
--- a/python/lsst/daf/butler/core/formatter.py
+++ b/python/lsst/daf/butler/core/formatter.py
@@ -23,6 +23,7 @@ __all__ = ("Formatter", "FormatterFactory")
 
 from abc import ABCMeta, abstractmethod
 import logging
+import copy
 from typing import ClassVar, Set, FrozenSet, Union, Optional, Dict, Any, Tuple, Type
 
 from .configSupport import processLookupConfigs, LookupKey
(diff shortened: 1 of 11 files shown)


python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -238,6 +238,11 @@ def put(self, inMemoryDataset, ref):
                                         Filename=tmpFile.name)
                 log.debug("Wrote file to %s via a temporary directory.", location.uri)
 
+        # Register a callback to try to delete the uploaded data if
+        # the ingest fails below
+        self._transaction.registerUndo("write", self.client.delete_object,
```
timj, line 243: @DinoBektesevic I've added this to the S3Datastore. Whether this is in the right place or not depends on whether upload_file or put_object are themselves transactional. If they fail and the object is effectively not in S3 then this is the right place. If they can partially work then instead of registerUndo we will have to do what we do in posixDatastore (use undoWith, catch the exception, reraise).

python/lsst/daf/butler/core/formatter.py
```diff
@@ -158,7 +159,34 @@ def toBytes(self, inMemoryDataset: Any) -> bytes:
...
+        ------
+        NotImplementedError
+            Raised if there is no ``extension`` attribute associated with
+            this formatter.
+        """
+        location = copy.deepcopy(location)
+        try:
+            location.updateExtension(cls.extension)
+        except AttributeError:
+            raise NotImplementedError("No file extension registered with this formatter")
```
TallJimbo: Use `raise NotImplementedError(...) from None` here if you want to avoid "while handing this exception..." noise in the traceback.

TallJimbo (approved):
Looks great.  Only one minor comment.

python/lsst/daf/butler/datastores/s3Datastore.py
DinoBektesevic: This will work. S3 API docs claim never to [add partial files](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html) which is what boto3 eventually calls for both object and file uploads.

In my Jira comment, I was actually concerned about the upload_file calls in the `_ExtractIngestInfo` functionality. That bit, in S3Datastore, has no undo transactions protections, but definitely should. However, on move actions it would require me to delete the files in the original location (whether local disk or an different S3 Bucket). For S3 Bucket then it becomes difficult to confirm file actually got deleted because the `delete_object` calls perpetually return HTML 204 response.

DinoBektesevic (approved):

python/lsst/daf/butler/datastores/s3Datastore.py
timj: Ok. I think my line does make things better and won't trigger any problems in the ingest line because transfer=None. I can leave the transaction handling for `ingest` to another time.

PR #215, opened 2019-12-17
DM-22609: partial rewrite of Registry on top of new Database interface

TallJimbo:
There are a lot of changes here, but a lot of it is just moving stuff around.  I've tried to clearly indicate on any commits that involve copying or moving code what's actually new; *please* only look at this commit-by-commit, and read the commit messages as you go.

It may also be helpful for reviewers to note that the commits mostly fall into three groups, and if you look at commits in the right order (which is not always how GitHub orders them, but I think some rebasing fixed that for now), the groups are pretty clear:

- Fix/improve existing code to get it ready for the new `Registry` implementation and slightly increased test coverage.  This is mostly in the new `Database` classes that were added on DM-22508 (#212), as this ticket is the first place those are used outside of tests.  This is 3d1083420734c6acf8719390f5578b8d1b519b4c - bdda46078a154b9d6960cb758b60c75e9bb43d1f.

- Copy `Registry` from `core/registry.py` into `registry/registry.py`, and then start copying functionality from `SqlRegistry` into it while coping tests from (mostly) `test_sqlRegistry.py` into `registry/tests/registry.py`.  The `Registry` method implementations change various amounts to make use of the new `Database` interface, but for the most part those changes are trivial.  At the end of this, we switch `Butler` over to the new `Registry.  This is - 9ac5fad0667fd7934d1c36d73e0484e9e81aa225 - f80c8d3a77a00a3fe1d71d62540e5bbc784c0f06.

- Delete the old `Registry` classes, merge the duplicate code in `schema.py` and `ddl.py` back together, and move code out of `core` to clean up the internal dependency relationships.  Contrary to earlier slack discussoins, I did not extend this to moving `Datastore`-only code out of `core`; I ran into the problem that led to [RFC-653](https://jira.lsstcorp.org/browse/RFC-653) while moving the `Registry`-only code, and that created enough doubt about how to name the reorganized files that I figured it'd be better to defer that (I did go with the RFC-653 proposal for what I did move, as that's inconsistent with our current conventions and I liked how it worked). cc67817ddfc4308244cced69ce7b71f427b25b1a - b1b68659d39ca5472cf080faa9627ddc95885b2c.


python/lsst/daf/butler/_butler.py
```diff
@@ -817,7 +821,8 @@ def import_(self, *, directory: typing.Optional[str] = None,
             filename = os.path.join(directory, filename)
         BackendClass = getClassOf(self.config["repo_transfer_formats"][format]["import"])
         with open(filename, 'r') as stream:
-            backend = BackendClass(stream)
+            backend = BackendClass(stream, self.registry)
```
timj, line 824: BackendClass takes a registry in the constructor but then you have to register that registry after you've constructed?

python/lsst/daf/butler/core/ddl.py
```diff
@@ -20,22 +20,19 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 from __future__ import annotations
 
-__all__ = ("SchemaConfig", "Schema", "TableSpec", "FieldSpec", "ForeignKeySpec")
```
timj, line 23: Can there be a docstring somewhere at the top saying what this file is for? I had to look up what DDL meant.

python/lsst/daf/butler/core/repoTransfers.py
```diff
@@ -349,46 +361,87 @@ class YamlRepoImportBackend(RepoImportBackend):
...
+                        DatasetRef(datasetType, d["data_id"], run=data["run"], id=d["dataset_id"]),
+                        formatter=doImport(d["formatter"])
+                    ),
+                    d.get("collections", [])
+                )
+                for d in data["records"]
+            )
+
+    def register(self, registry: Registry):
+        for run in self.runs:
```
timj, line 420: Docstring?

python/lsst/daf/butler/registry/_config.py
```diff
@@ -42,27 +51,41 @@ def getDialect(self):
...
-        Otherwise the `db` key is parsed and the correct class is determined
-        from a list of aliases found under `clsMap` key of the registry config.
+        The appropriate class is determined by parsing the `db` key to extract
+        the dialect, and then looking that up under the `engines` key of the
+        registry config.
+        """
+        dialect = self.getDialect()
+        if dialect not in self["engines"]:
+            raise ValueError(f"Connection string dialect has no known aliases. Received: {dialect}")
+        databaseClass = self.get(("engines", dialect))
```
timj: Maybe use:
```python
databaseClass = self["engines", dialect]
```
since you don't want the side effect of returning `None` by mistake.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -0,0 +1,1330 @@
...
+            ).fetchall()
+            dimensions = DimensionGraph(self.dimensions, names=(r[0] for r in result) if result else ())
+            datasetType = DatasetType(name=name,
+                                      storageClass=storageClass,
+                                      dimensions=dimensions)
+            self._datasetTypes[name] = datasetType
+        return datasetType
+
+    def getAllDatasetTypes(self) -> FrozenSet[DatasetType]:
+        r"""Get every registered `DatasetType`.
```
timj: No longer needs to be a raw string.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -0,0 +1,1330 @@
...
+        except sqlalchemy.exc.IntegrityError as err:
+            raise ConflictingDefinitionError(
+                f"Constraint violation while associating datasets with collection {collection}. "
+                f"This probably means that one or more datasets with the same dataset type and data ID "
+                f"already exist in the collection, but it may also indicate that the datasets do not exist."
+            ) from err
+
+    @transactional
+    def disassociate(self, collection: str, refs: List[DatasetRef]):
+        r"""Remove existing Datasets from a collection.
```
timj: No need for raw.

timj (approved):
Looks good to me. The reorganization is fine. I can get used to the weird underscores in the file names (I see matplotlib does that). Minor style comments.

python/lsst/daf/butler/_butler.py
TallJimbo: Ah, good catch.  I'll remove the `registry` argument from the later methods.  I'd still like to keep `register` as a separate method, because the constructor only does reads and that does writes.

PR #216, opened 2019-12-18
DM-22677: Relocate scripts to python module

timj:


python/lsst/daf/butler/script/dumpButlerConfig.py
```diff
@@ -0,0 +1,100 @@
...
+    if subset is not None:
+        config = config[subset]
+
+    try:
+        config.dump(outfh)
+    except AttributeError:
+        print(config, file=outfh)
+
+
+def main():
```
TallJimbo, line 89: Would it be worthwhile having this take an `args` to pass to `parse_args()` that would take `sys.argv[1:]` or similar?  That seems like it's a missing piece of making this more usable, but I also don't really a scenario where we'd actually use that.

python/lsst/daf/butler/script/dumpButlerConfig.py
```diff
@@ -0,0 +1,100 @@
...
+    args = build_argparser().parse_args()
+
+    if args.verbose:
+        logging.basicConfig(level=logging.DEBUG)
+
+    try:
+        dumpButlerConfig(args.root, args.searchpath, args.subset, sys.stdout)
+    except Exception as e:
+        print(f"{e}", file=sys.stderr)
```
TallJimbo: I'm not sure I like this traceback squashing here. In a really mature program that knows about all of its common failure modes, that can give you more friendly error messages, but in code that's under development, so often it just hides what went wrong and forces you to run again, which I find super frustrating.

python/lsst/daf/butler/core/repoTransfers.py
```diff
@@ -165,8 +165,8 @@ def saveDatasets(self, refs: Iterable[DatasetRef], *,
             path generated by the datastore.  If `None`, the `FileDataset`
             returned by `Datastore.export` will be used directly.
 
-        Note
-        ----
+        Notes
+        -----
```
TallJimbo, line 169: I think I've also got this fix on a branch for you to review :slightly_smiling_face: .

TallJimbo (approved):
Two minor comments on one script that probably apply to all of them.

python/lsst/daf/butler/script/dumpButlerConfig.py
timj: I believe I'm following standard practice here: https://developer.lsst.io/stack/argparse-script-topic-type.html?highlight=autoprogram

python/lsst/daf/butler/script/dumpButlerConfig.py
timj: I got so wrapped up in ensuring that I returned non-zero exit status that I forgot I could just let it die and also get non-zero exit status.

PR #217, opened 2020-01-13
DM-21768: Vectorize dataset insertion in Registry

TallJimbo:


python/lsst/daf/butler/registry/_registry.py
```diff
@@ -657,84 +657,86 @@ def find(self, collection: str, datasetType: Union[DatasetType, str], dataId: Op
         return self._makeDatasetRefFromRow(result, datasetType=datasetType, dataId=dataId)
 
     @transactional
-    def addDataset(self, datasetType: Union[DatasetType, str],
-                   dataId: DataId, run: str, producer: Optional[Quantum] = None,
-                   recursive: bool = False, **kwds: Any) -> DatasetRef:
-        """Adds a Dataset entry to the `Registry`
+    def insertDatasets(self, datasetType: Union[DatasetType, str], dataIds: Iterable[DataId],
+                       run: str, *, producer: Optional[Quantum] = None, recursive: bool = False
+                       ) -> List[DatasetRef]:
```
natelust, line 662: github seems to be showing a darker green rectangle, is it a tab?

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -657,84 +657,86 @@ def find(self, collection: str, datasetType: Union[DatasetType, str], dataId: Op
...
+        """Insert one or more datasets into the `Registry`
```
natelust, line 663: are dataIds themselves iterable? If not, despite the change in names, it might be nice to have this method take either a single dataId or Interable of DataIds, to prevent needing a bunch of trivial [dataId] all over the code. or maybe make a separate interface called InsertDataset that trivial defers to this method?

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -657,84 +657,86 @@ def find(self, collection: str, datasetType: Union[DatasetType, str], dataId: Op
...
+                row[dimension.name] = value
+            rows.append(row)
+        # Actually insert into the dataset table.
+        datasetIds = self._db.insert(self._tables.dataset, *rows, returnIds=True)
+        # Resolve the DatasetRefs with the autoincrement IDs we generated.
+        refs = [ref.resolved(id=datasetId, run=run) for datasetId, ref in zip(datasetIds, refs)]
+        # Associate the datasets with the run as a collection.  Note that we
+        # do this before inserting component datasets so recursing doesn't try
+        # to associate those twice.
+        self.associate(run, refs)
```
natelust, line 718: Im a bit worried this is a non-atomic in so far as the data get inserted but things can get interrupted such that they never end up being associated

python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -852,20 +848,20 @@ def testCalibrationLabelIndirection(self):
...
         registry.registerRun(run1)
         for detector in (1, 2, 3):
-            registry.addDataset(flat, dict(instrument="DummyCam", calibration_label="first_night",
-                                           physical_filter="dummy_i", detector=detector),
-                                run=run1)
-            registry.addDataset(flat, dict(instrument="DummyCam", calibration_label="second_night",
-                                           physical_filter="dummy_i", detector=detector),
-                                run=run1)
+            registry.insertDatasets(flat, [dict(instrument="DummyCam", calibration_label="first_night",
+                                                physical_filter="dummy_i", detector=detector)],
```
natelust, line 852: here (and below with the darker green)

natelust (approved):
Minor thoughts

python/lsst/daf/butler/registry/_registry.py
TallJimbo: I think those are just new spaces.

python/lsst/daf/butler/registry/_registry.py
TallJimbo: Data IDs are indeed iterable (they're `collections.abc.Mapping`s).  Given that the actual high-level APIs are in `Butler`, and one of those (`put`) is already scalar rather than vector, let's see how many other places in the codebase would use `insertDataset` before adding it.

python/lsst/daf/butler/registry/_registry.py
TallJimbo: The `transactional` decorator on `insertDatasets` runs all of this inside a transaction, so it's all definitely atomic.  The failure mode if we had tried to recursively associate later would have just been a lot of unnecessary I/O, as the database replaced a bunch of rows with identical rows.

python/lsst/daf/butler/_butler.py
```diff
@@ -701,14 +703,54 @@ class used.  `FileDataset.path` attributes may be modified to put
...
+        # The actual data structure:
+        groupedData: GroupedData = defaultdict(dict)
+        # And the nested loop that populates it:
         for dataset in datasets:
-            for i, ref in enumerate(dataset.refs):
-                dataset.refs[i] = self.registry.addDataset(ref.datasetType, ref.dataId,
-                                                           run=self.run, recursive=True)
+            resolvedRefs = []
+            for ref in dataset.refs:
+                groupedData[ref.datasetType][ref.dataId] = (dataset, resolvedRefs)
```
natelust, line 747: This will cause all entries in the (nested) dicts to share the same list, which I dont think is what you want

python/lsst/daf/butler/core/repoTransfers.py
```diff
@@ -437,17 +437,36 @@ def load(self, datastore: Datastore, *,
...
+            # resolved DatasetRefs to replace the unresolved ones as we
+            # reorganize the collection information.
+            i = 0
             for fileDataset, collectionsForDataset in records:
-                # For now, we ignore the dataset_id we pulled from the file
-                # and just insert without one to get a new autoincrement value.
-                # Eventually (once we have origin in IDs) we'll preserve them.
-                fileDataset.refs = [self.registry.addDataset(datasetType, ref.dataId, run=run,
-                                                             recursive=True) for ref in fileDataset.refs]
+                n = len(fileDataset.refs)
```
natelust: This is a bit messy, maybe it would be better do build the "n"s in a list while you generate the input to insertDatasets, and then just loop over those accumulated "n"s here

natelust (approved):

python/lsst/daf/butler/_butler.py
TallJimbo: It is what I want, because the list is associated with `dataset`.  I'll add a comment to that effect.

PR #218, opened 2020-01-13
DM-23024: Allow multiple datasets to be associated with a single file on ingest

timj:
The most work was in changing all the code that used FileDataset.ref to instead use FileDataset.refs.

cc/ @parejkoj 

I will now check what other packages need to change because of FileDataset changes.


python/lsst/daf/butler/core/formatter.py
```diff
@@ -60,10 +64,11 @@ class Formatter(metaclass=ABCMeta):
     are supported (`frozenset`).
     """
 
-    def __init__(self, fileDescriptor: FileDescriptor):
+    def __init__(self, fileDescriptor: FileDescriptor, dataId: "DataCoordinate" = None):
```
TallJimbo: If you use `from __future__ import annotations`, you can write `DataCoordinate` without the quotes.

python/lsst/daf/butler/core/formatter.py
```diff
@@ -77,6 +82,11 @@ def fileDescriptor(self) -> FileDescriptor:
         (`FileDescriptor`, read-only)"""
         return self._fileDescriptor
 
+    @property
+    def dataId(self) -> Dict:
+        """DataId associated with this formatter (`dict`)"""
```
TallJimbo: If constructor takes a `DataCoordinate`, that's what the type is going to be here, too, right?

python/lsst/daf/butler/core/repoTransfers.py
```diff
@@ -57,10 +57,11 @@ class RepoTransferFormatConfig(ConfigSubset):
...
     """A struct that represents a dataset exported to a file.
     """
-    __slots__ = ("ref", "path", "formatter")
+    __slots__ = ("refs", "path", "formatter")
 
-    ref: DatasetRef
-    """Registry information about the dataset (`DatasetRef`).
+    refs: List[DatasetRef]
+    """Registry information about the dataset. (`DatasetRef` or `list` of
+    `DatasetRef`).
```
TallJimbo: This is the doc for the attribute, not the ctor arg, so I think the type is just "`list` of `DatasetRef`" here.

TallJimbo (approved):
Looks good aside from some doc/annotation type stuff.  And it seems you've duplicated `butler.py` in addition to modifying `_butler.py` (presumably when resolving a rebase conflict).

timj:
My rebase completely failed to notice `_butler.py` turn up. 👎

python/lsst/daf/butler/core/formatter.py
timj: Yes, I should have gone back and fixed that

python/lsst/daf/butler/core/utils.py
```diff
@@ -54,9 +54,10 @@ def iterable(a):
 
     There are three cases, when the input is:
 
-    - iterable, but not a `str` -> iterate over elements
+    - iterable, but not a `str`  or Mapping -> iterate over elements
       (e.g. ``[i for i in a]``)
     - a `str` -> return single element iterable (e.g. ``[a]``)
+    - a Mapping -> return single element iterable
```
TallJimbo, line 60: This makes me a little nervous, as I think it's hard to remember what makes a class qualify as a `Mapping`.  Could we at least make this not the default behavior?  Or maybe allow general `isinstance` exceptions via an optional kwarg, i.e.:
`iterable(x, except=collections.abc.Mapping)` (though that's not terribly readable either).

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -246,6 +246,22 @@ def getStoredItemInfo(self, ref):
...
                               file_size=record["file_size"])
 
+    def getStoredItemInfoForPath(self, pathInStore):
+        # Docstring inherited from GenericBaseDatastore
+        records = list(self.registry.fetchOpaqueData(self._tableName, path=pathInStore))
+        if len(records) == 0:
+            return records
+
+        for r in records:
+            print(r["path"])
```
TallJimbo: Debug relic.

python/lsst/daf/butler/_butler.py
```diff
@@ -649,14 +648,21 @@ def remove(self, datasetRefOrType, dataId=None, *, delete=True, remember=True, *
...
         ref = self.registry.find(self.collection, datasetType, dataId)
         if delete:
-            for r in itertools.chain([ref], ref.components.values()):
-                # If dataset is a composite, we don't know whether it's the
-                # parent or the components that actually need to be removed,
-                # so try them all and swallow errors.
-                try:
+            # There is a difference between a concrete composite and virtual
+            # composite. In a virtual composite the datastore is never
+            # given the top level DatasetRef. In the concrete composite
```
TallJimbo, line 653: Does this mean there is a difference in behavior at the butler level when we try to delete different kinds of composites?  Or is it just butler's responsibility rather than datastore's to iterate over components in the virtual composite case?

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -246,21 +247,22 @@ def getStoredItemInfo(self, ref):
...
 
-        return [StoredFileInfo(formatter=r["formatter"],
-                               path=r["path"],
-                               storageClass=self.storageClassFactory.getStorageClass(r["storage_class"]),
-                               checksum=r["checksum"],
-                               file_size=r["file_size"])
-                for r in records]
+        Returns
+        -------
+        refs : `set` of `int`
```
TallJimbo: I'd call this something other than "refs" (same with the method name).  Maybe "ids"?

python/lsst/daf/butler/_butler.py
```diff
@@ -660,7 +660,10 @@ def remove(self, datasetRefOrType, dataId=None, *, delete=True, remember=True, *
                 self.datastore.remove(ref)
             elif ref.isComposite():
                 for r in ref.components.values():
-                    self.datastore.remove(r)
+                    # If a dataset was removed previously but remembered
+                    # in registry, skip the removal in the datastore.
+                    if self.registry.getDatasetLocations(r):
+                        self.datastore.remove(r)
```
TallJimbo, line 668: Note that this attempts to remove the dataset from this particular datastore if it is present in any datastore.  I think that's okay for now, but we should open a ticket to address it in the future.  The more I think about it, the more I think we'd be better off if butler had a dict of datastores instead of having `ChainedDatastore`, so it could refer to them by name in this and other places.

TallJimbo (approved):
Looks good; only minor comments, though this increases my unease with `ChainedDatastore` as a way to associate multiple `Datastores` with a butler, and I'd like to address that eventually.

python/lsst/daf/butler/core/utils.py
timj: Do you ever want iterable to return the keys of a Mapping? The problem here is dataIds and compatbility with old YAML files.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
timj: I think you were looking at an old PR.

python/lsst/daf/butler/_butler.py
timj: In one case butler sends the composite ref to the datastore and datastore stores its own entries because it is writing a single file. In the virtual case (I still get these the wrong way round) the butler keeps the top level composite ref to itself and only sends the components to the datastore.  Previously we were doing a "send all refs regardless" approach and mostly things were throwing exceptions because either the ref was never known to butler, or else after the first ref was removed the component refs were already deleted. Now at least the two paths are separate just as in getDirect.

python/lsst/daf/butler/_butler.py
timj: This is fine so long as the datastores used by the registry haven't changed because of a config change. I can filter the returned list by asking for the equivalent from the datastore itself.

timj: I think I've fixed it now so that it does use the datastore name before deciding whether to remove.

python/lsst/daf/butler/core/utils.py
TallJimbo: That's a fair point; I think you're right that this is the only behavior we'd ever actually want for mappings - my concern is more that it's not the behavior I would have naively expected.  But it's not a big deal either way.

I do think Python mappings would be much cleaner if they weren't iterable at all and you had to explicitly say `keys()`, but of course there's nothing we can do about that.

PR #220, opened 2020-01-16
DM-22648: Add DcrModel to Gen3 butler

isullivan:
Should be merged at the same time as DM-23008, and before DM-21917.
The `subfilter` dimension is used for modeling chromatic effects such as Differential Chromatic Refraction in coadds.

### DIFF
--- a/config/datastores/posixDatastore.yaml
+++ b/config/datastores/posixDatastore.yaml
@@ -8,5 +8,5 @@ datastore:
     # valid_first and valid_last here are YYYYMMDD; we assume we'll switch to
     # MJD (DM-15890) before we need more than day resolution, since that's all
     # Gen2 has.
-    default: "{collection}/{datasetType}.{component:?}/{tract:?}/{patch:?}/{label:?}/{abstract_filter:?}/{physical_filter:?}/{visit:?}/{datasetType}_{component:?}_{tract:?}_{patch:?}_{label:?}_{abstract_filter:?}_{physical_filter:?}_{calibration_label:?}_{visit:?}_{exposure:?}_{detector:?}_{instrument:?}_{skymap:?}_{skypix:?}_{run}"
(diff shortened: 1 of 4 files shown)


config/datastores/posixDatastore.yaml
```diff
@@ -8,5 +8,5 @@ datastore:
     # valid_first and valid_last here are YYYYMMDD; we assume we'll switch to
     # MJD (DM-15890) before we need more than day resolution, since that's all
     # Gen2 has.
-    default: "{collection}/{datasetType}.{component:?}/{tract:?}/{patch:?}/{label:?}/{abstract_filter:?}/{physical_filter:?}/{visit:?}/{datasetType}_{component:?}_{tract:?}_{patch:?}_{label:?}_{abstract_filter:?}_{physical_filter:?}_{calibration_label:?}_{visit:?}_{exposure:?}_{detector:?}_{instrument:?}_{skymap:?}_{skypix:?}_{run}"
+    default: "{collection}/{datasetType}.{component:?}/{tract:?}/{patch:?}/{label:?}/{abstract_filter:?}/{subfilter:?}/{physical_filter:?}/{visit:?}/{datasetType}_{component:?}_{tract:?}_{patch:?}_{label:?}_{abstract_filter:?}_{physical_filter:?}_{calibration_label:?}_{visit:?}_{exposure:?}_{detector:?}_{instrument:?}_{skymap:?}_{skypix:?}_{run}"
```
hsinfang, line 11: As you change this do you mind changing the default in `config/datastores/s3Datastore.yaml` as well?  (I know... the duplication is not ideal... :(

isullivan: Thank you for catching this!

config/dimensions.yaml
```diff
@@ -70,6 +70,20 @@ dimensions:
...
+    subfilter:
+      doc: >
+        A mathematical division of an abstract_filter. Subfilters are used to
+        model wavelength-dependent effects such as differential chromatic
+        refraction.
+      keys:
+      -
+        name: name
+        type: string
+        length: 8
```
natelust: So, this is number of characters, if someone has a longish abstract filter name, or I guess if you have a crazy number of sub-bands you could run over this. Might be worth adding a check in the script that adds subbands to make sure the name is under 8 characters

natelust (approved):

config/dimensions.yaml
timj: We already set abstract_filter to 16 so you need to be at least that long (and probably bigger). Or is this string an addendum to abstract_filter that is guaranteed to be < 8?

isullivan: This should typically be abstract_filter plus two more characters.

isullivan: A related question: Is it important here that `length` is a power of 2, or is that just a coincidence throughout this file?

PR #223, opened 2020-01-23
DM-23080: dimension+registry refactoring

TallJimbo:


python/lsst/daf/butler/registry/dimensions/query.py
```diff
@@ -0,0 +1,118 @@
...
+from ...core import DataCoordinate, Dimension, DimensionElement, DimensionRecord, Timespan
+from ...core.dimensions.schema import makeElementTableSpec
+from ...core.utils import NamedKeyDict
+from ..interfaces import Database, DimensionRecordStorage
+from ..queries import QueryBuilder
+
+
+class QueryDimensionRecordStorage(DimensionRecordStorage):
+    """A read-only record storage implementation that aggregates the values of
+    another element for which it is a required or implied dimension.
```
natelust: This comment could be made a little clearer

python/lsst/daf/butler/registry/dimensions/query.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Parameters
+    ----------
+    db : `Database`
+        Interface to the database engine namespace that will hold these
```
natelust: The use of the word namespace here is a little confusing

python/lsst/daf/butler/registry/dimensions/spatial.py
```diff
@@ -0,0 +1,78 @@
...
+    Parameters
+    ----------
+    db : `Database`
+        Interface to the database engine namespace that will hold these
+        dimension records.
+    element : `DimensionElement`
+        The element whose records this storage will manage.
+    table : `sqlalchemy.schema.Table`
+        The logical table for the element.
+    commonSkyPixOverlapTable : `sqlalchemy.schema.Table`, optional
```
natelust, line 81: This does not seem optional in the init method

python/lsst/daf/butler/registry/dimensions/table.py
```diff
@@ -0,0 +1,92 @@
...
+            [self._table.columns[name] for name in RecordClass.__slots__]
+        ).select_from(
+            self._table
+        ).where(sqlalchemy.sql.and_(*whereTerms))
+        row = self._db.query(query).fetchone()
+        if row is None:
+            return None
+        return RecordClass(*row)
+
+    def insert(self, *records: DimensionRecord):
```
natelust, line 116: So records is plural but its type is DimensionRecord. Is the type iterable or is the type hint wrong

python/lsst/daf/butler/registry/interfaces/_dimensions.py
```diff
@@ -25,7 +25,7 @@
 from abc import ABC, abstractmethod
 from typing import Optional
 
-from sqlalchemy.sql import FromClause
+import sqlalchemy
```
natelust, line 28: seems like overkill to import everything. Why was this changed? and or why not just import sqlachemy.sql

python/lsst/daf/butler/registry/dimensions/spatial.py
```diff
@@ -53,10 +56,18 @@ def __init__(self, db: Database, element: DimensionElement, *, table: sqlalchemy
...
-    def getCommonSkyPixOverlapTable(self, dataId: Optional[DataId] = None) -> sqlalchemy.schema.Table:
-        # Docstring inherited from
-        # DimensionRecordStorage.getCommonSkyPixOverlapTable.
-        return self._commonSkyPixOverlapTable
+    def join(
+        self,
+        builder: QueryBuilder, *,
+        regions: Optional[NamedKeyDict[DimensionElement, sqlalchemy.sql.ColumnElement]] = None,
+        timespans: Optional[NamedKeyDict[DimensionElement, Timespan[sqlalchemy.sql.ColumnElement]]] = None,
+    ):
```
natelust, line 114: I may not be following along with code flow, but what if this function gets called before one in which regions are added, should what is being added here get added some how?

python/lsst/daf/butler/registry/dimensions/table.py
```diff
@@ -60,12 +62,25 @@ def clearCaches(self):
...
+    def join(
+        self,
+        builder: QueryBuilder, *,
+        regions: Optional[NamedKeyDict[DimensionElement, sqlalchemy.sql.ColumnElement]] = None,
+        timespans: Optional[NamedKeyDict[DimensionElement, Timespan[sqlalchemy.sql.ColumnElement]]] = None,
+    ):
+        assert regions is None, "This implementation does not handle spatial joins."
+        joinDimensions = list(self.element.graph.required)
+        joinDimensions.extend(self.element.implied)
+        joinOn = builder.startJoin(self._table, joinDimensions, self.element.RecordClass.__slots__)
```
natelust, line 87: Im slightly uncomfortable directly accessing __slots__ here, though I dont think it can hurt....

python/lsst/daf/butler/registry/queries/_builder.py
```diff
@@ -94,50 +101,13 @@ def joinDimensionElement(self, element: DimensionElement):
...
-        if element is self.summary.universe.commonSkyPix:
-            return
-        assert element in self.summary.spatial,\
-            f"{element}'s spatial information is not relevant for this query."
-        table = self._dimensionStorage[element].getCommonSkyPixOverlapTable(self.summary.dataId)
-        assert table is not None, f"No relationship found from the common skypix to {element}."
-        dimensions = NamedValueSet(element.graph.required)
-        dimensions.add(self.summary.universe.commonSkyPix)
-        self.joinTable(table, dimensions)
+        storage = self._dimensionStorage[element]
```
natelust, line 103: do you want a normal lookup failure or something with a custom message if the element is not in the dimensionStorage

python/lsst/daf/butler/registry/interfaces/_dimensions.py
```diff
@@ -48,6 +57,74 @@ class DimensionRecordStorage(ABC):
     subclasses to provide them is not a significant burden.
     """
 
+    @classmethod
+    @abstractmethod
+    def initialize(cls, db: Database, element: DimensionElement, *,
+                   context: Optional[StaticTablesContext] = None) -> DimensionRecordStorage:
+        """Construct an instance of this class using a standardized interface.
```
natelust, line 64: This is an odd docstring, because the words you used also exactly describe new/init

natelust (approved):
Some questions, otherwise good.

python/lsst/daf/butler/registry/dimensions/query.py
TallJimbo: I think I meant to say "engine and namespace" instead of "engine namespace". I am trying hard to use "namespace" instead of "schema" because the latter is even more overloaded.

python/lsst/daf/butler/registry/dimensions/table.py
TallJimbo: It's a variadic argument, and type annotations for variadic arguments (like those for the type yielded by a generator) are supposed to be for a single element.

python/lsst/daf/butler/registry/interfaces/_dimensions.py
TallJimbo: I've come to the conclusion that SQLAlchemy isn't well suited to `from <module> import <symbol>`, because so many of its symbols are tiny and don't provide enough context - things like `Sequence`, `Table`, `and_` etc.  More musings on this subject in a post from a few days ago [on slack](https://lsstc.slack.com/archives/C2JPL8U78/p1579878585048400).

And `import sqlalchemy.sql` would still invoke `sqlalchemy/__init__.py` - the only reason to use it would be to force the `sql` subpackage to also be imported, but SQLAlchemy is designed for that not to be necessary.

python/lsst/daf/butler/registry/dimensions/spatial.py
TallJimbo: The base class docs (looks like I need to add a comment pointing to them here) clarify that this method should only ever be called by `QueryBuilder.joinDimensionElement` (this is a visitor pattern thing), and it's responsible for guaranteeing that `regions` is provided (even if it is an empty dict, which is fine) only when appropriate.

python/lsst/daf/butler/registry/dimensions/table.py
TallJimbo: I've at least [documented](https://pipelines.lsst.io/v/daily/py-api/lsst.daf.butler.DimensionRecord.html#lsst.daf.butler.DimensionRecord) that `__slots__` is considered part of the public API for `DimensionRecord`, but I have always wondered whether it'd be better to add a differently-named class attribute with the exact same elements instead.  If you have a strong preference for that, this would not be a bad ticket to make that change.

python/lsst/daf/butler/registry/queries/_builder.py
TallJimbo: Normal failure is fine for now; I think this is guaranteed to succeed with the only `DimensionRecordStorageManager` implementation we have right now, and how to handle manager implementations that do not guarantee this (because they're in a multi-layer Registry in which only the base layer has tables for all dimensions) is sufficiently unsettled that I'm sure this line will have to change in other ways if we don't continue to guarantee its success.

python/lsst/daf/butler/registry/interfaces/_dimensions.py
TallJimbo: Not quite - `__new__` (which is what this is more like) is not usually expected to have the same signature (what I mean by "standardized") across all subclasses of an ABC, and that's also what makes `__new__` inappropriate (and maybe disallowed?) as an `abstractmethod`.

PR #224, opened 2020-01-28
DM-21246: butler cleanup and support for deferred run/collection passing

TallJimbo:


python/lsst/daf/butler/_butler.py
```diff
@@ -85,29 +104,77 @@ class Butler:
...
+                 searchPaths: Optional[List[str]] = None,
+                 writeable: Optional[bool] = None):
+        if butler is not None:
+            if config is not None or searchPaths is not None or writeable is not None:
+                raise TypeError("Cannot pass 'config', 'searchPaths', or 'writeable' "
+                                "arguments with 'butler' argument.")
+            self.registry = butler.registry
+            self.datastore = butler.datastore
+            self._storageClasses = butler.storageClasses
+            self._composites = butler.composites
```
timj: If the attribute has changed to an internal attribute of `_composites` how do we copy them from another butler as `composites`? Similarly for the lines above and below. Either I'm missing something or our test coverage is poor.

python/lsst/daf/butler/_butler.py
```diff
@@ -470,21 +607,41 @@ def getDeferred(self, datasetRefOrType: typing.Union[DatasetRef, DatasetType, st
             A `dict` of `Dimension` link name, value pairs that label the
             `DatasetRef` within a Collection. When `None`, a `DatasetRef`
             should be provided as the first argument.
+        collection : `str`, optional
```
timj, line 610: collection is documented twice.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -216,7 +217,16 @@ def __init__(self, config, registry, butlerRoot=None):
...
         self._tableName = self.config["records", "table"]
-        registry.registerOpaqueTable(self._tableName, self.makeTableSpec())
+        try:
+            registry.registerOpaqueTable(self._tableName, self.makeTableSpec())
+        except ReadOnlyDatabaseError:
+            # If the database is read only and we just tried and failed to
+            # create a table, it means someone is trying to create a read-only
+            # butler client for an empty repo.  That should be okay, as long
+            # as they then try to get any datasets before some other client
+            # creates the table.  Chances are they'rejust validating
```
timj, line 227: Typo in comment rejust

timj (approved):
Some nice cleanups here. I don't think we are testing construction of a Butler from another butler though.

python/lsst/daf/butler/_butler.py
TallJimbo: Yup, this is poor test coverage in `daf_butler`, though code exercising it in obs_base is tested in `ci_hsc` (both of them, I think), and those did fail - Jenkins just hadn't gotten that far yet when I put this into review.

Instead of adding tests to `daf_butler`, though, I may see if I can just remove the `butler` argument and instead change the `obs_base` code to stop using it - this ticket was originally created because I wanted to make a butler before I knew the run/collection in that code, and I added the `butler` argument so I could quickly make new butlers with different runs/collections as a workaround.  I figure less code to test is even better than more tests.  But if that doesn't work, I'll add a test to `daf_butler`.

timj: It doesn't seem crazy to seed a new butler from another butler and then change the collection. A test seems trivial (create the butler then create another from it).

PR #226, opened 2020-01-28
DM-22776: Update Gen3 isrTask for BF corrections

czwa:
Add BrighterFatterKernel storageClass.

### DIFF
--- a/config/storageClasses.yaml
+++ b/config/storageClasses.yaml
@@ -140,6 +140,8 @@ storageClasses:
     pytype: numpy.ndarray
   StrayLightData:
     pytype: lsst.ip.isr.StrayLightData
+  BrighterFatterKernel:
+    pytype: lsst.cp.pipe.makeBrighterFatterKernel.BrighterFatterKernel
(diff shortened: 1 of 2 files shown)


config/storageClasses.yaml
```diff
@@ -140,6 +140,8 @@ storageClasses:
     pytype: numpy.ndarray
   StrayLightData:
     pytype: lsst.ip.isr.StrayLightData
+  BrighterFatterKernel:
+    pytype: lsst.cp.pipe.makeBrighterFatterKernel.BrighterFatterKernel
```
natelust, line 144: Do you need to make a formatter for this?

natelust (approved):

config/storageClasses.yaml
czwa: I have added an entry to config/datastores/formatters.yaml to be safe.  The test I ran ingested a BrighterFatterKernel from the gen2 butler, and that explicitly required a formatter entry (and so didn't cause any issues)

PR #227, opened 2020-01-31
DM-23275: Add back makeRepo support for outfile

timj:


python/lsst/daf/butler/_butler.py
```diff
@@ -291,6 +292,12 @@ def makeRepo(root: str, config: Union[Config, str, None] = None, standalone: boo
 
         if standalone:
             config.merge(full)
+        if outfile is not None:
+            # When writing to a separate location we must include
+            # the root of the butler repo in the config else it won't know
+            # where to look.
+            config["root"] = uri.geturl()
+            uri = outfile
```
kfindeisen: Oops, I didn't realize that `config["root"]` needed to be updated.

I'm a little nervous about reusing `uri` to mean first a normalized repository root, and then the config location. Why not have a separate `config.dumpToUri(outfile)` on this branch? It would be more robust to future code changes, especially ones below this point.

An alternative would be assigning `outfile = uri` if the argument is `None`, as this would keep the meaning of both variables unchanged.

python/lsst/daf/butler/registry/_config.py
```diff
@@ -85,7 +86,17 @@ def replaceRoot(self, root: str):
...
+        if BUTLER_ROOT_TAG not in self["db"]:
+            return
+
+        # This can currently only work with file scheme
+        uri = ButlerURI(root)
+        if not uri.scheme or uri.scheme == "file":
+            path = uri.path
+        else:
+            raise RuntimeError(f"Root replacement only works for file URIs")
+        self["db"] = replaceRoot(self["db"], path)
```
kfindeisen: I don't understand this change. What does it have to do with `outfile` support? Should it be a separate commit?

tests/test_butler.py
```diff
@@ -788,6 +788,31 @@ def testFileLocations(self):
...
+    def setUp(self):
+        self.root = tempfile.mkdtemp(dir=TESTDIR)
+        self.root2 = tempfile.mkdtemp(dir=TESTDIR)
+
+        self.tmpConfigFile = os.path.join(self.root, "different.yaml")
+        Butler.makeRepo(self.root, config=Config(self.configFile),
+                        outfile=self.tmpConfigFile)
+        print(f"YAML written to {self.tmpConfigFile}")
+        c = Config(self.tmpConfigFile)
+        print(c)
```
kfindeisen: Printing the config isn't really a test. Can you replace this with an assertion (e.g., that the recovered config is equal aside from the `root` field)?

kfindeisen: It seems that there are two more cases that could be tested: making `outfile` a directory (if this is not allowed, please say so in the documentation) and making it  a URI. I can't think of any obvious failure modes besides invalid/nonwritable location...

kfindeisen (changes requested):
I have some minor quibbles, but I'm happy to see this merged once you clarify what the `_config.py` changes are for. Thanks for taking care of this so quickly!

python/lsst/daf/butler/registry/_config.py
timj: This was needed because of the `.geturl()` change. When the output config file was a relative path it was all fine but in the tests it's an absolute path and that means that `root` is set to `file://` and `replaceRoot` does not understand URIs (so it ended up with `sqlite:file:...` which does not work).  I made it part of this commit because I want tests to pass on this commit and not defer fixing the tests until a later commit. I realize that we aren't entirely clear on whether we prefer tests to pass over "separation of concerns" in commits.

python/lsst/daf/butler/_butler.py
timj: ok.

Yes, if root is not in the config file it no longer knows how to find it's registry/datastore.

tests/test_butler.py
timj: Those print statements are some debugs that should not have been committed.

python/lsst/daf/butler/registry/_config.py
kfindeisen: I think the former is "obviously" correct. If one commit breaks the build and another fixes it, then the commits are no longer self-contained changes.

kfindeisen (approved):

python/lsst/daf/butler/registry/_config.py
timj: I did think about moving it into replaceRoot but for a datastore you generally will want file: and s3: to work as the root.

tests/test_butler.py
timj: Those two options (URI and directory only) do work but now I'm wondering about testing them since one put/get should be a sufficient test whereas at the moment I'm rerunning every butler test (because it's simpler that way). I may have to stop inheriting from the posix datastore test and do a one of test just for makeRepo outfile.

timj:
@kfindeisen I rewrote the tests to add your suggested changes and that made it clear that there was also a bug in posix datastore not working with file URIs. Fixing that meant I no longer needed special case code in registry.

PR #229, opened 2020-02-04
DM-23174: Consolidate daf_butler test code

kfindeisen:
This PR moves the test utilities previously located in the `tests/` directory to a new `lsst.daf.butler.tests` package (which is not included in `lsst.daf.butler`). It also adds some extra utilities designed for test code in other packages, such as functions for creating and configuring mock repositories (see lsst/pipe_base#114 for an example of where this would be useful).


COPYRIGHT
```diff
@@ -1,3 +1,4 @@
 Copyright 2016-2019 The Board of Trustees of the Leland Stanford Junior University, through SLAC National Accelerator Laboratory
-Copyright 2018, 2019 Association of Universities for Research in Astronomy
-Copyright 2015, 2018, 2019 The Trustees of Princeton University
+Copyright 2018-2019 Association of Universities for Research in Astronomy
+Copyright 2015, 2018-2019 The Trustees of Princeton University
+Copyright 2019 University of Washington
```
timj: 2020?

python/lsst/daf/butler/tests/_examplePythonTypes.py
```diff
@@ -25,8 +25,113 @@
...
+
+    storage = StorageClass(name, *args, **kwargs)
+    try:
+        storageRegistry.registerStorageClass(storage)
+    except ValueError:
+        storage = storageRegistry.getStorageClass(name)
+
+    try:
+        # if lsst.daf.butler.datastores.fileLikeDatastore.FileLikeDatastore
+        formatterRegistry = butler.datastore.formatterFactory
```
timj: Naively I'm now wondering what happens for ChainedDatastore which won't work with this. For a ChainedDatastore you need to do something like:
```python
try:
    datastores = butler.datastore.datastores
except AttributeError:
    datastores = [butler.datastore]

for datastore in datastores:
    try:
        formatterRegistry = datastore.formatterFactory
    except AttributeError:
        ...
```

python/lsst/daf/butler/tests/_examplePythonTypes.py
```diff
@@ -25,8 +25,113 @@
...
+    storageRegistry = butler.datastore.storageClassFactory
+
+    storage = StorageClass(name, *args, **kwargs)
+    try:
+        storageRegistry.registerStorageClass(storage)
+    except ValueError:
+        storage = storageRegistry.getStorageClass(name)
+
+    try:
+        # if lsst.daf.butler.datastores.fileLikeDatastore.FileLikeDatastore
```
timj: It took me a while to understand this comment so I think I'd prefer something like "Will fail if the datastore lacks a formatter factory"

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -0,0 +1,243 @@
...
+    Notes
+    -----
+    Since the values in ``dataIds`` uniquely determine the repository's
+    data IDs, the fully linked IDs can be recovered by calling
+    `expandUniqueId`, so long as no other code has inserted dimensions into
+    the repository registry.
+    """
+    if not config:
+        config = Config()
+        config[".datastore.cls"] = "lsst.daf.butler.datastores.inMemoryDatastore.InMemoryDatastore"
```
timj: In our internal butler code we prefer to write hierarchical keys as a tuple: `config["datastore", "cls"]` so we don't need to worry about delimiters (next two lines as well).

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -0,0 +1,243 @@
...
+        config[".datastore.checksum"] = False  # In case of future changes
+        config[".registry.db"] = "sqlite:///:memory:"
+    Butler.makeRepo(root, config=config, **kwargs)
```
timj: If you are allowing this function to take `kwargs` then you should use the return value of `makeRepo` and pass that to the `Butler()` constructor instead of `root`.  That protects you against `outfile` being used.

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -0,0 +1,243 @@
...
+
+    Returns
+    -------
+    butler : `lsst.daf.butler.Butler`
+        A Butler referring to a new collection in the repository at ``root``.
+        The collection is (almost) guaranteed to be new.
+    """
+    # Create a "random" collection name
+    # Speed matters more than cryptographic guarantees
+    collection = "test" + "".join((str(i) for i in np.random.randint(0, 10, size=8)))
```
timj, line 102: Since each test butler is only associated with a single python process couldn't we also use a global counter?

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -0,0 +1,243 @@
...
+    return Butler(butler=repo, run=collection)
+
+
+def _makeRecords(dataIds, universe):
```
timj, line 106: @TallJimbo are these next two functions generically useful such that they should be somewhere more discoverable outside of tests?

python/lsst/daf/butler/tests/_examplePythonTypes.py
```diff
@@ -25,8 +25,113 @@
...
 
+__all__ = ("ListAssembler", "MetricsAssembler", "MetricsExample", "registerMetricsExample")
+
+
 import copy
-from lsst.daf.butler import CompositeAssembler
+from lsst.daf.butler import CompositeAssembler, StorageClass
+
+
+def registerMetricsExample(butler):
```
timj, line 35: Should we change the existing tests to use this approach instead of using config files? This function only seems to be used in one test at the moment so I'm not sure what it's for.

tests/test_testRepo.py
```diff
@@ -0,0 +1,163 @@
...
+
+from lsst.daf.butler.tests import (makeTestRepo, makeTestCollection, addDatasetType, expandUniqueId,
+                                   MetricsExample, registerMetricsExample)
+
+
+class ButlerUtilsTestSuite(lsst.utils.tests.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        # Repository should be re-created for each test case, but
+        # this has a prohibitive run-time cost at present
```
timj, line 43: We need to work this out. `tests/test_butler.py` creates huge numbers of temp directories with their own butlers and this test class should not be any different. For example:
```
python tests/test_butler.py
.........................................................x......
----------------------------------------------------------------------
Ran 64 tests in 15.860s
```
and most of those 64 tests are calling `makeRepo` in their own temp directory.

tests/test_testRepo.py
```diff
@@ -0,0 +1,163 @@
...
+        cls.root = tempfile.mkdtemp()
```
timj: Other butler tests create the temp directory inside TESTDIR.

tests/test_testRepo.py
timj: This test on this branch is really fast for me:
```
$ time python tests/test_testRepo.py 
........
----------------------------------------------------------------------
Ran 8 tests in 0.386s

OK

real	0m1.711s
user	0m1.070s
sys	0m0.604s
```

tests/test_testRepo.py
```diff
@@ -0,0 +1,163 @@
...
+        self.butler = makeTestCollection(self.creatorButler)
+
+    def testButlerValid(self):
+        self.butler.validateConfiguration()
+
+    def testButlerKwargs(self):
+        # outfile has the most obvious effects of any Butler.makeRepo keyword
+        with tempfile.TemporaryDirectory() as temp:
+            path = os.path.join(temp, 'oddConfig.py')
+            makeTestRepo(self.root, {}, outfile=path)
```
timj, line 75: I need to think why this works fine given that self.root won't include a butler.yaml file so the Butler created inside makeTestRepo should fail because it won't know about the config written to `path`.  I think at least that the Butler returned will have a different config to the one in `path`.

tests/test_testRepo.py
```diff
@@ -0,0 +1,163 @@
...
+                      [{"instrument": "notACam", "detector": 5},
+                       {"instrument": "dummyCam", "detector": 5}])
+        self.assertIn(dict(expandUniqueId(self.butler, {"physical_filter": "k2020"})),
+                      [{"instrument": "notACam", "physical_filter": "k2020"},
+                       {"instrument": "notACam", "physical_filter": "k2020"}])
+        with self.assertRaises(ValueError):
+            expandUniqueId(self.butler, {"tract": 42})
+
+
+class MyMemoryTestCase(lsst.utils.tests.MemoryTestCase):
```
timj: I'm trying not to use lsst.utils.tests in daf_butler tests so please remove this and the init call below. (we have it in the Fits tests because those use afw and need to move at some point -- I also see one sneaked into the dbauth test). We are trying to be very careful about adding dependencies that we don't really need.

tests/test_testRepo.py
timj: Maybe we should use https://pypi.org/project/pytest-profiling/ and compare timings.

timj (changes requested):
Thanks for this clean up. I have a few questions before I do a final sign off. In particular I'd like @TallJimbo to comment on a couple of the test routines.

python/lsst/daf/butler/tests/_examplePythonTypes.py
kfindeisen: I know it's unpythonic, but might it be worth adding an abstract method to `Datastore` that does this (and raises, of course, if there is no formatter at all)? It would be more robust to other `Datastore` classes popping up in the future.

python/lsst/daf/butler/tests/_testRepo.py
kfindeisen: It might be worth clarifying in the documentation that this is the desired pattern. While reading the [current docs](https://pipelines.lsst.io/v/d_2020_02_04/py-api/lsst.daf.butler.Butler.html) shows that you can do this, it's not at all obvious what the advantages/disadvantages of any particular config to `Butler()` are. The distinction between `daf.butler.Config` and `daf.butler.ButlerConfig` is also pretty confusing.

python/lsst/daf/butler/tests/_testRepo.py
kfindeisen: Maybe? I'll admit I'd be tempted to use a static counter if Python had static variables, but I'm not sure I understand Python's memory model well enough to guarantee it's safe(r). In particular, can we be sure that each test butler will be process-confined, given that we can't really say "write your test suite this way and only this way"?

python/lsst/daf/butler/tests/_examplePythonTypes.py
kfindeisen: I assumed that the existing config files made other, possibly more complex, changes relative to the default config (the sheer number of files is a bit :dizzy_face:).

I'm already using this on [DM-22599](https://github.com/lsst/pipe_base/blob/tickets/DM-22599/tests/test_pipelineTaskTests.py), and being able to use `MetricsExample` instead of `numpy.Array` as a dummy dataset makes the test assertions much easier to read (at the cost of more boilerplate overall). I'll admit, though, that I don't expect it to be of much use in testing real `PipelineTask` classes, as there you could freely use the real data types. If you don't think it's useful here, I could move it to `pipe_base/tests/`.

tests/test_testRepo.py
kfindeisen: I'll change it for consistency, but, for the record, having to clean up those directories every time a test fails is a *real* pain.

tests/test_testRepo.py
kfindeisen: I propose splitting that into another ticket, possibly a high-priority one for February. While I am worried about what's going on, the pipeline test framework is something we need ASAP, and this ticket has already dragged on for two weeks for various reasons (e.g., the Butler API changing due to concurrent development). There is no guarantee that we can track down the cause of the slowdown (which, as far as we know, affects only my computer) in any particular amount of time.

python/lsst/daf/butler/tests/_examplePythonTypes.py
timj: Add this to `datastore.py`:
```python
@property
def datastores(self):
    return [self]
```
?

python/lsst/daf/butler/tests/_testRepo.py
timj: For most people they can't really tell the difference but for outfile it should be the difference between it working and not working at all. A `Config` is a hierarchical dict. A `ButlerConfig` knows about butler configurations and can expand defaults.

python/lsst/daf/butler/tests/_testRepo.py
timj: I'm not insisting on changes here.

python/lsst/daf/butler/tests/_examplePythonTypes.py
timj: If this is useful to you then leave it here. I was more wondering whether I could make use of it in test_butler.py but I don't require that to happen on this ticket.

tests/test_testRepo.py
timj: Ok. Create a new ticket. It is concerning that you see such a problem. I guess it's a good thing that you have demonstrated that you can run all tests with a shared butler and changing collections for each test.

python/lsst/daf/butler/tests/_testRepo.py
kfindeisen: Even if the code I originally wrote didn't work, I don't think there's any way I could have determined on my own that this was the fix.

timj: My point is that if you have asked for the configuration file to be written somewhere other than root there is an expectation that you would know to use that outfile in your Butler constructor. In your case since you are not in charge of kwargs and do not want to spend time analyzing them, your only safe option is to use the returned config as the argument to Butler.

python/lsst/daf/butler/tests/_testRepo.py
TallJimbo: The first one isn't; it really does feel like a test helper.  The second one might be more generally useful, but I think it's a special case of more general functionality that we want to make more accessible than it is currently.  So I think I'd file it under a more general work to make `queryDimensions` more convenient rather than moving this function now.

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -0,0 +1,243 @@
...
+                else:
+                    castType = key.dtype().python_type
+                try:
+                    castValue = castType(value)
+                except TypeError:
+                    castValue = castType()
+                expandedValue[key.name] = castValue
+            for key in dimension.metadata:
+                if not key.nullable:
+                    expandedValue[key.name] = key.dtype().python_type(value)
```
TallJimbo, line 145: Mostly to check that I'm reading this correctly (and maybe suggest that the answer be added to the docs): this just tries to populate any other required fields of the generated records by casting the data ID value for that dimension to the appropriate type for that field?

It's possible that won't work as the set of columns and types grows to realistic production complexity, but it seems like it's a reasonable approach for now.  And if/when that breaks, I'd rather just have some shared test data that provides all of the correctly-related dimension records one might want in a test (e.g. something derived from the ci_hsc output database might work), though of course I understand that this is not the ticket for such a switch.

TallJimbo (approved):

timj (approved):

python/lsst/daf/butler/tests/_testRepo.py
kfindeisen: ~I'm not going for "realistic production complexity". This is specifically for unit testing, i.e. minimal working environments for a specific kind of Butler/PipelineTask functionality.~

Would it help if I added something more explicit to the docs that this method is only for if you *don't care* how your test data are structured, you just want simple Butler queries to work as expected?

kfindeisen: Oh wait... are you saying the default config will get more complex? That would be a problem. :worried:

tests/test_testRepo.py
kfindeisen: Done as [DM-23357](https://jira.lsstcorp.org/browse/DM-23357).

python/lsst/daf/butler/tests/_examplePythonTypes.py
kfindeisen: I'm trying to write a regression test for this problem. Is there an easy way to get a `ChainedDatastore` repository that doesn't support `MetricsExample` (so that I can see if `registerMetricsExample` registers the formatter correctly)?

timj: I think the easiest is to copy the chained datastore yaml test config and stop it loading the metrics formatters. Then use that.

kfindeisen: I ended up going with the original double-`try` approach. Adding a property to `Datastore` broke `ABCDataStore`.

python/lsst/daf/butler/tests/_testRepo.py
TallJimbo: Yes, default config will get more complex, but let's not worry about that now. The rest of this discussion is getting into the philosophy of how we design and test daf_butler, and we should have that (elsewhere) at some point.  Look for a community post in the next few days.

PR #231, opened 2020-02-13
DM-23503: Deal with / in dataId values in file templates

timj:
Previously the templates would create extra unexpected directories.


TallJimbo (approved):
Looks good.

As a heads-up, I was planning to drop support for "collection" as a placeholder on DM-21849 (6f6a1254a0b9e33b77f5342945f5aa0d32334fa7 now, though I'm sure I'll rebase before I'm done) - it's already a synonym for "run", and after RFC-663 I think it'll be more confusing than convenient.  Not anything you need to do anything about now, but I thought I'd mention it as potentially related.

PR #232, opened 2020-02-21
DM-22655: Genericize gen2to3.py to be useable with any gen2 repo

parejkoj:


timj:
I've been thinking about the cpBias problem over night. I'm starting to think that putting instrument specific dataset type entries in daf_butler is a mistake.  These aren't used by butler at all anyhow since they are only used for ingest.  It would be much better if we put them in obs_decam itself since that's the only instrument that ingests cpBias. We already have the Instrument class returning the formatter for raw so why can't we have the instrument class have a mapping for translating an instrument-specific dataset type to a formatter? It's generalizing the raw formatter lookup.

TallJimbo:
I think per-DatasetType configurations for Formatters absolutely need to go in *repo-level* butler configs, but I totally buy @timj's argument as a reason they should get there via some means other than something in `daf_butler/config/*`.  I don't think we have another mechanism that would work well right now, but we already have other arguments pushing us towards having a set of butler config files in obs_base, and extending that to other obs packages seems reasonable.

timj:
In this case this is entirely an ingest issue just like raw and once the formatter is set the butler doesn't need to know about it. Generalizing the raw formatter lookup to other dataset types doesn't seem like a big stretch.

TallJimbo:
Ah, yup, I had forgotten that this was a case where we'd never put.  I think I proposal still stands for the general case (which I imagine will come back up), but I'm glad we don't need that mechanism right now.

timj (approved):
Approving this but the ticket to fix the cpBias issue is DM-23729.

PR #233, opened 2020-02-28
DM-23678: Flush the write before reading in S3

timj:
On some operating systems reading from a file (within the same process)
that has been written to but not closed will not work.  Flush the
buffer before reading.

Closing the file would result in it being deleted.


SimonKrughoff (approved):
Looks good.

PR #234, opened 2020-03-03
DM-23711: Support shell variables in includeConfigs directive

timj:


TallJimbo (approved):
Looks good.  I don't suppose you can easily do anything about this either way (except maybe document it), but do you know if `expandvars` will complain if referenced envvars don't exist, vs. just expanding them to "", the way shells typically (and unfortunately!) do?

timj:
It's documented to leave the variable in there untouched. Given that this is only used for reading configs you immediately get a big complaint about the file not being found.

timj:
I think complaining about that is a fairly obvious indicator of a problem since it will report the file name that includes the `$`

timj:
If you really want I can check for the presence of a `$` and raise my own exception before the FileNotFoundError arrives.

TallJimbo:
No need for special checking. Leaving the var untouched should be enough of a diagnostic; I'm just glad it doesn't do what the shell does, as that has a much better chance of resolving to something incorrect that might exist.

PR #235, opened 2020-03-03
DM-23718: Replace dots in file templates with underscore

timj:
Sometimes physical_filter definitions include spaces and it's
much easier to use a file name if spaces are not there.

### DIFF
--- a/tests/test_templates.py
+++ b/tests/test_templates.py
@@ -76,6 +76,12 @@ class TestFileTemplates(unittest.TestCase):
                             "run/2/calexp/00052/U-trail-run_2",
                             self.makeDatasetRef("calexp", run="run/2", conform=False))
 
+        # Check that "." are replaced in the file basename, but not directory.
+        dataId = {"instrument": "dummy", "visit": 52, "physical_filter": "g.10"}
+        self.assertTemplate(tmplstr,
+                            "run.2/calexp/00052/g_10-trail-run_2",
+                            self.makeDatasetRef("calexp", run="run.2", dataId=dataId, conform=False))
+
         with self.assertRaises(FileTemplateValidationError):
             FileTemplate("no fields at all")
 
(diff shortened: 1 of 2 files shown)


kfindeisen (approved):

parejkoj:
@timj : you technically can't be a reviewer of this because you created it, but can you please take a look? I'll leave it to you to update the ticket and PR description.

We should probably squash these two commits together?

I'm concerned about any effects this may have downstream if there are templates in use that are only `basename` that have `'.'` in them.

timj:
You mean you are worried about a template that refers to the special . directory but has no file component at all?  A template that has no file in it is going to break.

parejkoj:
The other way around: a template that has no `/` in it at all (and thus no directory, just basename).

timj:
We can trap a template that ends in a /

timj:
This doesn’t affect existing files in data stores. Only future ones.

tests/test_templates.py
```diff
@@ -128,11 +134,13 @@ def testComponent(self):
...
         self.assertTemplate(tmplstr, "run2_c_output_v52", refMetricOutput)
 
-        tmplstr = "{run}_{component:?}_{visit}"
-        self.assertTemplate(tmplstr, "run2_52", refMetric)
-        self.assertTemplate(tmplstr, "run2_output_52", refMetricOutput)
-        self.assertTemplate(tmplstr, "run2_maskedimage.variance_52", refMaskedImage)
-        self.assertTemplate(tmplstr, "run2_output_52", refMetricOutput)
+        # We want this template to have both a directory and basename, to
+        # test that the right parts of the output are replaced.
+        tmplstr = "{run}_{component:?}/{visit}"
```
timj: I think if you changed this template to `"{component:?}/{run}_{component:?}_{visit}"` then you would test the directory and file part of the replacement without needing the extra test above with the filer.

timj (commented):
Looks good. Yes squash since we aren't doing the space replacement.

tests/test_templates.py
parejkoj: That's a good idea. I left the other test, since it's a little more obvious what's happening in it.

timj:
@parejkoj use `git commit --amend --author "..."` to update the author of the commit you just squashed so it's not me.

PR #236, opened 2020-03-04
DM-21849: overhaul collections system

TallJimbo:
There are a few mostly-separate projects here (and a lot of miscellaneous prep work), combined in one branch to yield only one set of disruptive changes downstream (rather than four).  I've tried hard to keep the commits for those projects separate, but sadly GitHub's chronological rather than topological ordering makes that particularly hard to see.  Here are the commits ordered topologically and grouped by project:

Various cleanups to existing code and other prep work (recommended reviewer @timj):
- 91929abdad2ddec797b848f235e81c50e76f820f
- 837bf0db514e30e6bfa53c53b4db3ed0ed192604
- b58233d549825e6020b175f010f1caa8fd2ad000
- 8be92eb399bea2e6d22c4963173692dd350d30c5 (note that this refactor is extended significantly in a1abbc16994b6f5999758f8cd043e84814db57b4)
- 35d718f0f34742a1fe7b5c935fa8c4c60d57941e
- 40748ecb2c73a7b83cd4b9e3d960d2bc8263639b
- b1599a7ff39edc97f6d30a120b615791278ff2e9
- 90180bfd7d8f50d35ef737fcf2c73b35f4294e9f
- 8815607513371a31b15b6ff7b555e155a8b7899a
- 2cfca3f381b7d5c8a3bcb1b2bf9d10deb84dd8b3
- 2ede6d09ac435fbf530975acbcdc14a8454d0d3a
- 832e86d7f952fc48b8a806d785f13191add03d6b

Reimplement collections in Registry ala prototype, and make runs a type of collection (recommended reviewer @andy-slac):
- 7e1bfa0349c96cbeb8a5d78bcc487e3837a6a7f8
- 05dcd339afe2c7354606d9cbabe751c4b8896b57
- 38ec71c84e1c6bc74a71d76874375c87015ef8b7

Move multi-collection support into Registry and Butler, add CHAINED collections (no preference for reviewer):
- 9d1416db26e9aabcd8ee954330ca08d7e8975bc5
- a1abbc16994b6f5999758f8cd043e84814db57b4
- b4f007c4d06ccc789c10f0037a7bc24de5fc2ff7
- 7baa4bc59c9f8814be01d6d3161bccb6fd84b07f
- b23fd52f2058f7a8f48f9e17d73b94c921e007a8
- f07595fc6b761407d130c2e6f3ab42d9dcb0f6ae
- 1bd64fd1bcd63c88dbf4f7772a5f18aa47a1b6c5

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -928,7 +928,7 @@ class Butler:
             Datasets to prune.  These must be "resolved" references (not just
             a `DatasetType` and data ID).
         disassociate : bool`, optional
-            Disassociate pruned datasets from ``self.collection`` (or the
+            Disassociate pruned datasets from ``self.collections`` (or the
             collection given as the ``collection`` argument).  Dataset that are
             not in this collection are ignored, unless ``purge`` is `True`.
         unstore : `bool`, optional
@@ -960,7 +960,7 @@ class Butler:
             composite datasets.  This will only prune components that are
             actually attached to the given `DatasetRef` objects, which may
             not reflect what is in the database (especially if they were
-            obtained from `Registry.queryDatasets`, which by does not include
+            obtained from `Registry.queryDatasets`, which does not include
             components in its results).
 
         Raises
@@ -1053,7 +1053,7 @@ class Butler:
                 # If we're disassociating but not purging, we can do that
                 # before we try to delete, and it will roll back if deletion
                 # fails.  That will at least do the right thing if deletion
-                # fails because the files couldn't actually be delete (e.g.
+                # fails because the files couldn't actually be deleted (e.g.
                 # due to lack of permissions).
                 for tag in tags:
                     # recursive=False here because refs is already recursive
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -70,7 +70,7 @@ from .core.safeFileIo import safeMakeDir
 from .core.utils import transactional, getClassOf
 from ._deferredDatasetHandle import DeferredDatasetHandle
 from ._butlerConfig import ButlerConfig
-from .registry import Registry, RegistryConfig
+from .registry import Registry, RegistryConfig, CollectionType, MissingCollectionError
 
 log = logging.getLogger(__name__)
 
@@ -107,15 +107,16 @@ class Butler:
         Incompatible with the ``config``, ``searchPaths``, and ``writeable``
         arguments.
     collection : `str`, optional
-        Collection to use for all input lookups.  May be `None` to either use
-        the value passed to ``run``, or to defer passing a collection until
-        the methods that require one are called.
+        Collection to use for all input lookups and (if a
+        `~CollectionType.TAGGED` collection) that new datasets should be
+        associated with in `put` and `ingest`.  May be `None` to either use the
+        value passed to ``run`` for lookups, or to defer passing a collection
+        until the methods that require one are called.  If the collection does
+        not exist, a `~CollectionType.TAGGED` collection will be created.
     run : `str`, optional
-        Name of the run datasets should be output to; also used as a tagged
-        collection name these dataset will be associated with.  If the run
-        does not exist, it will be created.  If ``collection`` is `None`, this
-        collection will be used for input lookups as well; if not, it must have
-        the same value as ``run``.
+        Name of the `~CollectionType.RUN` collection datasets should be output
+        to in `put` and `ingest`.  If ``collection`` is `None`, this collection
+        will be used by default for input lookups as well.
     searchPaths : `list` of `str`, optional
         Directory paths to search when calculating the full Butler
         configuration.  Not used if the supplied config is already a
@@ -166,13 +167,18 @@ class Butler:
         self.run = run
         if collection is None and run is not None:
             collection = run
-        if self.run is not None and collection != self.run:
-            raise ValueError(
-                "Run ({}) and collection ({}) are inconsistent.".format(self.run, collection)
-            )
         self.collection = collection
         if self.run is not None:
             self.registry.registerRun(self.run)
+        if self.collection is not None:
+            try:
+                self._collectionType = self.registry.getCollectionType(self.collection)
+            except MissingCollectionError:
+                # Collection does not exist; create a new one.
+                self.registry.registerCollection(self.collection, CollectionType.TAGGED)
+                self._collectionType = CollectionType.TAGGED
+        else:
+            self._collectionType = None
 
     GENERATION: ClassVar[int] = 3
     """This is a Generation 3 Butler.
@@ -500,6 +506,7 @@ class Butler:
             dataId: Optional[DataId] = None, *,
             producer: Optional[Quantum] = None,
             run: Optional[str] = None,
+            collection: Optional[str] = None,
             **kwds: Any) -> DatasetRef:
         """Store and register a dataset.
 
@@ -519,6 +526,12 @@ class Butler:
         run : `str`, optional
             The name of the run the dataset should be added to, overriding
             ``self.run``.
+        collection : `str` or `False`, optional
+            The name of a `~CollectionType.TAGGED` collection to associate the
+            dataset with, overriding ``self.collection``.  This collection
+            must have already been added to the `Registry`.  May also be
+            `False` to skip association with ``self.collection`` even when that
+            is a `~CollectionType.TAGGED` collection.
         kwds
             Additional keyword arguments used to augment or construct a
             `DataCoordinate`.  See `DataCoordinate.standardize`
@@ -547,6 +560,14 @@ class Butler:
                 raise TypeError("No run provided.")
             run = self.run
 
+        if collection is None:
+            collection = self.collection
+            collectionType = self._collectionType
+        elif collection is False:
+            collectionType = None
+        else:
+            collectionType = self.registry.getCollectionType(collection)
+
         isVirtualComposite = self._composites.shouldBeDisassembled(datasetType)
 
         # Add Registry Dataset entry.  If not a virtual composite, add
@@ -560,11 +581,15 @@ class Butler:
             components = datasetType.storageClass.assembler().disassemble(obj)
             for component, info in components.items():
                 compTypeName = datasetType.componentTypeName(component)
-                compRef = self.put(info.component, compTypeName, dataId, producer=producer, run=run)
+                compRef = self.put(info.component, compTypeName, dataId, producer=producer, run=run,
+                                   collection=False)  # We don't need to recursively associate.
                 self.registry.attachComponent(component, ref, compRef)
         else:
             # This is an entity without a disassembler.
             self.datastore.put(obj, ref)
+
+        if collectionType is CollectionType.TAGGED:
+            self.registry.associate(collection, [ref])  # this is already recursive by default
 
         return ref
 
@@ -832,13 +857,18 @@ class Butler:
         collection : `str`, optional
             Collection to remove the datasets from, overriding
             ``self.collection``.  Ignored if ``disassociate`` is `False`.
+            Must be a `~CollectionType.TAGGED` collection, unless ``purge``
+            is `True`, in which case it must be a `~CollectionType.RUN`.
         purge : `bool`, optional
             If `True` (`False` is default), completely remove the dataset from
             the `Registry`.  To prevent accidental deletions, ``purge`` may
             only be `True` if all of the following conditions are met:
 
-             - ``disassociate`` is `True`
-             - ``unstore`` is `True`
+             - Either ``collection`` is a `~CollectionType.RUN`, or
+               ``collection`` is `None` and ``self.run`` is not `None`,
+               and all given datasets are in that run;
+             - ``disassociate`` is `True`;
+             - ``unstore`` is `True`;
              - none of the given datasets are components of some other dataset.
 
             This mode may remove provenance information from datasets other
@@ -891,11 +921,26 @@ class Butler:
         #
         if not self.isWriteable():
             raise TypeError("Butler is read-only.")
-        if disassociate:
+        if purge:
+            if collection is None:
+                collection = self.run
+                if collection is None:
+                    raise TypeError("No run provided but purge=True.")
+            collectionType = self.registry.getCollectionType(collection)
+            if collectionType is not CollectionType.RUN:
+                raise TypeError(f"Cannot purge from collection '{collection}' "
+                                f"of non-RUN type {collectionType.name}.")
+        elif disassociate:
             if collection is None:
                 collection = self.collection
                 if collection is None:
-                    raise TypeError("No collection provided.")
+                    raise TypeError("No collection provided but disassociate=True.")
+                collectionType = self._collectionType
+            else:
+                collectionType = self.registry.getCollectionType(collection)
+            if collectionType is not CollectionType.TAGGED:
+                raise TypeError(f"Cannot disassociated from collection '{collection}' "
+                                f"of non-TAGGED type {collectionType.name}.")
         if purge:
             if not disassociate:
                 raise TypeError("Cannot pass purge=True without disassociate=True.")
@@ -955,7 +1000,8 @@ class Butler:
                                   "the repository in a corrupted state.") from err
 
     @transactional
-    def ingest(self, *datasets: FileDataset, transfer: Optional[str] = None, run: Optional[str] = None):
+    def ingest(self, *datasets: FileDataset, transfer: Optional[str] = None, run: Optional[str] = None,
+               collection: Optional[str] = None,):
         """Store and register one or more datasets that already exist on disk.
 
         Parameters
@@ -978,6 +1024,12 @@ class Butler:
         run : `str`, optional
             The name of the run ingested datasets should be added to,
             overriding ``self.run``.
+        collection : `str` or `False`, optional
+            The name of a `~CollectionType.TAGGED` collection to associate the
+            dataset(s) with, overriding ``self.collection``.  This collection
+            must have already been added to the `Registry`.  May also be
+            `False` to skip association with ``self.collection`` even when that
+            is a `~CollectionType.TAGGED` collection.
 
         Raises
         ------
@@ -1012,6 +1064,14 @@ class Butler:
                 raise TypeError("No run provided.")
             run = self.run
 
+        if collection is None:
+            collection = self.collection
+            collectionType = self._collectionType
+        elif collection is False:
+            collectionType = None
+        else:
+            collectionType = self.registry.getCollectionType(collection)
+
         # Reorganize the inputs so they're grouped by DatasetType and then
         # data ID.  We also include a list of DatasetRefs for each FileDataset
         # to hold the resolved DatasetRefs returned by the Registry, before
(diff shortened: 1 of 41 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -106,36 +97,106 @@ class Butler:
...
+        collections must either already exist or be specified to be created
+        by other arguments.
     run : `str`, optional
-        Name of the run datasets should be output to; also used as a tagged
-        collection name these dataset will be associated with.  If the run
-        does not exist, it will be created.  If ``collection`` is `None`, this
-        collection will be used for input lookups as well; if not, it must have
-        the same value as ``run``.
+        Name of the run datasets should be output to.  If the run
+        does not exist, it will be created.  If ``collections`` is `None`, it
```
timj, line 115: Maybe say explicitly here that if `run` is not set it is a readonly butler?

python/lsst/daf/butler/_butler.py
```diff
@@ -783,94 +888,188 @@ def datasetExists(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
...
+                                    f"of non-TAGGED type {collectionType.name}.")
+        if purge:
+            if not disassociate:
+                raise TypeError("Cannot pass purge=True without disassociate=True.")
+            if not unstore:
+                raise TypeError("Cannot pass purge=True without unstore=True.")
+            refs = list(refs)
+            for ref in refs:
+                # isComponent isn't actually reliable, because the dataset type
+                # name isn't the whole story, and that's mildly concerning
```
timj, line 1035: If the dataset type name is `A.b` how do you get that if it isn't a component?

python/lsst/daf/butler/_butler.py
```diff
@@ -783,94 +888,188 @@ def datasetExists(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
...
+                        # datastore is never given the top level DatasetRef. In
+                        # the concrete composite the datastore knows all the
+                        # refs and will clean up itself if asked to remove the
+                        # parent ref.  We can not check configuration for this
+                        # since we can not trust that the configuration is the
+                        # same. We therefore have to ask if the ref exists or
+                        # not.  This is consistent with the fact that we want
+                        # to ignore already-removed-from-datastore datasets
+                        # anyway.
+                        if self.datastore.exists(ref):
```
timj, line 1080: I still worry that this hides problems with composites since non-existence might be a bug.

python/lsst/daf/butler/_butler.py
```diff
@@ -893,6 +1092,16 @@ def ingest(self, *datasets: FileDataset, transfer: Optional[str] = None, run: Op
         run : `str`, optional
             The name of the run ingested datasets should be added to,
             overriding ``self.run``.
+        tags : `Iterable` [ `str` ], optional
+            The names of a `~CollectionType.TAGGED` collections to associate
+            the dataset with, overriding ``self.tags``.  These collections
+            must have already been added to the `Registry`.
+        collection : `str` or `False`, optional
```
timj: There is no collection argument.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -930,18 +978,35 @@ def disassociate(self, collection: str, refs: List[DatasetRef]):
...
+            Raise adding new datasets to the given ``collection`` is not
+            allowed.
         """
-        rows = [{"dataset_id": _checkAndGetId(ref), "collection": collection}
-                for ref in _expandComponents(refs)]
-        self._db.delete(self._tables.dataset_collection, ["dataset_id", "collection"], *rows)
+        collectionFieldName = self._collections.getCollectionForeignKeyName()
+        collectionRecord = self._collections.find(collection)
+        if collectionRecord.type is CollectionType.RUN:
+            raise TypeError(f"Collection '{collection}' has type {collectionRecord.type.name}, not TAGGED.")
```
timj: This error message seems odd since you are checking it IS RUN, not it's not TAGGED. This is going to get really confusing if extra collection types turn up. Can you say `is not TAGGED` instead?

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -1101,18 +1166,76 @@ def insertDimensionData(self, element: Union[DimensionElement, str],
...
+            ignored.
+        collectionType : `CollectionType`, optional
+            If provided, only yield collections of this type.
+        flattenChains : `bool`, optional
+            If `True` (`False` is default), recursively yield the child
+            collections of matching `~CollectionType.CHAINED` collections.
+        includeChains : `bool`, optional
+            If `True`, yield records for matching `~CollectionType.CHAINED`
+            collections.  Default is the opposite of ``flattenChains``: include
+            either CHAINED collections or their children, but not both.
```
timj, line 1217: These aren't used at all. Should we document that they are unused parameters or delete them? I'm not sure if this is some more general interface that you are matching.

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -819,7 +819,7 @@ def check():
             elif bad:
                 raise DatabaseConflictError(f"Conflict in sync on column(s) {bad}.")
             inserted = False
-        return result, inserted
+        return {k: v for k, v in zip(returning, result)} if returning is not None else None, inserted
```
timj, line 822: The return type changed but the documentation didn't? Was the return type wrong previously?

python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -33,8 +34,14 @@
...
     StorageClass,
     ddl,
+    YamlRepoImportBackend
 )
-from .._registry import Registry, ConflictingDefinitionError, OrphanedRecordError
+from .._registry import Registry, CollectionType, ConflictingDefinitionError, OrphanedRecordError
+from ..wildcards import DatasetTypeRestriction
+
+
+DATA_DIR = os.path.normpath(os.path.join(os.path.split(__file__)[0],
```
timj: I think `os.path.dirname()` is preferred over `os.path.split()[0]`.

python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -33,8 +34,14 @@
...
+                                         "../../../../../../tests/data/registry"))
```
timj: It might be a little easier to let the tests specify the data dir themselves and pass it in to this class?

python/lsst/daf/butler/registry/wildcards.py
```diff
@@ -0,0 +1,751 @@
...
+        ------
+        TypeError
+            Raised if an unsupported type is found in the expression.
+        """
+        assert expression is not None
+        # See if we were given ...; just return that if we were.
+        if expression is ...:
+            if not allowAny:
+                raise TypeError("This expression may not be unconstrained.")
+            return ...
```
timj, line 129: The doc string says this should return None.

python/lsst/daf/butler/registry/wildcards.py
```diff
@@ -0,0 +1,751 @@
...
+    """
+
+    items: List[Tuple[str, Any]]
+    """Two-item tuples that relate string values to other objects
+    (`list` [ `tuple` [ `str`, `Any` ] ]).
+    """
+
+
+class DatasetTypeRestriction:
+    """An immutable set-like object that represents a restriction on the
```
timj, line 267: inherit from `collections.abc.Set` ?

python/lsst/daf/butler/script/validateButlerConfiguration.py
```diff
@@ -102,7 +97,7 @@ def validateButlerConfiguration(root, datasetTypes=None, ignore=None, quiet=Fals
 
     # The collection does not matter for validation but if a run is specified
     # in the configuration then it must be consistent with this collection
```
timj: I think this comment is now completely irrelevant

timj (approved):
Sorry it took me so long to make my way through this huge change. It looks good to me although I got a bit derailed in the low level query classes.

timj:
@andy-slac can you please take a look at the commits you were assigned so we can move towards merging?

andy-slac:
Sorry, I missed that, will look at it ASAP.

python/lsst/daf/butler/registry/_collectionType.py
```diff
@@ -0,0 +1,48 @@
...
+
+    Within a particular run, there may only be one dataset with a particular
+    dataset type and data ID.
+    """
+
+    TAGGED = 2
+    """Datasets can be associated with and removed from ``TAGGED`` collections
+    arbitrarily.
+
+    Within a particular run, there may only be one dataset with a particular
```
andy-slac: run? Should it be a "tagged collection"?

python/lsst/daf/butler/registry/interfaces/_collections.py
```diff
@@ -0,0 +1,349 @@
...
+class CollectionRecord(ABC):
+    """A struct used to represent a collection in internal `Registry` APIs.
+
+    User-facing code should always just use a `str` to represent collections.
+
+    Parameters
+    ----------
+    name : `str`
+        Name of the collection.
+    id : `int`
```
andy-slac: This does not exist any more?

python/lsst/daf/butler/registry/interfaces/_collections.py
```diff
@@ -0,0 +1,349 @@
...
+            Specification for the table that should reference the collection
+            table.  Will be modified in place.
+        prefix: `str`, optional
+            A name to use for the prefix of the new field; the full name may
+            have a suffix (and is given in the returned `ddl.FieldSpec`).
+        onDelete: `str`, optional
+            One of "CASCADE" or "SET NULL", indicating what should happen to
+            the referencing row if the collection row is deleted.  `None`
+            indicates that this should be an integrity error.
+        kwds
```
andy-slac: `**kwds`?

python/lsst/daf/butler/registry/interfaces/_collections.py
```diff
@@ -0,0 +1,349 @@
...
+            Additional keyword arguments are forwarded to the `ddl.FieldSpec`
+            constructor (only the ``name`` and ``dtype`` arguments are
+            otherwise provided).
+
+        Returns
+        -------
+        fieldSpec : `ddl.FieldSpec`
+            Specification for the field being added.
```
andy-slac, line 296: So it should both add FieldSpec to a TableSpec and return it?

python/lsst/daf/butler/registry/interfaces/_collections.py
```diff
@@ -0,0 +1,349 @@
...
+            Specification for the table that should reference the run table.
+            Will be modified in place.
+        prefix: `str`, optional
+            A name to use for the prefix of the new field; the full name may
+            have a suffix (and is given in the returned `ddl.FieldSpec`).
+        onDelete: `str`, optional
+            One of "CASCADE" or "SET NULL", indicating what should happen to
+            the referencing row if the collection row is deleted.  `None`
+            indicates that this should be an integrity error.
+        kwds
```
andy-slac: `**kwds`

python/lsst/daf/butler/registry/interfaces/_collections.py
```diff
@@ -0,0 +1,349 @@
...
+        raise NotImplementedError()
+
+    @classmethod
+    @abstractmethod
+    def getCollectionForeignKeyName(cls, prefix: str = "collection") -> str:
+        raise NotImplementedError()
+
+    @classmethod
+    @abstractmethod
+    def getRunForeignKeyName(cls, prefix: str = "run") -> str:
```
andy-slac, line 352: Docstring may be useful for these two methods

python/lsst/daf/butler/registry/interfaces/_collections.py
```diff
@@ -0,0 +1,349 @@
...
+        raise NotImplementedError()
+
+    @abstractmethod
+    def refresh(self):
+        """Ensure all other operations on this manager are aware of any
+        collections that may have been registered by other clients since it
+        was initialized or last refreshed.
```
andy-slac, line 373: Hmm, does this need to be exposed in a public interface? Is there a potential consistency issue here? I can imagine that some operations will require consistent view of the database contents and `refresh()` should be called for them, but this probably needs to be a part of the same transaction, not as a separate method (otherwise the question is what is going to happen when new collection is added between refresh() and a call to other method)? Or is it for something else entirely?

python/lsst/daf/butler/registry/interfaces/_collections.py
```diff
@@ -0,0 +1,349 @@
...
+        ------
+        TransactionInterruption
+            Raised if this operation is invoked within a `Database.transaction`
+            context.
+
+        Notes
+        -----
+        Concurrent registrations of the same collection should be safe; nothing
+        should happen if the types are consistent, and integrity errors due to
+        inconsistent types should happen before any database changes are made
```
andy-slac: I guess that if collection with the same name already exists and has the same type then it will be returned normally from this method. What happens if type is different, is it an error, will it throw an exception (maybe worth documenting here)?

python/lsst/daf/butler/registry/interfaces/_collections.py
```diff
@@ -0,0 +1,349 @@
...
+        Parameters
+        ----------
+        name : `str`
+            Name of the collection.
+
+        Returns
+        -------
+        record : `CollectionRecord`
+            Object representing the collection, including its type and ID.
+            If ``record/type == CollectionType.RUN``, this will be a
```
andy-slac: `record/type` -> `record.type`?

python/lsst/daf/butler/registry/interfaces/_collections.py
```diff
@@ -0,0 +1,349 @@
...
+        Parameters
+        ----------
+        key
+            Internal primary key value for the collection.
+
+        Returns
+        -------
+        record : `CollectionRecord` or `None`
+            Object representing the collection, including its type and name.
+            If ``record/type == CollectionType.RUN``, this will be a
```
andy-slac: `record/type` -> `record.type`?

python/lsst/daf/butler/registry/interfaces/_collections.py
```diff
@@ -0,0 +1,349 @@
...
+            `RunRecord` instance.  `None` if no collection with the given
+            ID exists in this layer.
+
+        Notes
+        -----
+        Collections registered by another client of the same layer since the
+        last call to `initialize` or `refresh` may not be found.
```
andy-slac, line 469: Should this method throw an exception for missing collection? My impression that `get()` is a "stronger" version of a `find()` method, usually if someone has a foreign key it means that object exists/existed. OTOH `find()` is a general search by name, it may be more reasonable to re return `None` from that method if nothing is found.

python/lsst/daf/butler/registry/collections/nameKey.py
```diff
@@ -0,0 +1,276 @@
...
+        self._timespan = timespan
+
+    def update(self, host: Optional[str] = None, timespan: Optional[Timespan[Optional[datetime]]] = None):
+        # Docstring inherited from RunRecord.
+        if timespan is None:
+            timespan = Timespan(begin=None, end=None)
+        row = {
+            "name": self.name,
+            TIMESPAN_FIELD_SPECS.begin.name: timespan.begin if timespan is not None else None,
+            TIMESPAN_FIELD_SPECS.end.name: timespan.end if timespan is not None else None,
```
andy-slac: `timespan is not None` is always `True`, there is a protection for that above.

python/lsst/daf/butler/registry/collections/nameKey.py
```diff
@@ -0,0 +1,276 @@
...
+            "host": host,
+        }
+        count = self._db.update(self._table, row, keys=["id"],
+                                values=[TIMESPAN_FIELD_SPECS.begin.name,
+                                        TIMESPAN_FIELD_SPECS.end.name,
+                                        "host"])
```
andy-slac: I'm looking at `Database.update()` method and I'm not sure these parameters match the declaration?

python/lsst/daf/butler/registry/collections/nameKey.py
```diff
@@ -0,0 +1,276 @@
...
+                kwds["table"] = self._tables.run
+                kwds["timespan"] = Timespan(
+                    begin=row[self._tables.run.columns[TIMESPAN_FIELD_SPECS.begin.name]],
+                    end=row[self._tables.run.columns[TIMESPAN_FIELD_SPECS.end.name]],
+                )
+                kwds["host"] = row[self._tables.run.columns.host]
+                record = NameKeyRunRecord(**kwds)
+            else:
+                kwds["type"] = type
+                record = NameKeyCollectionRecord(**kwds)
```
andy-slac: It's a bit unconventionally looking, why not just `record = NameKeyCollectionRecord(type=type)`? Same for `NameKeyRunRecord` above.

python/lsst/daf/butler/registry/collections/nameKey.py
```diff
@@ -0,0 +1,276 @@
...
+            if wildcard.patterns:
+                sql = sqlalchemy.sql.select(
+                    [self._tables.collection.columns.name]
+                ).select_from(
+                    self._tables.collection
+                ).where(
+                    wildcard.makeWhereExpression(self._tables.collection.columns.name)
+                )
+                for row in self._db.query(sql).fetchall():
+                    yield self._records[row["name"]]
```
andy-slac: If collection was just created in database and is not in `self._records` this will raise `LookupError`, I'm not sure this is an expected behavior.

andy-slac (approved):
I checked three commits that were specifically assigned to me, looks OK, small bunch of comments.

python/lsst/daf/butler/_butler.py
TallJimbo: I was thinking of the converse, in the deferred-virtual-composite case, e.g. we write `jointcal_wcs` as a standalone dataset and later use it as the `wcs` component of some `Exposure` dataset.

python/lsst/daf/butler/_butler.py
TallJimbo: Agreed; I was just trying to handle the common case here, and defer a real solution to DM-23671.

python/lsst/daf/butler/registry/_registry.py
TallJimbo: Ah, that's a bug; I'd intended to forward them down to the `query.iter` call, which is already set up to handle them.  I'll do that now.

python/lsst/daf/butler/registry/interfaces/_database.py
TallJimbo: Exactly - the return type and documentation were previously inconsistent, and I decided here that the documentation was actually the more useful behavior and changed the code accordingly.

python/lsst/daf/butler/registry/wildcards.py
TallJimbo: Yeah, in an earlier (squashed away) version of this I was converting `...` to `None` when going from public to internal APIs, and then I realized it was a bad idea to have two different sentinel types to represent the same concept.  Docs fixed.

python/lsst/daf/butler/registry/wildcards.py
TallJimbo: Can't; it can't always define `__len__` (maybe other limitations, but at least that).  I'll add a comment to that effect.

python/lsst/daf/butler/registry/interfaces/_collections.py
TallJimbo: As this entirely class hierarchy is internal to `Registry`, it's not a fully-public interface, and that reflects the fact that I was thinking that any operations of they type you described would be `Registry`-level operations, not something internal to instances of this class.  Right now, this is only used when constructing `Registry` instance that points at an existing repository.

python/lsst/daf/butler/registry/interfaces/_collections.py
TallJimbo: It should be an error; will document.

python/lsst/daf/butler/registry/interfaces/_collections.py
TallJimbo: Will do.  In keeping with Python conventions, I'll also rename it from `get` to `__getitem__`.

python/lsst/daf/butler/registry/collections/nameKey.py
TallJimbo: A later commit makes this a moot point.

PR #237, opened 2020-03-12
DM-22599: Develop PipelineTask unit test framework

kfindeisen:
This PR adds a user guide for the code in `daf.butler.tests._testRepo`, which was overlooked when that code was merged. Guides should also be written for the other code in `daf.butler.tests`, but I don't feel qualified to do so myself.

### DIFF
--- a/doc/lsst.daf.butler/use-in-tests.rst
+++ b/doc/lsst.daf.butler/use-in-tests.rst
@@ -1,0 +1,79 @@
+.. py:currentmodule:: lsst.daf.butler.tests
+
+.. _using-butler-in-tests:
+
+##############################
+Using the Butler in unit tests
+##############################
+
+This document describes how to use the Butler when testing Butler-dependent code in circumstances where one should not use a repository associated with a specific telescope, such as in unit tests.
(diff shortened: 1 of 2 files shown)


doc/lsst.daf.butler/use-in-tests.rst
```diff
@@ -0,0 +1,79 @@
...
+Once a test collection is created, datasets can be read or written as usual:
+
+.. code-block:: py
+   :emphasize-lines: 1, 4
+
+   butler = butlerTests.makeTestCollection(repo)
+
+   exposure = makeTestExposure()
+   butler.put(exposure, "calexp", dataId)
+   processCalexp(dataId)
```
mrawls: Is `processCalexp` a thing, or is that supposed to be a vague pseudocode example?

mrawls (approved):

doc/lsst.daf.butler/use-in-tests.rst
kfindeisen: Pseudocode to avoid explicitly depending on the tasks framework. I'll add a comment.

PR #243, opened 2020-03-26
DM-24018: fix bug in unrestricted queries over collections

TallJimbo:
The code path for iterating over collections with the (default) no-restriction expression `...` didn't take into account arguments that indicated how to handle chained collections, causing downstream code to die when it specified that it wanted their children, rather than the chained collections themselves.


parejkoj (approved):
Thanks for the quick fix, and adding a test of this case.

PR #244, opened 2020-03-26
DM-15890: Use nanoseconds to represent time in database

andy-slac:
Most of the code now uses astropy Time for timestamp, conversion to
nanoseconds happens in sqlalchemy type decorator. For YAML we need to
convert astropy time into Python datetime. Added two constants for
min/max times (in astropy format) supported by registry.

### DIFF
--- a/python/lsst/daf/butler/core/ddl.py
+++ b/python/lsst/daf/butler/core/ddl.py
@@ -57,10 +57,12 @@ EPOCH = astropy.time.Time("1970-01-01 00:00:00", format="iso", scale="tai")
 stored in the database.
 """
 
-MAX_TIME = astropy.time.Time("2038-01-19 03:14:07", format="iso", scale="tai")
+MAX_TIME = astropy.time.Time("2100-01-01 00:00:00", format="iso", scale="tai")
 """Maximum time value that we can store. Assuming 64-bit integer field we
-can actuially store higher values but we intentionally limit it to 2**31-1
-seconds since epoch.
+can actually store higher values but we intentionally limit it to arbitrary
+but reasonably high value. Note that this value will be stored in registry
+database for eternity, so it should not be changed without proper
+consideration.
 """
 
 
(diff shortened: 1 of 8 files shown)


python/lsst/daf/butler/core/ddl.py
```diff
@@ -125,14 +143,55 @@ def process_result_value(self, value, dialect):
         return ConvexPolygon.decode(super().process_result_value(value, dialect))
 
 
+class TimeRangeWarning(Warning):
```
timj: Is this used now?

python/lsst/daf/butler/core/ddl.py
```diff
@@ -31,21 +31,39 @@
...
 
+_LOG = logging.getLogger(__name__)
+
+# These constants can be used by client code
+EPOCH = astropy.time.Time("1970-01-01 00:00:00", format="iso", scale="tai")
+"""Epoch for calculating time delta, this is the minimum time that can be
+stored in the database.
+"""
+
+MAX_TIME = astropy.time.Time("2038-01-19 03:14:07", format="iso", scale="tai")
```
timj: I'm fine with this being 2100 or even higher. We are using nanosecond integers so 32bit seconds doesn't mean anything.

python/lsst/daf/butler/core/ddl.py
```diff
@@ -125,14 +143,55 @@ def process_result_value(self, value, dialect):
...
+    def process_bind_param(self, value, dialect):
+        # value is astropy.time.Time or None
+        if value is None:
+            return None
+        if not isinstance(value, astropy.time.Time):
+            raise TypeError(f"Unsupported type: {type(value)}, expected astropy.time.Time")
+        # sometimes comparison produces warnings if input value is in UTC
+        # scale, transform it to TAI before doing anyhting
+        value = value.tai
+        # anyhting before epoch or after MAX_TIME is truncated
```
timj: typo in comment

timj (approved):
Looks great. Thank you.

python/lsst/daf/butler/core/ddl.py
andy-slac: Hmm, I though I cleaned it, will try again 🙂

PR #246, opened 2020-04-01
DM-24245: Convert daf_butler script to Click as a test for that package

n8pease:


python/lsst/daf/butler/script/butler.py
```diff
@@ -0,0 +1,115 @@
...
+
+pass_verbosity = click.make_pass_decorator(Verbosity, ensure=True)
+
+
+def verbose_option(f):
+    def callback(ctx, param, value):
+        verbose = ctx.ensure_object(Verbosity)
+        verbose.verbose = value
+        return value
+    return click.option('-v', '--verbose', is_flag=True,
```
timj: In daf_butler we always use double quotes (unless we are inside an f-string or something similar).

tests/test_commandLine.py
```diff
@@ -0,0 +1,138 @@
...
+        with runner.isolated_filesystem():
+            result = runner.invoke(butler.cli, ['create', 'here'])
+            self.assertEqual(result.exit_code, 0)
+            # verify the just-created repo validates without error, resulting
+            # in no output.
+            result = runner.invoke(butler.cli, ['validate-config', 'here'])
+            self.assertEqual(result.exit_code, 0)
+            self.assertEqual(result.stdout, '')
+
+    # How can we test that validate-config finds invalid config data? (Hack
```
timj: You can do this by seeding the butler with a test config file that is broken if you want. The tests directory has a bunch of configs in there.

timj (commented):
This does look interesting. I think we do want to put each sub-command in its own file if at all possible since I worry that that script/butler.py file is going to get very large.  Does click allow you to predefine standard options so that you can apply them to all the subcommands?

@jonathansick I would be interested in your opinion on this? We haven't adopted it yet but we are seeing whether it will work for us. I think there is a sphinx plugin that understands the option handling.

jonathansick:
In SQuaRE we're already almost exclusively using Click along with sphinx-click for our command line interfaces. So I'd have no problem with adding sphinx-click to the standard listing of Sphinx extensions in `documenteer[pipelines]`.

in the `conf.py` file, do:

```py
extensions.append("sphinx_click.ext"]
```

Integrate your CLI reference into the docs with something like: https://github.com/lsst-sqre/ltd-conveyor/blob/master/docs/cli.rst

Of course, add https://pypi.org/project/sphinx-click/ to your requirements.txt / Python environment.

I can add `sphinx-click` to the documenteer dependencies in the upcoming documenteer 0.6.0 if you want to go ahead with this.

timj:
@jonathansick that sounds positive. Thank you. We were intending on a little RFC once we've decided that this is where we want to go.

n8pease:
@timj click does allow standard options. To have a set of standard options that can be applied to chosen commands, reusable options, like `verbose_option` in this script, are grouped into a function, e.g. `standard_options()`, and that is used to decorate the command functions that accept/implement the standard options. There is an example [here](https://github.com/pallets/click/issues/108#issuecomment-44691173).

There is a second way, which is to make them options of the top-level script/command. For example to enable verbose output globally the command could be `butler.py -v <command> <command args & opts>` - that `-v` value is accessible by all the subcommands.

n8pease:
I think it should be easy to separate each command into its own script. I'll take a crack at that.

timj:
This work was done in #258

PR #247, opened 2020-04-01
DM-24290: PostgreSQL database handles case when namespace is not provided incorrectly.

DinoBektesevic:
Solves issue described in [DM-24290 point 1.a](https://jira.lsstcorp.org/browse/DM-24290) when namespace is not provided it is set to an non-existent DSN "schema" key. Went unnoticed because all the tests provide explicit namespace arguments.

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -260,9 +260,9 @@ class Butler:
 
         Parameters
         ----------
-        root : `str`
-            Filesystem path to the root of the new repository.  Will be created
-            if it does not exist.
+        root : `str` or `ButlerURI`
+            Path or URI to the root location of the new repository. Will be
+            created if it does not exist.
         config : `Config` or `str`, optional
             Configuration to write to the repository, after setting any
             root-dependent Registry or Datastore config options.  Can not
@@ -334,10 +334,13 @@ class Butler:
             if not os.path.isdir(uri.ospath):
                 safeMakeDir(uri.ospath)
         elif uri.scheme == "s3":
-            s3 = boto3.resource("s3")
-            # implies bucket exists, if not another level of checks
-            bucket = s3.Bucket(uri.netloc)
-            bucket.put_object(Bucket=uri.netloc, Key=uri.relativeToPathRoot)
+            # implies bucket already exists
+            s3 = boto3.client("s3")
+            if uri.relativeToPathRoot != ".":
+                # to be dir-equivalent, S3 this requires URIs ending on '/'
+                if not uri.geturl().endswith('/'):
+                    uri.updateFile(uri.relativeToPathRoot+'/')
+                s3.put_object(Bucket=uri.netloc, Key=uri.relativeToPathRoot+'/')
         else:
             raise ValueError(f"Unrecognized scheme: {uri.scheme}")
         config = Config(config)
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -340,7 +340,7 @@ class Butler:
                 raise ValueError(f"Bucket {uri.netloc} does not exist!")
             s3 = boto3.client("s3")
             # don't create S3 key when root is at the top-level of an Bucket
-            if not uri.path != "/":
+            if not uri.path == "/":
                 s3.put_object(Bucket=uri.netloc, Key=uri.relativeToPathRoot)
         else:
             raise ValueError(f"Unrecognized scheme: {uri.scheme}")
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -69,6 +69,7 @@ from .core import (
 from .core.repoRelocation import BUTLER_ROOT_TAG
 from .core.safeFileIo import safeMakeDir
 from .core.utils import transactional, getClassOf
+from .core.s3utils import bucketExists
 from ._deferredDatasetHandle import DeferredDatasetHandle
 from ._butlerConfig import ButlerConfig
 from .registry import Registry, RegistryConfig, CollectionType
@@ -335,13 +336,12 @@ class Butler:
                 safeMakeDir(uri.ospath)
         elif uri.scheme == "s3":
             # bucket must already exist
-            s3 = boto3.resource("s3")
-            bucket = s3.Bucket(uri.netloc)
-            if bucket.creation_date is None:
+            if not bucketExists(uri.netloc):
                 raise ValueError(f"Bucket {uri.netloc} does not exist!")
+            s3 = boto3.client("s3")
             # don't create root Key if root is at top-level of an Bucket
             if uri.relativeToPathRoot != ".":
-                bucket.put_object(Bucket=uri.netloc, Key=uri.relativeToPathRoot)
+                s3.put_object(Bucket=uri.netloc, Key=uri.relativeToPathRoot)
         else:
             raise ValueError(f"Unrecognized scheme: {uri.scheme}")
         config = Config(config)
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -329,18 +329,19 @@ class Butler:
 
         # for "file" schemes we are assuming POSIX semantics for paths, for
         # schemeless URIs we are assuming os.path semantics.
-        uri = ButlerURI(root)
+        uri = ButlerURI(root, forceDirectory=True)
         if uri.scheme == "file" or not uri.scheme:
             if not os.path.isdir(uri.ospath):
                 safeMakeDir(uri.ospath)
         elif uri.scheme == "s3":
-            # implies bucket already exists
-            s3 = boto3.client("s3")
+            # bucket must already exist
+            s3 = boto3.resource("s3")
+            bucket = s3.Bucket(uri.netloc)
+            if bucket.creation_date is None:
+                raise ValueError(f"Bucket {uri.netloc} does not exist!")
+            # don't create root Key if root is at top-level of an Bucket
             if uri.relativeToPathRoot != ".":
-                # to be dir-equivalent, S3 this requires URIs ending on '/'
-                if not uri.geturl().endswith('/'):
-                    uri.updateFile(uri.relativeToPathRoot+'/')
-                s3.put_object(Bucket=uri.netloc, Key=uri.relativeToPathRoot+'/')
+                bucket.put_object(Bucket=uri.netloc, Key=uri.relativeToPathRoot)
         else:
             raise ValueError(f"Unrecognized scheme: {uri.scheme}")
         config = Config(config)
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -339,8 +339,8 @@ class Butler:
             if not bucketExists(uri.netloc):
                 raise ValueError(f"Bucket {uri.netloc} does not exist!")
             s3 = boto3.client("s3")
-            # don't create root Key if root is at top-level of an Bucket
-            if uri.relativeToPathRoot != ".":
+            # don't create S3 key when root is at the top-level of an Bucket
+            if not uri.path != "/":
                 s3.put_object(Bucket=uri.netloc, Key=uri.relativeToPathRoot)
         else:
             raise ValueError(f"Unrecognized scheme: {uri.scheme}")
--- a/python/lsst/daf/butler/core/config.py
+++ b/python/lsst/daf/butler/core/config.py
@@ -32,7 +32,6 @@ import yaml
 import sys
 from yaml.representer import Representer
 import io
-import posixpath
 from typing import Sequence, Optional, ClassVar
 
 try:
@@ -807,9 +806,9 @@ class Config(collections.abc.MutableMapping):
                 uri = ButlerURI(os.path.join(uri.ospath, defaultFileName))
             self.dumpToFile(uri.ospath, overwrite=overwrite)
         elif uri.scheme == "s3":
-            head, filename = posixpath.split(uri.path)
-            if "." not in filename:
-                uri.updateFile(defaultFileName)
+            if not uri.dirLike and "." not in uri.basename():
+                uri = ButlerURI(uri.geturl(), forceDirectory=True)
+            uri.updateFile(defaultFileName)
             self.dumpToS3File(uri, overwrite=overwrite)
         else:
             raise ValueError(f"Unrecognized URI scheme: {uri.scheme}")
--- a/python/lsst/daf/butler/core/config.py
+++ b/python/lsst/daf/butler/core/config.py
@@ -251,6 +251,7 @@ class Config(collections.abc.MutableMapping):
                 self.__initFromYamlFile(uri.ospath)
         else:
             raise RuntimeError("Unhandled config file type:%s" % uri)
+        self.configFile = str(path)
 
     def __initFromS3YamlFile(self, url):
         """Load a file at a given S3 Bucket uri and attempts to load it from
@@ -290,7 +291,6 @@ class Config(collections.abc.MutableMapping):
         log.debug("Opening YAML config file: %s", path)
         with open(path, "r") as f:
             self.__initFromYaml(f)
-        self.configFile = path
 
     def __initFromYaml(self, stream):
         """Loads a YAML config from any readable stream that contains one.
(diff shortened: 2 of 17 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -333,10 +333,13 @@ def makeRepo(root: str, config: Union[Config, str, None] = None, standalone: boo
             if not os.path.isdir(uri.ospath):
                 safeMakeDir(uri.ospath)
         elif uri.scheme == "s3":
-            s3 = boto3.resource("s3")
-            # implies bucket exists, if not another level of checks
-            bucket = s3.Bucket(uri.netloc)
-            bucket.put_object(Bucket=uri.netloc, Key=uri.relativeToPathRoot)
+            # implies bucket already exists
+            s3 = boto3.client("s3")
+            if uri.relativeToPathRoot != ".":
```
timj: Maybe say that this test is to see whether the gen3 repo represents the root of the bucket.

python/lsst/daf/butler/_butler.py
```diff
@@ -333,10 +333,13 @@ def makeRepo(root: str, config: Union[Config, str, None] = None, standalone: boo
...
+                # to be dir-equivalent, S3 this requires URIs ending on '/'
+                if not uri.geturl().endswith('/'):
+                    uri.updateFile(uri.relativeToPathRoot+'/')
+                s3.put_object(Bucket=uri.netloc, Key=uri.relativeToPathRoot+'/')
```
timj: Since you now have to add the `/` again here what is the point of the previous call to `updateFile`?

python/lsst/daf/butler/_butler.py
```diff
@@ -333,10 +333,13 @@ def makeRepo(root: str, config: Union[Config, str, None] = None, standalone: boo
...
-            s3 = boto3.resource("s3")
-            # implies bucket exists, if not another level of checks
-            bucket = s3.Bucket(uri.netloc)
-            bucket.put_object(Bucket=uri.netloc, Key=uri.relativeToPathRoot)
+            # implies bucket already exists
+            s3 = boto3.client("s3")
+            if uri.relativeToPathRoot != ".":
+                # to be dir-equivalent, S3 this requires URIs ending on '/'
+                if not uri.geturl().endswith('/'):
+                    uri.updateFile(uri.relativeToPathRoot+'/')
```
timj: If ButlerURI had a constructor flag along the lines of `forceDirectory=True` (defaulting to false) then maybe we wouldn't need these hacks.

python/lsst/daf/butler/_butler.py
DinoBektesevic: `updateFile` will update and remember the addition of `/` to the uri, this manipulates the `uri.path` attribute. The `uri.relativeToPathRoot` strips the left hand side separator again so i have to add it again.

I can not use the `uri.path` option because the named directory will be precluded by a `/`. If I try inserting this what I will get is a key named `/` under which I will have a key named `dir` (`dir` will be file-like in this case). In addition, if you try to inset keys starting with `/` from the web interface to S3 it will complain that it's not allowed. 

The following, hopefully, demonstrates this better than I can put it into words: 

```import lsst.daf.butler as dafButler
uri = dafButler.ButlerURI("s3://bucket/dir")
print(uri)
print(uri.relativeToPathRoot)
print(uri.path)
```
outputs:
```
's3://bucket/dir'
'dir'
'/dir'
```
After updating the `path` attribute via `updateFile:
```
uri.updateFile(uri.relativeToPathRoot+'/')
print(uri)
print(uri.relativeToPathRoot)
print(uri.path)
```
outputs
```
's3://bucket/dir/'
'dir'
'/dir/'
```

The end conclusion is that I need to know in advance whether I'm pointing to a directory, in which case I need no leading separators but I do require a trailing separator, or if I'm pointing to a file in which case I don't need any separators except the ones that define the tree structure. 

I update the URI to append the separator in case that at some point later anyone uses that URI to construct other URIs they don't get burned when they use `updateFile` to find out their root directory in the URI disappeared. My thought process was that this is one of the rare places where it's safe to acknowledge that we expect a directory so ensuring that is explicitly written to the URI could not hurt. 

To also be explicit about another point: I do not need that `s3.put_object` call there explicitly. As long as that base URI is updated appropriately, the root directory will implicitly be created during first `butler.put` call. That will call `S3Datastore.put` which will put some relative path of the form `root/possibly_yet_another_dir/filename.ext` when all 3 keys will be created: 2 directories `root/` and `root/possibly_yet_another_dir/` and the file key `root/possibly_yet_another_dir/filename.ext`. This actually should occur *almost* every time because *almost* every time there is a `root/butler.yaml` key being inserted. I was afraid of the `outfile` parameter however so I didn't dare to leave it to chance.

timj: Any time I see a `/` being added explicitly to a path it indicates a hack.  There are two things we know from makeRepo:

1. The URI must refer to a "directory" and not a file.
2. You don't know if the butler.yaml will be written to that location but the root URI will be rewritten into that config if it's relocated.

So it seems to me that changing the ButlerURI code to support something like:

```python
uri = ButlerURI(root, forceDirectory=True)
```
would fix your problems?  If you say it's a directory we always add a `/` to the end of the string in the constructor if there isn't one there already and you get exactly what you are doing her but without the hack. You also guarantee a "directory" like URI is written to the output config if the output config is written somewhere else.  You can also ensure that `relativeToPathRoot` always includes a trailing slash since the URI will know it's a directory URI (if forceDirectory is false (the default) it means you can't tell -- you can still set the isDirectory internal flag if you see a trailing slash).

On a related note I think the reason you can't get `<butlerRoot>` to work properly is that you aren't setting `self.configFile` in `__initFromS3YamlFile`.

DinoBektesevic: Sorry it took me a while, I was swamped with Uni-related work and just couldn't catch up. I made the fixes requested. Hope you like the changes. 

Specifics I'm concerned about: 
1) did I guess all the places where it's safe to assume that I can `forceDirectory` 
2) do you like the `Butler.makeRepo` better like this 

I changed test_location.py because I had a hard time keeping up with what is testing what and whether all cases are covered. I hope you don't mind the re-arrangement of the test data. Mostly from what I've seen it's all my mess originally anyhow. All the same original tests are kept except 1 and some new ones added. I can revert if required.

TallJimbo (approved):
(Trivial) Registry changes look good to me, and I've only quickly scanned the rest - I'm happy to leave it to you and @timj to work out the URI handling, since I historically haven't followed that closely and would need to spend some time getting up to speed on it to understand the subtleties.

python/lsst/daf/butler/core/location.py
```diff
@@ -159,7 +169,11 @@ def relativeToPathRoot(self):
             p = PurePath(self.path)
         else:
             p = PurePosixPath(self.path)
-        return str(p.relative_to(p.root))
+        if self.dirLike:
+            relToRoot = str(p.relative_to(p.root)) + '/'
+        else:
+            relToRoot = str(p.relative_to(p.root))
```
timj: I think that rather than duplicating the bit of code that does all the work I'd prefer:
```python
relToRoot = str(p.relative_to(p.root))
if self.dirLike and not relToRoot.endswith("/"):
    relToRoot += "/"
```

python/lsst/daf/butler/core/location.py
```diff
@@ -113,17 +113,27 @@ class ButlerURI:
...
             raise ValueError("Supplied URI must be either string or ParseResult")
 
-        parsed = self._fixupFileUri(parsed, root=root, forceAbsolute=forceAbsolute)
+        parsed, dirLike = self._fixupFileUri(parsed, root=root,
+                                             forceAbsolute=forceAbsolute,
+                                             forceDirectory=forceDirectory)
+
+        # relativeToPathRoot indiscriminately strips trailing separators.
+        # Record whether we have trailing separator (dir-like) or not. Note:
+        # updated on ButlerURI.updateFile call
```
timj: I don't think this note about updateFile is necessary.

python/lsst/daf/butler/core/location.py
```diff
@@ -255,8 +277,15 @@ def _fixupFileUri(parsed, root=None, forceAbsolute=False):
         to be turned into absolute paths before they can be used.  This is
         always done regardless of the ``forceAbsolute`` parameter.
 
+        AWS S3 differentiates between keys with trailing POSIX separators (i.e
+        `/dir` and `/dir/`) whereas POSIX does not neccessarily. This must be
+        tracked manually if network location is to be treated interchangeably
```
timj: What does this note mean by "tracked manually".  I thought we were tracking this in this class. The note implies the reader has to do something explicit but I don't know what.

python/lsst/daf/butler/core/location.py
```diff
@@ -285,37 +314,59 @@ def _fixupFileUri(parsed, root=None, forceAbsolute=False):
...
+                    replacements["path"] += sep
+                    dirLike = True
             else:
                 raise RuntimeError("Unexpectedly got confused by URI scheme")
 
             # ParseResult is a NamedTuple so _replace is standard API
             parsed = parsed._replace(**replacements)
 
-        return parsed
+        elif parsed.scheme == "s3":
```
timj: Isn't this code path generically true for all non-file URIs? I don't think it needs to be limited to s3.

tests/config/basic/s3Datastore.yaml
```diff
@@ -6,6 +6,7 @@ datastore:
     calexp: "{run}/{datasetType}.{component:?}/{datasetType}_v{visit}_f{physical_filter:?}_{component:?}"
     metric: "{run}/{datasetType}.{component:?}/{instrument:?}_{datasetType}_v{visit:08d}_f{physical_filter}_d{detector:?}_{component:?}"
     test_metric_comp: "{run}/{datasetType}.{component:?}/{datasetType}_v{visit:08d}_f{instrument}_{component:?}"
+    metric2: "{run}/{datasetType}.{component:?}/{tract:?}/{patch:?}/{physical_filter:?}/{instrument:?}_{visit.name:?}"
```
timj: Can you change all the ones that start with `{run}` to be `{run:/}` -- they are meant to be doing that but we missed the s3 ones.

tests/test_butler.py
```diff
@@ -696,6 +696,7 @@ def testPutTemplates(self):
         # dependency); should not change template (at least the way we're
         # defining them  to behave now; the important thing is that they
         # must be consistent).
+
```
timj: We don't really need that extra line

timj (approved):
Some minor comments but this does look better

python/lsst/daf/butler/core/location.py
DinoBektesevic: Because `_fixupFileUri` is a static method I wanted to make it clear that anyone that uses it as a `ButlerURI._fixupFileUri(parsed)` needs to be aware of that fact and consider whether the returned `dirLike` flag is important for them or not.

python/lsst/daf/butler/core/location.py
DinoBektesevic: I believe the conditionals cover all 3 scenarios we explicitly consider - no scheme, explicit file scheme and s3 scheme. I was afraid of assuming a posix separator as a default for any future schemes.

tests/config/basic/s3Datastore.yaml
DinoBektesevic: I can do that. I do have a question though. 

The `posixDatastore.yaml` and `s3Datastore.yaml` in `daf_butler/tests/config` both have entries of the type `{run}`.                        
The `posixDatastore.yaml` and `s3Datastore.yaml` in `daf_butler/config` both have entries of the type `{run:/}`. 

Do you want me to change the entries in `daf_butler/tests/config/posixDatastore.yaml` too? There are multiple different `posixDatastore*.yaml` files in there, should I edit those as well? To boot there are `{run}` type entries in `templates-withdefault.yaml`, `templates-nodefault*.yaml` too. Were those also meant to be updated?

As it stands no configs in `daf_butler/tests/config/` use the `{run:/}` template.

timj: They are meant to -- we mostly didn't notice. Change them all if you like.

python/lsst/daf/butler/core/location.py
timj: I believe a `/` path separator is part of the URI standard so you can assume that -- scheme-less is special and has to be dealt with differently but everything else should be the same shouldn't it?

tests/test_location.py
```diff
@@ -106,7 +140,6 @@ def testRelativeRoot(self):
 
         pathInStore = "relative/path/file.ext"
         loc1 = factory.fromPath(pathInStore)
-
```
timj: Can you put this line back (and squash it) to minimize diff? Removing the line doesn't give us anything.

DinoBektesevic: I don't know where they keep going, these empty lines that is. Thanks!

python/lsst/daf/butler/core/location.py
timj: Also note that this method is called fixUpFileURI so that's not a helpful name any more. It looks like the os2posix won't trigger for S3 anyhow so you can probably use the code above without a special not-file URI else.

DinoBektesevic: I removed else, renamed to fixupPathURI. Does that work for you?

python/lsst/daf/butler/core/config.py
```diff
@@ -251,6 +251,7 @@ def __initFromFile(self, path):
                 self.__initFromYamlFile(uri.ospath)
         else:
             raise RuntimeError("Unhandled config file type:%s" % uri)
+        self.configFile = str(path)
```
timj, line 253: This might fix your butlerRoot woes.

python/lsst/daf/butler/core/location.py
```diff
@@ -285,37 +310,57 @@ def _fixupFileUri(parsed, root=None, forceAbsolute=False):
...
+        # URI is dir-like if explicitly stated or if it ends on a separator
+        endsOnSep = parsed.path.endswith(posixpath.sep)
+        if forceDirectory or endsOnSep:
+            dirLike = True
+            # only add the separator if it's not already there
+            if not endsOnSep:
+                parsed = parsed._replace(path=parsed.path+posixpath.sep)
+
+        if dirLike is None:
+            raise RuntimeError("ButlerURI.dirLike attribute not set succesfully.")
```
timj: Typo successfully.

python/lsst/daf/butler/core/location.py
timj: Please remove the final sentence. I still find it confusing and we already know this is a static method so people need to track the return values.

timj (approved):
looks good.

python/lsst/daf/butler/core/location.py
DinoBektesevic: Removed.

python/lsst/daf/butler/core/config.py
DinoBektesevic: You are right that butler root was not correctly inferred or propagated from butler config file.  That is now fixed. 

I don't think there is a reasonable solution for default values, such as `curdir` for PosixDatastore, without making Butler capable of creating, at the very least, new buckets from scratch. I've always avoided doing it because the way permissions, regions, availability zones, VPCs etc. are set up matters. That means I could, eventually, be in a situation where someone might want to be able to specify these various details during Butler.makeRepo. I keep asking whether this is eventually wanted or not, but insofar it feels like a "no" to me.

timj: I don't really follow you. ButlerRoot exists as a convenience to allow you to say that the datastore is located relative to where the butler.yaml file was found. That's it. It's not meant to be about creating new buckets.  If the butler.yaml is on a local disk and the datastore is in S3 then your butler.yaml would not use butlerRoot for the datastore.root entry.

DinoBektesevic: Maybe I just got confused, up to recently `repoRelocation` would also assume current working directory was the default for `butlerRoot` and that made a lot of commands like `makeButlrRepo.py` work. I see this was changed end of November. I guess I was not paying attention.

DinoBektesevic:
After talking with Tim it was decided a clearer way than the `ButlerURI.updateFile("")` was wanted so `ButlerURI.split, `ButlerURI.basename` and `ButlerURI.dirname` was added. 

Extra handling of weird edge-cases added.
Fixes for certain situations where URI can, by implication, only be a directory were made. 

This finally allows for `<butlerRoot>` tags in configs to be properly replaced by config file location. URIs without trailing `/` can finally be used to instantiate a butler too (f.e. `Butler("s3://bucket/root")`)

DinoBektesevic:
> You also don't want to be trying to reconstruct the URI string like that because you will miss other bits like fragments.

Maybe I'm just slightly confused as to how are fragments meant to be used if we're missing the basename of the URI, but here are the desired changes. I thought magically expanding head value might be confusing when compared to `os.path.split` behaviour but I can see why making sure it's easy to round-trip through ButlerURI is important too.

PR #248, opened 2020-04-01
DM-13403: numpy types fail in butler dataIds

andy-slac:
At least one database backend cannot handle `numpy.int64` type which is
a subclass of `numbers.Integral`. This patch replaces all values of
Integral type in a DataCoordinate with ints.


TallJimbo (approved):
Looks good.  I approve of `DataCoordinate.standardize` as the place to put this - while there may be other ways one could pass numpy types into the database, that will cover all data IDs (and hence the vast majority of places one is _likely_ to pass a numpy type), and it might save us from non-database trouble with `instance(numpy.int64, int)` being `False` in the future.

PR #249, opened 2020-04-02
DM-24285: fitsExposureFormatter fails to read "Exposure" entries correctly

czwa:
fitsExposureFormatter now uses the afw.image.ExposureFitsReader methods to access the full exposure as well as subcomponents.

### DIFF
--- a/python/lsst/daf/butler/formatters/fitsExposureFormatter.py
+++ b/python/lsst/daf/butler/formatters/fitsExposureFormatter.py
@@ -42,27 +42,6 @@ class FitsExposureFormatter(Formatter):
             self._metadata = self.readMetadata()
         return self._metadata
 
-    def readImageComponent(self, component):
-        """Read the image, mask, or variance component of an Exposure.
-
-        Parameters
-        ----------
-        fileDescriptor : `FileDescriptor`
-            Identifies the file to read and parameters to be used for reading.
-        component : `str`, optional
-            Component to read from the file.  Always one of "image",
-            "variance", or "mask".
-
-        Returns
-        -------
-        image : `~lsst.afw.image.Image` or `~lsst.afw.image.Mask`
-            In-memory image, variance, or mask component.
-        """
-        # TODO: could be made more efficient *if* Exposure type objects
-        # held the class objects of their components.
-        full = self.readFull()
-        return self.fileDescriptor.storageClass.assembler().getComponent(full, component)
-
     def readMetadata(self):
         """Read all header metadata directly into a PropertyList.
 
@@ -97,24 +76,53 @@ class FitsExposureFormatter(Formatter):
         bboxFromMetadata(self.metadata)  # always strips
         makeSkyWcs(self.metadata, strip=True)
 
-    def readInfoComponent(self, component):
-        """Read a component held by ExposureInfo.
+    def readComponent(self, component, parameters=None):
+        """Read a component held by the Exposure.
 
         Parameters
         ----------
         component : `str`, optional
             Component to read from the file.
+        parameters : `dict`, optional
+            If specified, a dictionary of slicing parameters that
+            overrides those in ``fileDescriptor``.
 
         Returns
         -------
         obj : component-dependent
             In-memory component object.
+
+        Raises
+        ------
+        KeyError
+            Raised if the requested component cannot be handled.
         """
-        from lsst.afw.image import LOCAL
-        from lsst.geom import Box2I, Point2I
-        parameters = dict(bbox=Box2I(minimum=Point2I(0, 0), maximum=Point2I(0, 0)), origin=LOCAL)
-        tiny = self.readFull(parameters)
-        return self.fileDescriptor.storageClass.assembler().getComponent(tiny, component)
+        componentMap = {'wcs': ('readWcs', False),
+                        'coaddInputs': ('readCoaddInputs', False),
+                        'psf': ('readPsf', False),
+                        'image': ('readImage', True),
+                        'mask': ('readMask', True),
+                        'variance': ('readVariance', True),
+                        }
+        method, hasParams = componentMap.get(component, None)
+
+        if method:
+            reader = ExposureFitsReader(self.fileDescriptor.location.path)
+            caller = getattr(reader, method, None)
+
+            if caller:
+                if parameters is None:
+                    parameters = self.fileDescriptor.parameters
+                if parameters is None:
+                    parameters = {}
+                self.fileDescriptor.storageClass.validateParameters(parameters)
+
+                if hasParams and parameters:
+                    return caller(**parameters)
+                else:
+                    return caller()
+        else:
+            raise KeyError(f"Unknown component requested: {component}")
 
     def readFull(self, parameters=None):
         """Read the full Exposure object.
@@ -140,10 +148,10 @@ class FitsExposureFormatter(Formatter):
             output = fileDescriptor.storageClass.pytype(fileDescriptor.location.path, **parameters)
         except TypeError:
             reader = ExposureFitsReader(fileDescriptor.location.path)
-            output = reader.read()
+            output = reader.read(**parameters)
         return output
 
-    def read(self, component=None):
+    def read(self, component=None, parameters=None):
         """Read data from a file.
 
         Parameters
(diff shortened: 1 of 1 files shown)


python/lsst/daf/butler/formatters/fitsExposureFormatter.py
```diff
@@ -109,11 +89,19 @@ def readInfoComponent(self, component):
...
+                        'coaddInputs': 'readCoaddInputs',
+                        'psf': 'readPsf',
+                        'image': 'readImage',
+                        'mask': 'readMask',
+                        'variance': 'readVariance',
+                        }
+        method = componentMap.get(component, None)
+        if method:
+            caller = getattr(reader, method, None)
+        if caller:
```
timj: I think this needs to be indented since if method is false caller will never be defined.

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
```diff
@@ -135,7 +123,12 @@ def readFull(self, parameters=None):
         if parameters is None:
             parameters = {}
         fileDescriptor.storageClass.validateParameters(parameters)
-        return fileDescriptor.storageClass.pytype(fileDescriptor.location.path, **parameters)
+        try:
+            output = fileDescriptor.storageClass.pytype(fileDescriptor.location.path, **parameters)
+        except TypeError:
+            reader = ExposureFitsReader(fileDescriptor.location.path)
+            output = reader.read(parameters)
```
timj: `**parameters` ?

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
```diff
@@ -96,8 +76,8 @@ def stripMetadata(self):
         bboxFromMetadata(self.metadata)  # always strips
         makeSkyWcs(self.metadata, strip=True)
 
-    def readInfoComponent(self, component):
-        """Read a component held by ExposureInfo.
+    def readComponent(self, component):
```
timj: I think this is going to have to take a parameters argument so that you can do things like pass a bounding box to readMask.

timj (commented):
Looks good but a couple of things need to be fixed. I'll leave the real review to @TallJimbo since he knows more about the afw bits.

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
```diff
@@ -109,11 +89,19 @@ def readInfoComponent(self, component):
         obj : component-dependent
             In-memory component object.
```
timj, line 93: Should this method raise an exception if the requested component is not understood? Rather than returning None?

czwa: Now raising KeyError.

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
czwa: Added here, and to read(), with appropriate handling to ensure parameters are not passed to methods that do not expect parameters.

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
```diff
@@ -96,24 +76,46 @@ def stripMetadata(self):
...
+
+        Raises
+        ------
+        KeyError
+            Raised if the requested component cannot be handled.
         """
-        from lsst.afw.image import LOCAL
-        from lsst.geom import Box2I, Point2I
-        parameters = dict(bbox=Box2I(minimum=Point2I(0, 0), maximum=Point2I(0, 0)), origin=LOCAL)
-        tiny = self.readFull(parameters)
```
TallJimbo, line 115: For the image, mask, and variance components, this now might return a pixel type other than the one declared by the StorageClass when that is something more specific than just `Exposure`.  I think that might be what the comment in the old code about "Exposure exposing the type of its image components" was getting at, and if so, that might require changes in afw to address (i.e. we might not have a machine way to discover that the image of an ExposureF is an ImageF, so we can't easily cast).

That would make this ticket a much bigger pain to fix, but it also may be a more minor problem than what this ticket already fixes, and hence might be a good trade if we open another ticket to deal with it later.  The question is whether this breaks Jenkins.

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
```diff
@@ -96,24 +76,46 @@ def stripMetadata(self):
...
+                        'variance': ('readVariance', True),
+                        }
+        method, hasParams = componentMap.get(component, None)
+
+        if method:
+            reader = ExposureFitsReader(self.fileDescriptor.location.path)
+            caller = getattr(reader, method, None)
+            if caller:
+                if hasParams and parameters:
+                    return caller(parameters)
```
timj: `**parameters`?

TallJimbo (approved):
One comment that _probably_ just needs to spawn a new ticket; Jenkins will tell us if it needs to be a blocker for this one.  Other than that I'd just echo Tim's comments.

python/lsst/daf/butler/formatters/fitsExposureFormatter.py
timj: It's all fine if you are using Exposure and Image everywhere in your storage classes. Note though that you know the python type so can't you do a cast immediately on read if the type you read isn't the same as the pytype that you expected? Can you construct an ImageF from an ImageI?

timj: I'd use `isinstance` rather than checking that the classes are identical -- this will let you handle `Image` and `ImageF`.

TallJimbo: > It's all fine if you are using Exposure and Image everywhere in your storage classes.

...and that's only fine if the PipelineTasks never expect the on-disk type to be coerced to a different one.  I seem to recall this being the case in ISR for some instruments, though I may be mixing up image->exposure conversion with pixel type conversion (though even that may be a problem here).

timj: but can you cast an ImageI to an ImageF? If you can then there's no problem.

RobertLuptonTheGood:
`imageI.convertF()`

timj:
@czwa it seems like it should be easy to deal with @TallJimbo 's concerns by using the `convertX` methods. Are you going to do that on this ticket?

TallJimbo:
Not _that_ easy, unless you hard-code facts like "you want `image.convertD()` to convert `Exposure.image` to `ExposureD.image` but `image.convertF()` to convert `Exposure.variance` to `ExposureD.variance`".  Those are pretty stable facts, but daf_butler doesn't seem like the right place to put them (which would be afw).

timj:
Oh. You are talking about a very very different thing to what I thought you were talking about. I was talking about the fact that ExposureFitsReader doesn't know the requested type for a component and so in theory you can get an ImageF but you had requested an ImageI. We can definitely solve that because we know that the StorageClass wants an ImageI python type so we know that we have to call convertI on it.   You seem to be talking about constructing Exposures with different types inside?  By default we get the exposure of the type you ask for and fall back to ExposureFitsReader if you were generically asking for Exposure. I think I'm confused a bit about your real issue.

TallJimbo:
@timj I was talking about the same thing you originally thought I was talking about, but my original wording was a bit confusing (hopefully the editing made it better).  If we've already (effectively) duplicating the afw component types in the storage class definitions (which is obvious now that I think about it), then I agree this is not nearly as bad as I thought.

czwa:
I was going to defer the typing issue to a new ticket, as I'm not entirely certain how this happens.  Is this an introduced issue from this ticket?  I thought the checking in FitsExposureFormatter.read() would raise if the storageClasses didn't match (and no component was requested).

I'm running Jenkins on this ticket branch, and would like to get it resolved so I can get back to the ISR/DECam/gen3 crosstalk issue that prompted the bugfix.

timj:
The discussion relates directly to the new code in that previously you read an ExposureF and then returned by definition an ImageF.  Now that we are using native type for reading the image component there is a risk that your storage class says you want ImageF but your readImage returns an ImageI.  This will cause the butler to raise an exception.  My point is that you already know how to fix this because you know that the pytype is ImageF and that ImageI is not an instance of ImageF.  You can have a dict that maps pytype to convert method and if `not isinstance` you call that convert method and return the result.

PR #250, opened 2020-04-02
DM-24314: Read through symlinks when ingesting as symlinks

timj:
This prevents us from having a repository that has multiple indirections to the real dataset.

### DIFF
--- a/tests/test_datastore.py
+++ b/tests/test_datastore.py
@@ -362,12 +362,16 @@ class DatastoreTests(DatastoreTestsBase):
         with self.assertRaises(FileNotFoundError):
             datastore.get(refInner)
 
-    def runIngestTest(self, func, expectOutput=True):
+    def _prepareIngestTest(self):
         storageClass = self.storageClassFactory.getStorageClass("StructuredData")
         dimensions = self.universe.extract(("visit", "physical_filter"))
         metrics = makeExampleMetrics()
         dataId = {"instrument": "dummy", "visit": 0, "physical_filter": "V"}
         ref = self.makeDatasetRef("metric", dimensions, storageClass, dataId, conform=False)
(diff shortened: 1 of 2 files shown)


kfindeisen (approved):
The change itself looks good. Why can't you create a unit test for it? Is it because every Gen 2 repository needs to be associated with an `obs` package?

timj:
Yes, obs packages are where a lot of the ingest stuff happens. I'll take a look to see if the butler ingest tests of symlink can be modified to make a double symlink.

tests/test_datastore.py
```diff
@@ -447,6 +451,24 @@ def failNotImplemented(obj, path, ref):
...
+            return
+        metrics, ref = self._prepareIngestTest()
+        with lsst.utils.tests.getTempFilePath(".yaml") as realpath:
+            with open(realpath, 'w') as fd:
+                yaml.dump(metrics._asdict(), stream=fd)
+            with lsst.utils.tests.getTempFilePath(".yaml") as sympath:
+                os.symlink(os.path.abspath(realpath), sympath)
+                datastore = self.makeDatastore()
+                datastore.ingest(FileDataset(path=os.path.abspath(sympath), refs=ref), transfer=mode)
+                uri = ButlerURI(datastore.getUri(ref))
```
kfindeisen: Consider adding blank lines as follows:
```suggestion
                os.symlink(os.path.abspath(realpath), sympath)

                datastore = self.makeDatastore()
                datastore.ingest(FileDataset(path=os.path.abspath(sympath), refs=ref), transfer=mode)

                uri = ButlerURI(datastore.getUri(ref))
```
I think that would make it easier to follow the logic (setup, operation being tested, tests).

PR #251, opened 2020-04-03
DM-24244: fix Oracle compatibility and reduce need for quoting

TallJimbo:


MichelleGower (approved):
With these changes was able to run ci_hsc_gen3 with Oracle.

PR #252, opened 2020-04-04
DM-24352: Add new auto and link transfer modes

timj:
* Auto selects the default based on context
* link tries hardlink and falls back to symlink

@DinoBektesevic in theory with this ticket you should be able to do a 2 to 3 conversion directly to S3.

### DIFF
--- a/python/lsst/daf/butler/datastores/posixDatastore.py
+++ b/python/lsst/daf/butler/datastores/posixDatastore.py
@@ -222,6 +222,47 @@ class PosixDatastore(FileLikeDatastore):
         info = self._extractIngestInfo(path, ref, formatter=formatter)
         self._register_datasets([(ref, info)])
 
+    def _overrideTransferMode(self, *datasets: FileDataset, transfer: Optional[str] = None) -> str:
+        # Docstring inherited from base class
+        if transfer != "auto":
+            return transfer
+        try:
+            # Assume first dataset is representative
+            self._pathInStore(datasets[0].path)
+        except RuntimeError:
+            transfer = "symlink"
+        else:
+            transfer = None
+        return transfer
+
+    def _pathInStore(self, path: str) -> str:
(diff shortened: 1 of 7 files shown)


python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -222,20 +222,55 @@ def _removeFileExists(path):
         info = self._extractIngestInfo(path, ref, formatter=formatter)
         self._register_datasets([(ref, info)])
 
+    def _overrideTransferMode(self, *datasets: FileDataset, transfer: Optional[str] = None) -> str:
+        # Docstring inherited from base class
+        if transfer != "auto":
+            return transfer
+        try:
+            # Assume first dataset is representative
+            self._pathInStore(datasets[0].path)
```
TallJimbo: This seems like it could yield pretty surprising behavior if the assumption is incorrect.  Better to check them all and disallow "auto" if the answer is inconsistent, even if that could be a lot of overhead.  After all, it's always possible to avoid that overhead by being explicit about the transfer mode.

An alternative would be to just make in-place the default so it's an error if the file is outside.  I'm *slightly* concerned that having symlinks be any kind of default (even the "default of last resort") could be bad, because it makes for a fragile repository.

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -222,20 +222,55 @@ def _removeFileExists(path):
...
+
+        Returns
+        -------
+        inStore : `str`
+            Path relative to datastore root.
+
+        Raises
+        ------
+        RuntimeError
+            Raised if the supplied path is outside the datastore.
```
TallJimbo: Given that in at least one context this is a totally normal condition (as opposed to an error) it might be better to return `None`.

TallJimbo (approved):

PR #253, opened 2020-04-06
DM-24365: Add relsymlink ingest transfer mode

timj:


python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -72,8 +72,8 @@ class ChainedDatastore(Datastore):
     Notes
     -----
     ChainedDatastore never supports `None` or `"move"` as an `ingest` transfer
-    mode.  It supports `"copy"`, `"symlink"`, and `"hardlink"` if and only if
-    its child datastores do.
+    mode.  It supports `"copy"`, `"symlink"`, `"relsymlink"`
+    and `"hardlink"` if and only if its child datastores do.
```
kfindeisen: Should this be "if and only if [all] its child datastores do"?

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -63,7 +63,8 @@ class PosixDatastore(FileLikeDatastore):
     Notes
     -----
     PosixDatastore supports all transfer modes for file-based ingest:
-    `"move"`, `"copy"`, `"symlink"`, `"hardlink"`, and `None` (no transfer).
+    `"move"`, `"copy"`, `"symlink"`, `"hardlink"`, `"relsymlink"`
+    and `None` (no transfer).
```
kfindeisen, line 67: There's a lot of documentation duplication on this ticket to keep track of all the strings. I wonder if it might be worth replacing these arguments with an `enum.Enum` in internal APIs like `Datastore`. On another ticket, of course.

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -289,8 +290,9 @@ def _extractIngestInfo(self, path: str, ref: DatasetRef, *, formatter: Type[Form
                 raise FileExistsError(f"File '{newFullPath}' already exists.")
             storageDir = os.path.dirname(newFullPath)
             if not os.path.isdir(storageDir):
-                with self._transaction.undoWith("mkdir", os.rmdir, storageDir):
-                    safeMakeDir(storageDir)
+                # Do not attempt to reverse directory creation
+                # because of race conditions
```
kfindeisen: Might help to say what this code is potentially racing with.

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -312,6 +314,16 @@ def _extractIngestInfo(self, path: str, ref: DatasetRef, *, formatter: Type[Form
                 with self._transaction.undoWith("symlink", os.unlink, newFullPath):
                     # Read through existing symlinks
                     os.symlink(os.path.realpath(fullPath), newFullPath)
+            elif transfer == "relsymlink":
+                # This is a standard symlink but using a relative path
+                fullPath = os.path.realpath(fullPath)
+
+                # Need the directory name to give to relative root
+                # A full file path confuses it into an extra ../
+                newFullPathRoot, _ = os.path.split(newFullPath)
```
kfindeisen, line 324: This seems a bit unsafe. It works as advertised only if `newFullPath` is either a `/`-terminated directory or a file. Can you guarantee that (i.e., is the output of `predictPathFromLocation` guaranteed to be in that form)?

kfindeisen (approved):
Looks good, mostly documentation questions, plus one possible bug.

python/lsst/daf/butler/datastores/posixDatastore.py
timj: Yes, I was starting to wonder about all the transfer modes.

python/lsst/daf/butler/datastores/posixDatastore.py
timj: It's racing with some other process doing an ingest itself. In theory you can have many jobs talking to one butler.

python/lsst/daf/butler/datastores/posixDatastore.py
timj: newFullPath is by definition an absolute path to a file. It's never a directory. It's the file that will be written to the datastore file system.

PR #254, opened 2020-04-07
DM-15536: add functionality needed to separate visit definition from ingest.

TallJimbo:


python/lsst/daf/butler/core/dimensions/elements.py
```diff
@@ -88,24 +175,20 @@ class DimensionElement:
...
         self.metadata.freeze()
         self.cached = cached
         self.viewOf = viewOf
+        self.alwaysJoin = alwaysJoin
 
-    def _finish(self, universe: DimensionUniverse):
+    def _finish(self, universe: DimensionUniverse, elementsToDo: Dict[str, DimensionElement]):
         """Finish construction of the element and add it to the given universe.
 
         For internal use by `DimensionUniverse` only.
```
timj, line 194: Even if it's internal I'd still like to know what elementsToDo is.

python/lsst/daf/butler/core/timespan.py
```diff
@@ -35,31 +35,116 @@
...
+
+        Returns
+        -------
+        overlaps : `Any`
+            The result of the overlap.  When ``ops`` is `operator`, this will
+            be a `bool`.  If ``ops`` is `sqlachemy.sql`, it will be a boolean
+            column expression.
+        """
+        return ops.and_(
+            ops.or_(self.end == None, other.begin == None, self.end >= other.begin), # noqa E711
```
timj: I think flake8 requires this to be `noqa: E711` otherwise you noqa everything. The colon is required for specific noqa.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -79,6 +89,74 @@
     )
 
 
+class DataIdRelationship(enum.Flag):
+    """Flag enumeration used to report relationships between data IDs by
+    `Registry.relateDataIds`.
+
+    The `DataIdRelationship` value for a pair of data IDs coerces to `True` in
+    boolean contexts if and only if `Registry.queryDimensions` would yield one
+    data ID when passed the the other as a constraint (via its ``dataId``
```
timj: Typo: the the

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -744,7 +822,8 @@ def insertDatasets(self, datasetType: Union[DatasetType, str], dataIds: Iterable
         # Expand data IDs and build both a list of unresolved DatasetRefs
         # and a list of dictionary rows for the dataset table.
         for dataId in dataIds:
-            ref = DatasetRef(datasetType, self.expandDataId(dataId, graph=datasetType.dimensions))
+            ref = DatasetRef(datasetType, self.expandDataId
+                             (dataId, graph=datasetType.dimensions))
```
timj: Where did this new line come from?

timj (approved):
I can't say that I understood all the dimensions changes but in general it looks okay.

python/lsst/daf/butler/core/timespan.py
TallJimbo: Thanks; I always forget which to use.  But while fixing this I discovered a handy trick to remind me in the future: if you get it right, flake8 will _also_ tell you if you only put one space before the comment instead of two, and if you get it wrong, it won't.

python/lsst/daf/butler/registry/_registry.py
TallJimbo: Must have been an errant keystroke, and I'm surprised flake8 didn't tell me about it.

tests/test_datasets.py
```diff
@@ -218,7 +218,8 @@ def testConstructor(self):
         # Construct an unresolved ref.
         ref = DatasetRef(self.datasetType, self.dataId)
         self.assertEqual(ref.datasetType, self.datasetType)
-        self.assertEqual(ref.dataId, self.dataId, msg=ref.dataId)
+        self.assertEqual(ref.dataId, DataCoordinate.standardize(self.dataId, universe=self.universe),
+                         msg=ref.dataId)
```
natelust, line 222: it seems like before this test was designed to test the special behavior in __eq__. As that is not longer the case, is there any reason for this line of the test at all? Was DataCoordinate.satandardize not tested before?

tests/test_datasets.py
```diff
@@ -238,7 +239,8 @@ def testConstructor(self):
         # as well as everything else.
         ref = DatasetRef(self.datasetType, self.dataId, id=1, run=run, components=components)
         self.assertEqual(ref.datasetType, self.datasetType)
-        self.assertEqual(ref.dataId, self.dataId, msg=ref.dataId)
+        self.assertEqual(ref.dataId, DataCoordinate.standardize(self.dataId, universe=self.universe),
+                         msg=ref.dataId)
```
natelust, line 243: same as above

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -80,6 +89,74 @@
     )
 
 
+class DataIdRelationship(enum.Flag):
```
natelust: Consider using typing.Optional and a dataclass for a cleaner interface

natelust (approved):
I think I am ok with this, noting that the implementation details will depend on changing the DataIdRelationship that we talked about 1:1. If you want me to look at how things shake out when you have made that change I will, but the algorithms seem ok.

tests/test_datasets.py
TallJimbo: Its value has gone down as it's become a more circular test, but I don't think it's gone down to zero, as it's still going through some logic in the `DatasetRef` constructor.

PR #255, opened 2020-04-07
DM-24375: Fix time precision issue in type decorator

andy-slac:
Precision of a Python float type is not enough to guarantee nanosecond
precision, need more careful handling when converting between astropy
time and nanoseconds.

### DIFF
--- a/python/lsst/daf/butler/core/ddl.py
+++ b/python/lsst/daf/butler/core/ddl.py
@@ -170,15 +170,25 @@ class AstropyTimeNsecTai(sqlalchemy.TypeDecorator):
             _LOG.warning("%s is later than max. time %s, max. time time will be used instead",
                          value, MAX_TIME)
             value = MAX_TIME
-        value = round((value - EPOCH).to_value("sec") * 1e9)
-        return int(value)
+
+        delta = value - EPOCH
+        # Special care needed to preserve nanosecond precision.
+        # Usually jd1 has no fractional part but just in case.
+        jd1, extra_jd2 = divmod(delta.jd1, 1)
+        nsec_per_day = 1_000_000_000 * 24 * 3600
+        value = int(jd1) * nsec_per_day + int(round((delta.jd2 + extra_jd2)*nsec_per_day))
+        return value
 
     def process_result_value(self, value, dialect):
         # value is nanoseconds since epoch, or None
         if value is None:
             return None
-        delta = astropy.time.TimeDelta(value * 1e-9, format="sec")
-        return EPOCH + delta
+        # Again special care needed to preserve precision
+        nsec_per_day = 1_000_000_000 * 24 * 3600
+        jd1, jd2 = divmod(value, nsec_per_day)
+        delta = astropy.time.TimeDelta(float(jd1), float(jd2)/nsec_per_day, format="jd", scale="tai")
+        value = EPOCH + delta
+        return value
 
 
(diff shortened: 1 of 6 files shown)


python/lsst/daf/butler/core/ddl.py
```diff
@@ -170,15 +170,25 @@ def process_bind_param(self, value, dialect):
...
                          value, MAX_TIME)
             value = MAX_TIME
-        value = round((value - EPOCH).to_value("sec") * 1e9)
-        return int(value)
+
+        delta = value - EPOCH
+        # Special care needed to preserve nanosecond precision.
+        # Usually jd1 has no fractional part but just in case.
+        jd1, extra_jd2 = divmod(delta.jd1, 1)
+        nsec_per_day = 1_000_000_000 * 24 * 3600
```
timj: Can we pull this out as a file constant rather than defining it in both routines?

tests/test_ddl.py
```diff
@@ -0,0 +1,99 @@
...
+        value = self.decor.process_bind_param(time, self.dialect)
+
+        value_max = self.decor.process_bind_param(ddl.MAX_TIME, self.dialect)
+        self.assertEqual(value, value_max)
+
+    def test_round_trip(self):
+        """Test precision of round-trip conversion.
+        """
+        # do tests at random points between epoch and max. time
+        times = [
```
timj, line 73: Can you add some times that use nanosecond precision and also some that are very near the end of a day and not just the start?

timj (approved):
Looks great. A couple of minor comments.

python/lsst/daf/butler/core/ddl.py
andy-slac: Moved everything to a separate module instead 🙂

python/lsst/daf/butler/core/time_utils.py
```diff
@@ -0,0 +1,140 @@
...
+    Parameters
+    ----------
+    time1, time2 : `astropy.time.Time`
+        Times to compare.
+    precision_nsec : `float`, optional
+        Precision to use for comparison in nanoseconds, default is one
+        nanosecond which is larger that round-trip error for conversion
+        to/from integer nanoseconds.
+    """
+    # To compare we need them in common scale, for simlicity just
```
timj: typo: simlicity

tests/test_time_utils.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # 1/2 nsec difference
+        time1 = Time("2000-01-01T00:00:00.123456789", format="isot", scale="tai")
+        time2 = time1 + TimeDelta(0.5e-9, format="sec")
+        self.assertTrue(time_utils.times_equal(time1, time2))
+        self.assertTrue(time_utils.times_equal(time2, time1))
+        self.assertFalse(time_utils.times_equal(time1, time2, .25))
+        self.assertFalse(time_utils.times_equal(time2, time1, .25))
+
+        # 1/2 mksec difference
```
timj: typo: mksec

timj (approved):

tests/test_time_utils.py
andy-slac: Do you prefer μsec? 🙂

timj: I've never seen mk used for micro before so I think I'd prefer (microsec) -- but of course a μ should work fine in the source code so I'm also fine with that...

PR #257, opened 2020-04-10
DM-24434: symlink bug in posixDatastore.py on Ubuntu

parejkoj:


timj (approved):

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -304,13 +304,13 @@ def _extractIngestInfo(self, path: str, ref: DatasetRef, *, formatter: Type[Form
                 with self._transaction.undoWith("link", os.unlink, newFullPath):
                     # Try hard link and if that fails use a symlink
                     try:
-                        os.link(fullPath, newFullPath)
+                        os.link(os.path.realpath(fullPath), newFullPath)
                     except OSError:
                         # Read through existing symlinks
                         os.symlink(os.path.realpath(fullPath), newFullPath)
```
erykoff: I have noticed a non-negligible overhead in `os.path.realpath`, so if this is being done on a lot of files, and is expected to fail often (and if it fails for one file it will fail for all) it might be slightly faster to do the `os.path.realpath()` outside the `try` block and not have to call `realpath` twice.

erykoff (approved):
Looks fine, but could be optimized a bit, see my comment.

PR #258, opened 2020-04-13
DM-24330: add ability to run an obs_base command via the butler command

n8pease:


python/lsst/daf/butler/cli/butler.py
```diff
@@ -0,0 +1,89 @@
...
+def getCli():
+    _addCommands(cli)
+    return cli
+
+
+def main():
+    verbose = len(sys.argv) > 1 and "-v" == sys.argv[1]
+    logging.basicConfig(level=logging.DEBUG if verbose else logging.INFO)
+    log.setLevel(lsst.log.Log.DEBUG if verbose else lsst.log.Log.INFO)
+    return getCli()()
```
n8pease: This file doesn't *have* to be called butler.py; it's called by the file at `bin/butler` which is the file name that causes the butler script to be named butler. This could be named e.g. `butlerCli.py` if that could reduce confusion between this and the file that owns the Butler class (`_butler.py`). I wasn't sure which way to go on that.

doc/lsst.daf.butler/index.rst
```diff
@@ -35,18 +35,9 @@ Using the Butler
...
-
-.. autoprogram:: lsst.daf.butler.script.dumpButlerConfig:build_argparser()
-   :prog: dumpButlerConfig.py
-   :groups:
-
-.. autoprogram:: lsst.daf.butler.script.validateButlerConfiguration:build_argparser()
-   :prog: validateButlerConfiguration.py
-   :groups:
-
+.. click:: lsst.daf.butler.script.butler:cli
```
timj: script.butler or cli.butler?

python/lsst/daf/butler/cli/butler.py
```diff
@@ -0,0 +1,89 @@
...
+
+from lsst.daf.butler.cli.opt.verbose import verbose_option
+import lsst.log
+from lsst.utils import doImport
+
+
+def _getPluginList():
+    pluginModules = os.environ.get("DAF_BUTLER_PLUGINS")
+    if pluginModules is None:
+        return None
```
timj: Shouldn't this return an empty list? Otherwise `_importPlugins` is going to get confused.

python/lsst/daf/butler/cli/butler.py
```diff
@@ -0,0 +1,89 @@
...
+                logging.debug("Exception during add_command: %s", err)
+
+
+def _addCommands(group):
+    pluginList = _getPluginList()
+    _importPlugins(pluginList, group)
+
+
+@click.group()
+# this group doesn't use the verbose command but does allow (ingore) it,
```
timj: Typo: ingore

python/lsst/daf/butler/cli/butler.py
timj: I don't think it really matters given namespacing. @TallJimbo ?

python/lsst/daf/butler/cli/butler.py
```diff
@@ -0,0 +1,89 @@
...
+log = lsst.log.Log.getLogger("lsst.daf.butler")
+
+
+def getCli():
+    _addCommands(cli)
+    return cli
+
+
+def main():
+    verbose = len(sys.argv) > 1 and "-v" == sys.argv[1]
```
timj: I thought click meant that sys.argv wasn't involved.  Are we causing trouble by wanting every command to have a `--verbose` flag?

python/lsst/daf/butler/cli/butler.py
```diff
@@ -0,0 +1,89 @@
...
+    logging.basicConfig(level=logging.DEBUG if verbose else logging.INFO)
+    log.setLevel(lsst.log.Log.DEBUG if verbose else lsst.log.Log.INFO)
```
timj: At some point we might also want `--log=WARN` and related options. See https://github.com/lsst/astro_metadata_translator/blob/master/python/astro_metadata_translator/bin/translateheader.py#L350-L355  (and decide which of --verbose and --log wins)

python/lsst/daf/butler/cli/cmd/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+from .create import *
```
timj: why is `*` being imported when you are only wanting `create` from here anyhow?

python/lsst/daf/butler/cli/cmd/dump_config.py
```diff
@@ -0,0 +1,53 @@
...
+import click
+
+from lsst.daf.butler import ButlerConfig
+
+__all__ = ['dump_config']
+
+
+@click.command(name='dump-config')
+@click.argument('repo')
+@click.option('--subset', '-s', type=str, help='Subset of a configuration to report. This can be any key in '
```
timj: single quotes should be double

python/lsst/daf/butler/cli/cmd/validate_config.py
```diff
@@ -0,0 +1,47 @@
...
+from lsst.daf.butler import Butler, ValidationError
+
+__all__ = ['validate_config']
+
+
+@click.command(name='validate-config')
+@click.argument('repo')
+@click.option('--quiet', '-q', is_flag=True, help="Do not report individual failures.")
+@click.option('--datasettype', '-d', type=str, multiple=True,
+              help="Specific DatasetType(s) to validate (can be comma-separated)")
```
timj: You need to reinstate these lines somewhere for the comma processing.

https://github.com/lsst/daf_butler/blob/master/python/lsst/daf/butler/script/validateButlerConfiguration.py#L110-L112

python/lsst/daf/butler/cli/cmd/validate_config.py
```diff
@@ -0,0 +1,47 @@
...
+
+    REPO is the filesystem path for an existing Butler repository.
+    '''
+    logFailures = not quiet
+    butler = Butler(config=repo)
+    try:
+        butler.validateConfiguration(logFailures=logFailures, datasetTypeNames=datasettype, ignore=ignore)
+    except ValidationError:
+        return False
+    return True
```
timj: Previously there was a message output saying that the configuration looked fine. Otherwise the user sees nothing. That can be a log.info message if you want but I think it would be helpful to retain it.

python/lsst/daf/butler/cli/cmd/dump_config.py
```diff
@@ -0,0 +1,53 @@
...
+              "the hierarchy such as '.datastore.root' where the leading '.' specified the delimiter for "
+              "the hierarchy.")
+@click.option("--searchpath", "-p", type=str, multiple=True,
+              help="Additional search paths to use for configuration overrides")
+@click.option("--file", "outfile", type=click.File("w"), default="-",
```
timj: This is clever.

python/lsst/daf/butler/cli/opt/repo.py
```diff
@@ -0,0 +1,35 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+import click
+
+__all__ = ["repo_option"]
+
+
+class repo_option:  # noqa: N801
```
timj, line 26: are these lower case class names a click style?

tests/test_cliOptionRepo.py
```diff
@@ -0,0 +1,68 @@
...
+    pass
+
+
+@click.command()
+@repo_option()  # required defaul val is False
+def repoNotRequired(repo):
+    pass
+
+
+class CliRepoOptinonSuite(unittest.TestCase):
```
timj: Should this be `Option` rather than `Optinon`?

python/lsst/daf/butler/cli/opt/repo.py
```diff
@@ -0,0 +1,35 @@
...
+    def __init__(self, required=False):
+        self.required = required
+
+    def __call__(self, f):
+        return click.option('--REPO',
```
timj: Does this mean we are insisting on upper case `--REPO` here? I'd much rather it was `--repo`.

tests/test_cliOptionRepo.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+@click.command()
+@repo_option(required=True)
+def repoRequired(repo):
+    pass
+
+
+@click.command()
+@repo_option()  # required defaul val is False
```
timj: typo: defaul

tests/test_commandLine.py
```diff
@@ -0,0 +1,105 @@
...
+        runner = click.testing.CliRunner()
+        with runner.isolated_filesystem():
+            result = runner.invoke(butler.getCli(), ["create", "here"])
+            self.assertEqual(result.exit_code, 0)
+
+            # test dumping to stdout:
+            result = runner.invoke(butler.getCli(), ["dump-config", "here"])
+            self.assertEqual(result.exit_code, 0)
+            # check for some expected keywords:
+            self.assertTrue("composites" in result.stdout)
```
timj: Hmm. Can't we read the output as YAML, parse that, and then check for `composites` in the `dict`?

Also, please use `assertIn`.

ups/daf_butler.table
```diff
@@ -11,3 +11,4 @@ setupRequired(afw)
 
 envPrepend(PATH, ${PRODUCT_DIR}/bin)
 envPrepend(PYTHONPATH, ${PRODUCT_DIR}/python)
+envPrepend(DAF_BUTLER_PLUGINS, lsst.daf.butler.cli.cmd)
```
timj: I don't think we want this for butler itself because the butler already knows where its own scripts are. The butler standard set should always work regardless.

For testing I'd create a trivial command line plugin that echoes back the subcommand name. In the test set the environment variable to include the path to that test code and run the new subcommand and check the output.
Note that you can create a class inline in a test and use getFullTypeName on that class and doImport will work on it (since Python will realize it has already been imported). -- the trick is triggering the read of DAF_BUTLER_PLUGINS if some other test has already triggered it.

tests/test_commandLine.py
```diff
@@ -0,0 +1,105 @@
...
+            self.assertTrue("datastore" in result.stdout)
+            self.assertTrue("storageClasses" in result.stdout)
+
+    def testDumpConfig_file(self):
```
timj: Can you check that the subsetting command line option works as well please?

tests/test_commandLine.py
```diff
@@ -0,0 +1,105 @@
...
+            with open("there", "r") as f:
+                contents = " ".join(f.readlines())
+                self.assertTrue("composites" in contents)
+                self.assertTrue("datastore" in contents)
+                self.assertTrue("storageClasses" in contents)
+
+    def testValidateConfig(self):
+        """Test validating a valid config."""
+        runner = click.testing.CliRunner()
+        with runner.isolated_filesystem():
```
timj: Also check that an unknown dataset type causes a bad exit status.

timj (commented):
This looks good. I do have a few comments though.

python/lsst/daf/butler/cli/butler.py
n8pease: there's a couple things here:
1. I wanted to be able to log the loading of plugins, this is before the invocation of the top-level Click object.
2. As far as I can tell, logging does not like to be set up more than once. It seems there are some ways to force it but I couldn't get consistent results. (in Python 3.8 `logging.basicConfig` adds an arg `force` which may help with this, but I think we are on 3.7?)

The best solution I could come up with was to have the verbose option handled outside of Click, and have it turned on or off globally. Not every command will have a --verbose flag. The usage is 
```
Usage: butler [OPTIONS] COMMAND [ARGS]...

Options:
  -v, --verbose  Turn on debug reporting.
  --help         Show this message and exit.

Commands:
  create           Create an empty Gen3 Butler repository.
  dump-config      Dump either a subset or full Butler configuration to...
  validate-config  Validate the configuration files for a Gen3 Butler...
```

timj: I'm fine with every command having this. If we only want one special option though I would be happy to abandon `--verbose` and replace it with `--log-level` and have it default to WARN. This gives people a lot more control over DEBUG and INFO.

n8pease: ok, I'll change `--verbose` to `--log-level`

python/lsst/daf/butler/cli/cmd/__init__.py
n8pease: I was following a pattern of letting modules declare their `__all__` and importing `*` generally. Is `*` inefficient or not good or anything? Or you'd prefer more explicit?

python/lsst/daf/butler/cli/cmd/validate_config.py
n8pease: Click does not support comma separated values. It does allow multiple of an option to be passed (where it's set `multiple=True`), so multiple dataset types would be `-d foo -d bar -d baz`. Can we work with that?
I see that the help string needs too be corrected.

n8pease: I did find one hack to allow comma delimited values, but requires that there not be a space after the comma `-d foo,bar,baz` or that the values be in quotes `-d 'foo, bar, baz'`. Neither seems great to me.

timj: I don't understand. Click doesn't need to know about it. Why can't you call my processCommas function once you've got the value and are about to pass it to the routine that needs it?   You mean that click refuses to let you right `--datasettypes a,b,c` at all????

python/lsst/daf/butler/cli/cmd/__init__.py
timj: In this case you know that by definition you are getting one command from one import so the `*` seems like it risks pollution with no real gain.

python/lsst/daf/butler/cli/cmd/validate_config.py
n8pease: Hang on, was I assuming too much? I thought you would want to be able to process `--datasettypes a, b, c` with spaces after commas. But now I've looked closer at `processCommas` it looks like it only supports *no* spaces after commas IE `--datasettypes a,b,c`. If that's what's desired, yes we can support that.

python/lsst/daf/butler/cli/opt/repo.py
n8pease: it seems most decorators are lower case names. In any event, click is using all lower case. In use this looks like
```
@click.command()
@repo_option()
def someCommand(repo) ...
```
I think it looks better and more consistent than `@RepoOption`

python/lsst/daf/butler/cli/opt/repo.py
n8pease: from Jim's document I thought he wanted `REPO`. I agree that `--repo` is better for a CLI option. (I've since discovered that in documentation using all caps for *argument* names is normal, but does not mean that *option* names should be all caps)

python/lsst/daf/butler/cli/cmd/validate_config.py
timj: Yes, I was asking to keep the current scheme where you can say `-d a,b -d c,d` and end up with `[a,b,c,d]`.

TallJimbo: Could we rename to `--dataset-types`?  I feel like `-`-separated words for command-line arguments is a pretty established convention in the Linux world, at least.

python/lsst/daf/butler/cli/butler.py
```diff
@@ -0,0 +1,174 @@
...
+import logging
+import os
+
+import lsst.log
+from lsst.utils import doImport
+
+
+def _initLogging(logLevel):
+    log = lsst.log.Log.getLogger("lsst.daf.butler")
+    if "critical" == logLevel:
```
timj: I thought I made a comment on this but I can't find it any more. Just in case it got lost: please don't do this with `if` statements. The logging module has methods for translating the text string to a logging level and lsst.log has method for converting a logging level to an lsst.log level. See astro_metadata_translator command line tool for example code.

python/lsst/daf/butler/cli/butler.py
```diff
@@ -0,0 +1,183 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import click
+import logging
+import os
+
+from lsst.daf.butler.cli import cmd as butlerCommands
```
timj: Please use relative imports in all this code

timj (commented):
Don’t use absolute imports when you are wanting relative imports.

python/lsst/daf/butler/cli/butler.py
```diff
@@ -0,0 +1,182 @@
...
+
+def _initLogging(logLevel):
+    numeric_level = getattr(logging, logLevel, None)
+    if not isinstance(numeric_level, int):
+        raise click.ClickException(f"Invalid log level: {logLevel}")
+    logging.basicConfig(level=numeric_level)
+
+
+def funcNameToCmdName(functionName):
+    """Change underscores, used in fucntions, to dashes, used in commands."""
```
timj: Typo: fucntions

python/lsst/daf/butler/cli/butler.py
```diff
@@ -0,0 +1,182 @@
...
+
+        Returns
+        -------
+        An imported module or None
+            The imported module, or None if the module could not be imported.
+        """
+        try:
+            return doImport(pluginName)
+        except (TypeError, ModuleNotFoundError, ImportError) as err:
+            logging.warning("Could not import plugin from %s, skipping.", pluginName)
```
timj: Does this add the message to the right logger?  I think we want
```python
```
at the top and then `'log.warning()`. I know that this logger is only going to be triggered from the command line tooling but we should be consistent about log hierarchies (unless you are going to tell me that it's required to use the default logger).

python/lsst/daf/butler/cli/butler.py
```diff
@@ -0,0 +1,182 @@
...
+        -------
+        List of string
+            The names of the commands that can be called by the butler command.
+        """
+        commands = self._getLocalCommands()
+        for pluginName in self._getPluginList():
+            plugin = self._importPlugin(pluginName)
+            if plugin is None:
+                continue
+            for command in plugin.__all__:
```
timj: Maybe use:
```python
commands.extend([funcNameToCmdName(command) for command in plugin.__all__])
```
?

python/lsst/daf/butler/cli/butler.py
```diff
@@ -0,0 +1,182 @@
...
+                if command == name:
+                    fullCommand = pluginName + "." + command
+                    try:
+                        cmd = doImport(fullCommand)
+                    except (TypeError, ModuleNotFoundError, ImportError) as err:
+                        logging.debug("Command import exception: %s", err)
```
timj: logging -> log (I won't mention again)

python/lsst/daf/butler/cli/butler.py
```diff
@@ -0,0 +1,182 @@
...
+            that can be called.
+
+        Returns
+        -------
+        An imported module or None
+            The imported module, or None if the module could not be imported.
+        """
+        try:
+            return doImport(pluginName)
+        except (TypeError, ModuleNotFoundError, ImportError) as err:
```
timj, line 76: Do we gain much by trying to restrict the exception handler to these three rather than always reporting the problem for all Exception?

python/lsst/daf/butler/cli/butler.py
```diff
@@ -0,0 +1,182 @@
...
+        Returns
+        -------
+        click.Command
+            A Command that wraps a callable command function.
+        """
+        name = cmdNameToFuncName(name)
+        localCmd = self._getLocalCommand(name)
+        if localCmd is not None:
+            return localCmd
+        for pluginName in self._getPluginList():
```
timj, line 155: Would this be simplified a lot of if you had a dict mapping command name to thing to import? You'd presumably then be able to catch people defining two identical subcommand names in different packages.  At the moment it seems that if two identical subcommands were defined you'd get the one that was ahead in the environment variable path.

python/lsst/daf/butler/cli/cmd/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+__all__ = ["create", "dump_config", "validate_config"]
```
timj: I think @TallJimbo and I agreed that we wanted these names reversed to be config dump and config validate.

python/lsst/daf/butler/cli/cmd/validate_config.py
```diff
@@ -0,0 +1,43 @@
...
+from ... import Butler, ValidationError
+from ..opt import dataset_type_option, repo_option
+
+
+@click.command(name="validate-config")
+@click.pass_context
+@repo_option(required=True)
+@click.option("--quiet", "-q", is_flag=True, help="Do not report individual failures.")
+@dataset_type_option(help="Specific DatasetType(s) to validate.")
+@click.option("--ignore", "-i", type=str, multiple=True,
```
timj: where does the comma splitting happen? Doesn't it need the same callback that dataset types get?

python/lsst/daf/butler/cli/util/__init__.py
```diff
@@ -1,10 +1,8 @@
-#!/usr/bin/env python
-
-# This file is part of daf_butler.
+# This file is part of obs_base.
```
timj: Not obs_base

python/lsst/daf/butler/cli/util/split_commas.py
```diff
@@ -0,0 +1,49 @@
+# This file is part of daf_butler.
```
timj: For utility functions we don't require a separate file per function. The reason I wanted separate files per sub command was that I was worried we were going to get many sub commands and things would get out of control. We envisaged that there would be support code needed for the subcommands and that would get confusing. You seem to have mitigated a lot of that worry with clever callbacks in click definitions but we'll see what happens when more complex commands turn up.

timj (approved):
Minor comments. Some really impressive test code.

python/lsst/daf/butler/cli/cmd/validate_config.py
n8pease: callback splitting is in definition of the dataset-type option. If we need some dataset-type options to take multiple and some not that can be added tot he option. https://github.com/lsst/daf_butler/pull/258/files#diff-191150039ddacf9a20d7744bc23e627aR37

timj: What I'm trying to say is that this `--ignore` option previously processed commas but I don't see it doing that in this implementation.

n8pease: Oh, I was looking at the wrong line. I'll add it to `ignore`

python/lsst/daf/butler/cli/butler.py
n8pease: nope, I'm just confused about logging. Will fix.

python/lsst/daf/butler/cli/butler.py
n8pease: no, probably not. will change to all Exception.

python/lsst/daf/butler/cli/butler.py
n8pease: good catch.

python/lsst/daf/butler/cli/util/split_commas.py
n8pease: I'll move utils into their own file. I think it's worth keeping commands and options in separate files at least for now.

PR #259, opened 2020-04-16
DM-23671: Change datastore remove to a trash + emptyTrash

timj:
Splits the remove step into two phases. The first moves a row from the dataset_location to the dataset_location_trash table. The second step deletes all the datasets that have been trashed. This will hopefully reduce further the chances of datasets being deleted and leaving the datastore in an inconsistent state.

I need to clean up some of the commits but tests all pass so it's ready for a look.

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1061,10 +1061,6 @@ class Butler:
                     self.registry.disassociate(tag, refs, recursive=False)
             if unstore:
                 try:
-                    # Point of no return: if Datastore.remove ever succeeds
-                    # or fails in a way that nevertheless changes repository
-                    # state, and then we get an exception (either in Datastore
-                    # or Registry) the repo might be corrupted.
                     for ref in refs:
                         # There is a difference between a concrete composite
                         # and virtual composite. In a virtual composite the
@@ -1078,12 +1074,15 @@ class Butler:
                         # to ignore already-removed-from-datastore datasets
                         # anyway.
                         if self.datastore.exists(ref):
-                            self.datastore.remove(ref)
+                            self.datastore.trash(ref)
                         if purge:
                             self.registry.removeDataset(ref)
+
+                    # Point of no return for removing artifacts
+                    self.datastore.emptyTrash()
                 except BaseException as err:
                     raise IOError("WARNING: an incomplete deletion may have put "
-                                  "the repository in a corrupted state.") from err
+                                  f"the repository in a corrupted state: {err}") from err
 
     @transactional
     def ingest(self, *datasets: FileDataset, transfer: Optional[str] = None, run: Optional[str] = None,
--- a/python/lsst/daf/butler/core/datasets/ref.py
+++ b/python/lsst/daf/butler/core/datasets/ref.py
@@ -20,7 +20,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 from __future__ import annotations
 
-__all__ = ["DatasetRef"]
+__all__ = ["DatasetRef", "FakeDatasetRef"]
 
 import hashlib
 from typing import Any, Dict, Iterable, Iterator, Mapping, Optional, Tuple
@@ -362,3 +362,44 @@ class DatasetRef:
     `unresolved` to add or remove this information when creating a new
     `DatasetRef`.
     """
+
+
+@immutable
+class FakeDatasetRef:
+    """A fake `DatasetRef` that can be used internally by butler where
+    only the dataset ID is available.
+
+    Should only be used when registry can not be used to create a full
+    `DatasetRef` from the ID.  A particular use case is during dataset
+    deletion when solely the ID is available.
+
+    Parameters
+    ----------
+    id : `int`
+        The dataset ID.
+    """
+    __slots__ = ("id",)
+
+    def __new__(cls, id: int):
+        self = super().__new__(cls)
+        self.id = id
+        return self
+
+    def __str__(self):
+        return f"dataset_id={self.id}"
+
+    def __repr__(self):
+        return f"FakeDatasetRef({self.id})"
+
+    def __eq__(self, other: FakeDatasetRef):
+        try:
+            return self.id == other.id
+        except AttributeError:
+            return NotImplemented
+
+    def __hash__(self) -> int:
+        return hash(self.id)
+
+    @property
+    def components(self):
+        return {}
--- a/python/lsst/daf/butler/core/datasets/ref.py
+++ b/python/lsst/daf/butler/core/datasets/ref.py
@@ -403,3 +403,7 @@ class FakeDatasetRef:
     @property
     def components(self):
         return {}
+
+    @staticmethod
+    def flatten(refs: Iterable[FakeDatasetRef], *, parents: bool = True) -> Iterator[DatasetRef]:
+        return DatasetRef.flatten(refs, parents=parents)
--- a/python/lsst/daf/butler/datastores/fileLikeDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileLikeDatastore.py
@@ -181,8 +181,6 @@ class FileLikeDatastore(GenericBaseDatastore):
                 ddl.FieldSpec(name="file_size", dtype=Integer, nullable=True),
             ]),
             unique=frozenset(),
-            foreignKeys=[ddl.ForeignKeySpec(table="dataset", source=("dataset_id",), target=("dataset_id",),
-                                            onDelete="CASCADE")]
         )
 
     def __init__(self, config, registry, butlerRoot=None):
@@ -233,6 +231,34 @@ class FileLikeDatastore(GenericBaseDatastore):
 
     def __str__(self):
         return self.root
+
+    @abstractmethod
+    def _artifact_exists(self, location):
+        """Check that an artifact exists in this datastore at the specified
+        location.
+
+        Parameters
+        ----------
+        location : `Location`
+            Expected location of the artifact associated with this datastore.
+
+        Returns
+        -------
+        exists : `bool`
+            True if the location can be found, false otherwise.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def _delete_artifact(self, location):
+        """Delete the artifact from the datastore.
+
+        Parameters
+        ----------
+        location : `Location`
+            Location of the artifact associated with this datastore.
+        """
+        raise NotImplementedError()
 
     def addStoredItemInfo(self, refs, infos):
         # Docstring inherited from GenericBaseDatastore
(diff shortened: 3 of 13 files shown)


python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -362,3 +362,44 @@ def flatten(refs: Iterable[DatasetRef], *, parents: bool = True) -> Iterator[Dat
...
+        return f"FakeDatasetRef({self.id})"
+
+    def __eq__(self, other: FakeDatasetRef):
+        try:
+            return self.id == other.id
+        except AttributeError:
+            return NotImplemented
+
+    def __hash__(self) -> int:
+        return hash((self.id,))
```
TallJimbo: What's the reasoning for the tuple inside the call?

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -1032,6 +1033,72 @@ def insertDatasetLocations(self, datastoreName: str, refs: Iterable[DatasetRef])
...
+            All the `DatasetRef` that are listed.
+        """
+
+        table = self._tables.dataset_location
+        result = self._db.query(
+            sqlalchemy.sql.select(
+                [table.columns.datastore_name, table.columns.dataset_id]
+            ).where(
+                sqlalchemy.sql.and_(table.columns.dataset_id.in_([ref.id for ref in refs]),
+                                    table.columns.datastore_name == datastoreName)
```
TallJimbo, line 1095: I don't know if this query scales when the number of refs is very large.  I don't know that it doesn't, but it might be good to get a database expert to weigh in (@andy-slac, maybe?).  My first thought would be that we might want to chunk these up, but I have no idea what the chunk size ought to be.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -1032,6 +1033,72 @@ def insertDatasetLocations(self, datastoreName: str, refs: Iterable[DatasetRef])
...
+        Parameters
+        ----------
+        datastoreName : `str`
+            Name of the datastore holding these datasets.
+        refs : `~collections.abc.Iterable` of `DatasetRef`
+            References to the datasets.
+        """
+        # We only want to move rows that already exist in the main table
+        filtered = self.checkDatasetLocations(datastoreName, refs)
+        self.canDeleteDatasetLocations(datastoreName, filtered)
```
TallJimbo, line 1049: Doing the query that filters down to the datasets that exist in the database separately from the insert into the trash table _might_ be unsafe in the presence of concurrent access - I think the SELECT query could be out of date by the time the INSERT query runs, even if they're in a transaction.  Unfortunately, fixing that would involve add INSERT...SELECT support to `Database` or punching a bit of a hole in its abstraction layer to let this code do it directly.  I'm going to have to move this code around a bit anyway on DM-21764, so I could have a try at fixing it there when I do.  That would need to involve merging `checkDatasetLocations` and `canDeleteDatasetLocations`.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -1065,7 +1132,59 @@ def getDatasetLocations(self, ref: DatasetRef) -> Set[str]:
...
+            sqlalchemy.sql.select(
+                [table.columns.dataset_id]
+            ).where(
+                table.columns.datastore_name == datastoreName
+            )
+        ).fetchall()
+        return {FakeDatasetRef(r["dataset_id"]) for r in result}
+
+    @transactional
+    def emptyDatasetLocationsTrash(self, datastoreName, refs):
```
TallJimbo: Missing type annotations here (I'm assuming you prefer to have them).  Same in the next method.

python/lsst/daf/butler/_butler.py
```diff
@@ -1078,12 +1074,15 @@ def prune(self, refs: Iterable[DatasetRef], *,
                         # to ignore already-removed-from-datastore datasets
                         # anyway.
                         if self.datastore.exists(ref):
-                            self.datastore.remove(ref)
+                            self.datastore.trash(ref)
```
TallJimbo, line 1077: I think we could now simplify a lot of this, removing the giant TODO comment block, and all the checks under the `if purge` blocks (which will be handled more rigorously by transaction rollback now).
 - Start a registry transaction context (I don't think the Datastore ones are useful here).
 - Expand datasets to components recursively.
 - Do the `if unstore` loop that calls `self.datastore.trash`.
 - Do the disassociate and and possibly the purge Registry operations.
 - End the transaction context, committing those changes (or raising if something went wrong).
 - Call `emptyTrash`.

I'd love to have someone else check my logic on that, of course, especially on whether it's safe if there are concurrent deletion operations and how it interacts with various kinds of composites in datastore (for virtual composites, are there ever records in `dataset_location`)?

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -595,6 +639,69 @@ def getUri(self, ref, predict=False):
...
+        artifactsToRemove = set()
+
+        for ref in trashed:
+            location, _ = self._get_dataset_location_info(ref)
+
+            if location is None:
+                raise FileNotFoundError(f"Requested dataset ({ref}) does not exist")
+
+            if not self._artifact_exists(location):
+                raise FileNotFoundError(f"No such file: {location.uri}")
```
TallJimbo: I suspect making this a silent no-op will make concurrent calls to `emptyTrash` safer, and maybe totally safe.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -595,6 +639,69 @@ def getUri(self, ref, predict=False):
 
         return location.uri
 
+    @transactional
+    def trash(self, ref):
+        """Indicate to the Datastore that a Dataset can be removed.
```
TallJimbo: FWIW, I have mostly been replacing "Dataset" with "dataset" when I see it in docs, since it's not actually a class.  Happy to stop doing that, but you and I should probably be consistent.

python/lsst/daf/butler/datastores/genericDatastore.py
```diff
@@ -108,19 +108,37 @@ def _register_datasets(self, refsAndInfos):
...
+        """Remove the dataset entries that have been trashed
+
+        Parameters
+        ----------
+        refs : iterable of `DatasetRef`
+           Datasets to be removed. These datasets must be present in the trash.
+        """
+        for ref in refs:
+            self.removeStoredItemInfo(ref)
+        self.registry.emptyDatasetLocationsTrash(self.name, refs)
```
TallJimbo: Is `_remove_trashed_registry_datasets` used anywhere?

python/lsst/daf/butler/datastores/inMemoryDatastore.py
```diff
@@ -354,28 +375,47 @@ def remove(self, ref):
         ------
         FileNotFoundError
             Attempt to remove a dataset that does not exist.
+        """
+
+        log.debug("Trash %s in datastore %s", ref, self.name)
+
+        # Check that this dataset is known to datastore
+        self._get_dataset_info(ref)
```
TallJimbo: Is this check necessary?  It'd probably be safer to make the deletion do nothing if the `ref` isn't actually in the datastore, to guard against the case where the a concurrent delete happens between the check and the move to trash.

python/lsst/daf/butler/datastores/inMemoryDatastore.py
```diff
@@ -354,28 +375,47 @@ def remove(self, ref):
...
-        remainingRefs = allRefs - theseRefs
-        if not remainingRefs:
-            del self.datasets[thisID]
+        self.registry.emptyDatasetLocationsTrash(self.name, trashed)
 
-        # Remove rows from registries
-        self._remove_from_registry(ref)
+        for refId in artifactsToRemove:
+            log.debug("Removing artifact %s from datastore %s", realID, ref)
+            del self.datasets[refId]
```
TallJimbo: I think it might be safer to do this operation (unlike the others) one dataset at a time, by first deleting the artifact, then deleting the internal records, and only then removing from the trash table.  That means a Datastore would _have_ to be willing to silently ignore artifact deletion if it fails because the artifact is already gone (i.e. because a previous deletion operation only got that far before it died).  But I think that's reasonable, and the failure mode of the current logic is more dangerous: a file could be left around even though it'd already been completely removed from all records (even the trash).

TallJimbo (approved):
This is a huge improvement. I've left a number of comments about possible further improvements, mostly concurrency edge cases. And most of those are about trying avoid query-then-operate logic, because I've learned that leaves one open to something unexpected happening in between.

But I'm also not sure all of those separate comments form a coherent big picture of how to make this rock-solid, at least not yet, especially when different varieties of composites are in play.  And I'm not sure we need rock-solid yet - this is already enough to get us from "easy to accidentally corrupt your repo" to "very difficult to accidentally corrupt your repo", so feel free to defer any of my comments to a future ticket.

python/lsst/daf/butler/core/datasets/ref.py
timj: Mainly the copying and pasting from the lines above and removing the items we no longer had...

python/lsst/daf/butler/registry/_registry.py
timj: At the moment this is only being called with a single ref and associated components so I don't think it's going to be unbounded. Of course in the future you might want to be doing bulk trashing rather than `for each ref: trash`.

python/lsst/daf/butler/registry/_registry.py
timj: I'm fine with you taking another stab at this. The reason I had to do it was that in some cases a DatasetRef can refer to components that have already been trashed (since we can't turn the component off within the DatasetRef) so switching things to only moving rows that are still there prevented me from adding rows to the trash that were already listed in the trash.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
timj: Ok. Concurrent access would suggest that we should also skip the earlier test since it implies that the trashed file has also been removed from the internal registry as well as having the file removed.

python/lsst/daf/butler/datastores/genericDatastore.py
timj: No. It was an early implementation. Well spotted.

python/lsst/daf/butler/datastores/inMemoryDatastore.py
timj: One option is to do a rename in the main trash emptying loop and then a final delete afterwards. You are right though that one reason why the file is no longer there is that some other process is emptying the trash at the same time and got to that file first (and the rename would also fail in that case as well). Multiple people emptying the trash simultaneously implies that we mostly have to suck up any errors and cross fingers. This is a problem if the file is left hanging around.

The trash emptying currently tries to delete everything. Another option is for emptyTrash to have a "delete everything mode" as now but also have a "delete only these refs and children" option (ie the refs that you only shortly before told to be trashed).

Concurrent accessing does raise the possibility that one process is emptying trash whilst another is putting the dataset that has been deleted. This will fail at the moment because the internal registry will still mention the file and the file will be on disk and since the emptyTrash is in a transaction I think that will be fine.

python/lsst/daf/butler/datastores/inMemoryDatastore.py
timj: In this particular case a concurrent delete should be impossible because in memory datastore is constrained to a particular python process. For the other datastores, I don't really know how to respond to multiple processes all trying to delete the same dataset. It sort of means that you have to say that `trash()` is never allowed to fail because you can never really tell if the file should be there or not.

python/lsst/daf/butler/datastores/inMemoryDatastore.py
TallJimbo: As we discussed a bit on slack, I think we're pretty much forced into the "ignore errors" behavior to deal with other aspects of concurrency.

For the last point on `put` collisions with deletions, I think that's a sufficiently rare case that it's fine if the `put` just fails, because it'll do so without doing any damage to the repository (as we've already done the work to make sure `put` is atomic).

PR #262, opened 2020-04-23
DM-24347: Allow a component dataset to be None

timj:
Sometimes when we store a dataset, not all the components are defined. Registry won't register them in that case but there are some code paths that will still cause a datastore to try to return a None component.


TallJimbo:
Code looks good; my only quibble is with the message on the final commit: `Registry` _does_ contain components that are `None`, probably more often than not - we just register all components in the StorageClass every time we put or ingest, regardless of whether they're present on the in-memory dataset or file.

I don't think that invalidates anything you've done here, but I think the message is wrong (or maybe just confusing, if I misunderstood it).

TallJimbo (approved):

timj:
Yes, you are right when you don't disassemble the in-memory dataset. If you disassemble then we do not register the `None` components. This means that when I added some tests for None to test_butler.py they were all blowing up in the cases where we do disassemble.

PR #263, opened 2020-04-23
DM-24543: handle conflicting CLI subcommand names

n8pease:
Refactor the plugin loader so that commands with the same name in
different plugins are found and an error rased.

### DIFF
--- a/python/lsst/daf/butler/cli/butler.py
+++ b/python/lsst/daf/butler/cli/butler.py
@@ -20,12 +20,18 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import click
+from collections import defaultdict
 import logging
 import os
+import yaml
 
 from . import cmd as butlerCommands
 from .utils import to_upper
 from lsst.utils import doImport
+
+# localCmdPkg identifies commands that are in this package, in the dict of
+# commands used in this file.
+localCmdPkg = "localPackage"
 
 log = logging.getLogger(__name__)
 
@@ -49,12 +55,23 @@ def cmdNameToFuncName(commandName):
 
 class LoaderCLI(click.MultiCommand):
 
+    def __init__(self, *args, **kwargs):
+        self.commands = None
+        super().__init__(*args, **kwargs)
+
     @staticmethod
     def _getPluginList():
+        """Get the list of importable yaml files that contain butler cli data.
+
+        Returns
+        -------
+        `list` [`str`]
+            The list of files that contain yaml data about a cli plugin.
+        """
         pluginModules = os.environ.get("DAF_BUTLER_PLUGINS")
-        if not pluginModules:
-            return []
-        return pluginModules.split(":")
+        if pluginModules:
+            return pluginModules.split(":")
+        return []
 
     @staticmethod
     def _importPlugin(pluginName):
@@ -79,25 +96,26 @@ class LoaderCLI(click.MultiCommand):
             return None
 
     @staticmethod
-    def _getLocalCommand(commandName):
-        """Search the commands provided by this package and return the command
-        if found.
+    def _mergeCommandLists(a, b):
+        """Combine two dicts whose keys are strings (command name) and values
+        are list of string (the package(s) that provide the named command).
 
         Parameters
         ----------
-        commandName : string
-            The name of the command to search for.
+        a : `defaultdict` [`str`: `list` [`str`]]
+            The key is the command name. The value is a list of package(s) that
+            contains the command.
+        b : (same as a)
 
         Returns
         -------
-        Command object
-            A Click command that can be executed.
+        commands : `defaultdict` [`str`: [`str`]]
+            For convenience, returns a extended with b. ('a' is modified in
+            place.)
         """
-        try:
-            return getattr(butlerCommands, commandName)
-        except AttributeError:
-            pass
-        return None
+        for key, val in b.items():
+            a[key].extend(val)
+        return a
 
     @staticmethod
     def _getLocalCommands():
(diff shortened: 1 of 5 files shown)


python/lsst/daf/butler/cli/butler.py
```diff
@@ -79,36 +90,80 @@ def _importPlugin(pluginName):
...
+            returns a extended with b
         """
-        try:
-            return getattr(butlerCommands, commandName)
-        except AttributeError:
-            pass
-        return None
+        for key, val in b.items():
+            try:
+                a[key].extend(val)
```
timj: I'm not a huge fan of using exceptions for expected behavior. Since this is internal API I think it might be cleaner to set up `commands` below as a `defaultdict`. Then you won't need the try at all. (since you test it as if it is public API that might lead to too much work having to ensure that the first dict is a defaultdict with the right default).

python/lsst/daf/butler/cli/butler.py
```diff
@@ -79,36 +90,80 @@ def _importPlugin(pluginName):
...
+        a : dict of string to list of string
+            key is command name, value is packages that provide the command
+        b : (same as a)
 
         Returns
         -------
-        Command object
-            A Click command that can be executed.
+        dict of string to list of string
+            returns a extended with b
```
timj: Please use capital letters at start of sentences. Also probably should note that you are returning this for convenience since `a` has been modified.

python/lsst/daf/butler/cli/butler.py
```diff
@@ -49,8 +49,19 @@ def cmdNameToFuncName(commandName):
...
+        self.commands = None
+        super().__init__(*args, **kwargs)
+
     @staticmethod
     def _getPluginList():
+        """Get the list of importable packages that contain butler commands.
+
+        Returns
+        -------
+        list of string
```
timj: Returns are supposed to have a variable name and follow the same format as parameters. Also there are multiple ways of writing this type but
```
`list` [`str`]
```
would be preferable to "list of string".  See https://developer.lsst.io/python/numpydoc.html?highlight=numpydoc#sequence-types

python/lsst/daf/butler/cli/butler.py
```diff
@@ -148,24 +199,14 @@ def get_command(self, context, name):
...
-                        log.debug("Command import exception: %s", err)
-                        context.fail("Could not import command {fullCommand}")
-                    return cmd
-        return None
+        if self.commands is None:
+            self.commands = self._getCommands()
+        if name not in self.commands:
+            return None
+        self._raiseIfDuplicateCommands(self.commands)
+        if self.commands[name][0] == "lsst.daf.butler.cli.cmd":
```
timj: Please set this as a constant somewhere since you have used it twice (you could even consider deriving it from `__name__` rather than hardcoding it).

tests/test_cliPluginLoader.py
```diff
@@ -59,19 +65,96 @@ def cli():
...
+        """Test the get_command function for commands in daf_butler."""
+        self.assertEqual(butler.LoaderCLI().get_command(None, "create"), cmd.create)
+
+    def test_getCommand_plugin(self):
+        """Test the get_command function for commands loaded as plugins.
+
+        Typically plugin commands are not in daf_butler, but here we mask the
+        local commands in the TestCLI class and expose them via the environment
+        variable."""
+        with patch.dict("os.environ", {"DAF_BUTLER_PLUGINS": "lsst.daf.butler.cli.cmd"}):
```
timj: Did not know about this method. I wrote `modified_environment` context manager in test_config.py that does exactly that.  I guess I need to change test_config to use unittest.mock.

tests/test_cliPluginLoader.py
```diff
@@ -59,19 +65,96 @@ def cli():
...
         runner = click.testing.CliRunner()
         result = runner.invoke(cli, "--help")
         self.assertEqual(result.exit_code, 0, result.stdout)
-        self.assertIn("Commands:\n  command-test", result.stdout)
+        self.assertIn("command-test", result.stdout)
+
+    def test_getLocalCommands(self):
+        """Test getting the daf_butler CLI commands."""
+        localCommands = butler.LoaderCLI._getLocalCommands()
+        self.assertEqual(localCommands, {"config-dump": ["lsst.daf.butler.cli.cmd"],
```
timj: Every time you add a new daf_butler subcommand I think you are going to have to edit multiple tests here. It might help you to define the command list once and then create this dict from that.

python/lsst/daf/butler/cli/butler.py
```diff
@@ -79,36 +90,80 @@ def _importPlugin(pluginName):
...
+
+        Returns
+        -------
+        dict of command name to list of packages name.
+            Key is the command name, value is the package(s) that contains the
+            command.
+        """
+        commands = {}
+        for pluginName in cls._getPluginList():
+            plugin = cls._importPlugin(pluginName)
```
timj: We have to keep an eye out for import overhead here. There is a worry that `butler create` could end up importing multiple obs and pipe_ packages in their entirety, which would be slow.

timj (approved):
Looks good. Some comments to ponder.

python/lsst/daf/butler/cli/butler.py
```diff
@@ -20,13 +20,17 @@
...
+from collections import defaultdict
 import logging
 import os
+import yaml
 
 from . import cmd as butlerCommands
 from .utils import to_upper
 from lsst.utils import doImport
 
+localCmdPkg = "lsst.daf.butler.cli.cmd"
```
timj: This looking like an import string does lead me to the mistaken impression that you might doImport this later on but the reality is it's a label used to tag the local commands and not something actionable by the code. Maybe add a comment that this is a label describing the local commands.

python/lsst/daf/butler/cli/butler.py
```diff
@@ -79,36 +94,101 @@ def _importPlugin(pluginName):
...
+        """
+        commands = defaultdict(list)
+        for pluginName in cls._getPluginList():
+            try:
+                with open(pluginName, "r") as resourceFile:
+                    resources = defaultdict(list, yaml.safe_load(resourceFile))
+            except Exception as err:
+                log.warning(f"Error loading commands from {pluginName}, skipping. {err}")
+                continue
+            if 'cmd' not in resources:
```
timj: double quotes please

timj (approved):
Looks good.

PR #267, opened 2020-04-29
DM-24658: Improve error reporting in connection string.

DinoBektesevic:
Stop silencing `DbAuthError` when cause is a YAML parsing issue. 
Reorganize dbAuth and connectionString test files, simplify and de-duplicate. 
Add test to ensure error propagates through ConnectionString class .


python/lsst/daf/butler/registry/_dbAuth.py
```diff
@@ -79,6 +83,8 @@ def __init__(self, path=None, envVar=None, authList=None):
                 "{:o}".format(secretPath, mode))
 
         try:
+            if not os.path.isfile(secretPath):
+                raise DbAuthNotFoundError(f"DbAuth file not found: {secretPath}")
```
ktlim: This should be removed and the proper exception thrown at (new) line 78.

python/lsst/daf/butler/registry/connectionString.py
```diff
@@ -91,13 +93,15 @@ def fromConfig(cls, registryConfig):
...
-            # re-raise permission error for safety
-            raise
-        except DbAuthError:
-            # credentials file doesn't exist or no match was found
+        except DbAuthNotFoundError:
+            # credentials file doesn't exist or no matches were found
             pass
+        except Exception:
+            # re-raise everything else
+            raise
```
ktlim: Is this necessary?  Leaving it out should cause the exception to propagate normally.

tests/config/dbAuth/registryConf2.yaml
```diff
@@ -0,0 +1,7 @@
+# minimal configuration required when using explicit 'host-port',
+# also tests '/' are placed correctly when not epxlicit in dbname
```
ktlim:
```suggestion
# also tests '/' are placed correctly when not explicit in dbname
```

ktlim (approved):

python/lsst/daf/butler/registry/_dbAuth.py
DinoBektesevic: Thanks! I guess I was blind.

PR #271, opened 2020-05-04
DM-24594: Implement integer synthetic PK for collections

andy-slac:
Existing collection manager class was refactored, most of the logic
moved into a base class which is now used to implement two separate
managers. New manager uses integer auto-increment column for collections
table as its PK, all dependent tables switched to integer FK too.
Unit tests were updated to test both implementations. Registry
configuration is still using current string-based manager.

### DIFF
--- a/python/lsst/daf/butler/registry/collections/_base.py
+++ b/python/lsst/daf/butler/registry/collections/_base.py
@@ -1,0 +1,438 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+__all__ = []
+
+from abc import abstractmethod
+import astropy.time
+import itertools
+from typing import (
+    Any,
+    Iterator,
+    NamedTuple,
+    Optional,
+    TYPE_CHECKING,
+)
+
+import sqlalchemy
+
+from ...core import ddl
+from ...core.timespan import Timespan, TIMESPAN_FIELD_SPECS
+from .._collectionType import CollectionType
+from ..interfaces import (
+    ChainedCollectionRecord,
+    CollectionManager,
+    CollectionRecord,
+    MissingCollectionError,
+    RunRecord,
+)
+from ..wildcards import CollectionSearch
+
+if TYPE_CHECKING:
+    from .database import Database
+
+
+def _makeCollectionForeignKey(sourceColumnName: str, collectionIdName: str, **kwargs) -> ddl.ForeignKeySpec:
+    """Define foreign key specification that refers to collections table.
+
+    Parameters
+    ----------
+    sourceColumnName : `str`
+        Name of the column in the referring table.
+    collectionIdName : `str`
+        Name of the column in collections table that identifies it (PK).
+    **kwargs
+        Additional keyword arguments passed directly to `ddl.ForeignKeySpec`.
+
+    Returns
+    -------
+    spec : `ddl.ForeignKeySpec`
+        Foreign key specification.
+
+    Notes
+    -----
+    This method assumes fixed name ("collection") of a collections table.
+    There is also a general assumption that collection primary key consists
+    of a single column.
+    """
+    return ddl.ForeignKeySpec("collection", source=(sourceColumnName,), target=(collectionIdName,),
+                              **kwargs)
+
+
+def makeRunTableSpec(collectionIdName: str, collectionIdType: type):
+    """Define specification for "run" table.
+
+    Parameters
+    ----------
+    collectionIdName : `str`
+        Name of the column in collections table that identifies it (PK).
+    collectionIdType
+        Type of the PK column in the collections table, one of the
+        `sqlalchemy` types.
+
+    Returns
+    -------
+    spec : `ddl.TableSpec`
+        Specification for run table.
+
+    Notes
+    -----
+    Assumption here and in the code below is that the name of the identifying
+    column is the same in both collections and run tables. The names of
+    non-identifying columns containing run metadata are fixed.
+    """
+    return ddl.TableSpec(
+        fields=[
+            ddl.FieldSpec(collectionIdName, dtype=collectionIdType, primaryKey=True),
+            TIMESPAN_FIELD_SPECS.begin,
+            TIMESPAN_FIELD_SPECS.end,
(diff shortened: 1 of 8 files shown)


python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -0,0 +1,485 @@
...
+    ----------
+    collectionIdName : `str`
+        Name of the column in collections table that identifies it (PK).
+    collectionIdType
+        Type of the PK column in the collections table, one of the
+        `sqlalchemy` types.
+
+    Returns
+    -------
+    spec : `ddl.TableSpec`
```
TallJimbo, line 94: Is it actually allowed to not have a description for the return value?  I often find myself writing something that's totally trivial/redundant/obvious, so if this is intentional rather than a mistake I'll start using it, too!

python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -0,0 +1,485 @@
...
+            TIMESPAN_FIELD_SPECS.end,
+            ddl.FieldSpec("host", dtype=sqlalchemy.String, length=128),
+        ],
+        foreignKeys=[
+            _makeCollForeignKey(collectionIdName, collectionIdName, onDelete="CASCADE"),
+        ],
+    )
+
+
+def makeCollChainTableSpec(collectionIdName: str, collectionIdType: type):
```
TallJimbo: I'd prefer to just expand "Coll" to "Collection" here and elsewhere - it's not really an expected abbreviation for someone coming new to the code, and the existence of both in different places makes it hard to remember which one is used in the name of some symbol of interest without looking it up.

python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -0,0 +1,485 @@
...
+        # Docstring inherited from RunRecord.
+        return self._timespan
+
+
+class DefaultChainedCollectionRecord(ChainedCollectionRecord):
+    """Default `ChainedCollectionRecord` implementation.
+
+    This method assumes the same chain table definition as produced by
+    `makeCollChainTableSpec` method. All column names in the table are
+    fixed and ahrd-coded in the methods.
```
TallJimbo: typo "ahrd"

python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -0,0 +1,485 @@
...
+                 collectionIdName: str):
+        self._db = db
+        self._tables = tables
+        self._collectionIdName = collectionIdName
+        self._records = {}  # indexed by record ID
+
+    @classmethod
+    def getCollectionForeignKeyName(cls, prefix: str = "collection") -> str:
+        # Docstring inherited from CollectionManager.
+        return f"{prefix}_id"
```
TallJimbo: I'd prefer to leave this method and the one below it abstract, I think, and let subclasses implement the suffix to be `_id` or `_name` to match the primary key.  We have a bit of an informal convention (or at least a lot precedent) for `id` fields being integers only.

python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -0,0 +1,485 @@
...
+        self._id = id
+
+    @property
+    def key(self) -> int:
+        # Docstring inherited from CollectionRecord.
+        return self._id
+
+    _id: Any
+    """Database-internal unique collection ID.
+    """
```
TallJimbo: This may be an indication we should just make `key` a regular attribute (and `__init__` argument) to `CollectionRecord` itself, which would then be concrete and directly usable.

python/lsst/daf/butler/registry/collections/synthIntKey.py
```diff
@@ -0,0 +1,133 @@
...
+            del self._nameCache[record.name]
+
+    def _getByName(self, name: str):
+        # Docstring inherited from DefaultCollectionManager.
+        if self._nameCache:
+            return self._nameCache.get(name)
+
+    _nameCache: Optional[Mapping[str, CollectionRecord]] = None
+    """Mapping that indexes records by name for fast search.
+    """
```
TallJimbo: Elsewhere in daf_butler I think we've only been documenting/annotating public instance attributes like this.  I think private ones probably shouldn't have docs, and I learned today that putting the type annotations in `__init__` is sufficient for mypy.  This may be something we should discuss more broadly before assuming precedent -> policy, but there is precedent.

python/lsst/daf/butler/registry/collections/nameKey.py
```diff
@@ -58,167 +47,25 @@
...
-            restriction = row[self._table.columns.dataset_type_name]
-            if not restriction:
-                restriction = ...  # we store ... as "" in the database
-            record = manager[key]
-            children.append((record.name, restriction))
-        return CollectionSearch.fromExpression(children)
-
-
-class AggressiveNameKeyCollectionManager(CollectionManager):
+class AggressiveNameKeyCollectionManager(DefaultCollectionManager):
```
TallJimbo: It's probably best to either drop "Aggressive" in the name here, or add it to both `DefaultCollectionManager` and `SynthIntCollectionManager`, since it's a property they all have in common (and we're not yet at the point where that change would break existing repos).

python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -0,0 +1,485 @@
...
+
+    Returns
+    -------
+    spec : `ddl.ForeignKeySpec`
+
+    Note
+    ----
+    This method assumes fixed name ("collection") of a collections table.
+    There is also a general assumption that collection primary key consists
+    of a single column.
```
TallJimbo, line 75: "Note" -> "Notes" (and in many other places).

But I really appreciate the clear descriptions of assumptions you're putting in these docs.

TallJimbo (approved):
Looks great!  I've left a number of comments, all of them minor at least in importance (and mostly in implied changes, I think).

python/lsst/daf/butler/registry/collections/_base.py
andy-slac: Hmm, good question. I think I saw examples of missing description for parameters when it is obvious, and I assumed return value is not that different from parameters. I guess I better ask at #dm-docs

andy-slac: Apparently what is obvious to me is not obvious to others :shocking: 🙂 
Will add something here.

python/lsst/daf/butler/registry/collections/_base.py
andy-slac: Agree, will do.

python/lsst/daf/butler/registry/collections/synthIntKey.py
andy-slac: I really wanted to have this member defined in `__init__` but that is a bit more code and documentation, this is why I used this shortcut of defining a private attribute as a class member and felt that as a class member it should be documented. Proabably I just need to add `__init__` and not expose this at class level.

python/lsst/daf/butler/registry/collections/nameKey.py
andy-slac: Agree. I was thinking about implementing different pre-fetching/caching policies in the DefaultCollectionManager, then we can instantiate various managers by either providing different subclasses which just pass correct parameters to a DefaultCollectionManager constructor or maybe adding configuration options for managers in YAML.

andy-slac: I renamed AggressiveNameKeyCollectionManager to NameKeyCollectionManager. My current thinking is that we'll probably have just two different classes that define different schema for collection tables (one for name keys, another for int keys) but runtime behavior (caching/pre-fetching) of these classes would be controlled by the options that we pass in the configuration (need to extend it to support this feature).

PR #272, opened 2020-05-04
DM-24560: make 'repo' an argument, by convention always the first.

n8pease:


python/lsst/daf/butler/cli/opt/repo.py
```diff
@@ -23,13 +23,34 @@
...
+        Indicates if the caller must pass this argument to the command, by
+        default True.
+    help : str, optional
+        The help text for this argument to append to the command's help text.
+        If None or '' then nothing will be appended to the help text (in which
+        case the command should document this argument directly in its help
+        text). By default the value of repo_argument.existing_repo.
+    """
+
+    will_create_repo = "REPO is a filesystem path for the new repository. " \
```
timj: This is a URI to the new butler root. It doesn't have to be a filesystem path.

python/lsst/daf/butler/cli/opt/repo.py
```diff
@@ -23,13 +23,34 @@
...
+                       "Will be created if it does not exist."
+    existing_repo = "REPO is the path to an existing data repository root " \
```
timj: This is a URI to an existing repository root or configuration file. Does not have to be a path.

timj (approved):
Looks okay although we have to be careful about path vs URI.

PR #273, opened 2020-05-05
Attempt to run mypy on a subset of the package (PR is for travis testing)

TallJimbo:
The setup here is probably a little unusual, because it turned out to be quite tricky to get mypy to determine that the root of our package tree is `lsst` (which has no `__init__.py`), not `python` (which also has no `__init__.py`) or `daf` (the first directory that _does_ have such a file).  The only solution I could find was to run mypy from the `python` directory, with a config that enables namespace package support.

This was difficult to diagnose because if you run mypy on files or directories, it mostly doesn't care what the package root is, and seems to be working - but it will not obey configuration that's targeted at a particular directory, and AFAICT the config option that's supposed to warn about that (`warn_unused_configs = True`) didn't actually warn (maybe it doesn't work on globs?).  Anyhow, if you run it with `-p lsst.daf.butler` instead of on files/directories, it will complain if it doesn't recognize the package, so that seems much safer.

Anyhow, the config disables checking for all of `lsst.*`, then re-enables it for just `lsst.daf.butler.registry.interfaces.*`; we can enable other sub-packages as we clean them up.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ bin/*
 .mypy_cache/
 .idea/
 .vscode/
+mypy.log
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,4 +5,7 @@ matrix:
     - python: '3.7'
       install:
       - pip install flake8
-      script: flake8
+      - pip install mypy
+      script:
+      - flake8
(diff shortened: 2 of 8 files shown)


TallJimbo:
Superseded by #276

PR #274, opened 2020-05-06
DM-24721: fixes for Oracle compatibility

TallJimbo:


python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -566,27 +566,33 @@ def _convertTableSpec(self, name: str, spec: ddl.TableSpec, metadata: sqlalchemy
...
+        # sure we don't add duplicate explicit or foreign key indexes for
+        # those.
+        allIndexes = {tuple(fieldSpec.name for fieldSpec in spec.fields if fieldSpec.primaryKey)}
         args.extend(
             sqlalchemy.schema.UniqueConstraint(
                 *columns,
                 name=self.shrinkDatabaseEntityName("_".join([name, "unq"] + list(columns)))
             )
-            for columns in spec.unique if columns not in spec.indexes
+            for columns in spec.unique
```
MichelleGower, line 578: Just checking whether this line also needs a "if columns not in allIndexes" like the below spec.indexes section.

MichelleGower (approved):
Just one question.   Otherwise good to merge.

python/lsst/daf/butler/registry/interfaces/_database.py
TallJimbo: I don't think so - at present, it's basically saying that if someone defines a unique constraint that's identical to the primary key, that's a bug they should fix because both of those are explicit choices (as opposed to automatically indexing FKs, which is not).

PR #275, opened 2020-05-07
Improve connection string matching by ignoring driver value.

DinoBektesevic:
Ignore driver during DB authentication matching and ensure Oracle-Wallet-like URL is correctly passed through connection string unchanged.

### DIFF
--- a/python/lsst/daf/butler/registry/_dbAuth.py
+++ b/python/lsst/daf/butler/registry/_dbAuth.py
@@ -88,7 +88,7 @@ class DbAuth:
         except Exception as exc:
             raise DbAuthError(f"Unable to load DbAuth configuration file: {secretPath}.") from exc
 
-    def getAuth(self, drivername, username, host, port, database):
+    def getAuth(self, dialectname, username, host, port, database):
         """Retrieve a username and password for a database connection.
 
         This function matches elements from the database connection URL with
@@ -96,8 +96,8 @@ class DbAuth:
 
         Parameters
         ----------
-        drivername : `str`
-            Name of database backend driver from connection URL.
+        dialectname : `str`
+            Database dialect, f.e. sqlite, mysql, postgresql, oracle, or mssql.
         username : `str` or None
             Username from connection URL if present.
         host : `str`
@@ -131,6 +131,10 @@ class DbAuth:
         the database connection URL, the dictionary must also contain a
         ``username`` item.
 
+        The ``url`` item of the authorization dictionaries can contain both the
+        dialect and the driver of the database URL but the driver is ignored
+        during matching.
+
         Glob-style patterns (using "``*``" and "``?``" as wildcards) can be
         used to match the host and database name portions of the connection
         URL.  For the username, port, and database name portions, omitting them
@@ -156,8 +160,8 @@ class DbAuth:
         the pattern ``postgresql://host.example.com:5432``, even if the default
         port for the connection is 5432.
         """
-        if drivername is None or drivername == "":
-            raise DbAuthError("Missing drivername parameter")
+        if dialectname is None or dialectname == "":
+            raise DbAuthError("Missing dialectname parameter")
         if host is None or host == "":
             raise DbAuthError("Missing host parameter")
         if database is None or database == "":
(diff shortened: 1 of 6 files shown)


python/lsst/daf/butler/registry/_dbAuth.py
```diff
@@ -88,16 +88,16 @@ def __init__(self, path=None, envVar=None, authList=None):
...
 
         This function matches elements from the database connection URL with
         glob-like URL patterns in a list of configuration dictionaries.
 
         Parameters
         ----------
-        drivername : `str`
-            Name of database backend driver from connection URL.
+        dialectname : `str`
+            Database dialect, f.e. sqlite, mysql, postgresql, oracle, or mssql.
```
ktlim:
```suggestion
            Database dialect, for example sqlite, mysql, postgresql, oracle, or mssql.
```

python/lsst/daf/butler/registry/_dbAuth.py
```diff
@@ -131,6 +131,10 @@ def getAuth(self, drivername, username, host, port, database):
         the database connection URL, the dictionary must also contain a
         ``username`` item.
 
+        The ``url`` item of the authorization dictionaries can contain both the
+        dialect and the driver of the database URL but the driver is ignored
```
ktlim: Isn't that even more confusing?  Better to forbid it, I think.

python/lsst/daf/butler/registry/connectionString.py
```diff
@@ -47,9 +47,8 @@ class ConnectionStringFactory:
 
     @classmethod
     def fromConfig(cls, registryConfig):
-        """Parses the 'db' key in the config, and if they exist username,
-        password, host, port and database keys, and returns an connection
-        string object.
+        """Parses the 'db' and, if they exist, username, password, host, port
```
ktlim: Maybe make the quoting consistent?  Add a comma before "and".  Add "the" before "given".  Maybe say that the `db` key is what specifies the initial/default connection string?

python/lsst/daf/butler/registry/connectionString.py
```diff
@@ -81,14 +90,15 @@ def fromConfig(cls, registryConfig):
             if getattr(conStr, key) is None:
                 setattr(conStr, key, regConf.get(key))
 
-        # when host is None we cross our fingers and return
-        if conStr.host is None:
+        # allow 3rd party authentication mechanisms by assuming connection
+        # string is correct when we can not recognize (dialect, host, database)
```
ktlim:
```suggestion
        # Allow 3rd party authentication mechanisms by assuming connection
        # string is correct when we can not recognize (drivername, host, database)
```

python/lsst/daf/butler/registry/connectionString.py
```diff
@@ -71,6 +70,16 @@ def fromConfig(cls, registryConfig):
...
+
+        Notes
+        -----
+        Matching requires drivername, host and database keys. If drivername is
+        not specified in the db string and host and database keys are not found
+        in the db string, or as explicit keys in the config, the matcher
+        returns an unchanged connection string.
+        Insufficiently specified connection strings are interpreted as
+        indication that a 3rd party authentication mechanism, f.e. Oracle
+        Wallet, are used and are not modified.
```
ktlim:
```suggestion
        an indication that a 3rd party authentication mechanism such as Oracle
        Wallet is being used and therefore are left unmodified.
```

ktlim (approved):

python/lsst/daf/butler/registry/connectionString.py
DinoBektesevic: Added a comment that `db` specifies the default connection string and is returned unchanged when no additional information about it is provided. Thanks!

python/lsst/daf/butler/registry/_dbAuth.py
DinoBektesevic: I was unsure of how opinionated we should make DbAuth. I raise an error when this happens and I have updated the docstrings to reflect that.

python/lsst/daf/butler/registry/_dbAuth.py
DinoBektesevic: Expanded on other `f.e.`s in my docstrings too. Thanks!

PR #276, opened 2020-05-07
DM-24780: add initial mypy support.

TallJimbo:
Copying from #273, where this work started:
> The setup here is probably a little unusual, because it turned out to be quite tricky to get mypy to determine that the root of our package tree is `lsst` (which has no `__init__.py`), not `python` (which also has no `__init__.py`) or `daf` (the first directory that _does_ have such a file).  The only solution I could find was to run mypy from the `python` directory, with a config that enables namespace package support.
>
> This was difficult to diagnose because if you run mypy on files or directories, it mostly doesn't care what the package root is, and seems to be working - but it will not obey configuration that's targeted at a particular directory, and AFAICT the config option that's supposed to warn about that (`warn_unused_configs = True`) didn't actually warn (maybe it doesn't work on globs?).  Anyhow, if you run it with `-p lsst.daf.butler` instead of on files/directories, it will complain if it doesn't recognize the package, so that seems much safer.
>
> Anyhow, the config disables checking for all of `lsst.*`, then re-enables it for just `lsst.daf.butler.registry.interfaces.*`; we can enable other sub-packages as we clean them up.

### DIFF
--- a/python/lsst/daf/butler/registry/interfaces/_database.py
+++ b/python/lsst/daf/butler/registry/interfaces/_database.py
@@ -175,8 +175,7 @@ class Database(ABC):
 
     `Database` itself has several underscore-prefixed attributes:
 
-     - ``_cs``: SQLAlchemy objects representing the connection and transaction
-        state.
+     - ``_connection``: SQLAlchemy object representing the connection.
      - ``_metadata``: the `sqlalchemy.schema.MetaData` object representing
         the tables and other schema entities.
 
--- a/python/lsst/daf/butler/registry/interfaces/_database.py
+++ b/python/lsst/daf/butler/registry/interfaces/_database.py
@@ -37,6 +37,7 @@ from typing import (
     List,
     Optional,
     Sequence,
+    Set,
     Tuple,
 )
 import warnings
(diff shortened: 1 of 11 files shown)


python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -136,7 +137,8 @@ def addTableTuple(self, specs: Tuple[ddl.TableSpec, ...]) -> Tuple[sqlalchemy.sc
         is just a factory for `type` objects, not an actual type itself,
         we cannot represent this with type annotations.
         """
-        return specs._make(self.addTable(name, spec) for name, spec in zip(specs._fields, specs))
+        return specs._make(self.addTable(name, spec)                     # type: ignore
+                           for name, spec in zip(specs._fields, specs))  # type: ignore
```
rra, line 142: I feel like there's probably something better than `# type: ignore`, but `collections.namedtuple` is very weird in the type system, and I couldn't come up with anything better either.

rra (approved):
LGTM. For the record, here's the flags that I usually work towards in mypy (which does not imply that you should use them all from the start):

```
disallow_untyped_defs = True
disallow_incomplete_defs = True
strict_equality = True
warn_redundant_casts = True
warn_unreachable = True
warn_unused_ignores = True
```

TallJimbo:
Thanks for the tips.  Those stricter flags just required a lot of `-> None` and `**kwargs: Any`, so I've turned them on (and I'm glad to know now that those should be habits).

PR #277, opened 2020-05-08
DM-24786: Add string length constraint checks to sqlite and fix component length

timj:


TallJimbo (approved):
Looks good!  It's a bit of a pity we had to add a new hook to the `Database` base class, but I don't see a good way around it right now.

PR #278, opened 2020-05-11
DM-24829: Use a magic string for no component in datastore record

timj:
We were using an empty string but Oracle forces those
to be NULL and that breaks things.

### DIFF
--- a/python/lsst/daf/butler/datastores/fileLikeDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileLikeDatastore.py
@@ -62,6 +62,9 @@ from .genericDatastore import GenericBaseDatastore
 
 log = logging.getLogger(__name__)
 
+# String to use when a Python None is encountered
+NULLSTR = "__NULL_STRING__"
+
 
 class _IngestPrepData(Datastore.IngestPrepData):
     """Helper class for FileLikeDatastore ingest implementation.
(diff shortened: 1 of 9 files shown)


MichelleGower (approved):
I did not search for database admin recommendations to see if they differ from this solution.  I started running ci_hsc_gen3 with Oracle using this ticket branch.   It has made it past running ISR at the time of this comment.  Dug around the output repo filenames a little bit to make sure this string didn't end up in a filename and couldn't find any.  Approving the changes for merging.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -62,6 +62,9 @@
 
 log = logging.getLogger(__name__)
 
+# String to use when a Python None is encountered
+NULLSTR = "__NULL_STRING__"
```
timj, line 66: @TallJimbo I am pondering whether this should be shorter. Will people want to use this constant elsewhere at some point?

TallJimbo: Let's not worry about it.  I think if we did need to use sentinal empty strings in more places, we'd move more than just the constant into `Database`, and I don't have any anticipated use cases of that right now.

TallJimbo (approved):
Looks good.

Thank goodness the f-string lint didn't affect multi-line strings where some of the lines have substitutions and others don't (where I vastly prefer to start all lines with `f`).

PR #279, opened 2020-05-12
DM-24584: create an ingestRaws butler command

n8pease:


timj:
Quick comment: For the comment with message "add support for ingest-raw command" -- daf_butler doesn't really care about ingest-raw in obs_base and the commit doesn't seem to be immediately obvious as being about ingest-raw.  Can you change the commit message to be more relevant to what is really going on in daf_butler?

python/lsst/daf/butler/cli/utils.py
```diff
@@ -65,3 +130,82 @@ def to_upper(context, param, value):
...
+    ------
+    click.ClickException
+        An exception to be handled by the Click CLI tool.
+    """
+    # "DAF_BUTLER_MOCK" matches the key in the variable DAF_BUTLER_MOCK,
+    # defined in the top of this file.
+    if "DAF_BUTLER_MOCK" in os.environ:
+        printFunctionInfo(func, *args, **kwargs)
+        return
+    try:
```
n8pease, line 208: I tried to implement this with a unittest.mock, but couldn't find a way to get the called mock object back out to the calling script for verification. I couldn't pass it in because butler wants to import the command, and even when replacing the object in sys.modules, import still provides a copy of the class, not the class itself :-(

timj:
Wasn't I supposed to be reviewing this PR?

python/lsst/daf/butler/cli/opt/__init__.py
```diff
@@ -19,6 +19,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from .config import config_option
+from .config_file import config_file_option
```
n8pease, line 23: test

python/lsst/daf/butler/cli/utils.py
```diff
@@ -19,9 +19,20 @@
...
+
 from ..core.utils import iterable
 
 
+# DAF_BUTLER_MOCK is set by some tests as an environment variable and indicates
+# to the cli_handle_exception function that instead of executing the command
+# implementation function it should print details about the called command to
+# stdout. These details are then used to verify the command function was loaded
+# and received expected inputs.
+DAF_BUTLER_MOCK = {"DAF_BUTLER_MOCK": ""}
```
timj, line 33: It reads like the variable here is the contents of the environment variable but environment variables can't be a dict so is the key in the dict the environment variable name and the value in the dict what is set by the user? Having the global use the same name as the key is confusing here.

python/lsst/daf/butler/cli/utils.py
```diff
@@ -49,6 +64,56 @@ def split_commas(context, param, values):
...
+    `dict` : [`str`, `str`]
+        The passed-in values in dict form.
+
+    Raises
+    ------
+    `click.ClickException`
+        Raised if the separator is not found in an entry, or if duplicate keys
+        are encountered.
+    """
+    if "," == separator or " " == separator:
```
timj, line 102: Sometimes for future expansion possibilities it's better to write this as

```python
if separator in (",", " "):
```

python/lsst/daf/butler/cli/utils.py
```diff
@@ -65,3 +130,82 @@ def to_upper(context, param, value):
...
+    function : `str`
+        The name of the function that was was expected to have been called.
+    expectedArgs : [`str`]
+        The values of the arguments that should have been passed to the
+        function.
+    expectedKwargs : `dict` [`str`, `str`]
+        The names and values of the kwargs that should have been passsed to the
+        funciton.
+    """
+    calledWith = eval(output)
```
timj, line 174: if we are going to be using `eval` on the output that we generated above, I think it would be safer all around if we made the "print" do a yaml dump and here we did a yaml safe_load.

python/lsst/daf/butler/cli/utils.py
```diff
@@ -65,3 +130,82 @@ def to_upper(context, param, value):
...
+               kwargs=kwargs))
+
+
+def verifyFunctionInfo(testSuite, output, function, expectedArgs, expectedKwargs):
+    """For unit testing butler subcommand call execution, compare a dict that
+    has been printed to stdout to expected data.
+
+    Parameters
+    ----------
+    testSuite : `unittest.Testsuite`
```
timj, line 160: Are you sure this isn't a `TestCase` rather than a `TestSuite`?

timj (commented):
Looks fine overall. If the cli wrappers are only going to be minimalist and never have business logic in them there is no real reason to split them into separate files in the cli tree unless the import overhead is really high.   Did only one of the commands get relocated to scripts?

n8pease:
@timj, I did only move one command function into script. I didn't want to do all of them without approval on the concept so I did this one to demonstrate. Since it seems like you agree with the concept I'll make a ticket to do the other butler commands.
CORRECTION: I did the commands in obs_base. There's a couple left in daf_butler. Will ticket.
UPDATE: the ticket is DM-24937

PR #280, opened 2020-05-12
DM-24807: Add type annotations to datastore

timj:


python/lsst/daf/butler/core/storedFileInfo.py
```diff
@@ -45,8 +47,9 @@ class StoredFileInfo(StoredDatastoreItemInfo):
     __slots__ = {"formatter", "path", "storageClass", "component",
                  "checksum", "file_size"}
 
-    formatter: str
-    """Fully-qualified name of Formatter."""
+    formatter: Union[str, Formatter, Type[Formatter]]
```
timj: This is quite problematic since it stops us from saying that we accept these three types but it will always be stored internally as a str.

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -311,7 +325,8 @@ def _lookupNames(self) -> Tuple[LookupKey]:
         return names
 
     @staticmethod
-    def flatten(refs: Iterable[DatasetRef], *, parents: bool = True) -> Iterator[DatasetRef]:
+    def flatten(refs: Iterable[Union[FakeDatasetRef, DatasetRef]], *,
+                parents: bool = True) -> Iterator[DatasetRef]:
```
TallJimbo: I don't think this one is the right way to fix whatever problem inspired it - if it's given a `FakeDatasetRef`, there's no way it's going to return only true `DatasetRefs`.  If the problem is that this is used to implement `FakeDatasetRef.flatten`, we should just `# type: ignore` that implementation.  There are probably other solutions, too, but they're more involved and that one seems pretty pragmatic.

python/lsst/daf/butler/core/datastore.py
```diff
@@ -277,23 +288,23 @@ def __init__(self, config, bridgeManager, butlerRoot=None):
...
+    def __str__(self) -> str:
         return self.name
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return self.name
 
     @property
-    def names(self):
+    def names(self) -> List[str]:
```
TallJimbo: Returning a `list` via a read-only property might be confusing - it's possible to write something like `obj.names[0] = "zero"`, and have it silently do nothing.  Could you return a `tuple` here instead?

python/lsst/daf/butler/core/datastore.py
```diff
@@ -277,23 +288,23 @@ def __init__(self, config, bridgeManager, butlerRoot=None):
...
 
-        Some datastores can have child datastores.
+        Can be different to ``name`` for a chaining datastore.
         """
         # Default implementation returns solely the name itself
         return [self.name]
 
     @contextlib.contextmanager
-    def transaction(self):
+    def transaction(self) -> Generator:
```
TallJimbo: I think you can give MyPy a bit more to work with by saying `Generator[None]`.

If that's the syntax for saying it yields `None`, that is - I'm not sure for `Generator`.  I've tended to use `Iterator` instead of `Generator` for functions decorated by `contextmanager`, because `Generator` feels like an implementation choice ather than an interface.  But `Generator` was put in `typing` by smarter people than me, so maybe I'm missing some reason it's better.

python/lsst/daf/butler/core/datastore.py
```diff
@@ -667,12 +687,13 @@ def export(self, refs: Iterable[DatasetRef], *,
         raise NotImplementedError(f"Transfer mode {transfer} not supported.")
 
     @abstractmethod
-    def validateConfiguration(self, entities, logFailures=False):
+    def validateConfiguration(self, entities: Iterable[Union[DatasetRef, DatasetType, StorageClass]],
+                              logFailures: bool = False) -> None:
```
TallJimbo, line 694: "Entity" might merit an alias.

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -262,7 +267,7 @@ def filterParameters(self, parameters, subset=None):
                              f" known parameters ({known})")
         return {k: parameters[k] for k in known if k in parameters}
 
-    def validateInstance(self, instance):
+    def validateInstance(self, instance: object):
```
TallJimbo: I have been using `typing.Any` instead of `object`; I _think_ they're equivalent.  We should probably standardize on one.

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -278,14 +283,15 @@ def validateInstance(self, instance):
         """
         return isinstance(instance, self.pytype)
 
-    def __eq__(self, other):
+    def __eq__(self, other: object):
```
TallJimbo: Huh, I guess mypy doesn't complain about no return type annotation here?  I've been adding them for `__eq__`, but maybe I shouldn't be.

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -353,7 +359,7 @@ class StorageClassFactory(metaclass=Singleton):
         is located in the ``storageClasses`` section.
     """
 
-    def __init__(self, config=None):
+    def __init__(self, config: Optional[Union[StorageClassConfig, str]] = None):
```
TallJimbo, line 379: Is `Union[StorageClassConfig, str, None]` better?  I'm not sure it is, but we should be intentional about establishing a convention one way or the other.

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -464,7 +470,7 @@ def processStorageClass(name, sconfig):
             processStorageClass(name, sconfig)
 
     @staticmethod
-    def makeNewStorageClass(name, baseClass=StorageClass, **kwargs):
+    def makeNewStorageClass(name, baseClass: Type = StorageClass, **kwargs) -> Type:
```
TallJimbo: I think `Type[StorageClass]` would further check that arguments and return values are subclasses of `StorageClass`.

python/lsst/daf/butler/core/storedFileInfo.py
TallJimbo: Looks like there's no particularly good answer (https://stackoverflow.com/questions/51737828/python-dataclasses-what-type-to-use-if-post-init-performs-type-conversion doesn't provide one, for example).

I would probably lean towards just defining `__init__` directly instead of `__post_init__` so the parameter type can differ from the attribute type.  That's how I was planning to handle the `TableSpec` problem, which is essentially identical to this one.  But of course that's not ideal either.

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -230,7 +248,7 @@ def exists(self, ref):
                 return True
         return False
 
-    def get(self, ref, parameters=None):
+    def get(self, ref: DatasetRef, parameters: Mapping[str, Any] = None) -> Any:
```
TallJimbo: Looks like this needs to be `Optional[Mapping[str, Any]]`.  Did mypy maybe not get enabled on this file?

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -377,12 +395,13 @@ def isDatasetAcceptable(dataset, *, name, constraints):
         # NotImplementedError being raised.
         allFailuresAreNotImplementedError = True
         for datastore, constraints in zip(self.datastores, self.datastoreConstraints):
+            okForChild: List[FileDataset]
```
TallJimbo, line 398: Ah, neat - I didn't know this syntax was possible.  I'm sure I'll use it in the future.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -307,10 +325,13 @@ def addStoredItemInfo(self, refs, infos):
             )
         self._table.insert(*records)
 
-    def getStoredItemInfo(self, ref):
+    def getStoredItemInfo(self, ref: Union[DatasetRef, FakeDatasetRef]) -> StoredFileInfo:
```
TallJimbo: `DatasetIdRef` (in `registry.interfaces`) is a type alias for this union.

Open to ideas for better names for that.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -799,7 +863,7 @@ def getUri(self, ref, predict=False):
 
         return location.uri
 
-    def get(self, ref, parameters=None):
+    def get(self, ref: DatasetRef, parameters: Mapping[str, Any] = None) -> Any:
```
TallJimbo: Another one that should be `Optional`.

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -19,22 +19,34 @@
...
 import shutil
-from typing import Iterable, Optional, Type
+from typing import (TYPE_CHECKING,
+                    Any,
+                    ClassVar,
+                    Iterable,
+                    Optional,
+                    Type,
+                    Union
+                    )
```
TallJimbo: Different formatting here than in the other files.

python/lsst/daf/butler/core/datastore.py
```diff
@@ -30,7 +30,19 @@
...
+                    ClassVar,
+                    Generator,
+                    Iterable,
+                    List,
+                    Mapping,
+                    Optional,
+                    Set,
+                    Type,
+                    Union,
+                    )
```
TallJimbo: Another differently-formatted import pattern.

python/mypy.ini
```diff
@@ -33,3 +33,12 @@ disallow_incomplete_defs = True
...
 warn_unused_ignores = True
+
+[mypy-lsst.daf.butler.datastores.*]
+ignore_missing_imports = False
+ignore_errors = False
+disallow_untyped_defs = True
+disallow_incomplete_defs = True
+strict_equality = True
+warn_unreachable = True
+warn_unused_ignores = True
```
TallJimbo, line 44: Can you also turn mypy on for a few `core` modules after the changes you've made to them?  Or does that just bring new errors?

TallJimbo (approved):
Lots of very minor comments, none of them critical.  Overall this is of course a very nice cleanup.

python/lsst/daf/butler/core/storageClass.py
timj: Yes, I'll use Any. What you are seeing is my experience with typing evolving over a week.

python/lsst/daf/butler/core/datastore.py
timj: I used Generator because I looked at what you had done in another example. I can change.

python/lsst/daf/butler/core/datastore.py
timj: That is a good point. I should be returning a tuple.

python/lsst/daf/butler/core/datasets/ref.py
timj: Yes, this is an implementation detail in that I didn't want to duplicate the code from DatasetRef. I'm starting to wonder if inheriting from DatasetRef and then stubbing out all the methods that we know won't work wouldn't be easier.

python/lsst/daf/butler/datastores/chainedDatastore.py
timj: It was enabled. I had convinced myself that all the parameter types where we assign a None default were automatically being tagged Optional by mypy (since clearly we are initializing with None). For that reason I stopped bothering with the explicit Optional. Maybe I was deluding myself.

python/lsst/daf/butler/core/storedFileInfo.py
timj: I think we had a reason for using `__post_init__` since it seemed designed for this and didn't result in me trying to ensure that I implemented `__init__` for the dataclass properly.

python/lsst/daf/butler/datastores/chainedDatastore.py
TallJimbo: Interesting!  Looks like you actually discovered a MyPy feature I didn't know about: https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-optional

But it sounds like disabling this behavior may become the default in the future, so we might want to be explicit with `Optional` anyway (and add that flag to our config).

timj: Ok. I will make it explicit but it seems weird to me since it's so clear that I clearly mean Optional from the defaulting.

python/lsst/daf/butler/core/storageClass.py
timj: We've been using `Optional[something] = None` elsewhere so I'm fine with doing that.

python/lsst/daf/butler/core/storedFileInfo.py
timj: I've rewritten it and things seem to work okay now.

python/lsst/daf/butler/core/datasets/ref.py
timj: I fixed it by adding a `allRefs` method so that mypy no longer gets confused abut the return type of the static method.

PR #286, opened 2020-05-19
DM-24979: config_file_option does not save the 'help' input argument

n8pease:


python/lsst/daf/butler/cli/opt/config_file.py
```diff
@@ -24,11 +24,15 @@
 
 
 class config_file_option:  # noqa: N801
-    def __init__(self, required=False, help=None):
+
+    defaultHelp = "The path to the config file."
```
timj, line 28: This help doesn't make it clear to reader which type of config this is. Is it the config file for a butler or the config file for a task? Given that the butler config file is only used in `butler create` (after that you get a choice of specifying the config file or the repo root) we may want to consider it's name more carefully.

tests/test_cliOption.py
```diff
@@ -0,0 +1,144 @@
...
+
+import click
+import click.testing
+import unittest
+import yaml
+
+from lsst.daf.butler.cli.opt import config_file_option, config_option, dataset_type_option
+
+
+class DatasetTypeSuite(unittest.TestCase):
```
timj: It's a TestCase not a Suite.

tests/test_cliOption.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def test_help(self):
+        """test capture of the help text"""
+        runner = click.testing.CliRunner()
+        result = runner.invoke(DatasetTypeSuite.cli, ["--help"])
+        self.assertEqual(result.exit_code, 0)
+        self.assertIn("the dataset type", result.stdout)
+
+
+class ConfigSuite(unittest.TestCase):
```
timj: Again TestCase not TestSuite

timj (approved):

python/lsst/daf/butler/cli/opt/config_file.py
n8pease: Is it always going to be a particular kind of configuration file for any use of --config-file? I guess I'm not clear if shared options should always be absolutely exactly the same, or if there is room for flexibility, and how much of that. `create` passes its own help text into config_file_option, maybe there shouldn't be a default help and the help arg should be required. I was trying to split the difference, it's possible/likely I haven't found the sweet spot just yet.

n8pease: Right now this option serves two purposes, for repo cfg (`butler create`) and pex_config in other places. Since we are considering changing the `create --config-file` option to a different name, this can be made more specific for pex_config files, when we implement that change.

PR #288, opened 2020-05-20
DM-24851: Add getURIs method

timj:
This supports composite disassembly in datastore and will also be used for virtual composite datasets.

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -825,6 +825,59 @@ class Butler:
         ref = self._findDatasetRef(datasetRefOrType, dataId, collections=collections, **kwds)
         return self.getDirect(ref, parameters=parameters)
 
+    def getURIs(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
+                dataId: Optional[DataId] = None, *,
+                predict: bool = False,
+                collections: Any = None,
+                run: Optional[str] = None,
+                **kwds: Any) -> Tuple[Optional[ButlerURI], Dict[str, ButlerURI]]:
+        """Returns the URIs associated with the dataset.
+
+        Parameters
+        ----------
+        datasetRefOrType : `DatasetRef`, `DatasetType`, or `str`
+            When `DatasetRef` the `dataId` should be `None`.
+            Otherwise the `DatasetType` or name thereof.
+        dataId : `dict` or `DataCoordinate`
+            A `dict` of `Dimension` link name, value pairs that label the
+            `DatasetRef` within a Collection. When `None`, a `DatasetRef`
+            should be provided as the first argument.
+        predict : `bool`
+            If `True`, allow URIs to be returned of datasets that have not
+            been written.
+        collections : Any, optional
+            Collections to be searched, overriding ``self.collections``.
+            Can be any of the types supported by the ``collections`` argument
+            to butler construction.
+        run : `str`, optional
+            Run to use for predictions, overriding ``self.run``.
+        kwds
+            Additional keyword arguments used to augment or construct a
+            `DataCoordinate`.  See `DataCoordinate.standardize`
+            parameters.
+
+        Returns
+        -------
+        primary : `ButlerURI`
+            The URI to the primary artifact associated with this dataset.
+            If the dataset was disassembled within the datastore this
+            may be `None`.
+        components : `dict`
+            URIs to any components associated with the dataset artifact.
+            Can be empty if there are no components.
+        """
+        ref = self._findDatasetRef(datasetRefOrType, dataId, allowUnresolved=predict,
+                                   collections=collections, **kwds)
+        if ref.id is None:  # only possible if predict is True
+            if run is None:
+                run = self.run
+                if run is None:
+                    raise TypeError("Cannot predict location with run=None.")
+            # Lie about ID, because we can't guess it, and only
+            # Datastore.getUri() will ever see it (and it doesn't use it).
+            ref = ref.resolved(id=0, run=self.run)
+        return self.datastore.getURIs(ref, predict)
+
     def getUri(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
                dataId: Optional[DataId] = None, *,
                predict: bool = False,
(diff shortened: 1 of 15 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -825,12 +825,65 @@ def get(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
...
+            if run is None:
+                run = self.run
+                if run is None:
+                    raise TypeError("Cannot predict location with run=None.")
+            # Lie about ID, because we can't guess it, and only
+            # Datastore.getURIs() will ever see it (and it doesn't use it).
+            ref = ref.resolved(id=0, run=self.run)
+        return self.datastore.getURIs(ref, predict)
+
+    def getURI(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
```
MichelleGower, line 881: I see what this is doing, but I was surprised to see this single URI form still here.   Might be worth having a comment pointing to getURIs as the more future proof function.

python/lsst/daf/butler/_butler.py
```diff
@@ -825,12 +825,65 @@ def get(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
...
+        Returns
+        -------
+        primary : `ButlerURI`
+            The URI to the primary artifact associated with this dataset.
+            If the dataset was disassembled within the datastore this
+            may be `None`.
+        components : `dict`
+            URIs to any components associated with the dataset artifact.
+            Can be empty if there are no components.
+        """
```
MichelleGower, line 868: Does it raise if dataset not found in Datastore or returns None, {}?  Question is more about the docstring than the code.

python/lsst/daf/butler/datastores/inMemoryDatastore.py
```diff
@@ -394,17 +439,10 @@ def getUri(self, ref: DatasetRef, predict: bool = False) -> str:
...
-                raise FileNotFoundError("Dataset {} not in this datastore".format(ref))
-            name = "{}#predicted".format(ref.datasetType.name)
-        else:
-            realID, _ = self._get_dataset_info(ref)
-            name = '{}'.format(id(self.datasets[realID]))
-
-        return "mem://{}".format(name)
+        uri, _ = self.getURIs(ref, predict)
+        if uri is None:
+            raise RuntimeError(f"Unexpectedly got no URI for in-memory datastore for {ref}")
```
MichelleGower: Just checking that the changing of the type of exception raised was intentional.

tests/test_butler.py
```diff
@@ -392,11 +397,65 @@ def testBasicPutGet(self):
 
     def testCompositePutGetConcrete(self):
         storageClass = self.storageClassFactory.getStorageClass("StructuredData")
-        self.runPutGetTest(storageClass, "test_metric")
+        butler = self.runPutGetTest(storageClass, "test_metric")
+
+        # Should *not* be disassembled
+        datasets = list(butler.registry.queryDatasets(..., collections="ingest"))
+        self.assertEqual(len(datasets), 1)
+        uri, components = butler.getURIs(datasets[0])
```
MichelleGower, line 405: I probably missed it, but is there a test of getURIs for a dataset type that isn't composite.  If I understood these tests correctly they test composite (with components) that was written as a single file, and the later one tests composite written as multiple files.  getURIs for a dataset type that isn't composite should give the same results as the single file, but doesn't guarantee that the code accidentally requires components.

python/lsst/daf/butler/_butler.py
```diff
@@ -878,17 +931,11 @@ def getUri(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
...
-                if run is None:
-                    raise TypeError("Cannot predict location with run=None.")
-            # Lie about ID, because we can't guess it, and only
-            # Datastore.getUri() will ever see it (and it doesn't use it).
-            ref = ref.resolved(id=0, run=self.run)
-        return self.datastore.getUri(ref, predict)
+        primary, _ = self.getURIs(datasetRefOrType, dataId=dataId, predict=predict,
+                                  collections=collections, run=run, **kwds)
+        if primary is None:
+            raise RuntimeError(f"Found dataset but no single URI retrieved for it {datasetRefOrType}")
```
MichelleGower: I'm not sure whether this error message (which is repeated a few times in the code) means found dataset in registry, but either 0 or > 1 URI retrieved for it (related to previous question about what getURIs does if dataset in registry but 0 datastore entries), or whether found dataset in datastore (and registry), but found > 1 URI for it.

MichelleGower (approved):
A couple comments and questions.    Otherwise good to merge.

python/lsst/daf/butler/_butler.py
timj: It raises if the dataset is missing and predict=False. ie you should never get `None, {}` returned. You either get something real, something that might be real, or an exception.

python/lsst/daf/butler/datastores/inMemoryDatastore.py
timj: This is really an assertion (but I never use assert). I could change it to AssertionError. It's really a "this is impossible" check since getURIs will never do disassembly (so components is false) and None should never be returned if getURIs returns.

tests/test_butler.py
timj: Good point. We don't actually run any of the tests here with a non-composite so adding such a thing is probably more than I want to do now.

python/lsst/daf/butler/_butler.py
timj: Yes. People shouldn't really use it. I left it around mostly because it *seems* more like the gen2 version. I'll ask on Slack.

python/lsst/daf/butler/_butler.py
timj: This is specifically trapping for the case where someone is using getURI and the dataset does exist but it's disassembled so getURI is at that point useless to you. The `if None` here also convinces mypy that I'm never going to return `Optional[ButlerURI]` but always `ButlerURI`.

PR #289, opened 2020-05-21
DM-25000: Make registry query methods work with composition again.

TallJimbo:


python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -457,6 +458,53 @@ def testDatasetTypeComponentQueries(self):
...
+        self.loadData(registry, "datasets.yaml")
+        # Test getting the child dataset type (which does still exist in the
+        # Registry), and check for consistency with
+        # DatasetRef.makeComponentRef.
+        collection = "imported_g"
+        parentType = registry.getDatasetType("permabias")
+        childType = registry.getDatasetType("permabias.wcs")
+        parentRefResolved = registry.findDataset(parentType, collections=collection,
+                                                 instrument="Cam1", detector=1)
+        self.assertIsNotNone(parentRefResolved)
```
timj: Maybe we could be more positive in this test and use:
```
self.assertIsInstance(parentRefResolved, DatasetRef)
```
?

python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -457,6 +458,53 @@ def testDatasetTypeComponentQueries(self):
...
+        self.assertEqual(childType, parentRefResolved.makeComponentRef("wcs").datasetType)
+        parentRefUnresolved = parentRefResolved.unresolved()
+        # Search for a single dataset with findDataset.
+        childRef1 = registry.findDataset("permabias.wcs", collections=collection,
+                                         dataId=parentRefUnresolved.dataId)
```
timj: Isn't the dataId of parentRefUnresolved the same as the dataId for parentRefResolved?

timj (approved):
Looks ok.

python/lsst/daf/butler/registry/tests/_registry.py
TallJimbo: Good catch; the unresolved variable's existence is a relic of a previous incarnation of the test, and I've removed it.

PR #291, opened 2020-05-27
DM-25080: finish typing fixes in daf.butler.registry

TallJimbo:


python/lsst/daf/butler/registry/_config.py
```diff
@@ -81,14 +83,26 @@ def makeDefaultDatabaseUri(self, root: str):
...
+        ----------
+        root : `str` or `None`
+            String to substitute for `BUTLER_ROOT_TAG`.  Passing `None` here is
+            allowed only as a convenient way to raise an exception
+            (`ValueError`).
+
+        Raises
+        ------
+        ValueError
+            Raised if ``butlerRoot`` is not set but a value is required.
```
andy-slac: `butlerRoot` -> `root` ?

python/lsst/daf/butler/registry/_dbAuth.py
```diff
@@ -260,8 +269,10 @@ def getUrl(self, url):
             components.username, components.hostname, components.port,
             components.path.lstrip("/"))
         hostname = components.hostname
+        assert hostname is not None
```
andy-slac, line 272: I am a bit worried about the use of `assert` for input validation, does it mean that we do not care about validation in optimized mode (or we never run in optimized mode)?

python/lsst/daf/butler/registry/queries/exprParser/exprTree.py
```diff
@@ -164,15 +170,15 @@ class TimeLiteral(Node):
     value : str
         Literal string value.
```
andy-slac, line 172: I think I forgot to update this docstring when we switched to astropy, the value type here should be `astropy.time.Time`, not `str`, my fault.

python/lsst/daf/butler/registry/queries/exprParser/exprTree.py
```diff
@@ -164,15 +170,15 @@ class TimeLiteral(Node):
...
     """
-    def __init__(self, value):
+    def __init__(self, value: str):
```
andy-slac: `astropy.time.Time` instead of `str`

python/lsst/daf/butler/registry/queries/exprParser/treeVisitor.py
```diff
@@ -63,19 +72,19 @@ def visitStringLiteral(self, value, node):
         """
 
     @abstractmethod
-    def visitTimeLiteral(self, value, node):
+    def visitTimeLiteral(self, value: str, node: Node) -> T:
```
andy-slac: Here too `astropy.time.Time`.

python/lsst/daf/butler/registry/queries/exprParser/treeVisitor.py
```diff
@@ -63,19 +72,19 @@ def visitStringLiteral(self, value, node):
...
         """Visit TimeLiteral node.
 
         Parameters
         ----------
-        value : `TimeLiteral`
+        value : `str`
```
andy-slac: And here `astropy.time.Time`.

python/lsst/daf/butler/registry/queries/expressions.py
```diff
@@ -108,51 +109,52 @@ class InspectionVisitor(TreeVisitor):
...
         # Docstring inherited from TreeVisitor.visitNumericLiteral
         pass
 
-    def visitStringLiteral(self, value, node):
+    def visitStringLiteral(self, value: str, node: Node) -> None:
         # Docstring inherited from TreeVisitor.visitStringLiteral
         pass
 
-    def visitTimeLiteral(self, value, node):
+    def visitTimeLiteral(self, value: str, node: Node) -> None:
```
andy-slac: One more `astropy.time.Time`

python/lsst/daf/butler/registry/queries/expressions.py
```diff
@@ -188,53 +190,58 @@ class ClauseVisitor(TreeVisitor):
...
+        return sqlalchemy.sql.literal(coerced)
 
-    def visitStringLiteral(self, value, node):
+    def visitStringLiteral(self, value: str, node: Node) -> sqlalchemy.sql.ColumnElement:
         # Docstring inherited from TreeVisitor.visitStringLiteral
-        return literal(value)
+        return sqlalchemy.sql.literal(value)
 
-    def visitTimeLiteral(self, value, node):
+    def visitTimeLiteral(self, value: str, node: Node) -> sqlalchemy.sql.ColumnElement:
```
andy-slac: And this is probably the last one that needs `astropy.time.Time`.

python/lsst/daf/butler/registry/wildcards.py
```diff
@@ -374,14 +407,13 @@ def _yieldCollectionRecords(
...
-    withRestrictions: bool = False,
     done: Optional[Set[str]] = None,
     flattenChains: bool = True,
     includeChains: Optional[bool] = None,
-) -> Iterator[CollectionRecord]:
+) -> Iterator[Tuple[CollectionRecord, DatasetTypeRestriction]]:
     """A helper function containing common logic for `CollectionSearch.iter`
-    and `CollectionQuery.iter`: yield a single `CollectionRecord` only if it
-    matches the criteria given in other arguments.
+    and `CollectionQuery.iter`: recursively yield `CollectionRecord` only they
```
andy-slac, line 415: Looks like `if` is missing (`only if they`)?

python/lsst/daf/butler/registry/wildcards.py
```diff
@@ -754,16 +788,60 @@ def iter(
...
+            Object responsible for managing the collection tables in a
+            `Registry`.
+        datasetType : `DatasetType`, optional
+            If given, only yield collections whose dataset type restrictions
+            include this dataset type.
+        collectionType : `CollectionType`, optional
+            If given, only yield collections of this type.
+        flattenChains : `bool`, optional
+            If `True` (default) recursively yield the child collections of
+            `~CollectionType.CHAINED` colelctions.
```
andy-slac: Typo: colelctions

andy-slac (approved):
Looks OK, few minor comments.

python/lsst/daf/butler/registry/_dbAuth.py
TallJimbo: The earlier call to `self.getAuth` with this as a parameter will do that input validation, so this assert is just a guard against that changing.

python/lsst/daf/butler/registry/queries/exprParser/exprTree.py
TallJimbo: Sad that mypy didn't catch these.  At first I thought it was just because I told it not to check the `ply` package, so it couldn't tell what the original type was but still enforced consistency.  But I tested that theory while fixing this, and it's not even testing consistency.  My guess is that because `astropy` doesn't provide type information, it figures it can't tell that `astropy.time.Time` isn't an alias to `str` or contravariant with it, but that's just a guess.

PR #292, opened 2020-05-27
DM-25072: Move afw code to obs_base

timj:
To support that move I have added `Config.fromYAML` and modifed the test createRepo config behavior.


python/lsst/daf/butler/core/config.py
```diff
@@ -75,7 +77,11 @@ class Loader(yaml.CSafeLoader):
 
     def __init__(self, stream):
         super().__init__(stream)
-        self._root = ButlerURI(stream.name)
+        # if this is a string and not a stream we may well lack a name
+        try:
+            self._root = ButlerURI(stream.name)
+        except AttributeError:
+            self._root = None
```
kfindeisen: The change in invariants is a bit worrying, since object fields that are `None` are a great source of bugs. In this case this seems to be effectively the "tagged class" anti-pattern (where a class can do very different things depending on its state); consider splitting into a class that always has `self._root` and one that never needs it, and presenting them to the user through a common base class.

kfindeisen: If the argument is not necessarily a stream, should it still be called `stream`?

python/lsst/daf/butler/core/config.py
```diff
@@ -99,6 +105,8 @@ def include(self, node):
             raise yaml.constructor.ConstructorError
 
     def extractFile(self, filename):
+        if self._root is None:
+            raise RuntimeError("Unable to load !include directive from YAML string")
```
kfindeisen: It doesn't seem like the ability to use `!include` should be affected by whether `self` represents a real file or not, at least as long as `filename` is absolute. Is there a ticket for fixing this later (would probably require reimplementing this method entirely)?

python/lsst/daf/butler/core/config.py
```diff
@@ -298,7 +322,7 @@ def __initFromYaml(self, stream):
         Parameters
         ----------
         stream
-            To a persisted config file in YAML format.
+            To a persisted config file in YAML format. Can be a string.
```
kfindeisen: No longer a `stream`...

If you're changing the documentation anyway, I suggest rewording to use complete sentences. It's a bit distracting to have to figure out that it's really "Stream from a persisted config..."

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -68,13 +72,16 @@ def makeTestRepo(root, dataIds, *, config=None, **kwargs):
...
+    defaults["datastore", "cls"] = "lsst.daf.butler.datastores.inMemoryDatastore.InMemoryDatastore"
+    defaults["datastore", "checksum"] = False  # In case of future changes
+    defaults["registry", "db"] = "sqlite:///:memory:"
+
+    if config:
+        defaults.update(config)
+
     # newConfig guards against location-related keywords like outfile
-    newConfig = Butler.makeRepo(root, config=config, **kwargs)
+    newConfig = Butler.makeRepo(root, config=defaults, **kwargs)
```
kfindeisen, line 84: I like this approach, but the name `defaults` is misleading once you call `update`. Maybe `mergedConfig` to distinguish it from the function parameter?

tests/test_cliPluginLoader.py
```diff
@@ -79,6 +79,9 @@ class Suite(unittest.TestCase):
     def setUp(self):
         butler.cli.commands = None
 
+    def tearDown(self):
+        butler.cli.commands = None
```
kfindeisen, line 83: If clearing `butler.cli.commands` is essential for test integrity, then I recommend the following:

    @staticmethod
    def _clearCommands():
        butler.cli.commands = None

    def setUp(self):
        self._clearCommands()
        self.addCleanup(self._clearCommands)

While the current `setUp` cannot raise exceptions (assuming `butler.cli.commands` exists), this may not be true in the future, and `addCleanup` provides a stronger guarantee that it will run.

kfindeisen: I'm a bit curious about the bug, though. Why is it helpful to set `butler.cli.commands = None` *both* before and after each test?

python/lsst/daf/butler/_butler.py
```diff
@@ -207,7 +206,6 @@ def __init__(self, config: Union[Config, str, None] = None, *,
             self.registry = butler.registry
             self.datastore = butler.datastore
             self.storageClasses = butler.storageClasses
-            self._composites = butler._composites
```
kfindeisen, line 210: For future reference, I think removing `Butler._composites` and removing the corresponding Butler config lines should have been a separate commit from moving `composites.yaml` and updating references to it, since it's a logically distinct change (and presumably the code would have worked without moving the file).

python/lsst/daf/butler/core/composites.py
```diff
@@ -71,7 +86,7 @@ def __init__(self, config, *, universe):
         # the values
         self._lut = processLookupConfigs(self.config[DISASSEMBLY_KEY], universe=universe)
 
-    def shouldBeDisassembled(self, entity):
+    def shouldBeDisassembled(self, entity: Union[DatasetRef, DatasetType, StorageClass]) -> bool:
```
kfindeisen, line 89: Is there a standard alias for this? I think I've seen this combination elsewhere in the `daf_butler` docs.

python/lsst/daf/butler/core/composites.py
```diff
@@ -103,11 +118,11 @@ def shouldBeDisassembled(self, entity):
             log.debug("%s will not be disassembled (not a composite)", entity)
             return False
 
-        matchName = "{} (via default)".format(entity)
+        matchName: Union[LookupKey, str] = "{} (via default)".format(entity)
```
kfindeisen, line 121: Isn't this always `str`???

python/lsst/daf/butler/core/assembler.py
```diff
@@ -70,11 +76,12 @@ class CompositeAssembler:
         `StorageClass` to be used with this assembler.
     """
 
-    def __init__(self, storageClass):
+    def __init__(self, storageClass: StorageClass):
+        assert storageClass is not None
```
kfindeisen, line 80: Why `assert` (and therefore `AssertionError`) and not `ValueError`?

python/lsst/daf/butler/core/assembler.py
```diff
@@ -70,11 +76,12 @@ class CompositeAssembler:
...
         self.storageClass = storageClass
 
     @staticmethod
-    def _attrNames(componentName, getter=True):
+    def _attrNames(componentName: str, getter: bool = True) -> Tuple[str, ...]:
         """Return list of suitable attribute names to attempt to use.
```
kfindeisen, line 85: (Pre-existing) inconsistent documentation.

I would argue that a list makes more sense than a tuple semantically, since it's an arbitrary-length sequence of like things...

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -961,13 +961,17 @@ def get(self, ref: DatasetRef, parameters: Optional[Mapping[str, Any]] = None) -
...
+            components: Dict[str, Any] = {}
             for getInfo in allGetInfo:
                 # assemblerParams are parameters not understood by the
                 # associated formatter.
                 usedParams.update(set(getInfo.assemblerParams))
 
                 component = getInfo.component
+
+                if component is None:
+                    raise RuntimeError(f"Internal error in datastore assembly of {ref}")
```
kfindeisen, line 973: A bit vague; are there a lot of ways to get a missing component?

python/lsst/daf/butler/core/location.py
```diff
@@ -87,7 +97,7 @@ def posix2os(posix):
 
     # Trailing "/" is stripped so we need to add back an empty path
     # for consistency
-    if posix.endswith(posixpath.sep):
+    if str(posix).endswith(posixpath.sep):
```
kfindeisen, line 100: Given that this code worked until now, is it necessary that `posix` be allowed to be a `PurePath`? If so, then the documentation needs to be updated.

python/lsst/daf/butler/core/location.py
```diff
@@ -270,31 +281,31 @@ def updateFile(self, newfile):
...
-        else:
-            pathclass = os.path
+        pathclass = posixpath if self.scheme else os.path
 
-        dir, _ = pathclass.split(self.path)
-        newpath = pathclass.join(dir, newfile)
+        # Mypy can't work out that these specific modules support split
+        # and join
+        dir, _ = pathclass.split(self.path)  # type: ignore
+        newpath = pathclass.join(dir, newfile)  # type: ignore
```
kfindeisen, line 289: Can you explicitly declare the types of `dir` and `newpath` for later code?

python/lsst/daf/butler/core/location.py
```diff
@@ -447,33 +458,35 @@ def __init__(self, datastoreRootUri, path):
 
         self._datastoreRootUri = datastoreRootUri
 
+        pathModule: types.ModuleType
         if self._datastoreRootUri.scheme == "file":
             pathModule = os.path
         else:
             pathModule = posixpath
```
kfindeisen, line 465: Can you use a ternary expression, like with `pathclass` above?

python/lsst/daf/butler/core/location.py
```diff
@@ -516,7 +529,7 @@ def relativeToPathRoot(self):
         stripped = p.relative_to(p.root)
         return str(posix2os(stripped))
 
-    def updateExtension(self, ext):
+    def updateExtension(self, ext: Optional[str]) -> None:
```
kfindeisen, line 532: Add ", optional" to the docstring.

python/lsst/daf/butler/core/s3utils.py
```diff
@@ -32,7 +40,7 @@
 from .location import ButlerURI, Location
 
 
-def getS3Client():
+def getS3Client() -> boto3.client:
```
kfindeisen, line 43: Does this work even if `boto3` can't be imported?

kfindeisen (approved):
I'm a bit concerned about how the YAML update was handled, since I think it will cause maintenance problems later. I'm fine with that being refactored later, however. Otherwise mostly style comments.

tests/test_cliPluginLoader.py
timj: I'm hoping that @n8pease fixes this properly at some point but I needed to fix it here otherwise I can't run tests on my local machine (the number of cores I have means that the test partitioning was triggering a problem whereby the left over state not being reset meant that real tests in command lines failed later. It's probably right that the tearDown is the one that makes the difference.

python/lsst/daf/butler/core/config.py
timj: Fundamentally it was a surprise to me to learn that yaml.load works fine on a string. I was about to start fiddling with IO.StringIO but then realized that wasn't necessary. `yaml.load` calls the first argument `stream` so I was also using stream. It's not documented what `stream` really means in the yaml docs.

python/lsst/daf/butler/core/config.py
timj: I don't think a second class is possible is it? This is a YAML plugin to handle `!include`.

Also, "stream" is what YAML calls it so I pretty much have to call it that.

python/lsst/daf/butler/core/config.py
timj: There's no ticket for fixing this since nothing has tickled this error yet. The problem is mostly associated with configs in S3 which need to then use the URI root to find the include file (which will be relative path). If the `!include` is absolute or you assume local root is `"."` then we could try to continue.  I haven't thought too deeply about the ramifications of this.

python/lsst/daf/butler/core/config.py
kfindeisen: If you mean the fact that `Loader` has `yaml.CSafeLoader` as a base class, it certainly should not constrain your parameter names, and I don't think it would constrain whether `Loader` itself has subclasses -- especially if you're relying on factory methods everywhere.

timj: I think that this argument being passed directly to the YAML loader suggests that `stream` is the right name.

python/lsst/daf/butler/core/config.py
timj: I've rearranged the code in a new commit. Now I use a relative URI up front if no other context exists and I fixed a but that meant that S3 absolute URIs in `!include` were not working properly (so you couldn't have a local YAML that referred to a remote YAML). Please take another look and see if this makes you happier.

PR #293, opened 2020-05-28
DM-25030 make an import butler subcommand

n8pease:


python/lsst/daf/butler/cli/butler.py
```diff
@@ -44,19 +44,38 @@ def _initLogging(logLevel):
...
+    # in cli/cmd/commands.py, and if those names are changed they must be
+    # changed here as well.
+    # It is expected that there will be very few butler command names that need
+    # to be changed because of e.g. conflicts with python keywords (as is done
+    # here and in cmdNameToFuncName for the 'import' command). If this becomes
+    # a common need then some way of doing this should be invented that is
+    # better than hard coding the function names into these conversion
+    # functions. An extension of the 'cli/resources.yaml' file (as is currently
+    # used in obs_base) might be a good way to do it.
+    if functionName == "butler_import":
```
timj, line 61: One other option that may be less prone to scaling problems could be to define a dict that maps from "import" to "butler_import" and then automatically generate the inverse dict. Then in these two functions you do the equivalent of:
```python
functionName = funcToCommand.get(functionName, functionName)
```
(I understand that my dict name there is non-optimal)

python/lsst/daf/butler/cli/opt/directory.py
```diff
@@ -0,0 +1,51 @@
...
+
+from ..utils import addArgumentHelp
+
+
+class directory_argument:  # noqa: N801
+    """Decorator to add a directory argument to a click command.
+
+    Parameters
+    ----------
+    required : bool, optional
```
timj: Backticks around the `bool` are needed for numpydoc.

python/lsst/daf/butler/script/butlerImport.py
```diff
@@ -0,0 +1,55 @@
...
+        in the current working directory, and `directory` is not `None`, it is
+        assumed to be in `directory`.  Defaults to "export.{format}".
+    output_run : `str`
+        The path to the location, the run, where datasets should be put.
+    transfer : `str`
+            The external data transfer type.
+
+    Returns
+    -------
+    butler : Butler
```
timj: backticks on Butler.

python/lsst/daf/butler/script/butlerImport.py
```diff
@@ -0,0 +1,55 @@
...
+    """Import data into a butler repository.
+
+    Parameters
+    ----------
+    repo : `str`
+        URI to the location of the repo.
+    directory : `str`, or None
+        Directory containing dataset files.  If `None`, all file paths must be
+        absolute.
+    export_file : `str`, optional
```
timj: This doesn't look like an optional parameter in the sense that you don't need to specify it. You do need to specify it but it can be None.

python/lsst/daf/butler/script/butlerImport.py
```diff
@@ -0,0 +1,55 @@
...
+    -------
+    butler : Butler
+        The butler intance, which can be used for test and verification.
+    """
+    butler = Butler(repo, run=output_run, writeable=True)
+    butler.import_(directory=directory,
+                   filename=export_file,
+                   transfer=transfer,
+                   format="yaml")
+    return butler
```
timj: I don't see anything using this return parameter. Why does it exist?

tests/test_cliCmdImport.py
```diff
@@ -0,0 +1,129 @@
...
+        self.run_test(["import", "here",
+                       "foo",
+                       "--output-run", "out"], expected)
+
+    def test_almostAll(self):
+        """Test all the parameters, except export_file which gets its own test
+        case below.
+        """
+        with self.runner.isolated_filesystem():
+            f = open("output.yaml", "w")
```
timj: What is the purpose of opening this file? It doesn't seem to be used anywhere.

python/lsst/daf/butler/script/butlerImport.py
```diff
@@ -0,0 +1,55 @@
...
+from .. import Butler
+
+
+def butlerImport(repo, directory, export_file, output_run, transfer):
+    """Import data into a butler repository.
+
+    Parameters
+    ----------
+    repo : `str`
+        URI to the location of the repo.
```
timj: URI to the location of the repo or URI to a config file describing the repo and its location.

tests/test_cliPluginLoader.py
```diff
@@ -84,31 +84,33 @@ def test_loadAndExecutePluginCommand(self):
         runner = click.testing.CliRunner()
         with command_test_env(runner):
             result = runner.invoke(butler.cli, "command-test")
-            self.assertEqual(result.exit_code, 0, result.output)
+            self.assertEqual(result.exit_code, 0, f"output: {result.output} exception: {result.exception}")
```
timj, line 84: It seems like you have fixed the commands caching so I think that means you can remove the setUp above.

timj (approved):
Looks okay.

python/lsst/daf/butler/cli/butler.py
n8pease: I thought about making some kind of data structure like that, but if I'm going to do even that much work I wanted to take a moment to figure out the best way to solve the problem. I think it might involve the plugin system, but I didn't want to spent that much time on it or add complexity, yet.

python/lsst/daf/butler/script/butlerImport.py
n8pease: leftover from trying to fix a unit test issue that ended up not being an issue. Not needed, will remove.

tests/test_cliCmdImport.py
n8pease: also leftover from addressing a unit test issue, the fix ended up becoming ExportFileCase. Will remove.

PR #294, opened 2020-05-29
DM-25153: Remove _ components from class path name when determining full type name

timj:


python/lsst/daf/butler/core/utils.py
```diff
@@ -131,15 +131,37 @@ def getFullTypeName(cls: Any) -> str:
...
+    if real_name != cleaned_name:
+        try:
+            test = doImport(cleaned_name)
+        except Exception:
+            # Could not import anything so return the real name
+            return real_name
+
+        # The thing we imported should match the class we started with
+        # despite the clean up. If it does not we return the real name
+        if test != cls:
```
TallJimbo: I think this can safely be `is not` rather than `!=`.

TallJimbo (approved):
Looks good!  Makes me really want to go rename `daf.butler.core` -> `daf.butler._core` (and probably more), but I'll resist that urge, at least until I have a real reason to.

TallJimbo (approved):
New changes look good!

PR #297, opened 2020-06-01
DM-25028 change butler create --config-file to --seed-config & DM-25151 in butler CLI add -h as an alias for --help

n8pease:


python/lsst/daf/butler/cli/butler.py
```diff
@@ -253,7 +253,7 @@ def get_command(self, context, name):
         return doImport(commands[name][0] + "." + cmdNameToFuncName(name))
 
 
-@click.command(cls=LoaderCLI)
+@click.command(cls=LoaderCLI, context_settings=dict(help_option_names=['-h', '--help']))
```
timj: General comment but in daf_butler we always prefer double quotes to single quotes.

python/lsst/daf/butler/cli/cmd/commands.py
```diff
@@ -50,7 +49,7 @@ def butler_import(*args, **kwargs):
 
 @click.command()
 @repo_argument(help=repo_argument.will_create_repo)
-@config_file_option(help="Path to an existing YAML config file to apply (on top of defaults).")
+@click.option("--seed-config", help="Path to an existing YAML config file to apply (on top of defaults).")
 @click.option("--standalone", is_flag=True, help="Include all defaults in the config file in the repo, "
               "insulating the repo from changes in package defaults.")
 @click.option("--override", "-o", is_flag=True, help="Allow values in the supplied config to override any "
```
timj: We probably want to remove the `-o` here since it's a very rare option and it's likely that we'll be using `-o` elsewhere.

timj: Also change text to "...supplied seed config to override all repo defaults"

python/lsst/daf/butler/tests/mockeredTest.py
```diff
@@ -0,0 +1,104 @@
...
+import click
+import click.testing
+import copy
+import unittest
+
+from ..cli.utils import clickResultMsg, mockEnvVar, Mocker
+from ..cli import butler
+
+
+class MockeredTestBase(unittest.TestCase, abc.ABC):
```
timj, line 32: "mockered"?

timj (approved):
Nice refactoring of the tests.

python/lsst/daf/butler/tests/mockeredTest.py
n8pease: at the risk of being too cute, I have a class called Mocker that wraps a static instance of MagicMock. This test uses a Mocker and so is Mockered.

python/lsst/daf/butler/cli/butler.py
n8pease: yeah. This was copy/paste and I didn't notice the quotes, will fix.

PR #298, opened 2020-06-02
DM-25034 make an install-external-data butler subcommand

n8pease:
help looks like this:

```
$ butler install-external-data --help
Usage: butler install-external-data [OPTIONS] REPO

  Install extenral data

  REPO is the URI or path to an existing data repository root or
  configuration file.

Options:
  --source TEXT            Source of external data
  --tract INTEGER          Tract identifier
  --visit-ccd INT INT ...  Visit and CCD of jointcal data to ingest.
  -h, --help               Show this message and exit.
```

### DIFF
--- a/python/lsst/daf/butler/script/installExternalData.py
+++ b/python/lsst/daf/butler/script/installExternalData.py
@@ -1,0 +1,78 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


TallJimbo:
Is this what we were calling `butler import` before?  This name feels like the name of a very ci_hsc-specific script.

n8pease:
@TallJimbo this replaces ci_hsc_gen2/bin.src/installExternalData.py I thought we wanted it turned into a butler subcommand?

TallJimbo:
That's actually a Gen2 butler operation.  I think the equivalent in Gen3 is just an invocation of `butler import`.

python/lsst/daf/butler/script/installExternalData.py
```diff
@@ -0,0 +1,78 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+import os
+from lsst.daf.persistence import Butler
```
timj, line 24: We are shutting down this PR but the red flag here is that daf_butler can not depend on daf_persistence for anything.

PR #301, opened 2020-06-02
DM-25178: Support write parameters in formatters

timj:


python/lsst/daf/butler/core/formatter.py
```diff
@@ -58,22 +59,34 @@ class Formatter(metaclass=ABCMeta):
         Identifies the file to read or write, and the associated storage
         classes and parameter information.  Its value can be `None` if the
         caller will never call `Formatter.read` or `Formatter.write`.
+    dataId : `DataCoordinate`, optional
+        Data ID associated with this formatter.
```
TallJimbo, line 63: I realize only the docs are new, but I'm curious why `Formatter` wants a data ID.

python/lsst/daf/butler/core/mappingFactory.py
```diff
@@ -189,9 +190,14 @@ def getFromRegistryWithMatch(self, targetClasses: Iterable[Any], *args: Any,
...
         # Supplied keyword args must overwrite registry defaults
-        registry_kwargs.update(kwargs)
-
-        return key, cls(*args, **registry_kwargs)
+        # We want this overwriting to happen recursively since we expect
+        # some of these keyword arguments to be dicts.
+        # Simplest to use Config for this
+        config_kwargs = Config(registry_kwargs)
+        config_kwargs.update(kwargs)
+        merged_kwargs = dict(config_kwargs)
```
TallJimbo, line 198: This can't actually convert the inner mappings back to `dict` as well.  Does that just not matter?

config/datastores/formatters.yaml
```diff
@@ -10,11 +10,6 @@ PhotoCalib: lsst.obs.base.fitsGenericFormatter.FitsGenericFormatter
 TransmissionCurve: lsst.obs.base.fitsGenericFormatter.FitsGenericFormatter
 Camera: lsst.obs.base.fitsGenericFormatter.FitsGenericFormatter
 Detector: lsst.obs.base.fitsGenericFormatter.FitsGenericFormatter
-ImageF: lsst.obs.base.fitsGenericFormatter.FitsGenericFormatter
-MaskX: lsst.obs.base.fitsGenericFormatter.FitsGenericFormatter
-Exposure: lsst.obs.base.fitsExposureFormatter.FitsExposureFormatter
-ExposureF: lsst.obs.base.fitsExposureFormatter.FitsExposureFormatter
-ExposureI: lsst.obs.base.fitsExposureFormatter.FitsExposureFormatter
```
TallJimbo, line 17: These get picked up because of StorageClass inheritance, I assume?

python/lsst/daf/butler/core/formatter.py
```diff
@@ -72,10 +72,11 @@ class Formatter(metaclass=ABCMeta):
     are supported. These param (`frozenset`).
     """
 
+    # Should we use ... to say that the parameters are unconstrained?
```
TallJimbo: FWIW, mypy hates `...` and I had to work pretty hard to make it work with the usage of `...` in Registry queries; I wouldn't use it unless it's very nice as a user-facing thing (GvR recommends enums as sentinals instead).

python/lsst/daf/butler/core/formatter.py
```diff
@@ -73,7 +73,7 @@ class Formatter(metaclass=ABCMeta):
     """
 
     # Should we use ... to say that the parameters are unconstrained?
-    supportedWriteParameters: ClassVar[Optional[Union[FrozenSet[str], Set[str]]]] = frozenset()
+    supportedWriteParameters: ClassVar[Optional[Union[FrozenSet[str], Set[str]]]] = None
```
TallJimbo: Might want to just use `MutableSet[str]` rather than the `Union` for this type.

TallJimbo (approved):
Looks good; thanks for the bite-size commits.

Comments are mostly questions rather than concerns.

python/lsst/daf/butler/core/formatter.py
timj: Because DECam raw files have no other way of knowing which detector you actually wanted the formatter to read.

python/lsst/daf/butler/core/mappingFactory.py
timj: I don't think it matters because everything lower down just gets passed directly anyhow and `Config` behaves like a dict. It might be that that final line isn't really needed either.

config/datastores/formatters.yaml
timj: They are duplicate entries in the YAML that were just confusing things. Copy and paste error I think. They weren't doing anything useful.

python/lsst/daf/butler/core/formatter.py
timj: FrozenSet isn't mutable though???

python/lsst/daf/butler/core/formatter.py
timj: I did see that recent sphinx fix that showed me that type and that is what gave me pause.

python/lsst/daf/butler/core/formatter.py
TallJimbo: Oh, right.  It would need to be `AbstractSet[str]`.  But that's probably what you want - mypy might complain later when you try to use mutators, but it would also complain similarly about the union.

PR #303, opened 2020-06-04
DM-25236: Remove quantum tables and columns from Registry schema.

TallJimbo:
These will be added back after further thought on DM-24613, but removing what we know to be incorrect should make those changes more likely to be backwards compatible and any migrations a bit easier.


andy-slac (approved):
Looks good.

PR #304, opened 2020-06-04
DM-25196 make query-collections and query-dataset-types butler commands

n8pease:


n8pease:
I've asked in Slack if `query-dataset-types --components` option is useful without `expression`. I'll drop the commit "remove components arg from queryDatasetTypes" if Jim (or anyone) says that parameter is still useful.

python/lsst/daf/butler/cli/opt/collection_type.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+
+    defaultHelp = "If provided, only list collections of this type."
+
+    choices = ["CHAINED", "RUN", "TAGGED"]
+
+    def makeCollectionType(self, context, param, value):
+        if value is None:
+            return value
+        value = value.upper()
```
timj, line 41: Click doesn't normalize case on choice options if you say `case_sensitive=False` ? Does click do minimum matching so that "chain" would work as well as "chained"?

python/lsst/daf/butler/script/queryCollections.py
```diff
@@ -0,0 +1,53 @@
...
+        If `True` (`False` is default), recursively yield the child collections
+        of matching `~CollectionType.CHAINED` collections.
+    include_chains : `bool` or `None`
+        If `True`, yield records for matching `~CollectionType.CHAINED`
+        collections.  Default is the opposite of ``flattenChains``: include
+        either CHAINED collections or their children, but not both.
+
+    Returns
+    -------
+    collections : [`str`, [`str`]]
```
timj: I'm trying to parse this return type indicator. Is that how we show that a dict is returned?

python/lsst/daf/butler/script/queryDatasetTypes.py
```diff
@@ -0,0 +1,30 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from .. import Butler
+
+
+def queryDatasetTypes(repo):
+    """Get the dataset types in a repository.
```
timj, line 47: Please add Returns section.

python/lsst/daf/butler/script/queryDatasetTypes.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    butler = Butler(repo)
+    datasetTypes = butler.registry.queryDatasetTypes()
+    return {'dataset-types': [datasetType.name for datasetType in datasetTypes]}
```
timj: I think I prefer "datasetTypes" as the key.

tests/test_cliCmdQueryCollections.py
```diff
@@ -0,0 +1,72 @@
...
+
+class QueryCollectionsScriptTest(unittest.TestCase):
+
+    def testGetCollections(self):
+        run = "ingest/run"
+        tag = "ingest"
+        expected = {"collections": ["ingest/run", "ingest"]}
+        runner = click.testing.CliRunner()
+        with runner.isolated_filesystem():
+            _ = Butler.makeRepo("here")
```
timj: Why are you assigning anyything at all to the return value?

The more common idiom though is for the return value from `Butler.makeRepo` and pass it to the first argument of the Butler constructor in the next line.

tests/test_cliCmdQueryCollections.py
```diff
@@ -0,0 +1,72 @@
...
+            _ = Butler("here", run=run, tags=[tag], collections=[tag])
```
timj: Can you add a comment here to say that the purpose of this call is to create some collections?

tests/test_cliCmdQueryCollections.py
```diff
@@ -0,0 +1,72 @@
...
+                                        flatten_chains=True,
+                                        include_chains=True))
+
+
+class QueryCollectionsScriptTest(unittest.TestCase):
+
+    def testGetCollections(self):
+        run = "ingest/run"
+        tag = "ingest"
+        expected = {"collections": ["ingest/run", "ingest"]}
```
timj: Don't repeat yourself here. use `[run, tag]` here

tests/test_cliCmdQueryDatasetTypes.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class QueryDatasetTypesScriptTest(unittest.TestCase):
+
+    def testQueryDatasetTypes(self):
+        expected = {"dataset-types": ["test"]}
+        runner = click.testing.CliRunner()
+        with runner.isolated_filesystem():
+            Butler.makeRepo("here")
+            butler = Butler("here", writeable=True)
```
timj: Again I have a slight preference for the return value of makeRepo being used here.

timj (approved):
Looks good. Some minor comments only.

PR #306, opened 2020-06-05
DM-21212: Update existing cp_pipe tasks to pipelineTasks

czwa:
Add StructuredDataDict formatter.


timj:
The reason I didn't make that a default formatter definition was that I was a bit concerned that we'd end up with a load of butler datasets that are random disorganized content (like pickle files) where the only validation is "looks like a dict".

czwa:
I noticed that the tests have explicit StructuredDataDict Yaml/Json/Pickle storageClasses with appropriate formatters.  I think I've been working under the idea that those would exist publicly, although I have no idea where I got that idea.  Being able to write a multi-level nested dataset is helpful in cp_pipe, especially one that needs no effort to read.

timj:
I'm more than happy for you to make the case that dumping native python dicts into YAML datasets is a useful thing to do.

TallJimbo (approved):

PR #309, opened 2020-06-11
DM-25327: Add file extension validation on ingest

timj:
Ostensibly this PR exists to support the new `PackagesFormatter` which I wanted to demonstrate could work by using the file format as a write parameter rather than embedding the file format in the formatter itself. This required that I change how Formatter.makeUpdatedLocation works and led to me realizing that the predict path location was not the right thing to do on ingest. We were simply assuming that the thing being ingested (with whatever extension it had) could always be handled by the gen3 formatter and we weren't checking. This means that in theory a .pickle file could be ingested and renamed to .yaml file.  Now we do the rename but copy across the file extension from the ingested file and complain if the ingested file has an unsupported extension.

### DIFF
--- a/python/lsst/daf/butler/core/formatter.py
+++ b/python/lsst/daf/butler/core/formatter.py
@@ -78,12 +78,13 @@ class Formatter(metaclass=ABCMeta):
     supported."""
 
     supportedExtensions: ClassVar[AbstractSet[str]] = frozenset()
-    """Set of all extensions supported by this formatter.  Only expected
-    to be populated by Formatters that write files. Any extension assigned
-    to the ``extension`` property will be automatically included in the
-    list of supported extensions."""
+    """Set of all extensions supported by this formatter.
 
-    def __init__(self, fileDescriptor: FileDescriptor, dataId: DataCoordinate = None,
+    Only expected to be populated by Formatters that write files. Any extension
+    assigned to the ``extension`` property will be automatically included in
+    the list of supported extensions."""
+
+    def __init__(self, fileDescriptor: FileDescriptor, dataId: Optional[DataCoordinate] = None,
                  writeParameters: Optional[Dict[str, Any]] = None,
                  writeRecipes: Optional[Dict[str, Any]] = None):
         if not isinstance(fileDescriptor, FileDescriptor):
--- a/python/lsst/daf/butler/core/formatter.py
+++ b/python/lsst/daf/butler/core/formatter.py
@@ -77,6 +77,12 @@ class Formatter(metaclass=ABCMeta):
     how a dataset is serialized. `None` indicates that no parameters are
     supported."""
 
+    supportedExtensions: ClassVar[AbstractSet[str]] = frozenset()
+    """Set of all extensions supported by this formatter.  Only expected
+    to be populated by Formatters that write files. Any extension assigned
+    to the ``extension`` property will be automatically included in the
+    list of supported extensions."""
+
     def __init__(self, fileDescriptor: FileDescriptor, dataId: DataCoordinate = None,
                  writeParameters: Optional[Dict[str, Any]] = None,
                  writeRecipes: Optional[Dict[str, Any]] = None):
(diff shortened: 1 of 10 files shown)


python/lsst/daf/butler/core/formatter.py
```diff
@@ -77,6 +77,12 @@ class Formatter(metaclass=ABCMeta):
     how a dataset is serialized. `None` indicates that no parameters are
     supported."""
 
+    supportedExtensions: ClassVar[AbstractSet[str]] = frozenset()
+    """Set of all extensions supported by this formatter.  Only expected
+    to be populated by Formatters that write files. Any extension assigned
+    to the ``extension`` property will be automatically included in the
+    list of supported extensions."""
```
TallJimbo: Should probably split this docstring into short + long descriptions.  I think right now it'll all end up in the short description.

python/lsst/daf/butler/core/formatter.py
```diff
@@ -77,6 +77,12 @@ class Formatter(metaclass=ABCMeta):
...
+
     def __init__(self, fileDescriptor: FileDescriptor, dataId: DataCoordinate = None,
```
TallJimbo: Should probably be `Optional[DataCoordinate]`.

python/lsst/daf/butler/datastores/posixDatastore.py
```diff
@@ -263,10 +263,13 @@ def _extractIngestInfo(self, path: str, ref: DatasetRef, *,
...
-            location = self.locationFactory.fromPath(template.format(ref))
-            newPath = formatter.predictPathFromLocation(location)
-            newFullPath = os.path.join(self.root, newPath)
+            # Work out the name we want this ingested file to have
+            # inside the datastore
+            location = self._calculate_ingested_datastore_name(ButlerURI(fullPath), ref, formatter)
+
+            newPath = location.pathInStore
+            newFullPath = location.path
+            print(newFullPath)
```
TallJimbo: Debug relic.

python/lsst/daf/butler/tests/testFormatters.py
```diff
@@ -62,3 +67,13 @@ def validateWriteRecipes(recipes: Optional[Mapping[str, Any]]) -> Optional[Mappi
...
+
+
+class LenientYamlFormatter(YamlFormatter):
+    """A test formatter that allows any file extension but always reads and
+    writes YAML."""
+    extension = ".yaml"
+
+    @classmethod
+    def validateExtension(cls, location: Location) -> None:
+        return
```
TallJimbo, line 79: How hard would it be to fix this test properly?  I think we've got a lot of tests that use invalid or unrealistic scenarios (though most of the problems are dimension combinations), and while I'm guilty of similar workarounds (i.e. the fact that `DatasetRef` has a `conform=False`) argument, I think this just adds technical debt.

TallJimbo (approved):
Most comments are little style things or obvious accidents.  I'm not a big fan of the test workaround, but if it's the pragmatic choice, fine.

python/lsst/daf/butler/tests/testFormatters.py
timj: I spent an hour on it and my main problem was trying to work out whether I deliberately used a mix of YamlFormatter and JsonFormatter or if they just happened to be the remnants from one of the other tests and we never noticed before because the format was irrelevant. This triggers solely in the chained datastore testing of constraints on ingest where there is a specialist datastore that overrides a formatter based on the dimensions rather than storage class.  I will take another look at it but I don't think I consider this to be technical debt as such -- it's a side effect of the complex testing options set up for one datastore. If I've dealt with everything else and this is my only remaining issue and I'm able to get it in the weekly today I'm inclined to merge it.

TallJimbo: Having already spent an hour looking for an easy-enough better solution is definitely enough due diligence for me.

PR #310, opened 2020-06-11
Add initial implementation of attribute manager (DM-25354)

andy-slac:
The interface for the manager defines few trivial operations, additional
methods will be added if needed as we discover new use cases. An
important part of this ticket is to define schema for metadata table
that should stay stable for long time. For this reason I decided to use a
simplest possible key-value storage with both key and value being
strings, reasonably long key (1kB) and value (64kB). I think this should
be sufficient for our current use cases.

### DIFF
--- a/python/lsst/daf/butler/registry/attributes.py
+++ b/python/lsst/daf/butler/registry/attributes.py
@@ -1,0 +1,114 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+"""The default concrete implementation of the class that manages
+attributes for `Registry`.
+"""
+
+__all__ = ["DefaultButlerAttributeManager"]
+
+from typing import (
+    ClassVar,
+    Iterable,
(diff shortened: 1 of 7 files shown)


python/lsst/daf/butler/registry/attributes.py
```diff
@@ -0,0 +1,114 @@
...
+    db : `Database`
+        Database engine interface for the namespace in which this table lives.
+    table : `sqlalchemy.schema.Table`
+        SQLAlchemy representation of the table that stores attributes.
+    """
+    def __init__(self, db: Database, table: sqlalchemy.schema.Table):
+        self._db = db
+        self._table = table
+
+    _META_TABLE_NAME: ClassVar[str] = "butler_attributes"
```
TallJimbo: In other managers I used "_META" as a prefix for things like this because they were the meta-tables that described which other tables existed.  I think I'd just use `_TABLE_NAME` and `_TABLE_SPEC` because there's only one table that's relevant for this manager.

python/lsst/daf/butler/registry/attributes.py
```diff
@@ -0,0 +1,114 @@
...
+                "value": value,
+            })
+        else:
+            try:
+                self._db.insert(self._table, {
+                    "name": name,
+                    "value": value,
+                })
+            except sqlalchemy.exc.IntegrityError as exc:
+                raise ButlerAttributeExistsError(f"attribute {name} already exists") from exc
```
TallJimbo, line 100: Maybe we won't be able to tell until we try to use this, but I suspect a version of this that uses `Database.sync` to implement "try to set k = v and raise if it is set to something else" might be more useful than the `insert` version, which will raise if k exists even if it's already set to the same value.

TallJimbo (approved):
Looks good!

Could you also take a quick look at `test_oracle.py`, and update the direct call to the `Registry` constructor there to include the new argument?  (I haven't found a good way to make that test call `Registry.fromConfig`, which now does a lot of important work, so it tends to be fragile - especially because we don't have a way to run it in CI.)

python/lsst/daf/butler/registry/attributes.py
andy-slac: I missed that in renaming, will fix.

python/lsst/daf/butler/registry/attributes.py
andy-slac: The use case I was thinking about is "I know this attribute should not exist and I want to create it". I suspect that updating attributes will be a rare process under user control, so `force` option should cover that case.

andy-slac:
I updated `test_oracle.py` too, but I have no way to test it.

PR #311, opened 2020-06-11
DM-25317 cleanup items for butler CLI

n8pease:


python/lsst/daf/butler/cli/butler.py
```diff
@@ -227,7 +227,7 @@ def list_commands(self, ctx):
         commands = self._getCommands()
         self._raiseIfDuplicateCommands(commands)
         log.debug(commands.keys())
-        return commands.keys()
+        return sorted(commands.keys())
```
timj: I don't think you need the `.keys()` any more.

python/lsst/daf/butler/cli/cmd/commands.py
```diff
@@ -64,7 +64,7 @@ def create(*args, **kwargs):
     cli_handle_exception(createRepo, *args, **kwargs)
 
 
-@click.command()
+@click.command(short_help="Dump buter config to stdout.")
```
timj: buter -> butler

timj (commented):
Not a review but just a couple of comments I spotted immediately.

tests/test_cliCmdConfigValidate.py
```diff
@@ -25,16 +25,25 @@
...
 
-TESTDIR = os.path.abspath(os.path.dirname(__file__))
+class ValidateTest(CliCmdTestBase):
 
+    defaultExpected = dict()
 
-class Suite(unittest.TestCase):
+    command = config_validate
+
+    def test_help(self):
```
timj: This definitely looks like it can go in the base class. It's identical everywhere.

tests/test_cliPluginLoader.py
```diff
@@ -74,12 +75,39 @@ def duplicate_command_test_env(runner):
             yield
 
 
+class FailedLoadTest(unittest.TestCase):
+
+    def setUp(self):
+        # The caplog fixture is a pytest feature to manipulate and read log
+        # output from the test.
```
timj: Does this mean log output from the logging package? We don't use pytest fixtures for that, we use unittest:
```python
with self.assertLogs(level="INFO") as cm:
   ...
```

timj (approved):
I left some minor comments yesterday that still need to be fixed. My main additional comment is that I don't understand the caplog thing and the commit message doesn't help. Also the `test_help` method is copied everywhere and should seemingly be in a base class.

tests/test_cliPluginLoader.py
n8pease: I was having trouble capturing logging for testing when testing with pytest, because it captures logs which is why I was using the pytest fixture. But this solution works better, does not require special casing.

tests/test_cliCmdConfigValidate.py
n8pease: I was having trouble with putting test_ functions in the base class, which is abstract. But making the most-derived class derive from the command test base and from TestCase works well.

PR #315, opened 2020-06-13
DM-25416: Fix documentation build from renamed formatters

timj:
Also includes a GitHub Action to run the build.


ktlim (approved):
Is there a reason for the ordering?  (I presume it carries through to the document.)  If not, perhaps alphabetical is better?

PR #316, opened 2020-06-18
DM-25355: Add support for specifying schema versions

andy-slac:
Schema versions are now specified in the registry config file and they
are stored to attributes table on schema initialization. When opening
existing database these versions are compared with configured ones,
exception is raised if they are not compatible. There is also a
groundwork for detecting schema changes by checking schema digests, it
is not used by anything yet.

Couple of unit tests needed updates to avoid repeated initialization of
existing schema.

### DIFF
--- a/config/registry.yaml
+++ b/config/registry.yaml
@@ -12,3 +12,13 @@ registry:
     collections: lsst.daf.butler.registry.collections.synthIntKey.SynthIntKeyCollectionManager
     datasets: lsst.daf.butler.registry.datasets.byDimensions.ByDimensionsDatasetRecordStorageManager
     datastores: lsst.daf.butler.registry.bridge.monolithic.MonolithicDatastoreRegistryBridgeManager
+  schema_versions:
+    core:
+      version: 0.0.1
+      digest: ""
+    dimensions:
+      version: 0.0.1
+      digest: ""
+    datastores:
+      version: 0.0.1
+      digest: ""
--- a/config/registry.yaml
+++ b/config/registry.yaml
@@ -12,13 +12,3 @@ registry:
     collections: lsst.daf.butler.registry.collections.synthIntKey.SynthIntKeyCollectionManager
     datasets: lsst.daf.butler.registry.datasets.byDimensions.ByDimensionsDatasetRecordStorageManager
     datastores: lsst.daf.butler.registry.bridge.monolithic.MonolithicDatastoreRegistryBridgeManager
-  schema_versions:
-    core:
-      version: 0.0.1
-      digest: ""
-    dimensions:
-      version: 0.0.1
-      digest: ""
-    datastores:
-      version: 0.0.1
-      digest: ""
--- a/python/lsst/daf/butler/registry/_registry.py
+++ b/python/lsst/daf/butler/registry/_registry.py
@@ -139,7 +139,7 @@ class Registry:
         collections = doImport(config["managers", "collections"])
         datasets = doImport(config["managers", "datasets"])
         datastoreBridges = doImport(config["managers", "datastores"])
-        versions = ButlerVersionsManager.fromConfig(config["schema_versions"])
+        versions = ButlerVersionsManager.fromConfig(config.get("schema_versions"))
 
         return cls(database, universe, dimensions=dimensions, attributes=attributes, opaque=opaque,
                    collections=collections, datasets=datasets, datastoreBridges=datastoreBridges,
--- a/python/lsst/daf/butler/registry/_registry.py
+++ b/python/lsst/daf/butler/registry/_registry.py
@@ -70,6 +70,7 @@ from ._collectionType import CollectionType
 from ._exceptions import ConflictingDefinitionError, InconsistentDataIdError, OrphanedRecordError
 from .wildcards import CategorizedWildcard, CollectionQuery, CollectionSearch, Ellipsis
 from .interfaces import ChainedCollectionRecord, RunRecord
+from .versions import ButlerVersionsManager
 
 if TYPE_CHECKING:
     from ..butlerConfig import ButlerConfig
@@ -138,9 +139,11 @@ class Registry:
         collections = doImport(config["managers", "collections"])
         datasets = doImport(config["managers", "datasets"])
         datastoreBridges = doImport(config["managers", "datastores"])
+        versions = ButlerVersionsManager.fromConfig(config["schema_versions"])
+
         return cls(database, universe, dimensions=dimensions, attributes=attributes, opaque=opaque,
                    collections=collections, datasets=datasets, datastoreBridges=datastoreBridges,
-                   create=create)
+                   versions=versions, writeable=writeable, create=create)
 
     def __init__(self, database: Database, universe: DimensionUniverse, *,
                  attributes: Type[ButlerAttributeManager],
@@ -149,6 +152,8 @@ class Registry:
                  collections: Type[CollectionManager],
                  datasets: Type[DatasetRecordStorageManager],
                  datastoreBridges: Type[DatastoreRegistryBridgeManager],
+                 versions: ButlerVersionsManager,
+                 writeable: bool = True,
                  create: bool = False):
         self._db = database
         self.storageClasses = StorageClassFactory()
@@ -164,6 +169,15 @@ class Registry:
                                                                  opaque=self._opaque,
                                                                  datasets=datasets,
                                                                  universe=self.dimensions)
+            context.addInitializer(lambda db: versions.storeVersions(self._attributes))
+
+        # This call does not do anything right now as we do not have a way to
+        # split tables between sub-schemas yet.
+        versions.checkVersionDigests()
+        if not create:
+            # verify that configured versions are compatible with schema
+            versions.checkStoredVersions(self._attributes, writeable)
+
         self._collections.refresh()
         self._datasets.refresh(universe=self._dimensions.universe)
 
--- a/python/lsst/daf/butler/registry/interfaces/_database.py
+++ b/python/lsst/daf/butler/registry/interfaces/_database.py
@@ -31,6 +31,7 @@ from abc import ABC, abstractmethod
 from contextlib import contextmanager
 from typing import (
     Any,
+    Callable,
     Dict,
     Iterable,
     Iterator,
@@ -100,6 +101,7 @@ class StaticTablesContext:
         self._foreignKeys: List[Tuple[sqlalchemy.schema.Table, sqlalchemy.schema.ForeignKeyConstraint]] = []
         self._inspector = sqlalchemy.engine.reflection.Inspector(self._db._connection)
         self._tableNames = frozenset(self._inspector.get_table_names(schema=self._db.namespace))
+        self._initializers: List[Callable[[Database], None]] = []
 
     def addTable(self, name: str, spec: ddl.TableSpec) -> sqlalchemy.schema.Table:
         """Add a new table to the schema, returning its sqlalchemy
@@ -140,6 +142,20 @@ class StaticTablesContext:
         """
         return specs._make(self.addTable(name, spec)                     # type: ignore
                            for name, spec in zip(specs._fields, specs))  # type: ignore
+
+    def addInitializer(self, initializer: Callable[[Database], None]) -> None:
+        """Add a method that does one-time initialization of a database.
+
+        Initialization can mean anything that changes state of a database
+        and needs to be done exactly once after database schema was created.
+        An example for that could be population of schema attributes.
+
+        Parameters
+        ----------
+        initializer : callable
+            Method of a single argument which is a `Database` instance.
+        """
+        self._initializers.append(initializer)
 
 
 class Database(ABC):
@@ -387,7 +403,12 @@ class Database(ABC):
                 with warnings.catch_warnings():
                     warnings.simplefilter("ignore", category=sqlalchemy.exc.SADeprecationWarning)
                     self._metadata.create_all(self._connection)
+                # call all initializer methods sequentially
+                for init in context._initializers:
+                    init(self)
         except BaseException:
+            # TODO: this is potentially dangerous if we run it on
+            # pre-existing schema.
             self._metadata.drop_all(self._connection)
             self._metadata = None
             raise
--- a/python/lsst/daf/butler/registry/versions.py
+++ b/python/lsst/daf/butler/registry/versions.py
@@ -148,12 +148,12 @@ class ButlerVersionsManager:
         self._tablesGroups: MutableMapping[str, List[sqlalchemy.schema.Table]] = {}
 
     @classmethod
-    def fromConfig(cls, schemaVersionConfig: Config) -> ButlerVersionsManager:
+    def fromConfig(cls, schemaVersionConfig: Optional[Config]) -> ButlerVersionsManager:
         """Make `ButlerVersionsManager` instance based on configuration.
 
         Parameters
         ----------
-        schemaVersionConfig : `Config`
+        schemaVersionConfig : `Config` or `None`
             Configuration object describing schema versions, typically
             "schema_versions" sub-object of registry configuration.
 
@@ -163,10 +163,11 @@ class ButlerVersionsManager:
             New instance of the versions manager.
         """
         versions = {}
-        for key, vdict in schemaVersionConfig.items():
-            version = VersionTuple.fromString(vdict["version"])
-            digest = vdict.get("digest")
-            versions[key] = VersionInfo(version, digest)
+        if schemaVersionConfig:
+            for key, vdict in schemaVersionConfig.items():
+                version = VersionTuple.fromString(vdict["version"])
+                digest = vdict.get("digest")
+                versions[key] = VersionInfo(version, digest)
         return cls(versions)
 
     @staticmethod
--- a/python/lsst/daf/butler/registry/versions.py
+++ b/python/lsst/daf/butler/registry/versions.py
@@ -1,0 +1,325 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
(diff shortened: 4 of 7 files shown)


config/registry.yaml
```diff
@@ -12,3 +12,13 @@ registry:
     collections: lsst.daf.butler.registry.collections.synthIntKey.SynthIntKeyCollectionManager
     datasets: lsst.daf.butler.registry.datasets.byDimensions.ByDimensionsDatasetRecordStorageManager
     datastores: lsst.daf.butler.registry.bridge.monolithic.MonolithicDatastoreRegistryBridgeManager
+  schema_versions:
+    core:
+      version: 0.0.1
+      digest: ""
+    dimensions:
+      version: 0.0.1
```
timj: Does this have to match the version number in dimensions.yaml?

andy-slac: This is the question that I forgot to ask with all other complications. I am not sure what how these versions relate to each other but we should try to un-confuse things here. I know that dimensions.yaml version is used internally and we can probably have multiple versions (and I think it's also a single integer number). 

 @TallJimbo, what are your thoughts on how that version is related to schema version? Do we actually need that one?

config/registry.yaml
```diff
@@ -12,3 +12,13 @@ registry:
...
+      digest: ""
+    datastores:
```
timj: Another quick point, but in theory the schema for datastores can depend on the datastore. s3datastore and posixDatastore do use the same schema for their internal table but that's not required (inMemoryDatastore doesn't use a table at all).

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -141,6 +143,20 @@ def addTableTuple(self, specs: Tuple[ddl.TableSpec, ...]) -> Tuple[sqlalchemy.sc
         return specs._make(self.addTable(name, spec)                     # type: ignore
                            for name, spec in zip(specs._fields, specs))  # type: ignore
 
+    def addInitializer(self, initializer: Callable[[Database], None]) -> None:
+        """Add a method that does one-time initialization of a database.
+
+        Initialization can mean anything that changes state of a database
+        and needs to be done exactly once after datanbase schema was created.
```
TallJimbo: typo: `datanbase`

python/lsst/daf/butler/registry/versions.py
```diff
@@ -0,0 +1,321 @@
...
+
+
+class IncompatibleVersionError(RuntimeError):
+    """Exception raised when configured version number is not compatible with
+    database version.
+    """
+    pass
+
+
+class VersionTuple:
```
TallJimbo: Unless this needs to be mutable, it seems like it might make sense to actually use `typing.NamedTuple` (as the base class here).  That will get you `__init__` and what I think are probably the right comparison operators for free, and if you call the class "Tuple" it's reasonable for users to expect indexing, iteration, etc. to work (and that makes sense, given the analogy with e.g. `sys.version_info`).

python/lsst/daf/butler/registry/versions.py
```diff
@@ -0,0 +1,321 @@
...
+
+class VersionInfo:
+    """Representation of version information as defined by configuration.
+
+    Parameters
+    ----------
+    version : `VersionTuple`
+        Version number in parsed format.
+    digest : `str`, optional
+        Optional digest of the corresponding part of the shema definition.
```
TallJimbo: typo: `shema`

python/lsst/daf/butler/registry/versions.py
```diff
@@ -0,0 +1,321 @@
...
+        self.digest = digest
+
+
+class ButlerVersionsManager:
+    """Utility class to manage and verify schema version compatibility.
+
+    Parameters
+    ----------
+    versions : `dict` [`str`, `VersionInfo`]
+        Mapping of the group name to correspoding schema version and digest.
```
TallJimbo: typo: `correspoding`

python/lsst/daf/butler/registry/versions.py
```diff
@@ -0,0 +1,321 @@
...
+    @classmethod
+    def fromConfig(cls, schemaVersionConfig: Config) -> ButlerVersionsManager:
+        """Make `ButlerVersionsManager` instance based on configuration.
+
+        Parameters
+        ----------
+        schemaVersionConfig : `Config`
+            Configutation object describibg schema versions, typically
+            "schema_versions" sub-object of registry configuration.
+
```
TallJimbo, line 159: Missing `Returns` section.

python/lsst/daf/butler/registry/versions.py
```diff
@@ -0,0 +1,321 @@
...
+        self._tablesGroups: MutableMapping[str, List[sqlalchemy.schema.Table]] = {}
+
+    @classmethod
+    def fromConfig(cls, schemaVersionConfig: Config) -> ButlerVersionsManager:
+        """Make `ButlerVersionsManager` instance based on configuration.
+
+        Parameters
+        ----------
+        schemaVersionConfig : `Config`
+            Configutation object describibg schema versions, typically
```
TallJimbo: typos: `Configutation` and `describibg`

python/lsst/daf/butler/registry/versions.py
```diff
@@ -0,0 +1,321 @@
...
+
+        Returns
+        -------
+        digest : `str`
+            String representation of the digest of the schema.
+
+        Notes
+        -----
+        It is not specified what kind of implementation is used to calculate
+        digest string. The only requrement for that is that result should be
```
TallJimbo: typo: requrement

python/lsst/daf/butler/registry/versions.py
```diff
@@ -0,0 +1,321 @@
...
+            tableSchemaRepr = [table.name]
+            schemaReps = []
+            for column in table.columns:
+                columnRep = f"COL,{column.name},{column.type}"
+                if column.primary_key:
+                    columnRep += ",PK"
+                if column.nullable:
+                    columnRep += ",NULL"
+                schemaReps += [columnRep]
+            for fkConstr in table.foreign_key_constraints:
```
TallJimbo, line 239: Do FK constraints actually affect schema compatibility?  In most of the scenarios I can think of (where there is always a logical FK but we may or may not define an actual constraint for some reason), adding or removing the constraint seems like it'd just be a patch-level change.

python/lsst/daf/butler/registry/versions.py
```diff
@@ -0,0 +1,321 @@
...
+        for that group.
+
+        Parameters
+        ----------
+        group : `str`
+            Schema group name, e.g. "core", or " dimensions".
+        table : `sqlalchemy.schema.Table`
+            Table schema.
+        """
+        self._tablesGroups.setdefault(group, []). append(table)
```
TallJimbo: space after `.`

python/lsst/daf/butler/registry/versions.py
```diff
@@ -0,0 +1,321 @@
...
+            attributes.set(f"schema_versions.{key}.version", str(vInfo.version))
+            # TODO: we could also store digest in the database but I'm not
+            # sure that digest calculation is stable enough at this point.
+
+    def checkVersionDigests(self) -> None:
+        """Compare current schema digest to a configured digest.
+
+        It calculates digest to all schema groups using tables added to each
+        group with `addTable` method. If digest is different from a configured
+        digest for the same group it generates logging warning message.
```
TallJimbo, line 289: _Only_ logging a warning seems like a fairly limited way to report problems.  Might make sense to have this at least return a `bool` as well so it could be used by callers that might want to raise exceptions instead.

config/registry.yaml
TallJimbo: I think it'd probably be ideal if the version in `dimensions.yaml` could be the only one, and also serve as a schema version.  We could also probably find a way to pass a version defined in this part of the config into the dimensions system, but I'm guessing it would require adding it as an argument to a bunch of constructors.  I definitely think we should have only one version number for dimensions.

I would be quite happy to use the `VersionTuple` class introduced on this ticket for the version information stored in `DimensionUniverse`, if we can organize the package structure to make that dependency okay.

As a side thought, I know the dimensions system needs some work to make backwards compatibility easier; very small changes (like adding a new HTM level) right now will make a totally new universe that knows nothing about the old one, and since dimensions are used to define tables (e.g. `dataset_collection_XXXX`) for some managers, that's a real problem.  The (vague) ideas I have for fixing this include:
 - Start including the dimension version number into `DimensionGraph.encode` output so we can at least tell when a `dataset_collection_XXXX` table was defined with a different dimension version than we have.
 - Have `dimensions.yaml` include multiple versions (under different `version: ` headers) of the complete dimensions definition.  Until a repo is fully migrated away from an old version, any older dimensions definitions used in that repo must be present in configuration as well.  If necessary we could add some kind of inheritance/override stuff to configs to reduce the duplication this implies.
 - Provide a way to translate a `DimensionGraph` from one version to another, by looking up the `DimensionUniverse` for both versions.  This would need to assume that the translated dimensions are present in both and have the same relationships, but could otherwise differ.

That's probably mostly not relevant for this ticket (it's more about making more changes into patch or minor changes instead of forcing all changes to be major changes), but this discussion reminded me that I wanted to discuss these ideas with you, and it gives you an idea of what the version in `dimensions.yaml` might do.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -164,6 +169,15 @@ def __init__(self, database: Database, universe: DimensionUniverse, *,
                                                                  opaque=self._opaque,
                                                                  datasets=datasets,
                                                                  universe=self.dimensions)
+            context.addInitializer(lambda db: versions.storeVersions(self._attributes))
+
+        # This call does not do anything right now as we do not have a way to
+        # split tables between sub-schemas yet.
```
TallJimbo, line 175: When we do, I assume the goal will be to do the digest only on the "static" tables, not the "dynamic" tables whose presence and definition depends on e.g. dataset type?  That seems much easier to implement, even though there would be a range of schema changes the digest checks would not catch (i.e. those that affect only dynamic tables).

The catch is that we treat all datastores more like "dynamic" tables (they're created by calls to `self._opaque`, and even when making a new repo that won't happen by this point in the code), so that may make it hard to create digests for those.

config/registry.yaml
TallJimbo: If the way this would work in practice is that the single datastore schema version would need to be updated whenever any datastore implementation changes its schema, that might still be okay (at least if all datastores continue to be implemented in daf_butler, not as extensions in other packages - that worries me a bit, but maybe that's just a hypothetical concern). It might cause us to declare some repos as incompatible when they actually aren't, which is better than not recognizing an incompatibility, but it could still be a problem.

Of course, we have the exact same problem (I think) with the core schema version and all of the registry managers.  If we change one of those, I assume that means we change the core schema version - but that implies that repos that didn't use the changed manager now report as incompatible with the codebase, but really aren't.

TallJimbo (approved):
Comments are little typos or big-picture questions about how this might be used in later tickets.  I'll summarize the latter in Jira rather than here, but the code looks fine and is okay to merge unless you think the big-picture stuff should hold it up.

python/lsst/daf/butler/registry/versions.py
andy-slac: 👍 I thought about just doing `collections.namedtuple` as I did not really like much of `typing` stuff, but I can do it too. This was supposed to be internal class but it looks like it's going to be more exposed (or moved somewhere) so I should make it right.

python/lsst/daf/butler/registry/versions.py
andy-slac: Even it it is a patch-level change you still probably want to know that your schema is different from a configured version and configured version needs to be updated (together with digest). For me anything that changes schema, even in a trivial way, needs a new version number, so that we can use migration tools to identify those versions.

python/lsst/daf/butler/registry/versions.py
andy-slac: I'm not sure that simple `bool` would be useful for clients. The use case for generating warnings is that when you update you code and change schema you still do not know the digest for the new schema. To calculate new digest you would run the code and with give you a warning that is expected and a new digest that you can you to update your configuration. Returning False to client does not produce much useful info to decide what needs to be done, the code will likely run if your changes are OK, it's just the configuration does not match your code.

python/lsst/daf/butler/registry/_registry.py
andy-slac: Yes, dynamic tables cannot be in digest as they can be made much later in the lifetime of the schema. For those tables I think we can only rely on developers to increment version numbers without help from digest.

config/registry.yaml
andy-slac: I'm OK with using version from `dimensions.yaml` but I worry that we may be overloading that version number with different purposes. The schema version number should reflect the database side of things, and that is an implementation detail and it could change if you use different implementation of a manager class for example. This could mean that we have to switch to incompatible version number even if `dimensions.yaml` does not change. I'm not entirely sure but I think that existing version number has some different purpose which does not depend on implementation detail like database schema.

python/lsst/daf/butler/registry/versions.py
andy-slac: And I could potentially add a separate application that calculates digests and then we could require that digest should always be correct in the configuration but that maybe too disruptive during development process (though it is possible to disable digest check now).

config/registry.yaml
andy-slac: Hmm, it gets complicated. I think that schema is not the whole story, there is also data compatibility. Could s3datastore and posixDatastore actually share the same database? If they cannot then we should not use the same version number for their schema.

That question also applies to other potential cases, I can imagine that sometimes we may need to do a data migration without changing schema (e.g. changing the format/meaning of string/BLOB data). That may also be accompanied by the change of the manager class but not necessarily. That sort of change also needs to be reflected in the schema version. And I think that configuration that we are going to store in the new attributes table is another example where the data in a table has to be consistent with the schema version number because that configuration actually determines the schema.

timj: S3 and Posix datastores do currently share the same database schema but nothing requires that they continue to do so.  It is defined in fileLikeDatastore at the moment.

andy-slac: My question was more about data, suppose that Posix datastore had written something to that table, could S3 datastore read and interpret those records, or at least understand that they are from different datastore and have to be ignored?

timj: Each datastore has its own table. They don't share content. You could point an S3 datastore at the posix table and it would work so long as somebody behind the scenes had uploaded the data to S3 without telling butler but that wasn't something we had considered being a standard scenario.

timj: And each butler can have multiple datastores in play each with their own tables and you can only work out how many you have by asking the datastore itself.

config/registry.yaml
TallJimbo: It's a good point that the version in `dimensions.yaml` is not quite a schema version.  But I also think it's all we need (pending the question of versioning managers that's being discussed on Jira) - I think we can say that the combination of the `dimensions.yaml` version with versions that would be embedded in the daf_butler codebase (I'm imagining an overall daf_butler schema version + a dimension manager version, but that's not the only possibility) totally controls the dimensions part of the schema.

Or more precisely, there isn't a well-defined "dimensions part of the schema", but if we think about the full schema "version" being an unorderable hash of many different human-meaningful identifiers (see my last Jira comment), then the `dimensions.yaml` version number is still a sufficient number to capture any changes in that file/yaml-section, even if it has to be hashed together with other in-code versions and config labels to map to the full schema.

python/lsst/daf/butler/registry/versions.py
TallJimbo: I can understand not liking a lot of `typing`, but IMO `typing.NamedTuple` has much nicer than the original `collections.namedtuple`.  I hope you'll agree.

python/lsst/daf/butler/registry/versions.py
TallJimbo: If there's no real use case, that's okay.  I just found it a little weird that the only way an error was reported was through logs.

config/registry.yaml
andy-slac: If there can be multiple datastores and/or datastores can be added/removed dynamically then I think we cannot manage that kind of setup via a single version number. Each datastore probably needs its own separate schema version number and each one needs to me migrated separately. I think in datastores case we likely don't need to migrate database schema between different datastores, so they can be managed completely independently, each one of them would have a linear history. It would still be beneficial to have the same tool (alembic) to manage datastores versions, so those versions need to be identifiable like for other managers (via configuration or in the source code).

python/lsst/daf/butler/registry/versions.py
andy-slac: I do not consider it an error yet, just a reminder for a developer that schema version may have changed. Digest calculation is not bullet-proof right now, if we can make it more reliable in the future we can proabably switch to exceptions or returning error status from this method.

config/registry.yaml
andy-slac: One more thought on dimensions - encoding of the dimension graph into `dataset_collection_XXXX` table names makes it very fragile w.r.t. adding or removing of the dimensions to/from the universe. There may be (complicated) workaround for that by adding a separate registry/table for mapping a set of dimensions to table name (or just a serial table number). It will be append-only and can be cached efficiently but it does add some extra logic.

TallJimbo: I was already planning to follow up (on DM-24660, or maybe DM-24613 to do it right the first time for quantum tables) on your suggestion on a previous ticket review to start saving the full `dataset_collection_XXXX` table names into the `dataset_type` table, instead of relying on the encoding producing the same thing each time.  I think if we encode the dimension version number, too, that may be enough to avoid that kind of logic, but we'll see.

config/registry.yaml
timj: I think each datastore class (s3, posix etc) needs its own version number.

PR #317, opened 2020-06-19
Switch to using a NullPool in sql connection

natelust:


MichelleGower (approved):
Changes look fine.  Tested with 20 ci_hsc_gen3 runs without any of the previous SSL error messages.  Please merge.

andy-slac:
Just a side comment - not using pool should indeed fix the problem but it may also have performance implications depending on how frequently our clients create/destroy connections. Maybe to keep in mind for the future when we are ready to profile things.

PR #318, opened 2020-06-21
DM-25385 make a pipetask command based on click

n8pease:


python/lsst/daf/butler/cli/butler.py
```diff
@@ -19,86 +19,149 @@
...
-    if commandName == "import":
-        commandName = "butler_import"
-    return commandName.replace("-", "_")
-
-
-class LoaderCLI(click.MultiCommand):
-
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+class LoaderCLI(click.MultiCommand, abc.ABC):
```
andy-slac, line 39: I think it needs docstring, it's hard to guess what its purpose is.

python/lsst/daf/butler/cli/butler.py
```diff
@@ -19,86 +19,149 @@
...
-        if pluginModules:
-            return [p for p in pluginModules.split(":") if p != '']
         return []
 
+    @classmethod
+    def _funcNameToCmdName(cls, functionName):
+        """Convert function name to the butler command name: change
+        underscores, (used in functions) to dashes (used in commands), and
+        change local-package command names that conflict with python keywords
+        to a leagal function name.
```
andy-slac: typo: leagal

python/lsst/daf/butler/cli/butler.py
```diff
@@ -19,86 +19,149 @@
...
         """
-        pluginModules = os.environ.get("DAF_BUTLER_PLUGINS")
-        if pluginModules:
-            return [p for p in pluginModules.split(":") if p != '']
         return []
 
+    @classmethod
+    def _funcNameToCmdName(cls, functionName):
+        """Convert function name to the butler command name: change
+        underscores, (used in functions) to dashes (used in commands), and
```
andy-slac, line 143: No need for comma after "underscores"?

python/lsst/daf/butler/cli/butler.py
```diff
@@ -210,58 +260,66 @@ def _raiseIfDuplicateCommands(commands):
...
-        commands : `list` [`str`]
-            The names of the commands that can be called by the butler command.
+    localCmdPkg = "lsst.daf.butler.cli.cmd"
+
+    @classmethod
+    def _funcNameToCmdName(cls, functionName):
+        """Convert function name to the butler command name: change
+        underscores, (used in functions) to dashes (used in commands), and
+        change local-package command names that conflict with python keywords
+        to a leagal function name.
```
andy-slac: I don't think you need to repeat whole docstring for the method that you are overriding, just add a comment like this:
https://github.com/lsst/daf_butler/blob/master/python/lsst/daf/butler/datastores/chainedDatastore.py#L354

python/lsst/daf/butler/cli/opt/config.py
```diff
@@ -21,21 +21,49 @@
...
+    ----------
+    help : `str`, optional
+        The help text to use for the option, by default `defaultHelp`
+    multiple : `bool`, optional
+        If true, multiple instances of the option may be passed in on the
+        command line, by default False
+    required : bool, optional
+        If true, the option is required to be passed in on the command line, by
+        default False
+    split_kv_sep : `str` or None, optional
```
andy-slac: There is `split_kv` parameter but no `split_kv_sep`.

python/lsst/daf/butler/cli/opt/config.py
```diff
@@ -21,21 +21,49 @@
...
-        self.required = required
+    def __init__(self, help=defaultHelp, multiple=False, required=False, split_kv=False):
         self.help = help
+        self.multiple = multiple
+        self.required = required
+        if split_kv:
+            self.callback = partial(split_kv_func,
+                                    separator=split_kv_separator,
+                                    multiple=multiple)
+        elif multiple is True:
```
andy-slac: `elif multiple is True` -> `elif multiple`?

python/lsst/daf/butler/cli/opt/config_file.py
```diff
@@ -22,18 +22,24 @@
 
 import click
 
+from ..utils import split_commas
+
 
 class config_file_option:  # noqa: N801
```
andy-slac, line 28: Docstring is needed for this class and other option classes.

python/lsst/daf/butler/cli/opt/directory.py
```diff
@@ -37,12 +37,12 @@ class directory_argument:  # noqa: N801
         The help text for this argument to append to the command's help text.
         If None or '' then nothing will be appended to the help text (in which
```
andy-slac: Can you change single quotes to double quotes here, it may be hard to distinguish it from single double quote in generated docs. Or maybe just say "empty string"?

timj (approved):
Looks okay. A comment in the pull request description providing an overview of the changes would have helped me a little since the PR title doesn't help (I really need to take another look at that dev guide rule) and the commit messages are very terse. I think what you've done here is rearrange the code a bit into a class so that the pipetask infrastructure can use it.

python/lsst/daf/butler/cli/opt/log_level.py
```diff
@@ -0,0 +1,61 @@
...
+        ----------
+        help : str, optional
+            The help text to use, by default defaultHelp
+        required : bool, optional
+            If true then the caller must pass this option when calling the
+            command. By default False
+        defaultValue : str, optional
+            The default value to use if this option is not required. By default
+            `log_level_option.defaultValue`
+        """
```
andy-slac: This docstring should be at class level.

python/lsst/daf/butler/cli/utils.py
```diff
@@ -102,7 +107,11 @@ def clickResultMsg(result):
     msg : `str`
         The message string.
     """
-    return f"output: {result.output} exception: {result.exception}"
+    msg = io.StringIO()
+    if result.exception:
+        traceback.print_tb(result.exception.__traceback__, file=msg)
+        msg.seek(0)
```
andy-slac, line 133: You can do `msg.getvalue()` to get whole contents without needing `seek(0)` and `read()`.

python/lsst/daf/butler/tests/cliOptionTestBase.py
```diff
@@ -22,16 +22,308 @@
...
+        determine if the option should be initialzied with a `parameterType`
+        keyword argument, and sets it accordingly (`ParameterType.OPTION` is
+        the standard default for parameters so this only sets it for
+        `ARGUMENT`)
+
+        Parameters
+        ----------
+        optTestBase : `OptTestBase` subclass
+            The test being executed.
+        parameterKwargs : `dict` [`str`: `Any`], optional
```
andy-slac: needs comma instead of colon:

    `dict` [`str`, `Any`]

same applies to docstrings below.

python/lsst/daf/butler/tests/cliOptionTestBase.py
```diff
@@ -22,16 +22,308 @@
...
+            A list of keyword arguments to pass to the parameter (Argument or
+            Option) constructor.
+
+        Returns
+        -------
+        cli : a click.Command function.
```
andy-slac: Is "a click.Command function" different from \``click.Command`\`?

python/lsst/daf/butler/tests/cliOptionTestBase.py
```diff
@@ -22,16 +22,308 @@
...
+            A click command function that can be invoked by the click test
+            invoker.
+        """
+        cliArgs = copy.copy(parameterKwargs) if parameterKwargs is not None else {}
+        if 'parameterType' not in cliArgs and optTestBase.isArgument and optTestBase.isParameter:
```
andy-slac: Strings need double quotes (many other places too).

python/lsst/daf/butler/tests/cliOptionTestBase.py
```diff
@@ -22,16 +22,308 @@
...
+        """The abbreviated flag that is used on the command line for the
+        option.
+        """
+        return f"-{self.shortOptionName}" if self.shortOptionName else None
+
+    @property
+    @abc.abstractmethod
+    def optionName(self):
+        """The option name, matches the option flag that appears on the command
+        line
```
andy-slac: Need more periods everywhere at the end of sentences.

python/lsst/daf/butler/tests/cliOptionTestBase.py
```diff
@@ -44,37 +336,275 @@ def run_command(self, cmd, args):
...
+        """
+        with testObj.runner.isolated_filesystem():
+            if testObj.valueType.exists:
+                # If the file or dir is expected to exist, create it since it's
+                # it doesn't exist because we're running in a temporary
+                # directory.
+                if testObj.valueType.dir_okay:
+                    os.makedirs(testObj.optionValue)
+                elif testObj.valueType.file_okay:
+                    _ = open(testObj.optionValue)
```
andy-slac: My guess is that you are creating a file here, should this be `open(..., "w")`? And I think it needs an explicit `close()` or `with open(...)`.

tests/test_cliCmdQueryDatasetTypes.py
```diff
@@ -55,9 +55,9 @@ def test_requiredMissing(self):
     def test_all(self):
         """Test all parameters."""
         self.run_test(["query-dataset-types", "here", "--verbose", "foo*", "--components"],
-                      self.makeExpected(repo="here", verbose=True, glob=("foo*", ), components=True))
+                      self.makeExpected(repo="here", verbose=True, glob=["foo*", ], components=True))
```
andy-slac: Can remove comma too to save two characters 🙂 (and below too)

andy-slac (approved):
Looks OK, few minor comments.

python/lsst/daf/butler/tests/cliOptionTestBase.py
n8pease: In this case `click.Command` is more correct. By "command function" I mean the function decorated by `@click.Command`, whereas a `click.Command` instance would wrap & contain a command function.

python/lsst/daf/butler/cli/log.py
```diff
@@ -0,0 +1,160 @@
...
+
+        lgr = logging.getLogger()
+        if lsstLog is not None:
+            # global logging config
+
+            lsstLog.configure_prop(_LOG_PROP.format(cls.longLogFmt if longlog else cls.normalLogFmt))
+
+            # Forward all Python logging to lsstLog
+            lgr.addHandler(lsstLog.LogHandler())
+        lgr.setLevel(logging.INFO)
```
andy-slac: If `lsstLog` is `None` then I think we probably need to call `logging.basicConfig(level=INFO)` instead of just `lgr.setLevel(logging.INFO)`.

python/lsst/daf/butler/cli/log.py
```diff
@@ -0,0 +1,160 @@
...
+
+        # also capture warnings and send them to logging
+        logging.captureWarnings(True)
+
+    @classmethod
+    def uninitLog(cls, components):
+        if lsstLog is not None:
+            lgr = logging.getLogger()
+            lgr.removeHandler(lsstLog.LogHandler())
```
andy-slac: Hmm 🤔 I'm not sure it is going to work. Do you have unit test for this?

python/lsst/daf/butler/cli/log.py
```diff
@@ -0,0 +1,160 @@
...
+            logger = logging.getLogger(component)
+            logger.setLevel(logging.NOTSET)
+
+    @classmethod
+    def setLogLevels(cls, logLevels):
+        """Set log level for one or more components or the root logger.
+
+        Parameters
+        ----------
+        longlog : `bool`
```
andy-slac: NSP - no such parameter :)

python/lsst/daf/butler/cli/log.py
```diff
@@ -0,0 +1,160 @@
...
+            (component, level), `component` is a logger name or an empty string
+            or `None` for root logger, `level` is a logging level name, one of
+            CRITICAL, ERROR, WARNING, INFO, DEBUG (case insensitive).
+        """
+
+        if isinstance(logLevels, dict):
+            logLevels = logLevels.items()
+
+        # configure individual loggers
+        for component, level in logLevels:
```
andy-slac: DO you need to do `level = level.upper()`?

python/lsst/daf/butler/cli/log.py
n8pease: @andy-slac can you say more about what part won't work? What would you like tested?

andy-slac: What I think is that `removeHandler` removes existing handler from the the handler list. `lsstLog.LogHandler()` make a new handler that was never added to that list, so removing it does not even make sense (it is probably no-op). What you want is to remove handler instance that you created earlier and added with `addHandler` , this means you need to remember that instance somewhere. As for testing, I think you can check the size of the `lgr.handlers` to see if removing changes its size (though `handlers` is an undocumented attribute)

timj: You can't test in daf_butler itself because daf_butler doesn't depend on lsst.log.

andy-slac: You can if you (accidentally) setup `log` or some other package that depends on `log`.

timj: That is true but you can't rely on it from a Jenkins point of view. ctrl_mpexec can test it though.

n8pease: I have tests in obs_base for features in daf_butler that depend on lsst.log. Let me know if you think that's not a good place.
I don't think I can depend on the actual number of handlers, tests may run before this test may affect it. I'm not even sure I can depend on a delta. I might have to figure out how to tell if a specific handler is in the list, or punt on being able to know.
The affect that persistent logging configuration has on unrelated unit tests that run in the same process is really giving me hell. I might have questions in Slack today.

timj: obs_base is good.   Logging does affect global state which is why logger configurations should only be triggered by command line tooling and not by module code.

PR #319, opened 2020-06-25
DM-25447: Add support for read-only components

timj:
This was quite a lot more code than I was expecting. There is still an issue over what to do with parameters with read-only components. For now they are sent to the component that is being used to calculate the read-only component and can't be sent to the assembler at all.

### DIFF
--- a/config/storageClasses.yaml
+++ b/config/storageClasses.yaml
@@ -23,6 +23,12 @@ storageClasses:
     pytype: lsst.afw.cameraGeom.Camera
   Detector:
     pytype: lsst.afw.cameraGeom.Detector
+  Box2I:
+    pytype: lsst.geom.Box2I
+  Extent2I:
+    pytype: lsst.geom.Extent2I
+  Point2I:
+    pytype: lsst.geom.Point2I
   Image: &Image
     pytype: lsst.afw.image.Image
     assembler: lsst.daf.butler.assemblers.exposureAssembler.ExposureAssembler
@@ -118,6 +124,10 @@ storageClasses:
       filter: Filter
       detector: Detector
       validPolygon: Polygon
+    readComponents:
+      bbox: Box2I
+      dimensions: Extent2I
+      xy0: Point2I
   ExposureF:
     inheritsFrom: Exposure
     pytype: lsst.afw.image.ExposureF
--- a/config/storageClasses.yaml
+++ b/config/storageClasses.yaml
@@ -67,6 +67,8 @@ storageClasses:
     pytype: lsst.afw.table.BaseCatalog
   PeakCatalog:
     pytype: lsst.afw.detection.PeakCatalog
+  Polygon:
+    pytype: lsst.afw.geom.Polygon
   SimpleCatalog:
     pytype: lsst.afw.table.SimpleCatalog
   SourceCatalog:
@@ -114,6 +116,8 @@ storageClasses:
       transmissionCurve: TransmissionCurve
       metadata: PropertyList
       filter: Filter
+      detector: Detector
+      validPolygon: Polygon
   ExposureF:
     inheritsFrom: Exposure
     pytype: lsst.afw.image.ExposureF
--- a/python/lsst/daf/butler/core/datasets/type.py
+++ b/python/lsst/daf/butler/core/datasets/type.py
@@ -117,7 +117,7 @@ class DatasetType:
 
     def __init__(self, name: str, dimensions: Union[DimensionGraph, Iterable[Dimension]],
                  storageClass: Union[StorageClass, str],
-                 *, parentStorageClass: Optional[Union[StorageClass, str]] = None,
+                 parentStorageClass: Optional[Union[StorageClass, str]] = None, *,
                  universe: Optional[DimensionUniverse] = None):
         if self.VALID_NAME_REGEX.match(name) is None:
             raise ValueError(f"DatasetType name '{name}' is invalid.")
@@ -128,7 +128,9 @@ class DatasetType:
                                  "a universe must be provided.")
             dimensions = universe.extract(dimensions)
         self._dimensions = dimensions
-        assert isinstance(storageClass, (StorageClass, str))
+        if not isinstance(storageClass, (StorageClass, str)):
+            raise ValueError("StorageClass argument must be StorageClass or str. "
+                             f"Got {storageClass}")
         self._storageClass: Optional[StorageClass]
         if isinstance(storageClass, StorageClass):
             self._storageClass = storageClass
@@ -140,6 +142,10 @@ class DatasetType:
         self._parentStorageClass: Optional[StorageClass] = None
         self._parentStorageClassName: Optional[str] = None
         if parentStorageClass is not None:
+            if not isinstance(storageClass, (StorageClass, str)):
+                raise ValueError("Parent StorageClass argument must be StorageClass or str. "
+                                 f"Got {parentStorageClass}")
+
             # Only allowed for a component dataset type
             _, componentName = self.splitDatasetTypeName(self._name)
             if componentName is None:
@@ -155,13 +161,16 @@ class DatasetType:
         # a component and is not specified when we don't
         _, componentName = self.splitDatasetTypeName(self._name)
         if parentStorageClass is None and componentName is not None:
-            raise RuntimeError(f"Component dataset type '{self._name}' constructed without parent"
-                               " storage class")
+            raise ValueError(f"Component dataset type '{self._name}' constructed without parent"
+                             " storage class")
         if parentStorageClass is not None and componentName is None:
-            raise RuntimeError(f"Parent storage class specified by {self._name} is not a composite")
+            raise ValueError(f"Parent storage class specified by {self._name} is not a composite")
 
     def __repr__(self) -> str:
-        return "DatasetType({}, {}, {})".format(self.name, self.dimensions, self._storageClassName)
+        parent = ""
+        if self._parentStorageClassName:
+            parent = f", parentStorageClass={self._parentStorageClassName}"
+        return f"DatasetType({self.name}, {self.dimensions}, {self._storageClassName}{parent})"
 
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, type(self)):
@@ -171,9 +180,15 @@ class DatasetType:
         if self._dimensions != other._dimensions:
             return False
         if self._storageClass is not None and other._storageClass is not None:
-            return self._storageClass == other._storageClass
+            if self._storageClass != other._storageClass:
+                return False
         else:
-            return self._storageClassName == other._storageClassName
+            if self._storageClassName != other._storageClassName:
+                return False
+        if self._parentStorageClass is not None and other._parentStorageClass is not None:
+            return self._parentStorageClass == other._parentStorageClass
+        else:
+            return self._parentStorageClassName == other._parentStorageClassName
 
     def __hash__(self) -> int:
         """Hash DatasetType instance.
@@ -181,7 +196,8 @@ class DatasetType:
         This only uses StorageClass name which is it consistent with the
         implementation of StorageClass hash method.
         """
-        return hash((self._name, self._dimensions, self._storageClassName))
+        return hash((self._name, self._dimensions, self._storageClassName,
+                     self._parentStorageClassName))
 
     @property
     def name(self) -> str:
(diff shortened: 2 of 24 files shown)


config/storageClasses.yaml
```diff
@@ -67,6 +67,8 @@ storageClasses:
     pytype: lsst.afw.table.BaseCatalog
   PeakCatalog:
     pytype: lsst.afw.detection.PeakCatalog
+  Polygon:
+    pytype: lsst.afw.geom.Polygon
```
TallJimbo, line 77: We can also probably save ourselves some work later by associating this storage class with the generic `readFits`/`writeFits` formatter in the datastore config.

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -165,6 +179,19 @@ def assemblerClass(self) -> Optional[Type]:
         self._assembler = doImport(self._assemblerClassName)
         return self._assembler
 
+    def allComponents(self) -> Dict[str, StorageClass]:
```
TallJimbo: If you make this `Mapping` instead of `Dict` now, it'll be easier to change the implementation to, say, `collections.ChainMap` instead of a `dict` copy later.

config/storageClasses.yaml
```diff
@@ -23,6 +23,14 @@ storageClasses:
...
   Detector:
     pytype: lsst.afw.cameraGeom.Detector
+  Box2I:
+    pytype: lsst.geom.Box2I
+  Extent2I:
+    pytype: lsst.geom.Extent2I
+  Point2I:
+    pytype: lsst.geom.Point2I
+  Filter:
+    pytype: lsst.afw.image.Filter
```
TallJimbo: This gets me wondering about whether we should allow (maybe just in this context, maybe more broadly) regular Python classes to be used as `StorageClasses` when their definitions are trivial, rather than requiring those to be added explicitly.

tests/test_datastore.py
```diff
@@ -205,7 +205,71 @@ def testBasicPutGet(self):
...
+
+        # Dummy dataId
+        dimensions = self.universe.extract(("visit", "physical_filter"))
+        dataId = {"instrument": "dummy", "visit": 428, "physical_filter": "R"}
+
+        for i, sc in enumerate(storageClasses):
+            with self.subTest(storageClass=sc.name):
+                # Create a different dataset type each time round
+                # so that a test failure in this subtest does not trigger
+                # a cascade of tests because of file clashes
```
TallJimbo, line 238: :100: .  Thanks for thinking about this.

python/lsst/daf/butler/datastores/inMemoryDatastore.py
```diff
@@ -293,12 +293,27 @@ def get(self, ref: DatasetRef, parameters: Optional[Mapping[str, Any]] = None) -
...
 
-        component = ref.datasetType.component()
+        # if this is a read only component we need to apply parameters
+        # before we retrieve the component. We assume that the parameters
+        # will affect the data globally, before the read-only component
+        # is selected.
+        if isReadOnlyComponent:
+            inMemoryDataset = writeStorageClass.assembler().handleParameters(inMemoryDataset, parameters)
+            # Then disable parameters for later
+            parameters = {}
```
TallJimbo, line 316: Is there a concrete case where this matters?  I'm just wondering if we should say, "you can't pass parameters if you want a read-only component" to try to make things a _little_ simpler.

Edit: after writing the above I saw the commit message that referenced an `origin` parameter to `bbox`.  I'd be okay with just dropping support for that (`origin=PARENT` is all we need - the box returned with `origin=LOCAL` is trivially calculable from the one returned with `origin=PARENT`).  I'd be fine with dropping the `xy0` and `dimensions` read components as being similarly redundant, though I gather that doesn't gain us as much.

python/lsst/daf/butler/core/formatter.py
```diff
@@ -385,6 +385,35 @@ def segregateParameters(self, parameters: Optional[Dict[str, Any]] = None) -> Tu
...
+        Raises
+        ------
+        NotImplementedError
+            Raised if this formatter refuses to answer the question.
+        ValueError
+            Raised if this formatter can not determine a relevant component
+            from the supplied options.
+        """
+        raise NotImplementedError("This formatter does not support read components")
+
```
TallJimbo: Would it make any sense to put this interface on the assembler?  Or maybe make it a fixed mapping in the StorageClass definitions?  I can't think of a case where it would need to depend on the actual formatter used for the components - whether a component can handle a particular `readComponent` seems conceptually like it's a general property of that component's storage class.

Would either or both of those options sidestep the need to write a row with the composite formatter to the datastore records?  The latter at least vaguely sounds like something I think you mentioned as a possibility on Slack, but I don't recall reading why you didn't go with that approach.

TallJimbo (approved):
Code looks good.  I have some of the same concerns on the big picture and increasing complexity that you do, but I'm mostly happy to trust that you've looked into alternatives more than I have and have chosen least-bad options.  On the off chance a thought I had for simplifying things might not be one you've considered, I've left a couple of comments on those big-picture things at various points in the PR.

One more quick big-picture thought: in terms of nomenclature, I wonder if we should not be calling these "readComponents" a type of "component" at all - maybe "properties", with what we're calling "allComponents" here (i.e. real components and properties) -> "attributes"?  Just a thought - I don't like those names enough better that I'd want to push for changing things this late in the game, but the naming right now feels like we started out by thinking of these as another kind of component, and by the end it became clear that they were more different than similar.

python/lsst/daf/butler/core/formatter.py
timj: You are absolutely right that this look up should be on the assembler and not the formatter. It is a function of the python type so the assembler does know which component is best. This could mean I can remove all the code that added the extra row although I need to ponder how the component can find out the composite storage class so it can get the right assembler. On the other hand the relationship of bbox to Exposure is something that can be worked out without caring about the formatter -- I was using the formatter because the formatter knew about the python type as well but that's clearly superfluous.  I might have to change DatasetType to add a compositeAssembler property.

This is what code review is all about! Granted it means a lot of the code on this ticket is irrelevant.

timj:
I called them read components because from the point of view of the caller they act exactly like components. The caller doesn't even know that some are read/write and some are read-only because they have no idea which of the components ends up being important for disassembly.  We still refer to them as "composite.component" and calling them a property or attribute might be easier to talk about internally (since talking about read-write component relevant for read-only component is annoying at times) but for the general interface they are components.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -141,7 +149,7 @@ def register(self, datasetType: DatasetType) -> Tuple[DatasetRecordStorage, bool
             inserted = False
         if inserted and datasetType.isComposite:
             for component in datasetType.storageClass.allComponents():
-                self.register(datasetType.makeComponentDatasetType(component))
+                pass  # self.register(datasetType.makeComponentDatasetType(component))
```
TallJimbo: Looks like you can now remove this whole block.

TallJimbo: And should we also add in this function to make sure no one tries to register just a component?  There are probably other places we could put that kind of check, but this is the only one I can think of where it might actually succeed (and mess up the registry) rather than failing with some possibly-less-friendly error.

tests/test_butler.py
```diff
@@ -686,14 +686,18 @@ def testGetDatasetTypes(self):
             butler.registry.insertDimensionData(*args)
 
         # When a DatasetType is added to the registry entries are not created
-        # for components.
+        # for components but querying them can reeturn the components.
```
TallJimbo: typo: reeturn

python/lsst/daf/butler/core/datasets/type.py
```diff
@@ -273,6 +274,21 @@ def makeComponentDatasetType(self, component: str) -> DatasetType:
...
+        dataset type.
+
+        Returns
+        -------
+        all : `list` of `DatasetType`
+            All the component dataset types. If this is not a composite
+            then returns an empty list.
+        """
+        if not self.isComposite():
+            return []
```
TallJimbo: Do you actually need this check?  I'd have thought `storageClass.allComponents()` would return an empty iterable in this case anyway.

python/lsst/daf/butler/core/datasets/type.py
```diff
@@ -81,12 +84,17 @@ class DatasetType:
     storageClass : `StorageClass` or `str`
         Instance of a `StorageClass` or name of `StorageClass` that defines
         how this `DatasetType` is persisted.
+    parentStorageClass : `StorageClass` or `str`
```
TallJimbo: Should probably be `, optional`.

python/lsst/daf/butler/core/datasets/type.py
```diff
@@ -131,6 +140,28 @@ def __init__(self, name: str, dimensions: Union[DimensionGraph, Iterable[Dimensi
...
+            else:
+                self._parentStorageClassName = parentStorageClass
+
+        # Temporarily warn if this is a component but is missing the
+        # parent storage class. Usually implies some place not using the right
+        # API.
+        _, componentName = self.splitDatasetTypeName(self._name)
+        if parentStorageClass is None and componentName is not None:
+            log.warning("Component dataset type '%s' constructed without parent storage class",
+                        self._name)
```
TallJimbo: Unless this fails tests that are a pain to fix, I'd be happy to bump this up to an exception.

python/lsst/daf/butler/core/datasets/type.py
```diff
@@ -179,6 +210,20 @@ def storageClass(self) -> StorageClass:
             self._storageClass = StorageClassFactory().getStorageClass(self._storageClassName)
         return self._storageClass
 
+    @property
+    def parentStorageClass(self) -> Optional[StorageClass]:
+        """`StorageClass` instance that defines how the composite associated
+        with this  `DatasetType` is persisted. Note that if DatasetType was
+        constructed with a name of a StorageClass then Butler has to be
+        initialized before using this property. Can be `None` if this is
+        not a component of a composite.
```
TallJimbo: Docstring should be split into brief + extended descriptions.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -899,8 +899,29 @@ def get(self, ref: DatasetRef, parameters: Optional[Mapping[str, Any]] = None) -
...
+        # component requested is listed as a read-only component for the
+        # composite storage class
+        isDisassembledReadOnlyComponent = False
+        if isDisassembled and refComponent:
+            # The composite storage class should be accessible through
+            # the component dataset type
+            compositeStorageClass = ref.datasetType.parentStorageClass
+
+            # Assume something else will handle things if we can not find
+            # it
```
TallJimbo: Is this actually a reasonable assumption?  Is there some other part of the system that could handle this case?

TallJimbo (approved):
Looks good!  I'm glad the last review pass was useful.  I don't have much to add in this pass - just minor comments - though some of the old ones are still relevant as well.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
timj: If the parent storage class is not available this will assume it's a real component being requested. At that point the branch below that looks up the disassembled component will fail and you'll get an error about the component not being recognized.

python/lsst/daf/butler/core/datasets/type.py
timj: I set it to be an exception to find where tests were breaking. I think a couple of tests were still morning (mostly the ones that set up fake datasetrefs). -- I'll take another look at them. Raising an exception will mean that this code will fail if an old registry is used when the code encounters the component dataset types but I imagine that's probably okay.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
timj: Yes. I did mean to remove that. I can add a check here.

python/lsst/daf/butler/core/datasets/type.py
timj: Yes. I was prematurely optimizing.

config/storageClasses.yaml
timj: That's an interesting thought albeit not for this ticket. So for `StorageClass("lsst.afw.image.Filter")` then rather than defaulting to `object` we assume that the pytype name is the storage class name... (we could even adopt the convention that names that start with upper case are never pytypes). Worth pondering.

python/lsst/daf/butler/datastores/inMemoryDatastore.py
timj: I'll ponder this some more when I write up the formatter/assembler documentation. Stating that parameters are never allowed for read-only components is the easiest approach.

PR #322, opened 2020-07-09
DM-25881: Add extra logging to S3 datastore

hsinfang:


python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -208,9 +213,10 @@ def _write_in_memory_to_artifact(self, inMemoryDataset: Any, ref: DatasetRef) ->
...
         try:
             serializedDataset = formatter.toBytes(inMemoryDataset)
+            log.debug(f"Writing file directly to: '{location.uri}''")
             self.client.put_object(Bucket=location.netloc, Key=location.relativeToPathRoot,
                                    Body=serializedDataset)
-            log.debug("Wrote file directly to %s", location.uri)
+            log.debug(f"Successfully wrote file directly to {location.uri}")
         except NotImplementedError:
             with tempfile.NamedTemporaryFile(suffix=location.getExtension()) as tmpFile:
                 formatter._fileDescriptor.location = Location(*os.path.split(tmpFile.name))
```
timj, line 222: I think we also need a log message around line 225 (new) where we upload this temporary file. This is the code path that will trigger for fits files.

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -113,6 +113,7 @@ def _artifact_exists(self, location: Location) -> bool:
         exists : `bool`
             True if the location can be found, false otherwise.
         """
+        log.debug(f"Checking if file exists: '{location.uri}''")
```
timj: As a general comment for debug log messages we tend to use the `%s` deferred formatting approach following the notion that there is no point stringifying `location.uri` if the debug message is not going to be issued. So all these should ideally be changed to:
```python
log.debug("something or other %s", location.uri)
```

timj (approved):
Looks good in general but I have a couple of quick changes that need to happen.

python/lsst/daf/butler/datastores/s3Datastore.py
hsinfang: Adding line 225 (new), before it attempts the put_object via a temporary file.

PR #323, opened 2020-07-11
DM-15465: Clean up utils usage with a view to supporting test running in pull requests

timj:
* Removed getPackageDir and replaced with package resources
* Remove spurious usage of lsst.utils.tests in one test file

Still have many doImport usages and one test uses a getTempFilePath call.


timj:
This now seems to be passing so that pull requests will run the tests and build the documentation.  It takes 8minutes to run and is currently not required for merging. Two minutes of that is building sphgeom so maybe we can upload a wheel as a "GitHub release" and pull that to save some time.

I wonder if we have to reconsider the pull_request and push builds. Are they the same once a pull request has been created?

python/lsst/daf/butler/tests/cliCmdTestBase.py
```diff
@@ -103,11 +103,11 @@ def run_missing(self, inputs, expectedMsg):
             followed by arguments, option keys and option values.
         expectedMsg : `str`
             An error message that should be present in stdout after running the
-            subcommand.
+            subcommand. Can be a regular expression string.
         """
         result = self.run_command(inputs)
         self.assertNotEqual(result.exit_code, 0, clickResultMsg(result))
-        self.assertIn(expectedMsg, result.stdout)
+        self.assertRegex(result.stdout, expectedMsg)
```
TallJimbo, line 133: If these tests cause problems ever again I vote for dropping them.  I don't think I've ever met a stringification test that added net value, even for classes that actually had broken stringification.

TallJimbo (approved):
Only big picture comment is that with all of the conditionals in `ButlerURI` I wonder if it'd work better as an ABC with separate implementations for remote URIs, POSIX paths, and now pkg_resources.  Or maybe a concrete class that delegated to an ABC or something.  Probably out of scope for this ticket, but something to think about.

timj:
Regarding ButlerURI returning subclasses for different URI schemes, that does seem worth thinking about. It would also allows the addition of more concrete methods such as "does this item exist" and "download this resource".

doc/lsst.daf.butler/configuring.rst
```diff
@@ -22,7 +22,7 @@ To construct a Butler configuration object (`~lsst.daf.butler.ButlerConfig`) fro
   Shell variables will be expanded.
   The contents of these files will then be inserted into the configuration at the same hierarchy as the ``includeConfigs`` directive, with priority given to the values defined explicitly in the parent configuration (for lists of include files later files overwrite content from earlier ones).
 * Each sub configuration class is constructed by supplying the relevant subset of the global config to the component Config constructor.
-* A search path is constructed by concatenating the supplied search path, the environment variable path (``$DAF_BUTLER_CONFIG_PATH``), and the daf_butler config directory (``$DAF_BUTLER_DIR/config``).
+* A search path is constructed by concatenating the supplied search path, the environment variable path (``$DAF_BUTLER_CONFIG_PATH``), and the daf_butler config directory (using the ``lsst.daf.butler`` package resource).
```
ktlim:
```suggestion
* A search path is constructed by concatenating the supplied search path, the environment variable path (``$DAF_BUTLER_CONFIG_PATH``), and the daf_butler config directory (within the ``lsst.daf.butler`` package resource).
```

.github/workflows/build.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Install xdist
+        run: pip install pytest-xdist
+
+      - name: Build and install
+        run: pip install -v .
+
+      - name: Run tests
+        run: pytest -r a -v -n 3
+
+      - name: Install documenteer
```
ktlim, line 42: Note for the future: Should package the install and build as an action that others can reuse.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Install sqlite
+        run: sudo apt-get install sqlite libyaml-dev
+
+      # It seems like pip install of requirements for a URL package
+      # does not install those dependencies
+      - name: Install sphgeom dependencies
+        run: pip install -r https://raw.githubusercontent.com/lsst/sphgeom/master/requirements.txt
+
+      - name: Install dependencies
```
ktlim, line 29: Note for the future: these four steps (deps, xdist, install ., pytest) could be packaged as an action for others to reuse.

setup.cfg
```diff
@@ -1,3 +1,50 @@
+[metadata]
+name = daf_butler
+description = An abstraction layer for reading and writing astronomical data to datastores
+author = Rubin Observatory Data Management
+author_email = support@lsst.org
```
ktlim: Who does this go to?

setup.cfg
```diff
@@ -1,3 +1,50 @@
...
+license = GPLv3+
```
ktlim: I thought this was discouraged in favor of `classifiers`: https://packaging.python.org/guides/distributing-packages-using-setuptools/#license

tests/test_dbAuth.py
```diff
@@ -420,14 +419,5 @@ def test_urlEncoding(self):
             "%2A%5B%5D@host.example.com:5432/my_database")
 
 
-class MemoryTester(lsst.utils.tests.MemoryTestCase):
```
ktlim, line 423: We're losing the open-fd test by doing this, right?

requirements.txt
```diff
@@ -0,0 +1,16 @@
+pyyaml >= 5.1
```
ktlim, line 1: Unfortunate that we need these twice: here and in `setup.cfg`.  Is this the standard way to maintain these?

python/lsst/daf/butler/core/config.py
```diff
@@ -25,11 +25,14 @@
...
 __all__ = ("Config", "ConfigSubset")
 
+from dataclasses import dataclass
 import collections
 import copy
 import logging
 import pprint
 import os
+import pkg_resources
+import posixpath
```
ktlim, line 35: You should add a comment to explain why this is preferred over `os.path` and `pathlib`.

python/lsst/daf/butler/core/config.py
```diff
@@ -353,7 +456,13 @@ def _processExplicitIncludes(self):
         # Search paths for config files
         searchPaths = [os.path.curdir]
         if self.configFile is not None:
-            searchPaths.append(os.path.abspath(os.path.dirname(self.configFile)))
+            if isinstance(self.configFile, str):
```
ktlim, line 461: If you could turn `str` into `ButlerURI` earlier you could get rid of this case and the one below.  But if you can turn both into a `Resource` that would be even better.

python/lsst/daf/butler/core/config.py
```diff
@@ -385,10 +494,37 @@ def _processExplicitIncludes(self):
...
+                                        break
+                                elif dir.scheme == "file":
+                                    # import private helper function
+                                    from .location import posix2os
+                                    filePath = posix2os(posixpath.join(dir.path, fileName))
+                                    if os.path.exists(filePath):
+                                        found = os.path.normpath(os.path.abspath(filePath))
+                                        break
+                                else:
+                                    # S3 is harder because we can't tell if
```
ktlim: You can use `head_object()` to determine if it is there by checking its metadata.

python/lsst/daf/butler/core/config.py
```diff
@@ -140,6 +148,66 @@ def extractFile(self, filename):
         return yaml.load(response["Body"], Loader)
 
 
+@dataclass
+class Resource:
```
ktlim, line 152: Is there a way to unify filesystem paths or even URIs with these `Resource` paths so that they appear to the rest of the code as a uniform class instead of having to type-case everywhere?

setup.py
```diff
@@ -0,0 +1,70 @@
...
+from setuptools import setup
+
+
+version = "0.1.0"
+with open("./python/lsst/daf/butler/version.py", "w") as f:
+    print(f"""
+__all__ = ("__version__", )
+__version__='{version}'""", file=f)
+
+# The purpose of this setup.py is to build enough of the system to
```
ktlim, line 14: Almost worth including it as a git submodule...

setup.py
```diff
@@ -0,0 +1,70 @@
...
+            with open(outpath, "w") as fh:
+                print(contents, file=fh)
+    # Write a simple __init__.py
+    if not failed:
+        init_path = os.path.join(utils_dir, "__init__.py")
+        if not os.path.exists(init_path):
+            with open(init_path, "w") as fh:
+                print("from .doImport import *", file=fh)
+
+# read the contents of your README file
```
ktlim:
```suggestion
# read the contents of our README file
```

ktlim (approved):

requirements.txt
timj: I will admit that @jonathansick tried to explain to me once why we have to duplicate requirements but I haven't really grokked it yet. You can't do "pip install requirements from setup.cfg" so a separate file gives you that.  setuptools doesn't know about requirements.txt so ignores them and uses the ones from setup.cfg.

tests/test_dbAuth.py
timj: Yes. We weren't using it anywhere else though. I had been thinking of using pytest-openfiles instead but I've never been able to convince pytest-openfiles that there was a leak when I experimented with it in the past.

setup.cfg
timj: Probably. I am using the classifiers so I'll remove this. It was mostly historical copy and paste.

setup.cfg
timj: No idea. It was a placeholder. Since I'm not uploading to PyPI I guess it's better to remove it.

.github/workflows/build.yaml
timj: An action can import another action?

python/lsst/daf/butler/core/config.py
timj: I can't turn an S3 URI into a `Resource` but I should be able to convert all the plain strings into schemeless URIs.  I had thought about adding a pkg_resources URI scheme to ButlerURI but ButlerURI can't do concrete things yet (like check for existence).

python/lsst/daf/butler/core/config.py
timj: Or use `s3utils.s3CheckFileExists()` I think.   None of the S3 support is tested at the moment. I think I'd need to add a s3config test that uploads some test configurations to the fake moto bucket and then tries to reconstruct the config from it.  I don't think @DinoBektesevic or @hsinfang are making use of DAF_BUTLER_CONFIG_PATH at the moment or using `!include` so I'm suggesting I defer implementation until we need it.

python/lsst/daf/butler/core/config.py
timj: Adding `pkg_resources://lsst.daf.butler/configs/datastore.yaml` URI support to ButlerURI might work but really needs `ButlerURI` to be updated to have concrete methods for checking existence and even opening a stream.  I think some of this ties in with @TallJimbo wondering if ButlerURI could be changed so that the constructor returns specialist S3ButlerURI, PosixFileURI and LocalFileURI implementations.

setup.py
timj: but then I'd still have to write a new `__init__.py` that didn't try to import all the C++ routines. I imagine that these github.com gets should be fairly stable for actions that themselves run on github.com.

.github/workflows/build.yaml
ktlim: A _workflow_ (which this is) uses actions.  I'm saying to package up the `run` scripts as a reusable action.

python/lsst/daf/butler/core/config.py
ktlim: I think this is the same as Jim's ABC comment above.  We're not talking about `Resource` as currently defined but a new `Resource` ABC.

tests/test_dbAuth.py
timj: I've added pytest-openfiles and it found a real problem: our "test repo" tests that were meant to be using in-memory registry for speed were in fact using sqlite files because we had the wrong argument for makeRepo. We should add it to the conda env.

requirements.txt
jonathansick: I don't understand the strategy here. If you've got a `setup.cfg`, then you shouldn't need this `requirements.txt` file at all. What is the `requirements.txt` file being used for?

timj: Ok. I thought we had a whole conversation about this back when you were here.  I was following: https://packaging.python.org/discussions/install-requires-vs-requirements/  to a certain extent except I wasn't really doing the "minimal in setup.cfg and maximal in requirements.txt approach".

I'm pretty sure you can't ask pip to install requirements specified in a setup.cfg so you absolutely need the requirements.txt for that (and you need to somehow install setuptools_cpp when building sphgeom before you even run setup.py since that's required to configure the C++ build). On the other hand setuptools doesn't look at requirements.txt and I don't see situations where people try to read the requirements.txt into their setup.py execution. Also requirements.txt lets people preinstall dependencies using pip before they try to install the package itself.  It seems like an inconsistent mess at first glance.

PR #324, opened 2020-07-12
DM-25776: Refactor DataCoordinates and add custom containers for them.

TallJimbo:


doc/lsst.daf.butler/dimensions.rst
```diff
@@ -55,14 +55,13 @@ Different datasets with the same `DatasetType` are always identified by the same
...
-Informal dictionary data IDs can be transformed into `DataCoordinate` instances by calling `DataCoordinate.standardize` (which is what most `Butler` and `Registry` APIs that accept data IDs do under the hood).
-
-`ExpandedDataCoordinate` is its maximal counterpart.
-While its iteration and views still behave like those of dictionary that maps only *required* dimensions to their primary key values (for compatibility with its base class), one can look up the primary key values for implied dimensions as well.
-`ExpandedDataCoordinate` also has additional attributes that hold all of the metadata fields associated with its dimensions.
+The data IDs returned by the `Butler` or `Registry` (and most of those used internally) are usually instances of the `DataCoordinate` class.
+`DataCoordinate` instances can have different states of knowledge about the dimensions they identify.
+They always contains at least the key-value pairs that correspond to its `DimensionGraph`'s `~DimensionGraph.required` subset -- that is, the minimal set of keys needed to fully identify all other dimensions in the graph.
```
timj: "contains" -> "contain".

Also, doesn't there need to be a `\ ` or something between `DimensionGraph` and the `s`?

python/lsst/daf/butler/core/dimensions/_dataCoordinateIterable.py
```diff
@@ -0,0 +1,764 @@
...
+        `False`.
+    check: `bool`, optional
+        If `True` (default) check that all data IDs are consistent with the
+        given ``graph`` and state flags at construction.  If `False`, no
+        checking will occur.
+    """
+    def __init__(self, dataIds: Collection[DataCoordinate], graph: DimensionGraph, *,
+                 hasFull: Optional[bool] = None, hasRecords: Optional[bool] = None,
+                 check: bool = True):
+        self._dataIds = dataIds
```
timj, line 288: Is it at all possible for `dataIds` to be a mutable collection which can then be modified by the caller? ie should this be `self._dataIds = tuple(dataIds)` ?

python/lsst/daf/butler/core/dimensions/coordinate.py
```diff
@@ -19,83 +19,326 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+#
+# Design Notes for DataCoordinate
```
timj: There's a lot of good information in here that might be better off in real documentation.

python/lsst/daf/butler/core/dimensions/coordinate.py
```diff
@@ -19,83 +19,326 @@
...
+#
+# Overview
+# --------
+#
+# DataCoordinate is one of the most important classes in daf_butler, and one of
+# the trickiest to implement.  This is its fourth iteration, with each being
+# quite different from the other.  There are two big reasons for this:
+#
+# - We want this object to behave much like a `dict`, because plain `dict`
+#   objects are frequently used as informal and data IDs, but Python's
```
timj: grammar: "used as informal and data IDs"

python/lsst/daf/butler/core/dimensions/coordinate.py
```diff
@@ -19,83 +19,326 @@
...
+#
+#   There's actually even an aspect of (A) that we're not (yet) dealing with -
+#   we may in the future want to allow alternate sets of identifiers for a
+#   particular set of dimensions (e.g. dayObs + seqNum in place of integer IDs
+#   for LSST exposures).  That means we could have a data ID that does have
+#   sufficient information to fully identify a set of dimensions, but it's not
+#   the canonical set of identifiers, and hence we can't tell whether that data
+#   ID is equivalent to some other one until we've done a database query to get
+#   to canonical form.  I'm hoping that when that day comes we'll not try to
+#   make those kinds of data IDs into DataCoordinates, but even an adjacent
```
timj: I was assuming that we'd do this by having special code that sees an instrument in a dataId, looks up the instrument class, runs a dataId demangler (that combines dayObs and seqNum into a exposure_id) and then returns a proper DataCoordinate. From that point on dayObs would not exist.

python/lsst/daf/butler/core/dimensions/coordinate.py
```diff
@@ -19,83 +19,326 @@
...
+# earlier version of the dimensions subpackage.  I'm not going to bother
+# refreshing my memory now, as most of that is now irrelevant, and I do
+# remember its fatal flaw: it had too many states, especially with regard to
+# what DimensionRecords (or earlier equivalent) it held.  That was really hard
+# on code that needed to make a state transition (say, fetch some additional
+# records from the database so it could fill in a filename template).  Should
+# it fetch everything it needs, regardless of whether some of it was already
+# present, or is the complexity of trying to compute a diff between what it has
+# and what it needs worthwhile?  Should it fetch more than it needs, because
+# one fetch for many records can be a lot efficient than N separate fetches?
```
timj: lot -> lot more

python/lsst/daf/butler/core/dimensions/coordinate.py
```diff
@@ -19,83 +19,326 @@
...
+#
+# The second version is what we've had in the codebase for over a year as of
```
timj: Just say "since June 2019" rather than "over a year as of this writing"

python/lsst/daf/butler/core/dimensions/coordinate.py
```diff
@@ -111,115 +354,169 @@ def standardize(mapping: Optional[Union[Mapping[str, Any], NamedKeyMapping[Dimen
...
+
+        Returns
+        -------
+        dataId : `DataCoordinate`
+            A data ID object that identifies the given dimensions.
+            ``dataId.hasFull()`` will return `True` if and only if
+            ``graph.implied`` is empty, and ``dataId.hasRecords()`` will never
+            return `True`.
+        """
+        assert len(graph.required) == len(values)
```
timj: If this ever triggers people are going to be confused. Can we have an explicit RuntimeError with error message instead?

python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -430,16 +430,18 @@ def format(self, ref: DatasetRef) -> str:
...
-        # templates (and possibly when constructing DatasetRefs), but doing so
-        # would break a ton of otherwise-useful tests that would need to be
-        # modified to provide a lot more metadata.
-        fields = {k: v for k, v in getattr(ref.dataId, "full", ref.dataId).items() if v is not None}
-
+        # Extract defined non-None dimensions from the dataId.
+        # This guards against Nones being explicitly present in the data ID
+        # (which can happen if, say, an exposure has no filter), as well as
+        # the case where only required dimensions are present (which in this
+        # context should only happen in unit tests; in generally we need all
```
timj: generally -> general

python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -448,12 +450,12 @@ def format(self, ref: DatasetRef) -> str:
             # not be true in some test code, but that test code is a pain to
             # update to be more like the real world while still providing our
             # only tests of important behavior.
-            skypix = [dimension for dimension in ref.dataId.graph if isinstance(dimension, SkyPixDimension)]
+            skypix = [dimension for dimension in ref.datasetType.dimensions
+                      if isinstance(dimension, SkyPixDimension)]
             if len(skypix) == 1:
-                fields["skypix"] = fields[skypix[0]]
-
-        # Extra information that can be included using . syntax
```
timj, line 455: Can you attach this comment to where extras is now set?

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -778,34 +779,53 @@ def expandDataId(self, dataId: Optional[DataId] = None, *, graph: Optional[Dimen
...
+                    keys[element.name] = None
+                    record = None
+                else:
+                    storage = self._dimensions[element]
+                    dataIdSet = DataCoordinateIterable.fromScalar(
+                        DataCoordinate.standardize(keys, graph=element.graph)
+                    )
+                    fetched = tuple(storage.fetch(dataIdSet))
+                    try:
+                        (record,) = fetched
```
timj, line 825: This is a nice way to ensure only one item is returned.

tests/test_dimensions.py
```diff
@@ -212,5 +251,388 @@ def testPickling(self):
...
+            Data IDs to select from.  Defaults to ``self.allDataIds``.
+
+        Returns
+        -------
+        selected : `DataCoordinateSequence`
+            ``n`` Data IDs randomly selected from ``dataIds`` with replacement.
+        """
+        if dataIds is None:
+            dataIds = self.allDataIds
+        return DataCoordinateSequence(self.rng.sample(dataIds, n),
```
timj, line 336: This reminded me that we have code in `python/lsst/daf/butler/tests/_testRepo.py` that uses numpy for this for no reason at all (it's the only explicit usage of numpy in the butler) where `rng.sample(range(10), 10)` would work fine.

timj (approved):
Looks like a nice cleanup.

python/lsst/daf/butler/core/dimensions/_dataCoordinateIterable.py
TallJimbo: I thought about that a lot.  Since sometimes these will be pretty big collections, I preferred trusting the user to logically transfer ownership over making a (shallow) copy whether one was needed or not.  But maybe that's just premature optimization.

python/lsst/daf/butler/core/dimensions/coordinate.py
TallJimbo: This reminded me that @jonathansick pointed me at https://jira.lsstcorp.org/browse/DM-21228, which recommends a `docs/dev` directory.  I'll move this there and leave a pointer to that as a code comment.

python/lsst/daf/butler/core/dimensions/coordinate.py
TallJimbo: That's consistent with my thoughts as well.  If the demangling never needs a `Registry`, no problem.  If it does, we'll need to see if the layers that currently transform dictionaries into `DataCoordinates` already have a `Registry` to use for that, and if not, whether we should defer that transformation to a lower level or find a way to pass a `Registry` to them.

python/lsst/daf/butler/core/dimensions/coordinate.py
TallJimbo: How about I add a message to the `assert`?  This is an internal API, and I feel like checking declared preconditions is one of the things `assert` is for - any code that calls this without guaranteeing it (doing the checking of its own arguments if necessary) is broken.

timj: I guess I really dislike assert as a thing and I don't really understand what's wrong with:
```python
if len(graph.required) != len(values):
   raise AssertionError("Some message that tells me what len(a) was something and that this is a worry")
```
a message would always help though.

TallJimbo: That's fair; I actually prefer `assert` because it raises `AssertionError` and that's the intent I want to convey, not because it can be disabled.  I think we've discussed this before: https://lsstc.slack.com/archives/C2JPL8U78/p1576168318008800, where raising `AssertionError` directly was considered (at least by one person) confusing, and I think the consensus was that `assert` was reasonable.  _I_ don't find raising `AssertionError` confusing, but I'm probably not the best judge of that for my own code.

Basically, I think `assert` is a really important _concept_, and I'm reluctant to avoid the language construct for it just because we're worried it _could_ be disabled, especially since I don't think being able to disable this check is obviously wrong.  It certainly could appear in a tight loop, and while I very much doubt the check is expensive, using `assert` means I don't have to _care_ if it's expensive.

doc/lsst.daf.butler/dimensions.rst
TallJimbo: It builds fine for me without the `\ ` in all of my local `package-docs` builds, even though I know it's not supposed to.  I suppose I should add it anyway.  Don't know why I've never encountered that problem.

timj: I bet it's the non-alphabetic apostrophe character that is the difference.

python/lsst/daf/butler/core/dimensions/_dataCoordinateIterable.py
timj: maybe update the docstring then to say that the caller should not modify it

python/lsst/daf/butler/core/dimensions/coordinate.py
TallJimbo: I've added messages to all of the asserts in this module but am leaving these asserts for now.

PR #325, opened 2020-07-14
DM-25957: Remove unnecessary numpy usage

timj:
* Uses the python random number generator rather than numpy.
* Protects the one remaining numpy import

### DIFF
--- a/python/lsst/daf/butler/tests/_testRepo.py
+++ b/python/lsst/daf/butler/tests/_testRepo.py
@@ -22,9 +22,7 @@
 
 __all__ = ["makeTestRepo", "makeTestCollection", "addDatasetType", "expandUniqueId"]
 
-
-import numpy as np
-
+import random
 from lsst.daf.butler import Butler, Config, DatasetType
 
 
@@ -106,7 +104,7 @@ def makeTestCollection(repo):
     """
     # Create a "random" collection name
     # Speed matters more than cryptographic guarantees
-    collection = "test" + "".join((str(i) for i in np.random.randint(0, 10, size=8)))
+    collection = "test" + str(random.randrange(1_000_000_000))
     return Butler(butler=repo, run=collection)
 
 
--- a/tests/test_butler.py
+++ b/tests/test_butler.py
@@ -30,7 +30,6 @@ import shutil
 import pickle
 import string
 import random
-import numpy as np
 
 try:
     import boto3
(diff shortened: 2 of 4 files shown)


python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -106,7 +104,7 @@ def makeTestCollection(repo):
     """
     # Create a "random" collection name
     # Speed matters more than cryptographic guarantees
-    collection = "test" + "".join((str(i) for i in np.random.randint(0, 10, size=8)))
+    collection = "test" + "".join((str(i) for i in Random().sample(range(10), 8)))
```
ktlim: Um, isn't this changing a sample with replacement to a sample without replacement?   Given the comment, maybe you don't really care (but one may be faster than the other).

kfindeisen: I'm not sure it's a good idea to construct a new `Random` every time, since that might introduce weird correlations (e.g., slowly varying seeds). I suggest storing a persistent `Random` in the module, or (I think equivalently?) using the `sample` function.

kfindeisen: `sample` seems like a very roundabout way of doing this (as does the original method, to be honest). I think this would be much simpler with:
```suggestion
    collection = "test" + str(random.randrange(1_000_000_000))
```

timj: I think we just want 8 fairly random digits to use in the collection name without having to keep track of a counter internally. Maybe I'm better off doing a single random number of 8 digits and then zero-padding it in the collection name?

timj: Instead is there a function that returns 8 random characters? (like is used in tempfiles).

tests/test_butler.py
```diff
@@ -868,7 +871,9 @@ def testPutTemplates(self):
         butler.registry.registerDatasetType(DatasetType("metric2", dimensions, storageClass))
         butler.registry.registerDatasetType(DatasetType("metric3", dimensions, storageClass))
 
-        dataId1 = {"instrument": "DummyCamComp", "visit": np.int64(423)}
+        # Check that the dimension system can handle a numbers.Integral type
+        np_int = np.int64(423) if np else 423
+        dataId1 = {"instrument": "DummyCamComp", "visit": np_int}
```
kfindeisen: This is a bit different from the "skip if absent" strategy described in the commit message; in particular, somebody running this test without numpy will not realize there was a fallback. Why not use `@unittest.skipUnless` (with the added advantage that the current test will be factored into smaller pieces)?

kfindeisen (approved):
Looks good, some minor implementation suggestions.

python/lsst/daf/butler/tests/_testRepo.py
timj: I was also pondering this:
```
>>> tempfile.mktemp(dir="", prefix="test")
'testzp6hm_z8'
```
which seems good enough.

kfindeisen: That would work, but it's deprecated since Python 2.3... 😦

PR #326, opened 2020-07-15
Keep time precision when exporting to YAML (DM-25885)

andy-slac:
The time is now saved as ISO string in TAI scale but with full precision
and using special tag `!butler_time/tai/iso`.


timj (approved):
Looks great.

PR #327, opened 2020-07-16
DM-25691: Some documentation on storage classes and formatters

timj:


doc/lsst.daf.butler/datastores.rst
```diff
@@ -0,0 +1,79 @@
...
+The `~lsst.daf.butler.datastores.inMemoryDatastore.InMemoryDatastore` currently only supports the ``constraints`` field.
+This allows the datastore to accept specific dataset types.
+
+In the future more features will be added to allow some form of cache expiry.
+
+ChainedDatastore
+================
+
+This datastore enables multiple other datastores to be combined into one.
+The datastore will be sent to every datastore in the chain and success is reported if any of the datastores accepts the dataset.
```
ktlim, line 94:
```suggestion
The dataset will be sent to every datastore in the chain, and success is reported if any of the datastores accepts the dataset.
```

doc/lsst.daf.butler/datastores.rst
```diff
@@ -0,0 +1,79 @@
...
+This specifies the fully qualified class name of the Python class implementing the datastore.
+The default is to use the `~lsst.daf.butler.datastores.posixDatastore.PosixDatastore`.
+All other keys depend on the specific datastore class that is selected.
+
+The default configuration values can be inspected at ``$DAF_BUTLER_DIR/config`` and current values can be obtained by calling ``butler config-dump`` on a butler repository.
+
+File-Based Datastores
+=====================
+
+The file-based datastores (for example `~lsst.daf.butler.datastores.posixDatastore.PosixDatastore` and `~lsst.daf.butler.datastores.s3Datastore.S3Datastore`) share configuration since they use formatters to serialize datasets to file artifacts using file name template schemes and also support composite disassembly.
```
erykoff: I do not know what "composite disassembly" means.  Rather, I did not before reading the other documentation.  A link here would be handy!

doc/lsst.daf.butler/datastores.rst
```diff
@@ -0,0 +1,79 @@
...
+#######################
+Datastore Configuration
+#######################
+
+A Butler `~lsst.daf.butler.Datastore` is configured in the ``datastore`` section of the top-level Butler YAML configuration.
+The only mandatory entry in the datastore configuration is the ``cls`` key.
+This specifies the fully qualified class name of the Python class implementing the datastore.
+The default is to use the `~lsst.daf.butler.datastores.posixDatastore.PosixDatastore`.
+All other keys depend on the specific datastore class that is selected.
+
```
erykoff, line 14: It would be nice up here at the top in the intro to put a list of the supported datastores (either class names or brief text description) and a link to the details below.

doc/lsst.daf.butler/datastores.rst
```diff
@@ -0,0 +1,79 @@
...
+The default configuration values can be inspected at ``$DAF_BUTLER_DIR/config`` and current values can be obtained by calling ``butler config-dump`` on a butler repository.
+
+File-Based Datastores
+=====================
+
+The file-based datastores (for example `~lsst.daf.butler.datastores.posixDatastore.PosixDatastore` and `~lsst.daf.butler.datastores.s3Datastore.S3Datastore`) share configuration since they use formatters to serialize datasets to file artifacts using file name template schemes and also support composite disassembly.
+
+The supported configurations are:
+
```
erykoff, line 34: Should the following be a bulleted list?

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
...
+Each `~lsst.daf.butler.DatasetType` is associated with a `~lsst.daf.butler.StorageClass`.
+This storage class declares the Python type, any components it may have (read-only or read-write), and an assembler that can be used to process read parameters.
+
+Composites
+^^^^^^^^^^
+
+A composite storage class declares that the Python type consists of discrete components that can be accessed individually.
+Each of these components must declare its storage class as well.
+
+For example, if a ``pvi`` dataset type has been associated with an ``ExposureF`` storage class, the user can `~lsst.daf.butler.Butler.get()` the full ``pvi`` and access the components as they would normally for an `~lsst.afw.image.ExposureF`, or if the user solely want the metadata header from the exposure they can ask for ``pvi.metadata`` and just get that.
```
erykoff: Ah, so `ExposureF` is a "composite storage class"?

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
...
+
+Read Parameters
+^^^^^^^^^^^^^^^
+
+Read parameters are used to adjust what is returned by the `~lsst.daf.butler.Butler.get()` call but there is a requirement that whatever those read parameters do to modify the `~lsst.daf.butler.Butler.get()` the Python type returned must match the type associated with the `~lsst.daf.butler.Butler.StorageClass` associated with the `~lsst.daf.butler.Butler.DatasetType`.
+For example this means that a read parameter that subsets an image is valid because the type returned would still be an image.
+
+If read parameters are defined then a `~lsst.daf.butler.CompositeAssembler.handleParameters()` method must be defined that understands how to apply these parameters to the Python object and should return a modified copy.
+This method must be written even if a `~lsst.daf.butler.Formatter` is to be used.
+There are two reasons for this, firstly, there is no guarantee that a particular formatter implementation will understand the parameter (and no requirement for that to be the case), and secondly there is no guarantee that a formatter will be involved in retrieval of the dataset.
```
erykoff: Period or semi-colon after "reasons for this".

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
...
+For many file-based formatter implementations a subclass of `~lsst.daf.butler.Formatter` can be used that has a much simplified interface.
+`~lsst.daf.butler.formatters.file.FileFormatter` allows a formatter implementation to be written using two methods: `~lsst.daf.butler.formatters.file.FileFormatter._readFile()` takes a local path to the file system and the expected Python type, and `~lsst.daf.butler.formatters.file.FileFormatter._writeFile()` takes the in-memory object to be serialized.
+
+Composites are not handled by `~lsst.daf.butler.formatters.file.FileFormatter`.
+
+.. note::
+  I'm not sure I understand why _writeFile() doesn't also take the path rather than requiring FileDescriptor to be used.
+  It's inconsistent with _readFile that does take the local path.
+  It's not much of a simplification as things stand.
+  Need to revisit that.
```
erykoff: Should this be ticketed?

doc/lsst.daf.butler/datastores.rst
```diff
@@ -0,0 +1,79 @@
...
+Datastore Configuration
+#######################
+
+A Butler `~lsst.daf.butler.Datastore` is configured in the ``datastore`` section of the top-level Butler YAML configuration.
+The only mandatory entry in the datastore configuration is the ``cls`` key.
+This specifies the fully qualified class name of the Python class implementing the datastore.
+The default is to use the `~lsst.daf.butler.datastores.posixDatastore.PosixDatastore`.
+All other keys depend on the specific datastore class that is selected.
+
+The default configuration values can be inspected at ``$DAF_BUTLER_DIR/config`` and current values can be obtained by calling ``butler config-dump`` on a butler repository.
```
erykoff: Ah, this was the most useful line in the documentation for a "user" like me.  I think this should also be put into the formatter_assembler documentation as well.

doc/lsst.daf.butler/datastores.rst
```diff
@@ -0,0 +1,79 @@
...
+In the future more features will be added to allow some form of cache expiry.
+
+ChainedDatastore
+================
+
+This datastore enables multiple other datastores to be combined into one.
+The dataset will be sent to every datastore in the chain, and success is reported if any of the datastores accepts the dataset.
+When a dataset is retrieved each datastore is asked for the dataset in turn and the first match is sufficient.
+This allows an in-memory datastore to be combined with a file-based datastore to enable simple in-memory retrieval for a dataset that has been persisted to disk.
+
```
erykoff, line 99: Is this something that can be created on-the-fly, or is it something that you need to know you want ahead of time when you create a repository?

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
...
+Formatters and assemblers provide the interface between Butler and the python types it is storing and retrieving.
+A Formatter is responsible for serializing a Python type to an external storage system and reading that serialized form back into Python.
+The serialization can be stored to a local file system or cloud storage or even a database.
+It is possible for a formatter to be globally configured to use particular parameters on write.
+On retrieval of datasets read parameters can be used that can, for example, return only a subset of the data.
+
+Assemblers are used to disassemble and reassemble composite datasets and can also be used to process read parameters that adjust how the retrieved dataset might be modified on get.
+
+Deciding which formatter or assembler to use is controlled by the storage class and corresponding dataset type.
+
```
erykoff, line 18: As mentioned above, a note here on how to get the default and current configs would be very helpful.  Looking at that output made the description of what all the parameters mean a lot clearer to me (on top of the examples, seeing the full set of options was very illuminating).

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
...
+You can do this using YAML anchors and references but the preferred approach is to use the ``inheritsFrom`` key in the storage class definition:
+
+.. code-block:: yaml
+
+   MaskedImageI:
+     inheritsFrom: MaskedImage
+     pytype: lsst.afw.image.MaskedImageI
+     components:
+       image: ImageI
+       mask: MaskX
```
erykoff: Is there a reason this example doesn't match what's in the default config (which doesn't have the components)?

erykoff (changes requested):
There is a lot of good information here, and a lot to digest!  I think a little more introduction and summary up top would be warranted for those coming at this fresh, and a smattering of questions/suggestions throughout.

rgruendl:
I have read through the documentation update.  Overall it looks fine to me although I am sure there are many nuances that are basically lost on me (this is not a criticism just a statement to place my feedback in context).  One item that raised a question was in the section of File vs Bytes.  Is there any construct that limits the use of Bytes (besides common sense)?  I am basically asking because I suspect this is the place where many an unwary user can run into problems (that might affect others than just theirselves).

doc/lsst.daf.butler/datastores.rst
```diff
@@ -0,0 +1,79 @@
+.. _daf_butler-datastores:
+
+#######################
+Datastore Configuration
+#######################
+
+A Butler `~lsst.daf.butler.Datastore` is configured in the ``datastore`` section of the top-level Butler YAML configuration.
+The only mandatory entry in the datastore configuration is the ``cls`` key.
+This specifies the fully qualified class name of the Python class implementing the datastore.
+The default is to use the `~lsst.daf.butler.datastores.posixDatastore.PosixDatastore`.
```
andy-slac: So it is both mandatory and has a default value?

doc/lsst.daf.butler/datastores.rst
```diff
@@ -0,0 +1,79 @@
...
+    Generally the default setting will be usable although it can be tuned per `~lsst.daf.butler.DatasetType`, `~lsst.daf.butler.StorageClass` or data ID.
+formatters
+    Mapping of `~lsst.daf.butler.DatasetType`, `~lsst.daf.butler.StorageClass` or data ID to a specific formatter class that understands the associated Python type and will serialize it to a file artifact.
+    The formatters section also supports the definitions of write recipes (bulk configurations that can be selected for specific formatters) and write parameters (parameters that control how the dataset is serialized; note it is required that all serialized artifacts be readable by a formatter without knowing which write parameters were used).
+constraints
+    Specify `~lsst.daf.butler.DatasetType`, `~lsst.daf.butler.StorageClass` or data ID that will be accepted or rejected by this datastore.
+composites
+    Controls whether composite datasets are disassembled by the datastore.
+    By default composites are not disassembled.
+    Disassembly can be controlled by `~lsst.daf.butler.DatasetType`, `~lsst.daf.butler.StorageClass` or data ID.
```
andy-slac: Not quite clear what is the type of `composites`. Is it True/False or some sort of mapping?

doc/lsst.daf.butler/datastores.rst
```diff
@@ -0,0 +1,79 @@
...
+
+ChainedDatastore
+================
+
+This datastore enables multiple other datastores to be combined into one.
+The dataset will be sent to every datastore in the chain, and success is reported if any of the datastores accepts the dataset.
+When a dataset is retrieved each datastore is asked for the dataset in turn and the first match is sufficient.
+This allows an in-memory datastore to be combined with a file-based datastore to enable simple in-memory retrieval for a dataset that has been persisted to disk.
+
+`~lsst.daf.butler.datastores.chainedDatastore.ChainedDatastore` has a ``datastores`` key that contains a list of datastore configurations that can match the ``datastore`` contents from other datastores.
```
andy-slac: Not sure what "can match" means here?

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
+.. _daf_butler_storageclass_formatters_assemblers:
+
+##########################################
+Storage Classes, Assemblers and Formatters
+##########################################
+
+Formatters and assemblers provide the interface between Butler and the python types it is storing and retrieving.
```
andy-slac: Capitalize Python for consistency.

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
...
+
+Assemblers are used to disassemble and reassemble composite datasets and can also be used to process read parameters that adjust how the retrieved dataset might be modified on get.
+
+Deciding which formatter or assembler to use is controlled by the storage class and corresponding dataset type.
+
+Storage Classes
+===============
+
+A Storage Class is fundamental to informing Butler how to deal with specific Python types.
+Each `~lsst.daf.butler.DatasetType` is associated with a `~lsst.daf.butler.StorageClass`.
```
andy-slac: Maybe mention where this association happens?

timj:
@rgruendl the files vs bytes area is still a bit unexplored. If a formatter writer only ever reads and writes files everything is fine. The bytes part is an optimization for S3 (and related) where the local form of the file is transient and the bytes have to be uploaded somewhere else. As I say in a note, we will need to finesse this a little.

doc/lsst.daf.butler/formatters.rst
timj: Because I only wanted to show the essence of an example and not show a fully worked real life example (It probably doesn't have components because we probably don't use them anywhere).

doc/lsst.daf.butler/datastores.rst
timj: You can turn a single datastore entry into a chained datastore without any problem. Things go wrong if you replace it with something that doesn't have any idea where the previous data are stored. In theory you could upload all the data locally to S3 and switch from a posix datastore to s3datastore without a problem. You can change a posix to a chained posix+in-memory or posix+s3+in-memory. For data that are only in the original posix datastore the data will come from that datastore. New writes will go to the old posix plus the new s3 say.

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
...
+For example, if a ``pvi`` dataset type has been associated with an ``ExposureF`` storage class, the user can `~lsst.daf.butler.Butler.get()` the full ``pvi`` and access the components as they would normally for an `~lsst.afw.image.ExposureF`, or if the user solely want the metadata header from the exposure they can ask for ``pvi.metadata`` and just get that.
+The implementation details of how that metadata is retrieved depend on the details of how the dataset was serialized within the datastore.
+
+Composites must declare **all** the components for the Python type and Butler requires that if a composite is dissassembled into its components and then reassembled to form the composite again, this operation must be lossless.
+
+Only composites have the potential to be disassembled into discrete file artifacts by datastore.
+Disassembly itself is controlled by the datastore configuration.
+In cases where the datastore uses a remote storage and only some components are required, then disassembly can significantly improve data access performance.
+
+Read-only Components
```
andy-slac: `read-only` may be confusing (I think everything is read-only once it is stored in butler?) I'm not sure this naming was discussed, I probably missed it. Could `computed` or something similar be an option still?

doc/lsst.daf.butler/datastores.rst
timj: I probably should also say somewhere that a key design of datastore is that once you have written it with a specific formatter, you can change your configuration to use different file templates or associate different formatters with that dataset type and you can still read the data. Only new writes will be affected.

doc/lsst.daf.butler/formatters.rst
timj: Step 1 was getting feedback from this review...

doc/lsst.daf.butler/formatters.rst
timj: Yes.

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
...
+^^^^^^^^^^^^^^^^^^^^
+
+Just as for components of a composite, if a storage class defines read-only components, it must also specify an assembler to support the calculation of that derived component.
+This should be implemented in the `~lsst.daf.butler.CompositeAssembler.getComponent()` method.
+
+Additionally, if the storage class refers to a composite, the datastore can be configured to disassemble the dataset into discrete artifacts.
+Since read-only components are derived and are not persisted themselves the datastore needs to be told which component should be used to calculate this derived quantity.
+To enable this the assembler must implement `~lsst.daf.butler.CompositeAssembler.selectResponsibleComponent()`.
+This method is given the name of the read-only component and a list of all available persisted components and must return one and only one relevant component.
```
andy-slac: Should this kind of knowledge belong to configuration instead?

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
...
+.. warning::
+
+  The formatter system has only been used to write datasets to files or to bytes that would be written to a file.
+  The interface may evolve as other types of datastore become available and make use of the formatter system.
+
+When ingesting files from external sources formatters are associated with each incoming file but these formatters are only required to support a `~lsst.daf.butler.Formatter.read()` method.
+They must though declare all the file extensions that they can support.
+This allows the datastore to ensure that the image being ingested has not obviously been associated with a formatter that does not recognize it.
+
+In the current implementation that is focussed entirely on external files in datastores, the location of the serialized data is available to the formatter using the `~lsst.daf.butler.Formatter.fileDescriptor` property.
```
andy-slac: focussed  -- we spell is as "focused" in Russia 😉

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
...
+This `~lsst.daf.butler.FileDescriptor` property makes the file location available as a `~lsst.daf.butler.Location` and also gives access to read parameters supplied by the caller and also defines the `~lsst.daf.butler.StorageClass` of the dataset being written.
```
andy-slac: and also ... and also ...

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
...
+On read the the storage class used to read the file can be different from the storage class expected to be returned by `~lsst.daf.butler.Datastore`.
```
andy-slac: the the

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
...
+This happens if a composite was written but a component from that composite is being read.
+
+File Extensions
+^^^^^^^^^^^^^^^
+
+Each formatter that reads or writes a file must declare the file extensions that it supports.
+For a formatter that supports a single extension this is most easily achieved by setting the class property `~lsst.daf.butler.Formatter.extension` to that extension.
+In some scenarios a formatter might support multiple formats that are controlled by write parameters.
+In this case the formatter should assign a frozen set to the `~lsst.daf.butler.Formatter.supportedExtensions` class property.
+It is then required that the class implement an instance property for ``extension`` that returns the extension that will be used by this formatter for writing the current dataset.
```
andy-slac: implement -> implements?

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
...
+
+File vs Bytes
+^^^^^^^^^^^^^
+
+Some datastores can stream bytes from remote storage systems and do not require that a local file is created before the Python object can be created.
+To support this use case an implementer can implement `~lsst.daf.butler.Formatter.fromBytes()` for reading in from a datastore and `~lsst.daf.butler.Formatter.toBytes()` for serializing to a datastore.
+If a formatter raises `NotImplementedError` when these byte-like methods are called the datastore will default to using the `~lsst.daf.butler.Formatter.read()` and `~lsst.daf.butler.Formatter.write()` methods making use of local temporary files.
```
andy-slac: byte-like -> byte-oriented (or something else)?

andy-slac (commented):
Not sure I understood much of it but I left few comments.

doc/lsst.daf.butler/formatters.rst
timj: I'm not happy with `read-only` but I merged that because we couldn't think of anything better. "computed component" or "derived component" do seem better.

doc/lsst.daf.butler/datastores.rst
timj: I'm trying to say that datastores is a list of datastore configurations that are the same as a configuration you would use for that datastore if it wasn't in a chain...

doc/lsst.daf.butler/datastores.rst
timj: This is what the default looks like (it's really there as an example):
```yaml
# True: concrete composite (composites are disassembled);
# False: virtual composite (composites are written monolithically).
# default applies to all composites that are not explicitly mentioned
# elsewhere.
default: false
# Use a dict rather than a list to allow easy merging of user config
# into default config. Config class has no way of merging lists.
# Types can be StorageClass names or DatasetType names, with DatasetType
# name always taking preference over StorageClass name.
disassembled:
  ExposureI: false
  ExposureF: false
  calexp: false
```

doc/lsst.daf.butler/datastores.rst
timj: Hmm. Yes. In the sense that in `python/lsst/daf/butler/configs/datastore.yaml` we have it listed. It's mandatory in the sense that it must be in the butler configuration and defaulted in the sense that we put a default in the default yaml... both statements are correct.

doc/lsst.daf.butler/formatters.rst
timj: I put it in the assembler so that the assembler can make a dynamic choice based on the available components. I also don't know how common composite disassembly is going to become so it wasn't clear to me that I really needed more complexity in the storage class configuration. The assembler is the thing that has to calculate the "computed component" so it seemed reasonable to expect the assembler to know which component it needs to use for that.

doc/lsst.daf.butler/formatters.rst
timj: In British English we spell it "focussed" but it seems that I should change it to conform.

doc/lsst.daf.butler/datastores.rst
andy-slac: It was confusing me, maybe it would help to say something more explicit like "Default butler configuration uses `...PosixDatastore`"?

doc/lsst.daf.butler/datastores.rst
andy-slac: I think if you drop "that can match ..." it will be clearer for me.

doc/lsst.daf.butler/datastores.rst
```diff
@@ -0,0 +1,79 @@
...
+
+root
+    The location of the "root" of the datastore "file system".
+    Usually the default value of ``<butlerRoot>/datastore`` can be left unchanged.
+    Here ``<butlerRoot>`` is a magic value that is replaced either with the location of the Butler configuration file or the top-level ``root`` as set in that ``butler.yaml`` configuration file.
+records
+    This sections defines the name of the registry table that should be used to hold details about datasets stored in the datastore (such as the path within the datastore and the associated formatter).
+    This only needs to be set if multiple datastores are to be used simultaneously within one Butler repository since the table names should not clash.
+create
+    A Boolean to define whether an attempt should be made to initialize the datastore by creating the directory.  Defaults to `True`.
```
TallJimbo: This doesn't feel like it belongs in configuration, and I think @andy-slac may be doing related work on `Registry` to make sure we only create tables in the database when explicitly requested, and fail hard if someone asks to create tables when the database is not empty.  We may want similar logic for datastores, especially if we already have a butler-level boolean we can pass down.

doc/lsst.daf.butler/datastores.rst
```diff
@@ -0,0 +1,79 @@
...
+templates
+    The template to use to construct "files" within the datastore.
+    The template uses data dimensions to do this.
+    Generally the default setting will be usable although it can be tuned per `~lsst.daf.butler.DatasetType`, `~lsst.daf.butler.StorageClass` or data ID.
```
TallJimbo: This probably merits a link to other docs, like maybe the `FileTemplate` class?

doc/lsst.daf.butler/datastores.rst
```diff
@@ -0,0 +1,79 @@
+.. _daf_butler-datastores:
+
```
TallJimbo, line 2: Putting a
```
.. py:currentmodule:: lsst.daf.butler
```
here would let you drop a lot of `~lsst.daf.butler` below.

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
...
+.. code-block:: yaml
+
+   MaskedImageI:
+     inheritsFrom: MaskedImage
+     pytype: lsst.afw.image.MaskedImageI
+     components:
+       image: ImageI
+       mask: MaskX
+
+If this approach is used the `~lsst.daf.butler.StorageClass` Python class created by `~lsst.daf.butler.StorageClassFactory` will inherit from the specific parent class and not the generic `~lsst.daf.butler.StorageClass`.
```
TallJimbo: `StorageClass` is a regular type, not a metaclass, so its instances are just regular instances, and can't inherit in a Python-type-system sense.  I suspect that's not what you meant to imply, but I think that's how it reads.

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
...
+
+Assemblers
+==========
+
+.. note::
+
+  The base class is called CompositeAssembler because it was first developed for composites.
+  Now might be a good time to rebrand it since non-composites need one if they use read parameters.
```
TallJimbo: This makes me wonder if maybe `StorageClass` _should_ be a metaclass, and `CompositeAssemblers` could just become custom `StorageClass` subclasses (e.g. we'd give the storage class yaml definition a `cls` key like so many others, and if it's left blank we'd dynamically create one or just use the `StorageClass` base class).  That may be unnecessarily meta (and create new problems), but it nicely solves the problem that both `StorageClass` and `CompositeAssembler` feel now like they're competing for the role of "daf_butler object that represents some Python type we can store", as there's less of a clear boundary than there once was.

A much simpler solution with less "elegance potential" might be to just rename `CompositeAssembler` something like `StorageClassCustomizer`.

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
...
+
+For read-only components parameters are handled by the composite component prior to deriving the read-only component.
+The assembler `~lsst.daf.butler.CompositeAssembler.handleParameters()` method will only be called in this situation if no formatter is used (such as with an in-memory datastore).
+
+Formatters
+==========
+
+Formatters are responsible for serializing a Python type to a storage system and for reconstructing the Python type from the serialized form.
+A formatter has to implement at minimum a `~lsst.daf.butler.Formatter.read()` method and a `~lsst.daf.butler.Formatter.write()` method.
+The ``write()`` method takes a Python object and serializes it somewhere and the ``read()`` method is optionally given a component name and returns the matching Python object.
```
TallJimbo, line 200: This is bigger than docs, of course, but I have always been a bit uncomfortable with `Formatter` as a single class _with state_ being responsible for both reading and writing.  The base class state (a `FileDescriptor` and a data ID) is totally reasonable for both cases, but any subclass-specific state has got to be different between read-mode and write-mode.  I don't know if there's any concrete harm (other than just potential harm) from having a dual state object like that, and it's undeniably useful to have one name to refer to both, but it just doesn't feel right.

doc/lsst.daf.butler/formatters.rst
```diff
@@ -0,0 +1,316 @@
...
+Details of where the artifact may be located within the datastore are passed to the constructor by the datastore as a `~lsst.daf.butler.FileDescriptor` instance.
+
+.. warning::
+
+  The formatter system has only been used to write datasets to files or to bytes that would be written to a file.
+  The interface may evolve as other types of datastore become available and make use of the formatter system.
```
TallJimbo, line 206: FWIW, I had actually imagined that `Formatters` would simply not be used by some datastores, or that they would have an analogous concept with a totally different interface.

doc/lsst.daf.butler/formatters.rst
TallJimbo: I do think some work here is worthwhile.  Some related considerations:
 - Some prominent file-based formatters (`FiitsExposureFormatter` in obs_base, at least) can't use `FileFormatter` and have to inherit from `Formatter` directly.  I don't recall why, or even if it's still a problem.
 - I don't see why `FileFormatter` (rather than base `Formatter`, or even code that calls it) is responsible for updating file extensions in `write`.  I also have hard time keeping track of whether one is supposed to modify `self.fileDescriptor` in place, or whether that might cause bad side-effects, and I'd really like to keep `Formatter` authors from having to deal with any of that.

It'd certainly be much easier to write `Formatters` if there was a simplified interface that _only_ dealt with local file paths, took care of extensions, and didn't require the author to learn about `FileDescriptor` at all.

TallJimbo (approved):
Looks good!  I didn't do a super close read (I figure the readers who are newer to this stuff will be better judges of pedagogy anyway), but I left a few comments mostly on design issues the docs raised, since I gather that was one of your goals here as well.

doc/lsst.daf.butler/datastores.rst
timj: Ok, although the default configs have now moved and are package resources...

doc/lsst.daf.butler/datastores.rst
timj: I thought a description list was more accurate for what this is. Granted there is no real formatting there that tells you it is a description list. Maybe I can boldify the headings.

doc/lsst.daf.butler/datastores.rst
timj: ok. For now it exists so I probably should leave it in with a caveat.

doc/lsst.daf.butler/formatters.rst
timj: I probably need to change the storageClasses.yaml to also say derived component.

doc/lsst.daf.butler/formatters.rst
timj: Yes, early on we planned to experiment with different styles of formatters and so I always assumed that Formatter was a generic concept. That's why the base class is not quite right for files. The lack of prototyping in this area has caused this problem. Cleaning up the base Formatter to be an actually FileFormatter is something we can look at but should not be on this ticket.

doc/lsst.daf.butler/formatters.rst
timj: If we start having different formatter classes for reading and writing isn't that going to be really easy to get out of sync? You also end up with having to track twice as many classes and have to have separate entries in the config for them.

doc/lsst.daf.butler/formatters.rst
timj: You are not wrong. Given the way things have evolved it's clear that we are associating one storage class with one python type and that we optionally associate one assembler that is specific to that python type. The storage class yaml already has a field in it for associating a specific "assembler" (fully-qualified class name) with the storage class (and for now we only have one of them and it's attached to three storage classes: Exposure, Image, Mask).

For this ticket I think renaming to StorageClassCustomizer is a good idea.

doc/lsst.daf.butler/formatters.rst
timj: I'll create a ticket for this. The urgency for it partly depends on how soon other people are going to be writing their own formatters.

timj: DM-26658

doc/lsst.daf.butler/formatters.rst
timj: I think I'm implying that they are subclasses. If you use inheritsFrom the storage class you get back is a subclass. For example the ImageI storage class has this inheritance:

```
 |  Method resolution order:
 |      StorageClassImageI
 |      StorageClassImage
 |      StorageClass
 |      builtins.object
```
if you create a StorageClass directly then it is not a subclass, but StorageClassFactory works a bit harder than that.

erykoff (approved):
Looks good to me.

PR #328, opened 2020-07-17
DM-25468: Add protection for repeated schema initialization

andy-slac:
Adds a check to declareStaticTables() that requires empty schema if
create argument is True. Needed workaround for a workaround in the
sqlite override of that method.


TallJimbo (approved):
Looks great!  Thanks for taking care of this very dangerous bad `drop_all` behavior!

PR #329, opened 2020-07-18
DM-26010: Move mypy execution to root directory, and run it on a directory.

TallJimbo:
This is the more common and better-supported way to run mypy, but it was previously blocked by the fact that our root (lsst) package was a namespace package.  This change should make no external difference in what mypy does right now, but I suspect it will be easier to maintain and I can already say it's much more compatible with at least some in-editor mypy integration tooling.

### DIFF
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -20,5 +20,4 @@ jobs:
         run: pip install mypy
 
       - name: Change to source directory and run mypy
-        working-directory: ./python
-        run: mypy lsst
+        run: mypy python/lsst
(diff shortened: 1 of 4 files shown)


timj (approved):
Thanks for this. Setuptools requiring the `__init__.py` was the final straw. I do wonder whether we are simply misunderstanding namespace packages or too many tools don't yet know what to do with them.

TallJimbo:
> I do wonder whether we are simply misunderstanding namespace packages or too many tools don't yet know what to do with them.

My sense is that they are just a flawed bit of functionality - a good idea whose implementation just leaves it too hard for tools to identify when a directory is a Python package and when it isn't.

PR #330, opened 2020-07-19
DM-25919: custom classes and new functionality for query results

TallJimbo:


python/lsst/daf/butler/registry/queries/_query.py
```diff
@@ -142,44 +282,334 @@ def extractDataId(self, row: RowProxy, *, graph: Optional[DimensionGraph] = None
         graph : `DimensionGraph`, optional
             The dimensions the returned data ID should identify.  If not
             provided, this will be all dimensions in `QuerySummary.requested`.
+        records : `Mapping` [ `str`, `Mapping` [ `tuple`, `DimensionRecord` ] ]
+            Records to use to return an `ExpandedDataCoordinate`.  If provided,
```
natelust: This wording is awkward

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -1317,6 +1317,54 @@ def queryDataIds(self, dimensions: Union[Iterable[Union[Dimension, str]], Dimens
...
+                              datasets: Any = None,
+                              collections: Any = None,
+                              where: Optional[str] = None,
+                              components: Optional[bool] = None,
+                              **kwargs: Any) -> Iterator[DimensionRecord]:
+        if not isinstance(element, DimensionElement):
+            element = self.dimensions[element]
+        dataIds = self.queryDataIds(element.graph, dataId=dataId, datasets=datasets, collections=collections,
+                                    where=where, components=components, **kwargs)
+        """Query for dimension information matching user-provided criteria.
```
natelust, line 1365: why is the doc-string after some code

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -743,6 +745,69 @@ def getExistingTable(self, name: str, spec: ddl.TableSpec) -> Optional[sqlalchem
...
+            name = f"tmp_{uuid.uuid4().hex}"
+        table = self._convertTableSpec(name, spec, self._metadata, prefixes=['TEMPORARY'],
+                                       schema=sqlalchemy.schema.BLANK_SCHEMA)
+        if table.key in self._tempTables:
+            if table.key != name:
+                raise ValueError(f"A temporary table with name {name} (transformed to {table.key} by "
+                                 f"Database) already exists.")
+        for foreignKeySpec in spec.foreignKeys:
+            table.append_constraint(self._convertForeignKeySpec(name, foreignKeySpec, self._metadata))
+        table.create(self._connection)
```
natelust, line 810: Random question can this allow a temp table to be created that shadows a non temp table? I am not sure what each db will do. If It does then depending on how databases handle this, the dropTemporaryTable may cause a dropped table. I find this unlikely behavior, but its worth verifying

python/lsst/daf/butler/registry/queries/_query.py
```diff
@@ -389,6 +426,38 @@ def _makeSubsetQueryColumns(self, *, graph: Optional[DimensionGraph] = None,
             columns.datasets = self.getDatasetColumns()
         return graph, columns
 
+    @contextmanager
+    def materialize(self, db: Database) -> Iterator[MaterializedQuery]:
```
natelust: I think the return type is a context manager

python/lsst/daf/butler/registry/queries/_query.py
```diff
@@ -613,3 +682,112 @@ def makeBuilder(self, summary: Optional[QuerySummary] = None) -> QueryBuilder:
...
+                 graph: DimensionGraph,
+                 whereRegion: Optional[Region],
+                 managers: RegistryManagers):
+        super().__init__(graph=graph, whereRegion=whereRegion, managers=managers)
+        self._table = table
+        self._spatial = tuple(spatial)
+        self._datasetType = datasetType
+        self._isUnique = isUnique
+
+    def isUnique(self) -> bool:
```
natelust, line 756: Why is this a function? If its to future proof things, it should be fine to just expose isUnique directly as an attribute to access, and it can be changed later if need be. If it is supposed to keep the value from being set from the outside, I think you probably meant to make this a property?

python/lsst/daf/butler/registry/queries/_results.py
```diff
@@ -59,6 +69,10 @@ class DataCoordinateQueryResults(DataCoordinateIterable):
         Mapping containing `DimensionRecord` objects for all dimensions and
         all data IDs this query will yield.  If `None` (default),
         `DataCoordinateIterable.hasRecords` will return `False`.
+        This is a nested mapping with `str` names of dimension elements as
+        outer keys, `DimensionRecord` instances as inner values, and
+        ``tuple(record.dataId.values())`` for the inner keys / outer values
+        (where ``record`` is the innermost `DimensionRecord` instance).
```
natelust: This documentation is not clear. Is it Dict[str, Dict[DimensionRecord, tuple(...)]]

python/lsst/daf/butler/registry/queries/_results.py
```diff
@@ -235,3 +249,193 @@ def constrain(self, query: SimpleQuery, columns: Callable[[str], sqlalchemy.sql.
                 for dimension in self.graph.required
             ])
         )
+
+
+class DatasetQueryResults(Iterable[DatasetRef]):
```
natelust, line 313: I dont think I had seen typing used as inheritance in this way

python/lsst/daf/butler/registry/queries/_results.py
```diff
@@ -235,3 +249,193 @@ def constrain(self, query: SimpleQuery, columns: Callable[[str], sqlalchemy.sql.
...
+        self._query = query
+        self._components = components
+        self._records = records
+        assert query.datasetType is not None, \
+            "Query used to initialize dataset results must have a dataset."
+        assert query.datasetType.dimensions == query.graph
+
+    __slots__ = ("_db", "_query", "_dimensions", "_components", "_records")
+
+    def __iter__(self) -> Iterator[DatasetRef]:
```
natelust, line 403: This is just open discussion, but would it be better to use typing.Generator here? I personally think it helps identify in a signature that the results will be single use. I dont know if analysis tools are or are not smart enough for that to happen automatically though

python/lsst/daf/butler/registry/queries/_results.py
```diff
@@ -235,3 +249,193 @@ def constrain(self, query: SimpleQuery, columns: Callable[[str], sqlalchemy.sql.
...
+        for row in self._query.rows(self._db):
+            parentRef = self._query.extractDatasetRef(row, records=self._records)
+            for component in self._components:
+                if component is None:
+                    yield parentRef
+                else:
+                    yield parentRef.makeComponentRef(component)
+
+    def byParentDatasetType(self) -> Iterator[ParentDatasetQueryResults]:
```
natelust, line 412: same as iter

python/lsst/daf/butler/core/dimensions/records.py
```diff
@@ -172,6 +172,14 @@ def fromDict(cls, mapping: Mapping[str, Any]) -> DimensionRecord:
...
         return cls(*values)
 
+    def __eq__(self, other: Any) -> bool:
+        if type(other) != type(self):
+            return False
+        return self.dataId == other.dataId
+
+    def __hash__(self) -> int:
+        return hash(self.dataId)
+
```
natelust, line 182: Being pedantic I think this should be 2 commits. I dont know if you are doing any squashing before merging, but just keep it in mind if you do any rebases. It is not high enough priority that I would insist on it on its own.

python/lsst/daf/butler/registry/queries/_builder.py
```diff
@@ -151,45 +138,115 @@ def joinDataset(self, datasetType: DatasetType, collections: Any, *,
...
-                rank=subquery.columns["rank"] if addRank else None
+                runKey=subquery.columns[runKeyName],
             )
+        else:
+            subquery = subquery.alias(datasetType.name)
+        self.joinTable(subquery, datasetType.dimensions.required, datasets=columns)
         return True
 
-    def joinTable(self, table: FromClause, dimensions: NamedValueSet[Dimension]) -> None:
+    def joinTable(self, table: sqlalchemy.sql.FromClause, dimensions: NamedValueSet[Dimension], *,
```
natelust, line 244: I think this method needs to verify the dimensions passed in, can contained inside self.dimensions, and either reject if they are not, or somehow do the query to update self.dimensions so that the join can happen.

natelust (approved):
I think there were a few points I found to think about, but nothing too major. Overall it looks good, but I over no promises that I didn't miss some small logic bug related to functionality calls that already exist :-)

python/lsst/daf/butler/registry/_registry.py
TallJimbo: :facepalm:

python/lsst/daf/butler/registry/interfaces/_database.py
TallJimbo: I'm using `uuid` to create a random name, so clashes should be practically (if not quite theoretically) impossible.

python/lsst/daf/butler/registry/queries/_query.py
TallJimbo: I wish it was; that'd be clearer to read.  But MyPy wants me to annotate the undecorated method, which it then transforms by knowing what the decorator does.  I suppose that's the only behavior that would actually make sense for decorators in general, but it is particularly unfortunate for this one, where the return type is so different.

python/lsst/daf/butler/registry/queries/_query.py
TallJimbo: It's a naming convention thing.  Because we almost always name methods starting with verbs, I don't like to start non-methods with verbs unless it's completely ambiguous that they aren't methods.  I don't know about everyone else, but this helps me remember whether or not I should add a `()` to any particular interface.

python/lsst/daf/butler/registry/queries/_results.py
TallJimbo: I don't know about the tooling, but I think `Iterator` is also single-pass (all you can do with it is call `next()`, after all).  It's `Iterable` that's unfortunately ambiguous, and I don't have a good solution for that.

python/lsst/daf/butler/registry/queries/_query.py
natelust: in that case, does it need to be called `isUnique` or would `unique` as an attribute (or property) do the same thing? Does `is` add anything?

python/lsst/daf/butler/registry/queries/_builder.py
TallJimbo: I took a closer look at this, and it seems I need to add checks to _both_ this method and `Query.makeBuilder` to make sure we only attempt to build the queries we can build correctly.

python/lsst/daf/butler/registry/queries/_query.py
TallJimbo: I think it's a bit clearer.  Just "unique" in only a slightly different context could be an attribute that is a container of things that are unique, rather than a bool, while "isUnique" is very clearly a bool, and I think a set of parens is a reasonable price to pay for that.

natelust: sounds reasonable

TallJimbo:
> Second line does not need to be an f string, same in future strings in the makeBuilder functions

For posterity, since we discussed this OOB: I prefer this style for readability, and as a way to avoid forgetting the "f" when re-wrapping.  @natelust notes that it does come with a performance penalty, but in formatting an exception message that's not really a concern (especially these, which represent logic errors, not something someone would catch).

timj:
When we update flake8 you will get a warning if your f-string has no variables to format.

TallJimbo:
> When we update flake8 you will get a warning if your f-string has no variables to format.

Even if it's whitespace-concatenated with an `f`-string that does has variables to format?  If so, I think we should disable that check.

timj:
Yes, even if it's a continuation of something that did have something to format. I disagree with you on whether it should be disabled.

TallJimbo:
> Yes, even if it's a continuation of something that did have something to format. I disagree with you on whether it should be disabled.

Ok, not worth me fighting this, then.  I'm going to leave this branch as it is because it's already through Jenkins and this is far from the only case where this occurs in daf_butler, and we'll cross that bridge when we get to it.

PR #331, opened 2020-07-19
DM-23276: Add repr for database connection objects

timj:
Reports the connection URL rather than trying to reproduce
the exact caller parameters for the object instantiation.

### DIFF
--- a/python/lsst/daf/butler/registry/interfaces/_database.py
+++ b/python/lsst/daf/butler/registry/interfaces/_database.py
@@ -208,6 +208,15 @@ class Database(ABC):
         self._connection = connection
         self._metadata: Optional[sqlalchemy.schema.MetaData] = None
 
+    def __repr__(self) -> str:
+        # Rather than try to reproduce all the parameters used to create
+        # the object, instead report the more useful information of the
+        # connection URL.
+        uri = str(self._connection.engine.url)
+        if self.namespace:
+            uri += f"#{self.namespace}"
+        return f'{type(self).__name__}("{uri}")'
+
     @classmethod
     def makeDefaultUri(cls, root: str) -> Optional[str]:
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -208,6 +208,15 @@ def __init__(self, *, origin: int, connection: sqlalchemy.engine.Connection,
...
         self._metadata: Optional[sqlalchemy.schema.MetaData] = None
 
+    def __repr__(self) -> str:
+        # Rather than try to reproduce all the parameters used to create
+        # the object, instead report the more useful information of the
+        # connection URL.
+        uri = str(self._connection.engine.url)
+        if self.namespace:
+            uri = uri + f"#{self.namespace}"
+        return type(self).__name__ + f"(\"{uri}\")"
```
TallJimbo: It's a little strange seeing an f-string and a `+` string concatenation in the same line.  Why not use the f-string to do the concatenation?

TallJimbo (approved):

python/lsst/daf/butler/registry/interfaces/_database.py
timj: Mostly because I balked at having so much text in the curly braces. I can switch it to:
```python
f"{type(self).__name__}(\"{uri}\")"
```

TallJimbo: I'm really bikeshedding now, so please feel free to just merge whenever you'd like, but this seems like a great place to use single quotes somewhere:
```
f"{type(self).__name__}('{uri}')"
```

timj: Yes. It did cross my mind 😉

PR #332, opened 2020-07-21
Add boto3 retries, as well as higher-level backoff retries

domZippilli:
Primary purpose of this PR is to add low-level retries to the S3Datastore module. This is done in two ways:

- Botocore is imported, and a botocore.config.Config object is passed into the boto3.client() method. This allows for retries built-in to [boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html#available-retry-modes).

- The more general-purpose [`backoff`](https://pypi.org/project/backoff/) module is now used to implement retries at the level of datastore operations generally. These are higher-cost retries, but should only be attempted in the event the boto3 retry code either (a) gives up or (b) doesn't retry at all. It doesn't appear trivial to distinguish between the two cases.
  - In case (a), there will be some risk of excessive retries, though the `backoff` parameters allow only 60 seconds for retries overall, so there is low risk of jobs hanging on unrecoverable errors.
  - In case (b), we need `backoff` so that some kind of retry is attempted at all. 

All other changes are minutia.

### DIFF
--- a/python/lsst/daf/butler/datastores/s3Datastore.py
+++ b/python/lsst/daf/butler/datastores/s3Datastore.py
@@ -40,10 +40,23 @@ from typing import (
     Optional,
     Type,
     Union,
+    Callable
 )
 
 # https://pypi.org/project/backoff/
-import backoff
+try:
+    import backoff
+except ImportError:
+    class Backoff():
+        @staticmethod
+        def expo() -> None:
+            return None
+
+        @staticmethod
+        def on_exception(func: Callable, *args: Any, **kwargs: Any) -> Callable:
+            return func
+
+    backoff = Backoff
 
 from lsst.daf.butler import (
     ButlerURI,
--- a/python/lsst/daf/butler/datastores/s3Datastore.py
+++ b/python/lsst/daf/butler/datastores/s3Datastore.py
@@ -63,22 +63,24 @@ if TYPE_CHECKING:
 
 log = logging.getLogger(__name__)
 
-# settings for "backoff" retry decorators. these retries are belt-and-suspenders along
-# with the retries built into Boto3, to account for semantic differences in errors between
-# S3-like providers.
+# settings for "backoff" retry decorators. these retries are belt-and-
+# suspenders along with the retries built into Boto3, to account for
+# semantic differences in errors between S3-like providers.
 retryable_io_errors = (
     # http.client
-    ImproperConnectionState, HTTPException, 
+    ImproperConnectionState, HTTPException,
     # urllib3.exceptions
-    RequestError, HTTPError, 
+    RequestError, HTTPError,
     # built-ins
     TimeoutError, ConnectionError)
 retryable_client_errors = (
     # botocore.exceptions
-    ClientError, 
+    ClientError,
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -5,12 +5,15 @@ sqlalchemy >= 1.3
 https://github.com/lsst/sphgeom/archive/master.tar.gz
 
 # optional
+backoff >= 1.10.0
```
ktlim: Do you need the specifications to be this precise? Typically minor version is sufficient, no?

ktlim (approved):

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -38,6 +42,9 @@
     Union,
 )
 
+# https://pypi.org/project/backoff/
+import backoff
```
ktlim: Do we want to try to protect against backoff not being present or just assume that it is?

requirements.txt
domZippilli: Changed to minor version.

python/lsst/daf/butler/datastores/s3Datastore.py
domZippilli: Added a no-op decorator for that protection.

requirements.txt
```diff
@@ -5,12 +5,15 @@ sqlalchemy >= 1.3
...
 # optional
+backoff >= 1.10
 boto3 >= 1.13
+botocore >= 1.15
 moto >= 1.3
 pandas >= 1.0
 numpy >= 1.17
 matplotlib >= 3.0.3
 pyarrow >= 0.16
+urllib3 >= 1.25
```
timj, line 16: I think urllib3 is part of python core so I'm not sure this is needed is it?

python/lsst/daf/butler/datastores/s3Datastore.py
```diff
@@ -30,14 +30,26 @@
...
     Type,
     Union,
+    Callable
 )
 
+# https://pypi.org/project/backoff/
+try:
+    import backoff
+except ImportError:
+    def backoff(func: Callable, *args: Any, **kwargs: Any) -> Callable:
```
timj: This doesn't quite work because the decorator is actually `backoff.on_exception`.

timj (changes requested):
Looks good. Thank you. One thing needs fixing though.

python/lsst/daf/butler/datastores/s3Datastore.py
domZippilli: oh yes, duh.

domZippilli: Revised it to wrap this function in a class, PTAL

requirements.txt
domZippilli: It's not. I think it's a dependency of boto3 so we had it in the build before, which is why we were getting it in stack traces.

timj: Oh yes. Sorry about that.

timj (approved):

PR #335, opened 2020-07-23
DM-25985: Add saving and checking of per-manager versions

andy-slac:
For each manager we now store its full name, version number, and schema
digest. Some control for that has been moved to a separate interface
which needs to be implemented by each manager class. Saved manager name
could be different from one specified in registry config due to
difference in module names.

### DIFF
--- a/python/lsst/daf/butler/registry/interfaces/_versioning.py
+++ b/python/lsst/daf/butler/registry/interfaces/_versioning.py
@@ -125,7 +125,10 @@ class VersionedExtension(ABC):
         -------
         digest : `str` or `None`
             String representation of the digest of the schema, ``None`` should
-            be returned if schema digest is not to be saved or checked.
+            be returned if schema digest is not to be saved or checked. The
+            length of the returned string cannot exceed the length of the
+            "value" column of butler attributes table, currently 65535
+            characters.
 
         Notes
         -----
--- a/python/lsst/daf/butler/registry/interfaces/_versioning.py
+++ b/python/lsst/daf/butler/registry/interfaces/_versioning.py
@@ -1,0 +1,195 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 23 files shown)


python/lsst/daf/butler/registry/interfaces/_versioning.py
```diff
@@ -0,0 +1,195 @@
...
+
+        Returns
+        -------
+        digest : `str` or `None`
+            String representation of the digest of the schema, ``None`` should
+            be returned if schema digest is not to be saved or checked.
+
+        Notes
+        -----
+        The is no exact definition of digest format, any string should work.
```
TallJimbo: typo: "The" -> "There"

python/lsst/daf/butler/registry/interfaces/_versioning.py
```diff
@@ -0,0 +1,195 @@
...
+        The only requirement for string contents is that it has to remain
+        stable over tim if schema does not change but it should produce
```
TallJimbo: typo: "tim" -> "time"

python/lsst/daf/butler/registry/interfaces/_versioning.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        return f"{cls.__module__}.{cls.__name__}"
+
+    @abstractmethod
+    def schemaDigest(self) -> Optional[str]:
+        """Return digest for schema piece managed by this extension.
+
+        Returns
+        -------
+        digest : `str` or `None`
```
TallJimbo, line 126: Should we consider making these `bytes` instead of `str`, to move the burden of something like base64 or hex encoding out of implementations of this class, and into something common (and also guarantee that some encoding like that _is_ used)?  I assume that essentially all of the time they'll start out as `bytes`.

TallJimbo (approved):
Looks good, only minor comments.

One big-picture comment: why use multiple inheritance instead of just making all of the manager ABCs inherit from `VersionedExtension`?  It looks like that's effectively required for all implementations anyway.

python/lsst/daf/butler/registry/interfaces/_versioning.py
andy-slac: Or Tim 😄

python/lsst/daf/butler/registry/interfaces/_versioning.py
andy-slac: My concern here was that we need to save them in the database as strings because attribute table is a string -> string mapping. We could require bytes digests but then I'd need to convert them into string format (e.g. with hex encoding) when saving and convert back to bytes when reading. I decided to leave the choice for encoding to extension, besides there may be a case when digest can be produced as an actual human-readable string.

andy-slac:
Re multiple inheritance - I was thinking that there may be some manager implementations that don't need version support (or may not even have database schema). Requiring those to implement versioning ABC looks like unnecessary burden. OTOH as you say all our current managers support versioning, so may be it will simplify things if we use versioning ABC as a base class for all of them, implementing that interface when version is not needed should be trivial task.

TallJimbo (approved):

PR #336, opened 2020-07-24
DM-24475: Rewrite of Butler URI and major clean up of S3 usage

timj:
* ButlerURI constructor is now a factory that returns a specific subclass for that scheme.
* ButlerURI is now concrete and can check for existence of the resource and download it.
* Add resource:// URI scheme for local python package resources.
* Add http(s):// URI scheme for easy web access.
* ButlerURI can now transfer from one URI to another, supports move and transactions.
* Rewrite datastore.export to use ButlerURI thus enabling export to S3 and use in S3Datastore for export to S3 or local file.
* Change all file access in Config class to use ButlerURI.  Full support for !include and includeConfigs with remote resources. You can now create a config from a HTTP resource.
* Rewrite datastore.ingest to use ButlerURI, enabling S3 and POSIX to ingest from all supported sources.
* Butler.makeRepo now no longer special cases S3 but can write the output config to any supported location.


timj:
Some ingest tests included use of ??? in file names. This broke the URI parser. Therefore I just added URI quoting and unquoting into the mix.  This found a few more bugs.

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -0,0 +1,1332 @@
...
+# Root path for this operating system
+OS_ROOT_PATH = Path().resolve().root
+
+
+def os2posix(ospath: str) -> str:
+    """Convert a local path description to a POSIX path description.
+
+    Parameters
+    ----------
+    path : `str`
```
rra: Parameter is `ospath`, not `path`.

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -0,0 +1,1332 @@
...
+        import boto3
+    except ImportError:
+        pass
+    from .datastore import DatastoreTransaction
+
+
+log = logging.getLogger(__name__)
+
+# Determine if the path separator for the OS looks like POSIX
+IS_POSIX = os.sep == posixpath.sep
```
rra, line 64: I'm uncomfortable with all of the use of `posixpath` since that's an intentionally undocumented API, which makes this a bit of a violation of the API contract. I think it would be possible to write everything directly in terms of `Path`. That said, after looking over all the code, I'm not sure it's worth it for theoretical future portability, since the conversion is not entirely obvious.

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -0,0 +1,1332 @@
...
+    quotePaths = True
+    """True if path-like elements modifying a URI should be quoted.
+
+    All non-schemeless URIs have to internally use quoted paths. Therefore
+    if a new file name is given (e.g. to updateFile or join) a decision must
+    be made whether to quote it to be consistent.
+    """
+
+    # mypy is confused without this
+    _uri: urllib.parse.ParseResult
```
rra, line 203: Can you not put the type annotation on the `self._uri` initialization in `__new__`? I would have expected that to work, but maybe mypy doesn't understand `__new__`.

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -0,0 +1,1332 @@
...
+        Returns
+        -------
+        exists : `bool`
+            `True` if the resource exists.
+        """
+        raise NotImplementedError()
+
+    def remove(self) -> None:
+        """Remove the resource."""
+        raise NotImplementedError()
```
rra, line 508: General comment on all methods of this type: consider making `ButlerURI` inherit from `abc.ABC` and then using the `@abstractmethod` annotation on these methods, removing the `raise` body. This does the same thing except it will be checked at instantiation and mypy will also check that you provide all necessary methods in derived classes.

It looks like in some cases subclasses intentionally don't implement some of these methods, but I think that calls for some additional class hierarchy (`ButlerWritableURI`, for instance) to improve the type system. Right now, there's no way for the caller to know whether a given instance of `ButlerURI` supports a particular method, and that would allow `isinstance` and similar techniques, and would also allow getting additional type information via explicit mypy casts.

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -0,0 +1,1332 @@
...
+        if not self.isabs() and not other.isabs():
+            return super().relative_to(other)
+
+        # if we have a relative path convert it to absolute
+        # relative to the supplied parent.  This is solely to handle
+        # the case where the relative path includes ".." but somehow
+        # then goes back inside the directory of the parent
+        if not self.isabs():
+            childUri = other.join(self.path)
+            return childUri.relative_to(other)
```
rra, line 778: This logic feels weird to me. A relative path is, by convention, relative to the current working directory, but here you're assuming that it's relative to the supplied path, which could produce far different results. Maybe what I'm asking for here is better documentation of what "relative" means in the context of a ButlerURI?

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -0,0 +1,1332 @@
...
+
+        # if one is schemeless and the other is not the base implementation
+        # will fail so we need to fix that -- they are both absolute so
+        # forcing to file is fine
```
rra: This comment goes with the code below, not with this block.

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -0,0 +1,1332 @@
...
+        if self.scheme == other.scheme:
+            return super().relative_to(other)
+
+        # Force everything to file
+        # mypy does not know that only ButlerFileURI can do force_to_file
+        return self._force_to_file().relative_to(other._force_to_file())  # type: ignore
```
rra: I think the problem is that mypy doesn't know that `other` is a `ButlerFileURI` and would reasonably complain that you're calling a method that may be invalid on some other `ButlerURI`. Rather than `# type: ignore`, I would add an explicit `cast` to `ButlerFileURI`, since at this point you have established that `other` is indeed a file URI.

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -0,0 +1,1332 @@
...
+
+    return posix
+
+
+def posix2os(posix: Union[PurePath, str]) -> str:
+    """Convert a POSIX path description to a local path description.
+
+    Parameters
+    ----------
+    posix : `str`
```
rra: Parameter can also be a `PurePath`.

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -0,0 +1,1332 @@
...
+        """Transfer the current resource to a new location.
+
+        Parameters
+        ----------
+        src : `ButlerURI`
+            Source URI.
+        transfer : `str`
+            Mode to use for transferring the resource. Generically there are
+            many standard options: copy, link, symlink, hardlink, relsymlink.
+            Not all URIs support all modes.
```
rra, line 657: Shouldn't this be an enum?

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -0,0 +1,1332 @@
...
+
+class ButlerGenericURI(ButlerURI):
+    """Generic URI with a defined scheme"""
+
+    def exists(self) -> bool:
+        """Test for existence and always return False."""
+        return False
+
+    def as_local(self) -> Tuple[str, bool]:
+        raise RuntimeError(f"Do not know how to retrieve data for URI '{self}'")
```
rra: What is the purpose of this class as opposed to throwing an exception from `ButlerURI.__new__`? That should ideally be in the documentation.

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -0,0 +1,1332 @@
...
+        ------
+        ValueError
+            Raised if this URI is schemeless and relative path and so can
+            not be forced to file absolute path without context.
+        """
+        if not self.isabs():
+            raise RuntimeError(f"Internal error: Can not force {self} to absolute file URI")
+        uri = self._uri._replace(scheme="file", path=urllib.parse.quote(os2posix(self.path)))
+        # mypy really wants a ButlerFileURI to be returned here
+        return ButlerURI(uri, forceDirectory=self.dirLike)  # type: ignore
```
rra, line 1252: You know that the return of `ButlerURI` will be a `ButlerFileURI`, so rather than `# type: ignore` I would add an explicit `cast`.

python/lsst/daf/butler/core/config.py
```diff
@@ -343,93 +258,30 @@ def fromYaml(cls, string: str) -> Config:
...
         """Load a file from a path or an URI.
 
         Parameters
         ----------
         path : `str`
-            Path or an URI to a persisted config file.
+            Path or a URI to a persisted config file.
         """
         uri = ButlerURI(path)
         if uri.path.endswith("yaml"):
```
rra: `uri.getExtension() == ".yaml"` instead?

tests/test_uri.py
```diff
@@ -0,0 +1,330 @@
...
+        for mode in ("copy", "link", "hardlink", "symlink", "relsymlink"):
+            dest = ButlerURI(os.path.join(self.tmpdir, f"dest_{mode}.txt"))
+            dest.transfer_from(src, transfer=mode)
+            self.assertTrue(dest.exists(), f"Check that {dest} exists (transfer={mode})")
+
+            with open(dest.ospath, "r") as fh:
+                new_content = fh.read()
+            self.assertEqual(new_content, content)
+
+            if "symlink" in mode:
```
rra: Also `relsymlink` too, yes? It would probably also be worth checking the value of the symlink.

rra (approved):
I had some nits, but overall this looks good to me. I will say that it all still feels rather complex, and I would be tempted to store some additional data to make some of the operations clearer. For example, I think several operations would be simpler if a `ButlerFileURI` created a `Path` object on instantiation and stashed it inside the object and used it for several of the APIs. But that may be gilding the lily, and you probably want to get on with other work.

The one thing that makes me nervous is that I'm worried that relative paths are still ill-defined (specifically, "relative to what?" since in at least one place it doesn't seem to be the current working directory). I don't know enough of the surrounding context to know if you need to retain the relativeness because the working directory will change over the lifetime of the `ButlerURI`. If not, I would be sorely tempted to eliminate relative paths by canonicalizing everything on instantiation, since you're carrying a ton of complexity in order to make relative paths absolute at the last possible moment.

It might be worth trying to do some explicit testing with Windows paths, since I'm not sure how drive letters and whatnot would interact with a lot of this file logic.

python/lsst/daf/butler/core/_butlerUri.py
timj: The posixpath docs explicitly say:

> Some of this can actually be useful on non-Posix systems too, e.g. for manipulation of the pathname component of URLs.

so that seems good enough for me since that's exactly what I'm using it for.

python/lsst/daf/butler/core/_butlerUri.py
timj: I did have abstractmethod to begin with but `__new__` completely messed that up because mypy never thinks you ever get a subclass back from the `ButlerURI` factory so it would get very upset if a the base class uses one of those abstract methods (because it would rightly say that that couldn't possibly work because the base class is an abc).

python/lsst/daf/butler/core/_butlerUri.py
timj: It existed when I had abstractmethod in the base class but I didn't want to break everything up front if someone used an unsupported URI scheme. In theory the "PurePath" portion of ButlerURI is fine with an unknown scheme, it should only complain if you want to do something with an actual remote resource.

python/lsst/daf/butler/core/config.py
timj: Yes. That would be better (I wrote that before getExtension existed).

tests/test_uri.py
timj: Yes, that's why I used `in` and not `==` (since relsymlink does include symlink).

python/lsst/daf/butler/core/_butlerUri.py
rra: Yes, true, although `pydoc` also provides a documentation link that 404s. But yes, I suspect this is low-risk.

python/lsst/daf/butler/core/_butlerUri.py
rra: Oh, yes. Hm, right, I can't see a way to work around that, so this is probably best. That's a frustrating limitation. Probably worth noting in a comment somewhere so that no one else goes through the same experiment and finds the same failure?

python/lsst/daf/butler/core/_butlerUri.py
rra: Right, but do you ever want to use it that way? There is a bit of safety benefit to just throwing an exception in `__new__` when given an unsupported URL. For example, you will catch and throw a clear exception for `htpt` URLs and other typos, whereas now I think you may end up with obscure `NotImplementedError` exceptions much later on.

tests/test_uri.py
rra: Oh, right, Python `in` works with strings as well as proper collections. I always find this confusing. To be honest, I personally would still write this as `if mode in ("symlink", "relsymlink")` because I think it's more obvious and easier to read, but that's a very minor nit.

python/lsst/daf/butler/core/_butlerUri.py
timj: You would think that it would be enough to declare what type `parsed` was but if I leave this line out it really doesn't like it. It must be the use of `__new__`.

python/lsst/daf/butler/core/_butlerUri.py
timj: Probably true. I think I'll punt to a new ticket because there are ramifications for the APIs that pass the transfer mode into this routine that I need to consider.

python/lsst/daf/butler/core/_butlerUri.py
timj: Ok. I'll try the extreme prejudice approach and see what happens. (I had to add `mem://` for the in-memory datastore...)

python/lsst/daf/butler/core/_butlerUri.py
timj: Thanks. That worked.

python/lsst/daf/butler/core/_butlerUri.py
timj: For a relative path relative to an absolute path I'm trying to say that the answer is by definition the relative path unless it has `../` in it, in which case it goes up one level. I think that's a fairly well defined definition of relative. The current working directory is not relevant.

timj:
Regarding the use of `ntpath` -- it's a good idea and I'm sure something will break in my os.path usage if I try it. I think though that I'd save that for a rainy day unless you have a trivial way of mocking `sys.builtin_module_names` so that os.path will magically turn into ntpath.

python/lsst/daf/butler/core/_butlerUri.py
timj: That worked. Thanks.

PR #338, opened 2020-07-30
DM-26160: Rewrite Location class to make proper use of ButlerURI

timj:
Without this we were getting inconsistencies when a datastore root was a path that had a character in it that would be escaped. This was because in some cases os.path was being used when a ButlerURI method should have been used.

### DIFF
--- a/python/lsst/daf/butler/core/_butlerUri.py
+++ b/python/lsst/daf/butler/core/_butlerUri.py
@@ -349,10 +349,14 @@ class ButlerURI:
             ButlerURI rules.
         tail : `str`
             Last `self.path` component. Tail will be empty if path ends on a
-            separator. Tail will never contain separators.
+            separator. Tail will never contain separators. It will be
+            unquoted.
         """
         head, tail = self._pathModule.split(self.path)
         headuri = self._uri._replace(path=head)
+
+        # The file part should never include quoted metacharacters
+        tail = urllib.parse.unquote(tail)
 
         # Schemeless is special in that it can be a relative path
         # We need to ensure that it stays that way. All other URIs will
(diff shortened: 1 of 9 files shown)


python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -66,6 +67,9 @@
 # Root path for this operating system
 OS_ROOT_PATH = Path().resolve().root
 
+# Regex for looking for URI escapes
+ESCAPES_RE = re.compile(r"%[A-B0-9]{2}")
```
ktlim: Not `A-F`?  I could see `=`, `,`, `/` ,`?` or `]` in a URI; they're all in `C-F`.

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -218,7 +222,10 @@ def __new__(cls, uri: Union[str, urllib.parse.ParseResult, ButlerURI],
             # Since sometimes people write file:/a/b and not file:///a/b
             # we should not quote in the explicit case of file:
             if "://" not in uri and not uri.startswith("file:"):
-                uri = urllib.parse.quote(uri)
+                if ESCAPES_RE.search(uri):
+                    log.warning("Possible double encoding of %s", uri)
```
ktlim, line 226: Should you test for this warning?

ktlim (approved):
This looks OK, and the tests seem to indicate that it is OK for files.  I do wonder a bit about whether things work properly for non-`file` schemes that include URI-encoded characters.

python/lsst/daf/butler/core/_butlerUri.py
timj: 😆 Yes A-F. Wow.

python/lsst/daf/butler/core/_butlerUri.py
timj: Can do.

timj:
I've added some tests for S3 URI quoting.

PR #340, opened 2020-07-31
DM-26196: Add skip-dimensions option to butler import

timj:


timj:
Used the wrong branch name -- reopened as #341

PR #341, opened 2020-07-31
DM-26198: Add skip-dimensions option to butler import

timj:


TallJimbo (approved):
Looks good. Only comment is that there's a typo in the second commit message.

PR #345, opened 2020-08-10
Add support for Webdav datastore

bgounon:
This implements the possibility of using a Webdav server as a Butler Datastore, using the `http://` or `https://` scheme. Should also work against a standard HTTP server for basic read.
This work is initially based on @DinoBektesevic previous work to bring S3 support to the Butler.
We expect that some sort of authentication is required by the Webdav endpoint, and currently support x509_proxy and token-based authentication.
The following environment variables must be set accordingly:
 * `WEBDAV_AUTH_METHOD` (acceptable values are `X509` or `TOKEN`)
 * `WEBDAV_PROXY_CERT`: path to the x509 proxy certificate (only when using `X509`auth)
 * `WEBDAV_BEARER_TOKEN`: the bearer token, as a simple string (only when using `TOKEN` auth)
 * `REQUESTS_CA_BUNDLE`: the `requests` library will read CA certificates from this directory. The endpoint's CA certificate must be present to avoid SSL failures.

The code was tested against the Webdav door of a dCache backend.


python/lsst/daf/butler/_butlerConfig.py
```diff
@@ -93,6 +93,11 @@ def __init__(self, other=None, searchPaths=None):
                     uri = ButlerURI(other, forceDirectory=True)
                 uri.updateFile("butler.yaml")
                 other = uri.geturl()
+            elif uri.scheme == "https":
```
timj, line 96: I missed this on my ButlerURI cleanup -- no code should be checking URI schemes outside of ButlerURI now. I'll have to clean this up later.

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -1172,10 +1172,38 @@ def read(self, size: int = -1) -> bytes:
...
         """Check that a remote HTTP resource exists."""
-        header = requests.head(self.geturl())
-        return True if header.status_code == 200 else False
+        log.debug("Checking if resource exists: %s", self.geturl())
+        r = self.session.head(self.geturl())
+
+        return True if r.status_code == 200 else False
+
+    def mkdir(self) -> None:
+
```
timj: Add a simple docstring

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -1196,14 +1225,61 @@ def as_local(self) -> Tuple[str, bool]:
         return tmpFile.name, True
 
     def read(self, size: int = -1) -> bytes:
-        # Docstring inherits
```
timj, line 1220: Put this comment back or add a real docstring

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -1196,14 +1225,61 @@ def as_local(self) -> Tuple[str, bool]:
...
+        log.debug("Reading from remote resource: %s", self.geturl())
         stream = True if size > 0 else False
-        r = requests.get(self.geturl(), stream=stream)
+        r = self.session.get(self.geturl(), stream=stream)
         if not stream:
             return r.content
         else:
             return next(r.iter_content(chunk_size=size))
 
+    def write(self, data: bytes, overwrite: bool = True) -> None:
```
timj, line 1269: Docstring

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -1196,14 +1225,61 @@ def as_local(self) -> Tuple[str, bool]:
...
+
+        if self.exists():
+            raise FileExistsError(f"Destination path '{self}' already exists.")
+
+        if transfer == "auto":
+            transfer = self.transferDefault
+
+        if isinstance(src, type(self)):
+            if transfer == "move":
+                self.session.request('MOVE', src.geturl(), headers={'Destination': self.geturl()})
```
timj, line 1317: daf_butler uses double quotes for strings so please fix all the quoting for consistency.

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -1196,14 +1225,61 @@ def as_local(self) -> Tuple[str, bool]:
...
+            else:
+                self.session.request('COPY', src.geturl(), headers={'Destination': self.geturl()})
+        else:
+            # Use local file and upload it
+            local_src, is_temporary = src.as_local()
+            files = {'file': open(local_src, 'rb')}
```
timj, line 1323: What closes this file handle?

python/lsst/daf/butler/core/webdavutils.py
```diff
@@ -0,0 +1,167 @@
...
+    X509: must set WEBDAV_PROXY_CERT
+    (path to proxy certificate used to authenticate requests)
+
+    TOKEN: must set WEBDAV_BEARER_TOKEN
+    (bearer token used to authenticate requests, as a single string)
+
+    NB: requests will read CA certificates in REQUESTS_CA_BUNDLE
+    It must be manually exported according to the system CA directory.
+    """
+    s = requests.Session()
```
timj: Can you use `session` as the variable name to be more in line with other places here.

python/lsst/daf/butler/core/webdavutils.py
```diff
@@ -0,0 +1,167 @@
...
+        try:
+            bearer_token = os.environ['WEBDAV_BEARER_TOKEN']
+        except KeyError:
+            raise KeyError("Environment variable WEBDAV_BEARER_TOKEN is not set")
+        s.headers = CaseInsensitiveDict({'Authorization': 'Bearer ' + bearer_token})
+    else:
+        raise ValueError("Environment variable WEBDAV_AUTH_METHOD must be set to X509 or TOKEN")
+
+    # s.verify = False
+    log.debug("Session configured and ready.")
```
timj, line 91: Maybe as a debug message it would be helpful to include which auth method was used.

python/lsst/daf/butler/core/webdavutils.py
```diff
@@ -0,0 +1,167 @@
...
+
+def getFileURL(path: Union[Location, ButlerURI, str]) -> str:
+
+    if isinstance(path, str):
+        filepath = path
+    elif isinstance(path, ButlerURI):
+        filepath = path.geturl()
+    elif isinstance(path, Location):
+        filepath = path.uri.geturl()
+
```
timj, line 201: This needs an `else`

python/lsst/daf/butler/datastores/webdavDatastore.py
```diff
@@ -0,0 +1,271 @@
...
+            if srcUri.scheme == "file":
+                raise RuntimeError(f"'{srcUri}' is not inside repository root '{rootUri}'. "
+                                   "Ingesting local data to WebdavDatastore without upload "
+                                   "to Webdav is not allowed.")
+            elif srcUri.scheme.startswith("http"):
+                if not srcUri.path.startswith(rootUri.path):
+                    raise RuntimeError(f"'{srcUri}' is not inside repository root '{rootUri}'.")
+        return path
+
+    def _extractIngestInfo(self, path: Union[str, ButlerURI], ref: DatasetRef, *,
```
timj, line 232: This whole method needs to be rewritten following the recent cleanups in s3datastore.py to use ButlerURI everywhere. It's probably the case that it's identical to the s3 one except for the existence check right at the end.

python/lsst/daf/butler/datastores/webdavDatastore.py
```diff
@@ -0,0 +1,271 @@
...
+        self._transaction.registerUndo("write", webdavDeleteFile, location)
+
+        return self._extractIngestInfo(location.uri.geturl(), ref, formatter=formatter)
+
+    def _overrideTransferMode(self, *datasets: Any, transfer: Optional[str] = None) -> Optional[str]:
+        if transfer != "auto":
+            return transfer
+        return "copy"
+
+    def _standardizeIngestPath(self, path: str, *, transfer: Optional[str] = None) -> str:
```
timj, line 208: This should now be identical to the s3 version. We can probably consider moving the s3 one to fileLikeDatastore.py and not having a webdav one.

python/lsst/daf/butler/datastores/webdavDatastore.py
```diff
@@ -0,0 +1,271 @@
...
+                formatter._fileDescriptor.location = Location(*os.path.split(tmpFile.name))
+                result = formatter.read(component=getInfo.component if isComponent else None)
+        except Exception as e:
+            raise ValueError(f"Failure from formatter '{formatter.name()}' for dataset {ref.id}"
+                             f" ({ref.datasetType.name} from {location.uri}): {e}") from e
+
+        return self._post_process_get(result, getInfo.readStorageClass, getInfo.assemblerParams,
+                                      isComponent=isComponent)
+
+    def _write_in_memory_to_artifact(self, inMemoryDataset: Any, ref: DatasetRef) -> StoredFileInfo:
```
timj, line 177: It's clear that this needs to be rewritten using ButlerURI to do all the work so that it would be generic -- but not on this pull request.

python/lsst/daf/butler/datastores/webdavDatastore.py
```diff
@@ -0,0 +1,271 @@
...
+        super().__init__(config, bridgeManager, butlerRoot)
+        b = ButlerURI(self.root)
+        self.session = getHttpSession()
+        if not folderExists(self.root, session=self.session):
+            try:
+                b.mkdir()
+            except ValueError:
+                raise ValueError(f"Can not create directory {self.root}, check permissions.")
+
+    def _artifact_exists(self, location: Location) -> bool:
```
timj, line 103: Note to self: Move to parent class and use ButlerURI

python/lsst/daf/butler/datastores/webdavDatastore.py
```diff
@@ -0,0 +1,271 @@
...
+        Returns
+        -------
+        exists : `bool`
+            True if the location can be found, false otherwise.
+        """
+        exists, _ = webdavCheckFileExists(location,
+                                          session=self.session)
+        return exists
+
+    def _delete_artifact(self, location: Location) -> None:
```
timj, line 119: Note to self: Move to parent class

python/lsst/daf/butler/datastores/webdavDatastore.py
```diff
@@ -0,0 +1,271 @@
...
+
+    defaultConfigFile = "datastores/webdavDatastore.yaml"
+    """Path to configuration defaults. Relative to $DAF_BUTLER_DIR/config or
+    absolute path. Can be None if no defaults specified.
+    """
+
+    def __init__(self, config: Union[DatastoreConfig, str],
+                 bridgeManager: DatastoreRegistryBridgeManager, butlerRoot: str = None):
+        super().__init__(config, bridgeManager, butlerRoot)
+        b = ButlerURI(self.root)
```
timj: You probably want a `forceDirectory=True` here.  I need to clean up all this to ensure that self.root is always a ButlerURI.

Also, `b` means nothing to the reader so can you say `uri` as the variable instead?

python/lsst/daf/butler/datastores/webdavDatastore.py
```diff
@@ -0,0 +1,271 @@
...
+        self.session = getHttpSession()
+        if not folderExists(self.root, session=self.session):
```
timj: Why can't this use `b.exists()`?

python/lsst/daf/butler/datastores/webdavDatastore.py
```diff
@@ -0,0 +1,271 @@
...
+        ----------
+        location : `Location`
+            Expected location of the artifact associated with this datastore.
+
+        Returns
+        -------
+        exists : `bool`
+            True if the location can be found, false otherwise.
+        """
+        exists, _ = webdavCheckFileExists(location,
```
timj: Use `location.uri.exists()`

python/lsst/daf/butler/datastores/webdavDatastore.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def _delete_artifact(self, location: Location) -> None:
+        """Delete the artifact from the datastore.
+
+        Parameters
+        ----------
+        location : `Location`
+            Location of the artifact associated with this datastore.
+        """
+        r = self.session.delete(location.uri.geturl())
```
timj: Use `location.uri.remove()`

python/lsst/daf/butler/datastores/webdavDatastore.py
```diff
@@ -0,0 +1,271 @@
...
+        if r.status_code not in [200, 202, 204]:
+            raise FileNotFoundError(f"Unable to delete resource {self}; status code: {r.status_code}")
+
+    def _read_artifact_into_memory(self, getInfo: DatastoreFileGetInformation,
```
timj, line 129: I'm going to have to rewrite this to use ButlerURI.

python/lsst/daf/butler/datastores/webdavDatastore.py
```diff
@@ -0,0 +1,271 @@
...
+                log.debug("Wrote file to %s via a temporary directory.", location.uri)
+
+        if self._transaction is None:
+            raise RuntimeError("Attempting to write artifact without transaction enabled")
+
+        self._transaction.registerUndo("write", webdavDeleteFile, location)
+
+        return self._extractIngestInfo(location.uri.geturl(), ref, formatter=formatter)
+
+    def _overrideTransferMode(self, *datasets: Any, transfer: Optional[str] = None) -> Optional[str]:
```
timj, line 203: Note to self: Move to parent class since it's only the POSIX datastore that is special and needs to override.

python/lsst/daf/butler/core/webdavutils.py
```diff
@@ -0,0 +1,167 @@
...
+def isWebdavEndpoint(path: Union[Location, ButlerURI, str]) -> bool:
+
+    filepath = getFileURL(path)
+
+    log.debug("Detecting HTTP endpoint type...")
+    r = requests.options(filepath)
+    return True if 'DAV' in r.headers else False
+
+
+def getFileURL(path: Union[Location, ButlerURI, str]) -> str:
```
timj: I think I'd keep this as a private function and not export it. I'd consider dropping `Location` support and just make callers use `location.uri`.

The str check isn't needed because in theory you can always do `ButlerURI(path).geturl()` for str and ButlerURI.

python/lsst/daf/butler/core/webdavutils.py
```diff
@@ -0,0 +1,167 @@
...
+
+    filepath = getFileURL(path)
+
+    log.debug("Checking if folder exists: %s", filepath)
+    r = session.head(filepath)
+    return True if r.status_code == 200 else False
+
+
+def isWebdavEndpoint(path: Union[Location, ButlerURI, str]) -> bool:
+
```
timj, line 181: Docstring missing.

python/lsst/daf/butler/core/webdavutils.py
```diff
@@ -0,0 +1,167 @@
...
+    log.debug("Checking if file exists: %s", filepath)
+
+    r = session.head(filepath)
+    return (True, int(r.headers['Content-Length'])) if r.status_code == 200 else (False, -1)
+
+
+def webdavDeleteFile(path: Union[Location, ButlerURI, str],
+                     session: Optional[requests.Session] = None) -> None:
+    """Check that a remote HTTP resource exists."""
```
timj: Need to document parameters

python/lsst/daf/butler/core/webdavutils.py
```diff
@@ -0,0 +1,167 @@
...
+
+    # s.verify = False
+    log.debug("Session configured and ready.")
+
+    return s
+
+
+def webdavCheckFileExists(path: Union[Location, ButlerURI, str],
+                          session: Optional[requests.Session] = None) -> Tuple[bool, int]:
+    """Check that a remote HTTP resource exists."""
```
timj: Need to document parameters

timj (changes requested):
Thanks for doing this. I will make a Jira ticket in our system so that we can track the work properly.

I have made lots of little comments and some of my comments are to myself to make cleanups. Since you started this before the cleanup to s3datastore.py you need to bring things in sync again.

Please add some webdav tests to tests/test_uri.py.

Also, is it possible to mock things so that you can add a proper datastore test? It would be great if you could add a webdav test class to test_butler.py like we do for s3 so we can be a bit more confident that things work.

python/lsst/daf/butler/core/webdavutils.py
bgounon: The problem with dropping `Location` support in `getFileURL` is that the callers aren't aware of the type of the `path` they send when calling `getFileURL`.
This means that we should either completely drop `Location` support in callers also, and make sure themselves are never called with a `Location` parameter somewhere higher in the code, **or** do the `isinstance(path, Location)` test every time before we call `getFileURL`, which is probably not desirable.

timj: Ok, this will go away when I do a clean up pass I think. Please make it a private API though.

timj:
Please rename your branch to tickets/DM-26310 -- I think that will require closing this pull request and opening a new one.

bgounon:
OK, closing this PR, new one at #351 
I'll address comments from this PR in the new one

PR #346, opened 2020-08-10
DM-24247: Fix skypix dimension aliasing in file template validation

timj:


TallJimbo (approved):
Looks good.

I gather we're just giving up on snake_case vs. camelCase consistency, at least in the CLI-facing stuff?  I see both styles here, even in the same method, and I find that _slightly_ worrying, but not enough to do more than just make sure that reflects an intentional slow transition to snake_case (in all new code?  in some places?) rather than accidental.

PR #347, opened 2020-08-11
DM-25627 make qgraph and run subcommands for pipetask

n8pease:


python/lsst/daf/butler/cli/butler.py
```diff
@@ -173,7 +175,9 @@ def _importPlugin(pluginName):
             return doImport(pluginName)
         except Exception as err:
             log.warning("Could not import plugin from %s, skipping.", pluginName)
-            log.debug("Plugin import exception: %s", err)
+            tb = io.StringIO()
+            traceback.print_tb(err.__traceback__, file=tb)
```
andy-slac: Could you use `traceback.format_tb()` to avoid temporary file instance?

python/lsst/daf/butler/cli/opt/collection_type.py
```diff
@@ -22,11 +22,11 @@
 
 import click
 
-from lsst.daf.butler.cli.utils import MWOption
+from lsst.daf.butler.cli.utils import MWOption, MWOptionDecorator
```
andy-slac: Could use relative import for this?

python/lsst/daf/butler/cli/utils.py
```diff
@@ -440,3 +567,259 @@ def make_metavar(self):
...
+        -------
+        flags : `tuple` [`str`]
+            The flags to use.
+        """
+        pass
+
+    @classmethod
+    def _update(cls):
+        """Use the click.Option to generate the kwarg name and the list of
+        flags
```
andy-slac: Here and few other places - need period at the end of sentences.

python/lsst/daf/butler/cli/utils.py
```diff
@@ -440,3 +567,259 @@ def make_metavar(self):
...
+    @abc.abstractmethod
+    def defaultHelp():
+        """The help default help text to use for the Option.
+
+        Returns
+        -------
+        defaultHelp : `str`
+            The default help text.
+        """
+        pass
```
andy-slac: This should probably be `raise NotImplementedError()` just in case someone calls this method, same for `optionFlags`.

python/lsst/daf/butler/cli/utils.py
```diff
@@ -246,20 +253,35 @@ def split_kv(context, param, values, choice=None, multiple=True, normalize=False
...
     separator : str, optional
         The character that separates key-value pairs. May not be a comma or an
         empty space (for space separators use Click's default implementation
         for tuples; `type=(str, str)`). By default "=".
     unseparated_okay : `bool`, optional
         If True, allow values that do not have a separator. They will be
         returned in the values dict as a tuple of values in the key '', that
-        is: `values[''] = (unseparated_values, )`.
+        is: `values[''] = (unseparated_values, )`. By default False.
+    return_type : `str`, one of "dict" or "tuple"
```
andy-slac: Instead of string you can accept type itself (`dict` or `tuple`), would at least reduce chance of typo (and need to check for that).

python/lsst/daf/butler/cli/utils.py
```diff
@@ -288,17 +310,44 @@ def norm(val):
...
+                    raise click.ClickException(f"Duplicate entries for '{k}' in '{values}'")
+                self.ret[key] = val
+            elif self.typ == "tuple":
+                self.ret.append((key, val))
+
+        def get(self):
+            if self.typ == "tuple":
+                return tuple(self.ret)
+            else:
+                return self.ret
```
andy-slac: Feels like two separate classes merged into one. You could have `RetDict` and `RetTuple` and only one `if` when you instantiate it.

python/lsst/daf/butler/cli/utils.py
```diff
@@ -246,20 +253,35 @@ def split_kv(context, param, values, choice=None, multiple=True, normalize=False
...
+        values, the value to the left of the separator will be the key and the
+        value to the right of the separator will be the value.
+        If "tuple" then the returned object will be a tuple. Each item in the
+        tuple will be 2-item tuple, the first item will be the value to the
+        left of the separator and the second item will be the value to the
+        right. By default "dict".
+    reverse_kv : bool
+        If true then for each item in values, the value to the left of the
+        separator is treated as the value and the value to the right of the
+        separator is treated as the key. By default False.
```
andy-slac, line 279: No docstring for `default_key` parameter?

python/lsst/daf/butler/cli/utils.py
```diff
@@ -398,9 +448,73 @@ def cli_handle_exception(func, *args, **kwargs):
...
+                            cls=OptionSection)(f)
+
+
+class MWPath(click.Path):
+    """Overrides click.Path to implement file-does-not-exist checking.
+
+    Changes the definition of `exists` so that `True` indicates the location
+    (file or directory) must exist, `False` indicates the location must *not*
+    exist, and `None` indicates that the file may exist or not. The standard
+    definition for the 1click.Path1 `exists` parameter is that for `True` a
```
andy-slac: `1click.Path1`

python/lsst/daf/butler/cli/utils.py
```diff
@@ -398,9 +448,73 @@ def cli_handle_exception(func, *args, **kwargs):
...
+        # section.
+        return click.option(f"--option-section-{str(uuid.uuid4())}",
+                            sectionText=self.sectionText,
+                            cls=OptionSection)(f)
+
+
+class MWPath(click.Path):
+    """Overrides click.Path to implement file-does-not-exist checking.
+
+    Changes the definition of `exists` so that `True` indicates the location
```
andy-slac: Double backticks for parameter/variable names?

python/lsst/daf/butler/cli/utils.py
```diff
@@ -398,9 +448,73 @@ def cli_handle_exception(func, *args, **kwargs):
...
+        if exists is None:
+            exists = False
+        super().__init__(exists, file_okay, dir_okay, writable, readable,
+                         resolve_path, allow_dash, path_type)
+
+    def convert(self, value, param, ctx):
+        """Called by click.ParamType to "convert values through types".
+        `click.Path` uses this step to verify Path conditions."""
+        if self.mustNotExist and os.path.exists(value):
+            self.fail(f'"{value}" should not exist.')
```
andy-slac: Maybe mention explicitly that you complain about file, message `"chewbacca" should not exist` may be confusing.

python/lsst/daf/butler/cli/utils.py
```diff
@@ -398,9 +448,73 @@ def cli_handle_exception(func, *args, **kwargs):
...
+        return super().convert(value, param, ctx)
+
+
 class MWOption(click.Option):
     """Overrides click.Option with desired behaviors."""
 
+    def __init__(self, *args, **kwargs):
+        # `forward` indicates wheather a subcommand should forward the value of
```
andy-slac: "indicates weather" maybe?  ⛅ 😺

python/lsst/daf/butler/cli/utils.py
```diff
@@ -440,3 +567,259 @@ def make_metavar(self):
...
+        dashes if one is declared, otherwise uses the first flag that begins
+        with a single dash."""
+        cls._update()
+        ret = None
+        for opt in cls._opts:
+            if ret is None or (ret.startswith("-") and opt.startswith("--")):
+                ret = opt
+            if ret.startswith("--"):
+                break
+        return ret
```
andy-slac: I had to read it twice to figure out what it does, here is my version which I think is more transparent:

    ret = None
    for opt in cls._opts:
        if ret.startswith("--"):
            return opt
        elif ret is None:  # maybe also needs `and opt.stratswith("-")`
            ret = opt
    return ret

python/lsst/daf/butler/cli/utils.py
```diff
@@ -440,3 +567,259 @@ def make_metavar(self):
...
+
+
+class ForwardOptions:
+    """Captures CLI options to be forwarded from one subcommand to future
+    subcommands executed as part of a single CLI command invocation
+    (called "chained commands").
+
+    Parameters
+    ----------
+    optionDecorators : `list` or `tuple` [`MWOptionDecorator`]
```
andy-slac: Should it be `mwOptions` instead of `optionDecorators`?

python/lsst/daf/butler/cli/utils.py
```diff
@@ -440,3 +567,259 @@ def make_metavar(self):
...
+
+    Attributes
+    ----------
+    functionKwargs : `dict` [`str`, `str`]
+        The cached kwargs (argument name and argument value) from subcommands
+        that were called with values passed in as options on the command line.
+    """
+
+    def __init__(self, mwOptions=None):
+        self.functionKwargs = {}
```
andy-slac, line 702: I'm confused, `mwOptions` is not used at all?

python/lsst/daf/butler/cli/utils.py
```diff
@@ -440,3 +567,259 @@ def make_metavar(self):
...
+
+    @staticmethod
+    def _getKwargsToSave(mwOptions, arguments, **kwargs):
+        """Get kwargs that should be cached for use by subcommands invoked in
+        the future.
+
+        Parameters
+        ----------
+        mwOptions : `list` or `tuple` [`MWOption`]
```
andy-slac, line 711: Can be `iterable` instead of `list or tuple`?

python/lsst/daf/butler/tests/cliOptionTestBase.py
```diff
@@ -445,6 +453,67 @@ def test_forCaseInsensitive_upperLower(self):
...
+        if self.valueType.min is not None:
+            vals.append(self.valueType.min - 1)
+        if self.valueType.max is not None:
+            vals.append(self.valueType.max + 1)
+        for val in vals:
+            helper = CliFactory.mocked(self)
+            self.run_test(helper.cli,
+                          self.makeInputs(self.optionFlag, val),
+                          self.verifyError,
+                          "")  # the error strings are too varied, don't test.f
```
andy-slac: "test.f"? You did not finish sentence 😸

tests/test_cliUtils.py
```diff
@@ -175,5 +187,250 @@ def cmd(things):
...
+                                    call("no_test_option"),
+                                    call("forwards", test="foo")))
+
+
 if __name__ == "__main__":
-    unittest.main()
+    if False:
+        ForwardObjectsTest.cli()
+    else:
+        unittest.main()
```
andy-slac: Why `if False`?

python/lsst/daf/butler/cli/utils.py
```diff
@@ -398,9 +448,73 @@ def cli_handle_exception(func, *args, **kwargs):
...
+
+    def __call__(self, f):
+        # Generate a parameter declaration that will be unique for this
+        # section.
+        return click.option(f"--option-section-{str(uuid.uuid4())}",
+                            sectionText=self.sectionText,
+                            cls=OptionSection)(f)
+
+
+class MWPath(click.Path):
```
timj, line 473: When specifying a butler configuration file you can't use anything that assumes that it can open a local file. From the ctrl_mpexec PR I see that this is used for butler configs. Butler configs can be URIs that may or may not include the "butler.yaml" at the end (they can be `s3://` etc).  You could use `ButlerURI.exists()` to see if it exists but better to pass the string to `Butler()` directly without trying to check it.

andy-slac (approved):
Looks OK in general, bunch of minor comments. My concern after reading through all new additions is that it looks like you have to fight click to force it to do what we want. To my taste it's too much code that we did not have to write.

tests/test_cliUtils.py
n8pease: was temp. will remove.

python/lsst/daf/butler/cli/utils.py
n8pease: well I guess it could be a dict but it really oughtn't be.

n8pease:
@andy-slac 

> Looks OK in general, bunch of minor comments. My concern after reading through all new additions is that it looks like you have to fight click to force it to do what we want. To my taste it's too much code that we did not have to write.

Why does it feel that way to you? I didn't feel like I was fighting or forcing click, rather extending it, where I wanted to slightly change behavior or get a bit of information. 

I think I could remove a lot of the code that "extends" Click by getting rid of chained commands and command forwarding and going back to your definitions of build, qgraph, and run. Maybe that's a better way to go?

andy-slac:
My impression that click is not cooperating is mostly from looking at `utils.py` and what you had to do for pipeline actions in `ctrl_mpexec`. `option_section` looks more like a hack in click than an extension. All extra classes that you added to support chaining/forwarding also tells me that click idea of how things should work is different from what we need. 

And I'm not a fan of chaining, so if you say you want to drop it with all extra classes I won't object 🙂 

Another idea that occurred to me - the list of options is big and probably growing, if we want to simplify reuse it may be worth to "package" related options into one decorator, i.e. instead of:
```
@output_option()
@output_run_option()
@extend_run_option()
@replace_run_option()
@prune_replaced_option()
def something()
```
do 
```
@all_output_options()
def something()
```
(that example is for ctrl_mpexec options but we may have something similar in other places).

n8pease:
That all makes sense. I had coded up a batch decorator for the butler config options, but had hoped chaining would prevent the need to use it. Of course it turns out that qgraph and run still accept the butler options anyway. Given that , and that the chaining implementation requires more hacking on Click than you'd like, and that the number of lines of code in this review is more than people would like, it's sounding to me like the way to go is to take out chaining and making the commands behave the same as with the existing ctrl_mpexec command line parser.

timj:
I'm not completely averse to chaining, but it's impossible to work out what chaining really means from this pull request. I thought we had agreed to a community post to discuss things.

n8pease:
A community post after getting code checked in, which assumed I could get it past review 🙃

timj:
Ok. Firstly, please fix the tests so that the PR can build.

Secondly, we have a chicken and egg problem. It's really hard to review the pipetask code without the context of understanding subcommands. It's a very large pull request to understand without that context. Since pipetask2 is not something that people are using it seems like we can only break this impasse by merging the ctrl_mpexec PR, writing the community post, letting people experiment, and then being prepared to rip the code out again.

python/lsst/daf/butler/cli/utils.py
n8pease: (fixed)

n8pease:
Ok, sounds fine. I just started a Jenkins build, once that's complete I'll check it in. I will start working on the community post tomorrow morning.

PR #348, opened 2020-08-11
DM-26008: Use new Config._fromPython to deserialize the pex_config files

timj:
Requires lsst/pex_config#59


ktlim (approved):
Don't you need to add `pex_config` to the table file then, or can this only be invoked if it has already been setup somewhere else?

timj:
I just pushed a fix to not use pexConfig in the tests.  Ideally we need to test this but that really means testing it in obs_base and moving the formatter to obs_base with all the other lsst-specific formatters. I need a new ticket for that.

ktlim (approved):

PR #349, opened 2020-08-11
DM-26301: Database work and cleanups in support of CALIBRATION collections

TallJimbo:


python/lsst/daf/butler/registry/interfaces/_versioning.py
```diff
@@ -173,7 +177,7 @@ def tableSchemaRepr(table: sqlalchemy.schema.Table) -> str:
             tableSchemaRepr = [table.name]
             schemaReps = []
             for column in table.columns:
-                columnRep = f"COL,{column.name},{column.type}"
+                columnRep = f"COL,{column.name},{column.type.compile(dialect=dialect)}"
```
andy-slac, line 180: It is likely the right thing to do but I'm not sure how this changes things. I presume that we always use the same dialect for one database and type mapping within the same dialect is always stable so it probably makes no difference. Or do I miss something?

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -335,31 +337,86 @@ def fromConnection(cls, connection: sqlalchemy.engine.Connection, *, origin: int
...
+        # `Connection.in_nested_transaction()` method.
+        savepoint = savepoint or self._connection.info.get(_IN_SAVEPOINT_TRANSACTION, False)
+        self._connection.info[_IN_SAVEPOINT_TRANSACTION] = savepoint
+        if self._connection.in_transaction() and savepoint:
             trans = self._connection.begin_nested()
         else:
-            # Use a regular (non-savepoint) transaction only for the outermost
-            # context.
+            # Use a regular (non-savepoint) transaction always for the
+            # outermost context.
```
andy-slac: I think this comment is only partially true, this branch is not only for outermost context, but also for inner ones without savepoint.

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -335,31 +337,86 @@ def fromConnection(cls, connection: sqlalchemy.engine.Connection, *, origin: int
...
+
+        Parameters
+        ----------
+        tables : `Iterable` [ `sqlalchemy.schema.Table` ], optional
+            A list of tables to lock for the duration of this transaction.
+            These locks are guaranteed to prevent concurrent writes, but only
+            prevent concurrent reads if the database engine requires that in
+            order to block concurrent writes.
+        """
+        raise NotImplementedError()
```
andy-slac, line 419: General comment about table locking - it is not a standard SQL concept and implementation across backends can be inconsistent. MySQL in particular (I'm not sure if we care about it much) has a nasty interaction between transactions and LOCK TABLES. In some cases it may be possible to use SERIALIZABLE transaction isolation to the same effect as LOCK TABLE, but that is case by case consideration. And with LOCK TABLE things are usually much faster for the client that can lock tables but it can also slow down all other concurrent clients.

andy-slac (approved):
Looks OK, few minor comments.

python/lsst/daf/butler/registry/interfaces/_versioning.py
TallJimbo: I added this because (on a commit that's not on this branch) I switched to using a PostgreSQL-only `int8range` type for timespans, and the previous code here raised an exception because there was no stringification for that in the generic dialect that is used by default.  I think you're right that in other cases it makes no difference, except _maybe_ if a SQLAlchemy upgrade changed what dialect-specific type it mapped the a generic type to - but I imagine that would be considered a major breaking change in SQLAlchemy.

python/lsst/daf/butler/registry/interfaces/_database.py
TallJimbo: I certainly plan to use this as little as possible, but it seems to be the only way to make certain multi-statement operations atomic while guaranteeing invariants that can't be written as constraints.

My hope was that the very specific and limited guarantees made by this interface would be something all backends could be expected to support.  It sounds like that's at least plausible for MySQL, even if it might be a pain, and that's good enough for me right now.

PR #350, opened 2020-08-12
DM-24575: Add some more exposure metadata fields

timj:


python/lsst/daf/butler/configs/dimensions.yaml
```diff
@@ -211,6 +211,35 @@ dimensions:
           doc: >
             Integer group identifier associated with this exposure by the
             acquisition system.
+        -
+          name: object
+          type: string
+          length: 64
+          doc: Object of interest for this observation or survey field name.
```
TallJimbo, line 236: This is probably not worth fighting, because I get how ubiquitous it is... but "object" is _super_ overloaded everywhere.  Is "target" taken, or does that imply something different in an important context?

TallJimbo (approved):
I think we should add all of these to `visit` as well, and add code to `DefineVisitsTask` to copy/average/whatever them from the constituent exposures.

As it is, an expanded data ID for `calexp` or `src` can't contain any of these values, because those are all `visit`-based, not `exposure`-based, and that's a bit of a problem, because people are probably interested in that metadata at least as often as they're interested in exposure metadata.

python/lsst/daf/butler/configs/dimensions.yaml
timj: I'm confused because I made the change to target_name this morning.... Maybe I didn't push. I also added them to visit.

timj: Yes, that's exactly what I did. Sorry. If you look on the obs_base ticket you'll see that it has target_name.

timj: Do you mean also to add ra/dec?

PR #351, opened 2020-08-13
Add WebDAV support to daf_butler

bgounon:
This is a new PR following #345 (now closed), after setting the new branch name.
Feature: Add WebDAV support to ButlerURI and add a webdav datastore.

### DIFF
--- a/doc/lsst.daf.butler/index.rst
+++ b/doc/lsst.daf.butler/index.rst
@@ -97,6 +97,9 @@ Example datastores
 .. automodapi:: lsst.daf.butler.datastores.s3Datastore
    :no-main-docstr:
    :headings: ^"
+.. automodapi:: lsst.daf.butler.datastores.webdavDatastore
+   :no-main-docstr:
+   :headings: ^"
 
 Example formatters
 ------------------
--- a/python/lsst/daf/butler/_butlerConfig.py
+++ b/python/lsst/daf/butler/_butlerConfig.py
@@ -93,6 +93,11 @@ class ButlerConfig(Config):
                     uri = ButlerURI(other, forceDirectory=True)
                 uri.updateFile("butler.yaml")
                 other = uri.geturl()
+            elif uri.scheme == "https":
+                if not uri.dirLike and "." not in uri.basename():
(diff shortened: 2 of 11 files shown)


bgounon:
@timj I think the current version addresses all your requests from previous PR (except the ones you annotated as "notes to self")
Except for one thing that I haven't figured out yet:
> Also, is it possible to mock things so that you can add a proper datastore test? It would be great if you could add a webdav test class to test_butler.py like we do for s3 so we can be a bit more confident that things work.

I'm not sure how to correctly mock things. I tried using the `responses` library to mock all responses from requests inside a `WebdavDatastoreButlerTestCase(FileLikeDatastoreButlerTests, unittest.TestCase)` class, as I'm doing in `test_uri.py` for example, but I'm having difficulties with this since I can't seem to find a way to activate response mocking on tests implemented in parent classes (usually done by setting the `@responses.activate` decorator on test methods). Any ideas on how to do this properly are welcome!

timj:
If you rebase your branch I can merge it (we always require rebase -- not merge from master to branch and back to master).

We can defer the webdav datastore testing for now -- although it's clearly useful since when I refactor things I will have no idea if I'm breaking your webdav interface without a mock. We use moto for s3 and it seems to work well (it lets us add a decorator to the entire test class). Is there any way for requests/responses to create a webdav service on localhost that can be used?

bgounon:
> If you rebase your branch I can merge it (we always require rebase -- not merge from master to branch and back to master).
> 
> We can defer the webdav datastore testing for now -- although it's clearly useful since when I refactor things I will have no idea if I'm breaking your webdav interface without a mock. We use moto for s3 and it seems to work well (it lets us add a decorator to the entire test class). Is there any way for requests/responses to create a webdav service on localhost that can be used?

Rebase done.

I've seen what is done with moto but couldn't find anything similar to use with webdav. Response is able to catch requests as they are made in the context of the class (in methods where the decorator is used), so it looked perfect for simple tests, but it shows its limits when testing complex HTTP interactions. I guess working with a local webdav service, dynamically started when testing could be a solution, but it will most likely add a new dependency on the webdav library we will need to use. I will look into it in the upcoming days.

PR #354, opened 2020-08-17
DM-26343: Use less expansive definition of extension

timj:
* Change ButlerURI.getExtension to return a single extension unless that extension matches a specific set (e.g. .gz).
* Change extension validation in Formatter to check the end of the file name directly rather than trying to use the source file extension (which is an unnecessary extra step).

These changes allow "flat_i_sim_1.4_blah.fits" to be ingested as a fits file.

### DIFF
--- a/python/lsst/daf/butler/core/_butlerUri.py
+++ b/python/lsst/daf/butler/core/_butlerUri.py
@@ -434,12 +434,25 @@ class ButlerURI:
         -------
         ext : `str`
             The file extension (including the ``.``). Can be empty string
-            if there is no file extension. Will return all file extensions
-            as a single extension such that ``file.fits.gz`` will return
-            a value of ``.fits.gz``.
+            if there is no file extension. Usually returns only the last
+            file extension unless there is a special extension modifier
+            indicating file compression, in which case the combined
+            extension (e.g. ``.fits.gz``) will be returned.
         """
+        special = {".gz", ".bz2", ".xz", ".fz"}
+
         extensions = self._pathLib(self.path).suffixes
-        return "".join(extensions)
+
+        if not extensions:
+            return ""
+
+        ext = extensions.pop()
+
+        # Multiple extensions, decide whether to include the final two
+        if extensions and ext in special:
+            ext = f"{extensions[-1]}{ext}"
+
+        return ext
 
     def join(self, path: str) -> ButlerURI:
         """Create a new `ButlerURI` with additional path components including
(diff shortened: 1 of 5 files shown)


erykoff:
Before I leave a review, is there a completely different path for `.fz` or is that an oversight here?

timj:
I wasn't sure what to do with `.fits.fz` since I could cover it completely by declaring that `.fz` is a supported extension in obs_base `FitsExposureFormatter`. There is no other usage of `.fz` I think.  That does remind me that I need to make a one line patch to the formatter in obs_base to make that explicit.   I can add `.fz` here as a special extension if you like.

python/lsst/daf/butler/core/_butlerUri.py
```diff
@@ -434,12 +434,25 @@ def getExtension(self) -> str:
...
             The file extension (including the ``.``). Can be empty string
-            if there is no file extension. Will return all file extensions
-            as a single extension such that ``file.fits.gz`` will return
-            a value of ``.fits.gz``.
+            if there is no file extension. Usually returns only the last
+            file extension unless there is a special extension modifier
+            indicating file compression, in which case the combined
+            extension (e.g. ``.fits.gz``) will be returned.
         """
+        special = {".gz", ".bz2", ".xz"}
```
erykoff: I think I see that `.fz` is not "special" because it is handled natively by the fits formatter.  Which is actually good, because then we can support `fits.fz` or `.fz` which I think are both acceptable.

tests/test_location.py
```diff
@@ -181,6 +181,22 @@ def testButlerUriSerialization(self):
...
+        """Test extension extraction."""
+
+        files = (("file.fits.gz", ".fits.gz"),
+                 ("file.fits", ".fits"),
+                 ("file.fits.xz", ".fits.xz"),
+                 ("file.fits.tar", ".tar"),
+                 ("file", ""),
+                 ("flat_i_sim_1.4_blah.fits.gz", ".fits.gz"),
+                 ("flat_i_sim_1.4_blah.txt", ".txt"),
+                 )
```
erykoff, line 196: If we do support `.fits` and `.fits.fz` then I think that we should explicitly test that here.

erykoff (approved):

python/lsst/daf/butler/core/_butlerUri.py
timj: Yes, at the moment even though `.fits.fz` is listed as a supported extension in the formatter it's irrelevant because the `.fz` on its own is going to be enough to match.

tests/test_location.py
timj: Since `.fz` is not listed as a special extension at the moment getExtension() will return `.fz` for `.fits.fz`.

erykoff: I think that's fine!  I would suggest putting those in the list so that it's clear what it will return (and that we know it's returning what we think it should return).

PR #355, opened 2020-08-19
DM-26378: Refactor S3/WebDAV datastores

timj:
Implements a remote datastore entirely in terms of ButlerURI so there is now no special datastore for S3 or WebDAV (apart from the stub classes).

cc/ @bgounon

PR #356, opened 2020-08-20
DM-26383: Add webdav datastore tests to daf_butler

bgounon:
Following merged PR #351, here we're adding explicit WebdavDatastore tests. This requires starting a local webdav server against which requests are run. Adds a dependency to `wsgidav` and `cheroot` packages, for testing only. Tests are skipped if dependencies cannot be met.
Also implementing ButlerURI.parent() method, which returns the parent directory of the current ButlerURI (same as dirname() for a file-like URI).

### DIFF
--- a/python/lsst/daf/butler/datastores/fileLikeDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileLikeDatastore.py
@@ -710,6 +710,9 @@ class FileLikeDatastore(GenericBaseDatastore):
             # Work out the name we want this ingested file to have
             # inside the datastore
             tgtLocation = self._calculate_ingested_datastore_name(srcUri, ref, formatter)
+            if not tgtLocation.uri.dirname().exists():
+                log.debug("Folder %s does not exist yet.", tgtLocation.uri.dirname().geturl())
+                tgtLocation.uri.dirname().mkdir()
 
             # if we are transferring from a local file to a remote location
             # it may be more efficient to get the size and checksum of the
--- a/python/lsst/daf/butler/datastores/fileLikeDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileLikeDatastore.py
@@ -711,7 +711,7 @@ class FileLikeDatastore(GenericBaseDatastore):
             # inside the datastore
             tgtLocation = self._calculate_ingested_datastore_name(srcUri, ref, formatter)
             if not tgtLocation.uri.dirname().exists():
-                log.debug("Folder %s does not exist yet.", tgtLocation.uri.dirname().geturl())
+                log.debug("Folder %s does not exist yet.", tgtLocation.uri.dirname())
                 tgtLocation.uri.dirname().mkdir()
 
             # if we are transferring from a local file to a remote location
--- a/tests/config/basic/webdavDatastore.yaml
+++ b/tests/config/basic/webdavDatastore.yaml
@@ -1,8 +1,6 @@
 datastore:
   cls: lsst.daf.butler.datastores.webdavDatastore.WebdavDatastore
-  # http://localhost:8080 is where the local webdav server listens
-  # during Butler unit tests, so this value is required for successful testing
-  root: http://localhost:8080/butlerRoot
+  root: http://anywebdavserver/butlerRoot
   templates: !include templates.yaml
   formatters: !include formatters.yaml
   composites: !include composites.yaml
--- a/tests/config/basic/webdavDatastore.yaml
+++ b/tests/config/basic/webdavDatastore.yaml
@@ -1,6 +1,6 @@
 datastore:
   cls: lsst.daf.butler.datastores.webdavDatastore.WebdavDatastore
-  root: https://some.webdav.server/butlerRoot
+  root: http://localhost:8080/butlerRoot
   templates: !include templates.yaml
   formatters: !include formatters.yaml
   composites: !include composites.yaml
--- a/tests/config/basic/webdavDatastore.yaml
+++ b/tests/config/basic/webdavDatastore.yaml
@@ -1,5 +1,7 @@
 datastore:
   cls: lsst.daf.butler.datastores.webdavDatastore.WebdavDatastore
+  # http://localhost:8080 is where the local webdav server listens
+  # during Butler unit tests, so this value is required for successful testing
   root: http://localhost:8080/butlerRoot
   templates: !include templates.yaml
   formatters: !include formatters.yaml
--- a/tests/test_butler.py
+++ b/tests/test_butler.py
@@ -44,8 +44,15 @@ except ImportError:
         """
         return cls
 
+try:
+    from cheroot import wsgi
+    from wsgidav.wsgidav_app import WsgiDAVApp
+except ImportError:
+    WsgiDAVApp = None
+
 import astropy.time
 from threading import Thread
+from tempfile import gettempdir
 from lsst.utils import doImport
 from lsst.daf.butler.core.utils import safeMakeDir
 from lsst.daf.butler import Butler, Config, ButlerConfig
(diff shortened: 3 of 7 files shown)


python/lsst/daf/butler/_butlerConfig.py
```diff
@@ -93,7 +93,7 @@ def __init__(self, other=None, searchPaths=None):
                     uri = ButlerURI(other, forceDirectory=True)
                 uri.updateFile("butler.yaml")
                 other = uri.geturl()
-            elif uri.scheme == "https":
+            elif uri.scheme.startswith("http"):
```
timj: This change should no longer be needed since I fixed that code in #355

python/lsst/daf/butler/datastores/webdavDatastore.py
```diff
@@ -181,6 +181,9 @@ def _write_in_memory_to_artifact(self, inMemoryDataset: Any, ref: DatasetRef) ->
             raise FileExistsError(f"Cannot write file for ref {ref} as "
                                   f"output file {location.uri} exists.")
 
+        if not location.uri.dirname().exists():
```
timj: Need to consider how we deal with this in the new RemoteFileDatastore that has replaced this implementation. For now we should probably just add it and absorb the one extra S3 call.

tests/config/basic/webdavDatastore.yaml
```diff
@@ -1,6 +1,6 @@
 datastore:
   cls: lsst.daf.butler.datastores.webdavDatastore.WebdavDatastore
-  root: https://some.webdav.server/butlerRoot
+  root: http://localhost:8080/butlerRoot
```
timj: Maybe add a comment to the YAML here noting that this is an explicit URL for the test server rather than something random.

tests/test_butler.py
```diff
@@ -1193,5 +1202,117 @@ def tearDown(self):
...
+        config.update({"datastore": {"datastore": {"root": rooturi}}})
+
+        self.datastoreStr = f"datastore={self.root}"
+        self.datastoreName = [f"WebdavDatastore@{rooturi}"]
+
+        Butler.makeRepo(rooturi, config=config, forceConfigRoot=False)
+        self.tmpConfigFile = posixpath.join(rooturi, "butler.yaml")
+
+    def tearDown(self):
+        rooturi = f"http://{self.serverName}/{self.root}"
```
timj: Might be safer to store this in the object in `setUp` so you can guarantee consistency in `rooturi`.

tests/test_butler.py
```diff
@@ -1193,5 +1202,117 @@ def tearDown(self):
...
+
+        if self.useTempRoot:
+            self.root = self.genRoot()
+        rooturi = f"http://{self.serverName}/{self.root}"
+        config.update({"datastore": {"datastore": {"root": rooturi}}})
+
+        self.datastoreStr = f"datastore={self.root}"
+        self.datastoreName = [f"WebdavDatastore@{rooturi}"]
+
+        Butler.makeRepo(rooturi, config=config, forceConfigRoot=False)
```
timj: `makeRepo` can take a `ButlerURI` (but make sure you use `forceDirectory`) if that makes things easier for you.

tests/test_butler.py
```diff
@@ -1193,5 +1202,117 @@ def tearDown(self):
...
+        """Starts a local webdav-compatible HTTP server,
+        Listening on http://localhost:8080
+        This server only runs when this test class is instantiated,
+        and then shuts down. Must be started is a separate thread.
+        """
+        root_path = gettempdir()
+
+        config = {
+            "host": "0.0.0.0",
+            "port": 8080,
```
timj: Do we have to use 8080? Does it deal with people who are running local web servers there already?  If you run the tests with `pytest -n5 tests/test_butler.py` won't it start multiple web servers, one for each subprocess? Don't they clash?

tests/test_butler.py
```diff
@@ -1193,5 +1202,117 @@ def tearDown(self):
...
+    def tearDown(self):
+        rooturi = f"http://{self.serverName}/{self.root}"
+        uri = ButlerURI(rooturi)
+        uri.remove()
+
+    def _serveWebdav():
+        """Starts a local webdav-compatible HTTP server,
+        Listening on http://localhost:8080
+        This server only runs when this test class is instantiated,
+        and then shuts down. Must be started is a separate thread.
```
timj, line 1318: What causes it to shut down? Can it be shutdown in `tearDownClass` ?

tests/test_uri.py
```diff
@@ -358,11 +358,16 @@ def testQuoting(self):
 class WebdavURITestCase(unittest.TestCase):
 
     def setUp(self):
+        os.environ["WEBDAV_AUTH_METHOD"] = "TOKEN"
```
timj: I think you should use unittest.mock.patch here to patch the environment so we know we aren't contaminating other tests that might be called later.  I haven't entirely thought about how this can be done without having the env set for each test separately.

timj (approved):
This looks good. I do have some comments that should be thought about before I can merge. If you are going to do further work I will add you to the daf_butler repo as a collaborator.

tests/test_butler.py
bgounon: Here we use a `daemon` thread, which is automatically killed once the main program finishes its execution.
Another solution would be to make the thread non-daemon and explicitly ask it to terminate in tearDown(), but I think the result would be identical.

tests/test_butler.py
bgounon: We don't necessarily have to use 8080, any port above 1024 is probably fine (we may run into permissions issues if using any port below 1024).
I've added some necessary checks : if there is another service running on the selected port, the webdav server will fail to start and all tests will raise an OSError and fail.

tests/test_uri.py
bgounon: Done in ebf3ebe.
`@unittest.mock.patch.dict` class decorator only applies to methods with `test` prefix in the name, which means it doesn't apply to setUp() and tearDown() for example: I had to manually decorate those methods in test_butler.py.

bgounon:
After rebase, I'm having trouble with `_extractIngestInfo` in `fileLikeDatastore` not transferring files as expected, I'll update the PR when this is figured out.

edit: the issue was with using POST (privously we used PUT) to transfer a local file to the webdav repository. Moving from POST to PUT fixed it.

tests/test_butler.py
bgounon: Running pytest with `-n5` for example does not seem to be an issue and succeeds. I'm not sure if it's something related to how `pytest` works with multithread, or if it's just the fact that, once the first webdav server is up, subsequent webdav server threads silently fail and tests all run against the first one.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -710,6 +710,9 @@ def _extractIngestInfo(self, path: Union[str, ButlerURI], ref: DatasetRef, *,
             # Work out the name we want this ingested file to have
             # inside the datastore
             tgtLocation = self._calculate_ingested_datastore_name(srcUri, ref, formatter)
+            if not tgtLocation.uri.dirname().exists():
+                log.debug("Folder %s does not exist yet.", tgtLocation.uri.dirname().geturl())
```
timj: I don't think you need the `geturl()` here because there is a `__str__` method that does this.

tests/test_butler.py
timj: I wonder if using port 0 works (it depends on whether you can ask the server what port it has been assigned).

timj: It seems plausible that they are sharing the same server. This seems a bit risky if the first subprocess completes before the others have finished talking to it. I imagine we are lucking out because the webdav tests are a small fraction of all the tests.

tests/test_butler.py
timj: If we do kill it in tearDownClass then that might make it more obvious that the other subprocesses are using it.

bgounon:
@timj I think this new version addresses all your observations.
In particular, the webdav server now dynamically chooses a free port on which to listen. This allows for multiple webdav servers to start simultaneously without interfering with each other (for example when using `pytest -nX`).
Also, the server is now gracefully shut down in tearDownClass instead of using daemon threads.

PR #360, opened 2020-08-25
DM-26445: Fix Timespan equality to be fuzzy

timj:


MichelleGower (approved):
String changes made debugging easier.  The new fuzzy checks for equal and not equal fixed the raw ingestion problem.

PR #363, opened 2020-09-01
DM-25760: add --glob to butler query-collections

n8pease:


n8pease:
I see the current linter errors. I have to run out for about an hour and will address these when I get back.

python/lsst/daf/butler/core/utils.py
```diff
@@ -344,3 +345,19 @@ def findFileResources(values: Iterable[str], regex: Optional[str] = None) -> Lis
...
+    """Translate glob-style search terms to regex.
+
+    Parameters
+    ----------
+    expressions : `list` [`str` or `...`]
+        A glob-style pattern string to convert, or an Ellipsis.
+
+    Returns
+    -------
+    expressions : [`str` or `...`]
```
timj: This docstring is wrong now.

python/lsst/daf/butler/core/utils.py
```diff
@@ -344,3 +345,19 @@ def findFileResources(values: Iterable[str], regex: Optional[str] = None) -> Lis
...
     return resources
+
+
+def globToRegex(expressions: List[Union[str, Ellipsis]]):
+    """Translate glob-style search terms to regex.
+
+    Parameters
+    ----------
+    expressions : `list` [`str` or `...`]
+        A glob-style pattern string to convert, or an Ellipsis.
```
timj: This docstring is now wrong (I saw the code originally had ellipsis and I had a long comment on not understanding that).

n8pease:
(all checks are passing now)

python/lsst/daf/butler/script/queryCollections.py
```diff
@@ -47,7 +51,11 @@ def queryCollections(repo, collection_type, flatten_chains, include_chains):
         collection names.
     """
     butler = Butler(repo)
+    kwargs = {}
+    if glob:
+        kwargs['expression'] = globToRegex(glob)
```
timj: From your earlier version handling of `...` for this did make me wonder whether there is an efficiency gain for globToRegex to spot `*` as a specil case -- it probably should warn if `*` is passed in with other globs and if they just pass in `*` it should probably return an empty list rather than a regex if we assume that empty list means no constraints (that would save a lot of unnecessary regex processing when we know there is no filtering).

python/lsst/daf/butler/script/queryCollections.py
```diff
@@ -47,7 +51,11 @@ def queryCollections(repo, collection_type, flatten_chains, include_chains):
         collection names.
     """
     butler = Butler(repo)
+    kwargs = {}
+    if glob:
```
timj: since globToRegex can take an empty list it might be clearer to write this as:

```python
filter = globToRegex(glob)
if filter:
    kwargs["expression"] = filter
```

python/lsst/daf/butler/script/queryCollections.py
```diff
@@ -47,7 +51,11 @@ def queryCollections(repo, collection_type, flatten_chains, include_chains):
...
+        kwargs['expression'] = globToRegex(glob)
     collections = butler.registry.queryCollections(collectionType=collection_type,
                                                    flattenChains=flatten_chains,
-                                                   includeChains=include_chains)
+                                                   includeChains=include_chains,
+                                                   **kwargs)
     return {'collections': list(collections)}
```
timj: Single quotes have crept into this file.

tests/test_utils.py
```diff
@@ -202,5 +204,21 @@ def test_multipleInputs(self):
...
+    def testGlobList(self):
+        """Test that a list of glob strings converts as expected to a regex and
+        returns in the expected list.
+
+        This assumes the conversion algorithm of globToRegex produces the same
+        results as in the right side of the assert, and verifies the
+        globToRegex function api.
+        """
+        testval = ["foo*", "bar"]
+        self.assertEqual(globToRegex(testval),
```
timj: I think a better test is to check that the result of globToRegex really does work properly. ie does it return something that matches `bar` and not `bars` or `abar` and matches `food` and not `flood`.

timj (approved):
Looks okay but I think I'd tweak the final test to check for outcomes and not specifics, and I think we should consider having a special case for `*` glob.

PR #364, opened 2020-09-01
DM-26568: Fix ButlerURI.exists for root S3 URIs

timj:


tests/test_location.py
```diff
@@ -148,6 +149,25 @@ def testButlerUri(self):
         uri4 = copy.deepcopy(uri3)
         self.assertEqual(uri4, uri3)
 
+    def testUriRoot(self):
+        os_path_root = pathlib.Path(__file__).absolute().root
```
hsinfang: I'm a bit confused with the mixed use of snake case and camel case in this test.  Isn't our convention still camelCase for variable names?

hsinfang (approved):

tests/test_location.py
timj: Yes, okay. daf_butler is trying to be snake case internally but sometimes the two desires come into tension. I've made consistent.

PR #367, opened 2020-09-04
DM-26620: Catch astropy.time warnings about dubious years

timj:
Any times from more than 5 years in the future will trigger
warnings when converting between UTC and TAI. Catch the places
where we are converting to TAI since those warnings are not
relevant to users (especially when they involve MAX_TIME
or simulated data from the future).

### DIFF
--- a/tests/test_time_utils.py
+++ b/tests/test_time_utils.py
@@ -25,6 +25,7 @@
 import unittest
 import warnings
 
+import astropy.utils.exceptions
 from astropy.time import Time, TimeDelta
 from lsst.daf.butler import time_utils
 from lsst.daf.butler.core.time_utils import astropy_to_nsec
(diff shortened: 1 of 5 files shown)


tests/test_time_utils.py
```diff
@@ -53,6 +54,17 @@ def test_max_time(self):
         value_max = time_utils.astropy_to_nsec(time_utils.MAX_TIME)
         self.assertEqual(value, value_max)
 
+        # Check that we do not warn inside our code for UTC in the future
+        with self.assertWarns(Warning):
+            time = Time("2101-01-01T00:00:00", format="isot", scale="utc")
```
andy-slac: I'm not sure I can parse that comment, the code seems to be testing that astropy makes warning here, but there is nothing for "our code"?

andy-slac (approved):
Looks good, one minor comment.

tests/test_time_utils.py
timj: I'm mostly hiding the warnings from pytest. If you'd rather I can put the warnings context manager and simple warnings filter in. Warning here at least tells us that there is a warning to hide later on.

andy-slac: No, I 'm happy with what this code does, it's just comment that confuses me. When I read it initially I understood "our code" as the code in daf_butler package but now I read it again and I think you mean next two lines by "our code".

andy-slac: Also there is a chance that in 100 years or less astropy will stop making warnings about 2101 and this unit test will break 🙂

timj: I fixed it properly.

PR #368, opened 2020-09-07
DM-26336: minor non-prototyping changes generated during query/dimensions prototyping

TallJimbo:
These commits are the only ones worth merging now from a branch that was mostly about prototyping new code without any effort to make it work.  They're mostly improvements to the named container module, which had ABCs for its mapping class but only concrete Set class; that's now been rectified.

### DIFF
--- a/python/lsst/daf/butler/configs/dimensions.yaml
+++ b/python/lsst/daf/butler/configs/dimensions.yaml
@@ -1,13 +1,14 @@
 dimensions:
   version: 0
   skypix:
+    # 'common' is the skypix system and level used to relate all other spatial
+    # dimensions.  Its value is a string formed by concatenating one of the
+    # other keys under the 'skypix' headerin (i.e. the name of a skypix system)
+    # with an integer level (with no zero-padding).
     common: htm7
-    htm7:
+    htm:
       class: lsst.sphgeom.HtmPixelization
-      level: 7
-    htm9:
-      class: lsst.sphgeom.HtmPixelization
-      level: 9
+      max_level: 24
 
   elements:
     instrument:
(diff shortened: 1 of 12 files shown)


python/lsst/daf/butler/configs/dimensions.yaml
```diff
@@ -2,12 +2,9 @@ dimensions:
   version: 0
   skypix:
     common: htm7
-    htm7:
+    htm:
```
andy-slac, line 9: A little bit confusing that you rename `htm7` to `htm` but keep `htm7` for `common`. I see how it works in Python code, so it should be OK if it's documented clearly.

python/lsst/daf/butler/core/dimensions/universe.py
```diff
@@ -144,11 +139,114 @@ def __new__(cls, config: Optional[Config] = None) -> DimensionUniverse:
...
+        Returns
+        -------
+        index : `int`
+            Sorting index for this dimension.
+
+        Notes
+        -----
+        The dimension sort order for a universe is consistent with the element
+        order (all dimensions are elements), and either can be used to sort
+        dimensions if used consistently.
```
andy-slac: If dimensions are elements do you need separate indices for them instead of reusing the element indices. If you only care about order and not particular values then it may be easier to guarantee consistency by having only one index.

andy-slac (approved):

python/lsst/daf/butler/core/dimensions/universe.py
TallJimbo: We also use the dimension-only indices in other places; I'll clarify that in the comment.

python/lsst/daf/butler/configs/dimensions.yaml
TallJimbo: Good point.  I hope to get rid of the notion of "common skypix" soon (with materialized overlap tables it should be unnecessary), so I think I will just document this with a comment in the YAML right now.

PR #369, opened 2020-09-08
DM-26333: add partial support for CALIBRATION collections

TallJimbo:
I'm creating this as a draft PR because it includes a breaking schema change without adding anew new functionality, and it will need to be accompanied by a testdata_ap_verify change (at least) before merging.  But it is otherwise ready for review, and I think it makes sense to get that out of the way now.

### DIFF
--- a/python/lsst/daf/butler/registry/_registry.py
+++ b/python/lsst/daf/butler/registry/_registry.py
@@ -60,6 +60,7 @@ from ..core import (
     NamedKeyMapping,
     NameLookupMapping,
     StorageClassFactory,
+    Timespan,
 )
 from . import queries
 from ..core.utils import doImport, iterable, transactional
@@ -518,7 +519,8 @@ class Registry:
         return self._datasets[name].datasetType
 
     def findDataset(self, datasetType: Union[DatasetType, str], dataId: Optional[DataId] = None, *,
-                    collections: Any, **kwargs: Any) -> Optional[DatasetRef]:
+                    collections: Any, timespan: Optional[Timespan] = None,
+                    **kwargs: Any) -> Optional[DatasetRef]:
         """Find a dataset given its `DatasetType` and data ID.
 
         This can be used to obtain a `DatasetRef` that permits the dataset to
@@ -538,6 +540,10 @@ class Registry:
             to search for the dataset, such as a `str`, `DatasetType`, or
             iterable  thereof.  See :ref:`daf_butler_collection_expressions`
             for more information.
+        timespan : `Timespan`, optional
+            A timespan that the validity range of the dataset must overlap.
+            If not provided, any `~CollectionType.CALIBRATION` collections
+            matched by the ``collections`` argument will not be searched.
         **kwargs
             Additional keyword arguments passed to
             `DataCoordinate.standardize` to convert ``dataId`` to a true
@@ -557,6 +563,18 @@ class Registry:
             Raised if the dataset type does not exist.
         MissingCollectionError
             Raised if any of ``collections`` does not exist in the registry.
+
+        Notes
+        -----
+        This method simply returns `None` and does not raise an exception even
+        when the set of collections searched is intrinsically incompatible with
+        the dataset type, e.g. if ``datasetType.isCalibration() is False``, but
+        only `~CollectionType.CALIBRATION` collections are being searched.
+        This may make it harder to debug some lookup failures, but the behavior
+        is intentional; we consider it more important that failed searches are
+        reported consistently, regardless of the reason, and that adding
+        additional collections that do not contain a match to the search path
+        never changes the behavior.
         """
         if isinstance(datasetType, DatasetType):
             storage = self._datasets[datasetType.name]
@@ -566,7 +584,10 @@ class Registry:
                                             universe=self.dimensions, **kwargs)
         collections = CollectionSearch.fromExpression(collections)
         for collectionRecord in collections.iter(self._collections, datasetType=storage.datasetType):
-            result = storage.find(collectionRecord, dataId)
+            if (collectionRecord.type is CollectionType.CALIBRATION
+                    and (not storage.datasetType.isCalibration() or timespan is None)):
+                continue
+            result = storage.find(collectionRecord, dataId, timespan=timespan)
             if result is not None:
                 return result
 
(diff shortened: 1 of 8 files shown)


python/lsst/daf/butler/core/datasets/type.py
```diff
@@ -91,10 +91,14 @@ class DatasetType:
     universe : `DimensionUniverse`, optional
         Set of all known dimensions, used to normalize ``dimensions`` if it
         is not already a `DimensionGraph`.
+    isCalibration : `bool`, optional
+        If `True`, this dataset type may be included in
+        `~CollectionType.CALIBRATION~ collections.
```
andy-slac: Trailing `~` should be backtick?

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -759,6 +780,80 @@ def disassociate(self, collection: str, refs: Iterable[DatasetRef]) -> None:
...
+        dataIds : `Iterable` [ `DataId` ], optional
+            Data IDs that should be decertified within the given validity range
+            If `None`, all data IDs for ``self.datasetType`` will be
+            decertified.
+
+        Raises
+        ------
+        TypeError
+            Raised if
+            ``collectionRecord.type is not CollectionType.CALIBRATION`` or if
```
andy-slac: `collectionRecord` is not in the parameters, it may confuse people who don't read the code. I think `certify` description for the same exception can be copied here too.

python/lsst/daf/butler/registry/datasets/byDimensions/tables.py
```diff
@@ -24,15 +24,17 @@
 __all__ = (
     "addDatasetForeignKey",
     "makeStaticTableSpecs",
-    "makeDynamicTableName",
-    "makeDynamicTableSpec",
+    "makeTagTableName",
+    "makeTagTableSpec",
```
andy-slac: `makeCalibTableName` and `makeCalibTableSpec` should be here too?

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
```diff
@@ -137,13 +164,148 @@ def disassociate(self, collection: CollectionRecord, datasets: Iterable[DatasetR
...
+        # The insert rows will have the same values for collection and
+        # dataset type.
+        protoInsertRow = {
+            self._collections.getCollectionForeignKeyName(): collection.key,
+            "dataset_type_id": self._dataset_type_id,
+        }
+        rowsToDelete = []
+        rowsToInsert = []
+        # Acquire a table lock to ensure there are no concurrent writes
+        # between the SELECT and the DELETE and INSERT queries based on it.
```
andy-slac: Usually row-level locking with "SELECT ... FOR UPDATE" is used for such cases to reduce lock granularity, maybe something to optimize later if we see performance issues.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -41,7 +48,7 @@
 
 
 # This has to be updated on every schema change
-_VERSION = VersionTuple(0, 1, 0)
+_VERSION = VersionTuple(2, 0, 0)
```
andy-slac: Skipping 1.x.y?

python/lsst/daf/butler/registry/interfaces/_datasets.py
```diff
@@ -168,11 +174,73 @@ def disassociate(self, collection: CollectionRecord, datasets: Iterable[DatasetR
...
+        ------
+        AmbiguousDatasetError
+            Raised if any of the given `DatasetRef` instances is unresolved.
+        ConflictingDefinitionError
+            Raised if the collection already contains a different dataset with
+            the same `DatasetType` and data ID and an overlapping validity
+            range.
+        TypeError
+            Raised if
+            ``collectionRecord.type is not CollectionType.CALIBRATION`` or if
```
andy-slac: `collectionRecord.type` -> `collection.type`?

python/lsst/daf/butler/registry/interfaces/_datasets.py
```diff
@@ -168,11 +174,73 @@ def disassociate(self, collection: CollectionRecord, datasets: Iterable[DatasetR
...
+        dataIds : `Iterable` [ `DataCoordinate` ], optional
+            Data IDs that should be decertified within the given validity range
+            If `None`, all data IDs for ``self.datasetType`` will be
+            decertified.
+
+        Raises
+        ------
+        TypeError
+            Raised if
+            ``collectionRecord.type is not CollectionType.CALIBRATION``.
```
andy-slac: Same here

tests/test_cliCmdQueryCollections.py
```diff
@@ -37,7 +37,7 @@ class QueryCollectionsCmdTest(CliCmdTestBase, unittest.TestCase):
     @staticmethod
     def defaultExpected():
         return dict(repo=None,
-                    collection_type=None,
+                    collection_type=tuple(CollectionType.__members__.values()),
```
andy-slac: Could use `CollectionType.all()` here?

andy-slac (approved):

python/lsst/daf/butler/core/datasets/type.py
kfindeisen: Please do build and proofread the docs when working on `daf_butler` and `pipe_base` -- the API docs are pretty much our only reference on Middleware at this point, so if they get corrupted that's a problem.

TallJimbo: I was thinking that the GH action at least took care of the "build" part, but this reminds me that while they do, the build doesn't return an error code when it fails, so it's not actually helping.  Anyhow, yes, agreed - it's just a matter of remembering that.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
TallJimbo: There was another commit on the same branch that went to 1.0.0.  I figure it's probably better to increment versions by commit rather than by branch when possible, but I could be persuaded otherwise.

andy-slac: Well, my mental model is that we are effectively squashing commits in the same PR, to me it means that version 1.0.0 will never be officially released and there is no point providing migration from 0.1.0 to 1.0.0 as there is no software that can work with that version. I'm OK with gaps in versions, it may just look odd to people who need to look at version numbers.

tests/test_cliCmdQueryCollections.py
TallJimbo: Hmm, at first I thought so, but while this seems to work, I think whether it works depends on whether `frozenset` preserves order (the way `dict` now does).  It might, but I have a hard time confirming that this behavior is guaranteed (and [a moderately recent Stack Overflow conversation](https://stackoverflow.com/questions/61414947/why-dont-python-sets-preserve-insertion-order) says the opposite with what looks like pretty solid sources).  @natelust, do you know?

natelust: Frozenset does not preserve insertion order (neither does set), if the set grows (memory is reallocated) the order will be different.

TallJimbo:
This has been rebased on DM-26680 (both the branch and the PR) and all review comments have been addressed.  I've also removed the version increments, and will add them back (along with those from previous DM-26680 tickets) to the tip of tickets/DM-26680 after merging.  I've got one more Jenkins run underway I'd like to complete first.

PR #370, opened 2020-09-08
DM-26630: Change to TEXT column for datastore file paths

timj:


python/lsst/daf/butler/core/ddl.py
```diff
@@ -281,6 +298,9 @@ def getSizedColumnType(self) -> sqlalchemy.types.TypeEngine:
             A SQLAlchemy column type object.
         """
         if self.length is not None:
+            # Last chance check that we are only looking at possible String
+            if self.dtype == sqlalchemy.String and not self.isStringType():
```
TallJimbo: I think `self.dtype is sqlalchemy.String` is safe here, and probably preferable.

TallJimbo (approved):

timj:
Closed in #371

PR #373, opened 2020-09-14
DM-26684: Add command-line tool for Butler.pruneCollection

n8pease:


python/lsst/daf/butler/cli/cmd/__init__.py
```diff
@@ -19,9 +19,9 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-__all__ = ["butler_import", "create", "config_dump", "config_validate", "query_collections",
-           "query_dataset_types"]
+__all__ = ["butler_import", "create", "config_dump", "config_validate", "prune_collection",
+           "query_collections", "query_dataset_types"]
```
timj: might make future changes easier if you make one per line so that a new one early in the alphabet doesn't cause multiple lines to reflow.

python/lsst/daf/butler/cli/cmd/__init__.py
```diff
@@ -19,9 +19,9 @@
...
 
 
-from .commands import (butler_import, create, config_dump, config_validate, query_collections,
-                       query_dataset_types)
+from .commands import (butler_import, create, config_dump, config_validate, prune_collection,
```
timj: similar comment about one per line here.

timj (commented):
Looks fine but can we have a test that it's actually deleting a collection? ie delete a collection and then show that queryCollections doesn't list it any more.

timj (approved):

PR #375, opened 2020-09-16
DM-26683: Make exporting dimension data friendlier

TallJimbo:


tests/test_butler.py
```diff
@@ -931,6 +931,9 @@ def runImportExportTest(self, storageClass):
         # Test that the repo actually has at least one dataset.
         datasets = list(exportButler.registry.queryDatasets(..., collections=...))
         self.assertGreater(len(datasets), 0)
+        # Add a DimensionRecord that's unused by those datasets.
+        skymapRecord = {"name": "example_skymap", "hash": (50).to_bytes(8, byteorder="little")}
+        exportButler.registry.insertDimensionData("skymap", skymapRecord)
         # Export those datasets.  We used TemporaryDirectory because there
         # doesn't seem to be a way to get the filename (as opposed to the file
         # object) from any of tempfile's temporary-file context managers.
```
timj: `tempfile.NamedTemporaryFile` ? Also, what does transfer-on-exist mean below?

python/lsst/daf/butler/transfers/_context.py
```diff
@@ -155,14 +150,16 @@ def saveDatasets(self, refs: Iterable[DatasetRef], *,
...
         self.saveDataIds(dataIds, elements=elements)
-        for (datasetType, run), records in datasets.items():
-            self._backend.saveDatasets(datasetType, run, *records)
 
     def _finish(self) -> None:
         """Delegate to the backend to finish the export process.
 
         For use by `Butler.export` only.
         """
+        for element in self._registry.dimensions.sorted(self._records.keys()):
```
timj, line 201: Does this sorting help us out at all with DM-26324 ?

python/lsst/daf/butler/transfers/_context.py
```diff
@@ -155,14 +150,16 @@ def saveDatasets(self, refs: Iterable[DatasetRef], *,
...
+            self._backend.saveDimensionData(element, *self._records[element].values())
+        for (datasetType, run), records in self._datasets.items():
```
timj: Could this benefit from a sort?

timj (approved):
I eventually realized that there were very few changes here but a lot of reorganization. I think this looks fine.

tests/test_butler.py
TallJimbo: I was vaguely hoping you knew what either of these meant when I saw them.  `git blame` has proved my guilt on both.  I'm just going to remove them, because I don't think they're helpful - nobody knows what `transfer-on-exist` means anymore, and I don't see an obvious way to replace temporary directory usage with temporary file usage here (or why that would be better).

python/lsst/daf/butler/transfers/_context.py
TallJimbo: It's a start, but we'd also need to sort the iterable provided on the line below, and that's harder, because we don't have a way to sort data IDs or dimension records (something I realized after my Jira post about trying it out).

python/lsst/daf/butler/transfers/_context.py
TallJimbo: I'll add one.  As in the previous case, we'll need to add sorting in the line below as well to make the order fully deterministic, but we might as well start.

PR #377, opened 2020-09-18
DM-26780: Add --presets for command line bulk overrides via file

timj:


python/lsst/daf/butler/cli/utils.py
```diff
@@ -667,3 +670,50 @@ def getFrom(ctx):
...
+    ctx : `click.context`
+        The context for the click operation. Used to extract the subcommand
+        name.
+    param : `str`
+        The parameter name.
+    value : `object`
+        The value of the parameter.
+    """
+    ctx.default_map = ctx.default_map or {}
+    cmd_name = ctx.info_name
```
n8pease, line 690: `Context.info_name` is weird: when used with a command that is in a Group, as our subcommands are, then it contains the name of the subcommand, and so this works. But if you were to use this callback with the top level command (`butler` or `pipetask`) then it would contain the name of the script, including any extension (`.py`). Currently this would be ok for our commands because the butler script is called `butler` (not `butler.py`), and similar for `pipetask`. But it might pop up with other scripts if they don't omit the extension. Maybe it's worth stripping `.py` off the end here if it exists, or warning if `.py` is found? Or at least it's something to be aware of, someone may hit it someday.

timj: This effectively becomes the key to use in the dict read from the yaml file. Things are all fine so long as this cmd_name is itself mentioned in the yaml file. Given that we are only using this for subcommands I'm inclined to leave it as is at the moment and let things break.

python/lsst/daf/butler/cli/opt/options.py
```diff
@@ -103,3 +103,12 @@ def makeCollectionType(context, param, value):
 verbose_option = MWOptionDecorator("-v", "--verbose",
                                    help="Increase verbosity.",
                                    is_flag=True)
+
+
+file_override_options = MWOptionDecorator("--options-file", "-@",
```
n8pease: I think the name should be `options_file_option`; basically `_option` indicates that it's an option (as opposed to an argument) and `options_file_` matches the name of the flag.

n8pease (approved):
Please change the option name as noted or let's discuss, otherwise it looks goo d.

timj:
I changed the option name. It was wrong before.

python/lsst/daf/butler/cli/opt/options.py
```diff
@@ -103,3 +103,12 @@ def makeCollectionType(context, param, value):
 verbose_option = MWOptionDecorator("-v", "--verbose",
                                    help="Increase verbosity.",
                                    is_flag=True)
+
+
+options_file_option = MWOptionDecorator("--options-file", "-@",
```
n8pease, line 108: of course this adds the `-@` short flag for butler as well. I don't mind either way, but I think it will work if you want to add it only in `pipetask`, where you use the `options_file_option` you can pass it there:
```
@click.command()
@options_file_option("-@")
def build(**kwargs):
    ...
```

timj: I want the short flag available everywhere. I realize that butler is not going to be a problem because the options are small but I didn't think it caused harm by having it be consistent.

PR #379, opened 2020-09-23
DM-26600: Add ability to remove a dataset type

timj:


python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -115,6 +115,22 @@ def refresh(self, *, universe: DimensionUniverse) -> None:
         self._byName = byName
         self._byId = byId
 
+    def deleteDatasetType(self, name: str) -> None:
```
timj: Not really sure what name is wanted here. prune vs delete vs pruneDatasetType vs deleteDatsetType. Should it be consistent with the registry method?

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -115,6 +115,22 @@ def refresh(self, *, universe: DimensionUniverse) -> None:
...
+        if componentName is not None:
+            raise ValueError(f"Cannot delete a dataset type of a component of a composite (given {name})")
+        storage = self._byName.get(compositeName)
+        if storage is None:
+            raise KeyError(f"DatasetType {name} is not registered.")
+
+        # Delete the row
+        self._db.delete(self._static.dataset_type, ["name"], {"name": name})
+
+        # Now delete the cache entries
```
timj: Should I pass in a universe and call `refresh` here instead?

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -115,6 +115,22 @@ def refresh(self, *, universe: DimensionUniverse) -> None:
         self._byName = byName
         self._byId = byId
 
+    def deleteDatasetType(self, name: str) -> None:
+        # Docstring inherited from DatasetRecordStorageManager.
+        compositeName, componentName = DatasetType.splitDatasetTypeName(name)
+        if componentName is not None:
+            raise ValueError(f"Cannot delete a dataset type of a component of a composite (given {name})")
+        storage = self._byName.get(compositeName)
+        if storage is None:
```
timj: Should this return silently?

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -115,6 +115,22 @@ def refresh(self, *, universe: DimensionUniverse) -> None:
...
+            raise KeyError(f"DatasetType {name} is not registered.")
+
+        # Delete the row
+        self._db.delete(self._static.dataset_type, ["name"], {"name": name})
```
timj: Should we catch IntegrityError here and raise something else? Should we even list a collection that uses this dataset type?

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
TallJimbo: I'd at least use the same name here as the registry method.  I think I used `prune` in `Butler.pruneDatasets` and `pruneCollections` because I wanted to signify partial deletion, e.g. untagging from a collection or unstoring from a Datastore, as I figured those would be the most common modes.  It looks like "remove" is beating "delete" 2-1 in other Registry precedent, and the "delete" (for `deleteOpaqueData`) is maybe a justifiable exception because it acts on "data" rather than countable entities.  So maybe "remove"?  I don't have a strong preference, but I gather you don't either.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
TallJimbo: Given that registration is a no-op if the dataset already exists, I suspect we will want deletion to be a no-op if it's already deleted.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
timj: `removeDatasetType` is fine with me. I did wonder whether `remove` would be preferred in the manager because you have a `register` there and not a `registerDatasetType`.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
TallJimbo: Yeah, I would probably catch and raise just to at least say that this is because there is at least one dataset in that collection.  Actually querying for which dataset types for an even better error message is something we ought to do, but there's no easy way to do it now; maybe add a comment to [DM-24939](https://jira.lsstcorp.org/browse/DM-24939) to come back to this there?

TallJimbo: Oh, we should also make sure that catch _doesn't_ catch the case where the delete failed because the dataset type was already gone but this registry client had a stale cache (I don't think that'd be an integrity error, though).

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
TallJimbo: I think this is safe enough, but dataset type deletion ought to be rare enough the cost of `refresh` shouldn't matter, I think.

tests/test_butler.py
```diff
@@ -589,6 +589,21 @@ def testPruneCollections(self):
...
+                                             butler.registry)
+        tmpFromRegistry = butler.registry.getDatasetType(tmpName)
+        self.assertEqual(tmpDatasetType, tmpFromRegistry)
+        butler.registry.pruneDatasetType(tmpName)
+        with self.assertRaises(KeyError):
+            butler.registry.getDatasetType(tmpName)
+
+        # Try and fail to delete a datasetType that is associated with data
+        with self.assertRaises(Exception):
+            butler.registry.pruneDatasetType(datasetType.name)
```
TallJimbo: If you can think of an easy way to do it, it'd also be nice to check here that nothing did get deleted.  I'm not at all worried about the current implementation, but a different caching model might lead to an implementation where that's less obvious.

TallJimbo (approved):
Looks good!  I think you found all of the points where I'd have questions.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
TallJimbo: Ah, yeah, `remove` alone is probably a better match for the storage, given that `register` symmetry (I was only thinking about the initial verb).

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -497,6 +497,32 @@ def registerDatasetType(self, datasetType: DatasetType) -> bool:
...
+        Parameters
+        ----------
+        name : `str`
+            Name of the type to be removed.
+
+        Raises
+        ------
+        KeyError
+            Raised if the dataset type is not registered.
+        sqlalchemy.exc.IntegrityError
```
andy-slac: I wish we had our own exception class for this sort of error instead of making high-level interface expose low-level/implementation exceptions.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
andy-slac: Agree `remove` should be fine.

tests/test_butler.py
```diff
@@ -589,6 +589,21 @@ def testPruneCollections(self):
...
+        tmpDatasetType = self.addDatasetType(tmpName, dimensions, storageClass,
+                                             butler.registry)
+        tmpFromRegistry = butler.registry.getDatasetType(tmpName)
+        self.assertEqual(tmpDatasetType, tmpFromRegistry)
+        butler.registry.pruneDatasetType(tmpName)
+        with self.assertRaises(KeyError):
+            butler.registry.getDatasetType(tmpName)
+
+        # Try and fail to delete a datasetType that is associated with data
+        with self.assertRaises(Exception):
```
andy-slac: Should you check for `IntegrityError` here instead?

andy-slac (approved):
Looks OK, few minor comments.

python/lsst/daf/butler/registry/_registry.py
timj: Maybe I should make one. I was pondering what to reraise although I guess we need a base class of RegistryDatabaseError and a subclass of RegistryIntegrityError -- trivial to add of course.

tests/test_butler.py
timj: Yes, but I was cranky about having to import sqlalchemy to make a test work.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
timj: Deleting something that isn't there doesn't trigger an error -- it becomes a no-op so is safe. In that scenario though (where we have discovered that things are being deleted) it might be better to call the refresh anyhow.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
timj: Actually, with the refresh call here and the delete being a no-op if it doesn't exist, I can remove the lines above that check for existence in the cache so this all gets much simpler.

python/lsst/daf/butler/registry/_registry.py
timj: It looks like `OrphanedRecordError` is already exactly the error I want.

PR #380, opened 2020-09-24
DM-26629: switch to calibration collections and remove calibration_label

TallJimbo:


python/lsst/daf/butler/_butler.py
```diff
@@ -664,12 +664,44 @@ def getDirect(self, ref: DatasetRef, *, parameters: Optional[Dict[str, Any]] = N
...
+        parameters : `dict`
+            Additional StorageClass-defined options to control reading,
+            typically used to efficiently read only a subset of the dataset.
+
+        Returns
+        -------
+        obj : `DeferredDatasetHandle`
+            A handle which can be used to retrieve a dataset at a later time.
+
+        Raises
```
timj, line 700: I don't think these docstrings are relevant.

python/lsst/daf/butler/_butler.py
```diff
@@ -664,12 +664,44 @@ def getDirect(self, ref: DatasetRef, *, parameters: Optional[Dict[str, Any]] = N
...
+        LookupError
+            Raised if no matching dataset exists in the `Registry` (and
+            ``allowUnresolved is False``).
+        ValueError
+            Raised if a resolved `DatasetRef` was passed as an input, but it
+            differs from the one found in the registry.
+        TypeError
+            Raised if no collections were provided.
+        """
+        return DeferredDatasetHandle(butler=self, ref=ref, parameters=parameters)
```
timj, line 709: In the interests of failing early, can we check that the supplied ref is actually resolved?

python/lsst/daf/butler/configs/datastores/webdavDatastore.yaml
```diff
@@ -8,6 +8,6 @@ datastore:
     # valid_first and valid_last here are YYYYMMDD; we assume we'll switch to
     # MJD (DM-15890) before we need more than day resolution, since that's all
     # Gen2 has.
-    default: "{run:/}/{datasetType}.{component:?}/{tract:?}/{patch:?}/{label:?}/{abstract_filter:?}/{subfilter:?}/{physical_filter:?}/{visit:?}/{datasetType}_{component:?}_{tract:?}_{patch:?}_{label:?}_{abstract_filter:?}_{physical_filter:?}_{calibration_label:?}_{visit:?}_{exposure:?}_{detector:?}_{instrument:?}_{skymap:?}_{skypix:?}_{run}"
```
timj: Yes, I should have a shared templates.yaml for these three datastores.

python/lsst/daf/butler/core/datasets/association.py
```diff
@@ -0,0 +1,49 @@
...
+@dataclass
+class DatasetAssociation:
+    """A struct that represents the membership of a single dataset in a
+    single collection.
+    """
+
+    __slots__ = ("ref", "collection", "timespan")
+
+    ref: DatasetRef
+    """Reference to a dataset.
```
timj: This can be resolved or unresolved?

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -1497,6 +1500,72 @@ def queryDimensionRecords(self, element: Union[DimensionElement, str], *,
...
+    ) -> Iterator[DatasetAssociation]:
+        """Iterate over dataset-collection combinations where the dataset is in
+        the collection.
+
+        This method is a temporary placeholder for better support for
+        assocation results in `queryDatasets`.  It will probably be
+        removed in the future, and should be avoided in production code
+        whenever possible.
+
+        Parameters
```
timj, line 1547: Missing some parameters here.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -1497,6 +1500,72 @@ def queryDimensionRecords(self, element: Union[DimensionElement, str], *,
...
+                    storage.datasetType.dimensions,
+                    tuple(row[name] for name in storage.datasetType.dimensions.required.names)
+                )
+                runRecord = self._collections[row[self._collections.getRunForeignKeyName()]]
+                ref = DatasetRef(storage.datasetType, dataId, id=row["id"], run=runRecord.name,
+                                 conform=False)
+                if collectionRecord.type is CollectionType.CALIBRATION:
+                    timespan = tsRepr.extract(row)
+                else:
+                    timespan = None
```
timj, line 1592: I really thought that mypy would get cranky about timespan here being a different type to the one two lines before. I always had to rewrite those as:

```python
timespan = tsRepr.extract(row) if collectionRecord.type is CollectionType.CALIBRATION else None
```
for mypy to work out that timespan can be `Optional[Timespan]`. Maybe I'm misremembering.

python/lsst/daf/butler/registry/queries/_builder.py
```diff
@@ -143,6 +143,13 @@ def joinDataset(self, datasetType: DatasetType, collections: Any, *,
             collections = CollectionSearch.fromExpression(collections)
         else:
             collections = CollectionQuery.fromExpression(collections)
+        # If we are searching all collections with no constraints, loop over
+        # RUN collections only, because that will include all datasets.
```
timj, line 147: Very good. Obvious in retrospect but I'm sure it all helps.

python/lsst/daf/butler/registry/wildcards.py
```diff
@@ -839,6 +839,12 @@ def iter(
                                         flattenChains=flattenChains, includeChains=includeChains):
             yield record
 
+    def __eq__(self, other: Any) -> bool:
+        if isinstance(other, CollectionQuery):
+            return self._search == other._search and self._patterns == other._patterns
+        else:
```
timj, line 844: The else doesn't really do much so could be dropped and just have the return as the last line in the method.

python/lsst/daf/butler/transfers/_context.py
```diff
@@ -79,8 +81,35 @@ def __init__(self, registry: Registry, datastore: Datastore, backend: RepoExport
...
+        Notes
+        -----
+        `~CollectionType.RUN` collections are also exported automatically when
+        any dataset referencing them is exported.  They may also be explicitly
+        exported this method to export the collection with no datasets.
+        Duplicate exports of collections are ignored.
+
+        Exporting a `~CollectionType.TAGGED` or `~CollectionType.CALIBRATION`
+        collection will cause its associations with exported datasets to also
+        be exported, but it does not export those datasets automatically.
```
timj, line 106: so you can end up with an empty calibration collection after import? Or a broken one because the references exist but the datasets don't?

python/lsst/daf/butler/transfers/_context.py
```diff
@@ -207,13 +238,106 @@ def _finish(self) -> None:
...
+                chains[record.name] = [child for child, _ in record.children]
+            else:
+                result.append(record.name)
+        result.sort()
+        # Sort all chains topologically, breaking ties lexicographically.
+        # Append these to 'result' and remove them from 'chains' as we go.
+        while chains:
+            unblocked = {
+                parent for parent, children in chains.items()
+                if not any(child in chains.keys() for child in children)
```
timj, line 295: is the keys() needed here? Normally `__contains__` would work on the dict directly.

timj (approved):
Looks good. A few minor comments and questions.

python/lsst/daf/butler/core/datasets/association.py
TallJimbo: Always resolved.  Will document.

python/lsst/daf/butler/registry/_registry.py
TallJimbo: I think I'd intended to delete both of those from the signature as well, but actual usage reveals that I do need `flattenChains` (but `includeChains` definitely doesn't make sense for this method).  Will fix.

python/lsst/daf/butler/registry/_registry.py
TallJimbo: Yup, I have the same recollection, and I'm not sure if mypy has gotten smarter or if this case is easier in a non-obvious way.  I do expect mypy will eventually get smart enough to infer the optional in essentially all cases like this, so I'm not inclined to fight it when it seems to.

python/lsst/daf/butler/registry/wildcards.py
TallJimbo: I've seen style guides (in other languages, I think?) that recommend this form to "preserve the symmetry of the branching" or some such, and I have never had a good sense for whether one form other the other was idiomatic in Python.  But if this was strange enough to you to merit a comment, it probably merits changing.

python/lsst/daf/butler/transfers/_context.py
TallJimbo: Empty is indeed possible (I think that's unavoidable if a partial export of a collection is allowed).  Broken should not be possible.

python/lsst/daf/butler/transfers/_context.py
TallJimbo: Not needed, I just thought it was easier to read.  I know `.keys` is basically always visual noise for some people, but when it's not extremely obvious that the target of `__contains__` or `__iter__` is a mapping I find it very helpful to have it called out.

timj: Ok, it looks odd to me because I'm then wondering what's special about the `keys` that means that `in` doesn't work (and surely it makes python work harder since it then has to generate keys and do the `__contains__` on that?)

PR #382, opened 2020-09-24
DM-26371: Make Quantum immutable

natelust:
Remove all the properties of Quantum that were expected to be
mutable. Make sure object was then immutable. Add a hash and
eq function based on immutable instance attributes.

### DIFF
--- a/python/lsst/daf/butler/core/quantum.py
+++ b/python/lsst/daf/butler/core/quantum.py
@@ -31,13 +31,13 @@ from typing import (
     Type,
     TYPE_CHECKING,
     Union,
+    Tuple,
+    Any
 )
-
-import astropy.time
 
 from lsst.utils import doImport
 
-from .named import NamedKeyDict
+from .named import NamedKeyDict, NamedKeyMapping
 
 if TYPE_CHECKING:
     from .dimensions import DataCoordinate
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/core/quantum.py
```diff
@@ -165,7 +131,7 @@ def initInputs(self) -> NamedKeyDict[DatasetType, DatasetRef]:
         with `DatasetType` instances as keys (names can also be used for
         lookups) and `DatasetRef` instances as values.
         """
-        return self._initInputs
+        return copy.copy(self._initInputs)
```
TallJimbo: Check out `NamedKeyDict.freeze`; I think it's exactly what you want here and in similar accessors (or maybe the constructor).  That'd come with changing the return type `NamedKeyMapping[DatasetType, DatasetRef]`.

python/lsst/daf/butler/core/quantum.py
```diff
@@ -205,70 +158,37 @@ def outputs(self) -> NamedKeyDict[DatasetType, List[DatasetRef]]:
...
+        cumHash = hash(self.taskClass)
+        cumHash ^= hash(self.dataId)
+        if isinstance(self.initInputs, abc.Mapping):
+            values = self.initInputs.values()
+        else:
+            values = self.initInputs
+        for item in values:
+            cumHash ^= hash(item)
+        if self.predictedInputs:
+            for item in self.predictedInputs.values():
```
andy-slac: Are these always ordered identically across different instances?

python/lsst/daf/butler/core/quantum.py
```diff
@@ -205,70 +158,37 @@ def outputs(self) -> NamedKeyDict[DatasetType, List[DatasetRef]]:
...
+                else:
+                    cumHash ^= hash(item)
+        if self.outputs:
+            for item in self.outputs.values():
+                if isinstance(item, abc.Iterable):
+                    for value in item:
+                        cumHash ^= hash(value)
+                else:
+                    cumHash ^= hash(item)
+        return cumHash
```
andy-slac: I'm not sure how stable it is to pack all inputs and outputs into a hash (ordering is the obvious question). OTOH it may be enough to use `taskClass` and `dataID` for a hash, in most cases this should be sufficiently unique.

andy-slac (approved):
Looks like an improvement, but I'm concerned again with the hash function, couple of questions related to that.

python/lsst/daf/butler/core/quantum.py
```diff
@@ -99,39 +84,26 @@ class Quantum:
     def __init__(self, *, taskName: Optional[str] = None,
                  taskClass: Optional[Type] = None,
                  dataId: Optional[DataCoordinate] = None,
-                 run: Optional[str] = None,
                  initInputs: Optional[Union[Mapping[DatasetType, DatasetRef], Iterable[DatasetRef]]] = None,
                  predictedInputs: Optional[Mapping[DatasetType, List[DatasetRef]]] = None,
```
andy-slac: With dropping `actualInputs` would it make sense to call it `inputs` instead of `predictedInputs`?

python/lsst/daf/butler/core/quantum.py
natelust: That is not enough to compare Quanta prepared from two different input collections. They might have different inputs for instance and those would be considered the same.

python/lsst/daf/butler/core/quantum.py
natelust: Good point. I switched this to calculating from a sorted list of hashes, which will always be the same. I also realized there was no need to do this on each hash call, as `Quantum` is not immutable, so I do the calculation once in the constructor.

python/lsst/daf/butler/core/quantum.py
andy-slac: Well, same hash does not mean object are the same, there is always `__eq__` test involved. I do not think you ever want to put quanta from different graphs into one collection, but even if you do it will still work event if their hashes are equal. My worry is that with a complicated hash it's hard to prove that it is always consistent with `__eq__`.

natelust: I wrote up a whole thing arguing back and forth based on hash table performance. I got to the end and deleted it all, because now I think you are right.

PR #384, opened 2020-09-26
DM-26947: generalize test support code and add test for Butler.get on calibrations

TallJimbo:
I forgot to ask for a review of 2062a8d13df520f52451bf7b0691c41faa2b63e0 when I added it to DM-26629 after the main review for that ticket was complete.  This branch address two obvious problems with that commit - lack of docs and tests - already, but I'd like to have that commit reviewed here as well as the code actually on this branch.

### DIFF
--- a/tests/test_simpleButler.py
+++ b/tests/test_simpleButler.py
@@ -21,17 +21,12 @@
 
 from __future__ import annotations
 
-import contextlib
-from io import StringIO
 import os
+import tempfile
 from typing import (
-    Generator,
+    Any,
     Iterable,
     Optional,
-    Set,
-    TextIO,
-    Tuple,
-    Union,
 )
 import unittest
 import unittest.mock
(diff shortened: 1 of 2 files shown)


tests/test_simpleButler.py
```diff
@@ -87,105 +93,14 @@ def makeRegistry(self) -> Registry:
...
-            helper = RepoExportContext(registry, datastore, backend=backend,
-                                       directory=directory, transfer=transfer)
-            yield helper
-        except BaseException:
-            raise
-        else:
-            helper._finish()
+    def makeButler(self, **kwargs: Any) -> Butler:
+        config = ButlerConfig()
+        config["registry", "db"] = "sqlite://"
```
timj: It might be clearer here if you use `:memory:` explicitly here like is done in https://github.com/lsst/daf_butler/blob/master/python/lsst/daf/butler/tests/_testRepo.py#L76

PR #386, opened 2020-09-30
DM-27007: Disable checksum calculation by default

timj:


MichelleGower (approved):
The changes are fine and should be merged.   Outside the scope of this ticket (but because of the have_sized line addition here), not sure I follow the have_sized pieces of the code.  In particular the switch to using the tgtLocation.uri.size.  I get the needing size to store in the database.  At first it reads like it is going to be a sanity check (does src == tgt) but it's not.  The assumption is that src does equal tgt (no check) and couldn't get the size from the source so trying to get from the target?

timj:
The `have_sized` logic is trying to say that if the file to be ingested is local, ask the local thing directly for the size and calculate the checksum on the local file. If that calculation has not happened, we instead ask the final location of the ingested file for size and checksum information. If this is a remote file then we have to ask the remote for the size and that involves additional network traffic (and checksum could involve redownloading the file).

PR #387, opened 2020-10-02
DM-27034: Overhaul dimension construction and add ABCs

TallJimbo:


python/lsst/daf/butler/core/dimensions/_config.py
```diff
@@ -0,0 +1,170 @@
...
+      with dictionary values that each define a `StandardTopologicalFamily`.
+
+    - packers: a nested dictionary whose entries define factories for a
+      `DimensionPacker` instance.
+
+    See the documentation for the linked classes above for more information
+    on the configuration syntax.
+    """
+    component = "dimensions"
+    requiredKeys = ("version", "elements", "skypix")
```
andy-slac, line 67: Just a question - will we still need multiple version support if dimensions are configured from database?

python/lsst/daf/butler/core/dimensions/_graph.py
```diff
@@ -108,9 +114,21 @@ def __new__(cls, universe: DimensionUniverse,
...
+            oldSize = len(conformedNames)
+            while True:
+                # iterate over a temporary copy so we can modify the original
+                for name in tuple(conformedNames):
+                    conformedNames.update(universe[name].required.names)
+                    conformedNames.update(universe[name].implied.names)
+                if oldSize == len(conformedNames):
+                    break
+                else:
+                    oldSize = len(conformedNames)
```
andy-slac: Feels like this should be a method of universe?

python/lsst/daf/butler/core/dimensions/construction.py
```diff
@@ -0,0 +1,198 @@
...
+    def visit(self, builder: DimensionConstructionBuilder) -> None:
+        """Modify the given builder object to include the entities this object
+        is responsible for constructing.
+
+        Parameters
+        ----------
+        builder : `DimensionConstructionBuilder`
+            Builder to modify in-place and from which dependencies can be
+            obtained.
+        """
```
andy-slac, line 95: What should happen if `visit()` is called while `hasDependenciesIn()` still returns True?

python/lsst/daf/butler/core/dimensions/_graph.py
```diff
@@ -403,6 +399,16 @@ def addToOrder(element: DimensionElement) -> None:
             self._primaryKeyTraversalOrder = order
         return order
 
+    @property
+    def spatial(self) -> NamedValueAbstractSet[TopologicalFamily]:
+        # TODO: docs
```
andy-slac: Can you add it on this branch? One below for `temporal()` too.

python/lsst/daf/butler/core/dimensions/_elements.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def __str__(self) -> str:
+        return self.name
+
+    def __repr__(self) -> str:
+        return f"{type(self).__name__}({self.name})"
+
+    def __eq__(self, other: Any) -> bool:
+        try:
+            return self.name == other.name
```
andy-slac, line 82: Should it also check that universes are the same?

python/lsst/daf/butler/core/dimensions/_elements.py
```diff
@@ -0,0 +1,297 @@
...
+
+    name: str
+    """Unique name for this dimension element (`str`).
+    """
+
+    @property
+    @abstractmethod
+    def required(self) -> NamedValueAbstractSet[Dimension]:
+        """Dimensions that are sufficient (often necessary) to uniquely
+        identify a record of this dimension element.
```
andy-slac: Hmm, I thought that required means necessary, I sort of associated these with identifying relationships in database speak. Is there an example where they are not necessary?

python/lsst/daf/butler/core/dimensions/_elements.py
```diff
@@ -0,0 +1,297 @@
...
+        return NamedValueSet(list(self.required) + list(self.implied)).freeze()
+
+    @property  # type: ignore
+    @cached_getter
+    def graph(self) -> DimensionGraph:
+        """Minimal graph that includes this element (`DimensionGraph`).
+
+        ``self.graph.required`` includes all dimensions whose primary key
+        values are sufficient (often necessary) to uniquely identify ``self``
+        (including ``self` if ``isinstance(self, Dimension)``.
```
andy-slac: One backtick is missing after `self`.

python/lsst/daf/butler/core/dimensions/_elements.py
```diff
@@ -0,0 +1,297 @@
...
+    @abstractmethod
+    def viewOf(self) -> Optional[str]:
+        """Name of another table this element's records are drawn from (`str`
+        or `None`).
+        """
+        raise NotImplementedError()
+
+    @property
+    def alwaysJoin(self) -> bool:
+        # TODO: docs
```
andy-slac: Here too.

python/lsst/daf/butler/core/dimensions/_elements.py
```diff
@@ -0,0 +1,297 @@
...
+        """
+        _, *alternateKeys = self.uniqueKeys
+        return NamedValueSet(alternateKeys).freeze()
+
+
+class DimensionCombination(DimensionElement):
+    """A `DimensionElement` that provides extra metadata and/or relationship
+    endpoint information for a combination of dimensions.
+    """
+    pass
```
andy-slac, line 300: Is this going to be extended with methods or is it just the type that is useful by itself?

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,411 @@
...
+        self,
+        name: str,
+        space: TopologicalSpace, *,
+        members: NamedValueAbstractSet[DimensionElement],
+    ):
+        super().__init__(name, space)
+        self.members = members
+
+    def choose(self, endpoints: NamedValueAbstractSet[TopologicalRelationshipEndpoint]) -> DimensionElement:
+        # Docstring inherited from TopologicalRelationshipEndpoint.
```
andy-slac: TopologicalRelationshipEndpoint -> TopologicalFamily

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,411 @@
...
+
+from .. import ddl
+from ..named import NamedValueAbstractSet, NamedValueSet
+
+from ._elements import Dimension, DimensionCombination, DimensionElement
+from ._topology import TopologicalFamily, TopologicalRelationshipEndpoint, TopologicalSpace
+from .construction import DimensionConstructionBuilder, DimensionConstructionVisitor
+
+
+class StandardTopologicalFamily(TopologicalFamily):
```
andy-slac, line 45: "standard" suggests that other things are non-standard, maybe other names like "database" or "default" are better for these classes and module (though I don't have strong feelings about it, up to you)?

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,411 @@
...
+    members: NamedValueAbstractSet[DimensionElement]
+    """The members of this family, ordered according to the priority used in
+    `choose` (first-choice member first).
+    """
+
+
+class StandardTopologicalFamilyConstructionVisitor(DimensionConstructionVisitor):
+    """A construction visitor for `StandardTopologicalFamily`.
+
+    This visitor depends on (and is thus visited after) its members.=
```
andy-slac: Extraneous `=`

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,411 @@
...
+    """A `Dimension` implementation that maps directly to a database table or
+    query.
+
+    Parameters
+    ----------
+    name : `str`
+        Name of the dimension.
+    required : `NamedValueAbstractSet` [ `Dimension` ]
+        Other dimensions whose keys are part of the compound primary key for
+        this dimension's (logical) table, as well as references to ther own
```
andy-slac: typo: ther

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,411 @@
...
+        table as foreign keys.
+    metadata : `NamedValueAbstractSet` [ `ddl.FieldSpec` ]
+        Field specifications for all non-key fields in this dimension's table.
+    cached : `bool`
+        Whether to cache the records of this dimension in memory when they are
+        fetched from the database.
+    viewOf : `str`, optional
+        Name of another `StandardDimension` or `StandardDimensionCombination`
+        whose records this dimension's records summarize.
+    uniqueKeys : `NamedValueAbstractSet [ `ddl.FieldSpec` ]
```
andy-slac: Backtick is missing

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,411 @@
...
+        return self._cached
+
+    @property
+    def viewOf(self) -> Optional[str]:
+        # Docstring inherited from DimensionElement.
+        return self._viewOf
+
+    @property
+    def uniqueKeys(self) -> NamedValueAbstractSet[ddl.FieldSpec]:
+        # Docstring inherited from DimensionElement.
```
andy-slac: DimensionElement -> Dimension

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,411 @@
...
+    """A `DimensionCombination` implementation that maps directly to a database
+    table or query.
+
+    Parameters
+    ----------
+    name : `str`
+        Name of the dimension.
+    required : `NamedValueAbstractSet` [ `Dimension` ]
+        Dimensions whose keys define the compound primary key for this
+        combinations's (logical) table, as well as references to ther own
```
andy-slac: `ther`

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,411 @@
...
+        (logical) table as foreign keys.
+    metadata : `Iterable` [ `ddl.FieldSpec` ]
+        Field specifications for all non-key fields in this element's table.
+    cached : `bool`
+        Whether to cache the records of this element in memory when they
+        are fetched from the database.
+    viewOf : `str`, optional
+        Name of another `StandardDimension` or `StandardDimensionCombination`
+        whose records this element's records summarize.
+    uniqueKeys : `NamedValueAbstractSet [ `ddl.FieldSpec` ]
```
andy-slac: backtick

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,411 @@
...
+    """A construction visitor for `StandardDimension` and
+    `StandardDimensionCombination`.
+
+    Parameters
+    ----------
+    name : `str`
+        Name of the dimension.
+    required : `Set` [ `Dimension` ]
+        Names of dimensions whose keys define the compound primary key for this
+        element's (logical) table, as well as references to ther own
```
andy-slac: ther

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,411 @@
...
+                self.name,
+                required=required,
+                implied=implied.freeze(),
+                metadata=self._metadata,
+                cached=self._cached,
+                viewOf=self._viewOf,
+                uniqueKeys=self._uniqueKeys,
+            )
+            required.add(dimension)
+            required.freeze()
```
andy-slac: This is potentially problematic if `StandardDimension` decides to make a copy of `required`. Can this be done in `StandardDimension` constructor?

andy-slac (commented):
I still have a lot to go through but need to stop for now. Will finish later, here is a bunch of comments/question I have so far.

python/lsst/daf/butler/core/dimensions/_config.py
TallJimbo: I think so, but probably only as something to help with migrations:
 - We will probably want to be able to have multiple `DimensionUniverse` instances active in the same process (so we need `DimensionUniverse._instances` to be a dict with some kind of version-like key, not just a complete singleton).
 - We probably want to be able to organize (in git, probably) migration scripts between dimensions definitions.

I do think a hash/digset would work as well or better than an integer version, though, though writing the digest function would be a pain.  I do not see any need for monotonicity.

python/lsst/daf/butler/core/dimensions/_elements.py
TallJimbo: The case of having more than one `DimensionUniverse` active is sufficiently rare and special that I don't think it needs to.  There's probably very little cost to adding that check _here_, but I think it's better to just define this equality comparison as only being well-defined within one universe than let this equality get out of sync with the requirements of the `Named*` collections, which also look only at the name and would be harder to change to look at the universe as well.

python/lsst/daf/butler/core/dimensions/_elements.py
TallJimbo: This is an old docstring that I just moved to a new location, and it's probably worth changing.  I was thinking about the (mostly hypothetical) case where there may be some other [set of] fields that might uniquely identify a dimension record, like a physical filter whose string name includes the instrument name (but this is not required now, even if it is sometimes the case).  I think your comment highlights that this wording just causes confusion.  I'll just say they're necessary here, and document any exceptions that come up where they come up.

python/lsst/daf/butler/core/dimensions/_elements.py
TallJimbo: Right now it's just the type, but I think it's at least likely that some methods on `DimensionElement` will move to be only here and not on `Dimension` in the future.

python/lsst/daf/butler/core/dimensions/_graph.py
TallJimbo: Certainly!  I'm glad you caught this, and I'm sorry you had to, because I had definitely intended to do it on this branch, and it just slipped through.

python/lsst/daf/butler/core/dimensions/construction.py
TallJimbo: The `visit` call would fail in a (currently) unspecified way.  I'll document that subclasses are permitted to assume that this is that this condition has been met, and that callers must guarantee it.  I don't think it's worth requiring subclasses to check it, given that the only caller is likely to be the `Builder` class and this part of the logic there is fairly straightforward.

python/lsst/daf/butler/core/dimensions/standard.py
TallJimbo: That's a good point, but I think I'll wait for my current tickets to be further along before I rename, because I'm think the hierarchy of implementations will expand a bit and that may clarify what the different roles are a bit better.

python/lsst/daf/butler/core/dimensions/standard.py
TallJimbo: Yes, don't know why I didn't think of that solution here.

python/lsst/daf/butler/core/dimensions/_config.py
andy-slac: Will a registry have multiple universe definition in it too? I agree that hash is a pain, could we use user-defined string/tag instead, may be easier to track those than integer versions.

python/lsst/daf/butler/core/dimensions/standard.py
andy-slac: Maybe we could write it `Docstring inherited from base class` to reduce entropy?

python/lsst/daf/butler/core/dimensions/_packer.py
```diff
@@ -205,23 +213,53 @@ def __call__(self, fixed: DataCoordinate) -> DimensionPacker:
             packer.  Must be expanded with all metadata known to the
             `Registry`.  ``fixed.hasRecords()`` must return `True`.
         """
-        assert fixed.graph.issuperset(self.fixed)
+        if not isinstance(self._fixed, DimensionGraph):
+            self._fixed = universe.extract(self._fixed)
+        if not isinstance(self._dimensions, DimensionGraph):
+            self._dimensions = universe.extract(self._dimensions)
```
andy-slac, line 224: Should this be done once in `__init__`?

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,407 @@
...
+    """A `Dimension` implementation that maps directly to a database table or
+    query.
+
+    Parameters
+    ----------
+    name : `str`
+        Name of the dimension.
+    required : `NamedValueSet` [ `Dimension` ]
+        Other dimensions whose keys are part of the compound primary key for
+        this dimension's (logical) table, as well as references to there own
```
andy-slac: there -> their

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,407 @@
...
+    cached : `bool`
+        Whether to cache the records of this dimension in memory when they are
+        fetched from the database.
+    viewOf : `str`, optional
+        Name of another `StandardDimension` or `StandardDimensionCombination`
+        whose records this dimension's records summarize.
+    uniqueKeys : `NamedValueAbstractSet` [ `ddl.FieldSpec` ]
+        Fields that can each be used to uniquely identify this dimension (given
+        values for all required dimensions).  The first of these is used as
+        (part of) this dimension's table's primary key, which others used to
```
andy-slac: which -> while?

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,407 @@
...
+        name: str, *,
+        required: NamedValueSet[Dimension],
+        implied: NamedValueAbstractSet[Dimension],
+        metadata: NamedValueAbstractSet[ddl.FieldSpec],
+        cached: bool,
+        viewOf: Optional[str],
+        uniqueKeys: NamedValueAbstractSet[ddl.FieldSpec],
+    ):
+        super().__init__(name)
+        required.add(self)
```
andy-slac, line 170: Is it worth making a copy before updating an argument?

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,407 @@
...
+    """A `DimensionCombination` implementation that maps directly to a database
+    table or query.
+
+    Parameters
+    ----------
+    name : `str`
+        Name of the dimension.
+    required : `NamedValueAbstractSet` [ `Dimension` ]
+        Dimensions whose keys define the compound primary key for this
+        combinations's (logical) table, as well as references to there own
```
andy-slac: their

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,407 @@
...
+    metadata : `NamedValueAbstractSet` [ `ddl.FieldSpec` ]
+        Field specifications for all non-key fields in this combination's
+        table.
+    cached : `bool`
+        Whether to cache the records of this combination in memory when they
+        are fetched from the database.
+    viewOf : `str`, optional
+        Name of another `StandardDimension` or `StandardDimensionCombination`
+        whose records this combination's records summarize.
+
```
andy-slac, line 243: `alwaysJoin` parameter is missing from docstring.

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,407 @@
...
+    viewOf : `str`, optional
+        Name of another `StandardDimension` or `StandardDimensionCombination`
+        whose records this element's records summarize.
+    uniqueKeys : `Iterable` [ `ddl.FieldSpec` ]
+        Fields that can each be used to uniquely identify this dimension (given
+        values for all required dimensions).  The first of these is used as
+        (part of) this dimension's table's primary key, which others used to
+        define unique constraints.  Should be empty for
+        `StandardDimensionCombination` definitions.
+    """
```
andy-slac, line 348: `alwaysJoin` missing here too

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,407 @@
...
+    """A construction visitor for `StandardDimension` and
+    `StandardDimensionCombination`.
+
+    Parameters
+    ----------
+    name : `str`
+        Name of the dimension.
+    required : `Set` [ `Dimension` ]
+        Names of dimensions whose keys define the compound primary key for this
+        element's (logical) table, as well as references to there own
```
andy-slac: there

python/lsst/daf/butler/core/dimensions/standard.py
```diff
@@ -0,0 +1,407 @@
...
+    cached : `bool`
+        Whether to cache the records of this element in memory when they
+        are fetched from the database.
+    viewOf : `str`, optional
+        Name of another `StandardDimension` or `StandardDimensionCombination`
+        whose records this element's records summarize.
+    uniqueKeys : `Iterable` [ `ddl.FieldSpec` ]
+        Fields that can each be used to uniquely identify this dimension (given
+        values for all required dimensions).  The first of these is used as
+        (part of) this dimension's table's primary key, which others used to
```
andy-slac: while

python/lsst/daf/butler/core/dimensions/standard.py
andy-slac: Lot's of copy-paste for parameter descriptions, would be nice if sphinx could automate that.

python/lsst/daf/butler/core/dimensions/_config.py
TallJimbo: No, a Registry instance will only ever have one universe.  User-defined strings means humans like us have to come up with names, and does not sound fun, but you're probably right that it's better than integers, at least in terms of avoiding clashes when there is branching.

python/lsst/daf/butler/core/dimensions/_universe.py
```diff
@@ -78,59 +88,53 @@ class DimensionUniverse:
...
+        cls,
+        config: Optional[Config] = None, *,
+        version: Optional[int] = None,
+        builder: Optional[DimensionConstructionBuilder] = None,
+    ) -> DimensionUniverse:
+        # Try to get a version first, to look for existing instances; try to
+        # do as little work as possible at this stage.
+        if version is None:
+            if builder is None:
+                from ._config import DimensionConfig
```
andy-slac: Why do you import here, is there import cycle somewhere?

andy-slac (commented):
Looks OK, few minor comments/questions.

python/lsst/daf/butler/core/dimensions/_config.py
andy-slac: So if registry only has one universe then multiple universes could only come from different registries? Could we make universe a property of a registry rather than having a singleton dict with all universes?

TallJimbo: The singleton dict is only needed for pickling, and that pickling is probably only needed for multiprocessing (https://github.com/lsst/daf_butler/blob/master/python/lsst/daf/butler/core/dimensions/universe.py#L333).

So there may be a way to allow multiple versions _or_ pickling but not both (and I think that's probably okay if migration scripts are the only use case for multiple versions), but it'd be a bit tricky.  And we may want the identifiers anyway to organize the migration scripts anyway.

python/lsst/daf/butler/core/dimensions/_packer.py
TallJimbo: Right now, it's not possible, because `__init__` is called before a `DimensionUniverse` exists, and it isn't possible to create a `DimensionGraph` until the universe exists. I'll add a comment to that effect.

python/lsst/daf/butler/core/dimensions/_universe.py
TallJimbo: There was a cycle in an earlier version. I managed to flatten that one out but forgot to change this; I'll do that now.

python/lsst/daf/butler/core/dimensions/standard.py
TallJimbo: Not in this case, I think, but I should document that behavior, and will.

python/lsst/daf/butler/core/dimensions/standard.py
TallJimbo: Yeah, I think copy-paste here is just the lesser of two evils.  It's probably a fundamental Python limitation, not a Sphinx one, at least as long as we care about having docstrings as well as HTML docs.

python/lsst/daf/butler/core/dimensions/_universe.py
andy-slac: 👍 There is another instance of this import below, just in case.

python/lsst/daf/butler/core/dimensions/_config.py
andy-slac: That unpickling worries me, it obviously depends on `_instances` being populated by something else. This should work with out current fork-based multiprocesing but that is not the only way to do multiprocessing. Maybe we should stop pickling universe and always get it from registry?

TallJimbo: The problem is that we need to pickle things like `DataCoordinate` and `DimensionGraph`, which have `universe` attributes.  The `multiprocessing` package does not provide any way to give a registry or universe or anything else to their unpickle logic other than via singletons, as far as I can tell; this is exactly the problem that the [python pickle docs](https://docs.python.org/3/library/pickle.html#persistence-of-external-objects) say to solve with a custom unpickler, but `multiprocessing` does not provide a hook to change the unpickler.

We may be able to slowly remove a lot of those `.universe` attibutes (maybe even all) as `DimensionUniverse` becomes mostly a factory for making `DimensionGraph`, and that might be a big overall improvement.  But there will be a lot of API breakage where we have to add `universe` arguments to many methods instead.

PR #388, opened 2020-10-02
Use released version of documenteer.

ktlim:


timj:
@ktlim I'm fine with this since Jonathan says he won't have instructions available for migration until next week.

.github/workflows/build.yaml
```diff
@@ -49,7 +49,7 @@ jobs:
         run: pytest -r a -v -n 3 --open-files
 
       - name: Install documenteer
-        run: pip install documenteer[pipelines]
+        run: pip install documenteer[pipelines]==0.5.8
```
timj: I think <0.6.0 is also an option.

TallJimbo (approved):
I'm all for just merging this.

ktlim:
I've got to get the syntax right first :)

PR #391, opened 2020-10-09
DM-24355: make Database.sync (and methods that use it) transaction-friendly.

TallJimbo:


python/lsst/daf/butler/registry/databases/postgresql.py
```diff
@@ -161,6 +161,18 @@ def replace(self, table: sqlalchemy.schema.Table, *rows: dict) -> None:
         query = query.on_conflict_do_update(constraint=table.primary_key, set_=data)
         self._connection.execute(query, *rows)
 
+    def ensure(self, table: sqlalchemy.schema.Table, *rows: dict) -> int:
+        # Docstring inherited.
+        if not (self.isWriteable() or table.key in self._tempTables):
+            raise ReadOnlyDatabaseError(f"Attempt to esnure into read-only database '{self}'.")
```
timj: "esnure"

python/lsst/daf/butler/registry/databases/sqlite.py
```diff
@@ -417,6 +433,17 @@ def replace(self, table: sqlalchemy.schema.Table, *rows: dict) -> None:
             )
         self._connection.execute(_Replace(table), *rows)
 
+    def ensure(self, table: sqlalchemy.schema.Table, *rows: dict) -> int:
+        if not (self.isWriteable() or table.key in self._tempTables):
```
timj: This check is starting to appear in many places. Could we instead do something like a

```python
self.assertWriteable(table)
```
with an optional message string?

python/lsst/daf/butler/registry/databases/sqlite.py
```diff
@@ -65,15 +65,16 @@ class _Replace(sqlalchemy.sql.Insert):
     pass
 
 
+# SQLite and PostgreSQL use similar syntax for their ON CONFLICT extension,
+# but SQLAlchemy only knows about PostgreSQL's, so we have to compile some
```
timj, line 69: That seems odd that sqlite doesn't support it in sqlalchemy. Is it just that sqlalchemy team haven't got around to it or there are problems with it? Could we submit a patch?

timj (approved):

python/lsst/daf/butler/registry/databases/sqlite.py
TallJimbo: I think it's _slightly_ new functionality in SQLite, but not that new.  I imagine a patch would be appreciated, but it'd be a while before I got around to it.

PR #393, opened 2020-10-16
DM-26407: Store dimensions configuration in registry database

andy-slac:
Dimensions are now stored in registry attributes table with a key
"config:dimensions.yaml", serialized as YAML data. This change means
that we cannot use SQLite in-memory datqabase for tests that make
multiple connections (multiple Butler instances) so I had to update many
unit tests to use on-disk database instead. With this commit dimensions
are still part of full butler config but they will be separated in next
commits.

DimensionConfig is now a separate optional argument to `makeRepo`, by
default it is loaded from `dimensions.yaml`. ButlerConfig does not
include dimensions any more. `butler` CLI adds an option
--dimension-config. `dimensions.yaml` dropped top-level `dimensions`
key, and this is exactly how it is stored in database.

DimensionConfig still inherits from ConfigSubset, this is mostly for
convenience of finding defaults, though defaults loading behavior has
been overriden, it only loads first file that it finds.

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -253,7 +253,8 @@ class Butler:
     """
 
     @staticmethod
-    def makeRepo(root: str, config: Union[Config, str, None] = None, standalone: bool = False,
+    def makeRepo(root: str, config: Union[Config, str, None] = None,
+                 dimensionConfig: Union[Config, str, None] = None, standalone: bool = False,
                  createRegistry: bool = True, searchPaths: Optional[List[str]] = None,
                  forceConfigRoot: bool = True, outfile: Optional[str] = None,
                  overwrite: bool = False) -> Config:
@@ -272,6 +273,9 @@ class Butler:
             configuration will be used.  Root-dependent config options
             specified in this config are overwritten if ``forceConfigRoot``
             is `True`.
+        dimensionConfig : `Config` or `str`, optional
+            Configuration for dimensions, will be used to initialize registry
+            database. Only used when ``createRegistry`` is `True`.
         standalone : `bool`
             If True, write all expanded defaults, not just customized or
             repository-specific settings.
@@ -374,9 +378,7 @@ class Butler:
         # Create Registry and populate tables
         if createRegistry:
             registryConfig = RegistryConfig(config.get("registry"))
-            # TODO: for now DimensionConfig comes from a full config, in the
-            # future it should become separate.
-            dimensionConfig = DimensionConfig(config.get("dimensions"))
+            dimensionConfig = DimensionConfig(dimensionConfig)
             Registry.createFromConfig(registryConfig, dimensionConfig=dimensionConfig, butlerRoot=root)
         return config
 
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -62,6 +62,7 @@ from .core import (
     DatasetRef,
     DatasetType,
     Datastore,
+    DimensionConfig,
     FileDataset,
     StorageClassFactory,
     Timespan,
@@ -371,7 +372,12 @@ class Butler:
         config.dumpToUri(configURI, overwrite=overwrite)
 
         # Create Registry and populate tables
-        Registry.fromConfig(config, create=createRegistry, butlerRoot=root)
+        if createRegistry:
+            registryConfig = RegistryConfig(config.get("registry"))
+            # TODO: for now DimensionConfig comes from a full config, in the
+            # future it should become separate.
+            dimensionConfig = DimensionConfig(config.get("dimensions"))
+            Registry.createFromConfig(registryConfig, dimensionConfig=dimensionConfig, butlerRoot=root)
         return config
 
     @classmethod
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -255,9 +255,8 @@ class Butler:
     @staticmethod
     def makeRepo(root: str, config: Union[Config, str, None] = None,
                  dimensionConfig: Union[Config, str, None] = None, standalone: bool = False,
-                 createRegistry: bool = True, searchPaths: Optional[List[str]] = None,
-                 forceConfigRoot: bool = True, outfile: Optional[str] = None,
-                 overwrite: bool = False) -> Config:
+                 searchPaths: Optional[List[str]] = None, forceConfigRoot: bool = True,
+                 outfile: Optional[str] = None, overwrite: bool = False) -> Config:
         """Create an empty data repository by adding a butler.yaml config
         to a repository root directory.
 
@@ -275,7 +274,7 @@ class Butler:
             is `True`.
         dimensionConfig : `Config` or `str`, optional
             Configuration for dimensions, will be used to initialize registry
-            database. Only used when ``createRegistry`` is `True`.
+            database.
         standalone : `bool`
             If True, write all expanded defaults, not just customized or
             repository-specific settings.
@@ -284,8 +283,6 @@ class Butler:
             may be good or bad, depending on the nature of the changes).
             Future *additions* to the defaults will still be picked up when
             initializing `Butlers` to repos created with ``standalone=True``.
-        createRegistry : `bool`, optional
-            If `True` create a new Registry.
         searchPaths : `list` of `str`, optional
             Directory paths to search when calculating the full butler
             configuration.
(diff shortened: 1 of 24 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -371,7 +376,10 @@ def makeRepo(root: str, config: Union[Config, str, None] = None, standalone: boo
         config.dumpToUri(configURI, overwrite=overwrite)
 
         # Create Registry and populate tables
-        Registry.fromConfig(config, create=createRegistry, butlerRoot=root)
+        if createRegistry:
```
timj: I see three places where we use createRegistry=False and those are just in tests. Given what we are currently doing with registry versioning I don't think setting this to False is ever right. There is further discussion on DM-26237.

One of them seems to involve creating a registry, then overwriting the config file with the standalone config (but in the same registry) and then comparing configs. That test could run in two distinct butlers so doesn't need createRegistry=False. The other one in test_butler is solely for testing overwrite=False (so the createRegistry is irrelevant). Finally there is one usage in test_testRepo that uses makeRepo to create a butler.yaml outside of the repository and doesn't care about the registry itself being set up -- I think that could be True without any problem.

python/lsst/daf/butler/_butler.py
```diff
@@ -371,7 +376,10 @@ def makeRepo(root: str, config: Union[Config, str, None] = None, standalone: boo
...
+            registryConfig = RegistryConfig(config.get("registry"))
```
timj: Doesn't this `config.get` subvert what `ConfigSubset` is trying to work out for itself? By this point it's a full ButlerConfig config so it shouldn't need to be given the "registry" key. We've also already created a RegistryConfig above so this may not be necessary.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -122,19 +126,70 @@ class Registry:
...
+        Parameters
+        ----------
+        config : `RegistryConfig` or `str`, optional
+            Registry configuration, if missing then default configuration will
+            be loaded from registry.yaml.
+        dimensionConfig : `DimensionConfig` or `str`, optional
+            Dimensions configuration, if missing then default configuration
+            will be loaded from dimensions.yaml.
+        butlerRoot : `str`, optional
+            Path to the repository root this `Registry` will manage.
```
timj, line 146: URI to repository root, not a path (I see that fromConfig is now wrong).

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -122,19 +126,70 @@ class Registry:
...
+
+        Returns
+        -------
+        registry : `Registry`
+            A new `Registry` instance.
+        """
+        if isinstance(config, str):
```
timj, line 153: `RegistryConfig(None)` already acts like `RegistryConfig()` so you don't need to second guess the `RegistryConfig` constructor here.  Something like:

```python
if not isinstance(config, RegistryConfig):
    if config is None or isinstance(config, str):
        config = RegistryConfig(config)
    else:
        raise TypeError(...)
```
should work.  Makes me think that ConfigSubset could do with a classmethod for doing that check to make it consistent with what `Config` already can do.  We seem to do similar checks in fromConfig everywhere and `RegistryConfig.fromOther(config)` would be able to return either `config` or `RegistryConfig(config)` or raise.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -122,19 +126,70 @@ class Registry:
...
+            config = RegistryConfig(config)
+        elif config is None:
+            config = RegistryConfig()
+        elif not isinstance(config, RegistryConfig):
+            raise TypeError(f"Incompatible Registry configuration type: {type(config)}")
+        config.replaceRoot(butlerRoot)
+
+        if isinstance(dimensionConfig, str):
```
timj, line 161: Case in point for having the `fromOther` (or `fromConfig` class method.

python/lsst/daf/butler/core/dimensions/_config.py
```diff
@@ -62,11 +63,81 @@ class DimensionConfig(ConfigSubset):
...
     defaultConfigFile = "dimensions.yaml"
 
+    def __init__(self, other: Union[Config, ButlerURI, str, None] = None, validate: bool = True,
+                 searchPaths: Optional[Iterable[Union[str, ButlerURI]]] = None):
+        # if argument is not None then do not load/merge defaults
+        mergeDefaults = other is None
+        super().__init__(other=other, validate=validate, mergeDefaults=mergeDefaults,
+                         searchPaths=searchPaths)
+
+    def _updateWithConfigsFromPath(self, searchPaths: Iterable[Union[str, ButlerURI]],
```
timj, line 94: Rather than copying all this code, can we have a class property that DimensionConfig can set to indicate whether we are finding first or merging all found?

python/lsst/daf/butler/script/createRepo.py
```diff
@@ -31,6 +32,8 @@ def createRepo(repo, seed_config=None, standalone=False, override=False, outfile
         URI to the location to create the repo.
     seed_config : `str` or `None`
         Path to an existing YAML config file to apply (on top of defaults).
+    dimension_config : `str` or `None`
```
timj, line 35: I think this should also be able to take a `Config`.

timj (commented):
I have looked at the Config code and have some comments.

If we have a makeRepoAndButler would that reverse some of the changes we had to make to tests?

python/lsst/daf/butler/registry/_registry.py
timj: Or in fact I could just make `ConfigSubset.__init__` do the type check to ensure that you aren't passing a RegistryConfig to a DimensionConfig constructor.

python/lsst/daf/butler/_butler.py
andy-slac: I believe that if `standalone` is false then `config` does not need to have all individual sub-keys. If "registry" key is missing from `config` then `RegistryConfig(config)` works differently - it loads defaults from `registry.yaml` but then also merges `config` contents, which I think is wrong. `RegistryConfig(config.get("registry"))` should be safer in that respect - if "registry" is missing then it does not try to merge `config`, only loads defaults.

python/lsst/daf/butler/registry/_registry.py
andy-slac: Hmm, in this case `butlerRoot` is only used as an argument to `RegistryConfig.replaceRoot()`. I'm not sure that we can use generic URI with something like `sqlite://<butlerRoot>/gen3.sqlite3`?

python/lsst/daf/butler/registry/_registry.py
andy-slac: I'm OK with whatever you do to ConfigSubset, but in general DimensionConfig inheriting from ConfigSubset is an accident, it's only because I don't want to reimplement all good stuff in that class. Maybe we need a separate Config subclass which can be used by both ConfigSubset and DimensionConfig?

python/lsst/daf/butler/core/dimensions/_config.py
andy-slac: I thought about but did not want to mess with the base class just yet. But you are right, it's easier to do via policy in the base class implementation.

python/lsst/daf/butler/script/createRepo.py
andy-slac: This is the CLI-level API, it only works with strings, so it does not make much sense to allow other types here.

python/lsst/daf/butler/_butler.py
andy-slac: I'll drop `createRegistry`, indeed it makes no sense to try to keep it any longer.

andy-slac:
Regarding `makeRepoAndButler` (or `makeRepoButler`) - it will not reverse changes to tests on this ticket but rather simplify a little bit few tests that I did not have to touch on this ticket.

python/lsst/daf/butler/script/createRepo.py
TallJimbo: I realize this is pre-existing, and so probably out of scope, but I don't see that this extra level of indirection adds any value.  Surely the two lines of implementation here qualify as the kind of "minimal" that we'd permit in the actual Click subcommand definition, right?  May be a question for @n8pease, actually.

TallJimbo (approved):
Looks good!

I have not commented on any of the config discussions; it seems like you and @timj have that under control and I would only add noise given my limited knowledge of how `ConfigSubset` works.

What appears (in GitHub, at least) to be the second commit has a typo, "otehrs".

python/lsst/daf/butler/script/createRepo.py
timj: Sometimes the implementations are quite long and sometimes they are very short. The original motivation was to simplify testing by being able to call the script/ version from test files (which is what the convention is for non-click command lines) without going through forking a subprocess. Of course now we are using the click test runner thing so that is no longer relevant.

python/lsst/daf/butler/registry/_registry.py
andy-slac: I want to rebase our development branch ASAP before we diverge too much from master, so I'm going to merge these changes and we can work on harmonizing Config classes later.

PR #396, opened 2020-10-19
DM-26324: Enable sorting of datasets during export

timj:


python/lsst/daf/butler/core/datasets/type.py
```diff
@@ -222,6 +222,15 @@ def __hash__(self) -> int:
         return hash((self._name, self._dimensions, self._storageClassName,
                      self._parentStorageClassName))
 
+    def __lt__(self, other: Any) -> bool:
+        """Sort using the dataset type name.
+        """
+        if not isinstance(other, type(self)):
+            other_name = str(other)
```
TallJimbo: What kind of heterogenous sorting is this intended to support?  I would have thought this could safely raise `TypeError` or something.

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -139,20 +139,25 @@ def __str__(self) -> str:
...
         # need to ensure that sorting a DatasetRef matches what you would
         # get if you sorted DatasetType+DataCoordinate
         # Use the dataset_id as a tie breaker
         if not isinstance(other, type(self)):
             return False
 
+        # Group by run if defined, takes precedence over DatasetType
+        self_run = "" if self.run is None else self.run
+        other_run = "" if other.run is None else other.run
+
```
TallJimbo, line 152: Sorting by run should make the dataset_id tiebreaker unnecessary.

TallJimbo (approved):
Looks great!  Thanks for taking care of my pickling problems.

python/lsst/daf/butler/core/datasets/type.py
timj: I can fail here no problem. I was mostly trying to not blow up if somebody ended up with a DatasetType and DatasetRef somehow in the same collection.

PR #399, opened 2020-10-21
DM-27256: Add JSON config support

timj:


tests/test_config.py
```diff
@@ -366,15 +366,29 @@ def testHierarchy(self):
...
+            self.assertEqual(c["formatters", "calexp"], 3)
+            self.assertEqual(c["testing"], "hello")
+
+        # Round trip JSON -> Config -> YAML -> Config -> JSON -> Config
+        c1 = Config.fromString(serialized["json"], format="json")
+        yaml = c1.dump(format="yaml")
+        c2 = Config.fromString(yaml, format="yaml")
+        json = c2.dump(format="json")
+        c3 = Config.fromString(json, format="json")
+        self.assertEqual(c3.toDict(), c1.toDict())
```
andy-slac, line 403: Would it be useful to add tests for exception when `format` is not recognized or format is different from content (you pass YAML string for `format="json"`)?

andy-slac (approved):
Looks OK, one minor suggestion.

tests/test_config.py
timj: done. You can always parse JSON with a YAML parser so the default for `fromString()` actually works fine in both cases.

PR #401, opened 2020-10-22
DM-27266: Switch format for dimensions in database to JSON

timj:
JSON is more compact and faster to parse.

### DIFF
--- a/python/lsst/daf/butler/registry/_registry.py
+++ b/python/lsst/daf/butler/registry/_registry.py
@@ -89,7 +89,7 @@ if TYPE_CHECKING:
 _LOG = logging.getLogger(__name__)
 
 # key for dimensions configuration in attributes table
-_DIMENSIONS_ATTR = "config:dimensions.yaml"
+_DIMENSIONS_ATTR = "config:dimensions.json"
 
 
 class Registry:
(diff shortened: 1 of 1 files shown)


python/lsst/daf/butler/registry/_registry.py
```diff
@@ -284,11 +284,11 @@ def __init__(self, database: Database, *,
             # store managers and their versions in attributes table
             context.addInitializer(lambda db: versions.storeManagersConfig())
             context.addInitializer(lambda db: versions.storeManagersVersions())
-            # dump universe config as yaml into attributes
+            # dump universe config as json into attributes (faster than YAML)
             # `dimCfg` is to make freaking mypy happy
```
timj: This is quite confusing to me because it means that mypy sees a path through the earlier code where dimensionConfig can still be None -- I can't see it though. We have an "if None" test above and in that branch we always set it or raise so it seems like a bug in mypy.  The odd thing is that it doesn't get upset by the line below that assigned dimensionConfig to dimCfg.

@TallJimbo can you see the problem?

andy-slac (approved):
Looks good.

python/lsst/daf/butler/registry/_registry.py
andy-slac: Maybe it gets confused by context manager which wraps assignment to `dimensionConfig`?

TallJimbo: I was confused when I saw it originally in the review, too, but I didn't see an obvious explanation.

andy-slac: Context managers can potentially suppress exceptions inside `with` clause, so this is probably the path that `mypy` has to assume still exists (I presume it does not analyze how our context manager is implemented). I think I also tried to add `assert dimensionConfig is not None` after that context but it did not help with mypy.

timj:
Just realized that we have the row label be `config:dimensions.yaml` -- I think I probably want to change that to `config:dimensions` and drop the `.yaml`.

python/lsst/daf/butler/registry/_registry.py
timj: I fixed the mypy problem -- mainly by calling dump first and then copying the resulting Optional[str] to a str

PR #403, opened 2020-10-22
DM-27280: Fix sorting of DataCoordinate

timj:


TallJimbo (commented):
Looking and thinking about this more closely, I'm a bit torn about what to do here.  The potential problem is that I've taken great pains to make sure `DataCoordinate.__eq__` does not depend on whether implied values are present, but `DataCoordinate.__lt__` now does.  And that means two `DataCoordinate` instances can now simultaneously satisfy `a == b` and `a < b`.

What do you think about making `DataCoordinate.__lt__` use `tuple(self.items())` (assuming that isn't buggy as discussed earlier on Slack - what I really mean is have it just sort on the `required` key-value pairs, which is _supposed_ to be what `items` returns), and then just use `astropy.table.Table.sort` to order the `query-datasets` output?

timj:
Switching to `tuple(self.items())` does work and of course you end up with:
```
type     run      id band instrument detector physical_filter exposure
---- ----------- --- ---- ---------- -------- --------------- --------
 raw HSC/raw/all  14    r        HSC        0           HSC-R   903344
 raw HSC/raw/all  17    r        HSC        1           HSC-R   903346
 raw HSC/raw/all  33    i        HSC        1           HSC-I   904014
 raw HSC/raw/all  11    r        HSC        4           HSC-R   903342
 raw HSC/raw/all  30    i        HSC        4           HSC-I   904010
 raw HSC/raw/all  13    r        HSC        5           HSC-R   903344
 raw HSC/raw/all  16    r        HSC        6           HSC-R   903346
 raw HSC/raw/all  32    i        HSC        6           HSC-I   904014
```
so yes, we'd have to do the sort in Table (and I have to fix Table creation to stop it inserting strings into all the columns.

timj:
And of course we don't want to sort on the id column at all so I will have to look a bit harder at that (maybe moving `id` to the end would work)

python/lsst/daf/butler/script/queryDatasets.py
```diff
@@ -78,22 +78,46 @@ def getAstropyTable(self, datasetTypeName):
...
+        dataset_table = AstropyTable(np.array(rows), names=columnNames, dtype=columnTypes)
+
+        # For sorting we want to ignore the id
+        # We also want to move temporal or spatial dimensions earlier
+        sort_first = ["type", "run"]
+        sort_early = []
+        sort_late = []
+        for dim in dimensions:
+            if dim.spatial or dim.temporal:
+                sort_early.append(str(dim))
```
TallJimbo: This will sort `visit` before `instrument`, which doesn't make sense, either.  Something like
```
sort_early.extend(dim.required.names)
```
instead of
```
sort_early.append(str(dim))
```
would solve that, but it'd lead to duplicates across the various `sort_` lists that you'd then need to remove.

timj: I can take another look. We haven't yet dealt with single runs having datasets with different instruments. One option is to simply state up front that instrument goes after run in the sorting and then leave everything else as is.

timj: Okay. Your fix worked fine.

PR #405, opened 2020-10-27
DM-27321: Add day_obs and seq_num to exposure record

timj:
Also rename exposure.name to exposure.obs_id to make it look like obsCore.

### DIFF
--- a/python/lsst/daf/butler/configs/dimensions.yaml
+++ b/python/lsst/daf/butler/configs/dimensions.yaml
@@ -201,7 +201,7 @@ elements:
         name: id
         type: int
       -
-        name: name
+        name: obs_id
         type: string
         length: 64
     requires: [instrument]
--- a/python/lsst/daf/butler/configs/dimensions.yaml
+++ b/python/lsst/daf/butler/configs/dimensions.yaml
@@ -143,6 +143,13 @@ elements:
     implies: [physical_filter, visit_system]
     metadata:
       -
+        name: day_obs
+        type: int
+        doc: >
+          Day of observation as defined by the observatory (YYYYMMDD format).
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/configs/dimensions.yaml
```diff
@@ -142,6 +142,12 @@ elements:
     requires: [instrument]
     implies: [physical_filter, visit_system]
     metadata:
+      -
+        name: day_obs
+        type: int
+        doc: >
+          Day of observation as defined by the observatory (YYYYMMDD format).
+          If a visit crosses multiple days this entry will be null.
```
RobertLuptonTheGood: Is this correct?  I thought it was the day when the observation is started;  LSST needs a dayObs for all data we take (not quite the same as this day_obs)

python/lsst/daf/butler/configs/dimensions.yaml
```diff
@@ -227,6 +233,18 @@ elements:
...
+      -
+        name: day_obs
+        type: int
+        doc: >
+          Day of observation as defined by the observatory (YYYYMMDD format).
+      -
+        name: seq_num
+        type: int
+        doc: >
+          Counter for the observation within a larger sequence. Contxt
```
RobertLuptonTheGood: "Context"

RobertLuptonTheGood (approved):

python/lsst/daf/butler/configs/dimensions.yaml
timj: I can change it for visit so that it returns the smallest day_obs if that's what you prefer.

RobertLuptonTheGood: Yes please.

PR #407, opened 2020-10-27
DM-26944: Move certifyCalibrations.py content to a butler command

czwa:
Add butler certify-calibrations.

### DIFF
--- a/python/lsst/daf/butler/cli/cmd/__init__.py
+++ b/python/lsst/daf/butler/cli/cmd/__init__.py
@@ -20,6 +20,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 __all__ = ("butler_import",
+           "certify_calibrations",
            "create",
            "config_dump",
            "config_validate",
@@ -28,11 +29,11 @@ __all__ = ("butler_import",
            "query_dataset_types",
            "query_datasets",
            "remove_dataset_type",
-           "certify_calibrations",
            )
 
 
 from .commands import (butler_import,
+                       certify_calibrations,
                        create,
                        config_dump,
                        config_validate,
@@ -41,5 +42,4 @@ from .commands import (butler_import,
                        query_dataset_types,
                        query_datasets,
                        remove_dataset_type,
-                       certify_calibrations,
                        )
--- a/python/lsst/daf/butler/cli/cmd/__init__.py
+++ b/python/lsst/daf/butler/cli/cmd/__init__.py
@@ -28,6 +28,7 @@ __all__ = ("butler_import",
            "query_dataset_types",
            "query_datasets",
            "remove_dataset_type",
+           "certify_calibrations",
            )
 
 
@@ -40,4 +41,5 @@ from .commands import (butler_import,
                        query_dataset_types,
                        query_datasets,
                        remove_dataset_type,
+                       certify_calibrations,
                        )
--- a/python/lsst/daf/butler/cli/cmd/commands.py
+++ b/python/lsst/daf/butler/cli/cmd/commands.py
@@ -202,3 +202,21 @@ def query_datasets(**kwargs):
         print("")
         table.pprint_all()
     print("")
+
+
+@click.command()
+@repo_argument(required=True)
+@click.argument('input-collection')
+@click.argument('output-collection')
+@click.argument('dataset-type-name')
+@click.option("--begin-date", "-b", type=str, default=None,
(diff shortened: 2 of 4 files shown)


python/lsst/daf/butler/cli/cmd/__init__.py
```diff
@@ -28,6 +28,7 @@
            "query_dataset_types",
            "query_datasets",
            "remove_dataset_type",
+           "certify_calibrations",
```
timj: Minor point but alphabetical order please.

python/lsst/daf/butler/cli/cmd/__init__.py
```diff
@@ -40,4 +41,5 @@
                        query_dataset_types,
                        query_datasets,
                        remove_dataset_type,
+                       certify_calibrations,
```
timj: Make alphabetical.

python/lsst/daf/butler/cli/cmd/commands.py
```diff
@@ -202,3 +202,21 @@ def query_datasets(**kwargs):
...
     print("")
+
+
+@click.command()
+@repo_argument(required=True)
+@click.argument('input-collection')
+@click.argument('output-collection')
+@click.argument('dataset-type-name')
+@click.option("--begin-date", "-b", type=str, default=None,
+              help="Datetime (TAI) of the beginning of the validity range for the certified calibrations.")
```
timj: You need to say if this is ISO format or MJD.

python/lsst/daf/butler/script/certifyCalibrations.py
```diff
@@ -0,0 +1,72 @@
...
+    end_date : `str`, optional
+        Date this calibration should stop being used.
+    search_all_inputs : `bool`, optional
+        Search all children of the inputCollection if it is a CHAINED
+        collection, instead of just the most recent one.
+    """
+    butler = Butler(repo, writeable=True)
+    registry = butler.registry
+    timespan = Timespan(
+        begin=astropy.time.Time(begin_date) if begin_date is not None else None,
```
timj: Might be more explicit to add `scale="tai"` here and below so we are completely certain these are parsed as TAI.

@TallJimbo should we be supporting the same syntax we support for queries? ie https://pipelines.lsst.io/v/daily/modules/lsst.daf.butler/queries.html#time-literals ? (the leading T is not needed of course -- I mean the `mjd/` and `/tai` stuff -- we have code for converting that string to an astropy Time.

Maybe we support ISO string now and later on can expand if asked (because an ISO string is compatible with how the T-string works).

python/lsst/daf/butler/script/certifyCalibrations.py
```diff
@@ -0,0 +1,72 @@
...
+        collection.
+    outputCollection : `str`
+        Data collection to store final calibrations.  If it already exists, it
+        must be a `~CollectionType.CALIBRATION` collection.  If not, a new
+        `~CollectionType.CALIBRATION` collection with this name will be
+        registered.
+    datasetTypeName : `str`
+        Name of the dataset type to certify.
+    begin_date : `str`, optional
+        Date this calibration should start being used.
```
timj: Should state format and time scale.

timj (approved):
Looks good. Main comment is about documentation of time formats.

python/lsst/daf/butler/cli/cmd/commands.py
```diff
@@ -202,3 +202,21 @@ def query_datasets(**kwargs):
...
         table.pprint_all()
     print("")
+
+
+@click.command()
+@repo_argument(required=True)
+@click.argument('input-collection')
+@click.argument('output-collection')
+@click.argument('dataset-type-name')
+@click.option("--begin-date", "-b", type=str, default=None,
```
timj: @n8pease might want to comment but can we not use the `-b` and `-e` short options to begin with until we get a better feel as to whether they are confusing. We use `-b` to mean butler repo in `pipetask` and we are trying not to reuse short options if we don't have to.

czwa: I've removed them, and updated `ci_cpp_gen3` to use the full `--begin-date` style formats.

python/lsst/daf/butler/script/certifyCalibrations.py
TallJimbo: Yeah, I brought this up on slack a while back and never followed up with a ticket as I'd promised: https://lsstc.slack.com/archives/C2JPL8U78/p1600964518003000

PR #409, opened 2020-10-28
DM-27293: Extend expression parser with new constructs and options

andy-slac:
Summary of changes:

- new mechanism for identifier substitution (a.k.a. bindparams), parser now takes a Mapping that can map identifier names into `Node` instances.
- `POINT(ra, dec)` declares sky coordinates, generates new node type in the tree
- time range can be specified with a tuple of 2 time literals, tuples also make new node type
- `[NOT] IN` supports more options on right hand side, can be used with time ranges and regions
- new `OVERLAPS` operator that can work with time ranges and regions.

Just as before parser only builds parse tree, actual work for transforming that into reasonable output is responsibility of visitors. Visitor interface was extended for new node types, but implementation of existing visitors need are vision to actually do something useful.


python/lsst/daf/butler/registry/queries/exprParser/exprTree.py
```diff
@@ -317,6 +317,32 @@ def __str__(self) -> str:
...
+    Tuple is used to represent time ranges, for now parser supports tuples
+    with two items, though this class can be used to represent different
+    number of items in sequence.
+
+    Attributes
+    ----------
+    items : tuple of Node
+        Expressions inside parentheses.
+    """
+    def __init__(self, items: Tuple[Node]):
```
TallJimbo: I think the type annotation here should either be `Tuple[Node, ...]` (for any number of elements) or `Tuple[Node, Node]` (for 2-tuples only).  I think you're supporting only 2-tuples in the syntax here, but I wasn't sure if you were trying to make the visitor interface extensible to more later.

TallJimbo (approved):
Looks good!  ~The only thing I can think of that I didn't see in the PR was identifiers with more than one `.` separator, but it may just be that this is already supported and I didn't know it.~ edited: I just didn't see the commit originally.

I was a bit surprised to see the bindparam support here, though, because I was convinced by your previous argument on Jira that we could just use the existing `Identifier` nodes to inject bindparams via the visitor interface.  Is there an advantage to doing it in the parser instead?

python/lsst/daf/butler/registry/queries/exprParser/exprTree.py
andy-slac: I'll add `...`, indeed the language grammar now allows only two items, but parsed tree is separate from grammar so you can build your own expression trees, and I can't exclude we can extend grammar with longer tuples if needed.

andy-slac:
Sorry I did not mention two-dot identifiers, but I indeed added support for that too. For now only one or two dots are allowed, if you think you need more it's easy to update lexer for that. I did not think that supporting infinite number of dots is useful so I limited that to what you asked on a ticket.

For bindparams - I thought about implementing that at visitor level but then figured out that it is much simpler to do it in parser. With visitor I'd have to provide some reasonable implementation in the base class and I generally prefer to keep implementation separate from abstract interfaces so it would probably need a new intermediate base class or a mixin for a visitor to support that feature. And overriding `visitIdentifier` would need more attention to do the right thing in that case (including avoiding potential recursion loops). Still if you want it to happen in visitor there are two options:

- I can move that implementation to a visitor base class
- you can ignore parser feature and re-implement it in a specific visitor class

TallJimbo:
Sounds good r.e. bindparams.  I don't know yet whether it will be necessary to implement them in specific visitors (e.g. because we need to parse before we have bindparam values available in some context), but having support in the parser certainly does no harm, and I don't think there's any need to add it to the base visitors.

PR #410, opened 2020-10-29
DM-26403: Reorganize ButlerURI into separate files

fubarwrangler:
Splitting butler URI backends into separate files for ease of eventual splitting off into separate package.


python/lsst/daf/butler/core/location.py
```diff
@@ -33,7 +33,7 @@
     Union,
 )
 
-from ._butlerUri import ButlerURI
+from .uri import ButlerURI
```
timj: I still wanted the directory to be called `_butlerUri`  -- which would mean this line doesn't change.  I understand that the s3utils and webdavutils imports into the tests then look weird but I think I can live with that.  In the longer term the s3utils and webdavutils tests should disappear because ButlerURI should be testing all that anyhow.  That leaves us only with the test_butler.py imports which need a bit more thought.

python/lsst/daf/butler/core/uri/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Should only expose ButlerURI
+
+from ._butlerUri import ButlerURI
+__all__ = ("ButlerURI",)
```
timj: We don't usually put `__all__` in `__init__.py` since for the way we use `__init__.py` it mostly is for forwarding symbols anyhow. The `__all__` here does not serve any purpose so can be removed.

timj (approved):

python/lsst/daf/butler/core/uri/_butlerUri.py
```diff
@@ -0,0 +1,668 @@
...
+                from .file import ButlerFileURI
+                subclass = ButlerFileURI
+            elif parsed.scheme == "s3":
+                from .s3 import ButlerS3URI
+                subclass = ButlerS3URI
+            elif parsed.scheme.startswith("http"):
+                from .http import ButlerHttpURI
+                subclass = ButlerHttpURI
+            elif parsed.scheme == "resource":
+                # Rules for scheme names disasllow pkg_resource
```
timj: Minor typo in this comment. Maybe that was mine originally 😄

python/lsst/daf/butler/core/location.py
fubarwrangler: You want me to rename the whole uri/ directory to _butlerUri/ then? Do I keep a file called _butlerUri.py in that directory too?

python/lsst/daf/butler/core/uri/__init__.py
fubarwrangler: Understood, wasn't sure will fix

python/lsst/daf/butler/core/uri/_butlerUri.py
fubarwrangler: Fixed!

python/lsst/daf/butler/core/location.py
timj: Yes, I want the directory renamed and the _butlerUri.py to still be in that subdirectory.

fubarwrangler:
This should be ready now, rebased into 2 commits

PR #413, opened 2020-10-30
DM-27343: Test for URI existence before reporting bad file extension

timj:
If ButlerConfig is constructed with a directory that
does not exist, the butler.yaml will not be appended.
In that case we want to report that the directory
does not exist, not that it does not have .yaml
extension.

Only do the existence check if we know we are about to raise.
It can be an expensive check and if we pass the extension
test we will get a FileNotFoundError later.


n8pease:
looks good.

PR #414, opened 2020-10-30
DM-27253: reimplement commonSkyPix overlaps with more general tables

TallJimbo:
This doesn't change the functionality available at a high level; it just defines tables and starts to define interfaces that we can use to materialize more spatial joins in the future.


python/lsst/daf/butler/registry/dimensions/governor.py
```diff
@@ -64,12 +64,13 @@ def __init__(self, db: Database, dimension: GovernorDimension, table: sqlalchemy
         self._dimension = dimension
```
andy-slac, line 62: From old commit - lines 50-53 above - duplicate `Parameters` heading.

python/lsst/daf/butler/registry/dimensions/overlaps.py
```diff
@@ -0,0 +1,191 @@
...
+
+    New instances should be constructed by calling `initialize`, not by calling
+    the constructor directly.
+
+    Parameters
+    ----------
+    db : `Database`
+        Interface to the underlying database engine and namespace.
+    elementStorage : `tuple` [ `DatabaseDimensionRecordStorage` ]
+        Storage objects for the elements this object will related.
```
andy-slac: `will related` - `will be related to`? And next parameter too.

python/lsst/daf/butler/registry/dimensions/static.py
```diff
@@ -56,27 +66,74 @@ class StaticDimensionRecordStorageManager(DimensionRecordStorageManager):
...
+        self._overlaps = overlaps
 
     @classmethod
     def initialize(cls, db: Database, context: StaticTablesContext, *,
                    universe: DimensionUniverse) -> DimensionRecordStorageManager:
         # Docstring inherited from DimensionRecordStorageManager.
+        # Start by initializing governor dimensions; those go both in the main
+        # 'records' mapping we'll pass to init, and a local dictionary that we
+        # can pass in when initializing storage for DatabaseDimensionElements.
+        governors: NamedKeyDict[GovernorDimension, GovernorDimensionRecordStorage] = NamedKeyDict()
```
andy-slac: Just curious - could there be syntax like:

    governors = NamedKeyDict[GovernorDimension, GovernorDimensionRecordStorage]()

to reduce strain on my eyes?

python/lsst/daf/butler/registry/dimensions/table.py
```diff
@@ -158,13 +228,560 @@ def sync(self, record: DimensionRecord) -> bool:
...
+                levels.sort(reverse=True)
+                # Start with the first level, which is the finest-grained one.
+                # Compute skypix envelope indices directly for that.
+                indices: Dict[int, Set[int]] = {levels[0]: set()}
+                for begin, end in system[levels[0]].pixelization.envelope(record.region):  # type: ignore
+                    indices[levels[0]].update(range(begin, end))
+                # Divide those indices by powers of 4 (and remove duplicates)
+                # work our way up to the last (coarsest) level.
+                for lastLevel, nextLevel in zip(levels[:-1], levels[1:]):
+                    factor = 4**(lastLevel - nextLevel)
```
andy-slac, line 693: Is this true for all known pixelization systems? Or do we only care about HTM?

andy-slac (approved):
Looks OK, though it's a bit dense in places and I'm not sure I could understand every line of it 🙂

python/lsst/daf/butler/registry/dimensions/overlaps.py
TallJimbo: :+1: was supposed to be "will relate".

python/lsst/daf/butler/registry/dimensions/static.py
TallJimbo: Good point!  That syntax doesn't always work (for example with `Optional` or `Union`), so I always forget to use it when it does.  I will here.

python/lsst/daf/butler/registry/dimensions/table.py
TallJimbo: I am fairly certain it is true for HTM, Q3C, and healpix (for the "nested" healpix numbering system that we'd use, at least).

PR #415, opened 2020-10-31
DM-27390: replace DimensionGraph.{encode,decode}

TallJimbo:


python/lsst/daf/butler/core/dimensions/_graph.py
```diff
@@ -229,46 +231,23 @@ def __str__(self) -> str:
...
         """
-        mask = 0
-        for dimension in self.dimensions:
-            index = self.universe.getDimensionIndex(dimension.name)
-            mask |= (1 << index)
-        return mask.to_bytes(self.universe.getEncodeLength(), byteorder="big")
+        hasher = hashlib.blake2b(digest_size=self.DIGEST_SIZE//2)
+        for name in self.required.names:
+            hasher.update(name.encode("ascii"))
+        return hasher.hexdigest()
```
andy-slac: I do not trust hashes in general due to potential collisions. In this case collisions are possible for cases like `["a", "b"]` vs just `["ab"]`. It may help adding a non-word separator character (e.g. TAB) to avoid this issue.

python/lsst/daf/butler/registry/dimensions/static.py
```diff
@@ -50,6 +52,33 @@
...
+            ddl.FieldSpec(
+                name="digest",
+                dtype=sqlalchemy.String,
+                length=DimensionGraph.DIGEST_SIZE,
+                primaryKey=True,
+            ),
+            ddl.FieldSpec(
+                name="dimension_name",
+                dtype=sqlalchemy.String,
+                length=64,
```
andy-slac: We should have probably defined a bunch of constants for these `64` magic numbers too.

andy-slac (commented):
Looks OK, one minor comment.

python/lsst/daf/butler/registry/dimensions/static.py
timj: Are we sure they don't get converted to TEXT...?

python/lsst/daf/butler/core/dimensions/_graph.py
TallJimbo: I'll at least add that separator; good idea.

I also considered just using an autoincrement integer for the ID (so there would be no `DimensionGraph` method to get it).  The problem with that is that we'd have to load essentially the entire table from the database in order to see if a particular graph already has an ID, though that's more about code complexity than performance because the table should always be small.  I also liked the idea of making the ID deterministic across different data repositories, but I don't have a concrete reason why that needs to be the case.  Do you think an autoincrement ID would be better here?

python/lsst/daf/butler/registry/dimensions/static.py
TallJimbo: At this point I'd be more inclined to just use `sqlalchemy.Text` and leave the length blank. So many of them are arbitrary, or at best "guess and check".

TallJimbo: (messages crossed) I think this does get translated to TEXT, but that's controlled by different code, so the magic number here is still a bit problematic.

timj: My main comment is to note that if you needed it to be exactly 64 then it probably isn't happening. The threshold is >32 for auto conversion to TEXT. You can use TEXT explicitly in the type if you want it to be TEXT. You can't though require a long varchar to actually be a varchar.

python/lsst/daf/butler/core/dimensions/_graph.py
andy-slac: Deterministic ID would be nice to have but I guess it's hard to construct or make it stable. If we do not need this ID outside database relation context then autoincrement is probably a better match for that. That table is small so it just a single query/round-trip to database, so I would not worry about performance, and code complexity should not be terrible.

TallJimbo: I've added another commit that uses autoincrement integer IDs instead; please take a look.  Once the review is finished I'll squash commits and probably just remove `DimensionGraph.digest`.

python/lsst/daf/butler/registry/dimensions/static.py
```diff
@@ -50,6 +53,33 @@
 _VERSION = VersionTuple(5, 0, 0)
 
 
+def _makeDimensionGraphTableSpecs() -> ddl.TableSpec:
```
andy-slac: Is this still used?

python/lsst/daf/butler/registry/dimensions/static.py
```diff
@@ -172,3 +217,168 @@ def schemaDigest(self) -> Optional[str]:
...
+        ----------
+        key : `int`
+            Integer used as the unique key for this `DimensionGraph` in the
+            database.
+
+        Returns
+        -------
+        graph : `DimensionGraph`
+            Retrieved graph.
+        """
```
andy-slac, line 352: Maybe add `Raises` section to docstring?

andy-slac (approved):

python/lsst/daf/butler/registry/dimensions/static.py
TallJimbo: I thought I'd deleted that.  I certainly will now.

PR #416, opened 2020-11-02
DM-26405: Move S3-backoff into actual S3 class

fubarwrangler:
Some question on which methods get all_errors or just client_errors for
backoff. For now this at least passes tests...

### DIFF
--- a/python/lsst/daf/butler/core/_butlerUri/s3.py
+++ b/python/lsst/daf/butler/core/_butlerUri/s3.py
@@ -45,8 +45,6 @@ from botocore.exceptions import ClientError
 from http.client import ImproperConnectionState, HTTPException
 from urllib3.exceptions import RequestError, HTTPError
 
-
-
 if TYPE_CHECKING:
     try:
         import boto3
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/core/_butlerUri/s3.py
```diff
@@ -39,6 +41,9 @@
 from ._butlerUri import ButlerURI
 from .s3utils import getS3Client, s3CheckFileExists, bucketExists
 
+from botocore.exceptions import ClientError
+from http.client import ImproperConnectionState, HTTPException
+from urllib3.exceptions import RequestError, HTTPError
```
ktlim, line 46: Since these are only used once, it may be better to import the module and use a fully-qualified name instead.  On the other hand, that makes everything in the module available (unnecessarily) as opposed to just the imported symbol(s).

ktlim (approved):

python/lsst/daf/butler/core/_butlerUri/s3.py
fubarwrangler: Is there any difference to python (performance or otherwise) to plucking only needed symbols from a module?

timj: I don't think so. It's more of a convention where you don't want ambiguous bare symbols turning up in the code. In this particular case I think `HTTPError` is fairly unambiguous where it came from so I'm fine with leaving it as is.

ktlim: Well, if it were obvious, the comments wouldn't be needed at point-of-use :).  But I'm also fine with leaving it.

fubarwrangler:
Rebased onto master, awaiting tests. Then I will merge.

PR #417, opened 2020-11-02
DM-27151: Support exposure.obs_id or detector.full_name in dataId

timj:


python/lsst/daf/butler/_butler.py
```diff
@@ -547,6 +547,62 @@ def _findDatasetRef(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
...
+                if dimension in newDataId:
+                    log.warning("DataId specified explicit %s dimension value of %s in addition to"
+                                " general record specifiers for it of %s.  Ignoring record information.",
+                                dimension, newDataId[dimension], str(values))
+                    continue
+
+                # Build up a WHERE expression -- use single quotes
+                def quote(s):
+                    if isinstance(s, str):
+                        return f"'{s}'"
```
TallJimbo, line 590: Might need to quote `astropy.time.Time` here, too.  But hopefully this will be replaced by bindparam support in the query expressions before that actually matters.

python/lsst/daf/butler/_butler.py
```diff
@@ -547,6 +547,62 @@ def _findDatasetRef(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
...
+                    else:
+                        return s
+
+                where = " AND ".join(f"{dimension}.{k} = {quote(v)}"
+                                     for k, v in values.items())
+
+                # Hopefully we get a single record that matches
+                records = list(self.registry.queryDimensionRecords(dimension, dataId=newDataId, where=where))
```
TallJimbo: I think you want to wrap this in `set` instead of `list` do be safe re duplicates.  I doubt it's necessary for any of the queries this would do in practice, but I'm not completely certain.

python/lsst/daf/butler/_butler.py
```diff
@@ -547,6 +547,62 @@ def _findDatasetRef(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
...
+                            log.debug("- %s", str(r))
+                        raise RuntimeError(f"DataId specification for dimension {dimension} is not uniquely"
+                                           f" constrained to a single dataset by {values}."
+                                           f" Got {len(records)} results.")
+                    raise RuntimeError(f"DataId specification for dimension {dimension} matched no records"
+                                       f" when constrained by {values}")
+
+                # With the matching record, it is always the "id" field
+                # that we need to use for the dataId
+                newDataId[dimension] = records[0].id
```
TallJimbo: I don't think this is guaranteed by anything more than convention (and something silly but otherwise valid like `{"physical_filter.band": 'r'}` would violate it).  This is clunky, but:
```
newDataId[dimension] = getattr(records[0], dimension.primaryKey.name
```
is the safe thing to do here.

TallJimbo (approved):
It's a real shame this can't just boil down to something like:
```
where = " AND ".join(f"{k}={quote(v)}" for k, v in dataId.items())
(ref,) = set(self.registry.queryDatasets(datasetType, collections=collections, where=where))
return self.datastore.get(ref)
```
(with numbers-of-results checking, etc.).

It's almost there, and I think it's just the calibration collection stuff that wouldn't work with that approach.  That's basically the same problem that still exists in QG gen and what some of my schema changes this week are intended to support fixing.  So I think we'll get there, and then this will all be easy to simplify, and that's probably the time to try to integrate this with `Registry`.  

The piece that might be worth moving elsewhere now is just the logic that takes a `dict` (and maybe a `DimensionUniverse`) and returns a `dict` and a `str` where expression, with the returned dict only having genuine dimensions as keys; I think that'd stay in more or less its current form no matter how we end up using the where expression.  I'm not sure where I'd put it, but if you make it a free function somewhere in `core` that wouldn't be bad.

python/lsst/daf/butler/_butler.py
timj: At this point `dimension` is a string and not a `Dimension`. I need to look up how to convert it to a `Dimension` that I can ask the primary key.

python/lsst/daf/butler/_butler.py
timj: I'm hoping people are not going to be forming dataIds where they are doing `exposure.datetime_begin = ...` but maybe they will. It is *unique* so it would work...

python/lsst/daf/butler/_butler.py
timj: I can do that. I hadn't realized that queryDimensionsRecords would ever return duplicates.

python/lsst/daf/butler/_butler.py
TallJimbo: `self.registry.dimensions[dimensionName]` will get you a `Dimension`.

python/lsst/daf/butler/_butler.py
TallJimbo: Heh, I guess we know that wouldn't work right now anyway, because none of the timespan fields do.

PR #423, opened 2020-11-03
DM-27354: Update for sphgeom pip install

jonathansick:
sphgeom no longer has a requirements.txt file (as of https://github.com/lsst/sphgeom/pull/30), so this PR removes the step from the build_and_test workflow that executes such a requirements.txt file.

Secondly, the install_requires dependency listing in setup.cfg now installs sphgeom from GitHub, making the sphgeom installation from a requirements.txt file unnecessary.

Note that this PR's build is expected to fail until the sphgeom PR is merged. Both PRs will need to be merged in quick succession.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ pyyaml >= 5.1
 astropy >= 4.0
 click >= 7.0
 sqlalchemy >= 1.3
-https://github.com/lsst/sphgeom/archive/master.tar.gz
+git+git://github.com/lsst/sphgeom@master#egg=lsst_sphgeom
 
 # optional
 backoff >= 1.10
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -34,8 +34,8 @@ jobs:
         run: pip install -r requirements.txt
 
       # We have two cores so we can speed up the testing with xdist
-      - name: Install xdist and openfiles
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -2,7 +2,7 @@ pyyaml >= 5.1
 astropy >= 4.0
 click >= 7.0
 sqlalchemy >= 1.3
-https://github.com/lsst/sphgeom/archive/master.tar.gz
+git+git://github.com/lsst/sphgeom@master#egg=lsst_sphgeom
```
timj, line 5: why is this better than the previous URL?

timj (approved):

requirements.txt
r-owen: I spent some time trying to find documentation about this. The following is the best link I found: https://pip.pypa.io/en/stable/reference/pip_install/#requirement-specifiers

and it pointed me to the following for the various git protocols: https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols
If I read this correctly I think the main advantage to `git+git` is speed. It can only be used for unauthenticated read, which is fine for this use case. The "cons" section of that document talks about the fact that the port is often blocked by corporate firewalls, which is worrisome.

timj:
We've done the big push so I'm fine with the sphgeom PR and this PR being merged at any time.

PR #425, opened 2020-11-04
DM-27418: Use safe_load/dump for YAML

timj:
There are also some other fixes here that came about when debugging the YAML change.

### DIFF
--- a/tests/test_butler.py
+++ b/tests/test_butler.py
@@ -129,14 +129,14 @@ class ButlerPutGetTests:
         cls.storageClassFactory = StorageClassFactory()
         cls.storageClassFactory.addFromConfig(cls.configFile)
 
-    def assertGetComponents(self, butler, datasetRef, components, reference):
+    def assertGetComponents(self, butler, datasetRef, components, reference, collections=None):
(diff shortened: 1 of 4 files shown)


TallJimbo (approved):
Test logic looks good to me.

tests/test_butler.py
```diff
@@ -129,14 +129,17 @@ def setUpClass(cls):
...
         dataId = datasetRef.dataId
         deferred = butler.getDirectDeferred(datasetRef)
+        kwds = {}
+        if collections is not None:
+            kwds["collections"] = collections
 
         for component in components:
             compTypeName = datasetType.componentTypeName(component)
-            result = butler.get(compTypeName, dataId)
+            result = butler.get(compTypeName, dataId, **kwds)
```
ktlim: Couldn't you always say `collections=collections` since it defaults to `None` anyway?

ktlim (approved):

tests/test_butler.py
timj: Thanks. For some reason I thought it defaulted to `...`

PR #426, opened 2020-11-05
DM-27435: Unification of file-like datastores

timj:


python/lsst/daf/butler/core/formatter.py
```diff
@@ -210,6 +210,27 @@ def write(self, inMemoryDataset: Any) -> str:
...
+        try:
+            # We know the arguments are incompatible
+            cls.fromBytes(cls, b"")  # type: ignore
+        except NotImplementedError:
+            return False
+        except Exception:
+            # There will be problems with the bytes we are supplying so ignore
+            pass
+        return True
+
```
TallJimbo, line 246: Is the long-term plan here to implement this in subclasses directly, and then make it abstract?

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -847,7 +845,61 @@ def _read_artifact_into_memory(self, getInfo: DatastoreFileGetInformation,
...
+                                             component=getInfo.component if isComponent else None)
+            except Exception as e:
+                raise ValueError(f"Failure from formatter '{formatter.name()}' for dataset {ref.id}"
+                                 f" ({ref.datasetType.name} from {uri}): {e}") from e
+        else:
+            # Read from file
+            with uri.as_local() as local_uri:
+                # Have to update the Location associated with the formatter
+                # because formatter.read does not allow an override.
+                # This could be improved.
```
TallJimbo, line 942: FWIW, this is indeed something I'd love to see fixed (but not on this ticket).

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -825,7 +830,69 @@ def _write_in_memory_to_artifact(self, inMemoryDataset: Any, ref: DatasetRef) ->
...
+        if uri.exists():
+            # Assume that by this point if registry thinks the file should
+            # not exist then the file should not exist and therefore we can
+            # overwrite it. This can happen if a put was interrupted by
+            # an external interrupt. The only time this could be problematic is
+            # if the file template is incomplete and multiple dataset refs
+            # result in identical filenames.
+            # Eventually we should remove the check completely (it takes
+            # non-zero time for network).
+            log.warning("Object %s exists in datastore for ref %s", uri, ref)
```
TallJimbo: Is it time to just remove this check now?

python/lsst/daf/butler/datastores/fileLikeDatastore.py
```diff
@@ -825,7 +830,69 @@ def _write_in_memory_to_artifact(self, inMemoryDataset: Any, ref: DatasetRef) ->
...
+                serializedDataset = formatter.toBytes(inMemoryDataset)
+                log.debug("Writing bytes directly to %s", uri)
+                uri.write(serializedDataset, overwrite=True)
+                log.debug("Successfully wrote bytes directly to %s", uri)
+            except NotImplementedError:
+                with tempfile.NamedTemporaryFile(suffix=uri.getExtension()) as tmpFile:
+                    # Need to configure the formatter to write to a different
+                    # location and that needs us to overwrite internals
+                    tmpLocation = Location(*os.path.split(tmpFile.name))
+                    formatter._fileDescriptor.location = tmpLocation
```
TallJimbo: Could we make this mutation a public method on the `Formatter` class without modifying all of its derived classes?

TallJimbo (approved):
Looks good.  I think all of my comments are on things you're already aware could use some cleanup, and none of them are critical for this ticket.

python/lsst/daf/butler/core/formatter.py
timj: Yes. I haven't worked out the best approach yet but there is a Formatter cleanup ticket. I wanted an API here though so that I could at least ask the formatter about it even if the implementation was a bit odd. I had hoped to get away with always using the presence of a method to declare it but it seems like I'm going to have to use the approach of a class property and implementation method.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
timj: The FileDescriptor class was one of the first classes that Pim wrote and I'm not really convinced I need it. Again I can deal with that on the formatter cleanup.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
timj: Ok.

python/lsst/daf/butler/datastores/fileLikeDatastore.py
timj: I can add a quick `Formatter.updateLocation` method although maybe with an underscore for now.

PR #427, opened 2020-11-05
DM-27422: @option_section() doesn't seem to work well with Sphinx

n8pease:


timj (approved):

n8pease:
Interesting, pushing a branch with no changes caused this PR to close itself.

PR #428, opened 2020-11-06
DM-27008: Allow for list of tuples to be passed to multilevel index dataframe

timothydmorton:
Previously, only a dictionary was allowed to pass to access columns of a DataFrame with a multi-level column index.  However, there are cases in Functor usage in which it is necessary to explicitly pass a list of tuples as column indices; this adds and tests that functionality to the butler parquet interface.

### DIFF
--- a/python/lsst/daf/butler/formatters/parquet.py
+++ b/python/lsst/daf/butler/formatters/parquet.py
@@ -21,7 +21,7 @@
 
 from __future__ import annotations
 
-__all__ = ("ParquetFormatter", )
+__all__ = ("ParquetFormatter",)
 
 import json
 import re
@@ -55,14 +55,14 @@ class _ParquetLoader:
     path : `str`
         Full path to the file to be loaded.
     """
-
     def __init__(self, path: str):
         self.file = pq.ParquetFile(path)
         self.md = json.loads(self.file.metadata.metadata[b"pandas"])
         indexes = self.md["column_indexes"]
         if len(indexes) == 1:
-            self.columns = pd.Index(name for name in self.file.metadata.schema.names
-                                    if not name.startswith("__"))
+            self.columns = pd.Index(
+                name for name in self.file.metadata.schema.names if not name.startswith("__")
+            )
         else:
             raw_columns = list(self._splitColumnnNames(len(indexes), self.file.metadata.schema.names))
             self.columns = pd.MultiIndex.from_tuples(raw_columns, names=[f["name"] for f in indexes])
@@ -89,13 +89,15 @@ class _ParquetLoader:
         tuple : `tuple` of `str`
             A multi-index column name tuple.
         """
-        pattern = re.compile(r"\({}\)".format(', '.join(["'(.*)'"] * n)))
+        pattern = re.compile(r"\({}\)".format(", ".join(["'(.*)'"] * n)))
         for name in names:
             m = re.search(pattern, name)
             if m is not None:
                 yield m.groups()
 
-    def _standardizeColumnParameter(self, columns: Dict[str, Union[str, List[str]]]) -> Iterator[str]:
+    def _standardizeColumnParameter(
+        self, columns: Union[List[str], List[tuple], Dict[str, Union[str, List[str]]]]
+    ) -> Iterator[str]:
         """Transform a dictionary index into a multi-index column into a
         string directly understandable by PyArrow.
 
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -111,21 +114,36 @@ def _standardizeColumnParameter(self, columns: Dict[str, Union[str, List[str]]])
...
-                    raise ValueError(f"Unrecognized value {value!r} for index {self.indexLevelNames[i]!r}.")
-            yield str(requested)
-
-    def read(self, columns: Union[str, List[str], Dict[str, Union[str, List[str]]]] = None
-             ) -> pd.DataFrame:
+        if isinstance(columns, list):
+            for requested in columns:
+                if not isinstance(requested, tuple):
+                    raise ValueError(
+                        "columns parameter for multi-index data frame must be either a dictionary or list of tuples."  # noqa
```
timj: What is this bare noqa doing?

timothydmorton: I thought the flake8 was failing because of a long line here, and thought this might fix it?

timj: If the line is too long, split the string in two.

```python
raise ValueError("columns parameter for multi-index data frame "
                 "must be either a dictionary or list of tuples.")
```

timothydmorton: Is this an OK thing to do?  Is there a better way to do this?

timj: Splitting strings like this is extremely common. It's used all over the place.

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -89,13 +90,15 @@ def _splitColumnnNames(n: int, names: Iterable[str]) -> Iterator[Tuple[str, ...]
...
+        pattern = re.compile(r"\({}\)".format(", ".join(["'(.*)'"] * n)))
         for name in names:
             m = re.search(pattern, name)
             if m is not None:
                 yield m.groups()
 
-    def _standardizeColumnParameter(self, columns: Dict[str, Union[str, List[str]]]) -> Iterator[str]:
+    def _standardizeColumnParameter(
+        self, columns: Union[List[tuple], Dict[str, Union[str, List[str]]]]
+    ) -> Iterator[str]:
```
sr525, line 100: I don't know if this passed PEP8 but I find it really hard to read and understand.

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -111,21 +114,36 @@ def _standardizeColumnParameter(self, columns: Dict[str, Union[str, List[str]]])
...
-
-    def read(self, columns: Union[str, List[str], Dict[str, Union[str, List[str]]]] = None
-             ) -> pd.DataFrame:
+        if isinstance(columns, list):
+            for requested in columns:
+                if not isinstance(requested, tuple):
+                    raise ValueError(
+                        "columns parameter for multi-index data frame must be either a dictionary or list of tuples."
+                    )
+                yield str(requested)
```
sr525, line 121: It seems odd to have made sure it is a list of tuples to immediately force it to be a string. Why not make it a string initially?

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -111,21 +114,36 @@ def _standardizeColumnParameter(self, columns: Dict[str, Union[str, List[str]]])
...
+                raise ValueError(
+                    "columns parameter for multi-index data frame must be either a dictionary or list of tuples."
+                )
+            if not set(self.indexLevelNames).issuperset(columns.keys()):
+                raise ValueError(
+                    f"Cannot use dict with keys {set(columns.keys())} "
+                    f"to select columns from {self.indexLevelNames}."
+                )
+            factors = [
+                iterable(columns.get(level, self.columns.levels[i]))
```
sr525: Is self.columns always a MultiIndex? From earlier in the code it looks like it can sometimes be an Index which seems not to have a .levels option. I'm not really familiar with pandas and could just be wrong.

I also find it quite confusing to have columns and self.columns which are different things.

python/lsst/daf/butler/formatters/parquet.py
timothydmorton: This is just adding an option for the `columns` keyword to be passed a list of tuples instead of just a dictionary of a particular format, which was the original implementation.

python/lsst/daf/butler/formatters/parquet.py
timothydmorton: The reason for all this gymnastics about the column names is because of funniness about how pyarrow deals with the column names of a table with a multi-level column index.  

In pandas alone, if you have a multilevel column index, you can index it with a straight-up tuple; e.g. `df[('meas', 'HSC-G', 'coord_ra')]` to grab that specific column.  To read that same column from a parquet table via pyarrow, however, you would have to pass `columns=["('meas', 'HSC-G', 'coord_ra')"]` to the parquet loader.  This is tiresome to have to remember to do in the functor front-end, and there are situations where requesting columns via the "pass a dictionary of column specifications which gets converted internally to the list of columns needed from pyarrow" doesn't work, so we need to pass the list of tuples (e.g., for composite functors, color functors, and maybe more).  So rather than needing to string-ify them within the `Functor` implementation or call, we do the string-ifying in the butler parquet loader.  This way also, the same `Functor` code could in principle be used directly on a dataframe, not necessarily on the dataref that will load from parquet.

python/lsst/daf/butler/formatters/parquet.py
timothydmorton: This whole "else" block I didn't actually change anything from what was previously implemented; I just moved it into an "else" block to be the code.  All I did in this function was to add the `if isinstance(columns, list):` block to handle the case where passing a list of tuples.

PR #429, opened 2020-11-07
DM-27374: Add ingest time support to queryDatasets

andy-slac:
Visitor now recognizes `ingest_date` identifier and generates dialect-specific code to convert the value to nanoseconds since epoch so that time can be compared to other timestamps.

### DIFF
--- a/doc/lsst.daf.butler/queries.rst
+++ b/doc/lsst.daf.butler/queries.rst
@@ -148,7 +148,9 @@ dimensions, e.g. ``visit`` identifier is used to represent a value of
 ``visit`` dimension in registry database. Dotted identifiers are mapped to
 tables and columns in registry database, e.g. ``detector.raft`` can be used
 for accessing raft name (obviously dotted names need knowledge of database
-schema and how SQL query is built).
+schema and how SQL query is built). A simple identifier with a name
+``ingest_date`` is used to reference dataset ingest time, which can be used to
+filter query results based on that property of datasets.
 
 Unary arithmetic operators
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -373,3 +375,5 @@ Few examples of valid expressions using some of the constructs:
     exposure.datetime_begin > T'58938.515'
 
     visit.datetime_end < T'mjd/58938.515/tai'
+
+    ingest_date < T'2020-11-06 21:10:00'
--- a/python/lsst/daf/butler/registry/queries/expressions.py
+++ b/python/lsst/daf/butler/registry/queries/expressions.py
@@ -22,11 +22,13 @@ from __future__ import annotations
 
 __all__ = ()  # all symbols intentionally private; for internal package use.
 
-from typing import Any, List, Optional, Tuple, TYPE_CHECKING, Union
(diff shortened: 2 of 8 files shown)


python/lsst/daf/butler/registry/queries/expressions.py
```diff
@@ -22,19 +22,78 @@
...
+    name : `str`
+        Identifier to categorize.
+
+    Returns
+    -------
+    isIngestDate : `bool`
+        True is returned if identifier name is ``ingest_date``.
+    """
+    # TODO: this is hardcoded for now, may be better to extract it from schema
+    # but I do not know how to do it yet.
```
TallJimbo, line 92: I wouldn't worry about this too much; I generally haven't so far - hard-coded strings aren't great, but compared to the opaque levels of indirection involved in avoiding them when referring to column names, I think they're not really that bad bad.

tests/test_butler.py
```diff
@@ -892,6 +892,25 @@ def testStringification(self):
...
+        datasets = list(butler.registry.queryDatasets(..., collections=..., where=where))
+        len1 = len(datasets)
+        self.assertEqual(len0, len1)
+
+        # no one will ever use this piece of software in 30 years
+        where = "ingest_date > T'2050-01-01'"
+        datasets = list(butler.registry.queryDatasets(..., collections=..., where=where))
+        len2 = len(datasets)
+        self.assertEqual(len2, 0)
+
```
TallJimbo: Would you mind moving this to `daf/butler/registry/tests/_registry.py`?  You can get all of the set up you need with:
```
registry = self.makeRegistry()
self.loadData(registry, "base.yaml")
self.loadData(registry, "datasets.yaml")
```
That will give us coverage in PostgreSQL, and I've found relying on the outputs of `runPutGetTest` to be a problem in the past because the context managers within it can make it hard to see what actually failed because they swallow exceptions.

doc/lsst.daf.butler/queries.rst
```diff
@@ -148,7 +148,9 @@ dimensions, e.g. ``visit`` identifier is used to represent a value of
 ``visit`` dimension in registry database. Dotted identifiers are mapped to
 tables and columns in registry database, e.g. ``detector.raft`` can be used
 for accessing raft name (obviously dotted names need knowledge of database
-schema and how SQL query is built).
+schema and how SQL query is built). A simple identifier with a name
+``ingest_date`` is used to reference dataset ingest time, it can be used to
+filter query results based in that property of datasets.
```
TallJimbo: "time, it can" -> "time, which can"

"base in that" -> "based on that"

TallJimbo (approved):

tests/test_butler.py
timj: This happens after runPutGetTest so there should be no issue with swallowing of exceptions.

There is a broader problem that I need to completely refactor test_butler.py so that it moves mostly into the tests module and is clearer to how it tests backends. It mostly is designed to switch datastores but it does of course need to have a testing.postgresql version as well.

andy-slac: Thanks for the hint, I'll move it to `_registry.py`. The test grew large, it's hard to navigate their classes and contents to find the right spot to stick something in 🙂

PR #430, opened 2020-11-10
DM-27490: Add support for collection docstrings.

TallJimbo:


python/lsst/daf/butler/registry/_registry.py
```diff
@@ -554,6 +559,33 @@ def setCollectionChain(self, parent: str, children: Any) -> None:
...
+
+    def setCollectionDocumentation(self, collection: str, doc: Optional[str]) -> None:
+        """Set the documentation string for a collection.
+
+        Parameters
+        ----------
+        name : `str`
+            Name of the collection.
+        docs : `str`, optional
+            Docstring for the collection with the given name.
```
timj: If you set doc to None here does it remove the existing docstring?  Maybe state that this value overwrites a previous value?

timj (approved):
looks great. One question.

python/lsst/daf/butler/registry/_registry.py
TallJimbo: It should remove the existing docstring.  I'll update the docstring and add a test to make sure it's not lying.

PR #431, opened 2020-11-10
DM-27154: Usability improvement suggestions for butler collection commands

n8pease:


python/lsst/daf/butler/tests/utils.py
```diff
@@ -55,3 +67,184 @@ def assertAstropyTablesEqual(self, tables, expectedTables):
...
+        """Join the given path to the butler root.
+
+        Parameters
+        ----------
+        path : `str`
+            The relative path to append to the butler root.
+
+        Returns
+        -------
+        `str`
```
timj: I'm expecting this to return a ButlerURI.

Also this is the wrong doc syntax for a return value. I'm not completely convinced that this deserves a method of its own.

python/lsst/daf/butler/tests/utils.py
```diff
@@ -55,3 +67,184 @@ def assertAstropyTablesEqual(self, tables, expectedTables):
...
+            datasetType = self.datasetType
+        ref = DatasetRef(datasetType, dataId, id=None)
+        metric = self._makeExampleMetrics()
+        if run:
+            butler = Butler(self.root, run=run)
+        else:
+            butler = self.butler
+        butler.put(metric, ref)
+
+    def addDatasetType(self, dimensions, datasetTypeName, storageClass):
```
timj: I don't understand why this is here. It's identical to `lsst.daf.butler.tests.addDatasetType` isn't it?

python/lsst/daf/butler/tests/utils.py
```diff
@@ -55,3 +67,184 @@ def assertAstropyTablesEqual(self, tables, expectedTables):
...
+        """
+        if not datasetType:
+            datasetType = self.datasetType
+        ref = DatasetRef(datasetType, dataId, id=None)
+        metric = self._makeExampleMetrics()
+        if run:
+            butler = Butler(self.root, run=run)
+        else:
+            butler = self.butler
+        butler.put(metric, ref)
```
timj: Probably should return the DatasetRef returned by this `put`.

python/lsst/daf/butler/tests/utils.py
```diff
@@ -55,3 +67,184 @@ def assertAstropyTablesEqual(self, tables, expectedTables):
...
+            dataset will be added to the root butler.
+        datasetType : ``DatasetType``, optional
+            The dataset type of the added dataset. If `None`, will use the
+            default dataset type.
+        """
+        if not datasetType:
+            datasetType = self.datasetType
+        ref = DatasetRef(datasetType, dataId, id=None)
+        metric = self._makeExampleMetrics()
+        if run:
```
timj, line 191: This isn't needed. You can say `self.butler.put(metric, ref, run=run)` below.

python/lsst/daf/butler/tests/utils.py
```diff
@@ -55,3 +67,184 @@ def assertAstropyTablesEqual(self, tables, expectedTables):
...
+        run : `str`, optional
+            The name of the run to create and add a dataset to. If `None`, the
+            dataset will be added to the root butler.
+        datasetType : ``DatasetType``, optional
+            The dataset type of the added dataset. If `None`, will use the
+            default dataset type.
+        """
+        if not datasetType:
+            datasetType = self.datasetType
+        ref = DatasetRef(datasetType, dataId, id=None)
```
timj: This is not necessary. You do not need to create a ref for the put.

python/lsst/daf/butler/tests/utils.py
timj: I think this whole routine can be something like:
```python
def addDataset(...):
    return self.butler.put(metric, datasetType if datasetType is not None else self.datasetType,
                           dataId=dataId, run=run)
```

python/lsst/daf/butler/tests/utils.py
```diff
@@ -55,3 +67,184 @@ def assertAstropyTablesEqual(self, tables, expectedTables):
...
+        self.addDimensionData("visit", dict(instrument="DummyCamComp",
+                                            id=424,
+                                            name="fourtwentyfour",
+                                            physical_filter="d-r",
+                                            visit_system=1))
+
+        self.addDataset({"instrument": "DummyCamComp", "visit": 423})
+        self.addDataset({"instrument": "DummyCamComp", "visit": 424})
+
+    def addDataset(self, dataId=None, run=None, datasetType=None):
```
timj: I don't think you want to allow dataId to be optional do you?

python/lsst/daf/butler/tests/utils.py
```diff
@@ -55,3 +67,184 @@ def assertAstropyTablesEqual(self, tables, expectedTables):
...
+        datasetType = DatasetType(datasetTypeName, dimensions, storageClass)
+        self.butler.registry.registerDatasetType(datasetType)
+        return datasetType
+
+    def addDimensionData(self, element, data):
+        """Insert dimension data for an element into the repository.
+
+        Parameters are the same as Registry.insertDimensionData.
+        """
+        self.butler.registry.insertDimensionData(element, data)
```
timj: I don't think this API gives us much other than obfuscate a standard API. Above I suggest a way to make it more useful.  It looks like you could move it to _testRepo.py as a helper and presumably adjust makeTestRepo to use it as well.

python/lsst/daf/butler/tests/utils.py
```diff
@@ -55,3 +67,184 @@ def assertAstropyTablesEqual(self, tables, expectedTables):
...
+
+    def join(self, path):
```
timj: I'm not completely sold that we want to hide this one line behind a level of abstraction. It's such a generic name and it doesn't describe what is being joined to what since it's a method on the test class and you aren't joining a path to the test class.  A descriptive name runs into the problem of being longer and meaning you aren't saving much.

python/lsst/daf/butler/tests/utils.py
```diff
@@ -55,3 +67,184 @@ def assertAstropyTablesEqual(self, tables, expectedTables):
...
+        self.addDimensionData("instrument", {"name": "DummyCamComp"})
+        self.addDimensionData("physical_filter", {"instrument": "DummyCamComp",
+                                                  "name": "d-r",
+                                                  "band": "R"})
+        self.addDimensionData("visit_system", {"instrument": "DummyCamComp",
+                                               "id": 1,
+                                               "name": "default"})
+        visitStart = astropy.time.Time("2020-01-01 08:00:00.123456789", scale="tai")
+        visitEnd = astropy.time.Time("2020-01-01 08:00:36.66", scale="tai")
+        self.addDimensionData("visit", dict(instrument="DummyCamComp",
```
timj: I think to get some added value from a new API it would be better if this method wasn't just a rename of `insertDimensionData`. If it took kwargs with lists of dicts (or a single dict on its own meaning one item -- I think we have a utils function for that) then it would be worth doing:

```python
self.addDimensionData(visit_system=instrument_system, visit=[visit1, visit2])
```
and `addDimensionData` would iterate through the keyword arguments and for each one iterate through the arguments.

tests/test_cliCmdPruneCollection.py
```diff
@@ -45,7 +54,7 @@ def testPruneCollections(self):
             # Add the run and the tagged collection to the repo:
             result = self.runner.invoke(butlerCli, ["create", repoName])
             self.assertEqual(result.exit_code, 0, clickResultMsg(result))
-            lsst.daf.butler.Butler(repoName, run=runName, tags=[taggedName])
+            Butler(repoName, run=runName, tags=[taggedName])
```
timj, line 59: Maybe add a comment noting that this is creating the collections behind the scenes but that we don't need the butler object itself.

tests/test_cliCmdQueryCollections.py
```diff
@@ -56,41 +75,159 @@ def test_all(self):
...
     def testGetCollections(self):
         run = "ingest/run"
         tag = "tag"
-        runner = LogCliRunner()
-        with runner.isolated_filesystem():
+        with self.runner.isolated_filesystem():
             butlerCfg = Butler.makeRepo("here")
             # the purpose of this call is to create some collections
-            _ = Butler(butlerCfg, run=run, tags=[tag], collections=[tag])
+            _ = Butler(butlerCfg, run=run, tags=[tag], collections=[tag], writeable=True)
```
timj: Don't need the `_ =`

tests/test_cliCmdQueryCollections.py
```diff
@@ -56,41 +75,159 @@ def test_all(self):
...
+        """
+        return (ref.id, parameters)
+
+    def setUp(self):
+        self.runner = LogCliRunner()
+
+    def testChained(self):
+        with self.runner.isolated_filesystem():
+
+            # Create a butler and add some chained collections
```
timj: ...and replace the datastore with a mock

tests/test_cliCmdQueryCollections.py
```diff
@@ -56,41 +75,159 @@ def test_all(self):
...
+            butlerCfg = Butler.makeRepo("here")
+            with unittest.mock.patch.object(Datastore, "fromConfig", spec=Datastore.fromConfig):
```
timj: I find the `with` a bit confusing here since it implies that something is freed or reversed when the block exits. This seems to be how it works though.

tests/test_cliCmdQueryCollections.py
```diff
@@ -56,41 +75,159 @@ def test_all(self):
...
+            datastore.export = _mock_export
+
+        """
+        for ref in refs:
+            yield FileDataset(refs=[ref],
+                              path="mock/path",
+                              formatter="lsst.daf.butler.formatters.json.JsonFormatter")
+
+    @staticmethod
+    def _mock_get(ref: DatasetRef, parameters: Optional[Mapping[str, Any]] = None
```
timj: Rather than copying this from test_simpleButler.py, please put the mocks in a DatastoreMock base class in the tests hierarchy.

python/lsst/daf/butler/_butler.py
```diff
@@ -1068,19 +1092,24 @@ def pruneCollection(self, name: str, purge: bool = False, unstore: bool = False)
...
         # the cases are almost the same, but here we can rely on Registry to
         # take care everything but Datastore deletion when we remove the
         # collection.
         if not self.isWriteable():
             raise TypeError("Butler is read-only.")
-        if purge and not unstore:
-            raise TypeError("Cannot pass purge=True without unstore=True.")
         collectionType = self.registry.getCollectionType(name)
+        if purge and not unstore:
+            raise PruneCollectionsArgsError(PruneCollectionsArgsError.Reason.PURGE_WITHOUT_UNSTORE,
```
timj: I'm fine with a new subclass for specificity but couldn't we also make three different exceptions? You could have a new base class of PruneCollectionsArgsError and then have three subclasses of that such as PurgeWithoutUnstoreCollectionsError etc. They could have the error message burned in and you wouldn't need the enum here and you wouldn't need to check the reason later on -- you could simply catch the specific exception and replace it with a new error message suitable for click users.

timj (approved):
Mainly comments on test infrastructure changes but also a comment on the new exception class.

python/lsst/daf/butler/tests/utils.py
n8pease: It fails when I call this and the run doesn't exist yet.

timj: It would be clearer then if you had a line here that created the run collection rather than without commentary relying on Butler to create it behind the scenes.

n8pease: maybe it's weird if/because I'm thinking about it wrong though...

timj: Without a comment here it all seems superfluous. Whereas:

```python
if run:
```
is explicit about what you need. @TallJimbo do you prefer the "create a new butler for each run" paradigm? Maybe that's better for the general user. If that is the case I'd like a comment to explain why you are having to create a new butler.

n8pease: I see. That makes sense, works.

TallJimbo: Outside of tests, users are never going to create new runs or even call `put` themselves (it'll all be operations like ingest or `pipetask` doing that).  So I think we can go with whatever maximizes test code readability.  I don't have a super strong preference; the "new butler for each run" pattern is maybe a bit more familiar to those coming from Gen2, but I agree that it's less explicit.

tests/test_cliCmdQueryCollections.py
n8pease: I didn't totally understand it (it's cargo culted from something @TallJimbo wrote), so I did more research. The deal is that `Datastore.fromConfig` gets patched: replaced with a function that returns a `magicmock`. The butler initializer calls `Datastore.fromConfig`, which is a `MagicMock` instance and calling it returns a new `MagicMock` instance, so now the butler's `self.datastore` is a magicmock. Init finishes, and `export` and `get` are monkey patched onto the `butler.datastore` magic mock. Then the `with` block exits, which *does* restore `Datastore.fromConfig` to its normal function, but `butler.datastore` is of course not changed because how would it and anyway we don't want it to be - it's why we did all the above anyway.

n8pease: However, all that seems kind of complicated and it seems like it's not necessary to patch `Datastore.fromConfig` - the butler init works just fine with it *not* patched. We do run into problems when we call `butler1.import_(filename=os.path.join(TESTDIR, "data", "registry", "base.yaml"))` without patching, but before making that call we can simply replace the `datastore` functions that need replacing, without any context manager. Doing this seems to be enough:

            butler1.datastore.export = self._mock_export
            butler1.datastore.get = self._mock_get
            butler1.datastore.ingest = MagicMock()

(note that instead of `datastore` be a MagicMock with `export` and `get` monkey patched on, I've replaced those two with our alternative impls, and made `ingest` a mock because it will get called in the course of calling `butler.ingest_`, but does not seem to need to do anything.)

n8pease: @TallJimbo if there's a reason to use the patch context manager, please let me know? In the meantime I'll change it to the simpler version I just described and you and @timj can 👍 or 👎 that impl if y'all want.

n8pease: (hmm, it seems to need the context manager in test_simpleButler. working on understanding why...)

n8pease: aha. The butlers were getting created slightly differently, the one that worked was being init'd with the config returned by `Butler.makeRepo`, and the one that did not was building a config itself, and needed `config["root"]` to be set.

TallJimbo: This was the first time I ever really used `unittest.mock` (other than small changes to tests others had written), so if you find a simpler way to accomplish what my code did, assume I just wasn't aware of that alternative.

PR #439, opened 2020-11-19
DM-27118: Add support for recreating logging configuration

andy-slac:
New pipetask option for multiprocessing spawn method needs support for
transporting global state from parent to child processes. Logging
configuration is a part of this global state, this commit adds support
for recording and replaying logging configuration changes.

### DIFF
--- a/python/lsst/daf/butler/cli/cliLog.py
+++ b/python/lsst/daf/butler/cli/cliLog.py
@@ -105,7 +105,7 @@ class CliLog:
             # calling this function multiple times in one program execution,
             # so do log a debug but don't log an error or fail, just make the
             # re-initialization a no-op.
-            log = logging.getLogger(__name__.partition(".")[2])
+            log = logging.getLogger(__name__)
             log.debug("Log is already initialized, returning without re-initializing.")
             return
         cls._initialized = True
@@ -297,7 +297,7 @@ class CliLog:
         """
         if cls._initialized or cls.configState:
             # Already initialized, do not touch anything.
-            log = logging.getLogger(__name__.partition(".")[2])
+            log = logging.getLogger(__name__)
             log.warning("Log is already initialized, will not replay configuration.")
             return
 
--- a/python/lsst/daf/butler/cli/cliLog.py
+++ b/python/lsst/daf/butler/cli/cliLog.py
@@ -78,6 +78,11 @@ class CliLog:
     """The log format used when the lsst.log package is not importable and the
     log is initialized with longlog=False."""
 
+    configState = []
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/cli/cliLog.py
```diff
@@ -273,3 +284,24 @@ def _recordComponentSetting(cls, component):
...
+        if cls._initialized or cls.configState:
+            # Already initialized, do not touch anything.
+            log = logging.getLogger(__name__.partition(".")[2])
+            log.warning("Log is already initialized, will not replay configuration.")
+            return
+
+        # execute each one in order
+        for call in configState:
+            method, *args = call
+            method(*args)
```
n8pease, line 307: why not use `functools.partial`? (binding arg vals to funcs is what it's made for)

andy-slac: I thought about that but decided that I want to avoid extra imports and complexity. There are certainly cases when `partial` is useful, but in this case it's easy to do things without it.

python/lsst/daf/butler/cli/cliLog.py
```diff
@@ -273,3 +284,24 @@ def _recordComponentSetting(cls, component):
...
+
+        Parameters
+        ----------
+        configState : `list` of `tuple`
+            Tuples contain a method as first item and arguments for the method,
+            in the same format as ``cls.configState``.
+        """
+        if cls._initialized or cls.configState:
+            # Already initialized, do not touch anything.
+            log = logging.getLogger(__name__.partition(".")[2])
```
timj: In all other butler code we use `__name__` and don't strip the `lsst.` (I see it's actually two places because something else crept in as well)

andy-slac: I'm trying to follow recommendations that say `lsst` should be dropped: https://developer.lsst.io/stack/logging.html#logger-names. I can certainly change that to just `__name__` if this is preferred.

timj: butler is special and has been for a long time -- if you drop it people will have to use lsst.daf.butler and daf.butler to turn on debug logging .

andy-slac: OK, I'll replace that and other cases if I find any.

timj: The only other one was in core/utils.py I think

andy-slac: Replaced 3 instances in a new commit.

timj (approved):

PR #441, opened 2020-11-23
DM-27689: extend static type checking and fix errors

TallJimbo:


python/lsst/daf/butler/_butler.py
```diff
@@ -105,14 +107,14 @@ class PurgeWithoutUnstorePruneCollectionsError(PruneCollectionsArgsError):
     purge is True but unstore is False.
     """
 
-    def __init__(self):
+    def __init__(self) -> None:
```
timj, line 112: How come this return type is needed is it?

python/lsst/daf/butler/_butler.py
```diff
@@ -674,26 +679,26 @@ def _findDatasetRef(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
...
 
             # Look up table for the first association with a dimension
-            guessedAssociation: dict[Any, dict[str, Any]] = defaultdict(dict)
+            guessedAssociation: Dict[Any, Dict[str, Any]] = defaultdict(dict)
 
             # Keep track of whether an item is associated with multiple
             # dimensions.
-            counter = Counter()
-            assigned: dict[Any, Set[str]] = defaultdict(set)
+            counter: Any = Counter()
```
timj: I'm not sure what this type is doing. We know it's not `Any` -- it should be `Counter` shouldn't it?

python/lsst/daf/butler/_butler.py
```diff
@@ -778,7 +783,11 @@ def quote(s):
                                        f" records when constrained by {values}")
 
                 # Get the primary key from the real dimension object
-                dimension = self.registry.dimensions[dimensionName]
+                dimension = self.registry.dimensions.getStaticDimensions()[dimensionName]
+                if not isinstance(dimension, Dimension):
```
timj, line 791: How does this ever trigger given that the lookup above would KeyError if it wasn't. Is this just a mypy hint?

python/lsst/daf/butler/_butler.py
```diff
@@ -1367,6 +1376,7 @@ def pruneDatasets(self, refs: Iterable[DatasetRef], *,
             if purge:
                 self.registry.removeDatasets(refs)
             elif disassociate:
+                assert tags
```
timj: Can I have an error text as well please?

python/lsst/daf/butler/_butler.py
```diff
@@ -1613,14 +1623,14 @@ def import_(self, *, directory: Optional[str] = None,
             if filename is None:
                 raise TypeError("At least one of 'filename' or 'format' must be provided.")
             else:
-                _, format = os.path.splitext(filename)
+                _, format = os.path.splitext(filename)  # type: ignore
```
timj, line 1630: Now I want `filename` to be a `ButlerURI`

python/lsst/daf/butler/core/_butlerUri/http.py
```diff
@@ -329,6 +329,8 @@ def session(self) -> requests.Session:
         if isWebdavEndpoint(baseURL):
             log.debug("%s looks like a Webdav endpoint.", baseURL)
             s = getHttpSession()
+        else:
```
timj, line 332: @bgounon I'm not entirely sure what you wanted to happen in this else -- can you confirm this is okay?

python/lsst/daf/butler/core/composites.py
```diff
@@ -137,6 +137,8 @@ def shouldBeDisassembled(self, entity: Union[DatasetRef, DatasetType, StorageCla
                 matchName = key
                 break
 
+        assert isinstance(disassemble, bool)
```
timj: Error message? Would this be needed if we declared the type at line 132 above?

python/lsst/daf/butler/core/config.py
```diff
@@ -873,7 +875,9 @@ def dumpToUri(self, uri: Union[ButlerURI, str], updateFile: bool = True,
         ext = uri.getExtension()
         format = ext[1:].lower()
 
-        uri.write(self.dump(format=format).encode(), overwrite=overwrite)
+        output = self.dump(format=format)
+        assert output is not None
```
timj: Error message? 😄

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -178,8 +180,7 @@ def read(self, component: Optional[str] = None) -> Any:
 
         return loader.read(**self.fileDescriptor.parameters)
 
-    def write(self, inMemoryDataset: Any) -> str:
+    def write(self, inMemoryDataset: Any) -> None:
```
timj, line 183: Sorry about that one. I thought I had caught them all.

timj (approved):
Looks great. Standard comment about error messages on asserts 😄

python/lsst/daf/butler/_butler.py
```diff
@@ -600,8 +605,8 @@ def _findDatasetRef(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
 
         # Process dimension records that are using record information
         # rather than ids
-        newDataId: dict[Any, Any] = {}
-        byRecord: dict[Any, dict[str, Any]] = defaultdict(dict)
+        newDataId: Dict[Any, Any] = {}
```
timj: Oh, I think you mentioned something about proper types for data Id keys and values?

python/lsst/daf/butler/_butler.py
TallJimbo: Yeah, MyPy special-cases `__init__` with type-annotated arguments to not need a return type annotation, but it needs the annotation here to be able to tell whether it should treat this as a typed method it should look at more closely or one it should ignore.  And in strict mode it demands that everything be a typed method.

python/lsst/daf/butler/_butler.py
TallJimbo: I don't know why I assumed `typing.Counter` didn't exist, but of course it does (first I tried `Dict`, but that failed because `most_common` isn't a member, and then I got lazy).  Fixed.

python/lsst/daf/butler/_butler.py
TallJimbo: Yes, just a mypy hint - indexing `DimensionUniverse` can return `Dimension` or `DimensionElement`, so it was this or an `assert isinstance` check.

python/lsst/daf/butler/_butler.py
TallJimbo: I'll let you take care of that one :slightly_smiling_face:

python/lsst/daf/butler/core/composites.py
TallJimbo: I should probably just upgrade this to a real exception because it's essentially external config validation.  Of course, it's only the tip of that iceberg, but I figure every little bit helps.

python/lsst/daf/butler/formatters/parquet.py
TallJimbo: This one wasn't being type-checked before, so you're absolved :slightly_smiling_face:

python/lsst/daf/butler/core/_butlerUri/http.py
timj: @TallJimbo I'm fine with this merging as is -- it's clearly fixing a problem and if @bgounon comes back with a different fix that can be a different ticket.

python/lsst/daf/butler/_butler.py
TallJimbo: This was a bit trickier than I expected, due to the messiness of `Mapping[Union[str, Dimension], ...]` vs. `Union[Mapping[str, ...], Mapping[Dimension, ...]]`.  Rather than fight that I just made these dicts use `str` keys and coerced the one place where they might have been given `Dimension` instances.

PR #442, opened 2020-11-24
DM-27344: Add butler query-dimension-records subcommand

n8pease:


python/lsst/daf/butler/cli/cmd/commands.py
```diff
@@ -269,3 +268,24 @@ def query_data_ids(**kwargs):
...
+@click.command(cls=ButlerCommand)
+@repo_argument(required=True)
+@element_argument(required=True)
+@datasets_option(help=unwrap("""An expression that fully or partially identifies dataset types that should
+                             constrain the yielded records. Only affects results when used with
+                             --collections."""))
+@collections_option(help=collections_option.help + " Only affects results when used with --datasets.")
+@where_option(help=whereHelp)
+@click.option("--no-check", is_flag=True,
+              help=unwrap("""Don't check the query before execution. By default the query is checked before it
```
timj, line 282: @TallJimbo what do we gain by having this option? When is disabling the check a good idea?

python/lsst/daf/butler/tests/utils.py
```diff
@@ -87,6 +87,7 @@ def readTable(textTable):
     """
     return AstropyTable.read(textTable,
                              format="ascii",
+                             data_start=2,  # skip the header row and the header row underlines.
```
timj, line 90: I'm a bit surprised that Astropy doesn't have a parser for the default pretty print table output. Part of me is wondering whether we should add an output format option to all the tabular output subcommands so that users can easily parse the output if they want. That would also let you specify csv format in tests for example.

timj (approved):

python/lsst/daf/butler/tests/utils.py
n8pease: Formatter options are not a bad idea. Another format might be an option to choose between Table and yaml, if there's ever going to be a case where a user may pipe the output of one command to the input of another? How to do naming/data structure with yaml would probably take a little thought though, or maybe you've got good ideas already.

In this case, `AstropyTable.read` normally does the Right Thing, but I ran into a situation where it was getting confused. The table had one column and one row, something like
```
name
----
foo
```
and it was reading the `----` as a value row. I spent a little time trying to figure out why but didn't really get anywhere. It seemed ok to have the unit test function start reading at idx 2 (3rd row, eh) since we're consistent with table output format (until we decide not to be... then we have to come up with a better or more elaborate fix)

python/lsst/daf/butler/cli/cmd/commands.py
TallJimbo: The check can generate false positives for some valid-but-rare queries.  Those all involve _wanting_ multiple instruments _and_ a WHERE constraint on a table derived from instrument (e.g. `exposure`) on a column in that table whose values are nevertheless instrument-agnostic (e.g. `exposure.exposure_time > 30`).  They then fall into two categories:
 - Cases where you do restrict the instruments (or skymaps) you want, but use `IN` instead of `=` and `OR`: `instrument IN ('HSC', 'DECam') AND exposure.exposure_time > 30`.  The checker just isn't smart enough to rewrite the `IN`.
 - Cases where you actually want to query all of the instruments (or skymaps) in the registry: just `exposure.exposure_time > 30`.  Note that I think we still want to check this case by default, because usually the user is thinking of a specific instrument but we don't know that.

Similar cases may exist for skymaps, but it's harder to think of practically useful cases given the attributes actually on the tract and patch tables.

I'm open to opinions on whether that's sufficiently niche as to make this option not worth its maintenance weight.

PR #446, opened 2020-11-26
DM-27169: Use FilterLabel in Exposure/ExposureInfo

kfindeisen:
This PR adds support for `FilterLabel` as a persisted form for filter information in decomposed exposures. Must be merged with lsst/afw#551. `Filter` remains both for backward-compatibility, and for consistency with the `Exposure` API (which temporarily also has `getFilter()` and `getFilterLabel()`). 

The plan for the `Exposure` class is to migrate code to using the `FilterLabel` class (through `getFilterLabel()`) on DM-27170, remove `Filter` and `getFilter()` and replace them with a new `getFilter()` that returns a `FilterLabel` on DM-27177, then remove `getFilterLabel()` on DM-27811. I propose that the components follow the same pattern to minimize user confusion.

### DIFF
--- a/python/lsst/daf/butler/configs/datastores/formatters.yaml
+++ b/python/lsst/daf/butler/configs/datastores/formatters.yaml
@@ -13,6 +13,7 @@ CoaddInputs: lsst.obs.base.formatters.fitsGeneric.FitsGenericFormatter
 VisitInfo: lsst.obs.base.formatters.fitsGeneric.FitsGenericFormatter
 ApCorr: lsst.obs.base.formatters.fitsGeneric.FitsGenericFormatter
 PhotoCalib: lsst.obs.base.formatters.fitsGeneric.FitsGenericFormatter
+FilterLabel: lsst.obs.base.formatters.fitsGeneric.FitsGenericFormatter
 TransmissionCurve: lsst.obs.base.formatters.fitsGeneric.FitsGenericFormatter
 Camera: lsst.obs.base.formatters.fitsGeneric.FitsGenericFormatter
 Detector: lsst.obs.base.formatters.fitsGeneric.FitsGenericFormatter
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/configs/datastores/formatters.yaml
```diff
@@ -13,6 +13,7 @@ CoaddInputs: lsst.obs.base.formatters.fitsGeneric.FitsGenericFormatter
 VisitInfo: lsst.obs.base.formatters.fitsGeneric.FitsGenericFormatter
 ApCorr: lsst.obs.base.formatters.fitsGeneric.FitsGenericFormatter
 PhotoCalib: lsst.obs.base.formatters.fitsGeneric.FitsGenericFormatter
+FilterLabel: lsst.obs.base.formatters.fitsGeneric.FitsGenericFormatter
```
timj, line 16: It seems to me that serializing in FITS is overkill. A FilterLabel is a couple of key/value pairs isn't it? YAML or JSON seems fine for that. We could fairly trivially make the FilterFormatter work for FilterLabel and Filter since Formatters know the pytype they are formatting.

timj (approved):
These changes are fine but I'd like to know the justification of writing FilterLabel as FITS.

python/lsst/daf/butler/configs/datastores/formatters.yaml
kfindeisen: Every component of `ExposureInfo`, now including `FilterLabel`, is held, managed, and persisted using the same framework; this makes the code much simpler and lets us avoid (or, at least, phase out) a lot of the special-case code of the old `ExposureInfo`. Given that `FilterLabel`/`ExposureInfo` already provides a FITS format in order to benefit from these advantages, why should we create a second format instead of reusing one that is supported out of the box? If the content of `FilterLabel` changes (which is quite possible, as the filter system evolves), should we need to remember to synchronize changes to two different representations?

In my view, the custom `FilterFormatter` was a necessary evil that standardization has made unnecessary.

PR #447, opened 2020-12-03
DM-27356: Add formatter for astropy table types

SimonKrughoff:


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -171,3 +171,7 @@ storageClasses:
     pytype: lsst.verify.Measurement
   Stamps:
     pytype: lsst.meas.algorithms.Stamps
+  AstroTable:
```
timj: I think I'd prefer AstropyTable since that's what it is.

SimonKrughoff: Oh, rats.  I missed that.  Incoming

python/lsst/daf/butler/formatters/astropyTable.py
```diff
@@ -0,0 +1,92 @@
...
+        Raises
+        ------
+        Exception
+            The file could not be written.
+        """
+        format_str = self.writeParameters.get('format', None)
+        if format_str:
+            inMemoryDataset.write(self.fileDescriptor.location.path, format=format_str)
+        else:
+            inMemoryDataset.write(self.fileDescriptor.location.path)
```
timj: is this else branch needed? I'm not entirely sure how `write()` works and the help docs don't really tell me, but if `format=None` is used is that the same as not specifying format? What does absence of format mean? Does it use the file extension? If it uses the file extension why do we need both options here?

timj (approved):
Looks good. One confusion from me over the writing.

python/lsst/daf/butler/formatters/astropyTable.py
SimonKrughoff: Good questions.  I guess this came from my wanting to be safe and I think this will work no matter how things work internal to `Table`

SimonKrughoff: I'd prefer to just leave it unless you strongly object

timj: I don't understand what the else is doing here. In the extension method we guarantee we will have a valid supported extension.  Then in the write method there is the same `.get()` but here the default is `None`. Why isn't the default the same as for `extension` above?  What does not specifying `format` actually mean? If there was a package variable with `DEFAULT_FORMAT = "ecsv"` and you used that in both extension and the `.get` here, wouldn't that remove the else completely?

SimonKrughoff: You're absolutely right.  I don't think we need the `format` option at all.  `astropy.Table` discovers the correct output format from the extension as we would expect.  I'm going to remove the else.

PR #448, opened 2020-12-07
DM-27835 pipetask run should always print stack trace and exception details

n8pease:


python/lsst/daf/butler/tests/cliCmdTestBase.py
```diff
@@ -50,9 +50,22 @@ def cli(self):
...
+        """Get the mock object to use in place of `mockFunc`. If not provided
+        will use the default provided by `unittest.patch`, this is usually a
+        `unittest.patch.MagicMock`."""
+        return DEFAULT
+
+    @property
+    @abc.abstractmethod
+    def mockFunc(self):
+        """The qualified name of the function to mock, will be passed to
+        unittest.mock.patch, see python docs for details. """
```
andy-slac, line 64: If it is a name and not the function may be better to call it `mockFuncName`?

tests/test_cliCmdConfigDump.py
```diff
@@ -37,6 +37,8 @@
 
 class ConfigDumpTest(CliCmdTestBase, unittest.TestCase):
 
+    mockFunc = "lsst.daf.butler.cli.cmd.commands.script.configDump"
```
andy-slac: Base class defines `mockFunc` as abstract property, and this is just an attribute. I guess it should work OK, but if you do not need it to be a property maybe define it as an attribute with `None` value in the base class too?

andy-slac (approved):

andy-slac:
BTW, the `build_and_test` checks are stuck, you need to rebase to master to fix that.

tests/test_cliCmdConfigDump.py
n8pease: Sadly, it must be a property to work with `@abstract`. But once this is done it satisfies the checker to make it an attribute in the base class. This is the way @timj prefers it, as I understand it.

python/lsst/daf/butler/tests/cliCmdTestBase.py
n8pease: grumbles; yes.

PR #449, opened 2020-12-08
DM-27685 Add butler make gen3 dcr subfilters subcommand

n8pease:
When the command funciton docstring spanned multiple lines the
argument help text was getting inserted into the middle of the
command function text.

### DIFF
--- a/python/lsst/daf/butler/cli/utils.py
+++ b/python/lsst/daf/butler/cli/utils.py
@@ -161,6 +161,11 @@ def addArgumentHelp(doc, helpText):
             doc = doc.split("\f")[0]
 
         doclines = doc.splitlines()
+        # The function's docstring may span multiple lines, so combine the
+        # docstring from all the first lines until a blank line is encountered.
+        # (Lines after the first blank line will be argument help.)
+        while len(doclines) > 1 and doclines[1]:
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/cli/utils.py
```diff
@@ -193,6 +193,14 @@ def addArgumentHelp(doc, helpText):
...
 
         doclines = doc.splitlines()
+        # The function's docstring may span multiple lines, so combine the
+        # docstring from all the first lines until a blank line is encountered.
+        # (Lines after the first blank line will be argument help.)
+        while True:
+            if len(doclines) > 1 and doclines[1]:
+                doclines[0] = " ".join((doclines[0], doclines.pop(1).strip()))
+            else:
+                break
```
TallJimbo: Looks like this could just be
```
while len(doclines) > 1 and doclines[1]:
    doclines[0] = " ".join((doclines[0], doclines.pop(1).strip()))
```

TallJimbo (approved):
Typo in commit message.

python/lsst/daf/butler/cli/utils.py
n8pease: nice one.

PR #450, opened 2020-12-11
DM-27985: Improvements to query system, especially temporal aspects

TallJimbo:


python/lsst/daf/butler/core/time_utils.py
```diff
@@ -108,6 +108,11 @@ def astropy_to_nsec(astropy_time: astropy.time.Time) -> int:
     return value
 
 
+# Two more constants: the integer nanoseconds equivalents to EPOCH and MAX_TIME
+MIN_NSEC = 0
+MAX_NSEC = astropy_to_nsec(MAX_TIME)
```
andy-slac: This is probably fine but in general executing complex code at import time may behave unpredictably. e.g. `logging` may not be configured yet.

python/lsst/daf/butler/registry/queries/_structs.py
```diff
@@ -470,3 +505,8 @@ class RegistryManagers:
     dimensions: DimensionRecordStorageManager
     """Manager for dimensions (`DimensionRecordStorageManager`).
     """
+
+    tsRepr: Type[TimespanDatabaseRepresentation]
```
andy-slac: `tsRepr` feels rather terse, I have to spend some metal energy to figure out that `ts` means `timespan`, maybe use full word for that (`timespanRepr`)?

python/lsst/daf/butler/core/timespan.py
```diff
@@ -41,36 +54,210 @@
...
+
+
+class _SpecialTimespanBound(enum.Enum):
+    """Enumeration to provide a singleton value for empty timespan bounds.
+
+    This enum's only member should generally be accessed via the
+    `Timespan.EMPTY` alias.
+    """
+
+    EMPTY = "Ø"
```
andy-slac: If this is meant to be printed then Unicode may potentially cause issues if terminal is not setup properly. Also grepping/searching may become more difficult.

python/lsst/daf/butler/core/timespan.py
```diff
@@ -41,36 +54,210 @@
...
+        creates an empty timespan.
+    padInstantaneous : `bool`, optional
+        If `True` (default) and ``begin == end`` *after discretization to
+        integer nanoseconds*, extend ``end`` by one nanosecond to yield a
+        finite-duration timespan.  If `False`, ``begin == end`` evaluates to
+        the empty timespan.
+    _nsec : `tuple` of `int`, optional
+        Integer nanosecond representation, for internal use by `Timespan` and
+        `TimespanDatabaseRepresentation` implementation only.  If provided,
+        all other arguments are are ignored.
```
andy-slac, line 100: Why not make separate constructor (static factory) which accepts `_nsec` only?

python/lsst/daf/butler/core/timespan.py
```diff
@@ -41,36 +54,210 @@
...
+    Timespans are half-open intervals, i.e. ``[begin, end)``.
+
+    Any timespan with ``begin > end`` after nanosecond discretization
+    (``begin >= end`` if ``padInstantaneous`` is `False`), or with either bound
+    set to `Timespan.EMPTY`, is transformed into the empty timespan, with both
+    bounds set to `Timespan.EMPTY`.  The empty timespan is equal to itself, and
+    contained by all other timespans (including itself).  It is also disjoint
+    with all timespans (including itself), and hence does not overlap any
+    timespan - this is the only case where ``contains`` does not imply
+    ``overlaps``.
```
andy-slac, line 122: I'm not sure I understand the utility of empty timespans but I probably missed discussion if there was any. I'll try to read the code and figure out how it is used though.

python/lsst/daf/butler/core/timespan.py
```diff
@@ -41,36 +54,210 @@
...
+                        warnings.simplefilter("ignore", category=astropy.utils.exceptions.AstropyWarning)
+                        if erfa is not None:
+                            warnings.simplefilter("ignore", category=erfa.ErfaWarning)
+                        if begin <= EPOCH:
+                            raise ValueError(f"Timespan.begin may not be earlier than {EPOCH}.")
+                        if end >= MAX_TIME:
+                            raise ValueError(f"Timespan.end may not be later than {MAX_TIME}.")
+                    raise ValueError("Infinite instantaneous timespans are not supported.")
+                elif padInstantaneous:
+                    end_nsec += 1
```
andy-slac, line 176: Will this make `[MAX_NSEC-1, +Inf)` timespan if both begin/end are very close to MAX_TIME?

python/lsst/daf/butler/core/timespan.py
```diff
@@ -41,36 +54,210 @@
...
+        time : `astropy.time.Time`
+            Time to use for the lower bound.
+
+        Returns
+        -------
+        instant : `Timespan`
+            A ``[time, time + 1ns)`` timespan.
+        """
+        nsec = astropy_to_nsec(time)
+        return Timespan(None, None, _nsec=(nsec, nsec))
```
andy-slac: I think this makes empty timespan, no?

python/lsst/daf/butler/core/timespan.py
```diff
@@ -80,66 +267,154 @@ def __str__(self) -> str:
...
+        # the second uses a strict inequality to make sure inf == inf isn't in
+        # play, and it's okay for the second to use a strict inequality only
+        # because we know non-empty Timespans have nonzero duration, and hence
+        # the second term is never false for non-empty timespans unless the
+        # first term is also false.
+        if isinstance(other, astropy.time.Time):
+            nsec = astropy_to_nsec(other)
+            return self._nsec[0] > nsec and self._nsec[1] > nsec
+        else:
+            return self._nsec[0] >= other._nsec[1] and self._nsec[1] > other._nsec[0]
```
andy-slac, line 373: I think that in a presence of empty timespans it becomes impossible to order things. If I read this right then for any non-empty timespan `ts` all of the below is true:

    empty < ts is False
    empty > ts is False
    ts < empty is False
    ts > empty is False

That probably means that `sort()` is not going to work correctly on any list that contains empty timespan.

python/lsst/daf/butler/core/timespan.py
```diff
@@ -376,10 +633,12 @@ class _CompoundTimespanDatabaseRepresentation(TimespanDatabaseRepresentation):
...
-    begin : `sqlalchemy.sql.ColumnElement`
-        SQLAlchemy object representing the begin (inclusive) endpoint.
-    end : `sqlalchemy.sql.ColumnElement`
-        SQLAlchemy object representing the end (exclusive) endpoint.
+    begin_nsec : `sqlalchemy.sql.ColumnElement`
+        SQLAlchemy object representing the lower (inclusive) bound, as a
+        64-bit integer column containing nanoseconds since ``time_utils.EPOCH`.
+    end_nsec : `sqlalchemy.sql.ColumnElement`
+        SQLAlchemy object representing the upper bound (exclusive), as a 64-bit
+        integer column containing nanoseconds since `time_utils.EPOCH`.
```
andy-slac: Constructor has a single `nsec` parameter which is a tuple, and also `name` which is not documented here.

python/lsst/daf/butler/core/timespan.py
```diff
@@ -390,116 +649,150 @@ class _CompoundTimespanDatabaseRepresentation(TimespanDatabaseRepresentation):
...
+                default=(None if nullable else sqlalchemy.sql.text(str(MAX_NSEC))),
                 **kwargs,
             ),
         )
 
     @classmethod
-    def getFieldNames(cls) -> Tuple[str, ...]:
+    def getFieldNames(cls, name: Optional[str] = None) -> Tuple[str, ...]:
         # Docstring inherited.
         return (f"{cls.NAME}_begin", f"{cls.NAME}_end")
```
andy-slac: Should it use `name` if name is not None?

andy-slac (commented):
I have to stop for today, here are the comments so far.

python/lsst/daf/butler/core/dimensions/_coordinate.py
```diff
@@ -621,6 +626,13 @@ def __init__(self, target: DataCoordinate):
 
     __slots__ = ("_target",)
 
+    def __repr__(self) -> str:
+        terms = [f"{d}: {self[d]!r}" for d in self._target.graph.elements.names]
+        return "{{{}}}".format(', '.join(terms))
+
+    def __str__(self) -> str:
+        return "\n".join(str(v) for v in self.values())
```
andy-slac, line 634: Is `\n` intentional here. Newlines don't usually work too well with `logging`.

python/lsst/daf/butler/core/dimensions/_coordinate.py
```diff
@@ -300,9 +300,10 @@ def __repr__(self) -> str:
         # So we print something that mostly looks like a dict, but doesn't
         # quote its keys: that's both more compact and something that can't
         # be mistaken for an actual dict or something that could be exec'd.
-        return "{{{}}}".format(
-            ', '.join(f"{d}: {self.get(d, '?')}" for d in self.graph.dimensions.names)
-        )
+        terms = [f"{d}: {self[d]!r}" for d in self.graph.required.names]
+        if self.hasFull():
+            terms.append("...")
+        return "{{{}}}".format(', '.join(terms))
```
andy-slac, line 306: `{{{}}}` makes me wonder if the result is going to be some sort of emoticon `}:‑)` 😈

python/lsst/daf/butler/core/_topology.py
```diff
@@ -171,25 +179,156 @@ def temporal(self) -> Optional[TopologicalFamily]:
...
+        -------
+        names : `tuple` [ `str` ]
+            Field name(s).  Guaranteed to be the same as the names of the field
+            specifications returned by `makeFieldSpecs`.
+        """
+        raise NotImplementedError()
+
+    @classmethod
+    @abstractmethod
+    def update(cls, extent: Optional[_R], *, name: Optional[str] = None,
```
andy-slac: Is there reason why `name` is a keyword-only argument for this method and not other methods?

python/lsst/daf/butler/core/_topology.py
```diff
@@ -171,25 +179,156 @@ def temporal(self) -> Optional[TopologicalFamily]:
...
+    @classmethod
+    @abstractmethod
+    def extract(cls, mapping: Mapping[str, Any], name: Optional[str] = None) -> Optional[_R]:
+        """Extract a region from a dictionary that represents a
+        database row in this representation.
+
+        Parameters
+        ----------
+        mapping : `Mapping` [ `str`, `Any` ]
+            A dictionary representing a database row containing a `Timespan`
```
andy-slac, line 292: `Timespan`? Should it be generic "extent" or "region"?

python/lsst/daf/butler/core/_topology.py
```diff
@@ -171,25 +179,156 @@ def temporal(self) -> Optional[TopologicalFamily]:
...
+            in this representation.  Should have key(s) equal to the return
+            value of `getFieldNames`.
+        name : `str`, optional
+            Name for the logical column; a part of the name for multi-column
+            representations.  Defaults to ``cls.NAME``.
+
+        Returns
+        -------
+        timespan : `Timespan` or `None`
+            Python representation of the timespan.
```
andy-slac: Timespan here too.

python/lsst/daf/butler/core/_topology.py
```diff
@@ -198,19 +337,136 @@ def fromSelectable(cls: Type[_S], selectable: sqlalchemy.sql.FromClause) -> _S:
...
+        Column containing the opaque byte-string, with automatic conversion to
+        `lsst.sphgeom.Region` implemented via SQLAlchemy hooks.
+    name : `str`
+        Name of the column.
+
+    Notes
+    -----
+    Unlike `TimespanDatabaseRepresentation`, this is a concrete class, because
+    we currently do not support any database-native spatial regions, and
+    instead rely on precomputed overlaps and and opaque (to the database) byte
```
andy-slac: `and and`

python/lsst/daf/butler/core/_topology.py
```diff
@@ -198,19 +337,136 @@ def fromSelectable(cls: Type[_S], selectable: sqlalchemy.sql.FromClause) -> _S:
...
+        return cls(selectable.columns[name], name)
+
+    @property
+    def name(self) -> str:
+        # Docstring inherited
+        return self._name
+
+    def flatten(self, name: Optional[str]) -> Iterator[sqlalchemy.sql.ColumnElement]:
+        # Docstring inherited
+        yield self.column
```
andy-slac, line 476: No `isNull()` implementation?

python/lsst/daf/butler/core/dimensions/_schema.py
```diff
@@ -176,25 +172,34 @@ def __init__(self, element: DimensionElement):
...
+    ) -> ddl.TableSpec:
         """Construct a complete specification for a table that could hold the
         records of this element.
 
         Parameters
         ----------
+        regRepr : `type` (`SpatialRegionDatabaseRepresentation` subclass)
+            Class object that specifies how spatial regions are represented in
+            the database.
+
```
andy-slac: Extra empty line

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -995,6 +995,28 @@ def getTimespanRepresentation(cls) -> Type[TimespanDatabaseRepresentation]:
...
+        objects are stored in this database.
+
+        `Database` does not automatically use the return type of this method
+        anywhere else; calling code is responsible for making sure that DDL
+        and queries are consistent with it.
+
+        Returns
+        -------
+        regRepr : `type` (`SpatialRegionDatabaseRepresention` subclass)
+            A type that encapsultes the way `lsst.sphgeom.Region` objects
```
andy-slac: Typo: encapsultes

python/lsst/daf/butler/registry/queries/_query.py
```diff
@@ -70,7 +69,7 @@ class Query(ABC):
     ----------
     graph : `DimensionGraph`
         Object describing the dimensions included in the query.
-    whereRegion : `lsst.sphgeom.Region`, optional
+    where : `lsst.sphgeom.Region`, optional
```
andy-slac: The actual parameter is still called `whereRegion`.

python/lsst/daf/butler/registry/databases/postgresql.py
```diff
@@ -196,18 +196,25 @@ def process_bind_param(self, value: Optional[Timespan],
...
-        lower = None if value.begin is None else time_utils.astropy_to_nsec(value.begin)
-        upper = None if value.end is None else time_utils.astropy_to_nsec(value.end)
+        if value.isEmpty():
+            return psycopg2.extras.NumericRange(empty=True)
+        else:
+            assert value._nsec[0] >= time_utils.MIN_NSEC, "Guaranteed by Timespan.__init__."
+            assert value._nsec[1] <= time_utils.MAX_NSEC, "Guaranteed by Timespan.__init__."
+            lower = None if value._nsec[0] == time_utils.MIN_NSEC else value._nsec[0]
+            upper = None if value._nsec[1] == time_utils.MAX_NSEC else value._nsec[1]
         return psycopg2.extras.NumericRange(lower=lower, upper=upper)
```
andy-slac: This should probably be indented with `else` block.

python/lsst/daf/butler/registry/queries/expressions/check.py
```diff
@@ -295,8 +172,9 @@ class InspectionVisitor(TreeVisitor[TreeSummary]):
         All known dimensions.
     """
 
-    def __init__(self, universe: DimensionUniverse):
+    def __init__(self, universe: DimensionUniverse, bindKeys: AbstractSet[str]):
```
andy-slac, line 177: Docstring for `bindKeys` could be useful.

python/lsst/daf/butler/registry/queries/expressions/check.py
```diff
@@ -410,10 +290,11 @@ class CheckVisitor(NormalFormVisitor[TreeSummary, InnerSummary, OuterSummary]):
         The dimensions the query would include in the absence of this
         expression.
     """
-    def __init__(self, dataId: DataCoordinate, graph: DimensionGraph):
+    def __init__(self, dataId: DataCoordinate, graph: DimensionGraph, bindKeys: AbstractSet[str]):
```
andy-slac, line 300: Docstring here too.

python/lsst/daf/butler/registry/queries/expressions/convert.py
```diff
@@ -0,0 +1,1004 @@
...
+            operation may match exactly.
+        ranges : `list` of `tuple`
+            List of (start, stop, step) tuples that represent ranges that the
+            left-hand side of the IN operation may match.
+        dtype : `type`
+            Type of the left-hand side operand for the IN expression.  Literals
+            should only be appended to if ``self.dtype is dtype``, and
+            ``ranges`` should only be appended to if ``dtype is int``.
+        node : `Node`
+            Original expression tree  nodethis converter represents; for use
```
andy-slac: Space on the left of ` nodethis` needs to move 4 characters to the right 🙂

python/lsst/daf/butler/registry/queries/expressions/convert.py
```diff
@@ -0,0 +1,1004 @@
...
+        if adapt is True:
+            adapt = adaptBinaryColumnFunc
+        elif adapt is False:
+            adapt = adaptIdentity
+        for lh in iterable(lhs):
+            if rhs is None:
+                self._binary[operator, lh, lh] = adapt(func, result if result is not None else lh)
+            else:
+                for rh in iterable(rhs):
+                    self._binary[operator, lh, rh] = adapt(func, result)
```
andy-slac, line 650: Is result allowed to be `None` here? Maybe add `assert` it it is not?

python/lsst/daf/butler/registry/queries/expressions/convert.py
```diff
@@ -0,0 +1,1004 @@
...
+            Operand, with ``operand.dtype`` and ``operator`` used to look up
+            the appropriate function.
+
+        Returns
+        -------
+        expression : `WhereClauseConverter`
+            Converter instance that represents the operation, created by
+            calling the registered function.
+        """
+        return self._unary[operator, operand.dtype](operand)
```
andy-slac, line 680: I guess this is expected to raise for unexpected combinations, maybe add `Raises` to docstring? Same for `applyBinary`.

python/lsst/daf/butler/registry/queries/expressions/convert.py
```diff
@@ -0,0 +1,1004 @@
...
+    Parameters
+    ----------
+    universe : `DimensionUniverse`
+        All known dimensions.
+    columns: `QueryColumns`
+        Struct that organizes the special columns known to the query
+        under construction.
+    elements: `NamedKeyMapping`
+        `DimensionElement` instances and their associated tables.
+    values: `Mapping`
```
andy-slac: `values` -> `bind`

python/lsst/daf/butler/registry/queries/expressions/convert.py
```diff
@@ -0,0 +1,1004 @@
...
+        # Docstring inherited from TreeVisitor.visitTimeLiteral
+        return ScalarWhereClauseConverter.fromLiteral(value)
+
+    def visitIdentifier(self, name: str, node: Node) -> WhereClauseConverter:
+        # Docstring inherited from TreeVisitor.visitIdentifier
+        if name in self.bind:
+            value = self.bind[name]
+            if isinstance(value, Timespan):
+                return TimespanWhereClauseConverter(self._tsRepr.fromLiteral(value))
+            return ScalarWhereClauseConverter.fromLiteral(self.bind[name])
```
andy-slac: `self.bind[name]` -> `value`

python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -1668,3 +1668,81 @@ def testIngestTimeQuery(self):
...
+                                                                     bind=bind,
+                                                                     where=where)}
+            )
+
+        # Try a bunch of timespan queries, mixing up the bounds themselves,
+        # where they appear in the expression, and how we get the timespan into
+        # the expression.
+
+        # t1 is before the start of i1, so this should not include i1.
+        self.assertEqual(ids[:i1], query("visit.timespan OVERLAPS (None, t1)"))
```
andy-slac: Hmm, I'm confused here, I guess "None" identifier is converted into Python `None` somehow but I don't remember seeing "None" handling in the code and it's not in the `bind`?

andy-slac (approved):
Looks OK, bunch of minor comments for docstrings mostly. I'm still not sure what to think about empty Timespan, its semantic looks confusing to me (but I'm easily confused by many things in daf_butler🙂).

python/lsst/daf/butler/core/timespan.py
TallJimbo: Yeah, this dates back to when there were also enum values for +/- inf, and it's just not useful anymore.  Will switch to `enum.auto()`.

python/lsst/daf/butler/core/time_utils.py
TallJimbo: Good point; I think I'll take the time now to hide these behind a compute-on-first-use guard.

python/lsst/daf/butler/core/timespan.py
TallJimbo: I wish I could, but that would still have to go through `__init__` (unless it called `object.__init__` directly or something, but I think that's definitely worse).  I really wish Python had real constructor overloading of some kind.

python/lsst/daf/butler/core/timespan.py
TallJimbo: I think so, and that'd effectively be a `(+Inf, +Inf)` timespan, which shouldn't be allowed.  I'll test that and lock it down.

TallJimbo: I think I misunderstood this comment originally (it's not a `(+Inf, +Inf)` case), but that still helped me improve some error messages and fix the problem.

python/lsst/daf/butler/core/timespan.py
TallJimbo: Oh, yikes, how did that slip through testing?  I thought the `padInstantaneous` logic would come into play, but you're completely right that it doesn't.

python/lsst/daf/butler/core/timespan.py
TallJimbo: Correct.  I figured this wasn't any worse than NaN messing with floating point numbers being ordered, but I could also rename these methods to e.g. `strictlyLess` and `strictlyGreater` if you think that would be any better.

python/lsst/daf/butler/core/dimensions/_coordinate.py
TallJimbo: Yes, intentional; `__repr__` is a single line, but I think there's good value from having one of those use multiple lines to make the output easier to read.  And given that built-in containers always call `repr` on their items, that's the one that should be single-line.

python/lsst/daf/butler/core/_topology.py
TallJimbo: Not really; I tend to just start forcing some kwargs after the total number starts to get large, because I find it very hard to read calls with too many positional args.  But I was probably overzealous here, and I'll just make these all positional.

python/lsst/daf/butler/registry/tests/_registry.py
TallJimbo: I added it as a constant entry in the dict in a higher-level class, but since then I've realized I should move it (probably make a `categorizeConstant` function or something) in order to make it more obvious and make it case-insensitive.  I'll probably just support `null` instead of `null` and `None`.

TallJimbo:
Thanks for the detailed review, @andy-slac.  I've pushed a few more commits and amended many more, and I think I've addressed all comments.  If you're up for it, two of the new ones may be worth another look.

- https://github.com/lsst/daf_butler/pull/450/commits/a26ee5c024b336aef66ea086c6005115a298ffc7
- https://github.com/lsst/daf_butler/pull/450/commits/82ccbb4fe23d7062dc060179f5180ab29639b630

The other new one, https://github.com/lsst/daf_butler/pull/450/commits/15afa4a0c4eefabf5871cfc1f972501999b8dece, is totally mechanical, so the most you'd want to look at is the commit message, I think.

python/lsst/daf/butler/core/timespan.py
andy-slac: I guess it's the same sot of trouble as NaN and for many people life would be easier if NaN did not exist at all 🙂 I do not know if it's worth renaming though, it depends on how these methods are used and I don't have complete picture for that in my head so it's up to you to decide.

python/lsst/daf/butler/registry/tests/_registry.py
andy-slac: +1 for `null` only, too confusing to have two different constants meaning the same thing.

python/lsst/daf/butler/registry/queries/expressions/check.py
```diff
@@ -194,10 +194,11 @@ def visitIdentifier(self, name: str, node: Node) -> TreeSummary:
...
             return TreeSummary()
-        if categorizeIngestDateId(name):
-            return TreeSummary(
-                hasIngestDate=True,
-            )
+        constant = categorizeConstant(name)
+        if constant is ExpressionConstant.INGEST_DATE:
+            return TreeSummary(hasIngestDate=True)
+        elif constant is not None:
+            return TreeSummary()
```
andy-slac: Maybe try to make it a bit more stable w.r.t. potential extension of ExpressionConstant, like:

    if constant is ExpressionConstant.INGEST_DATE:
        ...
    elif constant is ExpressionConstant.NULL:
        ....
    assert constant is None, "..."

python/lsst/daf/butler/core/time_utils.py
```diff
@@ -37,135 +37,154 @@
...
 
 _LOG = logging.getLogger(__name__)
 
 
-def astropy_to_nsec(astropy_time: astropy.time.Time) -> int:
-    """Convert astropy time to nanoseconds since epoch.
-
-    Input time is converted to TAI scale before conversion to
+class TimeConverter(metaclass=Singleton):
+    """A singleton that provides methods for mapping TAI times to integer
```
andy-slac, line 46: A bit surprising to see singleton used to solve this problem, but probably fine. I guess I just don't like singletons in general 🙂.

andy-slac:
I checked recent commits, they look fine, couple of minor comments.

PR #452, opened 2020-12-16
DM-28042: configure test temporary directory location via envvar

TallJimbo:


n8pease:
Looks good overall. The one thing that I noticed is that `base = os.environ.get("LSST_DAF_BUTLER_TEST_TMP", TESTDIR)` is repeated often enough, I wish there was a good way to reuse that, but it seems to appear in enough different places (`setUp`, `setUpClass`) that it's not a slam-dunk to use a common base class, and putting it in a helper function & importing that from e.g. `lsst.daf.butler.tests` results in more lines of code, not less, though the helper function may still be more reusable & refactorable in the long run.

timj:
@n8pease I was about to say the same thing. There are many cases where we seem to have the same two lines repeated over and over. Also, the other butler env vars start with DAF_BUTLER_ and not LSST_DAF_BUTLER_...

TallJimbo:
Happy to switch to `DAF_BUTLER_...`.

Not sure I can get the number of repeated lines down to one instead of two, but I'll give it a shot; I should be able to at least make it a different two lines so the number of times the envvar name appears in a string is drastically reduced.

TallJimbo:
Ok, completely refactored version is up (just amended the last commit).  All of the logic is now in some functions in `daf.butler.tests.utils`, including the `shutils` bit that we use to remove temporary directories.  The envvar name (now `DAF_BUTLER_TEST_TMP`) appears in just two places - one of those utility functions, and the SConstruct code that propagates it.

PR #453, opened 2020-12-17
DM-28095: Catch exception and warn instead when storage class is missing.

TallJimbo:


python/lsst/daf/butler/registry/_registry.py
```diff
@@ -1227,8 +1227,12 @@ def queryDatasetTypes(self, expression: Any = ..., *, components: Optional[bool]
...
                 if components and datasetType.isComposite():
                     # Automatically create the component dataset types
-                    for component in datasetType.makeAllComponentDatasetTypes():
-                        yield component
+                    try:
+                        componentsForDatasetType = datasetType.makeAllComponentDatasetTypes()
+                    except KeyError as err:
+                        _LOG.warn(f"Could not load storage class {err} for {datasetType.name}; "
+                                  "if it has components they will not be included in query results.")
+                    yield from componentsForDatasetType
```
erykoff: I assume if `makeAllComponentDatasetTypes` throws an exception, then `componentsForDatasetType` is `None`.  Is it okay to do `yield from None`?

erykoff (commented):
Are there additional tests to go with this?

python/lsst/daf/butler/registry/_registry.py
TallJimbo: Good catch!  I think it's actually an undefined variable, so the `yield from` is definitely going to break and I need to move it into a `try/else`, probably.

TallJimbo:
> Are there additional tests to go with this?

While trying to describe why I thought this was untestable, I came up with an idea for why it might not be; in the real world, I think you'd to need two different versions of daf_butler being used on one repo to encounter this, but I _think_ we may have already punched a testing-only loophole that I could reuse here...will give it a shot.

TallJimbo:
Now with tests!

It was 10x the work of the fix itself, but I caught a bug (`logging.warn` -> `logging.warning`, saves us a deprecation warning about the warning), and learned some stuff about how to test log messages.

erykoff (approved):
Looks good.

PR #454, opened 2021-01-04
DM-27153: default instrument (and skymap) in butler (and registry!), when possible

TallJimbo:


python/lsst/daf/butler/_butler.py
```diff
@@ -199,30 +194,16 @@ class Butler:
...
-    chained collection, just by passing multiple items is ``collections``::
+    from several others (as well)::
 
         butler = Butler("/path/to/repo", run="u/alice/DM-50000/a",
                         collections=["u/alice/DM-50000/a",
                                      "u/bob/DM-49998",
-                                     "raw/hsc"])
+                                     "HSC/defaults"])
+
+  *  This butler will `put` new datasets to the run ``u/alice/DM-50000/a``.
```
timj: This indentation looks odd to me.

python/lsst/daf/butler/_butler.py
```diff
@@ -232,64 +213,46 @@ class Butler:
...
         if butler is not None:
             if config is not None or searchPaths is not None or writeable is not None:
                 raise TypeError("Cannot pass 'config', 'searchPaths', or 'writeable' "
                                 "arguments with 'butler' argument.")
-            self.registry = butler.registry
+            self.registry = butler.registry.copy(defaults)
             self.datastore = butler.datastore
             self.storageClasses = butler.storageClasses
             self._config: ButlerConfig = butler._config
+            # These intentially invoke property setters.
```
timj: typo: intentionally

What "these" are being referred to here?

python/lsst/daf/butler/_butler.py
```diff
@@ -146,34 +145,20 @@ class Butler:
...
+        See :ref:`daf_butler_collection_expressions` for more information.
+        These collections are not registered automatically and must be
+        manually registered before they are used by any method, but they may be
+        manually registered after the `Butler` is initialized.
     run : `str`, optional
-        Name of the run datasets should be output to.  If the run
-        does not exist, it will be created.  If ``collections`` is `None`, it
-        will be set to ``[run]``.  If this is not set (and ``writeable`` is
-        not set either), a read-only butler will be created.
-    tags : `Iterable` [ `str` ], optional
```
timj, line 167: I think making tagged collections explicit rather than implicit is probably a good thing.

python/lsst/daf/butler/_butler.py
```diff
@@ -182,6 +167,16 @@ class Butler:
...
         provided, a read-write butler is created if any of ``run``, ``tags``,
         or ``chains`` is non-empty.
+    inferDefaults : `bool`, optional
+        If `True` (default) infer default data ID values from the values
+        present in the datasets in ``collections``: if all collections have the
+        same value (or no value) for a governor dimension, that value will be
+        the default for that dimension.  Nonexistent collections are ignored.
+        If a default value is provided explicitly for a governor dimension via
+        ``**kwargs``, no default will be inferred for that dimension.
+    **kwargs : `str`
```
timj, line 177: I wasn't initially okay with forever more using kwargs here but as I think further it does seem reasonable for the defaults to be the kwargs as they are for other butler APIs.

python/lsst/daf/butler/_butler.py
```diff
@@ -436,7 +399,7 @@ def makeRepo(root: str, config: Union[Config, str, None] = None,
 
     @classmethod
     def _unpickle(cls, config: ButlerConfig, collections: Optional[CollectionSearch], run: Optional[str],
-                  tags: Tuple[str, ...], writeable: bool) -> Butler:
+                  defaultDataId: Dict[str, str], writeable: bool) -> Butler:
```
timj, line 401: I'm assuming we don't have any pickled butlers anywhere that are now going to break? Am I right in thinking that almost all our butler pickling use cases are for pipetask forking?

python/lsst/daf/butler/_butler.py
```diff
@@ -803,21 +770,17 @@ def quote(s: Any) -> str:
             # those of the dataset type requested, because there may be extra
             # dimensions that provide temporal information for a validity-range
             # lookup.
-            dataId = DataCoordinate.standardize(dataId, universe=self.registry.dimensions, **kwds)
+            dataId = DataCoordinate.standardize(dataId, universe=self.registry.dimensions,
```
timj, line 772: Should the queryDimensionRecords call a few lines above also know about defaults? In that if the caller has decided against specifying the instrument could that return too many results?

python/lsst/daf/butler/core/dimensions/_coordinate.py
```diff
@@ -145,12 +146,16 @@ def standardize(
...
-            and ``universe`` must be provided unless ``mapping`` is already a
-            `DataCoordinate`.
+            If not provided, will be inferred from the keys of ``mapping`` and
+            ``**kwargs``, and ``universe`` must be provided unless ``mapping``
+            is already a `DataCoordinate`.
         universe : `DimensionUniverse`
             All known dimensions and their relationships; used to expand
             and validate dependencies when ``graph`` is not provided.
+        defaults : `DataCoordinate`, optional
+            Default dimension key-value pairs to use when needed as.  These are
```
timj: I don't think the "as" should be in that sentence.

python/lsst/daf/butler/registry/_defaults.py
```diff
@@ -0,0 +1,142 @@
...
+@immutable
+class RegistryDefaults:
+    """A struct used to provide the default collections searched or written to
+    by a `Registry` or `Butler` instance.
+
+    Parameters
+    ----------
+    collections : `str` or `Iterable` [ `str` ], optional
+        An expression specifying the collections to be searched (in order) when
+        reading datasets.
```
timj: isn't this docstring slightly wrong since it's searching all the collections to determine defaults and is not reading datasets.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -1058,6 +1058,10 @@ def expandDataId(self, dataId: Optional[DataId] = None, *, graph: Optional[Dimen
         records : `Mapping` [`str`, `DimensionRecord`], optional
             Dimension record data to use before querying the database for that
             data, keyed by element name.
+        noDefaults : `bool``, optional
```
timj: negative bools can get confusing. Is there a reason we can't have `withDefaults` defaulting to `True`?

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -1364,13 +1373,14 @@ def queryDatasets(self, datasetType: Any, *,
             `re.Pattern`, and iterables thereof.  The special value `...` can
             be used to query all dataset types.  See
             :ref:`daf_butler_dataset_type_expressions` for more information.
-        collections
+        collections: optional
```
timj, line 1376: Needs a type here I think.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -1418,8 +1428,13 @@ def queryDatasets(self, datasetType: Any, *,
         refs : `queries.DatasetQueryResults`
             Dataset references matching the given query criteria.
 
+        Raises
```
timj: Isn't this line duplicated immediately below?

python/lsst/daf/butler/registry/datasets/byDimensions/summaries.py
```diff
@@ -0,0 +1,358 @@
...
+_T = TypeVar("_T")
+
+
+class CollectionSummaryTables(Generic[_T]):
+    """Structure that holds the table or table specification objects that
+    summarize the contents of collections.
+
+    Parameters
+    ----------
+    datasetType
```
timj, line 67: Missing the type

python/lsst/daf/butler/registry/summaries.py
```diff
@@ -0,0 +1,328 @@
...
+    def copy(self) -> GovernorDimensionRestriction:
+        """Return a deep copy of this object.
+
+        Returns
+        -------
+        copy : `GovernorDimensionRestriction`
+            A copy of ``self`` that can be modified without modifying ``self``
+            at all.
+        """
+        return GovernorDimensionRestriction(NamedKeyDict((k, set(v)) for k, v in self.items()))
```
timj, line 146: Should this use `type(self)` instead of the explicit class name?

python/lsst/daf/butler/registry/summaries.py
```diff
@@ -0,0 +1,328 @@
...
+    def copy(self) -> CollectionSummary:
+        """Return a deep copy of this object.
+
+        Returns
+        -------
+        copy : `CollectionSummary`
+            A copy of ``self`` that can be modified without modifying ``self``
+            at all.
+        """
+        return CollectionSummary(datasetTypes=self.datasetTypes.copy(), dimensions=self.dimensions.copy())
```
timj, line 290: Similarly `type(self)` here just in case a subclass ever happens.

tests/test_butler.py
```diff
@@ -190,13 +188,13 @@ def runPutGetTest(self, storageClass, datasetTypeName):
 
         # Put with a preexisting id should fail
         with self.assertRaises(ValueError):
-            butler.put(metric, DatasetRef(datasetType, dataId, id=100))
+            ref = butler.put(metric, DatasetRef(datasetType, dataId, id=100))
```
timj: why do we need the `ref` returned here? We know it fails. Is this some mypy thing?

tests/test_simpleButler.py
```diff
@@ -356,6 +356,45 @@ def testGetCalibration(self):
...
+        Many tests that use default run already exist in ``test_butler.py``, so
+        that isn't tested here.  And while most of this functionality is
+        implemented in `Registry`, we test it here instead of
+        ``daf/butler/tests/registry.py`` because it shouldn't depend on the
+        database backend at all.
+        """
+        butler = self.makeButler(writeable=True)
+        butler.import_(filename=os.path.join(TESTDIR, "data", "registry", "base.yaml"))
+        butler.import_(filename=os.path.join(TESTDIR, "data", "registry", "datasets.yaml"))
+        # Need to actually set defaults later, not at construction, because
```
timj, line 373: Can you though test that the kwargs work with `Butler` construction? Maybe by creating new butler from this butler with different defaults.

Is there a test where you have two instruments registered and so defaulting fails?

Is there a test where you specify an explicit instrument default? The tests here seem to be using the implicit "we only have one instrument in this collection so use that" code path.

timj (approved):
Looks great. Only minor comments.

python/lsst/daf/butler/_butler.py
TallJimbo: The "these" lines below got removed, but I forgot to remove the comment.

python/lsst/daf/butler/_butler.py
TallJimbo: That is certainly my understanding.

python/lsst/daf/butler/_butler.py
TallJimbo: Good question - not something I had thought about myself - but it should happen internally/automatically (note that we only need to pass the defaults explicitly because `DataCoordinate.standardize` is not a `Registry` method).

python/lsst/daf/butler/registry/datasets/byDimensions/summaries.py
TallJimbo: I wasn't sure how to document the type here, given that it's a generic type parameter.

python/lsst/daf/butler/registry/_defaults.py
TallJimbo: Well, these _are_ the default collections searched when someone doesn't say what collections to search for datasets.  I think the problem is that the docstring doesn't talk about using these collections to determine default data ID values, which these may _also_ be used for.

python/lsst/daf/butler/registry/_registry.py
TallJimbo: No real preference; happy to switch.

python/lsst/daf/butler/registry/summaries.py
TallJimbo: I generally prefer to use the class name explicitly unless I've otherwise guaranteed that the `__init__` signatures of subclasses must be the same, but neither choice is particularly satisfying.

tests/test_butler.py
TallJimbo: I think it's just a copy-paste artifact; will remove.  We don't actually run MyPy on the tests (every time I do a "improve typing" ticket I give that a shot and then run away from the huge number of changes that would be needed).

tests/test_simpleButler.py
TallJimbo: More tests added.

python/lsst/daf/butler/registry/_registry.py
TallJimbo: The type is `Any`/`object` (or it's at least more types than we ever bother to annotate), and I _think_ the preferred style is to just leave that out.

python/lsst/daf/butler/_butler.py
TallJimbo: Deleted lines; I've now deleted the comment as well.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -1058,6 +1058,10 @@ def expandDataId(self, dataId: Optional[DataId] = None, *, graph: Optional[Dimen
         records : `Mapping` [`str`, `DimensionRecord`], optional
             Dimension record data to use before querying the database for that
             data, keyed by element name.
+        withDefaults : `bool``, optional
```
timj: Extra backtick here

PR #456, opened 2021-01-05
DM-28220: Update docstring for import_

parejkoj:


python/lsst/daf/butler/_butler.py
```diff
@@ -1594,24 +1594,25 @@ def import_(self, *, directory: Optional[str] = None,
...
+        """Import datasets into this repository that were exported from a
+        different butler repository via `~lsst.daf.butler.Butler.export`.
 
         Parameters
         ----------
         directory : `str`, optional
-            Directory containing dataset files.  If `None`, all file paths
-            must be absolute.
+            Directory containing dataset files to import from. If `None`,
+            ``filename`` and all file paths specified therein must be absolute.
```
mrawls: This edit does enhance clarity for the existing behavior, thank you.

That said, do we truly need a `directory` argument? Why not require that `filename` be the full path?

I specified what turned out to be a completely superfluous `directory` along with a full path `filename`. Now that I review what the code actually does, it only used the latter. In the midst of an operation requiring two repos and a script you write to invoke `import_` and some kind of `export.yaml` file all floating around, one fewer things to specify would be welcome.

timj: The exported yaml isn't required to be in the same place as all the exported data files.

mrawls: Ohhhh, the exported yaml doesn't necessarily have absolute filepath information for each data file in it. Argh. So my call to `import_` *did* use the `directory` argument. I mistakenly thought the only time `directory` was used was to correctly locate the exported yaml itself. Let me think up a way to make this clearer.

python/lsst/daf/butler/_butler.py
```diff
@@ -1594,24 +1594,25 @@ def import_(self, *, directory: Optional[str] = None,
...
         filename : `str` or `TextIO`, optional
             A stream or name of file that contains database information
             associated with the exported datasets.  If this a string (name) and
```
mrawls:
```suggestion
            associated with the exported datasets. Likely the file created by `~lsst.daf.butler.Butler.export`. If this a string (name) and
```

python/lsst/daf/butler/_butler.py
mrawls:
```suggestion
            Directory containing dataset files to import. If `None`,
            ``filename`` and all file paths specified therein to each dataset must be absolute.
```

TallJimbo (approved):
Looks great, thanks for the improvement!

PR #458, opened 2021-01-07
DM-28257: allow passing a pathlib.Path

timothydmorton:
This allows a `pathlib.Path` object to be passed as `config` param when initializing a Butler.

### DIFF
--- a/python/lsst/daf/butler/core/config.py
+++ b/python/lsst/daf/butler/core/config.py
@@ -33,6 +33,7 @@ import pprint
 import os
 import yaml
 import sys
+from pathlib import Path
 from yaml.representer import Representer
 import io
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/core/config.py
```diff
@@ -204,6 +205,8 @@ def __init__(self, other=None):
 
         if other is None:
             return
+        if isinstance(other, Path):
```
timj: I was imagining you'd add the instance check lower down where we check str, ButlerURI and then handoff to ButlerURI to deal with Path directly. Also need to change the docstring above to add `Path` along with `str` and `ButlerURI`.

timj (approved):
Looks good. Main comment is to not have Config try to interpret the `Path` but to forward it on to `ButlerURI`.

PR #459, opened 2021-01-12
DM-26688: Add command-line tool for Registry.associate

n8pease:


n8pease:
This work is built on top of the branch tickets/DM-26689, it will get checked in after that one.

python/lsst/daf/butler/script/_associate.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def associate(repo, collection, dataset_type, collections, where, find_first):
+    """Add existing datasets to ta CHAINED collection.
+    """
+
+    butler = Butler(repo, writeable=True)
+
+    if butler.registry.getCollectionType(collection) is not CollectionType.CHAINED:
+        raise RuntimeError("Can only add datasets to a chained collection.")
```
TallJimbo: Instead of this check, let's do
```
butler.registry.registerCollection(collection, CollectionType.TAGGED)
```

- This will create the collection if it does not exist, which will be very convenient, and it will raise an exception if it exists with the wrong type.
- The type needs to be `TAGGED`, not `CHAINED`

python/lsst/daf/butler/script/_associate.py
```diff
@@ -0,0 +1,45 @@
...
+
+    refs = QueryDatasets(
+        butler=butler,
+        glob=dataset_type,
+        collections=collections,
+        where=where,
+        find_first=find_first,
+        show_uri=False,
+        repo=None
+    )
```
TallJimbo, line 42: I like the idea of calling the script rather than the native method to share logic like transforming globs to regexes.  Do we need to worry about the script _also_ doing unnecessary things (like making astropy Tables) at all?

TallJimbo (approved):

python/lsst/daf/butler/script/_associate.py
n8pease: QueryDatasets generates astropy tables on demand via the getter (`getTables`), so it's structured to not create these if they're not needed. (I do see a mistake in passing the QueryDatasets results to the butler without using the accessor, will fix)

PR #461, opened 2021-01-15
DM-28314: Initial JSON investigation

timj:
This implements the simplest possible JSON support.

```
json = ref.to_json()
ref = DatasetRef.from_json(json, universe=universe)
```

A universe is required although I imagine we could make it optional if we could have a support serialization package that could be queried for a default universe to use.

To implement these I've also created `to_simple` and `from_simple` pairs for `DatasetRef`, `DatasetType`, `DimensionGraph` and `DataCoordinate` that convert the object to a dict or list and then take that and convert it back to the object.

### DIFF
--- a/python/lsst/daf/butler/core/datasets/ref.py
+++ b/python/lsst/daf/butler/core/datasets/ref.py
@@ -38,6 +38,7 @@ from ..configSupport import LookupKey
 from ..utils import immutable
 from ..named import NamedKeyDict
 from .type import DatasetType
+from ..json import from_json_generic, to_json_generic
 
 if TYPE_CHECKING:
     from ...registry import Registry
@@ -158,28 +159,6 @@ class DatasetRef:
         # Compare tuples in the priority order
         return (self_run, self.datasetType, self.dataId) < (other_run, other.datasetType, other.dataId)
 
-    def to_json(self, minimal: bool = False) -> str:
-        """Convert this class to JSON form.
-
-        The class type is not recorded in the JSON so the JSON decoder
-        must know which class is represented.
-
-        Parameters
-        ----------
-        minimal : `bool`, optional
-            Use minimal serialization. Requires Registry to convert
-            back to a full type.
-
-        Returns
-        -------
-        json : `str`
-            The class in JSON string format.
-        """
-        # For now use the core json library to convert a dict to JSON
-        # for us.
-        import json
-        return json.dumps(self.to_simple(minimal=minimal))
-
     def to_simple(self, minimal: bool = False) -> Union[int, Dict]:
         """Convert this class to a simple python type suitable for
         serialization.
@@ -192,7 +171,7 @@ class DatasetRef:
 
         Returns
         -------
-        as_dict : `dict`
+        simple : `dict` or `int`
             The object converted to a dictionary.
         """
         if minimal and self.id is not None:
@@ -235,8 +214,8 @@ class DatasetRef:
             The special graph of all known dimensions.
             Can be `None` if a registry is provided.
         registry : `lsst.daf.butler.Registry`, optional
-            Registry to use to convert simple name of a DatasetType to
-            a full `DatasetType`. Can be `None` if a full description of
+            Registry to use to convert simple form of a DatasetRef to
+            a full `DatasetRef`. Can be `None` if a full description of
             the type is provided along with a universe.
 
         Returns
@@ -279,33 +258,8 @@ class DatasetRef:
         return cls(datasetType, dataId,
                    id=simple["id"], run=simple["run"], hasParentId=simple["hasParentId"])
 
-    @classmethod
-    def from_json(cls, json_str: str,
-                  universe: Optional[DimensionUniverse] = None,
-                  registry: Optional[Registry] = None) -> DatasetRef:
-        """Convert a JSON string created by `to_json` and return a
-        `DatsetRef`.
-
-        Parameters
-        ----------
-        json_str : `str`
-            Representation of the dimensions in JSON format as created
-            by `to_json()`.
-        universe : `DimensionUniverse`
-            The special graph of all known dimensions. Passed directly
-            to `from_simple()`.
-        registry : `lsst.daf.butler.Registry`, optional
-            Registry to use to convert simple name of a DatasetType to
-            a full `DatasetType`. Passed directly to `from_simple()`.
-
-        Returns
-        -------
-        datasetType : `DatasetRef`
-            Newly-constructed object.
-        """
-        import json
-        as_dict = json.loads(json_str)
-        return cls.from_simple(as_dict, universe=universe, registry=registry)
+    to_json = to_json_generic
+    from_json = classmethod(from_json_generic)
 
     @classmethod
     def _unpickle(
--- a/python/lsst/daf/butler/core/datasets/ref.py
+++ b/python/lsst/daf/butler/core/datasets/ref.py
@@ -74,9 +74,6 @@ class DatasetRef:
         not be created in new code, but are still supported for backwards
         compatibility.  New code should only pass `False` if it can guarantee
         that the dimensions are already consistent.
-    hasParentId : `bool`, optional
-        If `True` this `DatasetRef` is a component that has the ``id``
-        of the composite parent.
 
     Raises
     ------
@@ -85,19 +82,17 @@ class DatasetRef:
         provided but ``run`` is not.
     """
 
-    __slots__ = ("id", "datasetType", "dataId", "run", "hasParentId")
+    __slots__ = ("id", "datasetType", "dataId", "run",)
 
     def __init__(
         self,
         datasetType: DatasetType, dataId: DataCoordinate, *,
         id: Optional[int] = None,
         run: Optional[str] = None,
-        hasParentId: bool = False,
         conform: bool = True
     ):
         self.id = id
         self.datasetType = datasetType
-        self.hasParentId = hasParentId
         if conform:
             self.dataId = DataCoordinate.standardize(dataId, graph=datasetType.dimensions)
         else:
@@ -190,7 +185,6 @@ class DatasetRef:
         # Convert to a dict form
         as_dict: Dict[str, Any] = {"datasetType": self.datasetType.to_simple(minimal=minimal),
                                    "dataId": self.dataId.to_simple(),
-                                   "hasParentId": self.hasParentId
                                    }
 
         # Only include the id entry if it is defined
@@ -250,7 +244,7 @@ class DatasetRef:
         datasetType = DatasetType.from_simple(simple["datasetType"], universe=universe, registry=registry)
         dataId = DataCoordinate.from_simple(simple["dataId"], universe=universe)
         return cls(datasetType, dataId,
-                   id=simple["id"], run=simple["run"], hasParentId=simple["hasParentId"])
+                   id=simple["id"], run=simple["run"])
 
     to_json = to_json_generic
     from_json = classmethod(from_json_generic)
@@ -262,15 +256,14 @@ class DatasetRef:
         dataId: DataCoordinate,
         id: Optional[int],
         run: Optional[str],
-        hasParentId: bool,
     ) -> DatasetRef:
         """A custom factory method for use by `__reduce__` as a workaround for
         its lack of support for keyword arguments.
         """
-        return cls(datasetType, dataId, id=id, run=run, hasParentId=hasParentId)
+        return cls(datasetType, dataId, id=id, run=run)
 
     def __reduce__(self) -> tuple:
-        return (self._unpickle, (self.datasetType, self.dataId, self.id, self.run, self.hasParentId))
+        return (self._unpickle, (self.datasetType, self.dataId, self.id, self.run))
 
     def __deepcopy__(self, memo: dict) -> DatasetRef:
         # DatasetRef is recursively immutable; see note in @immutable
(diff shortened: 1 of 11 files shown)


timj:
@TallJimbo would you mind taking a quick look at this initial implementation before I continue on.

Thinking out loud about the client/server usage, it might be a good idea to add a `minimalist=True` argument to the `to_json` method to let it return something minimal for wire transport to a server. For example for a `DatasetRef` you in theory only need the `id` and the server can work out what you meant, and for `DatasetType` the name is sufficient. This would then need the `from_json()` methods to take an optional registry argument to let them do a registry lookup to create the full object in the server.  The server should return fully-specified results to the client otherwise that could result in additional registry lookups to the server to resolve everything later on.

So client Registry always uses `to_json(minimalist=True)`, server always used `from_json(registry=R)` and server always generates JSON with `to_json(minimalist=False)`. Or something like that.

TallJimbo (approved):
Looks great!  Sorry it took me so long to take a look.

I definitely like standardizing on `to_json` and `from_json`, and those delegating to `to_simple` and `from_simple` and the way you're handling the `DimensionUniverse` in those.

I'm trying to think of a better name to replace "simple" but haven't come up with one.  I'd have probably used "pod" for Plain Old Data, but as I think we've discussed I apparently think of that terminology as way more common that it actually is.

What do you think about a decorator or mixin to inject the `to_json` and `from_json` methods that delegate to `to_simple` and `from_simple`?  Seems like we should do something like that to make these mechanical, but that may not be compatible with universe handling and/or MyPy.

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -152,6 +159,108 @@ def __lt__(self, other: Any) -> bool:
...
+        -------
+        simple : `dict` or `int`
+            The object converted to a dictionary.
+        """
+        if minimal and self.id is not None:
+            # The only thing needed to uniquely define a DatasetRef
+            # is the integer id so that can be used directly if it is
+            # resolved and if it is not a component DatasetRef
+            if not self.isComponent():
+                return self.id
```
TallJimbo: Would it simplify the `from_simple` implementation much to return a dict with just the `id` key here?  If we someday add an `origin` term to the primary key, that'll be an easier transition.

And, if our primitives always use `Dict[str, Any]` as their "simple" representation (i.e. we go in a similar direction with `DatasetType`), we'll be able to more easily add a `Protocol` (thanks, Python 3.8) for the `to_simple`/`from_simple` interfaces and get a bit more type safety when we start using them.

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -152,6 +159,108 @@ def __lt__(self, other: Any) -> bool:
...
+
+            ref = registry.getDataset(simple)
+            if ref is None:
+                raise RuntimeError(f"No matching dataset found in registry for id {simple}")
+            return ref
+
+        # Minimalist component will just specify component and id and
+        # require registry to reconstruct
+        if set(simple) == {"id", "component"}:
+            if registry is None:
```
TallJimbo, line 225: It seems like it might be more efficient to swap these checks, and assume the input was produced with `minimal=True` if `registry is not None`.  I have no idea if that's the kind of "more efficient" that could matter down the road, but it does seem conceivable that we'll spend a lot of time in this code in the client/server butler.

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -152,6 +159,108 @@ def __lt__(self, other: Any) -> bool:
...
+                raise ValueError("Registry is required to construct component DatasetRef from integer id")
+            parent_ref = registry.getDataset(simple["id"])
```
TallJimbo: ...but a bigger issue for performance is almost certainly the fact that this call isn't vectorized, and it'll be death-by-many-tiny-queries on the server whenever anyone sends a long list of minimal `DatasetRefs` over the wire.  We can of course work around that by not using the minimal form, but maybe this is a sign that we should be using `FakeDatasetRef` in more places and/or just reworking what a `DatasetRef` is supposed to be.  Not a problem for this ticket, I think, but potentially a big one for the client/server butler project (unless sending non-minimal DatasetRefs around ends up being totally fine).

Also note that if we started out with a `DatasetRef` such that `ref.dataId.hasRecords() is True`, that will not be the case after round-trip through minimal form (and changing that would be many-tiny-queries x20).  That's probably okay, it's something to look out for (and probably document).

python/lsst/daf/butler/core/datasets/type.py
```diff
@@ -472,6 +474,94 @@ def _lookupNames(self) -> Tuple[LookupKey, ...]:
...
+        Returns
+        -------
+        datasetType : `DatasetType`
+            Newly-constructed object.
+        """
+        if isinstance(simple, str):
+            if registry is None:
+                raise ValueError(f"Unable to convert a DatasetType name '{simple}' to DatasetType"
+                                 " without a Registry")
+            datasetTypes = list(registry.queryDatasetTypes(simple))
```
TallJimbo: I suspect this is going to be a fair bit slower than `Registry.getDatasetType`, and it should be impossible to get more than one result anyway; there's a unique constraint in the database, and the pattern-matching logic won't be in play unless the object passed in is a `re.Pattern` object, not a `str`.

Also, this minimal form should be safe from the many-tiny-queries problem that `DatasetRef` suffers from, at least for our current manager implementations, because those aggressively fetch all `DatasetTypes` from the database and stuff them in dict, so no DB calls should be needed here.  Hopefully that sort of aggressive fetching and/or caching scales up to the full science-user database...

python/lsst/daf/butler/core/datasets/type.py
```diff
@@ -472,6 +474,94 @@ def _lookupNames(self) -> Tuple[LookupKey, ...]:
...
+            a full `DatasetType`. Can be `None` if a full description of
+            the type is provided along with a universe.
+
+        Returns
+        -------
+        datasetType : `DatasetType`
+            Newly-constructed object.
+        """
+        if isinstance(simple, str):
+            if registry is None:
```
TallJimbo, line 532: Same question as with `DatasetRef` as to whether checking for `registry is None` in the outer block would speed things up and/or simplify the code.

python/lsst/daf/butler/core/dimensions/_records.py
```diff
@@ -175,6 +181,105 @@ def __reduce__(self) -> tuple:
...
+        mapping = {name: getattr(self, name) for name in self.__slots__}
+        # If the item in mapping supports simplification update it
+        for k, v in mapping.items():
+            try:
+                mapping[k] = v.to_simple(minimal=minimal)
+            except AttributeError:
+                if isinstance(v, lsst.sphgeom.Region):
+                    # Match YAML serialization
+                    mapping[k] = {"cls": f"lsst.sphgeom.{type(v).__name__}",
+                                  "encoded": v.encode().hex()}
```
TallJimbo: I actually noticed the other day that the sphgeom class is already included in the encoded representation, and you can use `Region.decode` to handle a byte string produced by any derived class.  It's might be too late to be worth switching the YAML representation, but we could get away with just `bytes` for this one, if we wanted to.

python/lsst/daf/butler/core/json.py
```diff
@@ -0,0 +1,91 @@
...
+    Returns
+    -------
+    constructed : Any
+        Newly-constructed object.
+    """
+    simple = json.loads(json_str)
+    try:
+        return cls.from_simple(simple, universe=universe, registry=registry)  # type: ignore
+    except AttributeError as e:
+        raise AttributeError(f"JSON deserialization requires {cls} has a from_simple() class method") from e
```
TallJimbo, line 100: With a `Protocol` for `from_simple`/`to_simple`, I think we could replace this run-time check with a mypy one.

TallJimbo (approved):
Looks good, even though I've left a lot of nitpicky comments - those are mostly musing about how things might or might not work as we'd hope when we start to use them for client/server registry.

Some of those line comments are about whether we should use `registry is not None` in `from_simple` to _assume_ that the incoming form is minimal.  I think that would simplify the code, and it should be just as good at supporting over-the-wire needs, given that client and server should be able to expect when something is using minimal form.  If we do go that route, it may be worth playing with making the minimal versions of things into separate methods, even though that multiplies the number of methods by two, and that's a lot of methods.  But if most classes don't actually have a minimal form, and don't need those methods, maybe that's better?  I'm not sure either way.

Finally, I think you've got the right amount of checking for invalid inputs here, in the sense that they're reasonable things the user might get wrong, but there's no attempt to do really exhaustive schema validation here (which I think is not something we should roll ourselves, if we want it).  I don't actually know enough about developing http interfaces to know if we _should_ want exhaustive schema validation - but if there's a chance that we do, that's another thing we should think about before really settling on this approach.

python/lsst/daf/butler/core/datasets/ref.py
timj: We can always use a dict given that the origin will turn up at some point.

Not so sure about turning DatasetType into `{"name": nam}` but that could be done.

python/lsst/daf/butler/core/datasets/ref.py
timj: I think we might want a specialist method in the server for processing the collection of JSON simple dataset refs. The server always has to send the full form to the client but the client can send the minimal form back to the server. We can tell up front in the server if we are processing a collection of them.

python/lsst/daf/butler/core/datasets/type.py
timj: I have no idea why I was using `query` and not `get` there...

python/lsst/daf/butler/core/dimensions/_records.py
timj: I noticed too that the YAML form was storing the class and then not using it. It seems like I messed up somewhere as we iterated to the final form. On the other hand the `cls` here does make it obvious that it's a region although using a single key in a dict of `encoded_region` would also do that.

python/lsst/daf/butler/core/json.py
timj: Sounds like I need to look up protocols. mypy was really annoying in this file.

python/lsst/daf/butler/core/datasets/ref.py
timj: I think the server is always going to pass in a registry parameter and the client is always going to call `from_simple` with a universe, but we can't enforce every client using the minimal parameter, so I always have to look at the content to see whether it is minimal or not.

timj:
Regarding schema validation, it seems to me that we'd probably want to look at pydantic for that https://pydantic-docs.helpmanual.io/usage/exporting_models/  -- it has json support. Naively we could make the `to_simple` methods return a `SerializedDatasetRef` (or whatever) that is defined by the pydantic model. The constructor for that would take the json plain-old-data and then we'd call `from_simple` on that `SerializedDatasetRef`. Pydantic would handle all the checks. Of course the other option is to make `DatasetRef` an actual pydantic model but that might be a complicated retrofit.   Supporting `DimensionsRecords` seems a bit harder but maybe we could create a SerializedVisitRecord and SerializedExposureRecord dynamically from the dimensions.yaml.

TallJimbo:
I'm mostly reasoning by analogy with Rust's Serde, which sounds pretty similar to pydantic, but if we go in that direction I'd lean towards:
 - `to_pydantic`/`from_pydantic` convert out objects to a corresponding pydantic model instance; may return `self` (and similar for `from`) if the class can easily just be a pydantic model itself;
 - make `to_json`/`from_json` delegate to that;
 - add `to_simple`/`from_simple` that also delegate to that, if it looks like we still have a need for them.

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -152,6 +158,103 @@ def __lt__(self, other: Any) -> bool:
...
+            if self.isComponent():
+                # We can still be a little minimalist with a component
+                # but we will also need to record the datasetType component
+                simple["component"] = self.datasetType.component()
+            return simple
+
+        # Convert to a dict form
+        as_dict: Dict[str, Any] = {"datasetType": self.datasetType.to_simple(minimal=minimal),
+                                   "dataId": self.dataId.to_simple(),
+                                   "hasParentId": self.hasParentId
```
timj: @TallJimbo I'm fairly certain that `hasParentId` is a left over for when we first changed the handling of components in registry and it was possible to have a component in registry or in datastore depending on how the put had been implemented.

It was added in #269 and I don't see any code that checks the value anywhere.

TallJimbo: I don't recall anything that uses it (I've forgotten the details how we identify component dataset refs), so I'm all for removing if it doesn't break anything.

timj: I removed it and everything is still working...

python/lsst/daf/butler/core/dimensions/_records.py
timj: Changed it to just use the "encoded_region" key. It can't use bytes because json requires strings.

python/lsst/daf/butler/core/datasets/ref.py
timj: I've changed DatasetRef to always return a dict.

It's probably the case that switching to pydantic at some point would result in everything being a dict so maybe I should switch DatasetType over as well in the interests of consistency even though it adds 8 extra characters to all the serializations.

python/lsst/daf/butler/core/datasets/type.py
timj: Is an "is None" test expensive? The instance test presumably costs more (as would a test for a dict with only "name" in it). The client is always going to call this with universe and the server is always going to call this with registry but I'm not sure we can rely on that to be true for all possible JSON serializations.

TallJimbo: I would guess that `is None` boils down to a pointer comparison in C, so it's definitely going to be faster than `isinstance`, but I'd bet `isinstance(obj, str)` is similarly fast when the object actually is a `str` (I'd bet both cases are dominated by just per-statement interpreter overheads). It's the case where it's going to return `False` that I don't know about; it seems like it _should_ be possible for that to be quite fast, too, for `str` and other builtins, even if it might be much slower for user-defined classes.

python/lsst/daf/butler/core/json.py
timj: I've added protocol and removed the ignore. I'm inclined to leave the try/except so you get a more obvious error message. mypy seems happy (mypy didn't use the try/except itself).

PR #465, opened 2021-01-26
DM-23862: Split data ID handling off from butlerTests.makeTestRepo

kfindeisen:
This PR refactors how data IDs are handled in `lsst.daf.butler.tests` to allow more flexible, more precise, and more intuitive test environment creation. The old API is discouraged but not deprecated. The PR also includes some other minor fixes and code improvements.

### DIFF
--- a/doc/lsst.daf.butler/use-in-tests.rst
+++ b/doc/lsst.daf.butler/use-in-tests.rst
@@ -51,6 +51,25 @@ If not, the `expandUniqueId` function can recover which values are assigned to w
    >>> butlerTests.expandUniqueId(repo, {"detector": 2})
    DataCoordinate({instrument, detector}, ('FancyCam', 2))
 
+The `addDataIdValue` function defines allowed data ID values after the repository has been created.
+Unlike the data ID parameter to `makeTestRepo`, `addDataIdValue` lets you optionally specify which values are related to which.
+Any required but unspecified relationships are filled arbitrarily, so you need only give the ones your tests depend on.
+
+.. code-block:: py
+
+   from lsst.daf.butlker.tests import addDataIdValue
+
+   addDataIdValue(butler, "skymap", "map")
+   # Tract requires a skymap; can assume it's "map" because no other options.
+   # Were there more than one skymap, this would choose one unpredictably.
+   addDataIdValue(butler, "tract", 42)
+   # Explicit specification.
+   addDataIdValue(butler, "tract", 43, skymap="map")
+   # Can map dimensions in a many-to-many relationship.
+   for patch in [0, 1, 2, 3, 4, 5]:
+       addDataIdValue(butler, "patch", patch, tract=42)
+       addDataIdValue(butler, "patch", patch, tract=43)
+
 The `addDatasetType` function registers any dataset types needed by the test (e.g., "calexp").
 As with registering the data IDs, this is a prerequisite for actually reading or writing any datasets of that type during the test.
 
--- a/doc/lsst.daf.butler/use-in-tests.rst
+++ b/doc/lsst.daf.butler/use-in-tests.rst
@@ -25,35 +25,14 @@ Creating and populating a repository
 ====================================
 
 The `lsst.daf.butler.tests.makeTestRepo` function creates an in-memory repository.
-In addition to the arguments used by `~lsst.daf.butler.Butler.makeRepo`, `makeTestRepo` takes a mapping of the data ID(s) needed by the test.
-This mapping does not create any datasets itself, but instead prepares the registry so that datasets with those IDs *could* be created later.
-
-The data ID mapping takes the form of a key for each data ID dimension, and a list of possible values.
-For example:
-
-.. code-block:: py
-
-   {"instrument": ["FancyCam"],
-    "exposure": [101, 102, 103],
-    "detector": [0, 1, 2, 3],
-    "skymap": ["map"],
-    "tract": [42, 43],
-    "patch": [0, 1, 2, 3, 4, 5],
-   }
-
-To avoid creating ambiguous data IDs, values are assigned to each other in an arbitrary one-to-one fashion, rather than trying every combination (e.g., in the above example all the patches might only be assigned to tract 42).
-In many tests this is not a problem, as a single data ID is all that's needed.
-If not, the `expandUniqueId` function can recover which values are assigned to which:
-
-.. code-block:: py
-
-   >>> import lsst.daf.butler.tests as butlerTests
-   >>> butlerTests.expandUniqueId(repo, {"detector": 2})
-   DataCoordinate({instrument, detector}, ('FancyCam', 2))
+The repository is optimized for speed rather than for production processing, but otherwise is similar to `lsst.daf.butler.Butler.makeRepo`.
 
 The `addDataIdValue` function defines allowed data ID values after the repository has been created.
-Unlike the data ID parameter to `makeTestRepo`, `addDataIdValue` lets you optionally specify which values are related to which.
-Any required but unspecified relationships are filled arbitrarily, so you need only give the ones your tests depend on.
+By default, related values are assigned to each other in an arbitrary one-to-one fashion.
+This is good enough for many tests, where a single data ID is all that's needed.
+
+If you need more control (or if the assignment algorithm produces inconsistent results), `addDataIdValue` lets you specify by keyword which values are related to which.
+Any unspecified relationships are still filled arbitrarily, so you need only give the ones your tests depend on.
 
 .. code-block:: py
 
--- a/python/lsst/daf/butler/tests/_testRepo.py
+++ b/python/lsst/daf/butler/tests/_testRepo.py
@@ -151,21 +151,7 @@ def _makeRecords(dataIds, universe):
         expandedIds[name] = []
         dimension = universe[name]
         for value in values:
-            expandedValue = {}
-            for key in dimension.uniqueKeys:
-                if key.nbytes:
-                    castType = bytes
-                else:
-                    castType = key.dtype().python_type
-                try:
-                    castValue = castType(value)
-                except TypeError:
-                    castValue = castType()
-                expandedValue[key.name] = castValue
-            for key in dimension.metadata:
-                if not key.nullable:
-                    expandedValue[key.name] = key.dtype().python_type(value)
-            expandedIds[name].append(expandedValue)
+            expandedIds[name].append(_fillAllKeys(dimension, value))
 
     # Pick cross-relationships arbitrarily
     for name, values in expandedIds.items():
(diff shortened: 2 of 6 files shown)


doc/lsst.daf.butler/use-in-tests.rst
```diff
@@ -25,31 +25,23 @@ Creating and populating a repository
...
-   }
+If you need more control (or if the assignment algorithm produces inconsistent results), `addDataIdValue` lets you specify by keyword which values are related to which.
+Any unspecified relationships are still filled arbitrarily, so you need only give the ones your tests depend on.
 
-To avoid creating ambiguous data IDs, values are assigned to each other in an arbitrary one-to-one fashion, rather than trying every combination (e.g., in the above example all the patches might only be assigned to tract 42).
-In many tests this is not a problem, as a single data ID is all that's needed.
-If not, the `expandUniqueId` function can recover which values are assigned to which:
-
-.. code-block:: py
+.. code-block::py
```
mrawls:
```suggestion
.. code-block:: py
```
Without the space, the example block does not render

doc/lsst.daf.butler/use-in-tests.rst
```diff
@@ -25,31 +25,23 @@ Creating and populating a repository
...
 
-   >>> import lsst.daf.butler.tests as butlerTests
-   >>> butlerTests.expandUniqueId(repo, {"detector": 2})
-   DataCoordinate({instrument, detector}, ('FancyCam', 2))
+   addDataIdValue(butler, "skymap", "map")
```
mrawls, line 41: Please include prerequisites, e.g., `from lsst.daf.butler.tests import addDataIdValue` (and, for bonus points, importing and instantiating `butler`)

doc/lsst.daf.butler/use-in-tests.rst
```diff
@@ -25,31 +25,23 @@ Creating and populating a repository
...
+   addDataIdValue(butler, "tract", 42)
+   addDataIdValue(butler, "tract", 43, skymap="map")  # Redundant; only one map
```
mrawls: Please clarify what redundant means here. Two concerns. First, it says "you do not need to specify relationships you will not use" as well as "any unspecified dimensions will be linked arbitrarily." I'm still not sure which of these statements (if either) is being demonstrated with the redundant keyword example.

Second, does redundant mean the example line will break things? Or is just bad practice/superfluous? Immediately makes me wonder, what if I did `skymap="ponies"` or `ponies="map"`...

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -341,17 +345,45 @@ def addDataIdValue(butler, dimension, value):
         The name of the dimension to gain a new value.
     value
         The value to register for the dimension.
+    **related
+        Any existing dimensions to be linked to ``value``.
+
+    Examples
+    --------
+
+    .. code-block::py
```
mrawls: This is identical to the example in the docs, yes? I'm 100% pro examples, but I worry this is "duplicated code" that could get out of sync. This one is also lacking a space between `::` and `py` and I have the same questions about the redundant keyword bit.

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -237,7 +338,65 @@ def expandUniqueId(butler, partialId):
...
+    dimensionRecord = fullDimension.RecordClass(**completeValue)
+    try:
+        butler.registry.syncDimensionData(dimension, dimensionRecord)
+    except sqlalchemy.exc.IntegrityError as e:
+        raise RuntimeError("Could not create data ID value. Automatic relationship generation "
+                           "may have failed; try adding keywords to assign a specific instrument, "
+                           "filter, etc. based on the nested exception message.") from e
+
+
+def addDatasetType(butler: Butler, name: str, dimensions: Set[str], storageClass: str) -> DatasetType:
```
mrawls, line 407: Thanks for adding all the function type checking and annotations throughout! I've been learning about them in book club 🙂

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -186,14 +184,117 @@ def _makeRecords(dataIds, universe):
...
+        expandedValue[key.name] = castValue
+    for key in dimension.metadata:
+        if not key.nullable:
+            expandedValue[key.name] = key.dtype().python_type(value)
+    return expandedValue
+
+
+def _matchAnyDataId(record: Mapping[str, Any], registry: Registry, dimension: Dimension):
+    """Matches a partial dimension record to an existing record along a
+    specific dimension.
```
mrawls, line 231: I'm a little surprised this functionality, as well as `_fillRelationships`, doesn't already exist somewhere you could import from elsewhere in daf_butler. (Though I don't know immediately where - perhaps it's tied up with way more than is needed for a test repo?).

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -237,7 +338,65 @@ def expandUniqueId(butler, partialId):
...
+    expandedValue.update(**related)
+    completeValue = _fillRelationships(fullDimension, expandedValue, butler.registry)
+
+    dimensionRecord = fullDimension.RecordClass(**completeValue)
+    try:
+        butler.registry.syncDimensionData(dimension, dimensionRecord)
+    except sqlalchemy.exc.IntegrityError as e:
+        raise RuntimeError("Could not create data ID value. Automatic relationship generation "
+                           "may have failed; try adding keywords to assign a specific instrument, "
+                           "filter, etc. based on the nested exception message.") from e
```
mrawls: `filter` --> `band` or `physical_filter` ?

tests/test_testRepo.py
```diff
@@ -43,13 +43,14 @@ def setUpClass(cls):
...
-            "instrument": ["notACam", "dummyCam"],
-            "physical_filter": ["k2020", "l2019"],
-            "visit": [101, 102],
-            "detector": [5]
-        }
-        cls.creatorButler = makeTestRepo(cls.root, dataIds)
+        cls.creatorButler = makeTestRepo(cls.root)
+        addDataIdValue(cls.creatorButler, "instrument", "notACam")
+        addDataIdValue(cls.creatorButler, "instrument", "dummyCam")
+        addDataIdValue(cls.creatorButler, "physical_filter", "k2020", band="k", instrument="notACam")
```
mrawls, line 49: As usual, test examples are illuminating. Yet, I have some questions. All the `**related` items being input here are nominally of type string (per the function definition), and I guess the list of existing dimensions to be linked could be as long as desired. So, as a user invoking `addDataIdValue`, how to best decide what is related and what's not? For example, am I just supposed to *know* that adding a `physical_filter` requires linking a `band` and an `instrument`? Or is it not required at all and is just used in this test as a feasible example?

tests/test_testRepo.py
```diff
@@ -80,21 +81,38 @@ def _checkButlerDimension(self, dimensions, query, expected):
...
-                                     {"instrument": "dummyCam", "visit": 101}])
-        self. _checkButlerDimension({"visit", "instrument"},
-                                    "visit=102",
-                                    [{"instrument": "notACam", "visit": 102},
-                                     {"instrument": "dummyCam", "visit": 102}])
-        self. _checkButlerDimension({"detector", "instrument"},
-                                    "detector=5",
-                                    [{"instrument": "notACam", "detector": 5},
-                                     {"instrument": "dummyCam", "detector": 5}])
+        self._checkButlerDimension({"instrument"},
```
mrawls, line 86: Whoa, these had *spaces* after `self.` before? Was that just broken or...?

tests/test_testRepo.py
```diff
@@ -80,21 +81,38 @@ def _checkButlerDimension(self, dimensions, query, expected):
...
+        self._checkButlerDimension({"visit", "instrument"},
+                                   "visit=102",
+                                   [{"instrument": "notACam", "visit": 102},
+                                    {"instrument": "dummyCam", "visit": 102}])
+        self._checkButlerDimension({"detector", "instrument"},
+                                   "detector=5",
+                                   [{"instrument": "notACam", "detector": 5},
+                                    {"instrument": "dummyCam", "detector": 5}])
+
+    def testAddDataIdValue(self):
```
mrawls, line 102: Possibly missing: cases where a dimension and value are given, but necessary linked additional dimensions and values are **not** given.

mrawls (commented):
I feel like I'm at a place where I understand 80% of what is happening, and I'm hoping you can look at my comments and help me get closer to 100% so I can feel good about finishing the review. Thanks 🙂

doc/lsst.daf.butler/use-in-tests.rst
kfindeisen: I think those two sentences refer to the same thing, just in different words. 😕 

"Redundant" here in that, since there's only one skymap to choose from, the tract must be associated with that skymap whether you ask for it or not.

Invalid values and invalid keywords both raise exceptions.

python/lsst/daf/butler/tests/_testRepo.py
kfindeisen: The problem is that in a real repo you don't want to do this; you want your detectors, exposures, visits, tracts, etc. to be dictated by real data. All these functions basically make up filler values.

python/lsst/daf/butler/tests/_testRepo.py
kfindeisen: My experience is that the function docstrings are a lot more visible than the how-to guide, but I don't want the guide to not have examples...

tests/test_testRepo.py
kfindeisen: I have no idea how that code slipped in, but yes, it is legal Python. 😂

tests/test_testRepo.py
kfindeisen: Yeah, that's something I don't know how to handle.

Looking at https://github.com/lsst/daf_butler/blob/master/python/lsst/daf/butler/configs/dimensions.yaml#L80, `physical_filter` requires `instrument`, but does not require `band`. As far as I know this isn't documented anywhere, even though you also need to know it for writing a `PipelineTask` (the testing of which is the main use case for this code) or even passing data IDs to `pipetask`.

My intent was that this module could be used with no more knowledge of the dimensions system than you need for `pipetask` or to write a `PipelineTask`. Ideally, _all_ these keywords would be optional, with `_fillRelationships` filling them in as needed. However, I found that `_fillRelationships` would often pick contradictory values (e.g., exposure 101 is associated with camera 1 and with the g3 filter, but g3 belongs to camera 2). I couldn't come up with an algorithm that _can't_ do that (but still finds any self-consistent mapping), so I opted for putting warnings in the error handler and the documentation about the need for human guidance. It turns out the old code had the same problem (but with no workaround), I just got lucky during testing. 😬 

Any advice would be much appreciated.

timj: In https://github.com/lsst/daf_butler/blob/master/python/lsst/daf/butler/configs/dimensions.yaml#L81-L82 you can see that physical_filter requires an instrument but implies a band. So a physical filter on its own is not unique but giving it an instrument makes it fully deterministic. Once you know the actual physical+instrument then you can determine the band implicitly.

python/lsst/daf/butler/tests/_testRepo.py
mrawls: Would it be possible to say "for example usage, see [URL here]" in the docstring? It just feels like every time we have docs in more than one place they get out of sync and we forget which one is right.

timj: Normal sphinx lets you read code snippets from a separate file so that it's guaranteed to be identical. I don't think I've every tried it from code sphinx docs but maybe it would work anyhow?

python/lsst/daf/butler/tests/_testRepo.py
mrawls: Ah, interesting. Consider adding something akin to, "Only for creating filler dimension data in a test repo, as datasets in a real repo have known dimension values." Otherwise I might be tempted to "borrow" functions like these to complete/tidy up/fill gaps in a real repo.

doc/lsst.daf.butler/use-in-tests.rst
mrawls: In this case, please make the comment a little more specific, e.g. "There's only one skymap, so specifying it isn't required, but it is permitted." (Stating things are "redundant" has me worried something will behaving in a butler-y way and break because there is already an instance of X in my repo.)

tests/test_testRepo.py
mrawls: I guess what I'm really getting at here is this. To improve the tests, is there some combo of things you could put in `**related` that would cause `addDataIdValue` to break because the inputs are self-inconsistent? Like if you told it band is r and physical_filter is HSC-G, I don't think that should be OK.

mrawls:
I think this is good to go, so long as you give some thought to unit test augmentation to catch nonsensical dimension combos.

python/lsst/daf/butler/tests/_testRepo.py
kfindeisen: I remember it working *very* badly in developer guide examples; evolution of the underlying code turned them into gibberish. I'll go with @mrawls's idea; it'll have the added benefit of making high-level documentation more visible.

PR #467, opened 2021-01-28
DM-28527: Move special-casing of '*' and empty lists into globToRegex.

TallJimbo:
All of the command-line scripts that used globToRegex previously had to special-case the empty list it would return when '\*' or an empty input list was encountered.  But we already had a consistent sentinal value - `...` - that we used for all of these cases.  By making globToRegex return that when '*' or an empty list is encountered, we can simplify the scripts and fix DM-28527, which was due to the lack of this special-casing in queryCollections.

### DIFF
--- a/python/lsst/daf/butler/core/utils.py
+++ b/python/lsst/daf/butler/core/utils.py
@@ -52,10 +52,14 @@ from typing import (
     Pattern,
     Type,
     TypeVar,
+    TYPE_CHECKING,
     Union,
 )
 
 from lsst.utils import doImport
+
+if TYPE_CHECKING:
+    from ..registry.wildcards import Ellipsis, EllipsisType
 
 
 _LOG = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


python/lsst/daf/butler/core/utils.py
```diff
@@ -52,11 +52,15 @@
...
     Type,
     TypeVar,
+    TYPE_CHECKING,
     Union,
 )
 
 from lsst.utils import doImport
 
+if TYPE_CHECKING:
+    from ._registry.wildcards import Ellipsis, EllipsisType
```
timj:
```
from ..registry.wildcards...
```
I think

TallJimbo: Yup, fixing that now, and mypy complained as well.

python/lsst/daf/butler/script/queryCollections.py
```diff
@@ -54,12 +54,7 @@ def queryCollections(repo, glob, collection_type, chains):
         A table containing information about collections.
     """
     butler = Butler(repo)
-    expression = globToRegex(glob)
-    # Only pass expression to queryCollections if there is an expression to
-    # apply; otherwise let queryCollections use its default value.
-    kwargs: Dict[str, Any] = {}
-    if expression:
-        kwargs["expression"] = expression
+    kwargs: Dict[str, Any] = {"expression": globToRegex(glob)}
```
n8pease: since you're not conditionally setting `kwargs["expression"]` you can get rid of `kwargs` and just add the argument `expression=globToRegex(glob)` to your call to `queryCollections`.

python/lsst/daf/butler/script/queryDatasetTypes.py
```diff
@@ -57,12 +57,7 @@ def queryDatasetTypes(repo, verbose, glob, components):
         collection names.
     """
     butler = Butler(repo)
-    expression = globToRegex(glob)
-    # Only pass expression to queryDatasetTypes if there is an expression to
-    # apply; otherwise let queryDatasetTypes use its default value.
-    kwargs = {}
-    if expression:
-        kwargs["expression"] = expression
+    kwargs = {"expression": globToRegex(glob)}
```
n8pease: same as above; no need for `kwargs`

PR #468, opened 2021-02-04
DM-28665: fix regression in query flexibility

TallJimbo:


timj (approved):
You seem to have a failing action.

TallJimbo:
Yup, I fixed a typing problem in some new script code but forgot to change the mocked test accordingly.  Now fully fixed.

PR #469, opened 2021-02-04
DM-28674: Allow # to mean fragment in schemeless URIs

timj:


python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -531,9 +531,12 @@ def format(self, ref: DatasetRef) -> str:
             output = output + literal + format(value, format_spec)
 
         # Replace periods with underscores in the non-directory part to
-        # prevent file extension confusion.
+        # prevent file extension confusion. Also replace # in the non-dir
+        # part to avoid confusion with URI fragments
         head, tail = os.path.split(output)
-        output = os.path.join(head, tail.replace(".", "_"))
+        tail = tail.replace(".", "_")
+        tail = tail.replace("#", "HASH")
```
natelust, line 538: We probably dont care about speed here (at the number of times this will be called its negligible), and what is more readable probably depends on who reads it, but FWIW, `tail.replace(".", "_").replace("#", "HASH")` is a bit faster (I personally find it easier to grok, but others might not). This is more a side comment and not an explicit request to change.

natelust (approved):
All looks fine, minor comment just as a side note, looks fine to go in as is.

PR #470, opened 2021-02-05
DM-27473: Update Config.merge to support dict as well as Config

timj:


tests/test_config.py
```diff
@@ -244,6 +244,16 @@ def testOperators(self):
         c2[".a.b"] = 5
         self.assertNotEqual(c1, c2)
 
+    def testMerge(self):
+        c1 = Config({"a": 1, "c": 2})
+        c2 = Config({"a": 3, "b": "b"})
```
MichelleGower: Having single-letter values which match the single letter key names (especially later when mapping "b" to "c") means that one has to read the code more carefully.   It would just make the reading of this test easier if the string values were either small words or the other end of the alphabet.

tests/test_config.py
```diff
@@ -244,6 +244,16 @@ def testOperators(self):
...
+        c1.merge(c2)
+        self.assertEqual(c1, {"a": 1, "b": "b", "c": 2})
+
```
MichelleGower, line 255: Should this also check that c2 stays its original value (which should also care of making sure c1 and c2 aren't pointing to the same object)?

python/lsst/daf/butler/core/config.py
```diff
@@ -666,7 +666,12 @@ def merge(self, other):
         other : `dict` or `Config`
             Source of configuration:
         """
-        otherCopy = copy.deepcopy(other)
+        if not isinstance(other, collections.abc.Mapping):
+            raise ValueError(f"Can only merge a Mapping into a Config, not {type(other)}")
```
MichelleGower: Curious why not TypeError since variable other is of the wrong type.

MichelleGower (approved):
A question about Exception type and some comments about the test case.  Approved for merging.

PR #475, opened 2021-02-11
DM-28648: Allow datastore to trust caller and ignore registry

timj:


python/lsst/daf/butler/core/datasets/type.py
```diff
@@ -393,6 +393,28 @@ def componentTypeName(self, component: str) -> str:
...
+            The composite dataset type.
+
+        Raises
+        ------
+        RuntimeError
+            Raised if this dataset type is not a component dataset type.
+        """
+        if not self.isComponent():
+            raise RuntimeError(f"DatasetType {self.name} must be a component to form the composite")
+        name, _ = self.nameAndComponent()
```
ktlim: Maybe `composite_name` to distinguish from `self.name`?

python/lsst/daf/butler/core/datasets/type.py
```diff
@@ -393,6 +393,28 @@ def componentTypeName(self, component: str) -> str:
...
+        if self.parentStorageClass is None:
+            raise ValueError(f"Parent storage class is not set. Unable to create composite type from {self}")
```
ktlim: Why `{self}` here and simply `{self.name}` above?

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -452,6 +456,69 @@ def _can_remove_dataset_artifact(self, ref: DatasetIdRef, location: Location) ->
             return False
         return True
 
+    def _get_expected_dataset_locations_info(self, ref: DatasetRef) -> List[Tuple[Location,
+                                                                                  StoredFileInfo]]:
+        """Predict the location and related file information of the requested
+        dataset in the this datastore.
```
ktlim:
```suggestion
        dataset in this datastore.
```

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -548,7 +618,23 @@ def _prepare_for_put(self, inMemoryDataset: Any, ref: DatasetRef) -> Tuple[Locat
             The associated `DatasetType` is not handled by this datastore.
         """
         self._validate_put_parameters(inMemoryDataset, ref)
+        return self._determine_put_formatter_location(ref)
 
+    def _determine_put_formatter_location(self, ref: DatasetRef) -> Tuple[Location, Formatter]:
```
ktlim, line 623: Differences from `predictLocation()` (now deleted) include passing the dataId to `getFormatter()` and not trapping exceptions from `makeUpdatedLocation()`.  Are you sure that this can be used everywhere the other was being used?

ktlim (approved):

python/lsst/daf/butler/datastores/fileDatastore.py
timj: @TallJimbo was putting that dataId in on #476 and it was never really used.

I did notice that exception trap and I think that only triggers if a non-file Formatter is somehow used. The `butler.put()` code assumes that can never happen and I'm pretty sure the URI prediction code shouldn't trigger it.

python/lsst/daf/butler/core/datasets/type.py
timj: Not entirely sure. Will unify.

PR #476, opened 2021-02-11
DM-28583: require data ID when constructing a formatter

TallJimbo:


parejkoj (approved):
I'm going to trust you on this being a necessary change here. Fewer optional parameters is usually a good thing anyway.

PR #478, opened 2021-02-25
unpinned pinned versions in requirements.txt

tgoldina:
- Unpinned the versions of responses, urllib3, and mock in requirements.txt
- Removed pip downgrading step in build.yaml (build_and_test GitHub action)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -23,6 +23,11 @@ jobs:
       - name: Install sqlite
         run: sudo apt-get install sqlite libyaml-dev
 
+      - name: Update pip/wheel infrastructure
+        run: |
+          python -m pip install --upgrade pip
+          pip install wheel
+
       - name: Install postgresql (server)
         run: sudo apt install postgresql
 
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -26,13 +26,6 @@ jobs:
       - name: Install postgresql (server)
         run: sudo apt install postgresql
 
-      # pip is pinned because of urllib pinning because of moto bug with
-      # responses
-      - name: Update pip/wheel infrastructure
-        run: |
-          python -m pip install --upgrade pip==20.2.4
```
timj, line 33: I'd like to keep the upgrading of `pip` but remove the version pin completely.

requirements.txt
```diff
@@ -14,12 +14,11 @@ pandas >= 1.0
...
-responses == 0.12.0
-urllib3 == 1.25
+responses >= 0.12.0
+urllib3 >= 1.25.10
 
-# We don't use mock directly, but something else (moto?) does and is
-# incompatible with mock == 4.0.3.
-mock == 4.0.2
+# We don't use mock directly, but something else (moto?) does
+mock >= 4.0.2
```
timj: I think we can remove the mock line completely.

timj (approved):
Looks great. Thanks for looking at this. Minor comments only.

PR #479, opened 2021-02-27
DM-28964: add incremental progress reporting to daf_butler

TallJimbo:
This creates an abstract interface to allow at least daf_butler and some downstream middleware code to report progress in a way that would let higher-level tools show progress bars.  It also makes a start at letting click-based command-line tools actually show those progress bars, but that's disabled by default as I'm not sure how it interacts with logging.

I haven't actually instrumented too many points in `daf_butler` - just a few known hot spots, like datastore transfers and loops over `expandDataId` in` Registry` - and there may not be much more we can do there, because we've actually done a pretty good job of vectorizing things, and that means a lot of slow operations are single commands from our perspective (and there's no way to e.g. get incremental progress on a query from the database).  But if the middleware pundits are happy with the approach I've taken here, I'll go ahead and instrument at least `BaseSkyMap.register`, `RawIngestTask`, and `ConvertRepoTask` (on this ticket), where we certainly have lots of loops that would benefit from this kind of reporting.

### DIFF
--- a/python/lsst/daf/butler/core/progress.py
+++ b/python/lsst/daf/butler/core/progress.py
@@ -1,0 +1,367 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import annotations
+
+__all__ = ("Progress", "ProgressBar", "ProgressHandler")
+
+from abc import ABC, abstractmethod
+from contextlib import contextmanager
+import logging
+from typing import (
+    ClassVar,
+    Collection,
+    ContextManager,
+    Generator,
+    Iterable,
+    Iterator,
+    Optional,
+    Protocol,
+    Tuple,
+    TypeVar,
+)
+
+_T = TypeVar("_T", covariant=True)
+_K = TypeVar("_K")
+_V = TypeVar("_V", bound=Collection)
+
+
+class ProgressBar(Protocol, Iterable[_T]):
+    """A structural interface for progress bars that wrap iterables.
+
+    An object conforming to this interface can be obtained from the
+    `Progress.bar` method.
+
+    Notes
+    -----
+    This interface is intentionally defined as the intersection of the progress
+    bar objects returned by the ``click`` and ``tqdm`` packages, allowing those
+    to directly satisfy code that uses this interface.
+    """
+
+    def update(self, n: int = 1) -> None:
+        """Increment the progress bar by the given amount.
+
+        Parameters
+        ----------
+        n : `int`, optional
+            Increment the progress bar by this many steps (defaults to ``1``).
+            Note that this is a relative increment, not an absolute progress
+            value.
+        """
+        pass
+
+
+class Progress:
+    """Public interface for reporting incremental progress in the butler and
+    related tools.
+
+    This class automatically creates progress bars (or not) depending on
+    whether a handle (see `ProgressHandler`) has been installed and the given
+    name and level are enabled.  When progress reporting is not enabled, it
+    returns dummy objects that can be used just like progress bars by calling
+    code.
+
+    Parameters
+    ----------
+    name : `str`
+        Name of the process whose progress is being reported.  This is in
+        general the name of a group of progress bars, not necessarily a single
+        one, and it should have the same form as a logger name.
+    level : `int`, optional
+        A `logging` level value (defaults to `logging.INFO`).  Progress
+        reporting is enabled if a logger with ``name`` is enabled for this
+        level, and a `ProgressHandler` has been installed.
+
+    Notes
+    -----
+    The progress system inspects the level for a name using the Python built-in
+    `logging` module, and may not respect level-setting done via the
+    ``lsst.log`` interface.  But while `logging` may be necessary to control
+    progress bar visibility, the progress system can still be used together
+    with either system for actual logging.
+    """
+
+    def __init__(self, name: str, level: int = logging.INFO) -> None:
+        self._name = name
+        self._level = level
+
+    # The active handler is held in a ContextVar to isolate unit tests run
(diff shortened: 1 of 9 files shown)


TallJimbo:
One really obvious flaw here is that the tests are failing, and it's because the global progress-handler state I've introduced interacts poorly with `pytest-xdist`'s multithreading.  I _thought_ had fixed that (and indeed, cannot reproduce this locally anymore) by using `ContextVar` to wrap the global, but maybe I misunderstood how that interacts with threading or how `pytest-xdist` actually works.  In any case, I'm at a bit of loss for how best to isolate this test from all of the others, and I'm hoping someone has a suggestion.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -882,7 +883,8 @@ def _prepIngest(self, *datasets: FileDataset, transfer: Optional[str] = None) ->
     def _finishIngest(self, prepData: Datastore.IngestPrepData, *, transfer: Optional[str] = None) -> None:
         # Docstring inherited from Datastore._finishIngest.
         refsAndInfos = []
-        for dataset in prepData.datasets:
+        progress = Progress("daf.butler.datastores.FileDatastore.ingest")
```
timj: in daf_butler we always use `lsst.daf.butler` in logger names because we were hoping that people outside of LSST would be interested and where it might be confusing for the `lsst` to be implicit.

timj (commented):
The user interface looks fine. Given this is a "is this okay to continue" review request I'm not looking in detail. As for xdist, it forks a new python proces and is guaranteed to run setUpClass and setUp. It probably also runs setup_module. It's all meant to be distinct so I'm not sure what's going on.

TallJimbo:
> As for xdist, it forks a new python process and is guaranteed to run setUpClass and setUp. It probably also runs setup_module. It's all meant to be distinct so I'm not sure what's going on.

Huh, that is at least not what I thought it did, so it will help me debug.  It seems like it would have to be running _other_ tests in the same process, between one `TestCase`'s `setUp` and test invocation.  But I should look more closely; maybe I missed a way this could happen just because I'm missing a `tearDown`.

timj:
I'm 99.9% sure that pytest guarantees that `setUp` runs immediately before the test is executed and then `tearDown` runs immediately after the test has completed.  All the problems I've had with xdist relate to access to external resources like files.

python/lsst/daf/butler/core/progress.py
```diff
@@ -0,0 +1,396 @@
...
+    code.
+
+    Parameters
+    ----------
+    name : `str`
+        Name of the process whose progress is being reported.  This is in
+        general the name of a group of progress bars, not necessarily as single
+        one, and it should have the same form as a logger names: the
+        fully-qualified name of the class, method, or function, minus the
+        initial "lsst" (and without any "implemntation detail" packages or
```
timj: Minus the inital "lsst" if outside of daf_butler....

python/lsst/daf/butler/core/progress.py
```diff
@@ -0,0 +1,396 @@
...
+        """Check whether this process should report progress.
+
+        Returns
+        -------
+        enabled : `bool`
+            `True` if there is a `ProgressHandler` set and a logger with the
+            same name and level as ``self`` is enabled.
+        """
+        if self._active_handler is not None:
+            logger = logging.getLogger(self._name)
```
timj, line 139: Does this work for `obs_base` which uses `lsst.log` and not `logging`?

python/lsst/daf/butler/core/progress.py
```diff
@@ -0,0 +1,396 @@
...
+                return True
+        return False
+
+    def at(self, level: int) -> Progress:
+        """Return a copy of this progress interface with a different level.
+
+        Parameters
+        ----------
+        level : `int`, optional
+            A `logging` level value (defaults to `logging.INFO`).  Progress
```
timj: This is a required parameter with no default isn't it?

python/lsst/daf/butler/core/progress.py
```diff
@@ -0,0 +1,396 @@
...
+            bar is iterable is handler-defined, but it may be updated manually.
+        desc: `str`, optional
+            A user-friendly description for this progress bar; usually appears
+            next to it.  If not provided, ``self.name`` is used (which is not
+            usually a user-friendly string, but may be appropriate for
+            debug-level progress).
+        total : `int`, optional
+            The total number of steps in this progress bar.  If not provided,
+            ``len(iterable)`` is used.  If that does not work, whether the
+            progress bar works at all is handled-defined, and hence this mode
```
timj: Typo "handled-defined"

python/lsst/daf/butler/core/progress.py
```diff
@@ -0,0 +1,396 @@
...
+            An arbitrary Python iterable to iterate over.
+        desc: `str`, optional
+            A user-friendly description for this progress bar; usually appears
+            next to it.  If not provided, ``self.name`` is used (which is not
+            usually a user-friendly string, but may be appropriate for
+            debug-level progress).
+        total : `int`, optional
+            The total number of steps in this progress bar.  If not provided,
+            ``len(iterable)`` is used.  If that does not work, whether the
+            progress bar works at all is handled-defined, and hence this mode
```
timj: handled-defined

python/lsst/daf/butler/core/progress.py
```diff
@@ -0,0 +1,396 @@
...
+            An arbitrary Python iterable that will be iterated over when the
+            returned `ProgressBar` is.  If `None`, whether the progress bar is
+            iterable is handler-defined, but it may be updated manually.
+        desc: `str`
+            A user-friendly description for this progress bar; usually appears
+            next to it
+        total : `int` or `None`
+            The total number of steps in this progress bar.  If `None``,
+            ``len(iterable)`` should be used.  If that does not work, whether
+            the progress bar works at all is handled-defined.
```
timj: handled-defined

tests/test_progress.py
```diff
@@ -0,0 +1,210 @@
...
+        seen = {}
+        for key, chunk in self.progress.iter_item_chunks(mapping.items()):
+            seen[key] = chunk
+        self.assertEqual(seen, mapping)
+        self.assertEqual(MockProgressBar.last.reported, [0, 2, 5])
+
+
+if __name__ == "__main__":
+    unittest.main()
+    print(Progress._setters)
```
timj: Maybe say this is for debugging? pytest won't ever see it.

timj (approved):
Looks okay. My main comment is how this interacts with lsst.log but since you use it in obs_base it seems like it's all working fine.

tests/test_progress.py
TallJimbo: I had actually intended to remove it.  Will do that.

python/lsst/daf/butler/core/progress.py
TallJimbo: I'll just remove the parenthetical.

python/lsst/daf/butler/core/progress.py
TallJimbo: I'm not sure if it works to set the level for a "logger" that really corresponds to a progress bar via `lsst.log`; I used `logging` to do that (while still using `lsst.log` for _actual_ logs).  I think that's okay, because I'm really just using the logging system as a way to get a global mapping of name->level, but it probably is worth documenting.

PR #481, opened 2021-03-01
DM-28717: Allow for Stamps formatter to handle bbox

MorganSchmitz:
One of four PRs for [DM-28717](https://jira.lsstcorp.org/browse/DM-28717), the others being afw, obs_base and meas_algorithms.

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -186,8 +186,12 @@ storageClasses:
     pytype: lsst.verify.Measurement
   Stamps:
     pytype: lsst.meas.algorithms.Stamps
+    parameters:
+      - bbox
   BrightStarStamps:
     pytype: lsst.meas.algorithms.brightStarStamps.BrightStarStamps
+    parameters:
+      - bbox
   AstropyTable:
     pytype: astropy.table.Table
   AstropyQTable:


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -184,8 +184,14 @@ storageClasses:
     pytype: lsst.verify.Measurement
   Stamps:
     pytype: lsst.meas.algorithms.Stamps
+    parameters:
+      - bbox
+      - origin
```
SimonKrughoff: The reader seems to only do something with the `bbox` parameter.  What am I missing?

python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -184,8 +184,14 @@ storageClasses:
...
   BrightStarStamps:
     pytype: lsst.meas.algorithms.brightStarStamps.BrightStarStamps
+    parameters:
+      - bbox
+      - origin
```
SimonKrughoff: Same as above

SimonKrughoff (approved):
Just a clarifying question

python/lsst/daf/butler/configs/storageClasses.yaml
MorganSchmitz: Hah! Thanks for asking. I'd just stupidly copied from Image. Removed and tested!

PR #482, opened 2021-03-02
DM-26302: Avoid long-lived connections in Database

andy-slac:
Database connections are now short-lived by default, they are created
either for the duration of a query or transaction. Additionally `session()`
context can be used for persistent connection that does not start
transaction, this is needed to manage lifetime of temporary tables.
Re-enabled connection pool use in database engines, this will need special
care in fork-based mutiprocessing.

### DIFF
--- a/python/lsst/daf/butler/registry/interfaces/_database.py
+++ b/python/lsst/daf/butler/registry/interfaces/_database.py
@@ -170,6 +170,100 @@ class StaticTablesContext:
         self._initializers.append(initializer)
 
 
+class Session:
+    """Class representing a persistent connection to a database.
+
+    Parameters
+    ----------
+    db : `Database`
+        Database instance.
+
+    Notes
+    -----
+    Instances of Session class should not be created by client code, the
+    `Database.session` should be used to create context for a session::
+
+        with db.session() as session:
+            session.method()
+            db.method()
+
+    In current implementation sessions can be nested and transactions can be
+    nested within a session. All nested sessions and transaction share the
+    same database connection.
+
+    Session class represents a limited subset of database API that requires
+    persistent connection to a database (e.g. temporary tables which have
+    lifetime of a session). Potentially most of the database API could be
+    associated with a Session class.
+    """
+    def __init__(self, db: Database):
+        self._db = db
+
+    def makeTemporaryTable(self, spec: ddl.TableSpec, name: Optional[str] = None) -> sqlalchemy.schema.Table:
+        """Create a temporary table.
+
+        Parameters
+        ----------
+        spec : `TableSpec`
+            Specification for the table.
+        name : `str`, optional
+            A unique (within this session/connetion) name for the table.
+            Subclasses may override to modify the actual name used.  If not
+            provided, a unique name will be generated.
+
+        Returns
+        -------
+        table : `sqlalchemy.schema.Table`
+            SQLAlchemy representation of the table.
+
+        Notes
+        -----
+        Temporary tables may be created, dropped, and written to even in
+        read-only databases - at least according to the Python-level
+        protections in the `Database` classes.  Server permissions may say
+        otherwise, but in that case they probably need to be modified to
+        support the full range of expected read-only butler behavior.
+
+        Temporary table rows are guaranteed to be dropped when a connection is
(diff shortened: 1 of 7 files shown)


python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -170,6 +170,100 @@ def addInitializer(self, initializer: Callable[[Database], None]) -> None:
...
+    nested within a session. All nested sessions and transaction share the
+    same database connection.
+
+    Session class represents a limited subset of database API that requires
+    persistent connection to a database (e.g. temporary tables which have
+    lifetime of a session). Potentially most of the database API could be
+    associated with a Session class.
+    """
+    def __init__(self, db: Database):
+        self._db = db
```
TallJimbo, line 200: Could the `Session` object hold its `Connection`, instead of having `Database` have an optional `Connection` that's really associated with a particular `Session`?  I bet that would avoid some `None`-checking, and it'd be a tiny step forward for thread/async safety (in that if we resolved other issues, there could be multiple parallel sessions associated with a single `Database`).

Is the problem that we would have then have to move the `transaction` interface here, too, and that's more disruptive (i.e. it's a big step towards your "solution 3")?

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -170,6 +170,100 @@ def addInitializer(self, initializer: Callable[[Database], None]) -> None:
...
+    """Class representing a persistent connection to a database.
+
+    Parameters
+    ----------
+    db : `Database`
+        Database instance.
+
+    Notes
+    -----
+    Instances of Session class should not be created by client code, the
```
TallJimbo:
```suggestion
    Instances of Session class should not be created by client code;
```

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -170,6 +170,100 @@ def addInitializer(self, initializer: Callable[[Database], None]) -> None:
...
+    `Database.session` should be used to create context for a session::
+
+        with db.session() as session:
+            session.method()
+            db.method()
+
+    In current implementation sessions can be nested and transactions can be
```
TallJimbo:
```suggestion
    In the current implementation sessions can be nested and transactions can be
```

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -831,7 +950,10 @@ def ensureTableExists(self, name: str, spec: ddl.TableSpec) -> sqlalchemy.schema
 
         Subclasses may override this method, but usually should not need to.
         """
-        assert not self._connection.in_transaction(), "Table creation interrupts transactions."
+        # TODO: if _engine is used to make a table then it uses separate
+        # connection and should not interfere with current transaction
+        assert self._session_connection is None or not self._session_connection.in_transaction(), \
+            "Table creation interrupts transactions."
```
TallJimbo, line 956: FWIW, it may be that this already wasn't a problem; it was only with Oracle where I was certain that DDL statements would break transactions (and it may be that with Oracle, using a separate connection for DDL doesn't save you; I'm not sure).  I left this assert in after we dropped Oracle because it seemed like a good idea to leave room for future databases (maybe even Oracle again) where that might be a problem, and without the check it would be very easy to write important code that would be hard to use with such a database.

TallJimbo (approved):

python/lsst/daf/butler/registry/interfaces/_database.py
andy-slac: I think that would require full "solution 3", basically any method that needs a connection should be moved into Session class. I agree that having Connection here is the most natural and safer way which reduces unintentional sharing but it would be a big change in API and require manual session management in all methods that did not do it before.

python/lsst/daf/butler/registry/interfaces/_database.py
andy-slac: I beleive that in Postgres CREATE TABLE is transactional, but I know at least one other RDBMS (mysql) that does not support it, in mysql CREATE TABLE implicitly commits transaction. In general DDL statements are not required to be transaction-safe so it is probably worth keeping this assert just in case.

PR #484, opened 2021-03-03
DM-29011: Add support for ButlerURI.walk()

timj:
Works for S3 and File URIs.

### DIFF
--- a/python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
+++ b/python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
@@ -830,7 +830,7 @@ class ButlerURI:
             The regex to use when searching for files within directories.
             By default returns all the found files.
         grouped : `bool`, optional
-            If `True` the results will be grouped by directly and each
+            If `True` the results will be grouped by directory and each
             yielded value will be an iterator over URIs. If `False` each
             URI will be returned separately.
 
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/core/_butlerUri/s3.py
```diff
@@ -226,3 +229,75 @@ def transfer_from(self, src: ButlerURI, transfer: str = "copy",
...
+        if not self.isdir():
+            raise ValueError(f"Can not walk a non-directory URI: {self}")
+
+        if isinstance(file_filter, str):
+            file_filter = re.compile(file_filter)
+
+        s3_paginator = self.client.get_paginator('list_objects_v2')
+
+        # Limit each query to a single "directory" to match os.walk
+        # We could download all keys at once with no delimiter and work
```
ktlim, line 265: Because it's paginated, if you're going to scan them all eventually anyway, I think you'd be better off doing the simple version — except that you then wouldn't have the same result ordering as `os.walk()`.

tests/test_uri.py
```diff
@@ -343,6 +395,40 @@ def testTransfer(self):
 
         dest.transfer_from(src, transfer="copy", overwrite=True)
 
+    def testWalk(self):
+        """Test that we can list an S3 bucket"""
+        # Files we want to create
+        expected = ("a/x.txt", "a/y.txt", "a/z.json", "a/b/w.txt", "a/b/c/d/v.json")
```
ktlim, line 409: I think you should have at least one test that extends over more than one page, which unfortunately means a lot of files, given the default max of 1000 keys.

ktlim (approved):

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
```diff
@@ -771,3 +788,81 @@ def transfer_from(self, src: ButlerURI, transfer: str,
...
+        Parameters
+        ----------
+        candidates : iterable [`str` or `ButlerURI`]
+            The files to return and directories in which to look for files to
+            return.
+        file_filter : `str`, optional
+            The regex to use when searching for files within directories.
+            By default returns all the found files.
+        grouped : `bool`, optional
+            If `True` the results will be grouped by directly and each
```
ktlim:
```suggestion
            If `True` the results will be grouped by directory and each
```

PR #485, opened 2021-03-05
DM-29073: Make ButlerURI immutable

timj:


python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
```diff
@@ -674,30 +710,84 @@ def size(self) -> int:
...
-        return type(self)(str(self), forceAbsolute=self.isabs())
+        return self
 
     def __deepcopy__(self, memo: Any) -> ButlerURI:
+        """Deepcopy the object.
+
+        Object is immutable so copy can return itself.
+        """
         # Implement here because the __new__ method confuses things
         return self.__copy__()
```
ktlim: ???  Mismatch with docstring.

timj: Technically the docstring is correct because copy does return self... but given that you called me out on it I will fix it.

timj:
@TallJimbo this PR looks bigger than it really is because of all the docstring fixes. I'm hoping you can skip over those relatively quickly.

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
```diff
@@ -160,6 +160,18 @@ def __new__(cls, uri: Union[str, urllib.parse.ParseResult, ButlerURI, Path],
...
+            # If we are being instantiated with a subclass, rather than
+            # ButlerURI, ensure that that subclass is used directly.
+            # This could lead to inconsistencies if this constructor
+            # is used externally outside of the ButlerURI.replace() method.
+            #   ButlerS3URI(urllib.parse.urlparse("file://a/b.txt"))
+            # will be a problem.
+            # This is needed to prevent a schemeless absolute URI become
+            # a file URI unexpectedly when calling updatedFile or
+            # updatedExtension
+            if cls != ButlerURI:
```
TallJimbo: I think this can safely be a `is not`.  Don't know if that means it should be.

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
```diff
@@ -373,46 +373,60 @@ def parent(self) -> ButlerURI:
...
-    def replace(self, **kwargs: Any) -> ButlerURI:
+    def replace(self, forceDirectory: bool = False, **kwargs: Any) -> ButlerURI:
         """Replace components in a URI with new values and return a new
         instance.
 
+        Parameters
+        ----------
+        forceDirectory : `bool`
+            Parameter passed to ButlerURI constructor to force this
+            new URI to be dir-like.
```
TallJimbo, line 403: Document `**kwargs`.

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
```diff
@@ -421,9 +435,15 @@ def updateExtension(self, ext: Optional[str]) -> None:
         ext : `str` or `None`
             New extension. If an empty string is given any extension will
             be removed. If `None` is given there will be no change.
+
+        Returns
+        -------
+        updated : `ButlerURI`
+            URI with the specified extension. Can return itself if
+            no extension was specified.
```
TallJimbo, line 463: Would it be worthwhile to optimize the case where the extension is already the given one, too?

python/lsst/daf/butler/core/config.py
```diff
@@ -118,8 +118,7 @@ def extractFile(self, filename):
         if requesteduri.scheme:
             fileuri = requesteduri
         else:
-            fileuri = copy.copy(self._root)
-            fileuri.updateFile(filename)
+            fileuri = copy.copy(self._root).updatedFile(filename)
```
TallJimbo: It looks like `self._root` is a `ButlerURI`, so this copy is now a no-op and can be removed.

TallJimbo (approved):

PR #486, opened 2021-03-08
DM-29106: allow Datastore to "forget" datasets and add a new Butler method to use it.

TallJimbo:
This is nominally ready for review, but it raises questions about the future that we should probably resolve before merging:

 - Is it okay that I'm introducing `Butler.removeRuns` and sort of abandoning `Butler.pruneCollection` with this ticket (see https://community.lsst.org/t/simplifying-and-limiting-deletion-in-gen3-butler-repos/4808)?

 - Is it sufficiently easy/important to make forget-rather-than-unstore automatic for absolute-URL ingests that I should just do it here, or is that a separate ticket?

### DIFF
--- a/python/lsst/daf/butler/datastores/chainedDatastore.py
+++ b/python/lsst/daf/butler/datastores/chainedDatastore.py
@@ -551,6 +551,10 @@ class ChainedDatastore(Datastore):
         self.trash(ref, ignore_errors=False)
         self.emptyTrash(ignore_errors=False)
 
+    def forget(self, refs: Iterable[DatasetRef]) -> None:
+        for datastore in tuple(self.datastores):
+            datastore.forget(refs)
+
     def trash(self, ref: DatasetRef, ignore_errors: bool = True) -> None:
         log.debug("Trashing %s", ref)
 
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -1573,6 +1573,13 @@ class FileDatastore(GenericBaseDatastore):
                             f"of {self.name}"
                         ) from e
 
+    @transactional
+    def forget(self, refs: Iterable[DatasetRef]) -> None:
+        # Docstring inherited.
+        refs = list(refs)
+        self.bridge.forget(refs)
+        self._table.delete(["dataset_id"], *[{"dataset_id": ref.getCheckedId()} for ref in refs])
+
     def validateConfiguration(self, entities: Iterable[Union[DatasetRef, DatasetType, StorageClass]],
                               logFailures: bool = False) -> None:
         """Validate some of the configuration for this datastore.
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -328,6 +328,8 @@ class FileDatastore(GenericBaseDatastore):
         location : `Location`
             Location of the artifact associated with this datastore.
         """
+        if location.pathInStore.isabs():
+            raise RuntimeError(f"Cannot delete artifact with absolute uri {location.uri}.")
         log.debug("Deleting file: %s", location.uri)
         location.uri.remove()
         log.debug("Successfully deleted file: %s", location.uri)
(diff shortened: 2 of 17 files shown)


python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1568,7 +1573,17 @@ def emptyTrash(self, ignore_errors: bool = True) -> None:
                                     ref.id, location.uri, self.name, e)
                         continue
                     else:
-                        raise FileNotFoundError(err_msg)
```
timj, line 1571: Python not scoping variables lexically really bites me sometimes...

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1568,7 +1573,17 @@ def emptyTrash(self, ignore_errors: bool = True) -> None:
...
+                        raise FileNotFoundError(
+                            f"Error removing dataset {ref.id} ({location.uri}) from internal registry "
+                            f"of {self.name}"
+                        ) from e
+
+    @transactional
+    def forget(self, refs: Iterable[DatasetRef]) -> None:
+        # Docstring inherited.
+        refs = list(refs)
+        self._bridge.forget(refs)
```
timj: shouldn't this be `self.bridge` rather than using the internal property?

timj (commented):
Datastore part looks okay.

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -551,6 +551,10 @@ def remove(self, ref: DatasetRef) -> None:
         self.trash(ref, ignore_errors=False)
         self.emptyTrash(ignore_errors=False)
 
+    def forget(self, refs: Iterable[DatasetRef]) -> None:
+        for datastore in self.datastores:
```
andy-slac: Maybe add:

    refs = list(refs)

before this line to avoid problems with true iterators?

python/lsst/daf/butler/registry/bridge/ephemeral.py
```diff
@@ -54,6 +54,9 @@ def insert(self, refs: Iterable[DatasetIdRef]) -> None:
         # Docstring inherited from DatastoreRegistryBridge
         self._datasetIds.update(ref.getCheckedId() for ref in refs)
 
+    def forget(self, refs: Iterable[DatasetIdRef]) -> None:
+        self._datasetIds.difference_update({ref.id for ref in refs})
```
andy-slac: I think there is no need to make a set first, `difference_update` accepts any iterable.

python/lsst/daf/butler/tests/_dummyRegistry.py
```diff
@@ -63,11 +63,11 @@ def fetch(self, **where: Any) -> Iterator[dict]:
             if all(d[k] == v for k, v in where.items()):
                 yield d
 
-    def delete(self, **where: Any):
+    def delete(self, columns: Iterable[str], *rows: dict):
         # Docstring inherited from OpaqueTableStorage.
         kept = []
         for d in self._rows:
-            if not all(d[k] == v for k, v in where.items()):
+            if not all(d[k] == v for where in rows for k, v in where.items()):
```
andy-slac: I'm not sure that I read it correctly or that it means what it is supposed to mean?

andy-slac (approved):
Looks good, one comment about _dummyRegistry change, but maybe because it's late night here 😴

python/lsst/daf/butler/tests/_dummyRegistry.py
TallJimbo: It think it's correct, but I should know better than to use nested comprehensions because the ordering is very much the opposite of how it reads.  I'll rewrite it; even I can't read what it was supposed to do completely anymore.

andy-slac: But does not it also say that row should be removed from `self._rows` only when it matches _all_ rows in `rows`?

TallJimbo: Ah, yeah, indeed, it was completely wrong, and we got away with it because the existing test code only deleted one thing at a time because that's all it could do before.  Working on a very different and much more readable version now.

PR #487, opened 2021-03-08
DM-28857 sort collections before pruning in `butler prune-collection`

n8pease:
For @TallJimbo, seeing if this approach will work. I'm leaving more notes in the ticket.

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1204,7 +1204,8 @@ class Butler:
             # Point of no return for removing artifacts
             self.datastore.emptyTrash()
 
-    def pruneCollection(self, name: str, purge: bool = False, unstore: bool = False) -> None:
+    def pruneCollection(self, name: str, purge: bool = False, unstore: bool = False,
+                        unlink: Optional[List[str]] = None) -> None:
         """Remove a collection and possibly prune datasets within it.
 
         Parameters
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -1156,7 +1156,8 @@ def datasetExists(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
         ref = self._findDatasetRef(datasetRefOrType, dataId, collections=collections, **kwds)
         return self.datastore.exists(ref)
 
-    def pruneCollection(self, name: str, purge: bool = False, unstore: bool = False) -> None:
+    def pruneCollection(self, name: str, purge: bool = False, unstore: bool = False,
+                        unlink: Optional[str] = None) -> None:
```
```suggestion
                        unlink: Optional[List[str]] = None) -> None:
```

python/lsst/daf/butler/_butler.py
```diff
@@ -1176,6 +1177,9 @@ def pruneCollection(self, name: str, purge: bool = False, unstore: bool = False)
         unstore: `bool`, optional
             If `True`, remove all datasets in the collection from all
             datastores in which they appear.
+        unlink: `list` [`str`]
```
```suggestion
        unlink: `list` [`str`], optional
```

python/lsst/daf/butler/_butler.py
```diff
@@ -1198,6 +1202,19 @@ def pruneCollection(self, name: str, purge: bool = False, unstore: bool = False)
...
+            chain = list(self.registry.getCollectionChain(parent))
+            try:
+                chain.remove(name)
+            except ValueError as e:
+                raise RuntimeError(f"{name} is not a child of {parent}") from e
+            self.registry.setCollectionChain(parent, chain)
+
+        if (unlink):
+            for parent in unlink:
+                remove(name, parent)
```
TallJimbo: I think we should move this down into the transaction block so the whole thing stays exception-safe.

TallJimbo (approved):

PR #488, opened 2021-03-12
DM-29203: Add option to flatten when setting collection chain.

TallJimbo:


python/lsst/daf/butler/registry/interfaces/_collections.py
```diff
@@ -165,6 +165,9 @@ def update(self, manager: CollectionManager, children: CollectionSearch) -> None
         children : `CollectionSearch`
             A collection search path that should be resolved to set the child
             collections of this chain.
+        flatten : `bool`, optional
+            If `True` (`False` is default), recursively flatten out any nested
```
andy-slac: There is no default for this method (and `optional` on previous line).

andy-slac (approved):
Looks good, one comment for docstring.

PR #489, opened 2021-03-15
DM-25965: Add ButlerURI.abspath() method

timj:
This is now very lightweight since URIs are immutable.
Forces scheme-less URIs to file URIs.

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -84,7 +84,7 @@ from .core.repoRelocation import BUTLER_ROOT_TAG
 from .core.utils import transactional, getClassOf
 from ._deferredDatasetHandle import DeferredDatasetHandle
 from ._butlerConfig import ButlerConfig
-from .registry import Registry, RegistryConfig, RegistryDefaults, CollectionType
+from .registry import Registry, RegistryConfig, RegistryDefaults, CollectionType, ConflictingDefinitionError
 from .registry.wildcards import CollectionSearch
 from .transfers import RepoExportContext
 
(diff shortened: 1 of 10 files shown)


python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
```diff
@@ -621,6 +621,17 @@ def isabs(self) -> bool:
...
 
+    def abspath(self) -> ButlerURI:
+        """Return URI using an absolute path.
+
+        Returns
+        -------
+        abs : `ButlerURI`
+            Absolute URI. For non-schemeless URIs this always returns itself.
+            Schemeless URIs are upgraded to file URIs.
+        """
```
TallJimbo, line 632: Is there anything you could `assert` on here to make sure a new derived class that needs to implement this method doesn't forget to?

python/lsst/daf/butler/_butler.py
```diff
@@ -1433,6 +1433,10 @@ def ingest(self, *datasets: FileDataset, transfer: Optional[str] = "auto", run:
             # associated with `dataset`.
             resolvedRefs: List[DatasetRef] = []
             for ref in dataset.refs:
+                if ref.dataId in groupedData[ref.datasetType]:
+                    raise RuntimeError(f"Ingest conflict. Dataset {dataset.path} has same DataId as "
+                                       f"other ingest dataset"
+                                       f" {groupedData[ref.datasetType][ref.dataId][0].path} ({ref.dataId})")
```
TallJimbo: `daf.butler.registry.ConflictingDefinitionError` might be a better exception type to raise here.

TallJimbo (approved):

TallJimbo:
After starting to look at `obs_base`, it occurred to me that `daf.butler.FileDataset` and should probably be changed to take `ButlerURI` as well, and I see that it hasn't been; I assume that means that right now we're taking `ButlerURI` objects in `obs_base`, stringifying them to pass to `Butler.ingest`, and then making them back into `ButlerURI` objects inside `Datastore`?

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
timj: All URIs are by definition meant to be absolute so this should always be the right answer. It's only schemeless that needs to be upgraded to absolute `file` URI (`file` is always absolute).

timj:
I fixed the FileDataset declaration to allow ButlerURI. I don't think we do stringify them as part of ingest so I don't think it's a problem. There are a couple of places where we call ButlerURI but that's lightweight now since it's immutable.

PR #491, opened 2021-03-17
DM-29252: Update code for sqlalchemy 1.4 compatibility

andy-slac:
Couple of small changes:
- URL class became immutable and needs special method to update it
- Table.columns attribute is not a simple list and cannot be
  added/joined, need to make a list out of it first


timj (approved):
Looks reasonable to me.

PR #492, opened 2021-03-23
DM-29288: Pass dataId to composite assembly

timj:
This is required to enable filter label fixup


timj:
We have decided against this. The expectation is that composite assembly is too late and formatters (even for components) should have handled whatever needed to be handled from the dataIds.

PR #495, opened 2021-03-25
DM-29414: Fix S3 size for missing file and add tests

timj:


DinoBektesevic (approved):
Looks good to me. Thanks.

PR #496, opened 2021-03-26
DM-29383: Implement basic caching for remote datasets in Datastore

timj:
Currently there is no cache expiry.

### DIFF
--- a/python/lsst/daf/butler/configs/datastores/fileDatastore.yaml
+++ b/python/lsst/daf/butler/configs/datastores/fileDatastore.yaml
@@ -18,7 +18,7 @@ datastore:
     # Instantiating a Butler from another Butler could propagate temporary
     # location. This can be left out or null used
     # to indicate a temporary directory.
-    root: null
+    root: /Volumes/MediaHD/Users/timj/work/lsstsw/build/daf_butler_test/cache
     # Use a dict over list to simplify merging logic.
     # Later could change it to be the timescale for caching:
     #   - Quantum (clear at end of quantum)
--- a/python/lsst/daf/butler/configs/datastores/fileDatastore.yaml
+++ b/python/lsst/daf/butler/configs/datastores/fileDatastore.yaml
@@ -11,3 +11,21 @@ datastore:
     physical_filter+detector+exposure: "{run:/}/{datasetType}.{component:?}/{exposure.day_obs}/{exposure.obs_id}/{datasetType}_{component:?}_{instrument:?}_{physical_filter}_{exposure.obs_id}_{detector.full_name}_{run}"
   formatters: !include formatters.yaml
   composites: !include composites.yaml
+  cached:
+    # Explicit location of a cache directory root that can be shared
+    # between multiple Butler instantiations. Without this the Butler
+    # will use a new temporary location each time.
+    # Instantiating a Butler from another Butler could propagate temporary
+    # location. This can be left out or null used
+    # to indicate a temporary directory.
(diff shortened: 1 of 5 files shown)


python/lsst/daf/butler/configs/datastores/fileDatastore.yaml
```diff
@@ -11,3 +11,21 @@ datastore:
...
   formatters: !include formatters.yaml
   composites: !include composites.yaml
+  cached:
+    # Explicit location of a cache directory root that can be shared
+    # between multiple Butler instantiations. Without this the Butler
+    # will use a new temporary location each time.
+    # Instantiating a Butler from another Butler could propagate temporary
+    # location. This can be left out or the magic value of "none" used
+    # to indicate a temporary directory.
+    root: none
```
TallJimbo: I think YAML uses `null`, and I bet pyyaml would convert that to `None` for you.

python/lsst/daf/butler/core/datastoreCacheManager.py
```diff
@@ -0,0 +1,281 @@
...
+    from .storageClass import StorageClass
+    from .configSupport import LookupKey
+
+log = logging.getLogger(__name__)
+
+
+class DatastoreCacheManagerConfig(ConfigSubset):
+    """Configuration information for `DatastoreCacheManager`."""
+
+    pass
```
TallJimbo: I think a docstring is enough that you don't need `pass` here.

python/lsst/daf/butler/core/datastoreCacheManager.py
```diff
@@ -0,0 +1,281 @@
...
+        This is relevant when reading or writing.
+
+        Parameters
+        ----------
+        entity : `StorageClass` or `DatasetType` or `DatasetRef`
+            Thing to test against the configuration. The ``name`` property
+            is used to determine a match.  A `DatasetType` will first check
+            its name, before checking its `StorageClass`.  If there are no
+            matches the default will be returned. If the associated
+            `StorageClass` is not a composite, will always return `False`.
```
TallJimbo: Last sentence looks like a copy-paste relic.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -996,6 +1007,11 @@ def _removeFileExists(uri: ButlerURI) -> None:
                     with formatter._updateLocation(tmpLocation):
                         formatter.write(inMemoryDataset)
                     uri.transfer_from(tmpLocation.uri, transfer="copy", overwrite=True)
+
+                    # Cache if required
+                    if self.cacheManager.should_be_cached(ref):
+                        self.cacheManager.move_to_cache(tmpLocation.uri, ref)
```
TallJimbo: If this is how we always use these methods, I wonder if they should be combined (not necessarily advocating that; just something to ponder).

TallJimbo (approved):

python/lsst/daf/butler/core/datastoreCacheManager.py
timj: That is correct. I had a differing opinion on this in a review last week but I'll remove the `pass`.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: Good idea. I changed it and that allowed the null cache to return `None` if it didn't want to cache anything.

PR #497, opened 2021-03-26
DM-29412: Update StampsBase so that it can serve as StorageClass for subclasses

jbkalmbach:


jbkalmbach:
Leaving `Stamps` and `BrightStarStamps` in so I don't break other people's code.

timj (approved):
Looks okay. Removing `Stamps` and `BrightStarStamps` can be done separately (we'd need to patch the dataset type definitions that use them in the pipelines).

cc/ @SimonKrughoff

PR #500, opened 2021-04-05
DM-29196: Use UUIDs as dataset_ids in registry

andy-slac:
This is implemented as a separate manager in addition to a standard
manager that defines schema with integer PK. This is mostly transparent
though there are many pieces of code that assumed that DatasetRef.id is
an integer and had to be changed. This also needs support for serializing
IDs to YAML and JSON formats. Unit tests were added for the new manager
class including repeated import for UUID-based datasets.

New code supports generation of deterministic IDs but ingest code has not
been updated yet to use this functionality. Configuration has been updated
to use UUID datasets manager by default.


python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -170,14 +178,20 @@ def to_simple(self, minimal: bool = False) -> Dict:
...
         if minimal and self.id is not None:
             # The only thing needed to uniquely define a DatasetRef
-            # is the integer id so that can be used directly if it is
+            # is its id so that can be used directly if it is
             # resolved and if it is not a component DatasetRef.
             # Store is in a dict to allow us to easily add the planned
             # origin information later without having to support
             # an int and dict in simple form.
-            simple: Dict[str, Any] = {"id": self.id}
+            simple: Dict[str, Any] = {"id": _id2json(self.id)}
```
timj: I think this is going to be simpler once I'm allowed to merge #471 because pydantic supports UUID natively (and will cause a change to the JSON again since we won't want the `uuid:` prefix). It will be interesting to see if pydantic handles the `DatasetId` type.

python/lsst/daf/butler/core/config.py
```diff
@@ -58,6 +59,24 @@
...
+    """
+    return dumper.represent_scalar("!uuid", str(data))
+
+
+def _uuid_constructor(loader, node):
+    if node.value is not None:
+        return uuid.UUID(hex=node.value)
+
+
+yaml.Dumper.add_representer(uuid.UUID, _uuid_representer)
```
timj: Do we use UUID in our configs anywhere? Or is this fixing the import/export yaml (and so I think should be in the yaml export/import handler (although adding it here fixes it there because it's global).

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -820,12 +826,103 @@ def insertDatasets(self, datasetType: Union[DatasetType, str], dataIds: Iterable
                                              f"dimension row is missing.") from err
         return refs
 
-    def getDataset(self, id: int) -> Optional[DatasetRef]:
+    @transactional
+    def importDatasets(self, datasets: Iterable[DatasetRef],
+                       idGenerationMode: DatasetIdGenEnum = DatasetIdGenEnum.UNIQUE) -> List[DatasetRef]:
+        """Insert one or more datasets into the `Registry`
```
timj: Import not insert? Something should explain why you would use this instead of insertDatasets. This is "importing refs" because you want to use the `.id` that has already been allocated. Whereas insertDatasets is always allocating an ID for you.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -820,12 +826,103 @@ def insertDatasets(self, datasetType: Union[DatasetType, str], dataIds: Iterable
...
+        if run is None:
+            if self.defaults.run is None:
+                raise TypeError("No run provided to ingestDatasets, "
+                                "and no default from registry construction.")
+            run = self.defaults.run
+
+        runRecord = self._managers.collections.find(run)
+        if runRecord.type is not CollectionType.RUN:
+            raise TypeError(f"Given collection is of type {runRecord.type.name}; RUN collection required.")
+        assert isinstance(runRecord, RunRecord)
```
timj: Is this here because runRecord could be None and it's keeping mypy happy? Wouldn't we have already crashed a couple of lines earlier with an `AttributeError`? This assert seemingly will only fail if we managed to get a `runRecord` that did have a `.type` but wasn't a `RunRecord`. I'm not sure that's worth checking is it?

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -273,11 +324,38 @@ def getCollectionSummary(self, collection: CollectionRecord) -> CollectionSummar
...
-        return _VERSION
-
     def schemaDigest(self) -> Optional[str]:
         # Docstring inherited from VersionedExtension.
         return self._defaultSchemaDigest(self._static, self._db.dialect)
+
+    _version: VersionTuple
+    """Schema version for this class."""
+
+    _storageClass: Type[ByDimensionsDatasetRecordStorage]
```
timj: I'm a bit confused by the term "storageClass" being used here to store something that isn't a `StorageClass`.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
```diff
@@ -430,3 +402,202 @@ def getDataId(self, id: int) -> DataCoordinate:
...
+        Returns
+        -------
+        datasetId : `uuid.UUID`
+            Dataset identifier.
+        """
+        if idGenerationMode is DatasetIdGenEnum.UNIQUE:
+            return uuid.uuid4()
+        else:
+            # WARNING: If you modify this code make sure that the order of
+            # items in the list never changes.
```
timj: maybe say 
```
# items in the "items" list below never changes.
```
?

python/lsst/daf/butler/registry/interfaces/_datasets.py
```diff
@@ -83,6 +112,40 @@ def insert(self, run: RunRecord, dataIds: Iterable[DataCoordinate]) -> Iterator[
         """
         raise NotImplementedError()
 
+    @abstractmethod
+    def import_(self, run: RunRecord, datasets: Iterable[DatasetRef],
+                idGenerationMode: DatasetIdGenEnum = DatasetIdGenEnum.UNIQUE) -> Iterator[DatasetRef]:
+        """Insert one or more dataset entries into the database.
+
+        Parameters
```
timj, line 121: Docs are missing the `run` parameter

python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -95,6 +101,8 @@ def makeRegistryConfig(self) -> RegistryConfig:
         config = RegistryConfig()
         if self.collectionsManager:
             config["managers"]["collections"] = self.collectionsManager
+        if self.datasetsManager:
+            config["managers"]["datasets"] = self.datasetsManager
```
timj: If this is a `Config` you can actually write:

```python
config["managers", "datasets"] = self.datasetsManager
```
because the tuple is unpacked into the hierarchy (but what you've written also works of course).

python/lsst/daf/butler/tests/_dummyRegistry.py
```diff
@@ -150,7 +153,8 @@ class DummyRegistry:
     def __init__(self):
         self._opaque = DummyOpaqueTableStorageManager()
         self.dimensions = DimensionUniverse()
-        self._datastoreBridges = DummyDatastoreRegistryBridgeManager(self._opaque, self.dimensions)
+        self._datastoreBridges = DummyDatastoreRegistryBridgeManager(
+            self._opaque, self.dimensions, sqlalchemy.BigInteger)
```
timj, line 157: Does this mean the tests are not using UUID?

python/lsst/daf/butler/transfers/_yaml.py
```diff
@@ -337,28 +337,23 @@ def load(self, datastore: Optional[Datastore], *,
...
             # Insert all of those DatasetRefs at once.
             # For now, we ignore the dataset_id we pulled from the file
             # and just insert without one to get a new autoincrement value.
             # Eventually (once we have origin in IDs) we'll preserve them.
-            resolvedRefs = self.registry.insertDatasets(
-                datasetType,
-                dataIds=dataIds,
-                run=run,
-            )
+            resolvedRefs = self.registry.importDatasets(datasets)
```
timj: The new method has simplified this so that's good.

python/lsst/daf/butler/transfers/_yaml.py
```diff
@@ -337,28 +337,23 @@ def load(self, datastore: Optional[Datastore], *,
...
             # Populate our dictionary that maps int dataset_id values from the
             # export file to the new DatasetRefs
             for fileId, ref in zip(dataset_ids, resolvedRefs):
```
timj, line 383: Shouldn't we be able to skip all this with UUIDs in import? It's needed for integer IDs but isn't this going to slow down import for no reason in the UUID case? Where we will have imported the UUIDs that we already had? Can we at least add a short circuit where `fileId == ref.id` ?

timj (approved):
Looks great.

python/lsst/daf/butler/core/datasets/ref.py
andy-slac: I can remove `uuid:` prefix if we don't want it, I added as a future-proof in case we want some other ID type that translates into JSON string.

timj: It's fine to leave it in for the moment but on the understanding that I might make it moot in a few days if I am finally allowed to merge my pydantic branch.

python/lsst/daf/butler/core/config.py
andy-slac: This is for import/export, somehow I though it would be nice to have it defined globally, but I can also move it to yaml import/export backend.

timj: Both classes are going to be loaded on code import anyhow so it doesn't really make a difference. As written it implies that UUID are supported in our configs so that was why I had to think a bit. Moving it makes it more explicit.

python/lsst/daf/butler/registry/_registry.py
andy-slac: I'll expand it.

python/lsst/daf/butler/registry/_registry.py
andy-slac: This was copied verbatim from `insertDatasets()`, I'll see if it can be removed.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
andy-slac: storageClass in this context means class (type) which implements `ByDimensionsDatasetRecordStorage` interface, but I see that it can be confused with butler StorageClass. I'll try to find a more distinct name for it.

python/lsst/daf/butler/registry/tests/_registry.py
andy-slac: Indeed, I just copied two lines above without much brain activity involved 🙂

python/lsst/daf/butler/tests/_dummyRegistry.py
andy-slac: I think it it does not really matter, `DummyDatastoreRegistryBridgeManager` does not use that type at all. But it is needed there because base class constructor uses it. Looking again at base class I'm not sure now why I added that attribute to the base class, maybe it should be defined in subclasses instead, I'll look again.

python/lsst/daf/butler/transfers/_yaml.py
andy-slac: I think this code is still generic and can work with both integers and UUIDs. I'll see if optimization can be added if original Id is the same as returned one.

python/lsst/daf/butler/registry/_registry.py
andy-slac: OK, this is indeed for mypy only, `runRecord` cannot be `None` here but it can be any other subclass of `CollectionRecord`. Checking that `runRecord.type` is `RUN` works to constrain the type but mypy still needs convincing.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
andy-slac: I renamed it to `_recordStorageType`.

python/lsst/daf/butler/tests/_dummyRegistry.py
andy-slac: This is complicated. `datasetIdColumnType` is defined in `DatastoreRegistryBridgeManager` so I need to initialize with something. `datasetIdColumnType` needs to be in the base class because FileDatastore uses it to make its table schema. Normally (in regular Registry) this type is obtained from `DatasetRecordStorageManager` (obviously it needs to be consistent) but in case of DummyRegistry there is no datasets manager at all. I do not know if unit tests actually instantiate FileDatastore from a DummyRegistry (they probably do). If we want to run those tests with both integer and UUID IDs then we need to extend them and probably define DummyRegistryUUID class. This sort of repetition is already very ugly,m I had to parameterize bunch of tests in `test_simpleButler.py` and it has all chances to grow if we add new manager implementations.

timj: I wouldn't worry too much about it for now. I had forgotten where DummyRegistry was being used -- the standalone datastore test really doesn't care what the id is.

python/lsst/daf/butler/transfers/_yaml.py
andy-slac: Looking at the code I don't think we can gain much by truing to optimize `fileId == ref.id` case. We still need to build a map from ids to refs for collection association. We can avoid overriding `FileDataset.refs` but that is probably a negligible something compared to actual database transaction. I think that while we still have two types of IDs present code structure that handles generic case should be reasonable. If there is a performance hit I'd like to see some profiling numbers before trying to improve it.

python/lsst/daf/butler/configs/registry.yaml
```diff
@@ -9,5 +9,5 @@ registry:
     opaque: lsst.daf.butler.registry.opaque.ByNameOpaqueTableStorageManager
     dimensions: lsst.daf.butler.registry.dimensions.static.StaticDimensionRecordStorageManager
     collections: lsst.daf.butler.registry.collections.synthIntKey.SynthIntKeyCollectionManager
-    datasets: lsst.daf.butler.registry.datasets.byDimensions.ByDimensionsDatasetRecordStorageManager
+    datasets: lsst.daf.butler.registry.datasets.byDimensions.ByDimensionsDatasetRecordStorageManagerUUID
```
TallJimbo: I think we might want to hold off on making this the default until some supporting changes are ready.  I'm thinking particularly of migration code for existing repos (and some way to organize it); I think it's a good policy to not change the default schema until we have a way to migrate old repos, and being able to migrate them, and I think encouraging early migration might help keep schema version branching in check.  And having this as an option will already unblock testing with limited registry (etc).

python/lsst/daf/butler/transfers/_yaml.py
timj: Okay. It looks like this could all get simpler if we knew we had UUIDs but I agree that it's not worth tackling on this ticket.

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -51,6 +53,12 @@ class AmbiguousDatasetError(Exception):
     """
 
 
+DatasetId = Union[int, uuid.UUID]
+"""A type-annotation alias for dataset ID which could be either integer or
+UUID.
+"""
```
TallJimbo, line 96: I am somewhat tempted to go in later and instead use a generic type parameter on `Registry` and everything else that cares about the `dataset_id` type (since that should always be consistent across one `Registry` instance, and it'd be nice for MyPy to check that).  But that's a lot of work, and it'd be useful mostly as way to use static analysis to help with the conversion that you've already.  Anyhow, just a thought, not something I'd even consider asking you to do on this ticket.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -820,12 +826,106 @@ def insertDatasets(self, datasetType: Union[DatasetType, str], dataIds: Iterable
...
+        dataset ID. If registry supports globally-unique dataset IDs (e.g.
+        `uuid.UUID`) then datasets which already exist in the registry will be
+        ignored if imported again.
+
+        Parameters
+        ----------
+        datasets :  `~collections.abc.Iterable` of `DatasetRef`
+            Datasets to be inserted. All `DatasetRef` instances must have
+            identical ``datasetType`` and ``run`` attributes. ``run``
+            attribute can be `None` and defaults to ``self.defaults.run``.
```
TallJimbo, line 729: I don't think `DatasetRef` will permit `run` attributes to be `None` if they are required to be resolved, unless you've changed that here (and I'd rather not change it here).

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -820,12 +826,106 @@ def insertDatasets(self, datasetType: Union[DatasetType, str], dataIds: Iterable
...
+            Datasets can specify ``id`` attribute which will be used for
+            inserted datasets. All dataset IDs must have the same type
+            (`int` or `uuid.UUID`), if type of dataset IDs does not match
+            configured backend then IDs will be ignored and new IDs will be
+            generated by backend.
```
TallJimbo, line 734: If it's not hard to do, I think rejecting inconsistent ID types as a fatal logic bug is probably better behavior; it could be a serious problem for higher-level code if it expected its IDs to be used as given but they were silently ignored.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -820,12 +826,106 @@ def insertDatasets(self, datasetType: Union[DatasetType, str], dataIds: Iterable
...
+
+        Returns
+        -------
+        refs : `list` of `DatasetRef`
+            Resolved `DatasetRef` instances for all given data IDs (in the same
+            order). If any of ``datasets`` has an ID which already exists in
+            the database then it will not be inserted or updated, but a
+            resolved `DatasetRef` will be returned for it in any case.
+
+
```
TallJimbo: Extra newline.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -820,12 +826,106 @@ def insertDatasets(self, datasetType: Union[DatasetType, str], dataIds: Iterable
...
+        runRecord = self._managers.collections.find(run)
+        if runRecord.type is not CollectionType.RUN:
+            raise TypeError(f"Given collection is of type {runRecord.type.name}; RUN collection required.")
+        assert isinstance(runRecord, RunRecord)
+
+        progress = Progress("lsst.daf.butler.Registry.insertDatasets", level=logging.DEBUG)
+        expandedDatasets = [
+            dataset.expanded(self.expandDataId(dataset.dataId, graph=storage.datasetType.dimensions))
+            for dataset in progress.wrap(datasets, f"Expanding {storage.datasetType.name} data IDs")]
```
TallJimbo: FYI, you'll probably get a rebase conflict around here, because I just added a logic branch that doesn't do data ID expansion; we'll probably want that logic in both `importDatasets` and `insertDatasets`.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
```diff
@@ -430,3 +402,200 @@ def getDataId(self, id: int) -> DataCoordinate:
...
+                idGenerationMode: DatasetIdGenEnum = DatasetIdGenEnum.UNIQUE) -> Iterator[DatasetRef]:
+        # Docstring inherited from DatasetRecordStorage.
+
+        # We do not reuse integer IDs from other repository as it is going
+        # to clash with auto-incremented IDs from this and other repos.
+        return self.insert(
+            run,
+            (dataset.dataId for dataset in datasets),
+            idGenerationMode
+        )
```
TallJimbo: I'd (maybe naively) have expected this method to just `raise`, since can't actually do what the caller wants, and doing what it can do is probably so surprising that even documenting it carefully won't stop someone from using it incorrectly.  But I suppose that's just a repeat of my comment on `Registry.importDatasets`.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
```diff
@@ -430,3 +402,200 @@ def getDataId(self, id: int) -> DataCoordinate:
...
+    """Namespace UUID used for UUID5 generation. Do not change. This was
+    produced by `uuid.uuid5(uuid.NAMESPACE_DNS, "lsst.org")`.
+    """
+
+    def insert(self, run: RunRecord, dataIds: Iterable[DataCoordinate],
+               idMode: DatasetIdGenEnum = DatasetIdGenEnum.UNIQUE) -> Iterator[DatasetRef]:
+        # Docstring inherited from DatasetRecordStorage.
+
+        # Iterate over data IDs, transforming a possibly-single-pass iterable
+        # into a list, and remembering any governor dimension values we see.
```
TallJimbo: Looks like the second half of this comment no longer belongs here.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
```diff
@@ -430,3 +402,200 @@ def getDataId(self, id: int) -> DataCoordinate:
...
+            })
+
+        yield from self._insert(run, dataIdList, rows, self._db.insert)
+
+    def import_(self, run: RunRecord, datasets: Iterable[DatasetRef],
+                idGenerationMode: DatasetIdGenEnum = DatasetIdGenEnum.UNIQUE) -> Iterator[DatasetRef]:
+        # Docstring inherited from DatasetRecordStorage.
+
+        # Iterate over data IDs, transforming a possibly-single-pass iterable
+        # into a list, and remembering any governor dimension values we see.
```
TallJimbo: Same as above: comment is out of date.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
```diff
@@ -430,3 +402,200 @@ def getDataId(self, id: int) -> DataCoordinate:
...
+            })
+
+        yield from self._insert(run, dataIdList, rows, self._db.ensure)
+
+    def _insert(self, run: RunRecord, dataIdList: List[DataCoordinate],
+                rows: List[Dict], insertMethod: Callable) -> Iterator[DatasetRef]:
+        """Common part of implementation of `insert` and `import_` methods.
+        """
+
+        governorValues = GovernorDimensionRestriction.makeEmpty(self.datasetType.dimensions.universe)
```
TallJimbo, line 564: This is where that governor-dimensions comment goes now.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
```diff
@@ -430,3 +402,200 @@ def getDataId(self, id: int) -> DataCoordinate:
...
+        rows = []
+        for dataId in dataIds:
+            dataIdList.append(dataId)
+            rows.append({
+                "id": self._makeDatasetId(run, dataId, idMode),
+                "dataset_type_id": self._dataset_type_id,
+                self._runKeyColumn: run.key,
+            })
+
+        yield from self._insert(run, dataIdList, rows, self._db.insert)
```
TallJimbo, line 532: If you want existing datasets with the same deterministic ID to be ignored rather than cause conflict->rollback (as the docs say), I think you want to use `self.db.ensure` here.   Same in `import_` below.  And ideally we'd then inspect the number of rows inserted (which `ensure` will return) to make sure it's the same as what we passed in, unless `idMode` is a deterministic one - that would catch collisions, which I guess we shouldn't worry about (but I can't help myself).

But there's still a problem with that - there's also a `db.insert` into the tags table would fail due to the unique constraint on dataset type + data ID + collection.  If we use `db.ensure` there, too, _this_ problem is solved, but then we can't tell the difference between inserting the exact same dataset twice and trying to insert a genuinely conflicting dataset with a different `dataset_id` but the same dataset type + data ID + collection.

So I think the easiest way out by far is to change all of the new docs for now that advertise that inserting the same dataset will do nothing.  I think the alternative is to both add a new constraint to the `dataset_tags_*` tables and upgrade `db.ensure` to pay attention to which constraint failed (something I think PostgreSQL can already do, but SQLite can only do in the very latest version that we just blocked upgrading to because of performance regressions).

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
```diff
@@ -430,3 +402,200 @@ def getDataId(self, id: int) -> DataCoordinate:
...
         )
+
+
+class ByDimensionsDatasetRecordStorageInt(ByDimensionsDatasetRecordStorage):
+    """Implementation of ByDimensionsDatasetRecordStorage which uses integer
+    auto-incremented column for dataset IDs.
+    """
+
+    def insert(self, run: RunRecord, dataIds: Iterable[DataCoordinate],
+               idMode: DatasetIdGenEnum = DatasetIdGenEnum.UNIQUE) -> Iterator[DatasetRef]:
```
TallJimbo, line 413: Can we have this raise if `idMode is not DatasetIdGenEnum.UNIQUE`?

python/lsst/daf/butler/transfers/_yaml.py
TallJimbo: Comment above this line about "once we have origin IDs" is now out of date.

tests/test_sqlite.py
```diff
@@ -210,17 +210,45 @@ def makeRegistry(self) -> Registry:
...
+
+class SqliteFileRegistrySynthIntKeyCollMgrUUIDTestCase(SqliteFileRegistryTests, unittest.TestCase):
+    """Tests for `Registry` backed by a SQLite file-based database.
+
+    This test case uses SynthIntKeyCollectionManager and
+    ByDimensionsDatasetRecordStorageManagerUUID.
+    """
+    collectionsManager = "lsst.daf.butler.registry.collections.synthIntKey.SynthIntKeyCollectionManager"
+    datasetsManager = \
+        "lsst.daf.butler.registry.datasets.byDimensions.ByDimensionsDatasetRecordStorageManagerUUID"
```
TallJimbo, line 251: I don't know if we want to start down this path of combinatorial manager testing, at least in terms of running all of these by default.

Probably best to punt this to another ticket (and get @timj thinking about it), but I think we should start marking the non-default-manager tests with something like
```
@unittest.skipUnless("DAF_BUTLER_TEST_EXHAUSTIVE" in os.environ, "Skipping optional tests")
```
and the figure out where we want that environment variable to be set.

TallJimbo (approved):

python/lsst/daf/butler/configs/registry.yaml
andy-slac: OK, I will undo this, my idea was that this is only relevant for new repos, old repos that have integers will still be usable with the old manager, but you are right, it's better to have this transition planned more explicitly.

timj: I'm not entirely sure why we wouldn't make this change. It only affects people doing `butler create`. At the very least I hope we've done one entire jenkins (with ci_hsc) done using the UUID manager otherwise it's not been tested anywhere.

tests/test_sqlite.py
timj: I imagine we always want to fully test the current managers but for the deprecated managers we only want to test basic functionality. Eventually we hope to disable old managers when DMCCB allows us to stop supporting old deployments.

python/lsst/daf/butler/configs/registry.yaml
TallJimbo: My concern was more about establishing good practices than this particular change.

python/lsst/daf/butler/registry/_registry.py
andy-slac: I did not change DatasetRef. As it is implemented right now it does not require resolved refs, I said above that they should be resolved but this is not a must. I was not quite sure that we want to require it here, if we strongly prefer resolved refs then I will change the language here and code that checks for `run`.

python/lsst/daf/butler/registry/_registry.py
andy-slac: I was considering this as a legit use case when for example importing from integer repo into UUID repo. If we disallow that then higher-level code will need to know how to convert ints into UUIDs (or other ID type) before import. Problem with that is that logic for calculating deterministic IDs is deliberately encapsulated into manager and high-level code does not need to care about that detail. If we want to allow high level code to do ID transform then we then need to expose datasets manager type to that code.

python/lsst/daf/butler/registry/_registry.py
andy-slac: OK, will keep that in mind.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
andy-slac: For integer repo this just re-implements what it was doing before, it does not do what's expected but it sort of works in a sense that it will break when you try to import same dataId. Potentially I think we could implement more expensive re-mapping of the IDs by querying existing datasets dataIds and learning their existing IDs instead of asking for auto-incremented ones. For now I'm happy to replace this with `raise` if this is the best way to handle it.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
andy-slac: This is one unanswered question that I had on JIRA. It is not quite clear to me whether `insert` should skip existing dataset or raise. After some thinking I decided that raising is probably the right thing to do, my idea is that ingesting the same (raw) dataset into the same repo should not happen in general, and we should only need to care about identical datasets on import. If this is not true and we actually want `insert` to behave exactly as `import` then `ensure` is the right thing to do. There is indeed an issue of which constraint is violated with `ensure` - PK/ID or uniqueness, this I think matters for detecting genuine ID collisions, I'm not sure how to handle this best.

tests/test_sqlite.py
andy-slac: The alternative may be to test each manager separately and in isolation and only test full registry with a default one. But I'm not sure how difficult would be to test manager code in isolation as it depends on bunch of things from registry.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
andy-slac: Will that break when ingest starts using `DatasetIdGenEnum` and someone will want to ingest something into old integers repo?

python/lsst/daf/butler/registry/_registry.py
TallJimbo: Ah, I hadn't realized that resolved weren't actually required here, just recommended.

python/lsst/daf/butler/registry/_registry.py
TallJimbo: That makes sense.  I think it might be another argument that `_importDatasets` should be a package-private method that's expected to be called by the import code, but probably not code outside daf_butler.

andy-slac: I'm OK with making it semi-private though this probably will not stop determined people from using it. I'll rename it and also expand docstring to explain why it is private. I am still unsure what everyone's position on UUID generation is. Is what I implemented currently sufficient or do we really want to make clients responsible for generating UUIDs (both random and deterministic)?

timj: I suppose the question is, if I create a DatasetRef and attach a uuid myself, can I get that ref ingested as-is?

andy-slac: The answer should be yes if you know that you are ingesting into UUID registry (which is what I want to try to hide from you).

timj: I'm actually happy for it to fail if a UUID DatasetRef is ingested into an integer registry.  UUID is meant to be the future. As time goes by there will be effectively zero old style registries. Once we've updated the main ones and change raw-ingest then if people want to ingest into an old registry can't they use the old software (or else have a configuration option in raw-ingest).

TallJimbo: I am happy with butler always being responsible for actually generating UUIDs.  I mostly just want to emphasize that the control we already give higher-level code in terms of what kind of UUID to generate still means there's a lot of things butler code can't do properly on its own (like know how to migrate existing repository content or correctly import integer-`dataset_id` exports).

andy-slac: I agree that we still depend on client for consistently defining policy for deterministic IDs. If we want butler to be in complete control of that we can make that policy a part of registry configuration.

I'll change import to always raise if an import of UUIDs is attempted into integer registry.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
TallJimbo: :+1: for raising.  Please just check that this is consistent with the documentation.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
TallJimbo: Yes, and I think that's a good thing, at least for now - I want to force the high level code/user to acknowledge what kind of ID they are getting and respond accordingly.

TallJimbo (approved):
New `reuseIds` commit looks good, except that the commit message has a typo ("ffect").

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
andy-slac: `insertDataset()` already documents that exception is raised when attempting to insert dataset with matching dataId/dataset/run.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
andy-slac: I have added an exception if non-UINQUE is passed to `import_` or `insert` for integer IDs.

PR #503, opened 2021-04-06
DM-28315: Create Registry ABC

timj:


python/lsst/daf/butler/registry/_registry.py
```diff
@@ -25,107 +25,127 @@
...
-        Database,
+        CollectionRecord,
         DatastoreRegistryBridgeManager,
     )
 
-
 _LOG = logging.getLogger(__name__)
 
 # key for dimensions configuration in attributes table
 _DIMENSIONS_ATTR = "config:dimensions.json"
```
andy-slac: This probably does not belong here.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -25,107 +25,127 @@
...
+        ----------
+        config : `RegistryConfig` or `str`, optional
+            Registry configuration, if missing then default configuration will
+            be loaded from registry.yaml.
+        """
+        if not isinstance(config, RegistryConfig):
+            if isinstance(config, (str, Config)):
+                config = RegistryConfig(config)
+            else:
+                raise ValueError("Incompatible Registry configuration: {}".format(config))
```
andy-slac: f-string?

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -25,107 +25,127 @@
...
+    @classmethod
+    def forceRegistryConfig(cls, config: Optional[Union[ButlerConfig,
+                                                        RegistryConfig, Config, str]]) -> RegistryConfig:
+        """Force the supplied config to a `RegistryConfig`.
+
+        Parameters
+        ----------
+        config : `RegistryConfig` or `str`, optional
+            Registry configuration, if missing then default configuration will
+            be loaded from registry.yaml.
```
andy-slac: The code does not handle `None` (or rather it raises for `None`).

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -25,107 +25,127 @@
...
     """
 
+    @classmethod
+    def forceRegistryConfig(cls, config: Optional[Union[ButlerConfig,
+                                                        RegistryConfig, Config, str]]) -> RegistryConfig:
+        """Force the supplied config to a `RegistryConfig`.
+
+        Parameters
+        ----------
+        config : `RegistryConfig` or `str`, optional
```
andy-slac: There is no default for an argument, it should not be `optional`.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -151,35 +171,16 @@ def createFromConfig(cls, config: Optional[Union[RegistryConfig, str]] = None,
         registry : `Registry`
             A new `Registry` instance.
         """
```
andy-slac, line 181: Maybe add a note that subclasses must override this method? Same for `fromConfig`.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -151,35 +171,16 @@ def createFromConfig(cls, config: Optional[Union[RegistryConfig, str]] = None,
...
-            raise TypeError(f"Incompatible Dimension configuration type: {type(dimensionConfig)}")
-
-        DatabaseClass = config.getDatabaseClass()
-        database = DatabaseClass.fromUri(str(config.connectionString), origin=config.get("origin", 0),
-                                         namespace=config.get("namespace"))
-        managerTypes = RegistryManagerTypes.fromConfig(config)
-        managers = managerTypes.makeRepo(database, dimensionConfig)
-        return cls(database, RegistryDefaults(), managers)
+        registry_cls, registry_config = cls.determineTrampoline(config)
+        return registry_cls.createFromConfig(config, dimensionConfig, butlerRoot)
```
andy-slac: Do you want to pass `registry_config` instead of `config` here?

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -429,8 +321,25 @@ def getCollectionType(self, name: str) -> CollectionType:
...
-        return self._managers.collections.find(name).type
+        raise NotImplementedError()
+
+    @abstractmethod
+    def _get_collection_record(self, name: str) -> CollectionRecord:
+        """Return the record for this collection.
 
+        Parameters
+        ----------
+        collection : `str`
```
andy-slac: collection -> name?

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -429,8 +321,25 @@ def getCollectionType(self, name: str) -> CollectionType:
...
+            Name of the collection for which the record is to be retrieved.
+
+        Returns
+        -------
+        record : `CollectionRecord`
+            The record for this collection.
+        """
+        raise NotImplementedError
```
andy-slac: `NotImplementedError()`, just for consistency 🙂

python/lsst/daf/butler/registry/_sqlRegistry.py
```diff
@@ -0,0 +1,1849 @@
...
+        CollectionRecord,
+        Database,
+        DatastoreRegistryBridgeManager,
+    )
+
+
+_LOG = logging.getLogger(__name__)
+
+# key for dimensions configuration in attributes table
+_DIMENSIONS_ATTR = "config:dimensions.json"
```
andy-slac: Same here, I don't think it belongs here.

python/lsst/daf/butler/registry/_sqlRegistry.py
```diff
@@ -0,0 +1,1849 @@
...
+
+
+class SqlRegistry(Registry):
+    """Registry interface based on SQLAlchemy.
```
andy-slac: interface -> implementation? (but interface is probably OK to depending on what you mean by interface 🙂 )

python/lsst/daf/butler/registry/_sqlRegistry.py
```diff
@@ -0,0 +1,1849 @@
...
+        Manager class implementing `DatasetRecordStorageManager`.
+    datastoreBridges : `type`
+        Manager class implementing `DatastoreRegistryBridgeManager`.
+    dimensionConfig : `DimensionConfig`, optional
+        Dimension universe configuration, only used when ``create`` is True.
+    writeable : `bool`, optional
+        If True then Registry will support write operations.
+    create : `bool`, optional
+        If True then database schema will be initialized, it must be empty
+        before instantiating Registry.
```
andy-slac: Constructor only has three arguments.

python/lsst/daf/butler/registry/_sqlRegistry.py
```diff
@@ -0,0 +1,1849 @@
...
+        # Intentionally invoke property setter to initialize defaults.  This
+        # can only be done after most of the rest of Registry has already been
+        # initialized, and must be done before the property getter is used.
+        self.defaults = defaults
+
+    def __str__(self) -> str:
+        return str(self._db)
+
+    def __repr__(self) -> str:
+        return f"Registry({self._db!r}, {self.dimensions!r})"
```
andy-slac: Registry - > SqlRegistry?

python/lsst/daf/butler/registry/_sqlRegistry.py
```diff
@@ -0,0 +1,1849 @@
...
+
+    def isWriteable(self) -> bool:
+        """Return `True` if this registry allows write operations, and `False`
+        otherwise.
+        """
```
andy-slac: No need to repeat docstrings for inherited methods? Probably applies to many methods below, I'm not going to check all docstrings, I assume they will disappear.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -287,109 +264,23 @@ def defaults(self, value: RegistryDefaults) -> None:
...
         This may be necessary to enable querying for entities added by other
-        `Registry` instances after this one was constructed.
+        registry instances after this one was constructed.
         """
-        self._managers.refresh()
+        raise NotImplementedError()
 
     @contextlib.contextmanager
+    @abstractmethod
     def transaction(self, *, savepoint: bool = False) -> Iterator[None]:
```
andy-slac, line 291: Do you need  `@contextmanager` for an abstract method?

timj:
I've redone the commit history to try to give git a chance to handle the changes in #500 when either that is rebased onto this or this is rebased onto that.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -287,109 +264,23 @@ def defaults(self, value: RegistryDefaults) -> None:
...
-        try:
-            with self._db.transaction(savepoint=savepoint):
-                yield
-        except BaseException:
-            # TODO: this clears the caches sometimes when we wouldn't actually
-            # need to.  Can we avoid that?
-            self._managers.dimensions.clearCaches()
-            raise
-
-    def resetConnectionPool(self) -> None:
```
andy-slac, line 311: This is probably SQL-specific, but removing it from interface will force clients to check the actual type of the Registry before calling it. Maybe better to keep it but provide default implementation that does nothing?

python/lsst/daf/butler/registry/_sqlRegistry.py
```diff
@@ -0,0 +1,1849 @@
...
+        Returns
+        -------
+        ref : `DatasetRef` or `None`
+            A ref to the Dataset, or `None` if no matching Dataset
+            was found.
+        """
+        ref = self._managers.datasets.getDatasetRef(id)
+        if ref is None:
+            return None
+        return ref
```
andy-slac: `if` looks superfluous.

python/lsst/daf/butler/registry/_sqlRegistry.py
```diff
@@ -0,0 +1,1849 @@
...
+        AmbiguousDatasetError
+            Raised if any ``ref.id`` is `None`.
+        OrphanedRecordError
+            Raised if any dataset is still present in any `Datastore`.
+        """
+        progress = Progress("lsst.daf.butler.Registry.removeDatasets", level=logging.DEBUG)
+        for datasetType, refsForType in progress.iter_item_chunks(DatasetRef.groupByType(refs).items(),
+                                                                  desc="Removing datasets by type"):
+            storage = self._managers.datasets.find(datasetType.name)
+            assert storage is not None
```
andy-slac: Using `self._managers.datasets[datasetType.name]` should remove the need for `assert`, there are few cases like this below.

python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -52,16 +52,17 @@
     ddl,
     Timespan,
 )
-from .._registry import (
-    CollectionSummary,
-    CollectionType,
+from .._sqlRegistry import Registry
```
andy-slac: Should it be `from .._registry import Registry` or `from .._sqlRegistry import SqlRegistry`?

tests/test_cliCmdPruneDatasets.py
```diff
@@ -349,7 +349,7 @@ def test_purgeWithDisassociate(self):
             exPruneDatasetsExitCode=1
         )
 
-    @patch.object(lsst.daf.butler.Registry, "getCollectionType",
+    @patch.object(lsst.daf.butler.registry.SqlRegistry, "getCollectionType",
```
andy-slac, line 352: I guess this would also need patching other registry classes if we ever want to use this test with non-SqlRegistry?

andy-slac (approved):
Looks good, small bunch of minor comments.

timj:
@andy-slac Thanks. Some of your comments are from existing code that I moved around but I will take a look at them. I was minimizing changes to the old `Registry` class since I know we have code changes for that class in flight and I was worried about the rebase conflict.

python/lsst/daf/butler/registry/_registry.py
timj: Indeed. It was left behind when the code was changed a while back.

python/lsst/daf/butler/registry/_registry.py
timj: Old code copied from old `Registry`...

python/lsst/daf/butler/registry/_registry.py
timj: Thanks. I made it work with None.

python/lsst/daf/butler/registry/_registry.py
timj: Yes, thanks. It wouldn't matter in this particular case because the subclass forces it to a `RegistryConfig` anyhow but it definitely was something I should have done.

python/lsst/daf/butler/registry/_registry.py
timj: No, but I thought it would be better to be more explicit as a guide.

python/lsst/daf/butler/registry/_registry.py
timj: Okay. mypy didn't complain so I ignored it. I had not noticed that ctrl_mpexec was relying on it.

python/lsst/daf/butler/registry/tests/_registry.py
timj: Yes, that's funny.

timj: It works because it's only used for type annotation.

tests/test_cliCmdPruneDatasets.py
timj: Yes, but for now I think it's fine because this is a command line tool test and not really a registry test.

python/lsst/daf/butler/registry/_sqlRegistry.py
timj: This has been wrong forever. I wasn't expecting you to *read* the whole of the Registry implementation... (just the diffs from the commit history). I'll clean it up.

python/lsst/daf/butler/registry/_sqlRegistry.py
timj: I didn't write any of this....

python/lsst/daf/butler/registry/_sqlRegistry.py
timj: I've removed them from SqlRegistry. Hope this doesn't mess with your insertDatasets docstring changes when the rebase happens.

PR #505, opened 2021-04-07
DM-29617: Experimentation with remote registry

timj:


python/lsst/daf/butler/core/serverModels.py
```diff
@@ -0,0 +1,114 @@
...
+                    params[k] = []
+                params["regex"].append(expression.pattern)
+            else:
+                if (k := "glob") not in params:
+                    params[k] = []
+                params["glob"].append(expression)
+        return cls(**params)
+
+
+class QueryDatasetsModel(BaseModel):
```
timj: @TallJimbo this is what I've started doing for modeling the parameters to the registry APIs.

python/lsst/daf/butler/core/dimensions/_records.py
```diff
@@ -75,6 +78,61 @@ def _subclassDimensionRecord(definition: DimensionElement) -> Type[DimensionReco
...
+            if type(v) is not SpecificSerializedDimensionRecord:
+                # If already a subclass, assume the right type.
+                return v
+            # Convert to specific subclass.
+            v = v.dict()
+
+        if not isinstance(v, collections.abc.Mapping):
+            raise ValueError(f"record can either be SpecificDimensionRecord or dict, not {v!r}")
+
+        universe = DimensionUniverse._instances[values["universe"]]
```
timj: @TallJimbo I managed to get pydantic to create specific subclasses of serialized records so that validation could be enabled. The main problem being that you need a Universe to get the relevant DimensionElement.  This works because I store the universe version in the model and can therefore get the universe from the singleton. This can potentially significantly simplify the general `from_simple` methods since we could embed the universe version in all the models and they would ask Universe for that version (I assume that other versions should always fail because those versions aren't the one stored in the registry table). Thoughts:

* Currently I'm assuming that we will pass the version number from the client to the server in a parameters model. The server will compare that version to its own version and then if they match would pass that universe to `from_simple` as now.
* Embedding universe version potentially leads to the universe version number appearing at every level of the serialization and in one client/server interaction it should be the same everywhere.
* If we aren't embedding version numbers how could the dimension record be associated with the right universe? Since DimensionUniverse is already a singleton could we have an API that declares the active universe? (maybe as a context manager). One issue is that the model is instantiated by fastapi before we see it so I think I might need to create a universe  when the server starts up since otherwise there might not be one active if the records are being sent from the client (which means a context manager wouldn't work either).
* Currently the version number of a universe is a private attribute. There needs to be a public API for it.
* How would we distinguish a SPHEREx universe version from a Rubin universe version?

python/lsst/daf/butler/core/utils.py
```diff
@@ -408,23 +408,40 @@ def findFileResources(values: Iterable[str], regex: Optional[str] = None) -> Lis
...
+        the provided expressions would match everything.
     """
+    if expressions is Ellipsis or expressions is None:
+        return Ellipsis
+    expressions = list(iterable(expressions))
     if not expressions or "*" in expressions:
         return Ellipsis
-    return [re.compile(fnmatch.translate(e)) for e in expressions]
+
+    nomagic = re.compile("^[A-Za-z_]+$")
```
TallJimbo: If this is ever applied to collection expressions we probably want to include `/` as well.

python/lsst/daf/butler/registry/_remoteRegistry.py
```diff
@@ -0,0 +1,431 @@
...
+        self._db = server_uri
+        self._defaults = defaults
+
+        # All PUT calls should be short-circuited if not writeable.
+        self._writeable = writeable
+
+        self._dimensions: Optional[DimensionUniverse] = None
+
+        # Does each API need to be sent the defaults so that the server
+        # can use specific defaults each time?
```
TallJimbo: I think I'd try first to just make the defaults purely client-side and never rely on the defaults in the server.

python/lsst/daf/butler/registry/_remoteRegistry.py
```diff
@@ -0,0 +1,431 @@
...
+
+        # Storage class information should be pulled from server.
+        # Dimensions should be pulled from server.
+
+    def __str__(self) -> str:
+        return str(self._db)
```
TallJimbo: Make sure this doesn't repeat [DM-29249](https://jira.lsstcorp.org/browse/DM-29249) (which I'm now reminded we still haven't fixed).

python/lsst/daf/butler/registry/_remoteRegistry.py
```diff
@@ -0,0 +1,431 @@
...
+
+    def __repr__(self) -> str:
+        return f"RemoteRegistry({self._db!r}, {self.dimensions!r})"
```
TallJimbo: Probably want `writeable` here too.

python/lsst/daf/butler/registry/_remoteRegistry.py
```diff
@@ -0,0 +1,431 @@
...
+        # Access /dimensions.json on server and cache it locally.
+        try:
+            response = httpx.get(str(self._db.join("universe")))
+            response.raise_for_status()
+        except httpx.RequestError as e:
+            print(f"An error occurred while requesting {e.request.url!r}.")
+            raise e
+        except httpx.HTTPStatusError as e:
+            print(f"Error response {e.response.status_code} while requesting {e.request.url!r}.")
+            raise e
```
TallJimbo: Is this kind of `print` logging desirable in this context for some reason, or are these debug relics?

python/lsst/daf/butler/registry/_remoteRegistry.py
```diff
@@ -0,0 +1,431 @@
...
+    #    in a transaction but if the client wants one transaction that
+    #    involves two server methods how does that work? The server will have
+    #    already closed the transaction.
+    @contextlib.contextmanager
+    def transaction(self, *, savepoint: bool = False) -> Iterator[None]:
+        # No-op for now
+        try:
+            yield
+        except BaseException:
+            raise
```
TallJimbo: As we've discussed a bit, I don't think there's any way out of this without changing the `SqlRegistry` interface.

In the meantime, if you can make this `raise NotImplementedError` or something to make sure nobody thinks it'll do something it doesn't, that'd be better, I think. But I understand if that's already impossible.

python/lsst/daf/butler/registry/_remoteRegistry.py
```diff
@@ -0,0 +1,431 @@
...
+        config = DimensionConfig.fromString(response.text, format="json")
+        self._dimensions = DimensionUniverse(config)
+        return self._dimensions
+
+    def refresh(self) -> None:
+        # Docstring inherited from lsst.daf.butler.registry.Registry
+
+        # Need to determine what to refresh.
+        # Might need a server method to return all the DatasetTypes up front.
+        pass
```
TallJimbo: It occurs to me that this is not really a safe thing to have in the API at all; even if there is no client state to refresh (yet), it should tell the server to refresh its in-memory state from the DB, so it can pick up e.g. new dataset type registrations from other server instances.  But we also really don't want each random client to be able to tell the server to do a full refresh.

In many places, we could probably do automatic refreshes when e.g. looking for something and we don't see it in the cache.  But we probably need to worry about making those incremental so we don't get O(N^2) scaling or similar from repeated cache misses and total refreshes.  And there are also places where we really can't know that the cache is incomplete, anyway, such as when iterating over dataset types or collections in order to generate a larger query.

I don't see a way out of this that isn't really difficult.  Might be best to start by adding a suite of `Manager` classes that never hold any kind of local state (that should be doable within the current APIs) to just avoid all of the caching complexity imposed by the current managers, and start over again at optimizing things based on what's actually slow (which will often put the caches in the client instead of the server `Registry` instance, though we should be careful not to optimize for the one-concurrent-user case).  Anyhow, that's definitely not a project for this ticket, but it's one of the big hard problems we'll have to solve.

python/lsst/daf/butler/registry/_remoteRegistry.py
```diff
@@ -0,0 +1,431 @@
...
+        # No-op for now
+        try:
+            yield
+        except BaseException:
+            raise
+
+    # insertOpaqueData + fetchOpaqueData + deleteOpaqueData
+    #    There are no managers for opaque data in client. This implies
+    #    that the server would have to have specific implementations for
+    #    use by Datastore. DatastoreBridgeManager also is not needed.
```
TallJimbo: I did have designs on being able to use the opaque tables for other things (e.g. backing a `SqlDatastore`'s actual dataset storage, not just its extra records, and using that to save SQuaSH metrics), but I'm not sure yet whether any of them are viable in the no-shared-DB execution environment.  So, yes, let's not add those methods here now, but I'd rather not rule out having something like them in the future.

python/lsst/daf/butler/registry/_remoteRegistry.py
```diff
@@ -0,0 +1,431 @@
...
+
+    # Should makeQueryBuilder be a public method? Nothing uses it outside
+    # of Registry.
```
TallJimbo: Does not need to be public.  If I can ever get back to working on the query system it would be wholesale replaced anyway.

python/lsst/daf/butler/registry/_remoteRegistry.py
```diff
@@ -0,0 +1,431 @@
...
+
+        Returns
+        -------
+        manager : `DatastoreRegistryBridgeManager`
+            Object that mediates communication between this `Registry` and its
+            associated datastores.
+        """
+        from ..tests._dummyRegistry import DummyDatastoreRegistryBridgeManager, DummyOpaqueTableStorageManager
+        return DummyDatastoreRegistryBridgeManager(DummyOpaqueTableStorageManager(),
+                                                   self.dimensions, int)
```
TallJimbo: I suspect I'll know more when I've looked at more of the PR, but I'm guessing it might make more sense to just move this method to `SqlRegistry` and not define it for `RemoteRegistry`.

python/lsst/daf/butler/core/serverModels.py
```diff
@@ -0,0 +1,107 @@
...
+
+import re
+
+from pydantic import BaseModel
+
+from .utils import iterable, globToRegex
+
+# DataId and kwargs can only take limited types
+DataIdValues = Union[str, int, float]
+DataId = Mapping[str, DataIdValues]
```
TallJimbo: There are already some very similar (but not quite identical) type aliases at

https://github.com/lsst/daf_butler/blob/f94a57823ca904996b74b8bab85282891f2d18a5/python/lsst/daf/butler/core/dimensions/_coordinate.py#L62

and

https://github.com/lsst/daf_butler/blob/f94a57823ca904996b74b8bab85282891f2d18a5/python/lsst/daf/butler/core/dimensions/_coordinate.py#L685

python/lsst/daf/butler/core/serverModels.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class ExpressionQueryParameter(BaseModel):
+    """Represents a specification for an expression query."""
+
+    regex: Optional[List[str]] = None
+    """List of regular expression strings."""
+
+    glob: Optional[List[str]] = None
+    """List of dataset type names or globs."""
```
TallJimbo: I'm a bit surprised to see globs mentioned at all here; I had considered those to be purely part of the CLI.

python/lsst/daf/butler/core/serverModels.py
```diff
@@ -98,10 +104,11 @@ class QueryDatasetsModel(BaseModel):
...
     collections: Optional[ExpressionQueryParameter] = None
+    dimensions: Optional[List[str]] = None
     dataId: Optional[DataId] = None
     where: Optional[str] = None
     findFirst: bool = False
     components: Optional[bool] = None
     bind: Optional[DataId] = None
     check: bool = True
-    kwargs: Optional[DataId] = None
+    keyword_args: Optional[DataId] = None  # mypy refuses to allow kwargs in model
```
TallJimbo: This problem goes away if you do a little bit more standardization/normalization of the arguments on the client before or when making the model, such as calling `DataCoordinate.standardize` to merge `dataId` and `**kwargs` to generate something that's definitely a `DataCoordinate`.  Could go in a `classmethod` factory on the model class.

python/lsst/daf/butler/registry/_remoteRegistry.py
```diff
@@ -382,7 +386,36 @@ def queryDatasets(self, datasetType: Any, *,
                       check: bool = True,
                       **kwargs: Any) -> queries.DatasetQueryResults:
         # Docstring inherited from lsst.daf.butler.registry.Registry
-        raise NotImplementedError()
+        if collections is not None:
+            print("Collections:", collections)
```
TallJimbo: Debug relic.

python/lsst/daf/butler/core/dimensions/_coordinate.py
```diff
@@ -59,12 +60,20 @@
 DataCoordinate.
 """
 
-DataIdValue = Union[str, int, None]
+# Pydantic will cast int to str if str is first in the Union.
+DataIdValue = Union[int, str, None]
```
TallJimbo, line 64: Another duplicate of https://github.com/lsst/daf_butler/blob/f94a57823ca904996b74b8bab85282891f2d18a5/python/lsst/daf/butler/core/dimensions/_coordinate.py#L62

python/lsst/daf/butler/core/dimensions/_records.py
TallJimbo: I'm not actually sure the extra automatic validation is worth it in this case.  I figured having separate classes for different dimensions was useful (as opposed to `SimpleNamespace` or `dict`) because there typos are a real problem and tab-completion of attributes is nice, and I was worried (paranoid?) about using a class with `__slots__` instead of `__dict__` storage for memory efficiency.  But I'm not sure that's really won out over the complexities of having dynamically-defined types even in the pure local `SqlRegistry` case.

And while we may not be able to rely on all clients giving the server consistent dimensions and want to do some validation there, surely there's some pydantic hook that lets you e.g. validate that the keys of a dict match some expectation?  (And if not, I imagine we could always explicitly check in the server code).

I'm already pretty uncomfortable with using the singleton dict keyed on version to work around the pickle limitations in `mulitprocessing` (if that let me use a custom `Unpickler`, I absolutely would, and drop the singleton entirely), as I think it dangerously conflates the kind of version number we want for schema management with something that's really an ugly implementation detail.  It'd be safer to use some kind of digest/hash instead of an incrementing number, at least in the singleton, and maybe elsewhere, at least for everything other than linear, "dimensions used by the Rubin product" human-meaningful schema versioning (which already has problems with e.g. SPHEREx, as you note).

python/lsst/daf/butler/core/serverModels.py
```diff
@@ -82,33 +93,62 @@ def from_expression(cls, expression: Any) -> ExpressionQueryParameter:
...
-                if (k := "regex") not in params:
-                    params[k] = []
                 params["regex"].append(expression.pattern)
-            else:
-                if (k := "glob") not in params:
-                    params[k] = []
+            elif isinstance(expression, str):
                 params["glob"].append(expression)
+            elif hasattr(expression, "name"):
+                params["glob"].append(expression.name)
```
TallJimbo, line 135: Level of duplicative logic is starting to worry me here, but I think there's a commit I haven't gotten to yet starts to address that...

python/lsst/daf/butler/core/serverModels.py
```diff
@@ -82,33 +93,62 @@ def from_expression(cls, expression: Any) -> ExpressionQueryParameter:
...
+
+class QueryDimensionRecordsModel(BaseModel):
+    """Information needed to query the dimension records."""
+
+    dataId: Optional[SerializedDataCoordinate] = None
+    datasets: Optional[DatasetsQueryParameter] = None
+    collections: Optional[ExpressionQueryParameter] = None
+    where: Optional[str] = None
+    components: Optional[bool] = None
+    bind: Optional[SimpleDataId] = None
```
TallJimbo: This is not a data ID; the mapping is from `str` to any type that a `DimensionRecord` attribute can have (so at least also `bool`, `float`, and `Timespan`, in addition to `int` and `str`).

TallJimbo (approved):
Looks good. Lots of comments, but nothing major that isn't a "think about for the future" thing; closest is that I'd like to avoid using the `DimensionUniverse` anywhere we have any other option, even if that means not having pydantic validation of `DimensionRecords`.

I'm about to disappear until later this evening, so while I haven't forgotten about the questions in Jira, it'll be a bit before I get to them.  I also haven't looked at the server code at all yet; I'll do that later, too.

python/lsst/daf/butler/registry/_remoteRegistry.py
timj: I cleaned that up over the weekend. It was the first client routine I wrote so it had some debug examples in there from the httpx documentation.

python/lsst/daf/butler/registry/_remoteRegistry.py
timj: Yes, I added it here for now because without it *everything* blows up because Butler tries to create a Datastore class by passing in the manager and this requires a "real" datastore manager even if Datastore is not being used. Without this I couldn't even demonstrate the `butler query-X` command lines worked.

python/lsst/daf/butler/core/serverModels.py
timj: Yes, I saw those and I think I changed this code over the weekend. I thought I had pushed before I sent out the review request but clearly I hadn't. I did change it to drop the `float` but should have added the `None`.

It's funny that both those definitions for DataId value are in the same file.

python/lsst/daf/butler/core/serverModels.py
timj: I'm starting to get concerned that you are reviewing an old version of this pull request somehow.

Anyhow, yes, I called it `glob` rather than `strings` because when I was using the test server directly (and not as part of the RemoteRegistry) it was easier to be able to say `cal*` to do the wildcard matching. There will be some users of the server who aren't using Python.

python/lsst/daf/butler/core/serverModels.py
timj: I had forgotten that I can call `DataCoordinate.standardize` without a registry and so can fold those into the DataId before I send it.

python/lsst/daf/butler/registry/_remoteRegistry.py
timj: That was removed yesterday.

python/lsst/daf/butler/core/dimensions/_coordinate.py
timj: It's the actual line but you have to put the int before the str to control Pydantic casting.

python/lsst/daf/butler/core/serverModels.py
timj: Oh, you are reviewing commit by commit. That is not a good idea on this pull request where much of the code is work in progress building up to something that works as I experiment and keep cleaning up as I learn more. I had not expected you to be reviewing like this.

timj: This doesn't seem like duplication though. It supports someone passing in a DatasetType but me wanting to send the name of that dataset type over the wire (and not the full DatasetType serialized form)

python/lsst/daf/butler/core/serverModels.py
TallJimbo: Thinking about this a bit more, maybe we make `CollectionSearch.fromExpression` (and similar) responsible for identifying glob-like strings and turning them into regexes.  It's already responsible for standardizing a bunch of other kinds of inputs, and that'd let us use the globs in the Python API all over the place.

python/lsst/daf/butler/core/dimensions/_coordinate.py
TallJimbo: Feel free to change the one in `_coordinate.py` to make pydantic happy; I'm pretty sure it's the only thing that cares about the order right now.

python/lsst/daf/butler/core/serverModels.py
TallJimbo: > Oh, you are reviewing commit by commit. That is not a good idea on this pull request

Yeah, I realized that you didn't expect this last night while reading your replies, a couple of threads before you realized that's what I'd done.  In any case, aside from these false positives (I'm going through and resolving all of the clear-cut ones I see; feel free to resolve any others yourself), I think it may have actually helped me wrap my head around it a bit as it gave me a chronological view rather than an alphabetical view of a big suite of changes.

TallJimbo: And yes, I just misunderstood what this was doing r.e. duplication - it just looked suspiciously like the code that standardizes dataset type expressions as query inputs, but of course it's doing something quite different.

python/lsst/daf/butler/core/dimensions/_records.py
timj: > surely there's some pydantic hook that lets you e.g. validate that the keys of a dict match some expectation

I think all validation requires a universe and the pydantic models don't have a universe (if they did everything would work anyhow). I think we'd have to put the validation directly in `from_simple` since that does have a universe. I will switch to that approach.

python/lsst/daf/butler/core/dimensions/_coordinate.py
timj: We are commenting on the line in `_coordinate.py`. This change fixes the order.

TallJimbo: :facepalm:

python/lsst/daf/butler/core/dimensions/_records.py
timj: I retained the dynamic pydantic model creation and now create a specialist class to validate everything in `from_simple`. It works well and if I send odd data types or extra fields it does now complain. An obvious problem is that I dynamically create a model every time I call `from_simple` on a record and that seems less than ideal. Caching it works but the hash for DimensionElement is the name of the element so that doesn't seem safe because the universe can change -- should the DimensionElement know the universe it came from?

TallJimbo: `DimensionElement.universe` does exist, so you could cache on that plus the name (if `DimensionUniverse` is hashable; I'm not sure about that).

I wouldn't add the unvierse to `DimensionElement.__hash__` because that's allowed to be equal for non-equal objects anyway; adding a universe comparison to `DimensionElement.__eq__` might be okay, but there's also all of the `Named*` containers that assume that name equality is equal to object equality (ooh, that might be a problem for non-unique DatasetType names...), so I'd prefer to just define the equality comparison as "only meaningful within a universe" and add the universe explicitly only in contexts (like that cache) where necessary.

timj: Thanks. This seems to work (including the universe).

python/lsst/daf/butler/registry/_remoteRegistry.py
timj: Auth is going to be done with tokens so I don't think this is going to be an issue here.

python/lsst/daf/butler/registry/_remoteRegistry.py
timj: SqlRegistry does not report writeable so I was following that.

python/lsst/daf/butler/registry/_remoteRegistry.py
timj: pipetask uses `registry.refresh` in multiple places so I felt I had to include the method for now.  I am happy to remove it from the ABC but that would make it impossible for pipetask use remote registry in the medium term (maybe that's not going to be a problem given that we'd first have to sort out the datastore/remote registry interaction).

python/lsst/daf/butler/registry/_remoteRegistry.py
timj: I think I'm edging towards having remote datastore registry provide higher level interfaces but it may well be that this interface does eventually turn up directly on server.

python/lsst/daf/butler/core/serverModels.py
timj: although given your other ticket it sounds like I should be adding a `datasetTypes` field to the model and storing the full form for transmission to the server so it can work out exactly which dataset types it is going to be using and not just the name.

TallJimbo: Good point.  FWIW, if the user just passed the name you wouldn't need to pass more than that to the server (because the server could do the expansion), but if they did pass more, it would become important to respect that.  And it might make sense to always do the expansion on the client anyway.

python/lsst/daf/butler/registry/_remoteRegistry.py
TallJimbo: :+1:  I'm not sure if that was intentional or not.

timj:
@TallJimbo FYI I've relocated the SqlRegistry and RemoteRegistry to `lsst.daf.butler.registries` matching the pluggable datastore approach. These registries should not be imported by default but should instead be loaded via configuration.

python/lsst/daf/butler/registry/_remoteRegistry.py
timj: This seems doable but the downside then is that the client has to reimplement the server logic of knowing when to call standardize or default the collection and run. If the client sends everything to the server this all happens without the client having to care. On the other hand doing it client side would simplify the interface and potentially reduce the size of the JSON. Although non-python clients won't have any idea how to call `DataCoordinate.standardize()`.

python/lsst/daf/butler/registry/_remoteRegistry.py
timj: I had it be a no-op so that it would let the client experiment with doing writes on the server without having to solve the hard problem.  Another way to look at it is that this might lead to us getting in to trouble later on when we do want to think about transaction handling.  I'll change it to raise for the moment.

python/lsst/daf/butler/core/serverModels.py
timj: In almost every case sql registry calls `self.expandDataId` on the arguments. I think a `Registry.standardizeDataId()` method that folds in dataId+kwargs+defaults would work. It doesn't look like RegistryDefaults is amenable to pydantic because it (1) takes `**kwargs` and (2) needs registry to finish construction (making it incompatible with being immutable). I'd need to make a to_simple/from_simple and also allow `RegistryDefaults` to be constructed from a trusted dataId+collections+run rather than requiring a registry to check the dataId.

python/lsst/daf/butler/core/serverModels.py
```diff
@@ -0,0 +1,286 @@
...
+    None,
+    title="Additional parameters to use when standardizing the supplied data ID.",
+)
+
+
+class QueryBaseModel(BaseModel):
+    """Base model for all query models."""
+
+    @validator("keyword_args", check_fields=False)
+    def check_keyword_args(cls, v, values) -> Optional[SimpleDataId]:  # type: ignore # noqa: N805
```
rra: Nit: I tend to name pydantic validators with a leading underscore since they're not intended to be part of the public interface of the class.

python/lsst/daf/butler/core/serverModels.py
```diff
@@ -0,0 +1,286 @@
...
+    )
+    collections: Optional[ExpressionQueryParameter] = Collections
+    dimensions: Optional[List[str]] = OptionalDimensions
+    dataId: Optional[SerializedDataCoordinate] = DataId
+    where: Optional[str] = Where
+    findFirst: bool = FindFirst
+    components: Optional[bool] = Components
+    bind: Optional[BindType] = Bind
+    check: bool = Check
+    keyword_args: Optional[SimpleDataId] = KeywordArgs  # mypy refuses to allow kwargs in model
```
rra, line 259: Can't you lift this up to the parent class since it's common in all subclasses? Or is that what the comment is about?

rra (commented):
Nothing jumps out at me as unreasonable here. I'm surprised you managed to get defining fields at the top level to work; I struggled with that for a while but mypy really didn't like it. I may steal your approach.

python/lsst/daf/butler/registries/remote.py
```diff
@@ -0,0 +1,522 @@
...
+        # Docstring inherited from lsst.daf.butler.registry.Registry
+        raise NotImplementedError()
+
+    @functools.lru_cache
+    def getCollectionType(self, name: str) -> CollectionType:
+        # Docstring inherited from lsst.daf.butler.registry.Registry
+        # This could use a local cache since collection types won't
+        # change.
+        path = f"v1/registry/collection/type/{name}"
+        response = httpx.get(str(self._db.join(path)))
```
rra: Does `ButlerURI` enforce that the server URI ends in a `/`? If not, you may want to enforce that.

rra (commented):
Yeah, the general shape of `remote.py` looks right to me. That's what I'd expect of the translation error: convert parameters to a pydantic model and then make the call and convert the JSON back to a model.

python/lsst/daf/butler/registries/remote.py
timj: It's annoyingly complicated in general but yes, ButlerURI knows if it represents a "directory".

python/lsst/daf/butler/core/serverModels.py
timj: I was going to build all the models and then work out which bits needed to be in the base.

PR #506, opened 2021-04-10
DM-29562: Rewrite dataset subquery logic to make it easier on query planners.

TallJimbo:


python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
```diff
@@ -400,17 +406,72 @@ def select(self, collection: CollectionRecord,
...
+                self._static.dataset,
+                onclause=(dataset_id_col == self._static.dataset.columns.id),
+                **static_kwargs,
             )
+            # Also constrain dataset_type_id in static table in case that helps
+            # generate a better plan.
+            # We could also include this in the JOIN ON clause, but my guess is
+            # that that's a good idea IFF it's in the foreign key, and right
+            # now it isn't.
+            query.where.append(self._static.dataset.columns.dataset_type_id == self._dataset_type_id)
```
andy-slac, line 435: I don't think that this is helping as the constraint on dataset_id is already very strict, but I hope it is not hurting either.

andy-slac (approved):
Looks good.

PR #507, opened 2021-04-12
DM-27412: Rework daf.butler.tests.makeTestCollection

kfindeisen:
This PR alters `makeTestCollection` to support an externally provided test ID.

### DIFF
--- a/python/lsst/daf/butler/tests/_testRepo.py
+++ b/python/lsst/daf/butler/tests/_testRepo.py
@@ -115,7 +115,7 @@ def makeTestRepo(root: str,
     return butler
 
 
-def makeTestCollection(repo: Butler) -> Butler:
+def makeTestCollection(repo: Butler, uniqueId: Optional[str] = None) -> Butler:
     """Create a read/write Butler to a fresh collection.
 
     Parameters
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -137,9 +140,11 @@ def makeTestCollection(repo: Butler) -> Butler:
...
     processing or analysis.
     """
-    # Create a "random" collection name
-    # Speed matters more than cryptographic guarantees
-    collection = "test" + str(random.randrange(1_000_000_000))
+    if not uniqueId:
+        # Create a "random" collection name
+        # Speed matters more than cryptographic guarantees
+        uniqueId = str(random.randrange(1_000_000_000))
+    collection = "test" + uniqueId
```
parejkoj: I'd put an underscore or other character between the `test` and the id, to make it easier to read (on the rare occasion when a person needs to read it).

tests/test_testRepo.py
```diff
@@ -65,7 +65,8 @@ def tearDownClass(cls):
         removeTestTempDir(cls.root)
 
     def setUp(self):
-        self.butler = makeTestCollection(self.creatorButler)
+        # TestCase.id is unique for each test
```
parejkoj: `id()` - otherwise it looks like a property/attribute to me.

It's unique for each "test method", but if you re-run with the same method it will be the same. Not sure if that needs to be made clear here.

parejkoj (approved):
Two small comments, but otherwise this is a good improvement.

PR #509, opened 2021-04-15
DM-29708: Make ButlerHttpURI compatible with non-webDAV HTTP backends

bgounon:
This update allows the user to make simple requests against pure http backends without ButlerHttpURI complaining
+ reworked session creation a little to make authentication (via env variables) optional 
+ cleaned up old useless code from before the ButlerURI rework


python/lsst/daf/butler/core/_butlerUri/http.py
```diff
@@ -84,11 +84,22 @@ def getHttpSession() -> requests.Session:
 
     log.debug("Creating new HTTP session...")
 
+    ca_bundle = None
+    try:
+        ca_bundle = os.environ['LSST_BUTLER_WEBDAV_CA_BUNDLE']
+    except KeyError:
+        log.warning("Environment variable LSST_BUTLER_WEBDAV_CA_BUNDLE is not set: "
```
timj: I don't think this statement about https is true. I can get the content from a GitHub web page using ButlerURI just fine so this statement is not specifically about https.  I think this warning should only happen if you are connecting to a webdav backend (I think you have code to know) with `isWebdavEndpoint`

python/lsst/daf/butler/core/_butlerUri/http.py
```diff
@@ -84,11 +84,22 @@ def getHttpSession() -> requests.Session:
...
+                    "HTTPS requests will fail. If you intend to use HTTPS, please "
+                    "export this variable.")
+    session.verify = ca_bundle
+
     try:
         env_auth_method = os.environ['LSST_BUTLER_WEBDAV_AUTH']
     except KeyError:
-        raise KeyError("Environment variable LSST_BUTLER_WEBDAV_AUTH is not set, "
-                       "please use values X509 or TOKEN")
+        log.debug("Environment variable LSST_BUTLER_WEBDAV_AUTH is not set, "
```
timj, line 99: I'd only write this message if it's a webdav server.

python/lsst/daf/butler/core/_butlerUri/http.py
```diff
@@ -376,6 +287,10 @@ def size(self) -> int:
 
     def mkdir(self) -> None:
         """Create the directory resource if it does not already exist."""
+        # Only available on WebDAV backends
+        if not isWebdavEndpoint(self.scheme + "://" + self.netloc):
```
timj: This scheme+netloc code seems like it needs to be a general method of ButlerURI that can be reused.

Since ButlerURI is now immutable, I think you probably want an `self.is_webdav_endpoint` `@property` that will test once on the server but then cache the response (it's never going to change).

timj (changes requested):
This looks okay but I'd like some minor changes:

* Add an is_webdav property that will cache the response from the server so you aren't continually asking the server if it's webdav each time (ButlerURI is immutable).
* Only warn about missing env vars if it's a webdav server.
* Consider adding a generic `ButlerURI.root()` method that returns the root URI (no path) [although if you've created the is_webdav property at least the current code will not be repeated in multiple places]

python/lsst/daf/butler/core/_butlerUri/http.py
bgounon: My problem here is that getHttpSession() currently is not aware of the server URL, so we cannot internally check if the server is a webdav endpoint or not. I could add an optional parameter to getHttpSession() though, but even pure http servers may require x509/token authentication, so maybe it could be useful in some cases, even if the backend is not webdav.
And in that case maybe change the name of the env variable to something like LSST_BUTLER_HTTP_AUTH

python/lsst/daf/butler/core/_butlerUri/http.py
bgounon: You are right, but it is not specifically about webdav neither. I changed the statement to make it clearer that it is mostly useful to trust non-default CAs, and made it a debug log instead of a warning log.

bgounon:
@timj new version pushed, let me know if there are more changes to address

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
```diff
@@ -308,6 +308,16 @@ def geturl(self) -> str:
...
 
+    def getrooturl(self) -> str:
+        """Return the base root URI in string form.
+
+        Returns
+        -------
+        url : `str`
+            String form of root URI.
+        """
+        return self._uri.scheme + "://" + self._uri.netloc
```
timj: To be consistent with the other APIs this should return a new ButlerURI and not a string.

python/lsst/daf/butler/core/_butlerUri/http.py
timj: I think the ability to add new CAs is worthwhile but we shouldn't complain if someone is doing https somewhere on GitHub and uses standard certificates. Now that it's a debug I'm not worried about the log message confusing people when it's not relevant.

python/lsst/daf/butler/core/_butlerUri/http.py
```diff
@@ -345,18 +263,16 @@ def session(self) -> requests.Session:
...
-
+        s = getHttpSession()
         ButlerHttpURI._session = s
         ButlerHttpURI._sessionInitialized = True
         return s
 
+    @property
+    def is_webdav_endpoint(self) -> bool:
+        """Check if the current endpoint implements WebDAV features"""
+        return isWebdavEndpoint(self.getrooturl())
```
timj: Cache this in `self._is_webdav` so that we don't need the call to the server each time.
Something like:

```python
@property
def is_webdav_endpoint(self) -> bool:
    """..."""
    try:
        return self._is_webdav
    except AttributeError:
        pass
    self._is_webdav = isWebdavEndpoint(str(self.root_uri())
    return self._is_webdav
```

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
```diff
@@ -308,6 +308,16 @@ def geturl(self) -> str:
         """
         return self._uri.geturl()
 
+    def getrooturl(self) -> str:
```
timj: I'd prefer this be `root_uri` I think.

python/lsst/daf/butler/core/_butlerUri/http.py
```diff
@@ -345,18 +263,16 @@ def session(self) -> requests.Session:
...
-        baseURL = self.scheme + "://" + self.netloc
-
-        if isWebdavEndpoint(baseURL):
-            log.debug("%s looks like a Webdav endpoint.", baseURL)
-            s = getHttpSession()
-        else:
-            raise RuntimeError(f"Only Webdav endpoints are supported; got base URL '{baseURL}'.")
-
+        s = getHttpSession()
         ButlerHttpURI._session = s
```
timj, line 250: Since we are storing this for all possible remote hosts we wouldn't be able to do the warnings per-host anyhow.

timj (approved):
Thanks. Minor comments only.

python/lsst/daf/butler/core/_butlerUri/http.py
```diff
@@ -345,18 +263,21 @@ def session(self) -> requests.Session:
...
         return s
 
+    @property
+    def is_webdav_endpoint(self) -> bool:
+        """Check if the current endpoint implements WebDAV features""" 
+        try:
+            return ButlerHttpURI._is_webdav
+        except AttributeError:
+            pass
+        ButlerHttpURI._is_webdav = isWebdavEndpoint(self.root_uri())
```
timj: Doesn't this mean that every subsequent ButlerURI thinks it's webdav even if it's for a completely different server? I think the cache should be indexed by the root_uri shouldn't it?

timj: It might be easier to have a non-class function that uses an `functools.lru_cache` that takes the `self.root_uri()` as argument. You'll get caching for free (ButlerURI can now be used as a key).

bgounon: I made `_is_webdav` class-wide to be coherent with what I do with `_session`, although I admit I'm unsure this is the best way to do it. In my tests, without using the class scope, `_session` and `_is_webdav` would be refreshed every time we change directory/file, even though it is the same endpoint/server.
I will look into `functools.lru_cache` which I didn't know before.
I'll also fix mypy/lint issues in an upcoming commit.

timj: Global cache is absolutely fine but as written it was caching a single value regardless of the URI root. We need a per-root cache which can doesn't need to be attached to the class itself (hence lru_cache decorator on a standalone function).

timj:
@bgounon would you like me to take over this pull request? I think you are most of the way there and I just need to add the global cache by host name.

bgounon:
> @bgounon would you like me to take over this pull request? I think you are most of the way there and I just need to add the global cache by host name.

Hi @timj, sorry for letting this PR hanging like that. I've mostly been busy with Qserv and RSP matters recently and haven't taken the time to finalize things here. I have no problem with you taking over for the cache implementation, of course.
Otherwise, I expect I should be able to come back to it the week after next.

timj (approved):
@bgounon I've added the webdav caching so I think I'm happy to merge it now. I'll merge next week unless you say something. Thanks for the cleanup.

PR #513, opened 2021-04-30
DM-29966: Fix logic for appending butler.yaml to remote butler config URIs

timj:


python/lsst/daf/butler/_butlerConfig.py
```diff
@@ -87,20 +86,27 @@ def __init__(self, other: Optional[Union[str, ButlerURI, Config]] = None,
...
+            # we add the default configuration name.
+            should_append = False
+            if uri.isdir():
+                should_append = True
+            elif not uri.isLocal and not uri.getExtension():
+                # For remote URI people can easily forget to add the trailing
+                # slash. Attempt to fix this up by assuming that if
+                # the URI is remote and has no file extension then it is
+                # probably meant to be a directory.
+                uri = ButlerURI(other, forceDirectory=True)
```
ktlim, line 98: Isn't `uri.isdir()` true after this?  If so, I think you can simplify the logic and get rid of the flag variable.

ktlim (approved):

python/lsst/daf/butler/_butlerConfig.py
timj: Okay. `isdir` check is cheap if `isLocal` is False so that shouldn't slow anything down.

PR #515, opened 2021-04-30
DM-27241: Add Butler.getArtifacts

timj:


python/lsst/daf/butler/_butler.py
```diff
@@ -1131,6 +1131,43 @@ def getURI(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
...
+    def getArtifacts(self, refs: Iterable[DatasetRef],
+                     destination: Union[str, ButlerURI], transfer: str = "auto",
+                     preserve_path: Optional[bool] = True) -> List[ButlerURI]:
+        """Retrieve the artifacts associated with the supplied refs.
+
+        Parameters
+        ----------
+        refs : iterable of `DatasetRef`
+            The datasets for which artifacts are to be retrieved.
+            A single ref can result in multiple artifacts. The refs must
```
timj, line 1144: I'm working on the assumption that the normal approach would be `butler.registry.queryDatasets` and then `butler.getArtifacts` on the result, so resolved ref is okay to be required here rather than having to support collection+dataId+datasetType.

python/lsst/daf/butler/_butler.py
```diff
@@ -1131,6 +1131,43 @@ def getURI(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
...
+            be resolved.
+        destination : `ButlerURI` or `str`
+            Location to write the artifacts.
+        transfer : `str`, optional
+            Method to use to transfer the artifacts. Must be one of the options
+            supported by `ButlerURI.transfer_from()`.
+        preserve_path : `bool`, optional
+            If `True` the full path of the artifact within the datastore
+            is preserved. If `False` the final file component of the path
```
ktlim, line 1153: What happens if the final file components collide, or should that never happen?

python/lsst/daf/butler/_butler.py
```diff
@@ -1131,6 +1131,43 @@ def getURI(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
...
+        -------
+        targets : `list` of `ButlerURI`
+            URIs of file artifacts in destination location. Order is not
+            preserved.
+
+        Notes
+        -----
+        For non-file datastores the artifacts written to the destination
+        may not match the representation inside the datastore. For example
+        a hierarchichal data structure in a NoSQL database may well be stored
```
ktlim:
```suggestion
        a hierarchical data structure in a NoSQL database may well be stored
```

python/lsst/daf/butler/cli/cmd/commands.py
```diff
@@ -455,3 +456,17 @@ def query_dimension_records(**kwargs):
...
+@query_datasets_options(showUri=False, useArguments=False, repo=False)
+@destination_argument(help="Destination URI of folder to receive file artifacts.")
+@transfer_option()
+@click.option("--preserve-path/--no-preserve-path", is_flag=True, default=True,
+              help="Preserve the datastore path to the artifact at the destination.")
+@options_file_option()
+def retrieve_artifacts(**kwargs):
+    """Retrieve file artifacts associated with datasets in a repository."""
+    transferred = script.retrieveArtifacts(**kwargs)
+    print(f"Number of artifacts retrieved into destination {kwargs['destination']}: {len(transferred)}")
```
ktlim, line 481: Is this intentionally a `print()` as opposed to a log or anything else?

Might it be useful for the list of artifact URIs to be an (optional) output?

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1282,6 +1282,63 @@ def getURI(self, ref: DatasetRef, predict: bool = False) -> ButlerURI:
...
+        -------
+        targets : `list` of `ButlerURI`
+            URIs of file artifacts in destination location. Order is not
+            preserved.
+        """
+        if not destination.isdir():
+            raise ValueError(f"Destination location must refer to a directory. Given {destination}")
+
+        # Source -> Destination
+        to_transfer: Dict[ButlerURI, ButlerURI] = {}
```
ktlim, line 1325: This presumes that the set of artifacts to be transferred is relatively small.  You might want to consider a generator-based implementation in case there are millions of artifacts (but we are not likely to reach that scale for quite a while, if ever).

tests/test_cliCmdRetrieveArtifacts.py
```diff
@@ -0,0 +1,85 @@
...
+        return list(ButlerURI.findFileResources([root]))
+
+    def testRetrieveAll(self):
+        runner = LogCliRunner()
+        with runner.isolated_filesystem():
+
+            for counter, (preserve_path, prefix) in enumerate((("--preserve-path",
+                                                               "ingest/run/test_metric_comp."),
+                                                              ("--no-preserve-path",
+                                                               "test_metric_comp_"))):
```
ktlim, line 65: Why `.` in one and `_` in the other?

python/lsst/daf/butler/core/datastore.py
```diff
@@ -638,6 +638,43 @@ def getURI(self, datasetRef: DatasetRef, predict: bool = False) -> ButlerURI:
         """
         raise NotImplementedError("Must be implemented by subclass")
 
+    @abstractmethod
+    def getArtifacts(self, refs: Iterable[DatasetRef],
```
ktlim: I think I'd prefer something like "copy" or "retrieve" than "get" here; typically we use "get" to return Python objects.

ktlim (approved):

python/lsst/daf/butler/_butler.py
timj: The transfer will fail and everything will stop. It shouldn't happen because we are meant to check that the file itself is unique based on the associated dimensions and collection.

python/lsst/daf/butler/cli/cmd/commands.py
timj: print is what @n8pease seemed to use in other click commands so I followed that. I don't think we use logging at all at the click layer. I can move it down to the "script" layer and use `log.info` for it so that it can be turned off completely.

I was a bit worried about printing all the final URIs out but I can add a verbose mode to do that. Debug logging will show the explicit "transferring A to B" message.

python/lsst/daf/butler/_butler.py
ktlim: That also means that an interrupted retrieve cannot be retried in place without cleaning up first.  Maybe that's obvious, but it should be documented.

python/lsst/daf/butler/cli/cmd/commands.py
ktlim: No need to use log.  Seems reasonable to have this be a print.  (Actually I should have removed that first question once I saw that you actually checked the output in the tests.)

python/lsst/daf/butler/datastores/fileDatastore.py
timj: I can do the transfer in the loop (allowing the refs from queryDatasets to go straight to transfer) but fundamentally I think the right answer is to use asyncio or something (especially for S3 to file). If I do that the command line tool won't be able to report the count of the number of refs either (which might be fine but I worried that that would leave the command line tool essentially blocking for minutes as the query happens and then the transfer from S3 happens).  In chained datastore it's a bit harder to save on memory because I have to filter before I forward the request to the child datastore.

tests/test_cliCmdRetrieveArtifacts.py
timj: because the file name uses a `_` to represent the component but the directory is allowed to use a `.` in it. I think what you are really asking is for a comment to say that.

python/lsst/daf/butler/core/datastore.py
timj: Okay. I'll use `retrieve` then. I wasn't sure how you wanted to play it (we do have `getURI`). I didn't want to use `copy` because we don't copy sometimes, we link.

python/lsst/daf/butler/_butler.py
timj: I can add a `clobber` option that will always overwrite. In theory I can make it transactional and have all the transfers roll back on error.

python/lsst/daf/butler/cli/cmd/commands.py
timj: Looks like click wants me to use `click.echo()` -- https://click.palletsprojects.com/en/7.x/utils/#printing-to-stdout

(which also lets me use colors in the output)

timj: I'll leave it as `print` for now since it is probably a bit odd for this subcommand to use `click.echo` but all the others to use `print`.

python/lsst/daf/butler/_butler.py
timj: Added a clobber/overwrite option. Could also check whether the file exists and skip the transfer if it does. Can leave that to a follow up ticket if that is wanted.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: The dict does actually help here because queries can result in repeated DatasetRef and this dict has the side effect of filtering those out so that you don't try to copy the same file multiple times.

timj:
@ktlim I've added a `--clobber` option to the command line and a simplistic `--verbose` that will report all the transferred URIs. I'll worry about parallelism on another ticket.

PR #516, opened 2021-05-05
DM-29908: Guard against race condition in aggressive collection fetching.

TallJimbo:
A better solution to this would be a manager that doesn't load collections quite as aggressively (e.g. cache on first use, or fetching those associated with a user, etc), and that will someday be important for scaling, too.  But it's a lot harder than this fix.


andy-slac (approved):
Looks good.

PR #517, opened 2021-05-06
DM-29849: Make emptyTrash more efficient

timj:


python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1648,59 +1607,43 @@ def emptyTrash(self, ignore_errors: bool = True) -> None:
...
-                    else:
-                        raise FileNotFoundError(err_msg)
-
-                for location, _ in fileLocations:
-
-                    if not self._artifact_exists(location):
-                        err_msg = f"Dataset {location.uri} no longer present in datastore {self.name}"
+        # It will also automatically delete the relevant rows from the
+        # trash table and the records table.
+        with self.bridge.emptyTrash(self._table, record_class=StoredFileInfo) as trashed:
```
timj: @TallJimbo this is the first pass at doing the join in the bridge. It still does the per-file test though. For this implementation I could assume the list of records is manageable and do an `IN` select on all the files to get back the refs in a set and compare the two. I didn't quite catch your idea in the meeting.

python/lsst/daf/butler/registry/bridge/monolithic.py
```diff
@@ -185,24 +184,54 @@ def check(self, refs: Iterable[DatasetIdRef]) -> Iterable[DatasetIdRef]:
...
+        # inconsistency would be missed by this query.
+        sql = records_table._table.select(
         ).select_from(
             self._tables.dataset_location_trash
         ).where(
-            self._tables.dataset_location_trash.columns.datastore_name == self.datastoreName
+            sqlalchemy.sql.and_(
+                self._tables.dataset_location_trash.columns.dataset_id
+                == records_table._table.columns.dataset_id,
+                self._tables.dataset_location_trash.columns.datastore_name == self.datastoreName
```
TallJimbo: Took me a while to figure out the right SQL (and that's easier that trying to explain it in words), but this is what I had in mind for a way to remove the per-dataset `_can_remove_dataset_artifact` call:
```suggestion
                self._tables.dataset_location_trash.columns.datastore_name == self.datastoreName,
                sqlalchemy.sql.not_(
                    path_column.in_(
                        # Subquery for all paths in dataset_locations (not trash),
                        # as we don't want to delete any artifact still in use
                        # there.
                        sqlalchemy.sql.select(
                            [path_column]
                        ).select_from(
                            self._tables.dataset_location.join(
                                records_table._table,
                                onclause=(
                                    self._tables.dataset_location.columns.dataset_id
                                    == records_table._table.columns.dataset_id
                                ),
                            )
                        ).where(
                            self._tables.dataset_location_trash.columns.datastore_name == self.datastoreName
                        )
                    )
                )
```
That requires
```
path_column = records_table._table.columns["path"]
```
to be defined earlier.  But it's kind of encapsulation-breaking for the generic bridge manager to know that the "path" column is special or even that a query over something like paths is relevant for this datastore, and I don't know how best to deal with that.

timj: Yes, up to now all we are saying is that there is an id in this table matching an id in that other one and when you do the join you can throw all the content at this class I'm giving you.  It does seem a bit of a step to jump from there to requiring that a path column exists. I can try your code and see how fast it makes it and compare it with doing a separate IN query. I guess it doesn't help that the `path` is a TEXT field and not part of an index.

timj: Also, now that I'm trying this, we still need to delete the rows from records and trash even if the file is owned by multiple refs. This is controlling whether to delete the artifact, not whether to delete the ref from datastore. Won't any combo query lose the rows that must be deleted even if the artifacts stay unless the resulting row has a flag in it saying whether to keep the artifact or delete it.

TallJimbo: We could rewrite the query to make the "can_delete_artifact" a returned `bool` column, but that'd complicate constructing `StoredItemInfo` from the result row.

How about we make this two queries instead of one (or N+1):
```
# Helper closure to generate the common join+where clause.
def join_records(select: sqlalchemy.sql.Select, location_table: sqlalchemy.schema.Table) -> sqlalchemy.sql.FromClause:
    return select.select_from(
        record_table._table.join(
            location_table,
            onclause=record_table._table.columns.dataset_id == location_table.columns.dataset_id,
        )
    ).where(
        location_table.columns.datastore_name == self.datastoreName
    )

# Tim's original query (`sql`)
info_in_trash = join_records(record_table._table.select(), self._tables.dataset_location_trash)

# Some helper subqueries
paths_not_in_trash = join_records(
    sqlalchemy.sql.select([record_table._table.columns["path"]]),
    self._tables.dataset_location,
).alias("paths_not_in_trash")
paths_in_trash = join_records(
    sqlalchemy.sql.select([record_table._table.columns["path"]]),
    self._tables.dataset_location_trash,
).alias("paths_in_trash")

# A query for paths that are referenced by datasets in the trash and datasets not in the trash.
paths_to_preserve = sqlalchemy.sql.select(
    [paths_in_trash.columns["path"]]
).select_from(
    paths_not_in_trash.join(
        paths_in_trash,
        onclause=paths_in_trash.columns["path"] == paths_not_in_trash.columns["path"]
    )
)
```
I'm thinking the bridge manager would execute both queries and return via the context manager both the list of `StoredItemInfo` and a `set` of paths that need to be preserved.  The string column name `"path"` would be another argument passed by the Datastore; if it's `None`, we just wouldn't do the second query.

And I'm less uncomfortable than I was yesterday with having the bridge class run this query for the datastore - it's essentially asking the datastore for the name of a column that uniquely identifies its artifacts and offering to find artifacts shared across the trash boundary, which seems reasonably generic.

And yes, I think we probably do want an index on `path` in the future, when we can do it as a (trivial) migration.  For now, I think we'll still get a huge speedup without it, because I'm 95% sure it's the latency of a per-dataset query that's killing us, and just moving the repeated no-index searches into a loop inside the database will be enough of a speedup for now.

timj: I did the `in` clause query outside of the bridge manager and timing went from 8 seconds to 5 seconds (`master` takes 13). The actual timing for deleting 1000 datasets (with real timing in the emptyTrash method) is something like.

Before:
```
INFO  2021-05-07T12:07:35.102-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1650)- Emptying trash
INFO  2021-05-07T12:07:35.106-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1653)- Requested trashed datasets
INFO  2021-05-07T12:07:42.471-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1706)- Trash complete
```

with join as on this ticket and in operator to create set of all refs per file:

```
INFO  2021-05-07T12:07:59.412-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1633)- Emptying trash
INFO  2021-05-07T12:07:59.440-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1649)- Done with database
INFO  2021-05-07T12:08:00.721-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1687)- Trash complete
```
so without all the command line overhead we have the whole thing complete now in just over a second whereas before it takes 7 seconds.

This seems good. I think doing it your way above probably leads to a lower memory overhead and less risk of IN operator going bonkers with a million files.

timj: For completeness, the table querying overhead in this version of the code is 3 seconds (so half the time for half the number of select calls).

timj:
@TallJimbo I've tried your new approach and the bad news is that the second query takes a long time. Here are some results from a repo with 10k datasets deleting a 5k collection.

The bottom line is that the extra query to find preserved artifacts is taking 14 seconds but doing a simple "give me all refs associated with these paths" takes a fraction of a second.  This may be down entirely to the lack of index on file_datastore_records. Dumping the sqlite table seems to indicate that there is no index on file_datastore_records at all and now I'm confused. 

# Original code:

INFO  2021-05-07T16:10:42.830-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1650)- Emptying trash
INFO  2021-05-07T16:10:42.884-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1653)- Requested trashed datasets
INFO  2021-05-07T16:11:34.477-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1706)- Trash complete

51 seconds

# One join in bridge

With join with records table but still doing check of single-file per multi ref

INFO  2021-05-07T16:13:12.185-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1609)- Emptying trash
INFO  2021-05-07T16:13:12.264-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1614)- Trash table queried
INFO  2021-05-07T16:13:37.261-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1649)- Trash complete

Trash time: 25 seconds

# Single join + separate IN

With standard join in bridge and then separate IN query

INFO  2021-05-07T16:24:07.506-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1634)- Emptying trash
INFO  2021-05-07T16:24:07.571-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1639)- Trash table queried
INFO  2021-05-07T16:24:07.690-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1654)- Database queries complete
INFO  2021-05-07T16:24:14.527-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1699)- Trash complete

Trash time: 7 seconds

# Double query in bridge

With double query in bridge for empty preserved artifacts

INFO  2021-05-07T16:17:38.926-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1635)- Emptying trash
INFO  2021-05-07T16:17:52.690-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1649)- Trash table queried
INFO  2021-05-07T16:17:59.273-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1708)- Trash complete

Noting that the query part now takes 14 seconds. The file delete itself is 7 seconds.

Trash time: 21 seconds

timj:
Now with an index and magically it is fast:

```
INFO  2021-05-07T17:05:14.431-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1636)- Emptying trash
INFO  2021-05-07T17:05:14.591-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1650)- Trash table queried
INFO  2021-05-07T17:05:21.870-0700 lsst.daf.butler.datastores.fileDatastore ()(fileDatastore.py:1709)- Trash complete
```

Diff is:

```diff
diff --git a/python/lsst/daf/butler/datastores/fileDatastore.py b/python/lsst/daf/butler/datastores/fileDatastore.py
index 7d276f50..a1669e5e 100644
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -228,6 +228,7 @@ class FileDatastore(GenericBaseDatastore):
                 ddl.FieldSpec(name="file_size", dtype=BigInteger, nullable=True),
             ],
             unique=frozenset(),
+            indexes=(("dataset_id",), ("path",), ("dataset_id", "path")),
         )
 
     def __init__(self, config: Union[DatastoreConfig, str],
```
I'm not entirely sure what the right answer is and "component" should be involved since it's dataset_id+component that is unique. I am surprised it didn't automatically add an index for dataset_id.   We don't have an index on dataset_locations either since it's dataset_id+datastoreName that is the unique quantity.

TallJimbo:
I'd guess that `(dataset_id, path)` as the only index for the records table would work as well as all three of those, but that's really just a guess; I'm slightly more confident that that removes the need for an index on `dataset_id` alone.  Sounds like we should also make `(dataset_id, component)` the primary key, especially if the table doesn't have a primary key already.

And yes, `dataset_locations` and `dataset_locations_trash` should have a unique constraint on `(datastore_name, dataset_id`) in that order, which will also create an index for them.

It _looks_ like we could get away with adding the index and unique constraints on this ticket (and patching repos we know about manually) without a formal migration, because those changes won't be noticed by @andy-slac 's digest function that checks to see if the schema in the database matches what daf_butler would create.  But the primary key changes would have to wait, and I'd want Andy to weigh on whether taking advantage of the digest function gap for the rest is inadvisable for some reason.

timj:
`(dataset_id, component)` is already the primary key for file_datastore_records. I'll see about adding the others though.

TallJimbo:
Ok, I guess that's also unique, so it shouldn't matter much whether that or `(dataset_id, component)` is the PK vs another unique index.

timj:
I can't work out how to add indexes to the dataset_location tables. Somehow it's ignoring the indexes attribute.

I'm also a bit worried that I messed up my testing since even if I add indexes manually to both tables I can't get the new double query to go fast. It's possible I was running the test with the `IN` version (which you can trigger by passing `None` in as the column name of interest. I'll revisit another day.

TallJimbo:
I would expect it to created the indexes specified in the tableSpec only if you're creating a new repo.

timj:
I've been creating new repos for testing but for some reason adding `indexes` to https://github.com/lsst/daf_butler/blob/master/python/lsst/daf/butler/registry/bridge/monolithic.py#L99 isn't working. Maybe something about the adding of the foreign key a few lines later?

TallJimbo:
Yeah, those FKs will create indexes on `dataset_id` alone (because the FK is on `dataset_id` alone), and would block redundant manual indexes on that alone.  I would not expect them to block other indexes, like those on a combination of `dataset_id` and something else.

timj:
That would be it. datastore_name and dataset_id are both primary keys already hence indexing request would be ignored. Next week I'll send you my "create a test repo" script so you can try this branch to see what I'm missing in terms of why it's so slow with the double query.

timj:
I finally read back up the discussion here and noticed that there was an extra index I had talked about early on and which in subsequent discussions I had ignored. Adding an index for "path" alone was what was missing. Doing that fixes things.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1647,60 +1633,85 @@ def emptyTrash(self, ignore_errors: bool = True) -> None:
...
+                        del path_map[info.path]
+
+                artifacts_to_keep = set(path_map)
+
+            for ref, info in trashed:
+
+                if info is None:
+                    # Should not happen for this implementation but need
+                    # to keep mypy happy.
+                    raise RuntimeError(f"Internal logic error in emptyTrash with ref {ref}.")
```
TallJimbo: Given the motivation, this feels like an `assert` (same for those below).

python/lsst/daf/butler/registry/interfaces/_bridge.py
```diff
@@ -192,24 +195,44 @@ def check(self, refs: Iterable[DatasetIdRef]) -> Iterable[DatasetIdRef]:
...
+            Table of records to query with the trash records.
+        record_class : `type` of `StoredDatastoreItemInfo`, optional
+            Class to use when reading records from ``records_table``.
+        record_column : `str`, optional
+            Name of the column in records_table that refers to the artifact.
+
+        Yields
+        ------
+        matches : iterable of `tuple` of (`DatasetIdRef`,
+                                          `StoredDatastoreItemInfo`)
```
TallJimbo: Does this line break work with Sphinx?  If so, I wish I'd known about it before; in the past I've just abbreviated types and provided less information because I couldn't figure out what to do.

python/lsst/daf/butler/registry/interfaces/_bridge.py
```diff
@@ -192,24 +195,44 @@ def check(self, refs: Iterable[DatasetIdRef]) -> Iterable[DatasetIdRef]:
         raise NotImplementedError()
 
     @abstractmethod
-    def emptyTrash(self) -> ContextManager[Iterable[DatasetIdRef]]:
+    def emptyTrash(self, records_table: Optional[OpaqueTableStorage] = None,
+                   record_class: Optional[Type[StoredDatastoreItemInfo]] = None,
+                   record_column: Optional[str] = None,
+                   ) -> ContextManager[Tuple[Iterable[Tuple[DatasetIdRef,
+                                                      Optional[StoredDatastoreItemInfo]]],
+                                       Set[str]]]:
```
TallJimbo: One solution to at least some of the mypy woes would be to make this function generic over the type that is currently `StoredDatasetItemInfo` here:
```
# Define a type parameter that can only be satisfied by types
# that inherit from StoredDatasetItemInfo.
_T = TypeVar("_T", bound=StoredDatastoreItemInfo)

class DatastoreRegistryBridge:
    ...
    def emptyTrash(self, records_table: Optional[OpaqueTableStorage] = None,
                   record_class: Optional[Type[_T]] = None,
                   record_column: Optional[str] = None,
                   ) -> ContextManager[Tuple[Iterable[Tuple[DatasetIdRef,
                                                      Optional[_T]]],
                                       Set[str]]]:
        ...
```
This (passing a type in explicitly) is a corner of generics I haven't really tried much, though, so I'm not certain it'll actually help.

python/lsst/daf/butler/registry/bridge/monolithic.py
```diff
@@ -185,24 +186,102 @@ def check(self, refs: Iterable[DatasetIdRef]) -> Iterable[DatasetIdRef]:
...
-            self._tables.dataset_location_trash
-        ).where(
-            self._tables.dataset_location_trash.columns.datastore_name == self.datastoreName
-        )
+
+        if records_table is None:
+            raise ValueError("This implementation requires a records table.")
+
+        if not isinstance(records_table, ByNameOpaqueTableStorage):
+            raise ValueError(f"Records table must support hidden attributes. Got {type(records_table)}.")
```
TallJimbo: I wonder if we should just make `OpaqueTableStorage._table` public, or add a public `table` property that returns it.

(Wouldn't change the need for this check, but "hidden" here reminded me that we're using an at least sort-of private attribute).

python/lsst/daf/butler/registry/bridge/monolithic.py
```diff
@@ -185,24 +186,102 @@ def check(self, refs: Iterable[DatasetIdRef]) -> Iterable[DatasetIdRef]:
...
+                location_table.columns.datastore_name == self.datastoreName
+            )
+
+        # SELECT records.dataset_id, records.path FROM records
+        #    JOIN records on dataset_location.dataset_id == records.dataset_id
+        #    WHERE dataset_location.datastore_name = datastoreName
+
+        # It's possible that we may end up with a ref listed in the trash
+        # table that is not listed in the records table. Such an
+        # inconsistency would be missed by this query.
```
TallJimbo, line 226: Possible only in the presence of a logic bug in our code, right?  It's not rigorously guarded by e.g. a direct foreign key, but I hope this isn't a state that a user can get themselves into by using the code correctly (or even by deleting things by hand).

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -230,6 +228,8 @@ def makeTableSpec(cls, datasetIdColumnType: type) -> ddl.TableSpec:
                 ddl.FieldSpec(name="file_size", dtype=BigInteger, nullable=True),
             ],
             unique=frozenset(),
+            indexes=[tuple(["dataset_id", "path"]), tuple(["path"]),
+                     tuple(["dataset_id", "component", "path"])],
```
TallJimbo: Does this last `(dataset_id, component, path)` index yield a measured improvement, too?  I _think_ that could remove the need to actually look in the table when querying for the `path` given the `dataset_id` and `component`, as all the information would be in the index, but the primary key should already be sufficient to still do that in `O(log N)` time but with more overhead, and this is getting to be a lot of indexes.

TallJimbo (approved):
Looks good!  I have one suggestion for the mypy situation that _might_ help (but what you have is ok).  Also a few other minor comments.

I'm in favor of the solution that requires adding the indexes to avoid a regression, but we do need to be careful with how we we stage that.  I think everybody maintaining a big repo could very quickly get a few `CREATE INDEX` commands run, but maybe a community announcement post with the instructions is in order?  I can of course take care of the repos at NCSA, but I am worried about DESC@NERSC, IN2P3, IDF, and OODS.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1647,60 +1633,85 @@ def emptyTrash(self, ignore_errors: bool = True) -> None:
...
+                # Only trashed refs still known to datastore will be returned.
+                location = info.file_location(self.locationFactory)
+
+                # If the file itself has been deleted there is nothing we
+                # can do about it. It is possible that trash has been run
+                # in parallel in another process or someone decided to delete
+                # the file. It is unlikely to come back and so we should still
+                # continue with the removal of the entry from the trash
+                # table.
+                if not self._artifact_exists(location):
```
timj: I think I might remove this check completely and move it to a catch below for FileNotFoundError. That way we save an existence check (which can add up with the S3 interface).

python/lsst/daf/butler/registry/interfaces/_bridge.py
timj: It does not work 👎

python/lsst/daf/butler/datastores/fileDatastore.py
timj: I added it because it seemed like the complement to the first one. From a quick test `path` alone works fine.

python/lsst/daf/butler/registry/bridge/monolithic.py
timj: Yes, it would be a logic bug.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: It's annoying that so many of these are needed.

python/lsst/daf/butler/registry/interfaces/_bridge.py
timj: I tried that and things broke badly. Putting in some additional asserts made things work.

python/lsst/daf/butler/registry/bridge/monolithic.py
timj: The Dummy registry table storage class doesn't have a table at all (it's all a python dict).

TallJimbo: Ok.  A property that can return `None` isn't terribly satisfying, but it may still be better than what we've got.  And yet maybe not enough better to be worth the trouble.

If the QuantumDirectory stuff all works out, that would require adjustments to the opaque-table interface anyway, so I guess we can come back to this then.

PR #519, opened 2021-05-11
DM-30145: Support IN operator in database deletes

timj:
In my quick tests of deleting 5000 datasets from a collection the bridge emptyTrash deletes go from 10 seconds to 0.5 second with this change.

### DIFF
--- a/python/lsst/daf/butler/registry/interfaces/_database.py
+++ b/python/lsst/daf/butler/registry/interfaces/_database.py
@@ -1459,11 +1459,21 @@ class Database(ABC):
                 # The set only has one element
                 clauses.append(column == v.pop())
 
-            # Now the IN operator
-            clauses.append(table.columns[name].in_(content[name]))
+            # The IN operator will not work for "infinite" numbers of
+            # rows so must batch it up into distinct calls.
+            in_content = list(content[name])
+            n_elements = len(in_content)
 
-            sql = sql.where(sqlalchemy.sql.and_(*clauses))
-            return self._connection.execute(sql).rowcount
+            rowcount = 0
+            iposn = 0
+            n_per_loop = 1_000  # Controls how many items to put in IN clause
+            for iposn in range(0, n_elements, n_per_loop):
+                endpos = iposn + n_per_loop
+                in_clause = table.columns[name].in_(in_content[iposn:endpos])
+
+                newsql = sql.where(sqlalchemy.sql.and_(*clauses, in_clause))
+                rowcount += self._connection.execute(newsql).rowcount
+            return rowcount
         else:
             whereTerms = [table.columns[name] == sqlalchemy.sql.bindparam(name) for name in columns]
             if whereTerms:
--- a/python/lsst/daf/butler/registry/interfaces/_database.py
+++ b/python/lsst/daf/butler/registry/interfaces/_database.py
@@ -1447,7 +1447,17 @@ class Database(ABC):
                     content[k].add(v)
             changing_columns = [col for col, values in content.items() if len(values) > 1]
 
-        if len(changing_columns) == 1:
+        if len(changing_columns) != 1:
+            # More than one column changes each time so do explicit bind
+            # parameters and have each row processed separately.
+            whereTerms = [table.columns[name] == sqlalchemy.sql.bindparam(name) for name in columns]
+            if whereTerms:
+                sql = sql.where(sqlalchemy.sql.and_(*whereTerms))
+            return self._connection.execute(sql, *rows).rowcount
+        else:
+            # One of the columns has changing values but any others are
+            # fixed. In this case we can use an IN operator and be more
+            # efficient.
             name = changing_columns.pop()
 
             # Simple where clause for the unchanging columns
@@ -1474,11 +1484,6 @@ class Database(ABC):
                 newsql = sql.where(sqlalchemy.sql.and_(*clauses, in_clause))
                 rowcount += self._connection.execute(newsql).rowcount
             return rowcount
-        else:
-            whereTerms = [table.columns[name] == sqlalchemy.sql.bindparam(name) for name in columns]
-            if whereTerms:
-                sql = sql.where(sqlalchemy.sql.and_(*whereTerms))
-            return self._connection.execute(sql, *rows).rowcount
 
     def update(self, table: sqlalchemy.schema.Table, where: Dict[str, str], *rows: dict) -> int:
         """Update one or more rows in a table.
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -1430,10 +1432,57 @@ def delete(self, table: sqlalchemy.schema.Table, columns: Iterable[str], *rows:
...
+                # The set only has one element
+                clauses.append(column == v.pop())
+
+            # The IN operator will not work for "infinite" numbers of
+            # rows so must batch it up into distinct calls.
+            in_content = list(content[name])
+            n_elements = len(in_content)
+            MAX_ELEMENTS_PER_IN = 1500
+            n_loops = math.ceil(n_elements / MAX_ELEMENTS_PER_IN)
+            n_per_loop = math.ceil(n_elements / n_loops)
```
ktlim: This seems kind of complicated; why not just take MAX_ELEMENTS_PER_IN at a time until you run out?  Then (because you would explicitly check) you also don't have the possibility that endpos is off the end (which I think you have now).

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -1430,10 +1432,57 @@ def delete(self, table: sqlalchemy.schema.Table, columns: Iterable[str], *rows:
...
+            iposn = 0
+            while iposn < n_elements:
+                endpos = iposn + n_per_loop
+                in_clause = table.columns[name].in_(in_content[iposn:endpos])
+                iposn = endpos
+
+                newsql = sql.where(sqlalchemy.sql.and_(*clauses, in_clause))
+                rowcount += self._connection.execute(newsql).rowcount
+            return rowcount
+        else:
```
ktlim: Sometimes it's more readable for the short, default/fallback case to come before the long, specialized case.

python/lsst/daf/butler/registry/interfaces/_database.py
timj: I was trying not to have a situation where there are 5001 rows and 1000 as MAX_ELEMENTS_PER_IN to make it run 6 times when I can make it run 5 times if I use 1001.

timj: Re "endpos off the end" -- I checked and python is perfectly fine with that so I don't need any special logic to handle overrun.

timj: I do realize the logic is not quite right regardless but the loop still covers everything.

python/lsst/daf/butler/registry/interfaces/_database.py
timj: Yes, when I started they were about the same length.

python/lsst/daf/butler/registry/interfaces/_database.py
ktlim: Except I think you still get 6 loops of 834 in the case you gave, not 5 of 1001.  (You'd need to replace the first computation with a `math.floor`.)  What's the use of having a MAX if it's not the maximum?  And is an extra loop/query really that much time (compared with the thousands that you are avoiding)?

ktlim: Just saw you have a new calculation.  It is better, but still a lot of comments and code for not a lot of gain.

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -1430,10 +1432,68 @@ def delete(self, table: sqlalchemy.schema.Table, columns: Iterable[str], *rows:
...
+        columns = list(columns)  # Force iterators to list
+
+        # More efficient to use IN operator if there is only one
+        # variable changing across all rows.
+        content: Dict[str, Set] = defaultdict(set)
+        if len(columns) == 1:
+            # Nothing to calculate since we can always use IN
+            column = columns[0]
+            changing_columns = [column]
+            content[column] = set([row[column] for row in rows])
```
andy-slac: No need to make `list` before turning it into `set`, iterable is OK.

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -1430,10 +1432,68 @@ def delete(self, table: sqlalchemy.schema.Table, columns: Iterable[str], *rows:
...
+        else:
+            for row in rows:
+                for k, v in row.items():
+                    content[k].add(v)
+            changing_columns = [col for col in content if len(content[col]) > 1]
```
andy-slac: ` [col for col, values in content.items() if len(values) > 1]` for extra efficiency?

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -1430,10 +1432,68 @@ def delete(self, table: sqlalchemy.schema.Table, columns: Iterable[str], *rows:
...
+            if n_loops == 0:
+                n_loops = 1
+            n_per_loop = math.ceil(n_elements / n_loops)
+
+            rowcount = 0
+            iposn = 0
+            while iposn < n_elements:
+                endpos = iposn + n_per_loop
+                in_clause = table.columns[name].in_(in_content[iposn:endpos])
+                iposn = endpos
```
andy-slac: It maybe easier to read as:

    for iposn in range(0, n_elements, n_per_loop):
        endpos = iposn + n_per_loop

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -1430,10 +1432,68 @@ def delete(self, table: sqlalchemy.schema.Table, columns: Iterable[str], *rows:
...
+
+                newsql = sql.where(sqlalchemy.sql.and_(*clauses, in_clause))
+                rowcount += self._connection.execute(newsql).rowcount
```
andy-slac, line 1485: With multiple DELETEs I think we want to run all of them in the same transactions, so this loop need a transaction context around it.

andy-slac (approved):

andy-slac:
Sorry, forgot to add comment before clicking Approved. Looks OK, but check my comment about transactions, I think we need to wrap it into a single transaction.

python/lsst/daf/butler/registry/interfaces/_database.py
timj: Yes. Whilst doing the school run I realized which way I had messed it up. I can remove the complication -- I just felt like I preferred a more even spread...

python/lsst/daf/butler/registry/interfaces/_database.py
andy-slac: I am wrong about that, transactions are managed at higher level, this code requires that it is called within transaction context but we do not want to start transactions here.

PR #521, opened 2021-05-12
DM-30140: Allow trash to take iterable of dataset refs

timj:
This is much much faster than calling trash on each ref separately. Also remove the existence check since if we are asking for a dataset to be trashed it is not really that important if the dataset has already disappeared and that could just be caused by someone else pruning a collection in parallel.

### DIFF
--- a/python/lsst/daf/butler/datastores/chainedDatastore.py
+++ b/python/lsst/daf/butler/datastores/chainedDatastore.py
@@ -45,10 +45,10 @@ from typing import (
 
 from lsst.utils import doImport
 from lsst.daf.butler import ButlerURI, Datastore, DatastoreConfig, DatasetTypeNotSupportedError, \
-    DatastoreValidationError, Constraints, FileDataset
+    DatastoreValidationError, Constraints, FileDataset, DatasetRef
 
 if TYPE_CHECKING:
-    from lsst.daf.butler import Config, DatasetRef, DatasetType, LookupKey, StorageClass
+    from lsst.daf.butler import Config, DatasetType, LookupKey, StorageClass
     from lsst.daf.butler.registry.interfaces import DatastoreRegistryBridgeManager
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


python/lsst/daf/butler/datastores/genericDatastore.py
```diff
@@ -148,7 +148,7 @@ def _move_to_trash_in_registry(self, ref: DatasetRef) -> None:
         # Note that a ref can point to component dataset refs that
         # have been deleted already from registry but are still in
         # the python object. moveToTrash will deal with that.
-        self.bridge.moveToTrash([ref])
+        self.bridge.moveToTrash(refs)
```
TallJimbo: Might be time to just remove `_move_to_trash_in_registry` and call this method directly.

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -629,8 +629,13 @@ def forget(self, refs: Iterable[DatasetRef]) -> None:
         for datastore in tuple(self.datastores):
             datastore.forget(refs)
 
-    def trash(self, ref: DatasetRef, ignore_errors: bool = True) -> None:
-        log.debug("Trashing %s", ref)
+    def trash(self, ref: Union[DatasetRef, Iterable[DatasetRef]], ignore_errors: bool = True) -> None:
+        if isinstance(ref, DatasetRef):
+            ref_label = str(ref)
+        else:
+            ref_label = "bulk datasets"
```
TallJimbo, line 636: Might be nice to put `len(ref)` in here.

TallJimbo (approved):

python/lsst/daf/butler/datastores/chainedDatastore.py
timj: Wouldn't that be a problem if ref was a generator? I assumed that `len()` would trigger the generator and empty it for later.

TallJimbo: Oh, good point.  If it's a generator or other lazy iterator `len` wouldn't work at all. Nevermind.

PR #523, opened 2021-05-18
DM-28650: Add Butler.transfer_from

timj:


python/lsst/daf/butler/_butler.py
```diff
@@ -1670,6 +1670,78 @@ def doImport(importStream: TextIO) -> None:
...
+        Notes
+        -----
+        Requires that any dimension definitions are already present in the
+        receiving Butler. The datastore artifact has to exist for a transfer
+        to be made but non-existence is not an error.
+
+        The source collection will be reconstructed in this butler using
+        the same names. If a dataset already exists in a RUN collection
+        of the same name no transfer will occur and this will not be an error
+        unless the DatasetRef uses UUID and there is a UUID mismatch.
```
TallJimbo: "dataset already exists" means "a dataset with the same dataset type and data ID exists", then?

I don't understand the reasoning for the UUID behavior exception, in that case.

python/lsst/daf/butler/_butler.py
```diff
@@ -1670,6 +1670,78 @@ def doImport(importStream: TextIO) -> None:
...
+    def transfer_from(self, source_butler: Butler, collection: str,
+                      transfer: str = "auto") -> List[DatasetRef]:
+        """Transfer datasets from a run in another Butler.
+
+        Parameters
+        ----------
+        source_butler : `Butler`
+            Butler from which the datasets are to be transferred.
+        collection : `str`
+            Name of the collection to transfer.
```
TallJimbo: Transferring only complete collections seems pretty limiting.  Could we make this take an iterable of input `DatasetRef` to transfer?

I don't want to get in the way of providing something useful, which this already is, but I am worried about it interfacing-squatting on something we'll ultimately want to be much more flexible.

python/lsst/daf/butler/_butler.py
```diff
@@ -1670,6 +1670,78 @@ def doImport(importStream: TextIO) -> None:
...
+                                         source_collection.type,
+                                         doc=source_collection.key)
+
+        # Get refs that have associated datastore artifacts in the source
+        refs_to_transfer = [ref for ref in source_butler.registry.queryDatasets(..., collections=collection)
+                            if source_butler.datastore.exists(ref)]
+
+        log.debug("Will be transferring %d datasets", len(refs_to_transfer))
+
+        # For bulk import they need to be in the same run. Assume that for now.
```
TallJimbo: I assume this is because `Registry._importDatasets` can only do one run at a time?  That might now be an artificial limitation, coming from:
 - `DatasetRef` can have both `id` and `run` or neither;
 - before UUIDs it didn't make sense at all to try to import datasets that already had IDs.

I'm not sure this is the ticket to fix it, but we should come back to this after autoincrement IDs are gone, if not sooner.

In the meantime, what happens if this assumption is violated?  Doesn't look like we raise an exception, unless it's coming from lower down.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1723,6 +1748,130 @@ def emptyTrash(self, ignore_errors: bool = True) -> None:
...
+
+        # Be explicit for mypy
+        if not isinstance(source_datastore, FileDatastore):
+            raise TypeError("Can only transfer to a FileDatastore from another FileDatastore, not"
+                            f" {type(source_datastore)}")
+
+        # Stop early if "direct" transfer mode is requested. That would
+        # require that the URI inside the source datastore should be stored
+        # directly in the target datastore, which seems unlikely to be useful
+        # since at any moment the source datastore could delete the file.
```
TallJimbo, line 1768: I actually think there are lots of cases where this is not a concern.  If I'm transferring out of a big shared repo into a local one that's on the same filesystem (or can see the same object store), I may well want to transfer a lot of raws or data release products with "direct" because I don't need my own copies and I can reasonably assume they won't be deleted.  But this starts to blur the boundary with some chained datastore use cases, so if we have a story for how to do that kind of thing without `transfer_from`, that's fine, except for making documenting it.

(FWIW, the chained datastore solution is at least a little messy because both the shared DB and my local registry are going to have a `dataset_location` table that thinks it is canonical, at least the way things stand now.)

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1723,6 +1748,130 @@ def emptyTrash(self, ignore_errors: bool = True) -> None:
...
+        #   destination butler. May need to have a flag to indicate whether
+        #   the dataset should be transferred. This will only happen if
+        #   the detached Butler has had a local ingest.
+
+        # What we really want is all the records in the source datastore
+        # associated with these refs. Or derived ones if they don't exist
+        # in the source.
+        source_records = source_datastore._get_stored_records_associated_with_refs(refs)
+
+        source_ids = set(source_records)
```
TallJimbo, line 1801: Something seems wrong here (maybe just variable names) from a type perspective.  Are these IDs or records?

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1723,6 +1748,130 @@ def emptyTrash(self, ignore_errors: bool = True) -> None:
...
+
+        # Now can transfer the artifacts
+        for source_ref, target_ref in zip(refs, target_refs):
+            if target_ref.id in target_records:
+                # Already have an artifact for this.
+                already_present.append(target_ref)
+                continue
+
+            # mypy needs to know these are always resolved refs
+            assert source_ref.id is not None, "Unexpectedly got unresolved ref inside datastore"
```
TallJimbo: `DatasetRef.getCheckedId` is designed for this.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1723,6 +1748,130 @@ def emptyTrash(self, ignore_errors: bool = True) -> None:
...
+                target_location = info.file_location(self.locationFactory)
+                if source_location == target_location:
+                    # Either the dataset is already in the target datastore
+                    # (which is how execution butler currently runs) or
+                    # it is an absolute URI.
+                    if source_location.pathInStore.isabs():
+                        # Just because we can see the artifact when running
+                        # the transfer doesn't mean it will be generally
+                        # accessible to a user of this butler. For now warn
+                        # but assume it will be accessible.
```
TallJimbo, line 1854: I'm not sure whether this should be an exception or silent, but warning seems like it's still going to be a problem if it's user error, while being annoying if the user knows what they are doing.  Do we want a flag so the user can say, "trust me on this"?

TallJimbo (approved):
Looks okay; lots of comments, but many of them are future concerns that need at most minor changes now.  I'm all for solutions that get most of this on master sooner rather than later, though, as it already looks super useful.

In addition to the line comments, I am a bit worried that the logic for how to handle the many kinds of conflicts is going to be very hard for the user to reason about; that's an intrinsic problem at some level, but it might be better solved demanding more from the user about how they want conflicts to be resolved (maybe even a config object of some kind), rather than guessing.

python/lsst/daf/butler/_butler.py
timj: I think for importDatasets if the ref already exists in the repo it silently skips it. If somehow the UUID does not match the ref in the registry but does match the DataId/DatasetType then it will complain won't it? We can imagine a scenario like this if people do a transfer, then prune their copy and rerun and forget and transfer again. Hence the comment about it being fine if the UUID is used and matches but not if the UUID differs.

python/lsst/daf/butler/_butler.py
timj: So leave the command line at the run level but change the butler api to iterable of refs -- got it. This does mean the caller has to specify the target run not the source run but does mean I don't have to care about chained collections or calibration collections.

python/lsst/daf/butler/_butler.py
timj: `_importDatasets` requires that I group by dataset type as well as only using a single run.

timj: Which is no longer a problem if transfer_run is changed to take refs and one target run.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: I'm really uncomfortable with a situation where a butler datastore has an absolute URI stored in it to a datastore owned by another butler. I am not talking about the case where the source datastore has an absolute URI pointing to somewhere else (such as a file system of raws) -- that is handled below when I look at the actual records and see an absolute URI in the source and have to decide whether I transfer that or use the absolute URI. "direct" here means store absolute URIs to the datasets in the source datastore.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: The keys of source_records are the ids.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: In my head this scenario was extremely unlikely because it only happens with ingested datasets and not those created by processing. Since the transfer does not (yet?) copy dimension records over and the receiving registry won't ever have seen these raw files it seems like we aren't really going to get this far in the code. If we turn this into the default mode for transferring content from one butler to another (and only use export/import for cases where you can't see both butlers from your one python process) then yes, we need the code higher up to insert missing dimension records and then ask the user whether absolute URIs should be transferred or left as-is. I was putting in a warning there as a stop-gap scenario where I'm not expecting it to happen but I don't want to raise an exception immediately -- but warning will generally prompt people to ask us what the morning means.

python/lsst/daf/butler/datastores/fileDatastore.py
TallJimbo: This is why I always write `.keys()`.  But I guess I'm in the minority.

timj: `keys()` seems redundant to me -- we know that when you iterate over a `dict` you get all the keys automatically.

timj: I added a comment in the line above.

python/lsst/daf/butler/_butler.py
```diff
@@ -1673,6 +1673,119 @@ def doImport(importStream: TextIO) -> None:
...
+            id_generation_mode = default_id_gen
+            if isinstance(refs_to_import[0].id, int):
+                # ID generation mode might need to be overridden when
+                # targetting UUID
+                id_generation_mode = id_gen_map.get(datasetType.name, default_id_gen)
+
+            log.debug("Importing %d refs with %s into run %s",
+                      len(refs_to_import), datasetType.name, run)
+
+            # No way to know if this butler's registry uses UUID.
```
TallJimbo, line 1755: Not sure it's a job for this ticket, but I do think we want a public API for this.  It's not just an implementation detail; it's a behavioral change that's important to a lot of higher-level code.

TallJimbo (approved):

python/lsst/daf/butler/_butler.py
timj: It's a transitory problem to a certain extent. Things break if you try to use UUID in an old registry and if I knew that I could strip the ids before calling the import. The question is whether we actually care about that kind of migration. One question I have is whether you are thinking of this as a long-term API or a quick hack. I could easily see a "private" attribute on the dataset manager that we retire when we retire integer IDs. Do you envisage a more public API on `Registry` that will return the `dataset_id_type` even though in the long term that would presumably be the type of `DatasetRef.id`?

TallJimbo: I think totally public - things in pipe_base and maybe obs_base do care (thinking especially about Quantum and QuantumGraph because I'm working on those now).
I imagine we would probably deprecate this API when we deprecate `int` support, but that depends on how much we want to hedge against changing again in the future.

If we wanted to keep it around forever, we might want to think about whether there should be some other flags, in addition to the type; those might last longer:
- `has_globally_unique_ids`
- `supports_deterministic_ids`

Of course, if we expect both of those to always be true forever after `int` is gone, then maybe just the `int` vs `uuid` type flag is enough.

timj: Okay. I'll do this on another ticket because it will also lead to a cleanup of obs_base ingest-raws.

PR #525, opened 2021-05-19
DM-29370: add 'amp' and 'detector' parameters to Exposure storage class

TallJimbo:


czwa (approved):
I assume that the lint warnings will still allow it to pass?

TallJimbo:
> I assume that the lint warnings will still allow it to pass?

I think those must be new since I made the PR.  I'll look into and fix them.

python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -35,8 +35,12 @@ storageClasses:
     pytype: lsst.afw.image.Image
     delegate: lsst.obs.base.exposureAssembler.ExposureAssembler
     parameters:
-      - bbox
-      - origin
+      - bbox    # Load a subimage defined by this box (`lsst.afw.geom.Box2I`).
+
+      - origin  # Coordinate system for ``bbox`` (`lsst.afw.image.ImageOrigin`)
+                #
```
timj: yamllint doesn't like this because it thinks of this line as being over-indented: https://yamllint.readthedocs.io/en/latest/rules.html#module-yamllint.rules.comments_indentation

timj: I'm sure it wants you to do something like:

```yaml
parameters:
  # Load subimage
  - bbox
  # Coordinate system
  #
  # blah
  - origin
```

TallJimbo: Thanks, should be fixed.  And it wasn't new since the PR; I just hadn't noticed the lint warning.  Did both of you just dig into the GHA checks on your own to spot those?  It all looked green to me, and I'd have never have seen the warnings if I wasn't looking for them, so I suspect I'm not looking at some important indicator.

timj: I think GitHub is doing some magic with YAML files and now surfacing the warnings. I don't know if it does it for all YAML files or ones for repos that have a yamllint config file or ones that run actions but it is definitely a newish feature.

timj: the github action only fails for errors and not warnings.

timj: I see there is a `-s` option to `yamllint` for strict mode. Maybe we should turn that on in the action.

TallJimbo: Oh, now I see what you saw - there's a box in the `Files changed` tab that I didn't see in the `Conversations` tab, and the latter is all I'd looked at until now.

PR #527, opened 2021-05-21
DM-30286: Add VERBOSE log level

timj:


andy-slac (approved):
Looks good. I am a little bit worried about defining new logging levels at library level, this has a potential for colliding with other libraries (if we sell daf_butler to outside world) which can cause some unhappiness.

timj:
I'm not entirely sure what else I can do. You are correct that this does globally register "VERBOSE" with the python logging system but the python logging system seems to leave little choice. There is no way to add a per-logger level name. If I do not call addLevelName then the log output says "Level 15" instead of "VERBOSE". By definition that can't affect any other log user who is setting level names but it could mean our "level 15" gets renamed by some other bit of code.

andy-slac:
I agree there is little else we can do here. Let's hope that other people are not doing the same thing as we do or they are not using Python logging. Or we could make call to addLevelName optional based on some documented envvar name, so that in case of collision it would be possible to disable it?

timj:
Adding an env var filter is something we can add later for some future complaint.

PR #529, opened 2021-05-26
DM-30335: Add special flag to support put on execution butler

timj:
`allow_put_of_predefined_dataset` can be set to `True` to allow a put to work even if the registry thinks it knows about the entry. The put will still fail if datastore knows about the dataset.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -245,6 +245,7 @@ class Butler:
             self.datastore = butler.datastore
             self.storageClasses = butler.storageClasses
             self._config: ButlerConfig = butler._config
+            self._allow_put_of_predefined_dataset = butler._allow_put_of_predefined_dataset
         else:
             self._config = ButlerConfig(config, searchPaths=searchPaths)
             if "root" in self._config:
@@ -259,6 +260,7 @@ class Butler:
                                                   butlerRoot=butlerRoot)
             self.storageClasses = StorageClassFactory()
             self.storageClasses.addFromConfig(self._config)
+            self._allow_put_of_predefined_dataset = self._config.get("allow_put_of_predefined_dataset", False)
         if "run" in self._config or "collection" in self._config:
             raise ValueError("Passing a run or collection via configuration is no longer supported.")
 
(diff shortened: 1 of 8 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -855,7 +857,32 @@ def put(self, obj: Any, datasetRefOrType: Union[DatasetRef, DatasetType, str],
...
+        # For an execution butler the datasets will be pre-defined.
+        # If the butler is configured that way datasets should only be inserted
+        # if they do not already exist in registry. Trying and catching
+        # ConflictingDefinitionError will not work because the transaction
+        # will be corrupted. Instead, in this mode always check first.
+        ref = None
+        if self._allow_put_of_predefined_dataset:
+            # Get the matching ref for this run.
+            existing_refs = list(self.registry.queryDatasets(datasetType, collections=run,
+                                                             dataId=dataId))
```
TallJimbo: I'd use `Registry.findDataset` instead here; that will be more efficient, and it will always return either a resolved ref or `None`.  Getting more than that should be impossible anyway due to database unique constraints.

I also think it'd be a lot cleaner if we actually put the resolved output refs in the QuantumGraph and only enabled this logic if the ref was already resolved (or even just called `butler.datastore.put`).  But that involves making the execution butler before we actually write out the QG (or generating UUIDs for predicted datasets out the butler, once we can assume UUIDs everwhere), so it's probably not worth worrying about right now.

TallJimbo (approved):

python/lsst/daf/butler/_butler.py
timj: Yes, Sorry. Will switch. I remembered getDataset but that needed the ID.  

I also just had to push a change that is bigger than I expected because writing the quick test demonstrated that calling datastore.put leads to the data file being deleted if it goes wrong and that is very bad if you were relying on the file already being there (and it also overwrites it). I've added a `datastore.knows()` API to work around that.

TallJimbo: New changes look good, too, though all of this special handling for execution butlers reinforces my thinking that we want to do something (like QuantumDirectory) to let us drop it someday.  But not today.  Merge away, whenever you're ready.

PR #532, opened 2021-06-07
DM-30653: Add v22 release notes

timj:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/doc/lsst.daf.butler/CHANGES.rst
+++ b/doc/lsst.daf.butler/CHANGES.rst
@@ -1,0 +1,47 @@
+Butler v22.0 2021-04-01
+=======================
+
+New Features
+------------
+
+* A Butler instance can now be configured with dataId defaults such as an instrument or skymap. [DM-27153]
+* Add ``butler prune-datasets`` command. [DM-26689]
+* Add ``butler query-dimension-records`` command [DM-27344]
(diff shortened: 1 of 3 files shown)


doc/lsst.daf.butler/CHANGES.rst
```diff
@@ -0,0 +1,61 @@
...
+* Allow ``butler.get()`` to support dimension record values such as exposure observing day or detector name in the dataID. [DM-27152]
+* Add "direct" ingest mode to allow a file to be ingested retaining the full path to the original file. [DM-27478]
+
+Bug Fixes
+---------
+
+* Fix temporal queries and clarify ``Timespan`` behavior. [DM-27985]
+
+Performance Enhancements
+------------------------
```
TallJimbo: Can we configure towncrier to drop these empty headings?

TallJimbo (approved):
The API changes I see here are (I think) all backwards compatible, i.e. they're new kwargs in old methods or new methods, and hence I think I would have called them new features instead.  Do we have different concepts of what API change means, or am I just mistaken about whether these were backwards compatible?

doc/lsst.daf.butler/CHANGES.rst
timj: I wrote these manually. This was me forgetting to remove the placeholders I had put in.

PR #533, opened 2021-06-09
DM-30373: Add butler collection-chain command line

timj:
## Checklist

- [ ] ran Jenkins


TallJimbo (approved):
Looks good.  If you're up for a little more work, a `--remove X` option that just does
```
children = list(registry.getCollectionChain(parent))
children.remove(X)
registry.setCollectionChain(parent, children)
```
and a `--pop` that does
```
children = registry.getCollectionChain(parent)[1:]
registry.setCollectionChain(parent, children)
```
would be super useful, too.

timj:
@TallJimbo I've added this but you should take a look before I merge.

Did you mean `--pop` to remove the first entry in the chain? This goes against the python `list.pop()` which removes the final entry.

Rather than add a `--remove X` option I added a `--mode` option. Doing this allowed `X` to be the `CHILDREN` that he subcommand was already having to handle and meant that I could have an `extend` mode as well as a `remove` mode.

python/lsst/daf/butler/cli/cmd/commands.py
```diff
@@ -497,3 +497,33 @@ def transfer_datasets(**kwargs):
     """
     number = script.transferDatasets(**kwargs)
     print(f"Number of datasets transferred: {number}")
+
+
+@click.command(cls=ButlerCommand)
+@repo_argument(required=True)
+@click.argument("parent", required=True, nargs=1)
+@click.argument("children", required=False, nargs=-1)
+@click.option("--pop", is_flag=True, default=False,
```
timj: @TallJimbo Could also do this as `--mode=pop` and let `CHILDREN` be given as integer indexes into the chain (so as to match `list.pop`) -- defaulting to pop off the end.

python/lsst/daf/butler/cli/cmd/commands.py
```diff
@@ -497,3 +497,33 @@ def transfer_datasets(**kwargs):
...
+              "Only relevant if the collection is newly created.")
+@click.option("--flatten/--no-flatten", default=False,
+              help="If `True` recursively flatten out any nested chained collections in children first.")
+@click.option("--mode",
+              type=click.Choice(["redefine", "extend", "remove"]),
+              default="redefine",
+              help="Update mode: "
+              "'redefine': Create new chain or redefine existing chain with the supplied CHILDREN. "
+              "'remove': Modify existing chain to remove the supplied CHILDREN. "
+              "'extend': Modify existing chain to extend it with the supplied CHILDREN.")
```
timj, line 520: Would `--mode=prepend` be useful?

TallJimbo: Yes, and more useful than `extend`, I think.  We may represent these things as Sequences in Python, but conceptually we (`pipetask` in particular) really treat them as LIFO stacks, where all of the mutations happen at the front (yeah, maybe I should have implemented them as reversed-lists in Python, but I figured ordering the container according to the search order is more important).

So I really did mean for `--pop` to act on the front, and I really only care that we provide ways to act on the front in general, but I could easily believe that also providing ways to act on the end is the best way to make it clear what all of the actions actually do.

python/lsst/daf/butler/cli/cmd/commands.py
TallJimbo: (see other thread - I think the difference here is that I don't think popping off the end is a good default)

timj: In Perl you `shift` to move things off the front of the list and `pop` to remove things from the end. I do worry that calling this `pop` goes against the default python definition of pop. Maybe something like:

```
$ butler collection-chain REPO --mode=pop PARENT index index index
```
would work where we default `index` to `-1`? (and since this is using `CHILDREN` allow multiple indices).

TallJimbo: How about `--mode=pop --side=front`, or `--mode=popfront`?

timj: I've just implemented the `--mode=pop` defaulting to element 0 and allowing `--mode=pop -1` to remove from the end.

timj: The main problem actually is that `-1` is an option so you have to do `-- -1` to make it work. From what you say that's not going to be a common problem and the positive integer always works as desired.

PR #534, opened 2021-06-10
DM-30685: Issue warning if lookup key dimension is not known to universe

timj:
Previously a confusing KeyError would be issued with the unknown key.
Now replace this with a warning message suggesting that there might be a typo
but still continue on without stopping.

This is mainly motivated by the presence of physical_filter in the
fileDatastore file template. Since this is in a default configuration
it affects all users, even those that define their own universe.

DM-26190 may give us the real fix for the fileDatastore problem.

## Checklist

- [x] ran Jenkins


tgoldina (approved):
Thank you for a quick fix. I've tested it, works for us.

PR #535, opened 2021-06-14
DM-30743: Defer creation of temporary datastore cache directory

timj:
Previously we always created it every single time someone instantiated
a Butler!

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-30743-bugfix.rst
+++ b/doc/changes/DM-30743-bugfix.rst
@@ -1,0 +1,2 @@
+Stop writing a temporary datastore cache directory every time a ``Butler`` object was instantiated.
+Now only create one when one is requested.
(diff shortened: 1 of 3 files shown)


erykoff (approved):
Looks reasonable, but you need to fix those mypy complaints.

Also, I'm curious how the `else` block was getting triggered before.  I guess sometimes `self.config.get("root")` is `None` even in regular usage?

timj:
Yes, the mypy has to be fixed. I never remember to run it locally before I push.

We were always creating the cache dir. Every single time you typed `butler` on the command line it was making that temp directory... No-one is specifying a special cache directory in there datastore config.

erykoff:
Does this need a specific test, or I guess it's incidentally tested by code that uses and doesn't use a cache?

timj:
The cache code is only used for remote datastores. We haven't broken local datastores. Things only break for remote datastores with the y-background files. I'll add a dataset type to one of the test configs that will be run by S3 to make sure it's not fundamentally broken.

PR #536, opened 2021-06-15
DM-30761: Forward 'bind' keyword argument in queryDatasets so it actually works.

TallJimbo:
## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/registries/sql.py
```diff
@@ -870,6 +870,7 @@ def queryDatasets(self, datasetType: Any, *,
                     dimensions=dimensions,
                     dataId=standardizedDataId,
                     where=where,
+                    bind=bind,
```
timj, line 873: Oops 😄 

I admit I had looked for some examples of `bind` recently and didn't find many.

PR #537, opened 2021-06-16
DM-30771: Allow components to be missing in datastore trust mode

timj:
When disassembling a composite we can't assume that all components
will exist. In trust mode assume that any components found means
that these should be passed on to the assembler.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -584,6 +584,20 @@ class FileDatastore(GenericBaseDatastore):
 
         if len(fileLocations) > 1:
             disassembled = True
+
+            # If trust is involved it is possible that there will be
+            # components listed here that do not exist in the datastore.
+            # Explicitly check for file artifact existence and filter out any
+            # that are missing.
+            if self.trustGetRequest:
+                fileLocations = [loc for loc in fileLocations if loc[0].uri.exists()]
+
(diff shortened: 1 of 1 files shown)


python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -584,6 +584,20 @@ def _prepare_for_get(self, ref: DatasetRef,
 
         if len(fileLocations) > 1:
             disassembled = True
+
+            # If trust is involved it is possible that there will be too
+            # many components listed here. We need to explicitly check
```
MichelleGower: I don't understand this first sentence.  Is it because some components appear in the execution butler because it has everything for the run but some components will not be written until a later Quantum execution?  The rest of the comment and the code makes sense.

MichelleGower (approved):
Just a question about a comment in the code.  Ran a quick bps test turning back on disassembly for pipelines_check and it worked.  Changes approved for merging.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: Registry has no concept of components so nothing in registry can tell me whether all components were written or only some of them. On read if you ask for a `calexp.wcs` it will know to look for exactly the `wcs` component so will fail or not depending on the presence of the file.

PR #538, opened 2021-06-17
DM-30784: Fixes to transfer_from to correctly filter out missing datasets and register dataset types

timj:
Turns out that I had forgotten a rather critical filter and was transferring all datasets, even those that had not file artifact. This is now fixed. This PR also adds dataset type registration.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1705,7 +1705,6 @@ class Butler:
         else:
             doImport(filename)
 
-    @transactional
     def transfer_from(self, source_butler: Butler, source_refs: Iterable[DatasetRef],
                       transfer: str = "auto",
                       id_gen_map: Dict[str, DatasetIdGenEnum] = None) -> List[DatasetRef]:
@@ -1738,6 +1737,11 @@ class Butler:
         to be made but non-existence is not an error.
 
         Datasets that already exist in this run will be skipped.
+
+        The datasets are imported as part of a transaction, although
+        dataset types are registered before the transaction is started.
+        This means that it is possible for a dataset type to be registered
+        even though transfer has failed.
         """
         if not self.isWriteable():
             raise TypeError("Butler is read-only.")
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1849,13 +1859,25 @@ def transfer_from(self, source_datastore: Datastore, refs: Iterable[DatasetRef],
...
+
+                # Not all components can be guaranteed to exist so this
+                # list has to filter those by checking to see if the
+                # artifact is really there.
+                records = [info for location, info in expected if location.uri.exists()]
+                if records:
+                    source_records[missing].extend(records)
+                else:
+                    log.warning("Asked to transfer dataset %s but no file artifacts exist for it.",
+                                missing)
```
andy-slac: Maybe also print `id_to_ref[missing]` for more complete diagnostics if it can help?

python/lsst/daf/butler/_butler.py
```diff
@@ -1755,67 +1763,94 @@ def transfer_from(self, source_butler: Butler, source_refs: Iterable[DatasetRef]
...
+                if not source_butler.datastore.exists(ref):
+                    continue
+
+            accepted_source_refs.append(ref)
             grouped_refs[ref.datasetType, ref.run].append(ref)
             grouped_indices[ref.datasetType, ref.run].append(i)
 
+            # Update counter since enumerate cannot be used since we
+            # are filtering the refs.
+            i += 1
```
andy-slac: I think you can make logic somewhat simpler and keep `enumerate` if you filter separately:

    if skip_missing:
        source_refs = [ref for ref in source_refs if source_butler.datastore.exists(ref)]

python/lsst/daf/butler/_butler.py
```diff
@@ -1705,10 +1705,10 @@ def doImport(importStream: TextIO) -> None:
         else:
             doImport(filename)
 
-    @transactional
```
andy-slac, line 1708: What is the reason you cannot keep `@transactional` here?

andy-slac (approved):
Looks good, couple of minor comments/questions.

python/lsst/daf/butler/_butler.py
timj: Because we register dataset types and we explicitly disallow that inside an existing transaction.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: I was pondering that. The UUID is enough to find it but more information won't hurt.

PR #542, opened 2021-06-23
DM-30718: Do not recreate DataCoordinate on butler.get if given a DataCoordinate

timj:
Previously a fully expanded DataCoordinate would be turned into a minimalist dict before converting it back to a DataCoordinate in an attempt to handle the detector/exposure record specifications in dataIds.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -533,56 +533,52 @@ class Butler:
         assert internalDatasetType is not None
         return internalDatasetType, dataId
 
-    def _findDatasetRef(self, datasetRefOrType: Union[DatasetRef, DatasetType, str],
-                        dataId: Optional[DataId] = None, *,
-                        collections: Any = None,
-                        allowUnresolved: bool = False,
-                        **kwds: Any) -> DatasetRef:
-        """Shared logic for methods that start with a search for a dataset in
-        the registry.
+    def _rewrite_data_id(self, dataId: Optional[DataId], datasetType: DatasetType,
+                         **kwds: Any) -> Tuple[Optional[DataId], Dict[str, Any]]:
+        """Rewrite a data ID taking into account dimension records.
+
+        Take a Data ID and keyword args and rewrite it if necessary to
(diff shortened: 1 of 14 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -533,56 +533,52 @@ def _standardizeArgs(self, datasetRefOrType: Union[DatasetRef, DatasetType, str]
...
-            Raised if a resolved `DatasetRef` was passed as an input, but it
-            differs from the one found in the registry.
-        TypeError
-            Raised if no collections were provided.
+        dataId : `dict` or `DataCoordinate`
+            The, possibly rewritten, dataId. If given a `DataCoordinate` and
+            no keyword arguments, the orginal dataId will be returned
+            unchanged.
+        kwds : `dict`
+            Any unused keyword arguments.
```
TallJimbo, line 577: I've been told to switch to `kwargs` in new code, and I think we usually put the leading `**` in the docs now, too.

python/lsst/daf/butler/_butler.py
```diff
@@ -776,6 +767,61 @@ def quote(s: Any) -> str:
...
+        collections : Any, optional
+            Collections to be searched, overriding ``self.collections``.
+            Can be any of the types supported by the ``collections`` argument
+            to butler construction.
+        allowUnresolved : `bool`, optional
+            If `True`, return an unresolved `DatasetRef` if finding a resolved
+            one in the `Registry` fails.  Defaults to `False`.
+        kwds
+            Additional keyword arguments used to augment or construct a
+            `DataId`.  See `DataId` parameters.
```
TallJimbo, line 798: Same doc comments as above.

TallJimbo (approved):
Looks good. Only comments are about fixing doc style issues that are probably my fault originally, since you're touching those lines anyway (or at least `git diff` thinks you are).

python/lsst/daf/butler/_butler.py
timj: Yes, this was your code moved around. Will change it.

PR #545, opened 2021-07-01
DM-30935: Add butler ingest-files subcommand

timj:
## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/_butler.py
```diff
@@ -1504,7 +1504,7 @@ def ingest(self, *datasets: FileDataset, transfer: Optional[str] = "auto", run:
         ----------
         datasets : `FileDataset`
             Each positional argument is a struct containing information about
-            a file to be ingested, including its path (either absolute or
+            a file to be ingested, including its URI (either absolute or
```
kfindeisen, line 1507: Was this a documentation error, or is this an API change to `Butler.ingest`? If the former, I suggest making it its own commit, for clarity.

python/lsst/daf/butler/cli/cmd/commands.py
```diff
@@ -539,3 +541,55 @@ def collection_chain(**kwargs):
...
+@click.argument("table_file", required=True)
+@click.option("--formatter", type=str,
+              help="Fully-qualified python class to use as the Formatter. If not specified the formatter"
+              " will be determined from the dataset type and datastore configuration.")
+@click.option("--id-generation-mode",
+              default="UNIQUE",
+              help="Mode to use for generating dataset IDs. The default creates a unique ID. Other options"
+              " are: 'DATAID_TYPE' for creating a reproducible ID from the dataID;"
+              " 'DATAID_TYPE_RUN' for creating a reproducible ID from the dataID and run.  The latter is"
+              " usually used for 'raw'-type data.",
```
kfindeisen: Is the dataset ID something users need to understand, or can they just assume that `UNIQUE` is acceptable for all non-`raw` datasets?

python/lsst/daf/butler/cli/cmd/commands.py
```diff
@@ -539,3 +541,55 @@ def collection_chain(**kwargs):
...
+              " comma separation or multiple command line options.")
+@click.option("--prefix",
+              type=str,
+              help="For relative paths in the table file, specify a prefix to use. The default is to"
+              " use the current working directory.")
+@transfer_option()
+def ingest_files(**kwargs):
+    """Ingest files from table file.
+
+    DATASET_TYPE is the name of the dataset type to associated with these
```
kfindeisen: Typo.

python/lsst/daf/butler/cli/cmd/commands.py
```diff
@@ -539,3 +541,55 @@ def collection_chain(**kwargs):
...
+              " files. This can be used to simplify the table file by removing repeated entries that are"
+              " fixed for all files to be ingested.  Multiple key/values can be given either by using"
+              " comma separation or multiple command line options.")
+@click.option("--prefix",
+              type=str,
+              help="For relative paths in the table file, specify a prefix to use. The default is to"
+              " use the current working directory.")
+@transfer_option()
+def ingest_files(**kwargs):
+    """Ingest files from table file.
```
kfindeisen, line 576: Given the comment in Jira that this tool does not create dimension records, should there be a disclaimer in the help text? I think most people would expect that functionality from the name "ingest".

python/lsst/daf/butler/script/ingest_files.py
```diff
@@ -0,0 +1,140 @@
...
+        URI string of the Butler repo to use.
+    dataset_type : `str`
+        The name of the dataset type for the files to be ingested.
+    formatter : `str`
+        Fully-qualified python class name for the `Formatter` to use
+        to read the ingested files.
+    run : `str`
+        The run in which the files should be ingested.
+    table_file : `str`
+        Path to a table file to read.
```
kfindeisen: Consider restating that the file must be in `astropy.table` format, so that a developer doesn't have to cross-check against the command-line docs.

python/lsst/daf/butler/script/ingest_files.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+
+    # Check that the formatter can be imported -- validate this as soon
+    # as possible before we read a potentially large table file.
+    if formatter:
+        doImport(formatter)
+    else:
+        formatter = None
+
+    # Force null prefix to None for API compatibility.
```
kfindeisen: Coming from C-family languages, I'm used to thinking of "null" as meaning `None`. Did you mean an empty string?

python/lsst/daf/butler/script/ingest_files.py
```diff
@@ -0,0 +1,140 @@
...
+    if not prefix:
+        prefix = None
+
+    # Convert the dataset ID gen more string to enum
```
kfindeisen:
```suggestion
    # Convert the dataset ID gen mode string to enum
```

python/lsst/daf/butler/script/ingest_files.py
```diff
@@ -0,0 +1,140 @@
...
+    ----------
+    repo : `str`
+        URI string of the Butler repo to use.
+    dataset_type : `str`
+        The name of the dataset type for the files to be ingested.
+    formatter : `str`
+        Fully-qualified python class name for the `Formatter` to use
+        to read the ingested files.
+    run : `str`
+        The run in which the files should be ingested.
```
kfindeisen: Order does not match function parameters, and `data_id` is missing entirely.

python/lsst/daf/butler/script/ingest_files.py
```diff
@@ -0,0 +1,140 @@
...
+
+    datasetType = butler.registry.getDatasetType(dataset_type)
+
+    universe = butler.registry.dimensions
+
+    # Convert any additional k=v strings in the dataId tuple to dict
+    # form.
+    common_data_id: Dict[str, Any] = {}
+    for id_str in data_id:
+        dimension_str, value = id_str.split("=")
```
kfindeisen: I suggest factoring the entire initialization of `common_data_id` into its own function, that will make the code easier to follow.

kfindeisen: Incidentally, I thought it was good style in CLI libraries to do such parsing at the CLI layer (so that, in this case, `--data-id` already returns its argument in the form of a `dict`). Is this kind of architecture discouraged by Click?

python/lsst/daf/butler/script/ingest_files.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # Cast the value to the right python type (since they will be
+        # strings at this point).
+        value = dimension.primaryKey.getPythonType()(value)
+
+        common_data_id[dimension_str] = value
+
+    # Read the table and determine the dimensions (the first column
+    # is the file URI).
+    table = Table.read(table_file)
```
kfindeisen: This comment is actually a bit confusing, because it suggests that `Table.read` also determines the dimensions. I suggest leaving it out (the calls are fairly self-explanatory, aside from trimming off the first column).

python/lsst/daf/butler/script/ingest_files.py
```diff
@@ -0,0 +1,140 @@
...
+
+    dimensions = table.colnames
+    dimensions.pop(0)
+
+    # Handle multiple dataIds per file by grouping by file.
+    refs_by_file = defaultdict(list)
+    n_datasets = 0
+    for row in table:
```
kfindeisen: I suggest making a separate function for the initialization of `datasets` and `n_datasets`. In particular, `refs_by_file` is not needed except as an intermediate step.

python/lsst/daf/butler/script/ingest_files.py
```diff
@@ -0,0 +1,140 @@
...
+        dataId: Dict[str, Any] = {k: row[k] for k in dimensions}
```
kfindeisen: I'm surprised `astropy.table` forces you to do this kind of inner loop. Is there no way to directly interpret a row or slice as a mapping?

python/lsst/daf/butler/script/ingest_files.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # The command line can override a column.
+        dataId.update(common_data_id)
+
+        standardized = DataCoordinate.standardize(dataId, graph=datasetType.dimensions)
+
+        ref = DatasetRef(datasetType, standardized, conform=False)
```
kfindeisen: Why have two separate steps instead of just calling `DatasetRef(..., conform=True)`?

tests/test_cliCmdIngestFiles.py
```diff
@@ -0,0 +1,106 @@
...
+    configFile = os.path.join(TESTDIR, "config/basic/butler.yaml")
+
+    def setUp(self):
+        self.root = makeTestTempDir(TESTDIR)
+        self.root2 = makeTestTempDir(TESTDIR)
+        self.testRepo = MetricTestRepo(self.root,
+                                       configFile=self.configFile)
+
+    def tearDown(self):
+        removeTestTempDir(self.root)
```
kfindeisen: What about `self.root2`?

I suggest using `addCleanup` instead, it will be easier to keep in sync.

tests/test_cliCmdIngestFiles.py
```diff
@@ -0,0 +1,106 @@
...
+                json.dump(data._asdict(), fd)
+            datasets.append(data)
+            files.append(outfile)
+
+        # Create some Astropy tables with ingest options.
+        tables = (
+            # Everything in the table with relative path.
+            (Table([files, [423, 424], ["DummyCamComp", "DummyCamComp"]],
+                   names=["Files", "visit", "instrument"]),
+             ("--prefix", self.root2)),
```
kfindeisen: I think this code would be easier to understand if you factored the data ID components into variables (e.g., `visits`, `instruments`). Otherwise it's hard to discern that 423 and 424 are visit numbers, and can be arbitrary.

tests/test_cliCmdIngestFiles.py
```diff
@@ -0,0 +1,106 @@
...
+                self.assertEqual(result.exit_code, 0, clickResultMsg(result))
+
+                butler = Butler(self.root)
+                refs = list(butler.registry.queryDatasets("test_metric_comp", collections=run))
+                self.assertEqual(len(refs), 2)
+
+                for i, data in enumerate(datasets):
+                    butler_data = butler.get("test_metric_comp", visit=423+i, instrument="DummyCamComp",
+                                             collections=run)
+                    self.assertEqual(butler_data, data)
```
kfindeisen: I suggest moving this code into a separate method -- it looks like it would depend only on `table`, `options`, and the data IDs? It would make the code clearer, and certainly easier to improve in the future, if the test assertions were insulated from the exact code used to create the test data.

tests/test_cliCmdIngestFiles.py
```diff
@@ -0,0 +1,106 @@
...
+             ("--prefix", self.root2)),
+            # Relative path with instrument factored out.
+            (Table([files, [423, 424]],
+                   names=["Files", "visit"]),
+             ("--data-id", "instrument=DummyCamComp", "--prefix", self.root2)),
+            # Absolute path with instrument factored out.
+            (Table([[os.path.join(self.root2, f) for f in files], [423, 424]],
+                   names=["Files", "visit"]),
+             ("--data-id", "instrument=DummyCamComp")),
+        )
```
kfindeisen: Having one method for all tests makes it hard to isolate or diagnose problems. I suggest making each of these entries its own test method, moving the initialization of `test*.json` into `setUp`, and factoring the loop as suggested above.

kfindeisen (changes requested):
Functionally, the code looks good, but I'm not happy with the organization into single very long functions. This makes the code hard to follow (for which I noticed you try to compensate with extra comments) and hard to modify in the future. I'd like to see both the utility code and the test case broken up into more manageable pieces; I'll be happy to approve once that's done.

python/lsst/daf/butler/script/ingest_files.py
timj: Funny you should ask and maybe @TallJimbo can explain. The reason I did it this way is because of mypy which insists that DatasetRef can only take a DataCoordinate even though it can also take a dict if conform=True. Maybe I should have asked @TallJimbo if I can fix the annotation for the DatasetRef constructor itself.

tests/test_cliCmdIngestFiles.py
timj: 423 and 424 aren't arbitrary -- they are the visit numbers pre-registered in the test repo.

python/lsst/daf/butler/script/ingest_files.py
TallJimbo: Yeah, sounds like the annotation should probably be changed to the `DataId` type alias.

python/lsst/daf/butler/_butler.py
timj: doc fix. I merged it with the commit that fixed the type annotation in FileDataset.

python/lsst/daf/butler/cli/cmd/commands.py
timj: If you want a particular ingested file to have the same UUID wherever you ingest it you must change this mode accordingly. For "raw" in `butler ingest-raws` we use DATAID_TYPE_RUN so that we guarantee that when a raw is ingested into the default collection you end up with the same UUID in every site (so summit, base, IDF, NCSA) -- this helps a lot with provenance. I can't tell with this command whether people want that reproducible UUID so I have to give people the option. Maybe the default should change to DATAID_TYPE_RUN (which means the UUID is derived from the dataId, the datasetType, and the run) but UNIQUE is probably fine as a default as well. So I think if you are running this command you can ignore it if you don't want reproducibility across multiple repos but I have to leave the option here for the cases where people do want it.

python/lsst/daf/butler/cli/cmd/commands.py
timj: Okay. I'll make it clear. For `butler ingest-raws` the dimension record for `exposure` can be created as part of ingest because the file metadata defines the dimension record. For general ingest we have no idea where to get the record information from -- for most use cases these ingests will be related to instrument/detector/exposure records so should already exist.

python/lsst/daf/butler/script/ingest_files.py
timj: Yes, I  mean an empty string (which I always though of as a null string in C).

python/lsst/daf/butler/script/ingest_files.py
timj: You are right that this could be done in the click layer but we got into the habit of passing kwargs straight through from click to the `script` implementation -- that's simpler when iterating on the interface. We also aren't treating the `script` hierarchy as part of the public API so in daf_butler we put the definitions of the subcommands in a single file and then immediately forward to an implementation `script` file. No-one is expected to be calling these apart from click. We always assumed that if we had some reusable functionality in a `script` it would be moved into the shared hierarchy.

python/lsst/daf/butler/script/ingest_files.py
timj: An `astropy.table.row.Row` is a mapping (for example the `row[k]` here) and in this code  I'm extracting all the columns that correspond to dimensions.  It doesn't look like I can do `row[1:]` to extract those columns as a slice. Another approach could be:

```python
dataId = dict(row)
path = dataId.pop(file_colname)
```
which I assume works even if people haven't named the first column (the first column name is arbitrary). The `{k: v...}` approach seemed more explicit in that I was asking exactly for the columns that I knew were the dimensions but as written those are meant to be self-consistent (I'm not filtering out columns that I don't recognize as dimensions but continuing to process the table).

python/lsst/daf/butler/script/ingest_files.py
timj: I changed the annotation and everything broke since the overwhelming usage for `conform=False` is for plain `dict` in test and validation code and to appease mypy I had made that illegal. There doesn't seem to be a way to turn a non-conforming `dict` into a non-conforming `DataCoordinate` so I think I'm going to have to tell `mypy` that the `self.dataId` itself can also be `DataId` and is not required to be a `DataCoordinate`.

timj: ...and to confirm, mypy is really upset if I declare that `DatasetRef.dataId` might not be a `DataCoordinate`. This all explains why we weren't telling mypy about the plain dict option. I'll go back to how it was I think and add an `# type: ignore` (which is what we do in `_butler.py`.

timj:
@kfindeisen I've rearranged things as requested so can you please take a second look?

python/lsst/daf/butler/cli/cmd/commands.py
```diff
@@ -539,3 +541,60 @@ def collection_chain(**kwargs):
...
+@click.option("--formatter", type=str,
+              help="Fully-qualified python class to use as the Formatter. If not specified the formatter"
+              " will be determined from the dataset type and datastore configuration.")
+@click.option("--id-generation-mode",
+              default="UNIQUE",
+              help="Mode to use for generating dataset IDs. The default creates a unique ID. Other options"
+              " are: 'DATAID_TYPE' for creating a reproducible ID from the dataID and dataset type;"
+              " 'DATAID_TYPE_RUN' for creating a reproducible ID from the dataID, dataset type and run."
+              " The latter is usually used for 'raw'-type data that will be ingested in multiple."
+              " repositories.",
```
kfindeisen, line 560: Given the warning not to use `butler ingest-files` for  raws, is this comment still relevant?

python/lsst/daf/butler/script/ingest_files.py
```diff
@@ -0,0 +1,197 @@
...
+    datasetType = butler.registry.getDatasetType(dataset_type)
+
+    # Convert the k=v strings into a dataId dict.
+    universe = butler.registry.dimensions
+    common_data_id = parse_data_id_tuple(data_id, universe)
+
+    # Read the table assuming that Astropy can work out the format.
+    table = Table.read(table_file)
+
+    datasets = extract_datasets_from_table(table, common_data_id, datasetType, formatter, prefix)
```
kfindeisen, line 107: Thanks, this looks much better, though the argument list for `extract_datasets_from_table` is a bit longer than I would have expected. 🤔

python/lsst/daf/butler/script/ingest_files.py
```diff
@@ -0,0 +1,197 @@
...
+        to read the ingested files. If `None` the formatter is read from
+        datastore configuration based on the dataset type.
+    prefix : `str`
+        Prefix to be used for relative paths. Can be `None` for current
+        working directory.
+
+    Returns
+    -------
+    datasets : `list` of `FileDataset`
+        The `FileDataset` object corresponding to the rows in the table.
```
kfindeisen:
```suggestion
        The `FileDataset` objects corresponding to the rows in the table.
```

tests/test_cliCmdIngestFiles.py
```diff
@@ -0,0 +1,119 @@
...
+            data = MetricsExample(summary={"int": i, "string": f"{self.id()}_{i}"})
+            outfile = f"test{i}.json"
+            with open(os.path.join(self.root2, outfile), "w") as fd:
+                json.dump(data._asdict(), fd)
+            datasets.append(data)
+            files.append(outfile)
+
+        # Associate the dataId with these datasets
+        self.files = files
+        self.datasets = datasets
```
kfindeisen: Seems like you could have just defined `self.files` and `self.datasets` right away, at the cost of only a little more verbosity.

kfindeisen (approved):
Thanks for making these changes!

python/lsst/daf/butler/cli/cmd/commands.py
timj: The photodiode camera data probably should use DATAID_TYPE_RUN since the `exposure` it's associated with will have been defined when ingesting the raw data itself. That's what I mean by "raw-type" as opposed to raw. We may well decide that we should always use DATAID_TYPE_RUN for any external files being ingested but I'm not ready to make that the default without more experience.

PR #546, opened 2021-07-07
DM-30977: Add python log handler to store LogRecord as they come in

timj:
These records can then be serialized in butler datastore.  This is assuming that lsst/log#55 is merged which will also need changes in daf_butler for initializing logging such that we no longer forward everything to lsst.log. Will need changes to ctrl_mpexec as well.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/cli/cliLog.py
+++ b/python/lsst/daf/butler/cli/cliLog.py
@@ -85,7 +85,7 @@ class CliLog:
     that need to be closed on reset."""
 
     @classmethod
-    def initLog(cls, longlog, log_file=()):
+    def initLog(cls, longlog: bool, log_tty: bool = True, log_file=()):
         """Initialize logging. This should only be called once per program
         execution. After the first call this will log a warning and return.
 
@@ -96,6 +96,9 @@ class CliLog:
         ----------
         longlog : `bool`
             If True, make log messages appear in long format, by default False.
+        log_tty : `bool`
+            Control whether a default stream handler is enabled that logs
+            to the terminal.
         log_file : `tuple` of `str`
             Path to files to write log records. If path ends in ``.json`` the
             records will be written in JSON format. Else they will be written
@@ -124,7 +127,9 @@ class CliLog:
             # MDC is set via ButlerMDC, rather than in lsst.log.
             lsstLog.usePythonLogging()
 
-        if longlog:
+        if not log_tty:
+            logging.basicConfig(force=True, handlers=[logging.NullHandler()])
+        elif longlog:
 
             # Want to create our own Formatter so that we can get high
             # precision timestamps. This requires we attach our own
--- a/python/lsst/daf/butler/cli/cliLog.py
+++ b/python/lsst/daf/butler/cli/cliLog.py
@@ -85,7 +85,7 @@ class CliLog:
     that need to be closed on reset."""
 
     @classmethod
-    def initLog(cls, longlog, log_file=None):
+    def initLog(cls, longlog, log_file=()):
         """Initialize logging. This should only be called once per program
         execution. After the first call this will log a warning and return.
 
@@ -96,10 +96,10 @@ class CliLog:
         ----------
         longlog : `bool`
             If True, make log messages appear in long format, by default False.
-        log_file : `str`
-            Path to file to write log records. If ends in ``.json`` the
+        log_file : `tuple` of `str`
+            Path to files to write log records. If path ends in ``.json`` the
             records will be written in JSON format. Else they will be written
-            in text format. If `None` no log file will be created. Records
+            in text format. If empty no log file will be created. Records
             will be appended to this file if it exists.
         """
         if cls._initialized:
@@ -154,9 +154,9 @@ class CliLog:
         ButlerMDC.add_mdc_log_record_factory()
 
         # Set up the file logger
-        if log_file:
-            handler = logging.FileHandler(log_file)
-            if log_file.endswith(".json"):
+        for file in log_file:
+            handler = logging.FileHandler(file)
+            if file.endswith(".json"):
                 formatter = JsonFormatter()
             else:
                 if longlog:
--- a/python/lsst/daf/butler/cli/cliLog.py
+++ b/python/lsst/daf/butler/cli/cliLog.py
@@ -27,8 +27,7 @@ try:
 except ModuleNotFoundError:
     lsstLog = None
 
-from lsst.daf.butler import ButlerMDC
-from ..core.logging import JsonFormatter
+from ..core.logging import JsonLogFormatter, ButlerMDC
 
 
 class PrecisionLogFormatter(logging.Formatter):
@@ -162,7 +161,7 @@ class CliLog:
         for file in log_file:
             handler = logging.FileHandler(file)
             if file.endswith(".json"):
-                formatter = JsonFormatter()
+                formatter = JsonLogFormatter()
             else:
                 if longlog:
                     formatter = PrecisionLogFormatter(fmt=cls.pylog_longLogFmt, style="{")
--- a/python/lsst/daf/butler/cli/cliLog.py
+++ b/python/lsst/daf/butler/cli/cliLog.py
@@ -28,7 +28,7 @@ except ModuleNotFoundError:
     lsstLog = None
 
 from lsst.daf.butler import ButlerMDC
-from ..core.logging import MDCDict
+from ..core.logging import MDCDict, JsonFormatter
 
 
 class PrecisionLogFormatter(logging.Formatter):
@@ -80,8 +80,12 @@ class CliLog:
     _initialized = False
     _componentSettings = []
 
+    _fileHandlers = []
+    """Any FileHandler classes attached to the root logger by this class
+    that need to be closed on reset."""
+
     @classmethod
-    def initLog(cls, longlog):
+    def initLog(cls, longlog, log_file=None):
         """Initialize logging. This should only be called once per program
         execution. After the first call this will log a warning and return.
 
@@ -92,6 +96,11 @@ class CliLog:
         ----------
         longlog : `bool`
             If True, make log messages appear in long format, by default False.
+        log_file : `str`
+            Path to file to write log records. If ends in ``.json`` the
+            records will be written in JSON format. Else they will be written
+            in text format. If `None` no log file will be created. Records
+            will be appended to this file if it exists.
         """
         if cls._initialized:
             # Unit tests that execute more than one command do end up
@@ -152,8 +161,22 @@ class CliLog:
 
         logging.setLogRecordFactory(record_factory)
 
+        # Set up the file logger
+        if log_file:
+            handler = logging.FileHandler(log_file)
+            if log_file.endswith(".json"):
+                formatter = JsonFormatter()
+            else:
+                if longlog:
+                    formatter = PrecisionLogFormatter(fmt=cls.pylog_longLogFmt, style="{")
+                else:
+                    formatter = logging.Formatter(fmt=cls.pylog_normalFmt, style="{")
+            handler.setFormatter(formatter)
+            logging.getLogger().addHandler(handler)
+            cls._fileHandlers.append(handler)
+
         # remember this call
-        cls.configState.append((cls.initLog, longlog))
+        cls.configState.append((cls.initLog, longlog, log_file))
 
     @classmethod
     def resetLog(cls):
@@ -177,6 +200,14 @@ class CliLog:
             logger = logging.getLogger(componentSetting.component)
             logger.setLevel(componentSetting.pythonLogLevel)
         cls._setLogLevel(None, "INFO")
+
+        # Remove the FileHandler we may have attached.
+        root = logging.getLogger()
+        for handler in cls._fileHandlers:
+            handler.close()
+            root.removeHandler(handler)
+
+        cls._fileHandlers.clear()
         cls._initialized = False
         cls.configState = []
 
--- a/python/lsst/daf/butler/cli/cliLog.py
+++ b/python/lsst/daf/butler/cli/cliLog.py
@@ -28,7 +28,7 @@ except ModuleNotFoundError:
     lsstLog = None
 
 from lsst.daf.butler import ButlerMDC
-from ..core.logging import MDCDict, JsonFormatter
+from ..core.logging import JsonFormatter
 
 
 class PrecisionLogFormatter(logging.Formatter):
@@ -151,15 +151,7 @@ class CliLog:
         # to the records. By default this is only used for long-log
         # but always enable it for when someone adds a new handler
         # that needs it.
-        old_factory = logging.getLogRecordFactory()
-
-        def record_factory(*args, **kwargs):
-            record = old_factory(*args, **kwargs)
-            # Make sure we send a copy of the global dict in the record.
-            record.MDC = MDCDict(ButlerMDC._MDC)
-            return record
-
-        logging.setLogRecordFactory(record_factory)
+        ButlerMDC.add_mdc_log_record_factory()
 
         # Set up the file logger
         if log_file:
@@ -200,6 +192,8 @@ class CliLog:
             logger = logging.getLogger(componentSetting.component)
             logger.setLevel(componentSetting.pythonLogLevel)
         cls._setLogLevel(None, "INFO")
+
+        ButlerMDC.restore_log_record_factory()
 
         # Remove the FileHandler we may have attached.
         root = logging.getLogger()
--- a/python/lsst/daf/butler/cli/cliLog.py
+++ b/python/lsst/daf/butler/cli/cliLog.py
@@ -172,7 +172,7 @@ class CliLog:
             cls._fileHandlers.append(handler)
 
         # remember this call
-        cls.configState.append((cls.initLog, longlog, log_file))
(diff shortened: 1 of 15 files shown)


andy-slac:
DM-30996 will most likely add `MDC` attribute to `LogRecord` which will be a dict-like object (or maybe SimpleNamespace but it should support dict methods too). I think we want to save its contents too in some portable form.

timj:
The new `ButlerLogRecord` can be as flexible as we want it to be. I can definitely make it scan the `logging.LogRecord` for additional dict-like objects and add them in to `ButlerLogRecord.extra` or add them into the expanded message.

python/lsst/daf/butler/core/logging.py
```diff
@@ -21,11 +21,185 @@
...
+
 logging.addLevelName(VERBOSE, "VERBOSE")
+
+
+class ButlerLogRecord(BaseModel):
+    """A model representing a `logging.LogRecord`.
+
+    A `~logging.LogRecord` always uses the current time in its record
+    when recreated and that makes it impossible to use it as a
+    serialization format. Instead have a local representation of a
```
andy-slac, line 173: You can overwrite LogRecord attributes after creating it, you just need to know which ones, and there is a `logging.makeLogRecord()` method which looks like it was specifically designed for supporting LogRecord serialization (probably as a dict). Pickling LogRecod directly is probably not the right thing to do, in particular new MDC attribute needs some care, you probably want to serialize it a as a simple dict, to avoid dependency on `log.lsst.MDCDict` class.

timj: Right, that's all why I created `ButlerLogRecord` so that we can have the flexibility to deal with MDC.

python/lsst/daf/butler/cli/cliLog.py
```diff
@@ -152,8 +161,22 @@ def record_factory(*args, **kwargs):
 
         logging.setLogRecordFactory(record_factory)
 
+        # Set up the file logger
+        if log_file:
+            handler = logging.FileHandler(log_file, delay=True)
```
andy-slac: Is `delay` useful? May be better to fail early if file cannot be open? Also there could be non-trivial interaction with potential `chdir` use.

python/lsst/daf/butler/cli/opt/options.py
```diff
@@ -113,6 +113,12 @@ def makeCollectionTypes(context, param, value):
                                     help="Make log messages appear in long format.",
                                     is_flag=True)
 
+log_file_option = MWOptionDecorator("--log-file",
+                                    default=None,
+                                    type=MWPath(file_okay=True, dir_okay=False, writable=True),
+                                    help="File to write log messages. If .json JSON log records will be"
```
andy-slac: `If .json` -> maybe `If ends with .json then ...`?

python/lsst/daf/butler/core/logging.py
```diff
@@ -94,3 +105,270 @@ def MDCRemove(cls, key: str) -> None:
...
+
+        record_dict["message"] = record.getMessage()
+
+        # MDC
+        if hasattr(record, "MDC"):
+            # mypy does not understand hasattr.
+            record_dict["MDC"] = dict(record.MDC)  # type: ignore
+        else:
+            # Use an empty MDC to prevent format failures.
+            record_dict["MDC"] = defaultdict(str)
```
andy-slac: Maybe

    record_dict["MDC"] = getattr(record, "MDC", defaultdict(str))

python/lsst/daf/butler/core/logging.py
```diff
@@ -94,3 +105,270 @@ def MDCRemove(cls, key: str) -> None:
...
+        # Allow byte or str streams since pydantic supports either.
+        first_char = str(first_line[0])
+        if first_char == "[":
+            # This is a ButlerLogRecords model serialization.
+            all = first_line + stream.read()
+            return cls.parse_raw(all)
+
+        # A stream of records with one record per line.
+        if first_char != "{":
+            raise RuntimeError(f"Unrecognized JSON log format. First lines is '{first_line}'")
```
andy-slac: typo: lines is

python/lsst/daf/butler/core/logging.py
```diff
@@ -94,3 +105,270 @@ def MDCRemove(cls, key: str) -> None:
...
+        if not first_line:
+            # Empty file, return zero records.
+            return cls.from_records([])
+
+        # Allow byte or str streams since pydantic supports either.
+        first_char = str(first_line[0])
+        if first_char == "[":
+            # This is a ButlerLogRecords model serialization.
+            all = first_line + stream.read()
+            return cls.parse_raw(all)
```
andy-slac, line 386: Does it need whole file content in memory to parse? Could be an issue with large files. But I guess to write it in this format you also need everything in memory.

python/lsst/daf/butler/core/logging.py
```diff
@@ -94,3 +105,270 @@ def MDCRemove(cls, key: str) -> None:
...
+        item = self.__root__[index]
+        if isinstance(item, list):
+            return type(self)(__root__=item)
+        else:
+            return item
+
+    def __reversed__(self) -> Iterator[ButlerLogRecord]:
+        return self.__root__.__reversed__()
+
+    def __delitem__(self, index: int) -> None:
```
andy-slac: Can `index` be a slice here too?

python/lsst/daf/butler/core/logging.py
```diff
@@ -94,3 +105,270 @@ def MDCRemove(cls, key: str) -> None:
...
+        # Ensure that every record uses the same format string.
+        return "\n".join(record.format(self.log_format) for record in self.__root__)
+
+    def _validate_record(self, record: Record) -> ButlerLogRecord:
+        if isinstance(record, ButlerLogRecord):
+            pass
+        elif isinstance(record, LogRecord):
+            record = ButlerLogRecord.from_record(record)
+        else:
+            raise ValidationError(f"Can only not append item of type {type(record)}")
```
andy-slac: `Can only not append`? 🙂

python/lsst/daf/butler/core/logging.py
```diff
@@ -94,3 +105,270 @@ def MDCRemove(cls, key: str) -> None:
...
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.records = ButlerLogRecords(__root__=[])
+
+    def emit(self, record: LogRecord) -> None:
+        self.records.append(record)
+
+
+class JsonFormatter(Formatter):
```
andy-slac: There is another class with the same name in this package, maybe call this `JsonLogFormatter`?

python/lsst/daf/butler/formatters/logs.py
```diff
@@ -0,0 +1,121 @@
...
+from lsst.daf.butler import ButlerLogRecords, ButlerLogRecord
+from .json import JsonFormatter
+
+
+class ButlerLogRecordsFormatter(JsonFormatter):
+    """Read and write log records in JSON format.
+
+    This is a naive implementation that treats everything as a pydantic.
+    model.  In the future this may be changed to be able to read
+    ``ButlerLogRecord`` one at time from the file and return a subset
```
andy-slac: Can do single backticks?

python/lsst/daf/butler/formatters/logs.py
```diff
@@ -0,0 +1,121 @@
...
+
+        Returns
+        -------
+        data : `object`
+            Data as Python object read from JSON file.
+
+        Notes
+        -----
+        Can read two forms of JSON log file. It can read a full JSON
+        document created from `ButlerLogRecords`, or a stream off standalone
```
andy-slac: typo: off

python/lsst/daf/butler/formatters/logs.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def _readFile(self, path: str, pytype: Optional[Type[Any]] = None) -> Any:
+        """Read a file from the path in JSON format.
+
+        Parameters
+        ----------
+        path : `str`
+            Path to use to open JSON format file.
+        pytype : `class`, optional
+            Not used by this implementation.
```
andy-slac: Is this true?

python/lsst/daf/butler/formatters/logs.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def _fromBytes(self, serializedDataset: bytes, pytype: Optional[Type[Any]] = None) -> Any:
+        """Read the bytes object as a python object.
+
+        Parameters
+        ----------
+        serializedDataset : `bytes`
+            Bytes object to unserialize.
+        pytype : `class`, optional
+            Not used by this implementation.
```
andy-slac: Same here

python/lsst/daf/butler/formatters/logs.py
```diff
@@ -0,0 +1,121 @@
...
+        if pytype is None:
+            pytype = ButlerLogRecords
+        elif not issubclass(pytype, ButlerLogRecords):
+            raise RuntimeError(f"Python type {pytype} does not seem to be a ButlerLogRecords type")
+
+        if serializedDataset.startswith(b"["):
+            return pytype.parse_raw(serializedDataset)
+
+        if not serializedDataset.startswith(b"{"):
+            raise RuntimeError(f"These bytes do not look like JSON -- starts with '{serializedDataset[0]}'")
```
andy-slac: I think this will fail with different exception if `serializedDataset` is empty

python/lsst/daf/butler/formatters/logs.py
```diff
@@ -0,0 +1,121 @@
...
+
+        # Filter out blank lines.
+        records = [ButlerLogRecord.parse_raw(line) for line in serializedDataset.split(b"\n") if line]
```
andy-slac: `split()` is probably not super-efficient space-wise for large data volumes.

andy-slac (approved):

python/lsst/daf/butler/cli/cliLog.py
timj: Maybe not. I ended up adding this to see if it made a difference when debugging a problem with not being able to close files, but I'll remove it now.

python/lsst/daf/butler/core/logging.py
timj: Yes -- this is all working on the assumption that at some point you had it all in a `ButlerLogRecords` pydantic model anyhow. I've been able to read half a million records and it's a bit slow but it works. I have been wondering about adding simple filtering on read (at least for the streaming read) to filter by logger name and log level).

python/lsst/daf/butler/core/logging.py
timj: Maybe. I hadn't thought about it. Will look.

python/lsst/daf/butler/core/logging.py
timj: Oh yes. Luckily we never export datastore formatters. I have been wondering whether I should export any of these classes -- forcing pipetask to pull them in explicitly from `lsst.daf.butler.core.loggng`.

python/lsst/daf/butler/formatters/logs.py
timj: No. I could not use it but since it's there I am using it. Thanks.

python/lsst/daf/butler/formatters/logs.py
timj: Good point. I've been wondering whether to also move this to `ButlerLogRecords` as `.from_raw()` since at the moment the formatter has the byte reading version but the class itself has the file reading version.

python/lsst/daf/butler/formatters/logs.py
timj: Is there a version of `.split()` somewhere that returns an iterator rather than a list? Or should I walk through the string myself looking for `\n` ?  I think this being a list means that we have all the bytes in memory, and all of them again in a list and all of them accumulating into the output list... On the plus size Datastore has an internal limit as to how big a dataset is allowed to be before using fromBytes.

timj: The internet is full of people asking for `str.isplit()` and it never happening.

andy-slac: I don't think there is an iterator version of split() in the standard library, but writing a simple generator method for it would be trivial.

python/lsst/daf/butler/core/logging.py
timj: Actually that's not quite the same because the original code makes a point of copying `record.MDC` into a new dict so that later changes to the MDC won't change the record after the fact. With copying I think it needs to be:

```
record_dict["MDC"] = dict(getattr(record, "MDC", {})
```
This does make the formatting a touch problematic if `LABEL` is not in MDC but I'm not sure what to do about that anyway since as written the record constructed from json won't be using defaultdict either.

Also, since this is a pydantic model it's a bit tricky to add keys at the top level so doing `record_dict(**record.MDC)` isn't really going to work.

andy-slac: So if we need more dynamic model for this record type we cannot use pydantic? Like with any other persistent data we need to be careful with schema changes, at some point we can decide to add/remove attributes to this class and then we need to think about compatibility between new code and what is already stored in butler (or old code and data saved by new release). Maybe we need this class to be a simple dict with arbitrary set of keys (though it would just move compatibility issue to a different level)?

timj: Depends what you mean by dynamic. At the moment with pydantic I've set up the model so that all the LogRecord items are included (I'm keeping the exc_info text distinct from the message itself for now) and all the arbitrary stuff goes into the `MDC` attribute which can accept anything. That seems flexible enough to me. Maybe I should rename MDC to `extra`?  I've never ingested records into LogStash so I don't know how well it will deal with the `MDC` dict or whether it really wants `record.LABEL` to be used and the `MDC` hierarchy missing completely.  The `LogRecord` formatting of `MDC[LABEL]` is not the same code as a the `ButlerLogRecord` serialization so I can convert the `MDC` part of the dict to a defaultdict at that point when calling `ButlerLogRecord.format()`

andy-slac: By dynamic I mean that someone could decide later that using `extra` keyword to logging methods is a great idea (in addition to MDC) and wants to save all those LogRecord attributes in the same JSON format as top-level keys. That set of attributes is potentially unbounded. Packing all of them into `MDC` or one `extra` attribute in this class may or may not work depending on how they are used, I guess we can only discover that later. We could probably reserve `extra` attribute for that future extension and hope that nesting all those extra attributes into one dictionary will work for everyone.

python/lsst/daf/butler/core/logging.py
```diff
@@ -21,16 +21,63 @@
...
 VERBOSE = (logging.INFO + logging.DEBUG) // 2
 """Verbose log level"""
 
+_LONG_LOG_FORMAT = "{levelname} {asctime} {name} {filename}:{lineno} - {message}"
+"""Default format for log records."""
+
 logging.addLevelName(VERBOSE, "VERBOSE")
 
 
+def _isplit(string: Union[bytes, str], sep: Optional[Union[bytes, str]]) -> Iterator[Union[bytes, str]]:
```
timj: @andy-slac This is what I came up with for a split generator. I initially wrote:

```diff
diff --git a/python/lsst/daf/butler/core/logging.py b/python/lsst/daf/butler/core/logging.py
index 2a6ac1ab..ede1e6fe 100644
--- a/python/lsst/daf/butler/core/logging.py
+++ b/python/lsst/daf/butler/core/logging.py
@@ -41,6 +41,30 @@ _LONG_LOG_FORMAT = "{levelname} {asctime} {name} {filename}:{lineno} - {message}
 logging.addLevelName(VERBOSE, "VERBOSE")
 
 
+def _isplit(string: Union[bytes, str], sep: Optional[Union[bytes, str]]) -> Iterator[Union[bytes, str]]:
+    """Generator version of split to avoid making potentially large list.
+
+    Notes
+    -----
+    Python `str` do not support `memoryview` so slicing the string on each
+    iteration to chop the front off will copy the entire remainder of the
+    string. This implementation therefore keeps track of the start and end
+    positions.
+    """
+    if sep is None:
+        sep = " " if isinstance(string, str) else b" "
+    end = None
+    while end != -1:
+        # the type signature above does not explain that the string and
+        # sep types must match up.
+        end = string.find(sep)  # type: ignore
+        if end != -1:
+            yield string[:end]
+            string = string[end + 1:]
+    yield string
+    return
+
```
but then learned that `string = string[end:]` generates a copy so you would still end up with double the memory being required if the string is large.

mypy hates the current implementation and I'm not entirely sure what I have to do to tell mypy:

* All the parameters given and returned are the same type and not arbitrary combinations.
* That the result of `string[0]` is an `int` if `string` is `bytes` -- mypy really gets upset about the line that converts the int back to bytes again.

andy-slac: What about something like 

    T = TypeVar('T', str, bytes)

    def _isplit(string: T, sep: T) -> Iterator[T]:
        begin = 0
        while True:
            end = string.find(sep, begin)
            if end == -1:
                yield string[begin:]
                return
            yield string[begin:end]
            begin = end + 1

timj: Ha. Serves me right for not reading the `str.find()` docs completely and so not noticing the `begin` parameter. Thanks.

PR #549, opened 2021-07-16
DM-31120: Switch all logging to python

timj:
Now lsst.log messages are forwarded to python.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/cli/cliLog.py
+++ b/python/lsst/daf/butler/cli/cliLog.py
@@ -19,38 +19,44 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import datetime
 import logging
-import os
 
 try:
     import lsst.log as lsstLog
 except ModuleNotFoundError:
     lsstLog = None
 
+from lsst.daf.butler import ButlerMDC
+from ..core.logging import MDCDict
 
-# logging properties
-_LOG_PROP = """\
-log4j.rootLogger=INFO, A1
-log4j.appender.A1=ConsoleAppender
-log4j.appender.A1.Target=System.err
-log4j.appender.A1.layout=PatternLayout
-log4j.appender.A1.layout.ConversionPattern={}
-"""
+
+class PrecisionLogFormatter(logging.Formatter):
+    """A log formatter that issues accurate timezone-aware timestamps."""
+
+    converter = datetime.datetime.fromtimestamp
+
+    use_local = True
+    """Control whether local time is displayed instead of UTC."""
+
+    def formatTime(self, record, datefmt=None):
+        """Format the time as an aware datetime."""
+        ct = self.converter(record.created, tz=datetime.timezone.utc)
+        if self.use_local:
+            ct = ct.astimezone()
+        if datefmt:
+            s = ct.strftime(datefmt)
+        else:
+            s = ct.isoformat(sep='T', timespec='milliseconds')
+        return s
 
 
 class CliLog:
     """Interface for managing python logging and ``lsst.log``.
 
-    .. warning::
-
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/core/logging.py
```diff
@@ -21,11 +21,51 @@
...
+        return str(self)
+
+
+class ButlerMDC:
+    """Handle setting and unsetting of global MDC records."""
+
+    _MDC = MDCDict()
+
+    @classmethod
+    def MDC(cls, key: str, value: str) -> str:
```
andy-slac, line 71: I know it's trivial but docstring could be helpful here for people who only see the generated docs and not the code.

python/lsst/daf/butler/core/logging.py
```diff
@@ -21,11 +21,51 @@
...
+        old_value = cls._MDC[key]
+        cls._MDC[key] = value
+        return old_value
+
+    @classmethod
+    def MDCRemove(cls, key: str) -> None:
+        del cls._MDC[key]
```
andy-slac: Maybe add a protection (`cls._MDC.pop(key, None)`) just in case someone tries to delete it twice. It is probably an error but may not be worth crashing the whole application for that?

python/lsst/daf/butler/core/logging.py
```diff
@@ -21,11 +21,51 @@
...
+        return "{" + ", ".join(items) + "}"
+
+    def __repr__(self) -> str:
+        return str(self)
+
+
+class ButlerMDC:
+    """Handle setting and unsetting of global MDC records."""
+
+    _MDC = MDCDict()
```
andy-slac, line 68: One potential inconsistency is that `log4cxx` MDC is per-thread, while this one is global. I don't think we ever used that per-thread feature on Python side in LSST code (the only place that I know of is C++ code in qserv). If it is ever needed we may need to change this to use thread-local storage (which opens another can of worms that was in some way solved in `lsst.log`).

python/lsst/daf/butler/cli/cliLog.py
```diff
@@ -109,48 +101,57 @@ def initLog(cls, longlog):
...
+        # that needs it.
+        old_factory = logging.getLogRecordFactory()
 
-    @classmethod
-    def getHandlerId(cls):
-        """Get the id of the lsst.log handler added to the python logger.
+        def record_factory(*args, **kwargs):
+            record = old_factory(*args, **kwargs)
+            record.MDC = ButlerMDC._MDC
+            return record
```
andy-slac: `configure_pylog_MDC` already does the same so we are going to have two factories creating `MDC` objects which is not optimal performance-wise. Maybe we need additional flag to `configure_pylog_MDC` to make `_MDC` part optional?

andy-slac (approved):

python/lsst/daf/butler/cli/cliLog.py
timj: I was wondering about that. The lsst.log factory is called first and then the python one overrides it. I was pondering making this code merge the MDC from the lsst.log side but it's not really worth it given that no butler code will ever set the lsst.log MDC.  I'll add a small change to `configure_pylog_MDC` to make it skip the MDC part if the MDC class is None.

python/lsst/daf/butler/core/logging.py
timj: Python logging itself does handle threading but I had not pondered the situation of multiple threads setting their own MDC. We haven't got any code that wants to do that so I think I'll punt for now but will add a comment about it.

python/lsst/daf/butler/cli/cliLog.py
timj: See lsst/log#57

PR #550, opened 2021-07-29
DM-31249: Add some test for ButlerURI.join()

timj:
Now raises if an absolute scheme-less URI is given. The test
uncovered a small problem with quoting of the join path and that
is also fixed.

## Checklist

- [x] ran Jenkins


SergeyPod (approved):
I confirm that patch provided by this pull request code raises exception as expected. Looks good to me.

PR #551, opened 2021-07-30
DM-31251: Allow butler.ingest to work with execution butler

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -723,8 +723,16 @@ class FileDatastore(GenericBaseDatastore):
             # Allow ButlerURI to use its own knowledge
             transfer = "auto"
         else:
-            raise ValueError("Some datasets are inside the datastore and some are outside."
-                             " Please use an explicit transfer mode and not 'auto'.")
+            # This can happen when importing from a datastore that
+            # has had some datasets ingested using "direct" mode.
+            # Also allow ButlerURI to sort it out but warn about it.
+            # This can happen if you are importing from a datastore
+            # that had some direct transfer datasets.
+            log.warning("Some datasets are inside the datastore and some are outside. Using 'split' "
+                        "transfer mode. This assumes that the files outside the datastore are "
+                        "still accessible to the new butler since they will not be copied into "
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -861,10 +869,18 @@ def _extractIngestInfo(self, path: Union[str, ButlerURI], ref: DatasetRef, *,
...
-                tgtLocation = self.locationFactory.fromPath(pathInStore)
+                if pathInStore:
+                    tgtLocation = self.locationFactory.fromPath(pathInStore)
+                elif transfer == "split":
+                    # Outside the datastore but treat that as a direct ingest
+                    # instead.
+                    tgtLocation = None
+                else:
+                    raise RuntimeError(f"Unexpected transfer mode encountered: {transfer} for"
+                                       f" URI {srcUri}")
```
natelust, line 883: it seems like this whole if block is combining two different bits of logic together. It might be the right thing, but I am having a bit of trouble following it, not knowing what some of the results of calls will be.

You have changed to `pathInStore is None and transfer is None` is this implying that you are using something like True, False, and None as conditions in normal operation? Is it ok that if `pathInStore` is None, transfer is not None, for the `if pathInStore` to evaluate to false based on None being false-y?

I think I understand what you are doing and I think it is fine, but I am sufficiently uncertain, that I want to highlight it for a second look by you.

natelust (approved):

python/lsst/daf/butler/datastores/fileDatastore.py
timj: @TallJimbo used `transfer=None` to mean "the files are already in the right place" and no transfer is needed. For `split` mode I need to handle the case where some are in and some are out. For the files inside the target location is relative to the datastore root. For the files outside there is also no transfer but that is "direct" mode and is handled by setting the target location to None (so the code below knows to use the source URI directly). It is combining two bits of logic but the `None` mode is most of the logic and it didn't seem like it was reasonable to add a new `split` path as a separate check that would duplicate the `None` -- the only change is to set that target URI to `None` in split mode.

PR #552, opened 2021-08-04
DM-31295: Cache boto client and add timer messages to logs for some file I/O

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/_butlerUri/file.py
+++ b/python/lsst/daf/butler/core/_butlerUri/file.py
@@ -217,8 +217,11 @@ class ButlerFileURI(ButlerURI):
         if transfer not in self.transferModes:
             raise ValueError(f"Transfer mode '{transfer}' not supported by URI scheme {self.scheme}")
 
-        log.debug(f"Transferring {src} [exists: {src.exists()}] -> "
-                  f"{self} [exists: {self.exists()}] (transfer={transfer})")
+        # Existence checks can take time so only try if the log message
+        # will be issued.
+        if log.isEnabledFor(logging.DEBUG):
+            log.debug("Transferring %s [exists: %s] -> %s [exists: %s] (transfer=%s)",
+                      src, src.exists(), self, self.exists(), transfer)
 
         # We do not have to special case ButlerFileURI here because
         # as_local handles that.
--- a/python/lsst/daf/butler/core/_butlerUri/http.py
+++ b/python/lsst/daf/butler/core/_butlerUri/http.py
@@ -328,8 +328,7 @@ class ButlerHttpURI(ButlerURI):
         if r.status_code != 200:
             raise FileNotFoundError(f"Unable to download resource {self}; status code: {r.status_code}")
         with tempfile.NamedTemporaryFile(suffix=self.getExtension(), delete=False) as tmpFile:
-            with time_this(log, log_prefix="timer", msg="Downloading %s to local file",
-                           args=(self,)):
+            with time_this(log, msg="Downloading %s to local file", args=(self,)):
                 for chunk in r.iter_content():
                     tmpFile.write(chunk)
         return tmpFile.name, True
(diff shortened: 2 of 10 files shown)


python/lsst/daf/butler/core/_butlerUri/file.py
```diff
@@ -217,8 +217,11 @@ def transfer_from(self, src: ButlerURI, transfer: str,
         if transfer not in self.transferModes:
             raise ValueError(f"Transfer mode '{transfer}' not supported by URI scheme {self.scheme}")
 
-        log.debug(f"Transferring {src} [exists: {src.exists()}] -> "
-                  f"{self} [exists: {self.exists()}] (transfer={transfer})")
+        # Existence checks can take time so only try if the log message
```
ktlim, line 220: In C++, log macros ensure that log message components are not even evaluated until after it is decided that the message will be issued.  There's no equivalent in Python?  Maybe the log method could take a callable that returns a dict of message components and that would only be called after the threshold check.

python/lsst/daf/butler/core/_butlerUri/http.py
```diff
@@ -340,9 +343,10 @@ def read(self, size: int = -1) -> bytes:
             The number of bytes to read. Negative or omitted indicates
             that all data should be read.
         """
-        log.debug("Reading from remote resource: %s", self.geturl())
```
ktlim, line 343: You don't think it could still be useful to know what the URL is, separate from timing?

python/lsst/daf/butler/core/utils.py
```diff
@@ -475,3 +478,58 @@ def isplit(string: T, sep: T) -> Iterator[T]:
...
+    log : `logging.Logger`, optional
+        Logger to use to report the timer message. The root logger will
+        be used if none is given.
+    msg : `str`, optional
+        Context to include in log message.
+    log_level : `int`, optional
+        Python logging level to use to issue the log message. If the
+        code block raises an exception the log message will automatically
+        switch to level ERROR.
+    log_prefix : `str`, optional
```
ktlim: Your invocations are verbose enough that I wonder if you want to simplify by defaulting to prefixing with `timer`.

ktlim (approved):

ktlim:
Maybe change the PR title to reflect the addition of session caching?

python/lsst/daf/butler/core/_butlerUri/file.py
timj: I've not seen anyone doing deferred calling of log message components in python. We could pass `self.exists` as an argument to some object which then calls it later. This is a bit like how the python docs suggest f-string formatting of log messages is supported: https://docs.python.org/3/howto/logging-cookbook.html#using-custom-message-objects -- can I not worry about this for today and leave it with the isEnabledFor?

python/lsst/daf/butler/core/_butlerUri/http.py
timj: I was vacillating on this. I know that the timer message will come out regardless so I thought that seeing log output of:
```
DEBUG reading URI X
DEBUG read URI X took Y seconds
```
was redundant. Sounds like you'd prefer the log message showing the intent as well as the log message with the result.

python/lsst/daf/butler/core/utils.py
timj: Okay. I'll default it to `timer`. I was also wondering about dropping the `log_` from the variable names so maybe I should do that as well.

python/lsst/daf/butler/core/_butlerUri/file.py
ktlim: Of course.  I was just complaining about the state of Python logging, not asking you to fix it :)

python/lsst/daf/butler/core/_butlerUri/http.py
timj: I suppose one argument in favor of the duplication is that they are in fact different loggers so can be configured with different levels.

ktlim: > different loggers

Exactly.

PR #553, opened 2021-08-06
DM-31337: Allow skipping calibration collections

andy-slac:
QueryBuilder joinDataset() method now has an improved logic for handling
calibration collections. If calibration collection name was not specified
explicitly but appeared inside chained collection then that collection is
silently skipped instead of raising NotImplementedError.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/queries/_builder.py
+++ b/python/lsst/daf/butler/registry/queries/_builder.py
@@ -143,6 +143,7 @@ class QueryBuilder:
             collections = CollectionSearch.fromExpression(collections)
         else:
             collections = CollectionQuery.fromExpression(collections)
+        explicitCollections = frozenset(collections.explicitNames())
         # If we are searching all collections with no constraints, loop over
         # RUN collections only, because that will include all datasets.
         collectionTypes: AbstractSet[CollectionType]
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/registry/queries/_builder.py
```diff
@@ -150,6 +154,9 @@ def joinDataset(self, datasetType: DatasetType, collections: Any, *,
             collectionTypes = {CollectionType.RUN}
         else:
             collectionTypes = CollectionType.all()
+            # Do not search calibration collections if requested
+            if self._skipCalibs:
+                collectionTypes -= {CollectionType.CALIBRATION}
```
timj: Is this the line that does all the work for this entire PR?

timj (approved):
This looks okay, but I think we should add `skipCalibs=True` to the relevant registry queries in the scripts directory so that we can make life easier for people. Once calibration queries are supported we can change it back to `False`. It doesn't seem to be worth it to add a command line option for `--skip-calibs` at the moment since don't command line users always want to skip?

python/lsst/daf/butler/registry/queries/_builder.py
andy-slac: Indeed, the rest of it is just passing that flag from registry methods to here.

timj: Which raises the question I bring up in Jira. If querying calibration collections never works, why isn't this one line there until it's fixed? I assume there are scenarios where we do successfully query calibration collections through some other path?

andy-slac:
I did not think about command line tools, I think that it makes sense for now to set `skipCalibs=True` for now, if we add command line option it will be harder to undo later.

andy-slac:
I also realized that I forgot to update remote registry methods for new argument, need another commit for that.

python/lsst/daf/butler/registry/queries/_builder.py
andy-slac: I want to play with what I suggested on Jira - we could quietly  skip calibration collections when searching in chained ones, but raise an exception if someone searches calibration collection directly. It needs figuring out what collections were requested, may not be very trivial.

timj (approved):
I'm a lot happier about this approach. Looks good.

PR #554, opened 2021-08-06
DM-13365: Proper Caching for remote object stores

timj:
Includes cache expiry.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/datastores/fileDatastore.yaml
+++ b/python/lsst/daf/butler/configs/datastores/fileDatastore.yaml
@@ -19,16 +19,21 @@ datastore:
     # location. This can be left out or null used
     # to indicate a temporary directory.
     root: null
+    expiry:
+      # Expiry mode and associated threshold.
+      # Options are:
+      # - null (no expiry)
+      # - files (threshold is number of files)
+      # - datasets (threshold is number of datasets)
+      # - size (threshold is size in bytes)
+      # - age (threshold is oldest file in seconds)
+      # Default for everyone is to limit the cache size to 4 datasets.
+      mode: datasets
+      threshold: 4
     # Default cache value. This will be the default decision if no specific
     # match is found in the "cacheable" section later on.
     default: false
     # Use a dict over list to simplify merging logic.
-    # Later could change it to be the timescale for caching:
-    #   - Quantum (clear at end of quantum)
-    #   - Pipeline (clear at end of pipeline)
-    # pipeline activator could call clear_cache with a parameter indicating
-    # the event.
     cacheable:
-      # The formatter associated with this storage class
-      # implements deferred loading of data
-      StrayLightData: true
+      Exposure: true
+      ExposureF: true
--- a/python/lsst/daf/butler/configs/datastores/fileDatastore.yaml
+++ b/python/lsst/daf/butler/configs/datastores/fileDatastore.yaml
@@ -19,6 +19,9 @@ datastore:
     # location. This can be left out or null used
     # to indicate a temporary directory.
     root: null
+    # Default cache value. This will be the default decision if no specific
+    # match is found in the "cacheable" section later on.
+    default: false
     # Use a dict over list to simplify merging logic.
     # Later could change it to be the timescale for caching:
     #   - Quantum (clear at end of quantum)
--- a/python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
+++ b/python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
@@ -27,7 +27,9 @@ import posixpath
 import copy
 import logging
 import re
+import tempfile
 
+from random import Random
 from pathlib import Path, PurePath, PurePosixPath
 
 __all__ = ('ButlerURI',)
@@ -403,14 +405,16 @@ class ButlerURI:
         parentPath = originalPath.parent
         return self.replace(path=str(parentPath), forceDirectory=True)
 
-    def replace(self, forceDirectory: bool = False, **kwargs: Any) -> ButlerURI:
+    def replace(self, forceDirectory: bool = False, isTemporary: bool = False, **kwargs: Any) -> ButlerURI:
         """Return new `ButlerURI` with specified components replaced.
 
         Parameters
         ----------
-        forceDirectory : `bool`
+        forceDirectory : `bool`, optional
             Parameter passed to ButlerURI constructor to force this
             new URI to be dir-like.
+        isTemporary : `bool`, optional
+            Indicate that the resulting URI is temporary resource.
         **kwargs
             Components of a `urllib.parse.ParseResult` that should be
             modified for the newly-created `ButlerURI`.
@@ -427,7 +431,8 @@ class ButlerURI:
         # Disallow a change in scheme
         if "scheme" in kwargs:
             raise ValueError(f"Can not use replace() method to change URI scheme for {self}")
-        return self.__class__(self._uri._replace(**kwargs), forceDirectory=forceDirectory)
+        return self.__class__(self._uri._replace(**kwargs), forceDirectory=forceDirectory,
+                              isTemporary=isTemporary)
 
     def updatedFile(self, newfile: str) -> ButlerURI:
         """Return new URI with an updated final component of the path.
@@ -525,7 +530,7 @@ class ButlerURI:
 
         return ext
 
-    def join(self, path: Union[str, ButlerURI]) -> ButlerURI:
+    def join(self, path: Union[str, ButlerURI], isTemporary: bool = False) -> ButlerURI:
         """Return new `ButlerURI` with additional path components.
 
         Parameters
@@ -537,6 +542,8 @@ class ButlerURI:
             referring to an absolute location, it will be returned
             directly (matching the behavior of `os.path.join()`). It can
             also be a `ButlerURI`.
+        isTemporary : `bool`, optional
+            Indicate that the resulting URI represents a temporary resource.
 
         Returns
         -------
@@ -594,7 +601,8 @@ class ButlerURI:
 
         # normpath can strip trailing / so we force directory if the supplied
         # path ended with a /
-        return new.replace(path=newpath, forceDirectory=path.endswith(self._pathModule.sep))
+        return new.replace(path=newpath, forceDirectory=path.endswith(self._pathModule.sep),
+                           isTemporary=isTemporary)
 
     def relative_to(self, other: ButlerURI) -> Optional[str]:
         """Return the relative path from this URI to the other URI.
@@ -723,6 +731,54 @@ class ButlerURI:
             # The caller might have relocated the temporary file
             if is_temporary and local_uri.exists():
                 local_uri.remove()
+
+    @classmethod
+    @contextlib.contextmanager
+    def temporary_uri(cls, prefix: Optional[ButlerURI] = None,
+                      suffix: Optional[str] = None) -> Iterator[ButlerURI]:
+        """Create a temporary URI.
+
+        Parameters
+        ----------
+        prefix : `ButlerURI`, optional
+            Prefix to use. Without this the path will be formed as a local
+            file URI in a temporary directory.
+        suffix : `str`, optional
+            A file suffix to be used.
+
+        Yields
+        ------
+        uri : `ButlerURI`
+            The temporary URI. Will be removed when the context is completed.
+        """
+        if prefix is None or prefix.isLocal:
+            prefix_path = prefix.ospath if prefix else None
+            _, path = tempfile.mkstemp(suffix=suffix, prefix=prefix_path)
+            temporary_uri = ButlerURI(path, isTemporary=True)
+            # To avoid inconsistencies in behavior, delete the file
+            # even though that might result in a name clash later.
+            try:
+                temporary_uri.remove()
+            except FileNotFoundError:
+                pass
+        else:
+            # Remote resource. Need to create a randomized file name.
+            characters = "abcdefghijklmnopqrstuvwxyz0123456789_"
+            rng = Random()
+            tempname = "".join(rng.choice(characters) for _ in range(16))
+            if suffix:
+                tempname += suffix
+            temporary_uri = prefix.join(tempname, isTemporary=True)
+
+        try:
+            yield temporary_uri
+        finally:
+            # It's okay if this does not work because the user removed
+            # the file.
+            try:
+                temporary_uri.remove()
+            except FileNotFoundError:
+                pass
 
     def read(self, size: int = -1) -> bytes:
         """Open the resource and return the contents in bytes.
--- a/python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
+++ b/python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
@@ -27,6 +27,7 @@ import posixpath
 import copy
 import logging
 import re
+import shutil
 import tempfile
 
 from random import Random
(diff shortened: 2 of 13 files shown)


python/lsst/daf/butler/configs/datastores/fileDatastore.yaml
```diff
@@ -19,13 +19,23 @@ datastore:
...
-    # Later could change it to be the timescale for caching:
-    #   - Quantum (clear at end of quantum)
-    #   - Pipeline (clear at end of pipeline)
-    # pipeline activator could call clear_cache with a parameter indicating
-    # the event.
     cacheable:
       # The formatter associated with this storage class
       # implements deferred loading of data
       StrayLightData: true
+      ExposureF: true
```
timj, line 39: Once StrayLightData is turned into a normal dataset we need to decide whether it's still worth caching. We may want to add `Exposure` to this list in addition to `ExposureF` (or fix things so that storage class inheritance works in lookups). Not sure what other large non-exposure datasets we'd like to cache.

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
```diff
@@ -724,6 +732,54 @@ def as_local(self) -> Iterator[ButlerURI]:
...
+    @contextlib.contextmanager
+    def temporary_uri(cls, prefix: Optional[ButlerURI] = None,
+                      suffix: Optional[str] = None) -> Iterator[ButlerURI]:
+        """Create a temporary URI.
+
+        Parameters
+        ----------
+        prefix : `ButlerURI`, optional
+            Prefix to use. Without this the path will be formed as a local
+            file URI in a temporary directory.
```
ktlim: This makes it sound like it might be in a `mkdtemp` directory (and perhaps have user umask) rather than being in an OS-specified one.  Also note that if the prefix contains a directory, that must exist (and be writable).

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
```diff
@@ -724,6 +732,54 @@ def as_local(self) -> Iterator[ButlerURI]:
...
+            The temporary URI. Will be removed when the context is completed.
+        """
+        if prefix is None or prefix.isLocal:
+            prefix_path = prefix.ospath if prefix else None
+            _, path = tempfile.mkstemp(suffix=suffix, prefix=prefix_path)
+            temporary_uri = ButlerURI(path, isTemporary=True)
+            # To avoid inconsistencies in behavior, delete the file
+            # even though that might result in a name clash later.
+            try:
+                temporary_uri.remove()
```
ktlim: This seems less than wonderful.  If you're not going to use any of the features of `mkstemp`, maybe it would be better to just create a random filename in `tempfile.gettempdir()`?

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1070,11 +1067,20 @@ def _removeFileExists(uri: ButlerURI) -> None:
                 uri.write(serializedDataset, overwrite=True)
                 log.debug("Successfully wrote bytes directly to %s", uri)
 
+                # It may be that this dataset should be cached locally
```
ktlim: Is this actually more efficient than writing to temp and then copying?  In particular, the serialization code might be inefficient, so executing it twice would be bad.

It's also a little confusing having the main code for `toBytes` being here in an `else:`, though I understand catching `NotImplementedError` is Pythonic.

I guess this could be rewritten as:
```python
if not self.cacheManager.should_be_cached(ref):
    try:
        toBytes()
    except NotImplementedException:
        pass
    else:
        uri.write()
        return self._extractIngestInfo()
# rest of write-to-temp-and-copy code goes here
```

 (Would be even nicer to be able to stream from the serializer to the URI, but that's another refactor or two down the road.)

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1085,6 +1091,14 @@ def _read_artifact_into_memory(self, getInfo: DatastoreFileGetInformation,
             The registry information associated with this artifact.
         isComponent : `bool`
             Flag to indicate if a component is being read from this artifact.
+        cache_ref : `DatasetRef`, optional
+            The DatasetRef to use when looking up the file in the cache.
+            This ref must have the same ID as the supplied ref but can
+            be a parent ref or component ref to indicate to the cache whether
+            a composite file is being requested from the cache or a component
+            file. Without this the cache will default to the suppled ref but
```
ktlim:
```suggestion
            file. Without this the cache will default to the supplied ref but
```

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1137,45 +1164,53 @@ def _read_artifact_into_memory(self, getInfo: DatastoreFileGetInformation,
...
+                        cache_msg = ""
+                        location_updated = True
+
+                        if self.cacheManager.should_be_cached(cache_ref):
+                            # In this scenario we want to ask if the downloaded
+                            # file should be cached but we should not cache
+                            # it until after we've used it (to ensure it can't
+                            # be expired whilst we are using it).
+                            can_be_cached = True
+                            cache_msg = " and likely cached"
```
ktlim, line 1199: Why "likely" here?  Isn't it guaranteed, at least initially?  So "will be"?

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1137,45 +1164,53 @@ def _read_artifact_into_memory(self, getInfo: DatastoreFileGetInformation,
...
-                        with time_this(log, msg="Reading%s from location %s %s with formatter %s",
-                                       args=(f" component {getInfo.component}" if isComponent else "",
-                                             uri, msg, formatter.name())):
-                            result = formatter.read(component=getInfo.component if isComponent else None)
-                except Exception as e:
-                    raise ValueError(f"Failure from formatter '{formatter.name()}' for dataset {ref.id}"
-                                     f" ({ref.datasetType.name} from {uri}): {e}") from e
+                with uri.as_local() as local_uri:
+
+                    # URI was remote and file was downloaded
```
ktlim: I think this should move next to the `if`, perhaps even after it.

tests/test_datastore.py
```diff
@@ -1056,7 +1061,26 @@ def setUp(self):
 
         # Create empty files
```
ktlim: Not empty anymore, are they?

tests/test_datastore.py
```diff
@@ -1056,7 +1061,26 @@ def setUp(self):
...
         for uri in self.files:
-            uri.write(b"")
+            uri.write(b"0123456789")
+
+        # Create a some composite refs with component files.
```
ktlim:
```suggestion
        # Create some composite refs with component files.
```

tests/test_datastore.py
```diff
@@ -1115,20 +1152,155 @@ def assertCache(self, cache_manager):
...
+        self.assertFalse(found.exists())
+
+        # There should be more files in the cache than normally allowed.
+        self.assertEqual(cache_manager.file_count, threshold + 2)
+
+        # We can still ask for the same file from the cache.
+        with cache_manager.find_in_cache(self.refs[2], ".txt") as found:
+            self.assertIsNotNone(found)
+
+        # But not the dataset that was stored after it.
```
ktlim: Isn't this comment backwards?  You keep both the file and the one after, since it's not `None`.

python/lsst/daf/butler/core/datastoreCacheManager.py
```diff
@@ -31,27 +31,194 @@
...
+
+    Returns
+    -------
+    parsed : `dict` of `str`, `str`
+        Parsed components of the file. These include:
+        - "id": The dataset ID,
+        - "component": The name of the component (can be `None`),
+        - "extension": File extension (can be `None`).
+    """
+    root_ext = cached_location.split(".", maxsplit=1)
```
ktlim, line 121: Is `rsplit()` safer?  Or even `root, _, ext = cached_location.rpartition(".")` ?

Or even `re.search("(.+?)(_([^_]+?))?(\..+)")` ?

python/lsst/daf/butler/core/datastoreCacheManager.py
```diff
@@ -31,27 +31,194 @@
...
+
+def _construct_cache_path(root: ButlerURI, ref: DatasetRef, extension: str) -> ButlerURI:
+    """Construct the full path to use for this dataset in the cache.
+
+    Parameters
+    ----------
+    ref : `DatasetRef`
+        The dataset to look up in or write to the cache.
+    extension : `str`
+        File extension to use for this file.
```
ktlim: I guess you're consistent everywhere, but maybe note that "." must be included in the extension?

python/lsst/daf/butler/core/datastoreCacheManager.py
```diff
@@ -31,27 +31,194 @@
...
+    Returns
+    -------
+    uri : `ButlerURI`
+        URI to use for this dataset in the cache.
+    """
+    # Dataset type component is needed in the name if composite
+    # disassembly is happening since the ID is shared for all components.
+    component = ref.datasetType.component()
+    component = f"_{component}" if component else ""
+    return root.join(f"{ref.id}{component}{extension}")
```
ktlim, line 99: Are you sure `ref.id` can't contain `_`?

python/lsst/daf/butler/core/datastoreCacheManager.py
```diff
@@ -31,27 +31,194 @@
...
+    root = root_ext.pop(0)
+    ext = "." + root_ext.pop(0) if root_ext else None
+
+    parts = root.split("_")
+    id_ = parts.pop(0)
+    component = parts.pop(0) if parts else None
+    return {"id": id_, "component": component, "extension": ext}
+
+
+class CacheEntry(BaseModel):
```
ktlim, line 131: Are you basically using `BaseModel` everywhere now instead of `dataclass`?

python/lsst/daf/butler/core/datastoreCacheManager.py
```diff
@@ -153,32 +355,104 @@ class DatastoreCacheManager(AbstractDatastoreCacheManager):
...
+
+    Notes
+    -----
+    Two environment variables can be used to override the cache directory
+    and expiration configuration:
+
+    * ``$DAF_BUTLER_CACHE_DIRECTORY``
+    * ``$DAF_BUTLER_CACHE_EXPIRATION_MODE``
+
+    The expiration mode should take the form a ``mode=threshold`` so for
```
ktlim:
```suggestion
    The expiration mode should take the form ``mode=threshold`` so for
```

python/lsst/daf/butler/core/datastoreCacheManager.py
```diff
@@ -153,32 +355,104 @@ class DatastoreCacheManager(AbstractDatastoreCacheManager):
...
-        root = self.config.get("root")
-        self._cache_directory = ButlerURI(root, forceAbsolute=True) if root is not None else None
+        # requested. Allow external override from environment.
+        root = os.environ.get("DAF_BUTLER_CACHE_DIRECTORY") or self.config.get("root")
+        self._cache_directory = ButlerURI(root, forceAbsolute=True,
+                                          forceDirectory=True) if root is not None else None
+
+        if self._cache_directory:
+            if not self._cache_directory.isLocal:
+                raise ValueError(f"Cache directory must be a file system. Got: {self._cache_directory}")
```
ktlim:
```suggestion
                raise ValueError(f"Cache directory must be on a local file system. Got: {self._cache_directory}")
```

python/lsst/daf/butler/core/datastoreCacheManager.py
```diff
@@ -31,27 +31,194 @@
...
+    return {"id": id_, "component": component, "extension": ext}
+
+
+class CacheEntry(BaseModel):
+    """Represent an entry in the cache."""
+
+    name: str
+    """Name of the file."""
+
+    size: int
```
ktlim, line 137: I'm wondering if it isn't almost as easy to get these from the filesystem each time rather than maintaining a possibly-out-of-sync in-memory copy.

But I guess it's easier to figure out datasets by id this way.

python/lsst/daf/butler/core/datastoreCacheManager.py
```diff
@@ -223,21 +497,277 @@ def move_to_cache(self, uri: ButlerURI, ref: DatasetRef) -> Optional[ButlerURI]:
...
+        cache_entries : iterable of `str`
+            The entries to remove from the cache. The values are the path
+            within the cache.
+        """
+        for entry in cache_entries:
+            path = self.cache_directory.join(entry)
+
+            # If the file happens to also exist in the exempt folder
+            # skip removal since a process is actively using the file.
+            if self._temp_exempt_directory.join(entry).exists():
```
ktlim: I'm worried that there still may be (brief but non-zero) windows where a race can happen.  Still, since you treat it as uncached in that case, it should be safe if less performant.

python/lsst/daf/butler/core/datastoreCacheManager.py
```diff
@@ -223,21 +497,277 @@ def move_to_cache(self, uri: ButlerURI, ref: DatasetRef) -> Optional[ButlerURI]:
...
+            n_files = len(self._cache_entries)
+            n_over = n_files - self._expiration_threshold
+            if n_over > 0:
+                sorted_keys = self._sort_cache()
+                keys_to_remove = sorted_keys[:n_over]
+                self._remove_from_cache(keys_to_remove)
+            return
+
+        if self._expiration_mode == "datasets":
+            # Count the datasets, using reverse date order
```
ktlim: Isn't this ascending timestamp order, not "reverse date order"?

python/lsst/daf/butler/core/datastoreCacheManager.py
```diff
@@ -223,21 +497,277 @@ def move_to_cache(self, uri: ButlerURI, ref: DatasetRef) -> Optional[ButlerURI]:
...
+            for key in self._sort_cache():
+                entry = self._cache_entries[key]
+                datasets[entry.ref].append(key)
+
+            n_datasets = len(datasets)
+            n_over = n_datasets - self._expiration_threshold
+            if n_over > 0:
+                keys_to_remove = []
+                removed = 0
+                for dataset in datasets:
```
ktlim: Can't you just iterate over `datasets[:n_over]` like above, skipping the whole `removed` thing?

ktlim (approved):

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
timj: You mean because I am already doing that logic in the `else` block below for the remote URI?

python/lsst/daf/butler/core/datastoreCacheManager.py
timj: It can only be a UUID or integer so yes, none of those have underscores. Would you like an `assert` to be included?

python/lsst/daf/butler/core/datastoreCacheManager.py
timj: You are worried about rogue files being found in the cache directory (especially for a shared cache)?

python/lsst/daf/butler/core/datastoreCacheManager.py
timj: Yes, I am generally doing that and in this case it's only creating one of them so there's no performance hit.

python/lsst/daf/butler/core/datastoreCacheManager.py
timj: The file sizes can't change so they are static information. Also I was under the impression that file systems don't really like repeated calls to `stat` if the cache grows big so limiting the cache check to a `readdir` rather than a `readdir`+`stat` seemed like a reasonable thing to do.

python/lsst/daf/butler/core/datastoreCacheManager.py
timj: Sorry -- I was thinking CacheRegistry not CacheEntry -- yes, there will be a bit of overhead for the validation but it seemed reasonable and the validation is very simple.

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
ktlim: Yes, if it works for one, it should work for the other.

python/lsst/daf/butler/core/datastoreCacheManager.py
ktlim: Oh, if this is from a field you control, then no problem.

python/lsst/daf/butler/core/datastoreCacheManager.py
ktlim: I was worried about the `id` field, but since you control that, there's no problem.

python/lsst/daf/butler/core/datastoreCacheManager.py
timj: I did think about using hard links exclusively and letting the cache expiration delete the file regardless. Then I worried that someone might put the cache dir on some file system where hard links wouldn't work for some reason and so I played it a bit safer by skipping the removal if it existed in both places. I'm not sure how to deal with the case of another process deleting the file in the instant before I create a soft link in the exempt directory (the hardlink race would cause the hardlink to fail to be made but would helpfully fall back to creating a soft link to a file that is no longer there).

python/lsst/daf/butler/datastores/fileDatastore.py
timj: It's not serializing the in-memory dataset twice -- the bytes are just bytes that are written twice. Rewriting it so that we always write to the local file first if the intent is to be cached seems reasonable (I hadn't really intended to cache these at all initially -- the tests all failed without because the tests all use very small datasets that can bypass the file writing stage -- hence the convoluted logic I ended up with). 

There is a related problem that if the temp directory is on a different file system to the cache directory we will have to (effectively) read all the bytes back into memory for the "move" since `os.rename` won't work.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: I think I say "likely" because if the `formatter.read` fails I will not be caching it.

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
timj: I've rewritten this to not involve mkstemp.

python/lsst/daf/butler/core/datastoreCacheManager.py
timj: Do you want to block merging until we try to make this robust? How much can we rely on hardlinks working within the temp directory? Hardlinks working mean that we can remove this specific test completely and try to make a hardlink. If the hardlink is made then there is no race (deleting the cache file won't delete the thing in the "exempt" directory). If the hardlink fails we will have to report that the file wasn't really cached. That is all great assuming hardlinks can be trusted. Softlinks are the reason there is a possible race.

timj: hardlinks are much cleaner because then the cache directory can expire as normal without having to keep the extra files around and the "exempt" directory can delete things independently (let's hope they aren't kept around for too long if disk space is tight)

python/lsst/daf/butler/datastores/fileDatastore.py
timj: Thanks. The rewrite is much clearer.

python/lsst/daf/butler/core/datastoreCacheManager.py
timj: Simplified.

timj:
@ktlim I've made all the requested changes. The one outstanding issue is worrying about the exemption race and whether we can force hardlinks and stop trying to fall back to symlinks.

python/lsst/daf/butler/core/datastoreCacheManager.py
ktlim: Since these are on the worker and all within the same directory tree, I think hardlinks are reliable.  But having any race be "uncached" is also a belt-and-suspenders safety precaution.

timj: I've removed the softlink option which also removes any need for cache expiry to know about the "exempt" directory. The cache doesn't have to be on the worker -- people can specify a directory that is on some shared file system if they want.

PR #557, opened 2021-08-12
DM-31384: Add --log-label option

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/cli/cliLog.py
+++ b/python/lsst/daf/butler/cli/cliLog.py
@@ -21,6 +21,7 @@
 
 import datetime
 import logging
+from typing import Tuple, Optional, Dict
 
 try:
     import lsst.log as lsstLog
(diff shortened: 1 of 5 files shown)


python/lsst/daf/butler/cli/cliLog.py
```diff
@@ -84,7 +85,8 @@ class CliLog:
     that need to be closed on reset."""
 
     @classmethod
-    def initLog(cls, longlog: bool, log_tty: bool = True, log_file=()):
+    def initLog(cls, longlog: bool, log_tty: bool = True, log_file: Tuple[str, ...] = (),
+                log_label: Optional[Dict] = None):
```
andy-slac: Maybe `Dict[str, str]`?

python/lsst/daf/butler/cli/opt/options.py
```diff
@@ -122,6 +122,13 @@ def makeCollectionTypes(context, param, value):
                                     " JSON log records will be written, else formatted text log records"
                                     " will be written. This file can exist and records will be appended.")
 
+log_label_option = MWOptionDecorator("--log-label",
+                                     default=None,
+                                     multiple=True,
+                                     callback=split_kv,
+                                     type=str,
+                                     help="Keyword=value pairs to add to JSON log records.")
```
andy-slac: I think it's not only about JSON, should we say "add to MDC of log records"?

andy-slac (approved):
Looks great, couple of minor comments.

python/lsst/daf/butler/cli/opt/options.py
timj: Okay. We don't yet allow people to specify the template for long-log but in the future we could do so and that would make the content of these labels available to the formatted logs.

PR #558, opened 2021-08-13
DM-31367: Add `butler register-dataset-type` subcommand

timj:
## Checklist

- [ ] ran Jenkins


python/lsst/daf/butler/cli/cmd/commands.py
```diff
@@ -599,3 +599,31 @@ def ingest_files(**kwargs):
...
     """
     script.ingest_files(**kwargs)
+
+
+@click.command(cls=ButlerCommand)
+@repo_argument(required=True)
+@click.argument("dataset_type", required=True)
+@click.argument("storage_class", required=True, nargs=1)
+@click.argument("dimensions", required=False, nargs=-1)
+@click.option("--is-calibration/--no-is-calibration", is_flag=True, default=False,
```
parejkoj, line 609: "no" instead of "not?

python/lsst/daf/butler/cli/cmd/commands.py
```diff
@@ -599,3 +599,31 @@ def ingest_files(**kwargs):
     raw camera exposures.
     """
     script.ingest_files(**kwargs)
+
+
+@click.command(cls=ButlerCommand)
+@repo_argument(required=True)
+@click.argument("dataset_type", required=True)
+@click.argument("storage_class", required=True, nargs=1)
```
parejkoj: Why `nargs=1` here, but not above? Maybe I'm misinterpreting the `click.argument` docs, but they suggest that passing `nargs` makes the result a tuple?

python/lsst/daf/butler/script/register_dataset_type.py
```diff
@@ -0,0 +1,76 @@
...
+
+    Parameters
+    ----------
+    repo : `str`
+        URI string of the Butler repo to use.
+    dataset_type : `str`
+        The name of the new dataset type.
+    storage_class : `str`
+        The name of the storage class associated with this dataset type.
+    dimensions : `tuple` of `str`
```
parejkoj: Is the butler standard for documenting sequence types different from the lsst numpydoc one?

```
`list` [`str`]
```

https://developer.lsst.io/python/numpydoc.html#sequence-types

python/lsst/daf/butler/script/register_dataset_type.py
```diff
@@ -0,0 +1,76 @@
...
+        Dimensions associated with this dataset type. Can be empty.
+    is_calibration : `bool`
+        If `True` this dataset type maybe included in calibration
```
parejkoj: "may be", I think?

python/lsst/daf/butler/script/register_dataset_type.py
```diff
@@ -0,0 +1,76 @@
...
+    Returns
+    -------
+    inserted : `bool`
+        `True` if the dataset type was added; `False` if it was already
+        there.
+
+    Notes
+    -----
+    Components of dataset types are not supported by this command since
+    components are handled automatically.
```
parejkoj: Maybe expand on "handled automatically" a little bit: it's not clear what the user needs to do? It looks like trying to do so would result in an error, so the docstring also needs a `Raises` section.

parejkoj (commented):
Handful of comments. Thanks for doing this!

python/lsst/daf/butler/cli/cmd/commands.py
timj: The convention is `--x` to enable and `--no-x` to disable and that seems to be what click uses elsewhere. I think I could leave off the negating option but then you'd have no way to set it explicitly rather than letting it be the default.

python/lsst/daf/butler/script/register_dataset_type.py
timj: The convention is mostly desperately waiting for sphinx to catch up so that we can delete it completely (since we use type annotations almost everywhere). I'll update it.

python/lsst/daf/butler/cli/cmd/commands.py
timj: `nargs=1` results in a scalar value (I checked by printing the values immediately passed to this function). It seems to default to 1 is required=True so I can remove it.

python/lsst/daf/butler/script/register_dataset_type.py
timj: There are many exceptions that could be raised by bad arguments (bad butler location, unrecognized dimension or storage class, use of a component) and it's not really worth trying to find all possible failure modes because this script API is not public. It's entirely a thin trampoline from the click side.

parejkoj: But there's an explicit `if component: raise ValueError`, which we should document; doing so makes the components thing more clear.

PR #561, opened 2021-08-18
DM-31366: Add confirmation request to butler prune-collection

n8pease:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/cli/cmd/commands.py
+++ b/python/lsst/daf/butler/cli/cmd/commands.py
@@ -29,6 +29,7 @@ from ..opt import (
     collections_argument,
     collections_option,
     components_option,
+    confirm_option,
     dataset_type_option,
     datasets_option,
     destination_argument,
@@ -166,10 +167,22 @@ def config_validate(*args, **kwargs):
               help="Before removing the given `collection` unlink it from from this parent collection.",
               multiple=True,
               callback=split_commas)
+@confirm_option()
 @options_file_option()
 def prune_collection(**kwargs):
     """Remove a collection and possibly prune datasets within it."""
-    script.pruneCollection(**kwargs)
+    result = script.pruneCollection(**kwargs)
+    if result.confirm:
+        print("The following collections will be removed:")
+        result.removeTable.pprint_all(align="<")
+        doContinue = click.confirm("Continue?", default=False)
+    else:
+        doContinue = True
+    if doContinue:
+        result.onConfirmation()
+        print("Removed collections.")
+    else:
+        print("Aborted.")
 
 
 pruneDatasets_wouldRemoveMsg = unwrap("""The following datasets will be removed from any datastores in which
@@ -243,13 +256,6 @@ dry_run_option = MWOptionDecorator(
                 Note that a dataset can be in collections other than its RUN-type collection, and removing it
                 will remove it from all of them, even though the only one this will show is its RUN
                 collection.""")
-)
-
-
-confirm_option = MWOptionDecorator(
-    "--confirm/--no-confirm",
-    default=True,
-    help="Print expected action and a confirmation prompt before executing. Default is --confirm."
 )
 
 
--- a/python/lsst/daf/butler/script/pruneCollection.py
+++ b/python/lsst/daf/butler/script/pruneCollection.py
@@ -19,15 +19,38 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+
+from astropy.table import Table
+from dataclasses import dataclass
+from typing import Callable, Dict, List, Optional, Union
 from .. import Butler
 from .. import (
     PurgeWithoutUnstorePruneCollectionsError,
     RunWithoutPurgePruneCollectionsError,
     PurgeUnsupportedPruneCollectionsError,
 )
+from . import QueryDatasets
+from ..registry import CollectionType
 
 
-def pruneCollection(repo, collection, purge, unstore, unlink):
+class PruneCollectionResult:
+    def __init__(self, confirm: bool) -> None:
+        # if `confirm == True`, will contain the astropy table describing data
+        # that will be removed.
+        self.removeTable: Union[None, Table] = None
+        # the callback function to do the work
+        self.onConfirmation: Union[None, Callable[[], None]] = None
+        # true if the user should be shown what will be removed before pruning
+        # the collection.
+        self.confirm: bool = confirm
+
+
+def pruneCollection(repo: str,
+                    collection: str,
+                    purge: bool,
+                    unstore: bool,
+                    unlink: List[str],
+                    confirm: bool) -> Table:
     """Remove a collection and possibly prune datasets within it.
 
     Parameters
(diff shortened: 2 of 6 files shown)


python/lsst/daf/butler/cli/cmd/commands.py
```diff
@@ -148,6 +149,12 @@ def config_validate(*args, **kwargs):
         raise click.exceptions.Exit(1)
 
 
+pruneCollection_willRemoveMsg = "The following collections will be removed:"
+pruneCollection_askContinueMsg = "Continue?"
+pruneCollection_didRemoveCollections = "Removed collections."
+pruneCollection_aborted = "Aborted."
```
timj: What's the motivation for these being defined here and not embedded directly in the print messages below? I don't see them being used multiple times.

python/lsst/daf/butler/script/pruneCollection.py
```diff
@@ -42,14 +59,43 @@ def pruneCollection(repo, collection, purge, unstore, unlink):
         Same as the ``unstore`` argument to ``Butler.pruneCollection``.
     unlink: `list` [`str`]
         Same as the ``unlink`` argument to ``Butler.pruneCollection``.
+    confirm : `bool`
+        Get results for what would be removed and return the results for
+        display & confirmation, with a completion function to run after
+        confirmation.
     """
```
timj, line 77: The return value needs to be documented. The `confirm` docs talk more about the return parameter than about the boolean itself.

python/lsst/daf/butler/script/pruneCollection.py
```diff
@@ -19,15 +19,32 @@
...
     PurgeUnsupportedPruneCollectionsError,
 )
+from . import QueryDatasets
+
 
+class PruneCollectionResult:
+    def __init__(self, confirm):
+        # if `confirm == True`, will contain the astropy table describing data
+        # that will be removed.
+        self.removeTable = None
```
timj: You are using type annotations elsewhere in this file so might be worth adding some here and to the `pruneCollection` definition below.

python/lsst/daf/butler/script/pruneCollection.py
```diff
@@ -42,14 +59,43 @@ def pruneCollection(repo, collection, purge, unstore, unlink):
...
+        queryDatasets = QueryDatasets(
+            repo=repo,
+            glob=None,
+            collections=[collection],
+            where=None,
+            find_first=True,
+            show_uri=False,
+        )
+
+        collections: Dict[str, int] = defaultdict(lambda: 0)
```
timj: I think we may want to seed this dict with the output of `registry.queryCollections()` on that collection with `includeChains` because we want to be able to report that chain collections will be deleted as well as RUNs. If you prune a chained collection and that collection includes another chained collection I don't think it will turn up in the list at the moment (we should make it appear with ndatasets = 0 or something (or maybe saying "CHAIN" or "CALIBRATION"). Or else add in all the chain/calibration collections in at the end.

timj (approved):
Looks great. Only real comment is about reporting non-RUN collections that are being cleared out.

python/lsst/daf/butler/script/pruneCollection.py
n8pease: interesting, you're saying `getDatasets` will get datasets including expansion of the first chained collection, but not datasets from chained collections inside that first chained collection? That seems really odd to me (but I don't know that much about the use case I guess?)

n8pease: interesting, you're saying that `butler.registry.queryDatasets` will expand the first chained collection, but not subsequent chained collections?

timj: No. I don't think I'm saying that at all. I'm saying that datasets are by definition only associated with RUN collections. You are therefore only reporting RUN collections in your summary even though there may be some CHAINED collections also being removed. I'm just asking that you report those chained collections.

python/lsst/daf/butler/cli/cmd/commands.py
n8pease: I think it promotes reuse in the future. I also think it's easier to proof read the user-facing text when it's in one place. If you want I can embed it though (?)

timj: I see now that you did the same thing for pruneDatasets. I'm not sure it does promote reuse does it given the `pruneCollection_` prefix and the fact that there also exists a `pruneDatasets_askContinueMsg = "Continue?"`? Wouldn't the case for reuse be significantly stronger if the "Continue?" variable was reused for both commands? Now there are two sets of these variables with no code reuse coming out of it. Can you at least assign the value of `pruneCollection_askContinueMsg` from the `pruneDataset` one?

The strongest argument for doing this would be if we were internationalizing the commands such that we could switch to Spanish based on the locale setting but I imagine there is some other infrastructure we should be using for that ([gnu `gettext`](https://docs.python.org/3/library/gettext.html)?) Are you interested in I18N?

n8pease: I don't think I want to use dialog snippets between commands, the need for subtle differences can become difficult (I've experienced this mostly in the shared option help text). But, it's an interesting point that both commands implement a verify-do/abort pattern that could be worth writing a separate shared implementation. Let's do it when we have a 3rd use case ;-). In the meantime I'll remove the separate definition here.
I crossed paths only a little with I18N ages ago when I was doing more automated testing work. I'm not specifically interested in it, but am also happy to do work that needs to get done.

python/lsst/daf/butler/script/pruneCollection.py
n8pease: @timj please have a look & make sure this does what you want now, thanks!

timj: Perfect:
```
$ butler prune-collection ../pipelines_check/DATA_REPO_WITH_COMP/ demo_collection_exe
Searching collections...
The following collections will be removed:
         Collection          Collection Type Number of Datasets
---------------------------- --------------- ------------------
demo_collection_exe          CHAINED         -                 
demo_collection_exe/YYYYMMDD RUN             21                
HSC/calib                    CALIBRATION     -                 
HSC/raw/all                  RUN             1                 
refcats                      RUN             2                 
Continue? [y/N]: 
```
That's a complete report of the scary deletes that are about to happen.

python/lsst/daf/butler/script/pruneCollection.py
```diff
@@ -42,14 +59,81 @@ def pruneCollection(repo, collection, purge, unstore, unlink):
...
-        butler.pruneCollection(collection, purge, unstore, unlink)
-    except PurgeWithoutUnstorePruneCollectionsError as e:
-        raise TypeError("Cannot pass --purge without --unstore.") from e
-    except RunWithoutPurgePruneCollectionsError as e:
-        raise TypeError(f"Cannot prune RUN collection {e.collectionType.name} without --purge.") from e
-    except PurgeUnsupportedPruneCollectionsError as e:
-        raise TypeError(
-            f"Cannot prune {e.collectionType} collection {e.collectionType.name} with --purge.") from e
+    class CollectionInfo:
+        def __init__(self, count, type=None):
```
timj: Can you please put a little bit of documentation in this class? In particular, type annotations. Note that if you turn mypy on for this file it really isn't going to be happy about `self.count` being an int or a string when code later on assumes that it will be just fine to increment it. Given that it's a private class you can get away with it but you'd probably want to write this differently once we go to mypy everywhere (in some future dream land).

n8pease: Added docs & type annotations. Let me know if it looks ok? During development I temporarily turned off the `disallow_untyped_defs` exception for the scripts folder, and made sure there were no reported untyped defs for this file, so hopefully this file is totally taken care of now.

python/lsst/daf/butler/script/pruneCollection.py
```diff
@@ -42,14 +64,89 @@ def pruneCollection(repo, collection, purge, unstore, unlink):
...
-        raise TypeError("Cannot pass --purge without --unstore.") from e
-    except RunWithoutPurgePruneCollectionsError as e:
-        raise TypeError(f"Cannot prune RUN collection {e.collectionType.name} without --purge.") from e
-    except PurgeUnsupportedPruneCollectionsError as e:
-        raise TypeError(
-            f"Cannot prune {e.collectionType} collection {e.collectionType.name} with --purge.") from e
+    class CollectionInfo:
+        """Lightweight container to hold the type of collection and the number
+        of datasets in the collection if applicable."""
+        def __init__(self, count: Union[None, int], type: Union[str]):
```
timj: You can write the first one as Optional[int] and why is the second one a union at all?

python/lsst/daf/butler/script/pruneCollection.py
```diff
@@ -42,14 +64,89 @@ def pruneCollection(repo, collection, purge, unstore, unlink):
...
+            self.count = count
+            self.type: str = type
```
timj: Does it need this type annotation if you are annotating the argument? You could simplify this by using a dataclass

n8pease: The type annotation was vestigial, thanks for catching (too bad mypy does not warn for the duplicate decl). 

I did not know about dataclass (and had lamented that namedtuple is (of course) immutable). Thanks for letting me know about that.

PR #563, opened 2021-08-24
DM-31507 Modify data-structures to aid ref counting

natelust:
It is desirable that variables get cleaned up when they go out of
scope, meaning that their ref count must go to zero. Some objects
kept closures or circular references which means they must be
cleaned by the garbage collector which can happen at unpredictable
times. Make these objects able to be dropped with reference counting.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/config.py
+++ b/python/lsst/daf/butler/core/config.py
@@ -58,6 +58,18 @@ except AttributeError:
     yamlLoader = yaml.SafeLoader  # type: ignore
 
 
+def _doUpdate(d, u):
+    if not isinstance(u, collections.abc.Mapping) or \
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/core/config.py
```diff
@@ -58,6 +58,18 @@
     yamlLoader = yaml.SafeLoader  # type: ignore
 
 
+def _doUpdate(d, u):
+    if not isinstance(u, collections.abc.Mapping) or \
```
ktlim, line 62: If you write this as:
```
if not (A and B):
```
then you get free line breaks due to the parentheses.

andy-slac (approved):
Looks good (and I guess you saw flake8 complaint).

PR #564, opened 2021-08-26
DM-31520: Allow butler collection-chain to have comma-separated children

timj:
This simplifies the movement of options for pipetask which do
include commas.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/cli/cmd/commands.py
+++ b/python/lsst/daf/butler/cli/cmd/commands.py
@@ -504,7 +504,7 @@ def transfer_datasets(**kwargs):
 @click.command(cls=ButlerCommand)
 @repo_argument(required=True)
 @click.argument("parent", required=True, nargs=1)
-@click.argument("children", required=False, nargs=-1)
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/cli/cmd/commands.py
```diff
@@ -529,7 +529,11 @@ def collection_chain(**kwargs):
     CHILDREN are the collections to be used to modify the chain. The exact
     usage depends on the MODE option. When the MODE is 'pop' the CHILDREN
     should be integer indices indicating collections to be removed from
-    the current chain.
+    the current chain. The supplied values will be split on comma. For example,
```
MichelleGower: Recommend moving the new text prior to the statement about modes (or put in new "paragraph") so it's not confused as a continuation of the pop information.

MichelleGower (approved):
Changes look good (just a comment about the help string).  Ran pipelines_check demo and checked results of query-collections and the chain includes the individual input collections.

PR #566, opened 2021-08-29
DM-30178: Preparing for sqlalchemy 2 API changes

andy-slac:
Trying to make our code compatible with sqlalchemy 2 by enabling `SQLALCHEMY_WARN_20=1`, running all unit tests and watching for warnings. Here are the changes so far (which are still compatible with 1.4):
- `select()` takes columns as positional arguments instead of a list
- `Row` conversion to `dict` works differently
- `row[name]` is replaced with `row.name` or `row._mapping[name]` (or `RowCursor.mappings()` in some cases)
- `connection.execute()` takes parameters as a dict or list of dicts, not as keyword argument
- `Engine.execute()` is deprecated, have to to use `Connection.execute()`
- related to previous item, some changes to connection/transaction handling are needed in our code

With all these changes the unit tests are now warning-free. Our unit tests probably/likely don't have 100% coverage so I also searched the code for known patterns, still there is a chance that I missed some cases that need fixing. Jenkins is happy with our current 1.4 version so hopefully nothing is broken.

I did not enable [`future=True`](https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#migration-to-2-0-step-four-use-the-future-flag-on-engine) flag yet, not sure if we want it at this point. Our code still tries to support sqlalchemy 1.3, if we drop that support then enabling `future=True` is probably the right thing to do.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -15,41 +15,69 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Set up Python
-        uses: actions/setup-python@v2
+      - uses: conda-incubator/setup-miniconda@v2
         with:
           python-version: 3.8
-
-      - name: Install sqlite
-        run: sudo apt-get install sqlite libyaml-dev
+          channels: conda-forge,defaults
+          channel-priority: strict
+          show-channel-urls: true
 
       - name: Update pip/wheel infrastructure
+        shell: bash -l {0}
         run: |
-          python -m pip install --upgrade pip
-          pip install wheel
+          conda install -y -q pip wheel
+
+      - name: Install sqlite
+        shell: bash -l {0}
+        run: |
+          conda install -y -q sqlite
 
       - name: Install postgresql (server)
         run: sudo apt install postgresql
 
       - name: Install postgresql Python packages
-        run: pip install psycopg2 testing.postgresql
+        shell: bash -l {0}
+        run: |
+          conda install -y -q psycopg2
+          pip install testing.postgresql
 
       - name: Install cryptography package for moto
-        run: pip install cryptography
+        shell: bash -l {0}
+        run: |
+          conda install -y -q cryptography
 
       - name: Install WebDAV packages for testing
-        run: pip install cheroot wsgidav
+        shell: bash -l {0}
+        run: |
+          pip install cheroot wsgidav
 
       - name: Install dependencies
+        shell: bash -l {0}
         run: |
           pip install -r requirements.txt
 
       # We have two cores so we can speed up the testing with xdist
       - name: Install pytest packages
-        run: pip install pytest pytest-flake8 pytest-xdist pytest-openfiles
+        shell: bash -l {0}
+        run: |
+          conda install -y -q \
+            flake8 \
+            pytest pytest-flake8 pytest-xdist pytest-openfiles
+
+      - name: List installed packages
+        shell: bash -l {0}
+        run: |
+          conda list
+          pip list -v
 
       - name: Build and install
-        run: pip install -v .
+        shell: bash -l {0}
+        run: |
+          pip install -v .
 
       - name: Run tests
-        run: pytest -r a -v -n 3 --open-files
+        shell: bash -l {0}
+        env:
+          SQLALCHEMY_WARN_20: 1
+        run: |
+          pytest -r a -v -n 3 --open-files
--- a/python/lsst/daf/butler/registry/collections/_base.py
+++ b/python/lsst/daf/butler/registry/collections/_base.py
@@ -281,7 +281,7 @@ class DefaultChainedCollectionRecord(ChainedCollectionRecord):
             self._table.columns.position
         )
         return CollectionSearch.fromExpression(
-            [manager[row[self._table.columns.child]].name for row in self._db.query(sql)]
+            [manager[row._mapping[self._table.columns.child]].name for row in self._db.query(sql)]
         )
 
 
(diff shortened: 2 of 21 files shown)


.github/workflows/build.yaml
```diff
@@ -15,41 +15,66 @@ jobs:
...
       - name: Install WebDAV packages for testing
-        run: pip install cheroot wsgidav
+        shell: bash -l {0}
+        run: |
+          pip install cheroot wsgidav
 
       - name: Install dependencies
+        shell: bash -l {0}
         run: |
           pip install -r requirements.txt
```
erykoff, line 57: I would recommend `conda install --quiet -y --file=requirements.txt` here if we want to move to an all-conda testbed rather than getting conda and pip mixed.  But at the same time I can see the argument for only using conda to do the non-pypi packages to replace the old ubuntu versions.

andy-slac: I tried `conda` but it does not know how to handle `git+git://github.com/lsst/sphgeom@master#egg=lsst_sphgeom` which is in our `requirements.txt`

erykoff: Oh good point.  We'd have to do some more rearchitecting to make that work, and is definitely not worth it at this point.  Something like https://stackoverflow.com/questions/59864503/direct-link-to-github-in-requirements-txt-while-using-conda

.github/workflows/build.yaml
```diff
@@ -15,41 +15,66 @@ jobs:
...
+          show-channel-urls: true
 
       - name: Update pip/wheel infrastructure
         run: |
-          python -m pip install --upgrade pip
-          pip install wheel
+          conda install -y -q pip wheel
+
+      - name: Install sqlite
+        shell: bash -l {0}
```
andy-slac, line 31: `bash -l` is needed to activate conda environment (via `.profile`)

timj: Why isn't this shell option needed for the pip/wheel install above?

python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -271,17 +271,17 @@ def _update(self, manager: CollectionManager, children: CollectionSearch) -> Non
...
+        ).select_from(
             self._table
         ).where(
             self._table.columns.parent == self.key
         ).order_by(
             self._table.columns.position
         )
         return CollectionSearch.fromExpression(
-            [manager[row[self._table.columns.child]].name for row in self._db.query(sql)]
+            [manager[row._mapping[self._table.columns.child]].name for row in self._db.query(sql)]
```
timj, line 284: It does seem like a strange decision to make `_mapping` a public method.

python/lsst/daf/butler/registry/databases/postgresql.py
```diff
@@ -113,8 +115,9 @@ def transaction(self, *, interrupting: bool = False, savepoint: bool = False,
 
     def _lockTables(self, tables: Iterable[sqlalchemy.schema.Table] = ()) -> None:
         # Docstring inherited.
-        for table in tables:
-            self._connection.execute(f"LOCK TABLE {table.key} IN EXCLUSIVE MODE")
+        with self._connection() as connection:
```
timj: So this context manager frees the connection object but doesn't have any effect on the table locking?

timj (approved):
Looks great.

python/lsst/daf/butler/registry/collections/_base.py
erykoff: Very strange to name it with an underscore.  But looking here: https://docs.sqlalchemy.org/en/14/changelog/migration_20.html it says:
> This Row behaves like a named tuple, in that it acts as a sequence but also supports attribute name access, e.g. row.some_column. However, it also provides the previous “mapping” behavior via the special attribute row._mapping, which produces a Python mapping such that keyed access such as row["some_column"] can be used.

So I guess you're supposed to access it with `row.some_column` which seems really inconvenient for dynamically generated column names.

python/lsst/daf/butler/registry/databases/postgresql.py
TallJimbo: The table locking will automatically end when the current transaction ends, and closing the connection will (I think) do that.

.github/workflows/build.yaml
andy-slac: Because I missed it? Thanks for noticing, I'll fix it. And I'm not quite sure where does it install if you don't activate it, maybe in `lib/python` instead of separate environment. Would be much easier if I could run an interactive shell to look at things.

python/lsst/daf/butler/registry/collections/_base.py
andy-slac: Yes, this is a standard issue with `namedtuple`, to avoid clashes with user-defined attributes all public class methods/attributes start with underscore. New SQLAlchemy version is more strict about Row class, it was an odd mix of dict/namedtuple before, not it is just a namedtuple(-like).

python/lsst/daf/butler/registry/databases/postgresql.py
andy-slac: I believe this relies on the fact that `_lockTables()` is called from `transaction()` and in that case connection only closes when that outermost transaction closes. I should have probably used `self._session_connection` here instead of `self._connection()` even though they produce identical result. Whole connection transaction management is a bit messy, sqlalchemy 2.0 tries to simplify it but it does not map too well into what we are trying to do. Maybe we need to review it at some point. And in this particular case I think I can improve it just by passing connection as an argument to this method.

andy-slac: I updated `_lockTables()`  argument list to pass `connection` directly from transaction(), this should avoid possible issues with connection lifetime.

.github/workflows/build.yaml
```diff
@@ -15,41 +15,67 @@ jobs:
...
-        run: pip install -v .
+        shell: bash -l {0}
+        run: |
+          pip install -v .
 
       - name: Run tests
-        run: pytest -r a -v -n 3 --open-files
+        shell: bash -l {0}
+        run: |
+          pytest -r a -v -n 3 --open-files
```
timj, line 83: Can you set the sqlalchemy environment variable here so that we have a chance to notice if a bunch of new warnings turn up?

PR #568, opened 2021-09-02
DM-31623: Support day_obs and seq_num in butler.put

timj:
## Checklist

- [x] ran Jenkins


TallJimbo (approved):
Looks good.  I think the ideal implementation for `put` alone could be much simpler than the one used for `get`, because we don't have to worry about temporal dimensions for calibration lookups.  But delegating the same code certainly makes more sense right now (I just think it might change if we want to generalize this further to other interfaces).

PR #569, opened 2021-09-08
DM-31596: Export DM-28920 calibrations for use at the summit

czwa:
Add export-calibs butler subcommand.

## Checklist

- [X] ran Jenkins: https://ci.lsst.codes/job/stack-os-matrix/34905/display/redirect


python/lsst/daf/butler/script/exportCalibs.py
```diff
@@ -0,0 +1,147 @@
...
+        exported.
+
+    Raises
+    ------
+    RuntimeError :
+        Raised if the output directory already exists.
+    """
+    butler = Butler(repo, writeable=False)
+
+    calibTypes = ["bias", "dark", "flat", "defects", "linearity", "bfk", "ptc", "crosstalk"]
```
timj: I'm a bit worried about this list being hard-coded. Do you have to patch this every time a new dataset type is added? What happens if you only want to export just the bias? Why can't you query for all dataset types that have `isCalibration=True` ?

python/lsst/daf/butler/script/exportCalibs.py
```diff
@@ -0,0 +1,147 @@
...
+            log.info("Saving dataset(s)")
+            export.saveDatasets(datasetsToExport)
+        except Exception as e:
+            log.warning("Did not save datasets %s due to %s.", exportable, e)
+
+    sortedDatasets = sorted(datasetsToExport, key=lambda x: x.datasetType.name)
+
+    return Table({"calibrationType": [ref.datasetType.name for ref in sortedDatasets],
+                  "run": [ref.run for ref in sortedDatasets],
+                  "instrument": [ref.dataId["instrument"] for ref in sortedDatasets],
```
timj: Hard-coding this seems wrong to me. What if a calibration dataset type has an extra dimension? The robust answer seems to be that for all the dataset types involved -- work out all the relevant dimensions (which is not difficult to do by throwing all the dimensions of each datasetType into a set) and then for each dimension build the column and then add the column to this table.

timj (commented):
This is a quick comment rather than a review just yet but I have concerns about the hard-coding of calibTypes and the return columns in the table. I'd like you to think about making this code generic so it doesn't hard-code the calibration dimensions or calibration dataset types.

python/lsst/daf/butler/script/exportCalibs.py
```diff
@@ -0,0 +1,152 @@
...
+    -------
+    exportCollections : `list` [`str`]
+        List of collections to save on export.
+    exportDatasets : `list` [`lsst.daf.butler.queries.DatasetQueryResults`]
+        Datasets to save on export.
+    """
+    exportCollections = []
+    exportDatasets = []
+    for calibType in datasetTypes:
+        associations = registry.queryDatasetAssociations(calibType, collections=[collection])
```
timj: I think this API takes the standard "collections" types of str, iterable, regex. This means you don't need to turn it into a list and your API here doesn't need to be restrictive. Should this also constrain the search to CALIBRATION collections with `collectionTypes` parameter?

python/lsst/daf/butler/script/exportCalibs.py
```diff
@@ -0,0 +1,152 @@
...
+
+def parseCalibrationCollection(registry, collection, datasetTypes):
+    """Search a calibration collection for calibration datasets.
+
+    Parameters
+    ----------
+    registry : `lsst.daf.butler.Registry`
+        Butler registry to use.
+    collection : `str`
+        Collection to search.  This should be a CALIBRATION
```
timj, line 39: Do you enforce that it should be a CALIBRATION collection? Maybe call `getCollectionType` on it to be sure and that would allow a nice error message?

python/lsst/daf/butler/script/exportCalibs.py
```diff
@@ -0,0 +1,152 @@
...
+from .. import Butler, CollectionType
+
+
+log = logging.getLogger(__name__)
+
+
+def parseCalibrationCollection(registry, collection, datasetTypes):
+    """Search a calibration collection for calibration datasets.
+
+    Parameters
```
timj, line 34: datasetTypes parameter is not documented.

python/lsst/daf/butler/script/exportCalibs.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    butler = Butler(repo, writeable=False)
+
+    calibTypes = [datasetType for datasetType in butler.registry.queryDatasetTypes(...)
+                  if datasetType.isCalibration()]
+
+    collectionsToExport = []
+    datasetsToExport = []
+
+    for collection in collections:
```
timj: I think you have implemented more than you need to here. It seems to me that you have implemented a chain traversal routine but we already have one. I think something like:

```python
for collection in butler.registry.queryCollections(collections, flatten=True, includeChains=True,
                                                   collectionType={CollectionType.CALIBRATION, CollectionType.CHAINED}):
    collectionsToExport.append(collection)
    collectionType = butler.registry.getCollectionType(collection)
    if collectionType == CollectionType.CALIBRATION:
        exportCollections, exportDatasets = parseCalibrationCollection(butler.registry,
                                                                       collection,
                                                                       calibTypes)
        collectionsToExport.extend(exportCollections)
        datasetsToExport.extend(exportDatasets)
```
Adjust to suit, but that leverages existing APIs and is I think a lot clearer.

python/lsst/daf/butler/script/exportCalibs.py
```diff
@@ -0,0 +1,152 @@
...
+        datasetsToExport = list(set(datasetsToExport))
+
+        for exportable in collectionsToExport:
+            try:
+                export.saveCollection(exportable)
+            except Exception as e:
+                log.warning("Did not save collection %s due to %s.", exportable, e)
+
+        try:
+            log.info("Saving dataset(s)")
```
timj: Can you say you are saving N datasets (it's already a list so counting them is not expensive).

python/lsst/daf/butler/script/exportCalibs.py
```diff
@@ -0,0 +1,152 @@
...
+            export.saveDatasets(datasetsToExport)
+        except Exception as e:
+            log.warning("Did not save datasets %s due to %s.", exportable, e)
```
timj: isn't this fairly catastrophic? Doesn't it mean nothing was exported at all? Why would we then print a nice table showing all the things we tried to export but failed at?

timj (approved):
Thank you for making it more general.

I have one more substantive comment involving your reimplementation of collection unchaining which I'm pretty sure is not needed.

python/lsst/daf/butler/script/exportCalibs.py
czwa: I've removed the `try`, which was a carryover from before all of the review improvements.

PR #571, opened 2021-09-09
DM-31700: Fix parquet formatter failure on zero-index tables

taranu:
## Checklist

- [x] ran Jenkins


timj:
Normally we only check the "added a release note" task as complete if we've added a release note.  It might be worth making a bugfix release note for this change though.

erykoff (approved):

PR #573, opened 2021-09-14
DM-30266 Modify serialization of some objects

natelust:
Add/Modify serialization of some objects to support a new serialized
format of QuantumGraphs. In particular this introduces a new method
on some objects to support direct construction if the inputs are
already trusted, skipping validation steps.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/datasets/ref.py
+++ b/python/lsst/daf/butler/core/datasets/ref.py
@@ -89,6 +89,29 @@ class SerializedDatasetRef(BaseModel):
             raise ValueError(f"datasetType ({values[d]}) can not be set if component is given ({v}).")
         return v
 
+    @classmethod
+    def direct(cls, *, id: Optional[Union[str, int]] = None, datasetType: Optional[Dict[str, Any]] = None,
+               dataId: Optional[Dict[str, Any]] = None, run: str = None, component: Optional[str] = None
+               ) -> SerializedDatasetRef:
+        """Construct a `SerializedDatasetRef` directly without validators.
+
+        This differs from the pydantic "construct" method in that the arguments
+        are explicitly what the model requires, and it will recurse through
+        members, constructing them from their corresponding `direct` methods.
+
+        This method should only be called when the inputs are trusted.
+        """
+        node = SerializedDatasetRef.__new__(cls)
+        setter = object.__setattr__
+        setter(node, 'id', uuid.UUID(id) if isinstance(id, str) else id)
+        setter(node, 'datasetType',
+               datasetType if datasetType is None else SerializedDatasetType.direct(**datasetType))
+        setter(node, 'dataId', dataId if dataId is None else SerializedDataCoordinate.direct(**dataId))
+        setter(node, 'run', run)
+        setter(node, 'component', component)
+        setter(node, '__fields_set__', {'id', 'datasetType', 'dataId', 'run', 'component'})
+        return node
+
 
 DatasetId = Union[int, uuid.UUID]
 """A type-annotation alias for dataset ID which could be either integer or
@@ -249,7 +272,8 @@ class DatasetRef:
     @classmethod
     def from_simple(cls, simple: SerializedDatasetRef,
                     universe: Optional[DimensionUniverse] = None,
-                    registry: Optional[Registry] = None) -> DatasetRef:
+                    registry: Optional[Registry] = None,
+                    datasetType: Optional[DatasetType] = None) -> DatasetRef:
         """Construct a new object from simplified form.
 
         Generally this is data returned from the `to_simple` method.
@@ -265,6 +289,11 @@ class DatasetRef:
             Registry to use to convert simple form of a DatasetRef to
             a full `DatasetRef`. Can be `None` if a full description of
             the type is provided along with a universe.
+        datasetType : DatasetType, optional
+            If datasetType is supplied, this will be used as the datasetType
+            object in the resulting DatasetRef instead of being read from
+            the `SerializedDatasetRef`. This is useful when many refs share
+            the same type as memory can be saved. Defaults to None.
 
         Returns
         -------
@@ -295,10 +324,13 @@ class DatasetRef:
             # this is for mypy
             raise ValueError("Unable to determine a usable universe")
 
-        if simple.datasetType is None:
+        if simple.datasetType is None and datasetType is None:
             # mypy
             raise ValueError("The DatasetType must be specified to construct a DatasetRef")
-        datasetType = DatasetType.from_simple(simple.datasetType, universe=universe, registry=registry)
+        if datasetType is None:
+            if simple.datasetType is None:
+                raise ValueError("Cannot determine Dataset type of this serialized class")
+            datasetType = DatasetType.from_simple(simple.datasetType, universe=universe, registry=registry)
 
         if simple.dataId is None:
             # mypy
--- a/python/lsst/daf/butler/core/ddl.py
+++ b/python/lsst/daf/butler/core/ddl.py
@@ -32,6 +32,8 @@ databases, a "schema" is also another term for a namespace.
 """
 from __future__ import annotations
 
+from lsst import sphgeom
+
 __all__ = ("TableSpec", "FieldSpec", "ForeignKeySpec", "Base64Bytes", "Base64Region",
            "AstropyTimeNsecTai", "GUID")
 
@@ -127,6 +129,15 @@ class Base64Bytes(sqlalchemy.TypeDecorator):
         # native `bytes`.
         return b64decode(value.encode("ascii")) if value is not None else None
 
+    @property
+    def python_type(self) -> Type[bytes]:
+        return bytes
+
+
+# create an alias, for use below to disambiguate between the built in
+# sqlachemy type
+LocalBase64Bytes = Base64Bytes
+
 
 class Base64Region(Base64Bytes):
     """A SQLAlchemy custom type for Python `sphgeom.Region`.
@@ -145,6 +156,10 @@ class Base64Region(Base64Bytes):
         if value is None:
             return None
         return Region.decode(super().process_result_value(value, dialect))
+
+    @property
+    def python_type(self) -> Type[sphgeom.Region]:
+        return sphgeom.Region
 
 
 class AstropyTimeNsecTai(sqlalchemy.TypeDecorator):
@@ -370,7 +385,13 @@ class FieldSpec:
         type : `type`
             Python type associated with this field's (SQL) `dtype`.
         """
-        return self.dtype().python_type
+        # to construct these objects, nbytes keyword is needed
+        if issubclass(self.dtype, LocalBase64Bytes):
+            # satisfy mypy for something that must be true
+            assert self.nbytes is not None
+            return self.dtype(nbytes=self.nbytes).python_type
+        else:
+            return self.dtype().python_type  # type: ignore
 
 
 @dataclass
--- a/python/lsst/daf/butler/core/dimensions/_coordinate.py
+++ b/python/lsst/daf/butler/core/dimensions/_coordinate.py
@@ -73,6 +73,25 @@ class SerializedDataCoordinate(BaseModel):
     dataId: Dict[str, DataIdValue]
     records: Optional[Dict[str, SerializedDimensionRecord]] = None
 
+    @classmethod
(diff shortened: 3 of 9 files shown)


python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -89,6 +89,29 @@ def _check_component(cls, v: Any, values: Dict[str, Any]) -> Any:  # noqa: N805
             raise ValueError(f"datasetType ({values[d]}) can not be set if component is given ({v}).")
         return v
 
+    @classmethod
+    def direct(cls, *, id: Optional[Union[str, int]] = None, datasetType: Optional[Dict[str, Any]] = None,
+               dataId: Optional[Dict[str, Any]] = None, run: str = None, component=None
+               ) -> SerializedDatasetRef:
+        """Construct a `SerializedDatasetRef` directly without validators.
+
+        This differs from PyDantics construct method in that the arguments are
```
TallJimbo:
```suggestion
        This differs from pydantic construct method in that the arguments are
```
This seems to be the spelling and only capitalization in use in their docs.

I'm also very concerned that we have to add these methods at all; it's bad enough that we have to add `to_simple` and `from_simple` to everything, but this really doesn't seem scale up to multiple levels of composition or normalization.  Why can't we use `construct`?

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -89,6 +89,29 @@ def _check_component(cls, v: Any, values: Dict[str, Any]) -> Any:  # noqa: N805
...
+        explicitly what the model requires, and it will recurse through
+        members, constructing them from their corresponding `direct` methods.
+
+        This method should only be called when the inputs are trusted.
+        """
+        node = SerializedDatasetRef.__new__(cls)
+        setter = object.__setattr__
```
TallJimbo, line 105: This degree of poking at class internals worries me a lot; we seem to be assuming a lot about pydantic implementation details.  If we can't use `construct` instead of adding these methods, could we at least use it inside the `direct` implementations?

python/lsst/daf/butler/core/dimensions/_records.py
```diff
@@ -298,7 +330,10 @@ def to_simple(self, minimal: bool = False) -> SerializedDimensionRecord:
                     # and also history. Here use a different approach.
                     # This code needs to be migrated to sphgeom
                     mapping[k] = v.encode().hex()
-
+                if isinstance(v, bytes):
+                    # We actually cant handle serializing out to bytes for hash
```
TallJimbo:
```suggestion
                    # We actually can't handle serializing out to bytes for hash
```

python/lsst/daf/butler/core/quantum.py
```diff
@@ -32,13 +32,98 @@
...
+               dataId: Optional[Dict],
+               datasetTypeMapping: Mapping[str, Dict],
+               initInputs: Mapping[str, Tuple[Dict, List[int]]],
+               inputs: Mapping[str, List[Tuple[Dict, List[int]]]],
+               outputs: Mapping[str, List[Tuple[Dict, List[int]]]],
+               dimensionRecords: Optional[Dict[int, Dict]]
+               ) -> SerializedQuantum:
+        """Construct a `SerializedQuantum` directly without validators.
+
+        This differs from PyDantics construct method in that the arguments are
```
TallJimbo:
```suggestion
        This differs from pydantic construct method in that the arguments are
```

python/lsst/daf/butler/core/dimensions/_records.py
```diff
@@ -159,6 +159,38 @@ class Config:
             }
         }
 
+    @classmethod
+    def direct(cls, *, definition, record) -> SerializedDimensionRecord:
+        """Construct a `SerializedDimensionRecord` directly without validators.
+
+        This differs from PyDantics construct method in that the arguments are
```
TallJimbo:
```suggestion
        This differs from pydantic construct method in that the arguments are
```

python/lsst/daf/butler/core/dimensions/_graph.py
```diff
@@ -58,6 +58,21 @@ class SerializedDimensionGraph(BaseModel):
 
     names: List[str]
 
+    @classmethod
+    def direct(cls, *, names: List[str]) -> SerializedDimensionGraph:
+        """Construct a `SerializedDimensionGraph` directly without validators.
+
+        This differs from PyDantics construct method in that the arguments are
```
TallJimbo:
```suggestion
        This differs from pyantic construct method in that the arguments are
```

python/lsst/daf/butler/core/dimensions/_coordinate.py
```diff
@@ -73,6 +73,25 @@ class SerializedDataCoordinate(BaseModel):
     dataId: Dict[str, DataIdValue]
     records: Optional[Dict[str, SerializedDimensionRecord]] = None
 
+    @classmethod
+    def direct(cls, *, dataId: Dict[str, DataIdValue], records: Dict[str, Dict]) -> SerializedDataCoordinate:
+        """Construct a `SerializedDataCoordinate` directly without validators.
+
+        This differs from PyDantics construct method in that the arguments are
```
TallJimbo:
```suggestion
        This differs from pydantic construct method in that the arguments are
```

python/lsst/daf/butler/core/quantum.py
```diff
@@ -101,6 +186,194 @@ def __init__(self, *, taskName: Optional[str] = None,
...
+                typeMapping[key.name] = key.to_simple()
+            # for each input type there are a list of inputs, collect them
+            tmp = []
+            for e in value:
+                simp = e.to_simple()
+                # This container will hold ids (hashes) that point to all the
+                # dimension records within the SerializedDatasetRef dataId
+                # These dimension records repeat in almost every DatasetRef
+                # So it is hugely wasteful in terms of disk and cpu time to
+                # store them over and over again.
```
TallJimbo, line 258: `hash` does not generate anything like a unique hash, so you are potentially throwing away a lot of records that aren't actually duplicates due to collisions.

But `DimensionRecords` are already hashable and equality comparable, with those delegating to `record.dataId` (which is also hashable and equality comparable).  All of the state for that is embedded in the record itself.  So you can (for each dimension element):
- use a regular `set` to deduplicate all of the records on write
- write all records for each dimension element to one JSON list
- save the data IDs of the records you want in the quanta (this is a bit bigger than the number you're saving, but I don't see a good way around it)
- load those JSON lists into dicts keyed by data ID
- look up records in those dicts via the data IDs saved in the quanta.

You might be able to get away with just using an autoincrement key instead to preserve more of the existing code here, but keying on the data ID and avoiding integers seems like it'll be a lot less fragile.

That said, this is _exactly_ the stuff my containers on DM-30332 were designed to handle; I'll come back to that in a high-level comment later.

TallJimbo (changes requested):
I've got some pretty big concerns here, and I'm not actually sure it's worth trying to resolve them on this branch - originally this ticket was just going to be about making quantum node IDs into UUIDs to make other Quantum serialization changes easier, but:

- this now overlaps enough with DM-30332 that I'd really like to at least try to reconcile the branches before merging, to see if we've solved some problems in fundamentally different ways;
- the `hash()`-keyed `DimensionRecord` normalization here is broken, but that is _the_ problem that DM-30332 aims to solve much more rigorously;
- if we can possibly avoid setting the precedent of adding these `direct` methods to all of our serialization structures, I think I'll pay off in the long run.

I'd also like @timj to sign off on the serialization changes first, if we do plan to merge this soon; this system is more in his domain than mine (note that he may also object to some of the things I'm doing on DM-30332).

So, where does this leave us?  I was hoping not to get back to DM-30332 until after some big query-system changes, because QG serialiization and state stuff in general is a lower priority from a high-level perspective than QG generation stuff right now.  But this ticket is mostly done, and there's a lot of content on the DM-30332 branches that I don't want to get too stale, either.  And if I can get DM-30332 (or at least the parts that are basically done) out for review soon, we could rebase this branch on that and delegate to it for the `DimensionRecord` normalization logic instead of fixing it independently.

There's also the big catch that I didn't try to avoid pydantic validation on DM-30332, and the fact that @natelust has done so here makes me worry that DM-30332 could push us off a performance cliff. Maybe I can learn enough from his work here to fix my branch in that respect, but I'm also worried that just turning off validation sort of defeats the purpose of using pydantic, and it might be a sign of bigger problems to come.  I'm been wondering for a while now whether QG is going to be the thing that finally pushes us into adding some compiled-language code to the middleware; doing graph algorithms with hundreds of thousands of nodes in Python just strikes me as bonkers, period.

Thoughts on how to proceed welcome.

python/lsst/daf/butler/core/datasets/ref.py
TallJimbo: Nate explained to me that this is because `construct` isn't recursive (which I can now see in the docs; it just didn't make sense without a better understanding of what `construct` does).  I can see that we're kind of boxed in; that makes me concerned about pydantic possibly being something that doesn't really meet our needs (which is not to say that I'm aware of anything better), but I can't fault this ticket for that.

timj: My motivation for pydantic was butler client/server and fastapi automatically vetting content before it is even passed to butler.

python/lsst/daf/butler/core/quantum.py
TallJimbo: Nate and I discussed this live, and the plan is to use an autoincrement key (a counter in this loop), in order to keep his serialization data structure and minimize further changes on this branch.  DM-30332 may come along and change that later, but we'll cross that bridge when we come to it.

python/lsst/daf/butler/core/ddl.py
```diff
@@ -126,6 +128,15 @@ def process_result_value(self, value: Optional[str], dialect: sqlalchemy.engine.
         # native `bytes`.
         return b64decode(value.encode("ascii")) if value is not None else None
 
+    @property
+    def python_type(self) -> Type[bytes]:
+        return bytes
+
+
+# create and alias, for use below to disambiguate between the built in
```
TallJimbo:
```suggestion
# create an alias, for use below to disambiguate between the built in
```

python/lsst/daf/butler/core/dimensions/_records.py
```diff
@@ -159,6 +159,38 @@ class Config:
...
+            return False
+        if self.record.keys() != other.record.keys():
+            return False
+        for key in self.record:
+            if self.record[key] != other.record[key]:
+                return False
+        return True
+
+    def __hash__(self) -> int:
+        return hash((self.definition, tuple((k, v) for k, v in self.record.items())))
```
TallJimbo: I'd recommend removing this (and `__eq__`) after you switch to the autoincrement keys in serialization.

python/lsst/daf/butler/core/dimensions/_records.py
```diff
@@ -353,6 +388,8 @@ def from_simple(cls, simple: SerializedDimensionRecord,
         if (reg := "region") in rec:
             encoded = bytes.fromhex(rec[reg])
             rec[reg] = lsst.sphgeom.Region.decode(encoded)
+        if (hsh := "hash") in rec:
+            rec[hsh] = bytes.fromhex(rec[hsh].decode())
```
TallJimbo, line 384: FWIW, my solution to this on DM-30332 was 6e9ed39064ff4f949e36569063323a3639465eec, uses a pydantic validator to do the conversion instead.  I can't tell from a quick glance which is better, so let's go with your version, which has been more thoroughly tested, and we can come back to it if necessary later.

timj:
There don't seem to be any tests that use the serialization `direct` methods. This seems to be something that daf_butler should be testing rather than expecting pipe_base to do it.

natelust:
@timj or @timj can one of you look this over once more, and or mark it reviewed? One or two small things (lint problem in pipe_base) and I hope tomorrow is the day.

python/lsst/daf/butler/core/datasets/ref.py
timj: @natelust can you respond to @TallJimbo 's question here please?

TallJimbo: He and I discussed this out of band.  I'm not thrilled, but I don't see a great alternative and I'm mostly disappointed in pydantic.  Hopefully we can figure out a way to clean it up later.

PR #574, opened 2021-09-15
DM-29531: Update IsrCalib so it can serve as StorageClass for subclasses

czwa:
Add formatter and storageClass information for IsrCalib.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/doc/changes/DM-29531.misc.rst
+++ b/doc/changes/DM-29531.misc.rst
@@ -1,0 +1,1 @@
+Add formatter and storageClass information for IsrCalib.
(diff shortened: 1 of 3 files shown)


timj (approved):
This looks fine. At some point we need to clear out the entries for the older calib classes but there's no rush for that.

You could add `inheritsFrom:` to the yaml for CrossTalkCalib for example and that would mean that the actual StorageClass for CrossTalkCalib will explicitly be related to IsrCalib but it's not required.

PR #575, opened 2021-09-16
DM-31583: recognize doomed queries before we execute them

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/queries/_builder.py
+++ b/python/lsst/daf/butler/registry/queries/_builder.py
@@ -163,6 +163,12 @@ class QueryBuilder:
         baseColumnNames.update(datasetType.dimensions.required.names)
         for rank, collectionRecord in enumerate(collections.iter(self._managers.collections,
                                                                  collectionTypes=collectionTypes)):
+            # Only include collections that (according to collection summaries)
+            # might have datasets of this type and governor dimensions
+            # consistent with the query's WHERE clause.
+            collection_summary = self._managers.datasets.getCollectionSummary(collectionRecord)
+            if not collection_summary.is_compatible_with(datasetType, self.summary.where.restriction):
+                continue
             if collectionRecord.type is CollectionType.CALIBRATION:
                 # If collection name was provided explicitly then say sorry,
                 # otherwise collection is a part of chained one and we skip it.
--- a/python/lsst/daf/butler/registry/queries/_builder.py
+++ b/python/lsst/daf/butler/registry/queries/_builder.py
@@ -161,6 +161,9 @@ class QueryBuilder:
         runKeyName = self._managers.collections.getRunForeignKeyName()
         baseColumnNames = {"id", runKeyName, "ingest_date"} if isResult else set()
         baseColumnNames.update(datasetType.dimensions.required.names)
+        if not findFirst:
+            calibration_collections = []
+            other_collections = []
         for rank, collectionRecord in enumerate(collections.iter(self._managers.collections,
                                                                  collectionTypes=collectionTypes)):
             # Only include collections that (according to collection summaries)
@@ -183,21 +186,53 @@ class QueryBuilder:
                             f"Find-first query for dataset type '{datasetType.name}' in CALIBRATION-type "
                             f"collection '{collectionRecord.name}' is not yet supported."
                         )
+                    else:
+                        calibration_collections.append(collectionRecord)
                 else:
                     # We can never find a non-calibration dataset in a
                     # CALIBRATION collection.
                     continue
-            ssq = datasetRecordStorage.select(collection=collectionRecord,
-                                              dataId=SimpleQuery.Select,
-                                              id=SimpleQuery.Select if isResult else None,
-                                              run=SimpleQuery.Select if isResult else None,
-                                              ingestDate=SimpleQuery.Select if isResult else None)
-            if ssq is None:
-                continue
-            assert {c.name for c in ssq.columns} == baseColumnNames
-            if findFirst:
+            elif findFirst:
+                # If findFirst=True, each collection gets its own subquery so
+                # we can add a literal rank for it.
+                ssq = datasetRecordStorage.select(
+                    collectionRecord,
+                    dataId=SimpleQuery.Select,
+                    id=SimpleQuery.Select if isResult else None,
+                    run=SimpleQuery.Select if isResult else None,
+                    ingestDate=SimpleQuery.Select if isResult else None,
+                )
+                if ssq is None:
+                    continue
+                assert {c.name for c in ssq.columns} == baseColumnNames
                 ssq.columns.append(sqlalchemy.sql.literal(rank).label("rank"))
-            subsubqueries.append(ssq.combine())
+                subsubqueries.append(ssq.combine())
+            else:
+                # If findFirst=False, we have one subquery for all CALIBRATION
+                # collections and one subquery for all other collections; we'll
+                # assemble those later after grouping by collection type.
+                other_collections.append(collectionRecord)
+        if not findFirst:
+            if other_collections:
+                ssq = datasetRecordStorage.select(
+                    *other_collections,
+                    dataId=SimpleQuery.Select,
+                    id=SimpleQuery.Select if isResult else None,
+                    run=SimpleQuery.Select if isResult else None,
+                    ingestDate=SimpleQuery.Select if isResult else None,
+                )
+                if ssq is not None:
+                    subsubqueries.append(ssq.combine())
+            if calibration_collections:
+                ssq = datasetRecordStorage.select(
+                    *calibration_collections,
+                    dataId=SimpleQuery.Select,
+                    id=SimpleQuery.Select if isResult else None,
+                    run=SimpleQuery.Select if isResult else None,
+                    ingestDate=SimpleQuery.Select if isResult else None,
+                )
+                if ssq is not None:
+                    subsubqueries.append(ssq.combine())
         if not subsubqueries:
             return False
         # Although one would expect that these subqueries can be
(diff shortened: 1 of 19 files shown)


python/lsst/daf/butler/registry/summaries.py
```diff
@@ -311,6 +317,60 @@ def union(self, *others: CollectionSummary) -> CollectionSummary:
...
+                    if values_in_self.isdisjoint(values_in_other):
+                        assert values_in_other, f"No valid values in restriction for dimension {governor}."
+                        if rejections is not None:
+                            rejections.append(
+                                f"No datasets with {governor.name} in {values_in_other} "
+                                f"in collection {name!r}."
+                            )
+                        return False
+        return True
+
```
andy-slac, line 373: General comment - I'm checking generated sphinx doc and I can't find anything generated for classes in this module, is this intentional?

python/lsst/daf/butler/registry/queries/_results.py
```diff
@@ -304,17 +304,78 @@ def findDatasets(self, datasetType: Union[DatasetType, str], collections: Any, *
...
+        exact : `bool`, optional
+            If `True`, run the full query and perform post-query filtering if
+            needed to account for that filtering in the count.  If `False`, the
+            result may be an upper bound.
+
+        Returns
+        -------
+        count : `int`
+            The number of rows the query would return, or an upper bound if
+            ``postprocess=False``.
```
andy-slac: `exact=False`?

python/lsst/daf/butler/registry/queries/_query.py
```diff
@@ -178,6 +185,138 @@ def datasetType(self) -> Optional[DatasetType]:
...
+        db : `Database`
+            Object managing the database connection.
+        region : `sphgeom.Region`, optional
+            A region that any result-row regions must overlap in order to be
+            yielded.  If not provided, this will be ``self.whereRegion``, if
+            that exists.
+
+        Returns
+        -------
+        messages : `Iterator` [ `str` ]
```
andy-slac, line 299: Should this be 

    Yields
    ------
    message : str

python/lsst/daf/butler/registry/queries/_results.py
```diff
@@ -304,17 +304,78 @@ def findDatasets(self, datasetType: Union[DatasetType, str], collections: Any, *
...
+            `True` if the query would (or might, depending on arguments) yield
+            result rows.  `False` if it definitely would not.
+        """
+        return self._query.any(self._db, execute=execute, exact=exact)
+
+    def explain_no_results(self) -> Iterator[str]:
+        """Return human-readable messages that may help explain why the query
+        yields no results.
+
+        Returns
```
andy-slac, line 367: Same here, should it be `Yields` instead of `Returns`?

python/lsst/daf/butler/registry/queries/_results.py
```diff
@@ -304,17 +304,78 @@ def findDatasets(self, datasetType: Union[DatasetType, str], collections: Any, *
...
+        Messages related to post-query filtering are only available if `rows`,
+        `any`, or `count` was already called with the same region (with
+        ``postprocess=True`` for the latter two).
+
+        At present, this method only returns messages that are generated while
+        the query is being built or filtered.  In the future, it may perform
+        its own new follow-up queries, which users may wish to short-circuit
+        simply by not continuing to iterate over its results.
+        """
+        return self._query.explain_no_results(self._db)
```
andy-slac, line 384: `yield from` instead of `return` (but probably does not matter in this case)?

python/lsst/daf/butler/registry/queries/_results.py
```diff
@@ -370,6 +431,77 @@ def expanded(self) -> DatasetQueryResults:
...
+        exact : `bool`, optional
+            If `True`, run the full query and perform post-query filtering if
+            needed to account for that filtering in the count.  If `False`, the
+            result may be an upper bound.
+
+        Returns
+        -------
+        count : `int`
+            The number of rows the query would return, or an upper bound if
+            ``postprocess=False``.
```
andy-slac: `exact=False`?

python/lsst/daf/butler/registry/queries/_results.py
```diff
@@ -370,6 +431,77 @@ def expanded(self) -> DatasetQueryResults:
...
+        raise NotImplementedError()
+
+    @abstractmethod
+    def explain_no_results(self) -> Iterator[str]:
+        """Return human-readable messages that may help explain why the query
+        yields no results.
+
+        Returns
+        -------
+        messages : `Iterator` [ `str` ]
```
andy-slac, line 500: Same here

python/lsst/daf/butler/registry/queries/_results.py
```diff
@@ -393,17 +525,26 @@ class ParentDatasetQueryResults(DatasetQueryResults):
...
         self._query = query
         self._components = components
         self._records = records
-        assert query.datasetType is not None, \
+        if datasetType is None:
+            datasetType = query.datasetType
+        assert datasetType is not None, \
             "Query used to initialize dataset results must have a dataset."
-        assert query.datasetType.dimensions == query.graph
+        assert datasetType.dimensions == query.graph
```
andy-slac: Message could be useful here too.

python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -127,6 +126,24 @@ def loadData(self, registry: Registry, filename: str):
...
+        expected : `list`
+            A list of `DataCoordinate` o `DatasetRef` objects that should be
+            equal to results of the query, aside from ordering.
+        """
+        self.assertCountEqual(list(results), expected)
+        self.assertEqual(results.count(), len(expected))
+        if expected:
+            self.assertTrue(results.any())
+        else:
+            self.assertFalse(results.any())
```
andy-slac, line 145: Is not `any()` allowed to be true even if `expected` is empty. Or do we only test cases when `any` is definitely false, or expected is never empty?

python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -2236,3 +2234,75 @@ def testBindInQueryDatasets(self):
...
+        # Default test dataset has two collections, each with both flats and
+        # biases.  Add a new collection with only biases.
+        registry.registerCollection("biases", CollectionType.TAGGED)
+        registry.associate("biases", registry.queryDatasets("bias", collections=["imported_g"]))
+        # First query yields two results, and involves no postprocessing.
+        query1 = registry.queryDataIds(["physical_filter"], band="r")
+        self.assertTrue(query1.any(execute=False, exact=False))
+        self.assertTrue(query1.any(execute=True, exact=False))
+        self.assertTrue(query1.any(execute=True, exact=True))
+        self.assertEqual(query1.count(exact=False), 2)
```
andy-slac, line 2262: Should it be `assertGreaterEqual`?

python/lsst/daf/butler/registry/queries/_builder.py
```diff
@@ -253,7 +253,18 @@ def joinDataset(self, datasetType: DatasetType, collections: Any, *,
...
+            # results; this should never get executed, but downstream code
+            # still needs to access the SQLAlchemy column objects.
+            ssq = datasetRecordStorage.select(
+                dataId=SimpleQuery.Select,
+                id=SimpleQuery.Select if isResult else None,
+                run=SimpleQuery.Select if isResult else None,
+                ingestDate=SimpleQuery.Select if isResult else None,
+            )
+            if findFirst:
+                ssq.columns.append(sqlalchemy.sql.literal(rank).label("rank"))
```
andy-slac, line 266: I guess we don't expect any dimension to be named `rank`?

andy-slac (approved):
Looks good, few comments for docstrings mostly.

python/lsst/daf/butler/registry/summaries.py
TallJimbo: Unintentional and now fixed.

python/lsst/daf/butler/registry/queries/_query.py
TallJimbo: I think either is correct, and it formally comes down to the subtlety of whether you want to declare that an iterator-returning function is implemented as a generator.  I generally prefer to hide that, which I gather is kind of unpythonic but better for defining abstract interfaces.

python/lsst/daf/butler/registry/queries/_results.py
TallJimbo: I'm pretty sure that if it does matter, it's only because of performance, and `return` is more likely to win on that question (but this gets back to whether to the iterator vs. generator type covariance question).

python/lsst/daf/butler/registry/tests/_registry.py
TallJimbo: This leaves `exact` defaulted to `True`, so it provides a stronger guarantee (at the cost of testing a bit less, since that code path is similar to just executing the query).

python/lsst/daf/butler/registry/tests/_registry.py
TallJimbo: This test is constructed such that equality is safe.  That makes it a bit more dependent on the implementation than I'd normally like in a test, but I think it's appropriate in this case to make sure I'm testing all the code paths (in other words, if the implementation changed enough to break this, the tests really should be modified).

python/lsst/daf/butler/registry/queries/_builder.py
TallJimbo: That is indeed a limitation, but it's not a new one on this branch (just shuffled code around).

PR #576, opened 2021-09-17
DM-31785: Improve efficiency of butler.transfer_from

timj:
* Add `datastore.mexists()` and `ButlerURI.mexists()`
* Use parallel queries of object store.
* Cache result of initial check for registry entries and reuse for check inside datastore.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
+++ b/python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
@@ -675,19 +675,15 @@ class ButlerURI:
             Mapping of original URI to boolean indicating existence.
         """
         exists_executor = concurrent.futures.ThreadPoolExecutor(max_workers=MAX_WORKERS)
-
-        def wrapper(uri: ButlerURI) -> Tuple[ButlerURI, bool]:
-            try:
-                exists = uri.exists()
-            except Exception:
-                exists = False
-            return (uri, exists)
-
-        future_exists = [exists_executor.submit(wrapper, uri) for uri in uris]
+        future_exists = {exists_executor.submit(uri.exists): uri for uri in uris}
 
         results: Dict[ButlerURI, bool] = {}
         for future in concurrent.futures.as_completed(future_exists):
-            uri, exists = future.result()
+            uri = future_exists[future]
+            try:
+                exists = future.result()
+            except Exception:
+                exists = False
             results[uri] = exists
         return results
 
--- a/python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
+++ b/python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
@@ -62,6 +62,11 @@ ESCAPES_RE = re.compile(r"%[A-F0-9]{2}")
 # Precomputed escaped hash
 ESCAPED_HASH = urllib.parse.quote("#")
 
+# Maximum number of worker threads for parallelized operations.
+# If greater than 10, be aware that this number has to be consistent
+# with connection pool sizing (for example in urllib3).
+MAX_WORKERS = 10
+
 
 class ButlerURI:
     """Convenience wrapper around URI parsers.
@@ -670,11 +675,7 @@ class ButlerURI:
         """
         import concurrent.futures
 
-        # The max worker count here must not exceed the max connection pools
-        # in whatever system is doing the network connection. The default
-        # is 10 for urllib3.
-        max_workers = 10
-        exists_executor = concurrent.futures.ThreadPoolExecutor(max_workers=max_workers)
+        exists_executor = concurrent.futures.ThreadPoolExecutor(max_workers=MAX_WORKERS)
 
         def wrapper(uri: ButlerURI) -> Tuple[ButlerURI, bool]:
             try:
--- a/python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
+++ b/python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
@@ -22,6 +22,7 @@
 from __future__ import annotations
 
 import contextlib
+import concurrent.futures
 import urllib.parse
 import posixpath
 import copy
@@ -673,8 +674,6 @@ class ButlerURI:
         existence : `dict` of [`ButlerURI`, `bool`]
             Mapping of original URI to boolean indicating existence.
         """
-        import concurrent.futures
-
         exists_executor = concurrent.futures.ThreadPoolExecutor(max_workers=MAX_WORKERS)
 
         def wrapper(uri: ButlerURI) -> Tuple[ButlerURI, bool]:
--- a/python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
+++ b/python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
@@ -41,6 +41,7 @@ from typing import (
     Iterable,
     Iterator,
     List,
+    Dict,
     Optional,
     Tuple,
     Type,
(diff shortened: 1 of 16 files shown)


python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
```diff
@@ -653,6 +654,43 @@ def exists(self) -> bool:
...
+        -------
+        existence : `dict` of [`ButlerURI`, `bool`]
+            Mapping of original URI to boolean indicating existence.
+        """
+        import concurrent.futures
+
+        # The max worker count here must not exceed the max connection pools
+        # in whatever system is doing the network connection. The default
+        # is 10 for urllib3.
+        max_workers = 10
```
ktlim: Seems strange to have a hard-coded constant like this.  Seems like it should be at minimum a class variable or perhaps a kwarg.

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
```diff
@@ -653,6 +654,43 @@ def exists(self) -> bool:
...
+        exists_executor = concurrent.futures.ThreadPoolExecutor(max_workers=max_workers)
+
+        def wrapper(uri: ButlerURI) -> Tuple[ButlerURI, bool]:
+            try:
+                exists = uri.exists()
+            except Exception:
+                exists = False
+            return (uri, exists)
+
+        future_exists = [exists_executor.submit(wrapper, uri) for uri in uris]
```
ktlim: Can't this all be replaced by `return dict(exists_executor.map(wrapper, uris))`?

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1949,21 +2127,57 @@ def transfer_from(self, source_datastore: Datastore, refs: Iterable[DatasetRef],
...
-                    source_records[missing].extend(records)
-                else:
-                    log.warning("Asked to transfer dataset %s but no file artifacts exist for it.",
-                                id_to_ref[missing])
+            def chunk_iterable(data: Collection[Any], chunk_size: int = 10_000) -> Iterator[Iterable[Any]]:
+                """Converts any iterable that can be sized to a chunked
+                iterable."""
+                it = iter(data)
+                for i in range(0, len(data), chunk_size):
+                    yield tuple(element for element in itertools.islice(it, chunk_size))
```
ktlim: I don't think you need the `element for element in`; simply converting to `tuple` is enough.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1244,6 +1248,175 @@ def knows(self, ref: DatasetRef) -> bool:
...
+            if the caller is not interested.
+
+        Returns
+        -------
+        existence : `dict` of [`DatasetRef`, `bool`]
+            Mapping from dataset to boolean indicating existence.
+        """
+        chunk_size = 10_000
+        dataset_existence: Dict[DatasetRef, bool] = {}
+        refs = list(refs)
```
ktlim: Why two different ways of chunking?  `itertools.islice()` below in `transfer_from()` versus materializing the list and using regular slicing here?

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1949,21 +2127,57 @@ def transfer_from(self, source_datastore: Datastore, refs: Iterable[DatasetRef],
...
-                # Not all components can be guaranteed to exist so this
-                # list has to filter those by checking to see if the
-                # artifact is really there.
-                records = [info for location, info in expected if location.uri.exists()]
-                if records:
-                    source_records[missing].extend(records)
-                else:
-                    log.warning("Asked to transfer dataset %s but no file artifacts exist for it.",
-                                id_to_ref[missing])
+            def chunk_iterable(data: Collection[Any], chunk_size: int = 10_000) -> Iterator[Iterable[Any]]:
```
ktlim: `Iterable` or `Collection`?

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1244,6 +1248,175 @@ def knows(self, ref: DatasetRef) -> bool:
...
+        n_refs = len(refs)
+        log.log(VERBOSE, "Checking for the existence of artifacts for %d datasets in datastore", n_refs)
+        n_found_total = 0
+        n_checked = 0
+        for i in range(0, n_refs, chunk_size):
+            chunk_result = self._mexists(refs[i:i + chunk_size], artifact_existence)
+            if log.isEnabledFor(VERBOSE):
+                n_results = len(chunk_result)
+                n_checked += n_results
+                n_found = 0
```
ktlim: `n_found = sum(chunk_result.values())`

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
```diff
@@ -653,6 +654,43 @@ def exists(self) -> bool:
...
+        """Check for existence of multiple URIs at once.
+
+        Parameters
+        ----------
+        uris : iterable of `ButlerURI`
+            The URIs to test.
+
+        Returns
+        -------
+        existence : `dict` of [`ButlerURI`, `bool`]
```
ktlim, line 674: I think that with proper handling this (and the similar return for `Datastore.mexists()`) can be replaced by a simpler `set of ButlerURI`, but that can be left for later.

ktlim (approved):

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
timj: Okay. Will move it to be a package constant for now. This will get revisited when I have to implement the parallelized transfer code.

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
timj: Will look. I'm a newbie on this sort of thing so was mostly using the code that was in bps as a template.

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
timj: I haven't yet understood how one set can do it. I can see how a found set and not-found set can do it.

python/lsst/daf/butler/core/_butlerUri/_butlerUri.py
timj: I changed it to use the dict approach recommended in the python docs and simplified it a lot.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: Am I allowed to assume that bool will translate as 1 or 0 because there is an implicit `int(True)` involved in this sum?

timj: Seems like I am. Will change.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: I'm a bit confused because I could have sworn that I had already replied to this comment. I can't find it though.

Some of this is history and some of it is the code (not just this bit) being terrified that it might be called with an `Iterator` and not a `Collection`. In this particular case it is probably best to only create new small lists rather than one big list.

There are probably many cases in the code where if we simply assume up top that it was a List then things would be simpler. The problem is that users tend to pass in either an iterator (as returned by `queryDatasets`) or a set (which is the standard advice for what to do with `queryDatasets` output). Iterator isn't Sized but set is. Set is a Collection but you can't index into it.

timj: I moved `chunk_iterable` to the utilities package and reimplemented it so it would work without needing `len()` to work. Even though it's only 3 lines long I think it is easier to use as `for chunk in chunk_iterable(x)` format but maybe it's one of those cases where:

```python
for chunk in chunk_iterable(data, 4):
    # use chunk
```
vs
```python
while chunk := tuple(itertools.islice(iter(data), 4)):
    # use chunk
```
is not that much harder to use if you can remember it.

PR #577, opened 2021-09-21
DM-31826: Make ButlerURI.transfer_from check for identical in and out files

timj:
Sometimes even if the URIs are different they point to the same
file. This must be checked to ensure we don't delete the file
by mistake. If they are identical the transfer is assumed
to have succeeded.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/_butlerUri/file.py
+++ b/python/lsst/daf/butler/core/_butlerUri/file.py
@@ -171,6 +171,12 @@ class ButlerFileURI(ButlerURI):
             is_temporary = local_uri.isTemporary
             local_src = local_uri.ospath
 
+            # Short circuit if the URIs are identical immediately.
+            if self == local_uri:
+                log.debug("Target and destination URIs are identical: %s, returning immediately."
+                          " No further action required.", self)
+                return
+
             # Default transfer mode depends on whether we have a temporary
             # file or not.
             if transfer == "auto":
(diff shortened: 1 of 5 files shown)


python/lsst/daf/butler/core/_butlerUri/file.py
```diff
@@ -196,9 +202,35 @@ def transfer_from(self, src: ButlerURI, transfer: str,
...
+            # (such as soft links in the path, or they really are the same).
+            # In that case log a message and return as if the transfer
+            # completed (it technically did). A temporary file download
+            # can't be the same so the test can be skipped.
+            if dest_stat and not is_temporary:
+                # realpath could be used on self but the only reliable
+                # way to know if these are the same file (could be hard-links)
+                # is to check with os.lstat. Do not read through soft links
+                # since it's simpler logic to delete and recreate.
+                local_src_stat = os.lstat(local_src)
```
andy-slac, line 224: Does "Do not read through soft links..." refer to `self`? `local_src` is already realpath'ed

andy-slac (approved):
Looks good

python/lsst/daf/butler/core/_butlerUri/file.py
timj: Good point. I don't think it harms anything to consistently use `os.lstat` here as well but I can fix the comment.

PR #578, opened 2021-09-28
DM-31941: Allow datastore.trash to remove unknown files and ingest to overwrite

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -902,8 +902,13 @@ class FileDatastore(GenericBaseDatastore):
                 checksum = self.computeChecksum(srcUri) if self.useChecksum else None
                 have_sized = True
 
-            # transfer the resource to the destination
-            tgtLocation.uri.transfer_from(srcUri, transfer=transfer, transaction=self._transaction)
+            # Transfer the resource to the destination.
+            # Allow overwrite of an existing file. This matches the behavior
+            # of datastore.put() in that it trusts that registry would not
+            # be asking to overwrite unless registry thought that the
+            # overwrite was allowed.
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1897,6 +1902,54 @@ def trash(self, ref: Union[DatasetRef, Iterable[DatasetRef]], ignore_errors: boo
         # the cache will simply be refilled.
         self.cacheManager.remove_from_cache(ref)
 
+        # If we are in trust mode there will be nothing to move to
+        # the trash table and we will have to try to delete the file
+        # immediately.
+        if self.trustGetRequest:
+            # Try to keep the logic below for a single file trash.
+            if isinstance(ref, DatasetRef):
+                refs = set([ref])
```
andy-slac: `refs = {ref}` avoids temporary list.

timj:
@andy-slac I updated the trash commit to add some test coverage by turning on trust mode.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1897,6 +1902,54 @@ def trash(self, ref: Union[DatasetRef, Iterable[DatasetRef]], ignore_errors: boo
...
+            else:
+                # Ensure we have not used up a generator for later.
+                ref = list(ref)
+                refs = set(ref)
```
andy-slac: `set(ref)` should work with any iterable, no need to make a `list` first.

andy-slac (approved):
Looks good, couple of nitpick comments.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: I was worried about emptying a generator but that's not relevant because in this branch I recreate `ref` at the end. Will fix.

PR #579, opened 2021-09-30
DM-31956: Don't lock tables in Database.sync.

TallJimbo:
Locking tables makes us susceptible to deadlocks when within an outer transaction block, but not locking opens up the possibility that a concurrent writer might modify a just-inserted row before the check() query runs.  But we already have logic to raise RuntimeError if that happens, which is not a bad outcome for conflicting concurrent writes.

We might want to switch the exception type to DatabaseConflictError if we want a conflicting concurrent write to look the same as a preexisting conflict, but I think knowing that somebody else (probably the same user in a different process) is what's causing the problem ismore useful.

## Checklist

- [x] ran Jenkins


andy-slac (approved):
Looks reasonable.

PR #581, opened 2021-09-30
DM-31976: Be more careful with run collection creation and dataset type registration

timj:
By default now configure Butler.transfer_from to not register
dataset types. As part of this reorganization consistency
is now ensured between source butler definitions and target
butler definitions.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1884,44 +1884,28 @@ class Butler:
         # Check to see if the dataset type in the source butler has
         # the same definition in the target butler and register missing
         # ones if requested. Registration must happen outside a transaction.
-        missing_dataset_types = set()
-        inconsistent_dataset_types = {}
         newly_registered_dataset_types = set()
         for datasetType in source_dataset_types:
-            is_missing = False
-            try:
-                # If another process is registering this just after this
-                # search then this process could report failure if not
-                # allowed to register. That is fine. Rerunning will fix
-                # the problem.
-                target_dataset_type = self.registry.getDatasetType(datasetType.name)
-            except KeyError:
-                is_missing = True
-            if is_missing:
-                if register_dataset_types:
-                    self.registry.registerDatasetType(datasetType)
+            if register_dataset_types:
+                # Let this raise immediately if inconsistent. Continuing
+                # on to find additional inconsistent dataset types
+                # might result in additional unwanted dataset types being
+                # registered.
+                if self.registry.registerDatasetType(datasetType):
                     newly_registered_dataset_types.add(datasetType)
-                else:
-                    missing_dataset_types.add(datasetType)
             else:
-                if datasetType != target_dataset_type:
-                    inconsistent_dataset_types[datasetType] = target_dataset_type
-
+                # If the dataset type is missing, let it fail immediately.
+                target_dataset_type = self.registry.getDatasetType(datasetType.name)
+                if target_dataset_type != datasetType:
+                    raise ConflictingDefinitionError("Source butler dataset type differs from definition"
+                                                     f" in target butler: {datasetType} !="
+                                                     f" {target_dataset_type}")
         if newly_registered_dataset_types:
             # We may have registered some even if there were inconsistencies
             # but should let people know (or else remove them again).
             log.log(VERBOSE, "Registered the following dataset types in the target Butler: %s",
                     ", ".join(d.name for d in newly_registered_dataset_types))
-
-        if missing_dataset_types:
-            raise KeyError("The following dataset types are not known to the target registry"
-                           " and must be added: " + ", ".join(d.name for d in missing_dataset_types))
-        if inconsistent_dataset_types:
-            msg = "\n".join(f" - {src} != {tgt}" for src, tgt in inconsistent_dataset_types.items())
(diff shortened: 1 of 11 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -1869,16 +1873,56 @@ def transfer_from(self, source_butler: Butler, source_refs: Iterable[DatasetRef]
...
+        inconsistent_dataset_types = {}
+        newly_registered_dataset_types = set()
+        for datasetType in source_dataset_types:
+            is_missing = False
+            try:
+                # If another process is registering this just after this
+                # search then this process could report failure if not
+                # allowed to register. That is fine. Rerunning will fix
+                # the problem.
+                target_dataset_type = self.registry.getDatasetType(datasetType.name)
```
TallJimbo: I think I might prefer a version of this logic that called `registerDatasetType` here when `register_dataset_types=True`, rather than registering only after first identifying it as missing.  Depending on how caching works that could save a database/server query or two (though right now I think our caching is sufficiently aggressive that it doesn't matter in `SqlRegistry`).

Entirely conceivable that that is actually a lot more complex with the way you want to bundle up the error messages, though, and if you've already tried it, ignore me.

python/lsst/daf/butler/_butler.py
```diff
@@ -1924,6 +1927,33 @@ def transfer_from(self, source_butler: Butler, source_refs: Iterable[DatasetRef]
...
+                target_type = self.registry.getCollectionType(run)
+            except MissingCollectionError:
+                target_type = None
+            if target_type is None:
+                run_doc = source_butler.registry.getCollectionDocumentation(run)
+                self.registry.registerRun(run, doc=run_doc)
+                missing_run_collections.add(run)
+            elif target_type != CollectionType.RUN:
+                raise ValueError(f"Run collection {run} in source butler is not a RUN in target Butler,"
+                                 f" is {target_type.name}")
```
TallJimbo: Moving `DatasetType` registration out of the transaction is _probably_ good even after DM-31956 lands (imminent; just want to write a detailed reply to Andy's comments and then merge the code as-is), because it could involve creating a new table and in some databases (but not SQLite or PostgreSQL!) that's illegal inside a transaction.  But I think moving the collection registration outside the transaction is going to be totally obsoleted by DM-31956:

- There's no scenario where we shouldn't register the collection if it doesn't exist.
- The DM-31956 fix alone is not going to be vulnerable to these kinds of race conditions, and after it merges, the logic here will just open us back up to races.

I think I'd just drop this commit.

timj: What I've ended up doing is retaining the logic here (checking it's not a chain in the target up front seems like a good idea) and just using it to give an informational log message about which runs are likely to be created. This also stops us calling registerRun once per dataset type even if the run hasn't changed.

python/lsst/daf/butler/_butler.py
timj: Using `registerDatasetType` directly (as was done before) means that we have no way of knowing whether it was newly-created or not. I don't suppose that really matters but it would be nice to let people know. It would help if `registerDatasetType` returned a boolean indicating whether it actually registered anything.

timj: Now that I look at it again, I think that if I just used `registerDatasetType` then that code would only trigger if we are registering dataset types. This would mean that the first we might know about a dataset type not being registered or being inconsistent might be half an hour later when we try to import some data with that dataset type and then rollback the entire transaction. Don't we want to have a hint up front?

TallJimbo: `registerDatasetType` does return a `bool` already.

> This would mean that the first we might know about a dataset type not being registered or being inconsistent might be half an hour later when we try to import some data with that dataset type and then rollback the entire transaction. Don't we want to have a hint up front?

I was thinking that even a half hour earlier is too late, at least for execution butler - we need to get that warning when we make the QG, as per DM-30845.  But yeah, more generally, having an earlier check does seem like a good idea.  How about doing this up front; this still minimizes DB calls and maximizes reuse of checks in daf_butler.
```
if register_dataset_types:
    if butler.registry.registerDatasetType(datasetType):  # let this raise if inconsistent
        log.verbose(...)
else:
    if butler.registry.getDatasetType(name) != datasetType):
        raise ConflictingDefinitionError(...)  # what registerDatasetTypes raises
```
?

Makes me wonder if we should put the else block into an `Registry.expectDatasetType` or an `expect=True` option to `registerDatasetType`, as I bet this isn't the only place that wants this logic.

python/lsst/daf/butler/_butler.py
TallJimbo: Ok, but please pare down the comment, then - this is now just purely informative, as (since DM-31965 has merged) there's nothing to do with race conditions or collection registration not being safe in transactions in play anymore.

TallJimbo (approved):

python/lsst/daf/butler/_butler.py
timj: I reworked the registerRun stuff to use the new boolean return value to decide whether to report a message.

python/lsst/daf/butler/_butler.py
timj: I rewrote as described.

PR #582, opened 2021-10-01
DM-31722: Clean up utility code usage

timj:
Requires lsst/utils#100

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/dimensions/_skypix.py
+++ b/python/lsst/daf/butler/core/dimensions/_skypix.py
@@ -229,12 +229,15 @@ class SkyPixConstructionVisitor(DimensionConstructionVisitor):
 
     def visit(self, builder: DimensionConstructionBuilder) -> None:
         # Docstring inherited from DimensionConstructionVisitor.
-        pixelization_class = doImportType(self._pixelizationClassName)
-        if not hasattr(pixelization_class, "MAX_LEVEL"):
-            raise TypeError(f"Skypix pixelization class {self._pixelizationClassName} does"
-                            " not have MAX_LEVEL")
-        PixelizationClass = pixelization_class
-        maxLevel = self._maxLevel if self._maxLevel is not None else PixelizationClass.MAX_LEVEL
+        PixelizationClass = doImportType(self._pixelizationClassName)
+        assert issubclass(PixelizationClass, Pixelization)
+        if self._maxLevel is not None:
+            maxLevel = self._maxLevel
+        else:
+            maxLevel = getattr(PixelizationClass, "MAX_LEVEL", None)
+            if maxLevel is None:
+                raise TypeError(f"Skypix pixelization class {self._pixelizationClassName} does"
+                                " not have MAX_LEVEL but no max level has been set explicitly.")
         system = SkyPixSystem(
             self.name,
             maxLevel=maxLevel,
--- a/python/lsst/daf/butler/core/dimensions/_skypix.py
+++ b/python/lsst/daf/butler/core/dimensions/_skypix.py
@@ -39,7 +39,7 @@ from typing import (
 import sqlalchemy
 
 from lsst.sphgeom import Pixelization
-from lsst.utils import doImport
+from lsst.utils import doImportType
 from .. import ddl
 from .._topology import TopologicalFamily, TopologicalRelationshipEndpoint, TopologicalSpace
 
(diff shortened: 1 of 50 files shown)


python/lsst/daf/butler/core/dimensions/_skypix.py
```diff
@@ -229,7 +229,11 @@ def hasDependenciesIn(self, others: AbstractSet[str]) -> bool:
 
     def visit(self, builder: DimensionConstructionBuilder) -> None:
         # Docstring inherited from DimensionConstructionVisitor.
-        PixelizationClass = doImport(self._pixelizationClassName)
+        pixelization_class = doImportType(self._pixelizationClassName)
+        if not hasattr(pixelization_class, "MAX_LEVEL"):
```
TallJimbo: I don't think we want this check (at least not here), as it makes it impossible to use a `PixelizationClass` that lacks `MAX_LEVEL` even if you've set the `max_level` dimension config explicitly.

python/lsst/daf/butler/core/logging.py
```diff
@@ -416,10 +412,17 @@ def from_raw(cls, serialized: Union[str, bytes]) -> ButlerLogRecords:
             return cls.parse_raw(serialized)
 
         # Filter out blank lines -- mypy is confused by the newline
-        # argument to split().
-        newline = "\n" if isinstance(serialized, str) else b"\n"
-        records = [ButlerLogRecord.parse_raw(line) for line in isplit(serialized, newline)  # type: ignore
-                   if line]
+        # argument to isplit() [which can't have two different types
+        # simultaneously] so we have to duplicate some logic.
```
TallJimbo, line 416: I'm curious what the mypy error was with the original code.

TallJimbo (approved):
As an aside: it'd be really nice to be able to click a button on all those CodeCov annotations and say, "don't worry about coverage for this line ever again".  I don't suppose you know of anything like that?

python/lsst/daf/butler/core/dimensions/_skypix.py
timj: Okay. I will have to rearrange the logic here so that it doesn't use that one-line if statement and only does the mypy check if it's actually going to call MAX_LEVEL.

python/lsst/daf/butler/core/logging.py
timj: It was complaining that Union[str, bytes] for the separator can't work for a `str` parameter that is being split because the `T` type can't then work out what the return type is. With the rewrite it knows that serialized is the same type as newline so can know what isplit returns. Fundamentally if a variable at any point has a Union type it can't be converted to a non-Union type without being assigned to a new variable.

timj:
The code coverage reports only turn up for new code on this pull request. If I merge this PR and no-one else ever touches that code, codecov will never report any coverage problem. I added new code with an if clause that is not triggered by any of the tests and so that counts against this PR as being a drop in coverage.  Codecov is complaining about this PR because there was a net decrease in code coverage because I added all these checks for mypy. I can ignore them and merge or I can create a few bad config files and try to create butlers from them.

python/lsst/daf/butler/core/logging.py
TallJimbo: > Fundamentally if a variable at any point has a Union type it can't be converted to a non-Union type without being assigned to a new variable.

I've definitely seen mypy "collapse" a union down to the remaining type(s) when one of them is rejected in other contexts  (e.g. if you `assert x is not None`, it can tell that `x` is `T` in `Optional[T`]).  But I guess there's not much value in trying to figure out why it couldn't do that here.

TallJimbo:
> I can ignore them and merge or I can create a few bad config files and try to create butlers from them.

I am all for ignoring them here.  I was just hoping we could ignore them forever (e.g. keep them from coming back again if we, say, `black`-formatted everything), but it sounds like it doesn't matter much anyway.

timj:
If we black-formatted everything then codecov would absolutely lose its mind because of the 16% of the codebase that has no code coverage at all. We'd have to take that on the chin. I don't think it's unreasonable for codecov to warn you when you've edited a line but don't have any test code to show that the edit made a difference.

python/lsst/daf/butler/core/logging.py
timj: I think adding 4 asserts (two in each if branch) might do it but I'm not sure if that's much of an improvement.

python/lsst/daf/butler/core/dimensions/_skypix.py
timj: I've reorganized it. Oddly mypy is not complaining that I have not asserted that this is a `lsst.sphgeom.Pixelization` when it's put into the `SkyPixSystem`.

python/lsst/daf/butler/core/logging.py
TallJimbo: Agreed; fine as-is.

PR #583, opened 2021-10-05
DM-31287: Add protection to datasets manager import method

andy-slac:
Dataset storage manager now has an additional protection against
inconsistent dataset definitions when importing datasets that use UUID
for dataset ID.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
+++ b/python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
@@ -4,7 +4,6 @@ __all__ = ("ByDimensionsDatasetRecordStorage",)
 
 from typing import (
     Any,
-    Callable,
     Dict,
     Iterable,
     Iterator,
@@ -27,11 +26,13 @@ from lsst.daf.butler import (
     DatasetType,
     SimpleQuery,
     Timespan,
+    ddl
 )
 from lsst.daf.butler.registry import ConflictingDefinitionError, UnsupportedIdGeneratorError
 from lsst.daf.butler.registry.interfaces import DatasetRecordStorage, DatasetIdGenEnum
 
 from ...summaries import GovernorDimensionRestriction
+from .tables import makeTagTableSpec
 
 if TYPE_CHECKING:
     from ...interfaces import CollectionManager, CollectionRecord, Database, RunRecord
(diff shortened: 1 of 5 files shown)


python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
```diff
@@ -662,6 +632,181 @@ def _insert(self, run: RunRecord, dataIdList: List[DataCoordinate],
...
+            protoTagsRow = {
+                "dataset_type_id": self._dataset_type_id,
+                collFkName: run.key,
+            }
+            tmpRows = [dict(protoTagsRow, dataset_id=dataset_id, **dataId.byName())
+                       for dataset_id, dataId in dataIds.items()]
+
+            with self._db.transaction():
+
+                # store all incoming dat in a temporary table
```
TallJimbo:
```suggestion
                # store all incoming data in a temporary table
```

TallJimbo (approved):
Looks good!  I hurt my brain thinking about two potential concerns, and I think my conclusions are that no changes are needed:

*Could we get away with fewer queries in `_validateImport`?*  I _think_ so, maybe, but if so it'd be a lot harder to read and more fragile, depending on the details of various foreign keys, so not worth doing since we don't seem to have a performance problem here.

*Is this subject to race conditions with READ COMMITTED isolation?*  Only in the sense that concurrent writes that involve identical datasets would cause constraint violations in the actual `INSERT`, instead of being ignored as they would be if not concurrent.  Since that should be rare and doesn't result in any broken invariants, I think we're good.

Does that make sense to you?

andy-slac:
> Could we get away with fewer queries in _validateImport?
There are now one query on dataset table and two queries on tags table there. They should be optimal in a sense that they should be using PK or unique indices, so they should avoid full table scan. If we trust our data completely, we can avoid some checks, e.g. if we believe that UUID4 never collide and there are no mistakes on user side. I prefer to be more paranoid for now until we learn more about how reliable our data is. There is certainly a concern with ever-growing size of these tables which will make queries slower over time, I do not know yet what to do about that.

> Is this subject to race conditions with READ COMMITTED isolation? 

Concurrent inserts are indeed a potential issue. I do not think we can solve it with locking (except with the whole table lock which we don't want), I believe Postgres does not know how to lock non-existing rows. I think getting an error on INSERT in this situation and re-trying the whole thing should be acceptable, the only question is that it may need human intervention until we teach the code to recognize this sort of failure.

(I rebased this against current master and restarted Jenkins, will merge when it finishes)

PR #586, opened 2021-10-20
DM-31769 Remove Placeholder from DatasetType

natelust:
Previously a placeholder was created to aid in graph creation
involving component dataset types. Graph creation code was modified
such that this no longer needs to be an attribute of DatasetType.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/datasets/type.py
+++ b/python/lsst/daf/butler/core/datasets/type.py
@@ -100,10 +100,7 @@ class DatasetType:
     parentStorageClass : `StorageClass` or `str`, optional
         Instance of a `StorageClass` or name of `StorageClass` that defines
         how the composite parent is persisted.  Must be `None` if this
-        is not a component. Mandatory if it is a component but can be the
-        special temporary placeholder
-        (`DatasetType.PlaceholderParentStorageClass`) to allow
-        construction with an intent to finalize later.
+        is not a component.
     universe : `DimensionUniverse`, optional
         Set of all known dimensions, used to normalize ``dimensions`` if it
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/core/datasets/type.py
```diff
@@ -302,33 +291,28 @@ def isCalibration(self) -> bool:
...
-    def finalizeParentStorageClass(self, newParent: StorageClass) -> None:
-        """Finalize the parent storage class definition.
+    def replaceParentStorageClass(self, newParent: StorageClass) -> None:
+        """Replace the parent storage class definition.
 
-        Replaces the current placeholder parent storage class with
-        the real parent.
+        Replaces the current parent storage class with a new parent, this
+        could cause failures when interacting with the registry. Only use
+        this method if you know what you are doing.
```
TallJimbo: This really scares me; `DatasetType` is otherwise immutable.  I'm guessing it's probably okay because this is only used on refs produced by `GraphBuilder` before they are ever released to the outside world, but there are still a lot of moving parts there that could get confused if something changes under them.

If we can't do this without breaking immutability, I think I might be happier if we just let `GraphBuilder` poke around at private `DatasetType` attributes directly so it's clear it's being evil, instead of sanitizing that evil as a method (even one with a prominent warning).

timj: I was also wondering why we need this method at all. It was specifically designed to fail if there wasn't a placeholder for the parent storage class. It seems like a regression to now allow anyone to change the parent storage class from anything to anything.

PR #587, opened 2021-10-20
DM-29955: Add ExposureInfo id getter

kfindeisen:
This PR adds an `id` component to `ExposureInfo`, in keeping with the changes made on lsst/afw#613 and lsst/obs_base#394.


parejkoj (approved):
Our first primitive storage type?

PR #589, opened 2021-10-26
DM-32290: Re-order table creation in dataset manager

andy-slac:
Tags and calib tables are now registered after their creation, this
avoids inconsistency if creation fails. Also added checks for table
existence when tables need to be read.

## Checklist

- [X] ran Jenkins


TallJimbo (approved):
Looks great; sorry it took me so long to get to the review.

PR #590, opened 2021-11-02
DM-32407: Fix YAML serialization of empty Timespans

andy-slac:
Add special YAML representer and constructor for  the _SpecialTimespanBound
class. Extend unit test with a trivial case for an empty Timespan.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/timespan.py
+++ b/python/lsst/daf/butler/core/timespan.py
@@ -47,6 +47,7 @@ import astropy.time
 import astropy.utils.exceptions
 import sqlalchemy
 import warnings
+import yaml
 
 # As of astropy 4.2, the erfa interface is shipped independently and
 # ErfaWarning is no longer an AstropyWarning
@@ -198,6 +199,9 @@ class Timespan:
     __slots__ = ("_nsec", "_cached_begin", "_cached_end")
 
     EMPTY: ClassVar[_SpecialTimespanBound] = _SpecialTimespanBound.EMPTY
+
+    # YAML tag name for Timespan
+    yaml_tag = "!lsst.daf.butler.Timespan"
 
     @classmethod
     def makeEmpty(cls) -> Timespan:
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/core/timespan.py
```diff
@@ -78,6 +79,53 @@ class _SpecialTimespanBound(enum.Enum):
...
+        value : `_SpecialTimespanBound`
+            _SpecialTimespanBound instance, can be ``None``.
+        """
+        if node.value is not None:
+            return _SpecialTimespanBound[node.value]
+        return None
+
+
+# YAML tag name for _SpecialTimespanBound
+_SpecialTimespanBound_yaml_tag = "!lsst.daf.butler._SpecialTimespanBound"
```
timj: I do wonder if we want to have this tag burned into the YAML rather than the notion of EMPTY since the docstring for this class says we should always access it as `TimeSpan.EMPTY`.  Maybe @TallJimbo has thoughts.

timj (approved):
Looks okay. One minor thought on tag naming.

python/lsst/daf/butler/core/timespan.py
TallJimbo: If we are allowed to make it whatever string we want, then yes, `lsst.daf.butler.Timespan.EMPTY` would be preferable.

If it's not difficult to override the YAML serialization to this extent, it'd arguably be even better to map the entire `Timespan` to a sentinal string value like "EMPTY", since it's really a property of the entire timespan rather than just one bound, but if it's a lot harder it's not worth the effort.

Do we know what the YAML looks like with this solution?

andy-slac: The format before/after this change is on the JIRA here: https://jira.lsstcorp.org/browse/DM-32407?focusedCommentId=340215&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-340215

I was not sure that `_SpecialTimespanBound` is never going to have any other enum values for it, so I decided to keep it open for possible extensions. I was also thinking about replacing whole empty Timespan with a special YAML key, but this would mean adding more knowledge about Timespan details to YamlRepoExportBackend/YamlRepoImportBackend. Right now these two classes only know that Timespan has begin/end timestamps and it works OK for all cases so far. Making YAML representation more complicated adds more logic to those two classes. Alternatively we can define a special YAML representer for Timespan class itself to encapsulate that logic where it belongs, that adds YAML dependency to Timespan.

TallJimbo: Adding a YAML representer to `Timespan` does feel like the right solution.

andy-slac: OK, I'll try something quick today on this ticket.

timj (approved):
Looks good to me. Thanks for reworking it.

PR #591, opened 2021-11-02
DM-32058 Fail dataset type query early on non extant type

natelust:
If a dataset type is used in queryDatasetTypes call, and it is a
type that is not known to the registry, the method should raise
instead of silently continuing. This was causing down stream
consumers of this method to get confusing registry issues.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registries/remote.py
+++ b/python/lsst/daf/butler/registries/remote.py
@@ -354,9 +354,13 @@ class RemoteRegistry(Registry):
         # Docstring inherited from lsst.daf.butler.registry.Registry
         raise NotImplementedError()
 
-    def queryDatasetTypes(self, expression: Any = ..., *, components: Optional[bool] = None
+    def queryDatasetTypes(self, expression: Any = ..., *, components: Optional[bool] = None,
+                          missing: Optional[List[str]] = None,
                           ) -> Iterator[DatasetType]:
         # Docstring inherited from lsst.daf.butler.registry.Registry
+        if missing is not None:
+            raise NotImplementedError("RemoteRegistry does not support the 'missing' parameter.")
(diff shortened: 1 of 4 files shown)


TallJimbo (approved):
This fixes the serious problem that users are seeing, and we should get it in the weekly one way or another.

As-is, I think this either leaves a branch deep in the query logic that can never fire, _or_ that branch fires when `DatasetType` objects rather than string names are passed to `queryDatasets`, and then produces different behavior when the dataset type is not registered (exception vs. no results with diagnostics).  Both of those possibilities (dead code and inconsistent behavior) are undesirable, but no results and exceptions are both better than _wrong_ results, which is what we're getting on master now.

If we do decide that we want no results with diagnostics, as the behavior on master was _supposed_ to be, then after fixing that (e.g. by catching the exception raised by `queryDatasetTypes`, which I think probably _should_ raise) we'll want the original unit tests, and they'll pass without the workarounds added on this branch.  If we want an exception instead, we should just replace that unit test code with a `with self.assertRaises` check; I don't think there's a scenario where the unit test workaround is what we want.

In the interest of getting this in the weekly, without leaving too much debt if a follow-up ticket takes time, I think I'd recommend just removing the now-failing test code and the `UNIT_TEST_DATASET_TYPE` logic added to support it.  If the follow-up ticket changes the behavior to return no results instead of raising, we can get the original tests back from `git`.  But if you're trying to get more than just into the weekly, please merge as-is and we'll try to make sure the followup doesn't take too long.

TallJimbo (approved):
@natelust and I pair-coded what we believe is a better solution to this (including better at passing downstream tests).  We approve each other's contributions.

PR #595, opened 2021-11-05
DM-32305: Some speed ups in Config

timj:
Some minor optimizations but nothing that makes a huge difference.

## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/core/config.py
```diff
@@ -70,6 +70,36 @@ def _doUpdate(d, u):
...
+    nextVal = None
+    isThere = False
+    if d is None:
+        # We have gone past the end of the hierarchy
+        pass
+    elif not must_be_dict and isinstance(d, collections.abc.Sequence):
+        # Check sequence first because for lists
+        # __contains__ checks whether value is found in list
+        # not whether the index exists in list. When we traverse
+        # the hierarchy we are interested in the index.
```
MichelleGower, line 84: I found "Check sequence first" confusing.  I think you mean checking if k is an index first before checking __contains__.

python/lsst/daf/butler/core/config.py
```diff
@@ -70,6 +70,36 @@ def _doUpdate(d, u):
...
+        try:
+            nextVal = d[int(k)]
+            isThere = True
+        except IndexError:
+            pass
```
MichelleGower, line 89: I'm not familiar with all of the Config code.  With that in mind, can d be a list of numbers (d = [10,20,30])?  If yes (list is an instance of collections.abc.Sequence), and if k is 30, then it will get an IndexError and isThere will be False, but based on the ValueError line below I think we're wanting isThere to be True.

python/lsst/daf/butler/core/config.py
```diff
@@ -219,7 +249,7 @@ def __init__(self, other=None):
         if isinstance(other, Config):
             self._data = copy.deepcopy(other._data)
             self.configFile = other.configFile
-        elif isinstance(other, collections.abc.Mapping):
+        elif isinstance(other, (dict, collections.abc.Mapping)):
```
MichelleGower, line 252: Maybe want to copy the efficiency comment from other code (line 618) here too

MichelleGower (approved):
A couple comments about comments and one code question.  Changes approved for merging.

python/lsst/daf/butler/core/config.py
timj: No. It's something like `config[".a.2.b"]` and `a` refers to a sequence, `2` to index 2 and then `b` is a key in the dict inside that sequence.  It's features like this that are great for one off lookups in a config but a disaster when doing many of them because you get the isinstance check for sequence every time (except now I've bypassed it for a non-hierarchical look up since it's impossible for a Config top level to be a sequence.

PR #596, opened 2021-11-05
DM-32467: Allow a Path object to be used in Butler constructor

timj:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/tests/test_uri.py
+++ b/tests/test_uri.py
@@ -25,6 +25,7 @@ import shutil
 import unittest
 import urllib.parse
 import responses
+import pathlib
 
 try:
     import boto3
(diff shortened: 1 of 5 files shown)


tests/test_uri.py
```diff
@@ -63,6 +64,10 @@ def testFile(self):
         self.assertFalse(uri.exists(), f"{uri} should not exist")
         self.assertEqual(uri.ospath, file)
 
+        path = pathlib.Path(file)
+        uri = ButlerURI(path)
+        self.assertEqual(uri.ospath, str(path))
```
mfisherlevine: Not sure what the creation of the `BtlerURI`, but just wanted to check that this is still OK with/without trailing slashes. If anything the test is more fragile, so it's clearly fine, but just thought I'd flag that potential difference, depending on the value of path in future.

mfisherlevine (approved):
Looks great.

tests/test_uri.py
timj: ButlerURI is what is used as soon as anything arrives inside butler. It's the class that handles the abstraction of local file vs remote object store.

timj: I'm not sure I understand what you mean about fragility. I do see that I should be comparing `uri.ospath` to `file` so I will fix that.

mfisherlevine: It's fine, I just didn't know exactly how equality was being done internally, so wanted to make sure this was trailing-slash safe (though I'd have been surprised if it wasn't). All good.

timj: `ButlerURI` is very careful about trailing slash handling.

PR #597, opened 2021-11-07
DM-32459: Perform follow-up queries to improve no-result diagnostics.

TallJimbo:
## Checklist

- [x] ran Jenkins
- [x] added a release note for user-visible changes to `doc/changes` (in `pipe_base`).

### DIFF
--- a/python/lsst/daf/butler/registry/queries/_query.py
+++ b/python/lsst/daf/butler/registry/queries/_query.py
@@ -281,6 +281,7 @@ class Query(ABC):
         self,
         db: Database, *,
         region: Optional[Region] = None,
+        followup: bool = True,
     ) -> Iterator[str]:
         """Return human-readable messages that may help explain why the query
         yields no results.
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/registry/queries/_query.py
```diff
@@ -322,6 +324,20 @@ def explain_no_results(
                 f"{self._filtered_by_join} result rows were filtered out because "
                 "one or more regions did not overlap."
             )
+        if not followup or self._filtered_by_join or self._filtered_by_where:
```
kherner: Personally I prefer if (not followup) or self._filtered_by_join or self._filtered_by_where: simply because it makes the meaning immediately clear for future readers who may not remember the order such statements are evaluated (i.e. to not get confused with not (x or y or z). Up to you though since the code works properly as-is.

kherner (approved):

PR #601, opened 2021-11-10
DM-32403: Add support for order_by/limit to query results

andy-slac:
SQL registry method `queryDimensionRecords` now returns special iterable
object instead of plain iterator. Iterables returned from `queryDataIds`
and `queryDimensionRecords` have new methods `order_by()` and `limit()`. Same
two methods were added to `Query` classes. Extended unit tests for query
methods to check this new functionality.

Note that remote registry does not support this functionality (yet)

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/timespan.py
+++ b/python/lsst/daf/butler/core/timespan.py
@@ -614,7 +614,7 @@ class TimespanDatabaseRepresentation(TopologicalExtentDatabaseRepresentation[Tim
 
     Compound: ClassVar[Type[TimespanDatabaseRepresentation]]
     """A concrete subclass of `TimespanDatabaseRepresentation` that simply
-    uses two separate fields for the begin (inclusive) and end (excusive)
+    uses two separate fields for the begin (inclusive) and end (exclusive)
     endpoints.
 
     This implementation should be compatible with any SQL database, and should
--- a/python/lsst/daf/butler/core/timespan.py
+++ b/python/lsst/daf/butler/core/timespan.py
@@ -758,6 +758,15 @@ class TimespanDatabaseRepresentation(TopologicalExtentDatabaseRepresentation[Tim
         -------
         lower : `sqlalchemy.sql.ColumnElement`
             A SQLAlchemy expression for a lower bound.
+
+        Notes
+        -----
+        If database holds ``NULL`` for a timespan then the returned expression
+        should evaluate to 0. Main purpose of this and `upper` method is to use
+        them in generating SQL, in particular ORDER BY clause, to guarantee a
+        predictable ordering. It may potentially be used for transforming
+        boolean user expressions into SQL, but it will likely require extra
+        attention to ordering issues.
         """
         raise NotImplementedError()
 
@@ -770,6 +779,11 @@ class TimespanDatabaseRepresentation(TopologicalExtentDatabaseRepresentation[Tim
         -------
         upper : `sqlalchemy.sql.ColumnElement`
             A SQLAlchemy expression for an upper bound.
+
+        Notes
+        -----
+        If database holds ``NULL`` for a timespan then the returned expression
+        should evaluate to 0. Also see notes for `lower` method.
         """
         raise NotImplementedError()
 
@@ -945,11 +959,11 @@ class _CompoundTimespanDatabaseRepresentation(TimespanDatabaseRepresentation):
 
     def lower(self) -> sqlalchemy.sql.ColumnElement:
         # Docstring inherited.
-        return self._nsec[0]
+        return sqlalchemy.sql.functions.coalesce(self._nsec[0], sqlalchemy.sql.literal(0))
 
     def upper(self) -> sqlalchemy.sql.ColumnElement:
         # Docstring inherited.
-        return self._nsec[1]
+        return sqlalchemy.sql.functions.coalesce(self._nsec[1], sqlalchemy.sql.literal(0))
 
     def flatten(self, name: Optional[str] = None) -> Iterator[sqlalchemy.sql.ColumnElement]:
         # Docstring inherited.
--- a/python/lsst/daf/butler/core/timespan.py
+++ b/python/lsst/daf/butler/core/timespan.py
@@ -749,6 +749,30 @@ class TimespanDatabaseRepresentation(TopologicalExtentDatabaseRepresentation[Tim
         """
         raise NotImplementedError()
 
+    @abstractmethod
+    def lower(self: _S) -> sqlalchemy.sql.ColumnElement:
+        """Return a SQLAlchemy expression representing a lower bound of a
+        timespan.
+
+        Returns
+        -------
+        lower : `sqlalchemy.sql.ColumnElement`
+            A SQLAlchemy expression for a lower bound.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def upper(self: _S) -> sqlalchemy.sql.ColumnElement:
+        """Return a SQLAlchemy expression representing an upper bound of a
+        timespan.
+
+        Returns
+        -------
+        upper : `sqlalchemy.sql.ColumnElement`
+            A SQLAlchemy expression for an upper bound.
+        """
+        raise NotImplementedError()
+
 
 class _CompoundTimespanDatabaseRepresentation(TimespanDatabaseRepresentation):
     """Representation of a time span as two separate fields.
@@ -918,6 +942,14 @@ class _CompoundTimespanDatabaseRepresentation(TimespanDatabaseRepresentation):
             return sqlalchemy.sql.and_(self._nsec[0] <= other, self._nsec[1] > other)
         else:
             return sqlalchemy.sql.and_(self._nsec[0] <= other._nsec[0], self._nsec[1] >= other._nsec[1])
+
+    def lower(self) -> sqlalchemy.sql.ColumnElement:
+        # Docstring inherited.
+        return self._nsec[0]
+
+    def upper(self) -> sqlalchemy.sql.ColumnElement:
+        # Docstring inherited.
+        return self._nsec[1]
 
     def flatten(self, name: Optional[str] = None) -> Iterator[sqlalchemy.sql.ColumnElement]:
         # Docstring inherited.
--- a/python/lsst/daf/butler/datastores/chainedDatastore.py
+++ b/python/lsst/daf/butler/datastores/chainedDatastore.py
@@ -581,7 +581,7 @@ class ChainedDatastore(Datastore):
         primary, components = self.getURIs(ref, predict)
         if primary is None or components:
             raise RuntimeError(f"Dataset ({ref}) includes distinct URIs for components. "
-                               "Use Dataastore.getURIs() instead.")
+                               "Use Datastore.getURIs() instead.")
         return primary
 
     def retrieveArtifacts(self, refs: Iterable[DatasetRef],
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -843,7 +843,7 @@ class FileDatastore(GenericBaseDatastore):
         -------
         info : `StoredFileInfo`
             Internal datastore record for this file.  This will be inserted by
-            the caller; the `_extractIngestInfo` is only resposible for
+            the caller; the `_extractIngestInfo` is only responsible for
             creating and populating the struct.
 
         Raises
@@ -1016,7 +1016,7 @@ class FileDatastore(GenericBaseDatastore):
         Returns
         -------
         info : `StoredFileInfo`
-            Information describin the artifact written to the datastore.
+            Information describing the artifact written to the datastore.
         """
         location, formatter = self._prepare_for_put(inMemoryDataset, ref)
         uri = location.uri
@@ -1588,7 +1588,7 @@ class FileDatastore(GenericBaseDatastore):
         primary, components = self.getURIs(ref, predict)
         if primary is None or components:
             raise RuntimeError(f"Dataset ({ref}) includes distinct URIs for components. "
-                               "Use Dataastore.getURIs() instead.")
+                               "Use Datastore.getURIs() instead.")
         return primary
 
     def retrieveArtifacts(self, refs: Iterable[DatasetRef],
--- a/python/lsst/daf/butler/registry/queries/_query.py
+++ b/python/lsst/daf/butler/registry/queries/_query.py
@@ -24,9 +24,11 @@ __all__ = ("Query",)
 
 from abc import ABC, abstractmethod
 from contextlib import contextmanager
+import dataclasses
 import enum
 import itertools
 from typing import (
+    ContextManager,
     Dict,
     Iterable,
     Iterator,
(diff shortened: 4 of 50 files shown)


python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -581,7 +581,7 @@ def getURI(self, ref: DatasetRef, predict: bool = False) -> ButlerURI:
         primary, components = self.getURIs(ref, predict)
         if primary is None or components:
             raise RuntimeError(f"Dataset ({ref}) includes distinct URIs for components. "
-                               "Use Dataastore.getURIs() instead.")
+                               "Use DataStore.getURIs() instead.")
```
TallJimbo:
```suggestion
                               "Use Datastore.getURIs() instead.")
```

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1588,7 +1588,7 @@ def getURI(self, ref: DatasetRef, predict: bool = False) -> ButlerURI:
         primary, components = self.getURIs(ref, predict)
         if primary is None or components:
             raise RuntimeError(f"Dataset ({ref}) includes distinct URIs for components. "
-                               "Use Dataastore.getURIs() instead.")
+                               "Use DataStore.getURIs() instead.")
```
TallJimbo:
```suggestion
                               "Use Datastore.getURIs() instead.")
```

python/lsst/daf/butler/core/timespan.py
```diff
@@ -749,6 +749,30 @@ def contains(self: _S, other: Union[_S, sqlalchemy.sql.ColumnElement]) -> sqlalc
...
+    def lower(self: _S) -> sqlalchemy.sql.ColumnElement:
+        """Return a SQLAlchemy expression representing a lower bound of a
+        timespan.
+
+        Returns
+        -------
+        lower : `sqlalchemy.sql.ColumnElement`
+            A SQLAlchemy expression for a lower bound.
+        """
+        raise NotImplementedError()
```
TallJimbo, line 771: I am a little worried about the behavior of these for `NULL`/`None` timespans; I think I may have left them out originally because it was very hard to get them to behave consistently for both `TimespanDatabaseRepresentation` implementations in the presence of `NULL`/`None`.

python/lsst/daf/butler/registry/queries/_query.py
```diff
@@ -61,6 +63,222 @@
...
+            name = name[1:]
+
+        dimension: Dimension
+        field_name: Optional[str] = None
+        if "." not in name:
+            # No dot, can be either a dimension name or field name (in any of
+            # the known dimensions)
+            if name in graph.dimensions.names:
+                dimension = graph.dimensions[name]
+            elif name in ("timespan_begin", "timespan_end"):
```
TallJimbo: If this isn't matching some other precedent, I'd prefer `timespan.begin` and `timespan.end` to match the Python syntax for accessing attributes on a `DataCoordinate`.

python/lsst/daf/butler/registry/queries/_query.py
```diff
@@ -61,6 +63,222 @@
...
+                matches = [dim for dim in graph if dim.temporal]
```
TallJimbo:
```suggestion
                matches = [element for element in graph.elements if element.temporal]
```
We want this to include `DimensionElement`s that are not `Dimension`s.

python/lsst/daf/butler/registry/queries/_query.py
```diff
@@ -61,6 +63,222 @@
...
+                    raise ValueError(
+                        f"Timespan exists in more than one dimesion: {matches},"
+                        " qualify timespan with specific dimension name.")
+                else:
+                    raise ValueError(
+                        f"Cannot find any temporal dimension for '{name}'.")
+            else:
+                # Can be a metadata name or any of unique keys
+                matches = [dim for dim in graph
+                           if name in dim.metadata.names or name in dim.uniqueKeys.names]
```
TallJimbo: This should be a loop over `graph.elements` as well.

python/lsst/daf/butler/registry/queries/_query.py
```diff
@@ -61,6 +63,222 @@
...
+                if len(matches) == 1:
+                    dimension = matches[0]
+                    field_name = name
+                elif len(matches) > 1:
+                    raise ValueError(
+                        f"Metadata '{name}' exists in more than one dimesion: {matches},"
```
TallJimbo: d
```suggestion
                        f"Metadata '{name}' exists in more than one dimension element: {matches},"
```

python/lsst/daf/butler/registry/queries/_query.py
```diff
@@ -61,6 +63,222 @@
...
+                        " qualify metadata name with dimension name.")
+                else:
+                    raise ValueError(
+                        f"Metadata '{name}' cannot be found in any dimension.")
+        else:
+            # qualified name, must be a dimension and field
+            dim_name, _, field_name = name.partition(".")
+            if dim_name not in graph.dimensions.names:
```
TallJimbo: Should be a loop over `graph.elements.names`.

python/lsst/daf/butler/registry/queries/_query.py
```diff
@@ -61,6 +63,222 @@
...
+                # Primary key is optional
+                field_name = None
+            else:
+                if field_name not in dimension.metadata.names and \
+                        field_name not in dimension.alternateKeys.names:
+                    raise ValueError(f"Field '{field_name}' does not exist in '{dimension}'.")
+
+        return dimension, field_name, ascending
+
+    def _field_column(self, table: sqlalchemy.schema.Table, dimension: Dimension, field_name: str
```
TallJimbo: Should be `DimensionElement` here, not `Dimension`

python/lsst/daf/butler/registry/queries/_query.py
```diff
@@ -339,6 +562,33 @@ def explain_no_results(
...
+
+        Parameters
+        ----------
+        db : `Database`
+            Object managing the database connection.
+        args : `str`
+            Names of the columns/dimensions to use for ordering. Column name
+            can be prefixed with minus sign to use descending ordering.
+        """
+        self._order_by = _OrderByHandler(args, self.graph, db)
```
TallJimbo: I believe that aside from this method, `Query` is immutable and "modifiers" always returns a new object.

Might be expedient to mostly ignore that for now, since I don't think anything depends on that immutabilty (only those `QueryResults` objects use `Query` directly), but it would be good to `return self` and document that that's what user should use, with `self` effectively consumed, in case that's what we want to do later.

python/lsst/daf/butler/registry/queries/_query.py
```diff
@@ -61,6 +63,222 @@
...
+        order_by_columns: List[sqlalchemy.sql.ColumnElement] = []
+        for dimension, field, asc in self._order_by:
+
+            column: sqlalchemy.sql.ColumnElement
+            if field is None:
+                # dimension name, we should know it
+                column = dim_columns[dimension.name]
+            else:
+                # This is messy, but I do not know a better way of how to get
+                # access to the table.
```
TallJimbo: I'm afraid it won't work, at least not in general - a dimension column in a query _could_ come from a dataset subquery, not a dimension table at all.  For example, if someone does `queryDatasets("raw", collections="mine")`, we will not join in the `exposure` and `detector` tables, because we can tell that all of the columns we need are available from the appropriate `dataset_tags_*` table.

But that's a much deeper and more general problem; we really need to know the fields we will order by back when we first make `QuerySummary` (probably something analogous to or merged with `QueryWhereClause.columns`), in order to ensure the needed tables are joined into the query (and in the case of materialized, ensure order by columns are included in the result columns).  Then `QueryBuilder._elements` is the mapping from `DimensionElement` to column that we need here.

I'll try to gather thoughts on how best to modify things in a single post after the line-by-line review.

python/lsst/daf/butler/registry/queries/_query.py
```diff
@@ -339,6 +562,33 @@ def explain_no_results(
...
+
+        Parameters
+        ----------
+        limit : `int`
+            Upper limit on the number of returned records.
+        offset : `int` or `None`
+            If not `None` then the number of records to skip before returning
+            ``limit`` records.
+        """
+        raise NotImplementedError()
```
TallJimbo: Like `order_by`, this should return a new `Query` and not modify self (or at least `self` after modifying it).

python/lsst/daf/butler/registry/queries/_results.py
```diff
@@ -709,3 +762,152 @@ def explain_no_results(self) -> Iterator[str]:
...
+        This method can modify the iterator in place and return the same
+        instance. Normally this method is used together with `order_by`
+        method.
+        """
+        raise NotImplementedError()
+
+
+class DatabaseDimensionRecordQueryResults(DimensionRecordQueryResults):
+    """
+    """
```
TallJimbo, line 970: Missing docstring.

You're also welcome to merge this in with the ABC (or not), as far as I'm concerned.  I do think we want an ABC _somewhere_ to distinguish between query results objects from `RemoteRegistry` vs. `SqlRegistry`, but I sort of feel like the right level for that is in a helper object held by a concrete+final `QueryResults` object, not the `QueryResults` object itself.  But I'm also not at all sure either way.

TallJimbo (commented):
As I noted in one or two of the line comments, this has a pretty big limitation, in that you can only order by metadata fields if something else has happened to ensure the table that provides those fields has been joined into the query, and this is frequently not the case, especially when datasets are included in the query.  In fact, I think with the way `queryDimensionRecords` works, even metadata fields from the dimension element whose records are being queried are not available for use in `order_by`, because `order_by` acts on a subquery generated by `queryDataIds` that may not join in that table.

The quick solution to this problem is to be less ambitious: remove support for ordering by metadata fields, because ordering by dimension keys should work fine.  While I'm sure our colleagues really want to `order_by(["day_obs", "seq_num"])`, we can tell them to `order_by(["exposure"])` as a workaround and they'll get the same order with a better chance of good index usage anyway.

Full support for metadata-field ordering would require moving this logic from `Query` to `QuerySummary` and `QueryBuilder`:

- Add a mapping similar to `QueryWhereClause.columns` to `QuerySummary`, probably via an analogous nested struct that processes the user's order_by argument and stores the columns it needs.  The important thing is that `QuerySummary.mustHaveTableJoined` include any `DimensionElement` whose metadata fields are included in the `order_by` list, and something in `QuerySummary` should remember the `order_by` list itself so it can be used later.

- Move the logic in `_OrderByHandler.query_columns` to [something called by] `QueryBuilder.finish`, operating on `QueryBuilder._elements`, and then passing the `DirectQuery` constructor just the list of sqlalchemy columns to order by.

- Make `Query.order_by` return a new `Query` by calling `Query.makeBuilder` with an augmented `QuerySummary` that includes the `order_by` request, use that to make a `QueryBuilder` and use that to make a new `Query` to return.

I am not sure it is worth taking the more difficult/complete road right now, as I think most needs can be met with dimension-key ordering.  But this is tied up with the question of how to handle RemoteRegistry - at least a bit - and I'll bring that up on Jira since that's where that conversation started.

andy-slac:
> I think with the way queryDimensionRecords works, even metadata fields from the dimension element whose records are being queried are not available for use in order_by, because order_by acts on a subquery generated by queryDataIds that may not join in that table.

I see that `queryDimensionRecords` works by doing `queryDataIds` first and then selecting records that match returned DataIds, I don't think there are sub-query involved there. In my dumb implementation of `queryDimensionRecords` I order `queryDataIds` results and then re-order records based on that ordering.

andy-slac:
And I think my general naive idea was that to order something based on metadata its corresponding dimension have to be in the query already. If you want to order based on metadata from some other dimension, we'd have to include that dimension into join, but that would be a different query with potentially different constraints?

TallJimbo:
> And I think my general naive idea was that to order something based on metadata its corresponding dimension have to be in the query already. If you want to order based on metadata from some other dimension, we'd have to include that dimension into join, but that would be a different query with potentially different constraints?

I think we resolved this question at the middleware standup, but for posterity, I am specifically worried about the case where the dimension _is_ in the query, but as an optimization we did not include its table, either because a dataset with that dimension is in the query, or because some dependent dimension is in the query.  The user won't be aware of those optimizations and at present can't control them, so it would be surprising for their order_by to stop working when they are adding new constraints to the query but aren't changing the dimensions.

python/lsst/daf/butler/core/timespan.py
andy-slac: If the stored value is NULL then NULL should be returned, the issue is of course that NULL can be handled differently (or unexpectedly) in ORDER BY. If we could agree on something specific that should be returned for NULL then we can wrap it into `COALESCE` to return a value instead of NULL. Should NULL be treated as EMPTY? Or maybe they should be ordered earlier than any non-NULL timespans, so that they don't appear in a query "give me the latest timestamp".

python/lsst/daf/butler/registry/queries/_query.py
andy-slac: Is there a simpler way to clone Query than doing `makeBuilder`? It feels a bit heavy-weight for changing one simple attribute.

python/lsst/daf/butler/registry/queries/_query.py
andy-slac: I thought I would simplify life for myself if I only use Dimensions, Elements don't have primary/alternative keys and elements that we have defined today don't have any metadata. Is it possible we'll have Elements with metadata?

andy-slac:
@TallJimbo, I'm now trying to do the right thing and follow your recipe:

> Make Query.order_by return a new Query by calling Query.makeBuilder with an augmented QuerySummary that includes the order_by request, use that to make a QueryBuilder and use that to make a new Query to return. 

but without much success. `Query.makeBuilder` does make a builder, but that builder does not have `_elements` populated (naturally). To populate it I'd have to join all dimensions again calling `finish(True)` but I don't think this is the correct thing to do here as it will join them with the already existing query (sort of creating self-join). Do I need to re-implement `makeBuilder` to do something different (my guess it's not really possible to fill `_elements` from existing query).

TallJimbo:
Ah, very good point; I was imaging that `makeBuilder` was much smarter than it actually is.  And I don't think that can really be fixed without changing much more than just `makeBuilder`; the problem is that a `Query` just doesn't carry around enough information to reconstruct it.  At this point I think it'd be better to go back to your original approach and just document the limitations.

python/lsst/daf/butler/registry/queries/_query.py
andy-slac: I re-worked the code to handle DimensionElements correctly.

andy-slac:
@TallJimbo, I think I managed to resolve the problem by delaying the construction of a `Query` until the latest possible moment when `order_by` and `limit` are already known. It needed a sort of another level of indirection by passing a factory method instead of `Query` object to a Result constructor. Could you re-review ~~35f3ab0~~ 20ea772  to see if it's reasonable or can be improved? 

For NULL timespans I think one possible option is to assume begin/end as 0 for them so that they sort before any actual timespan, that should make things work as expected for people looking for the most recent values. For advanced handling we could filter NULLs using user expressions (not sure that it is supported right now though).

TallJimbo (approved):
:+1: New version looks good; the extra indirection is pragmatic way to get the functionality in without a major rewrite of the queries subpackage, even though it does add some complexity that makes doing that rewrite feel a bit more important to me (to be clear, most of the problems were preexisting by far).

As for timespans, I'm pretty happy with any ordering.  My main concern was actually with the endpoint SQLALchemy accessors on `TimespanDatabaseRepresentation` themselves, and in particular whether we can define consistent ordered _comparison_ operations on those in the presence of NULL.  I think that only matters if we _also_ add support for them in the boolean expression language, so all this ticket does is make that seem a bit easier (when it's actually quite hard).  So perhaps we should just add a disclaimer to the docs or comments on those attributes saying that `Timespan` methods should always be used for comparisons in boolean expressions, and attributes should not be exposed to the expression language in the future without a lot of attention to detail.

andy-slac:
I have added this note to `lower` method (and similar to `upper`):

        Notes
        -----
        If database holds ``NULL`` for a timespan then the returned expression
        should evaluate to 0. Main purpose of this and `upper` method is to use
        them in generating SQL, in particular ORDER BY clause, to guarantee a
        predictable ordering. It may potentially be used for transforming
        boolean user expressions into SQL, but it will likely require extra
        attention to ordering issues.

PR #603, opened 2021-11-12
DM-32491: Add Butler.get_repo_uri class method for searching repository index

timj:
Pretty straightforward implementation of a repository index.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -87,6 +87,7 @@ from .core.repoRelocation import BUTLER_ROOT_TAG
 from .core.utils import transactional
 from ._deferredDatasetHandle import DeferredDatasetHandle
 from ._butlerConfig import ButlerConfig
+from ._butlerRepoIndex import ButlerRepoIndex
 from .registry import (
     Registry,
     RegistryConfig,
@@ -303,24 +304,26 @@ class Butler:
 
         Notes
         -----
-        The index of butler repositories is found by looking for a
-        configuration file at the URI pointed at by the environment
-        variable ``$BUTLER_REPOSITORY_INDEX``. The configuration file
-        is a simple dictionary lookup of the form:
-
-        .. code-block:: yaml
-
-           label1: uri1
-           label2: uri2
-
-        and can be in YAML or JSON format.
+        See `~lsst.daf.butler.ButlerRepoIndex` for details on how the
+        information is discovered.
         """
-        index_uri = os.environ.get("BUTLER_REPOSITORY_INDEX")
-        if index_uri is None:
-            raise KeyError("No repository index defined in enviroment variable BUTLER_REPOSITORY_INDEX.")
+        return ButlerRepoIndex.get_repo_uri(label)
 
-        config = Config(index_uri)
-        return config[label]
+    @classmethod
+    def get_known_repos(cls) -> Set[str]:
+        """Retrieve the list of known repository labels.
+
+        Returns
+        -------
+        repos : `set` of `str`
+            All the known labels. Can be empty if no index can be found.
+
+        Notes
+        -----
+        See `~lsst.daf.butler.ButlerRepoIndex` for details on how the
+        information is discovered.
+        """
+        return ButlerRepoIndex.get_known_repos()
 
     @staticmethod
     def makeRepo(root: str, config: Union[Config, str, None] = None,
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -281,6 +281,47 @@ class Butler:
     code.
     """
 
+    @classmethod
+    def get_repo_uri(cls, label: str) -> ButlerURI:
+        """Look up the label in a butler repository index.
+
+        Parameters
+        ----------
+        label : `str`
+            Label of the Butler repository to look up.
+
+        Returns
+        -------
+        uri : `ButlerURI`
+            URI to the Butler repository associated with the given label.
+
+        Raises
+        ------
+        KeyError
+            Raised if the label is not found in the index, or if an index
+            can not be found at all.
+
+        Notes
+        -----
+        The index of butler repositories is found by looking for a
+        configuration file at the URI pointed at by the environment
+        variable ``$BUTLER_REPOSITORY_INDEX``. The configuration file
+        is a simple dictionary lookup of the form:
+
+        .. code-block:: yaml
+
+           label1: uri1
+           label2: uri2
+
+        and can be in YAML or JSON format.
+        """
+        index_uri = os.environ.get("BUTLER_REPOSITORY_INDEX")
+        if index_uri is None:
+            raise KeyError("No repository index defined in enviroment variable BUTLER_REPOSITORY_INDEX.")
+
+        config = Config(index_uri)
+        return config[label]
+
     @staticmethod
     def makeRepo(root: str, config: Union[Config, str, None] = None,
                  dimensionConfig: Union[Config, str, None] = None, standalone: bool = False,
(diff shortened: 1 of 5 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -281,6 +281,47 @@ def __init__(self, config: Union[Config, str, None] = None, *,
...
+           label1: uri1
+           label2: uri2
+
+        and can be in YAML or JSON format.
+        """
+        index_uri = os.environ.get("BUTLER_REPOSITORY_INDEX")
+        if index_uri is None:
+            raise KeyError("No repository index defined in enviroment variable BUTLER_REPOSITORY_INDEX.")
+
+        config = Config(index_uri)
```
ktlim: Could it be worth saving this rather than reloading it on every query?  Most of the time there would be exactly one call to this function, but I could see some cases where there might be more than one.  Also, I think it might be useful to provide a method to list all the labels to enable the user to be asked to select one.

tests/test_butler.py
```diff
@@ -482,6 +482,26 @@ def testConstructor(self):
         self.assertIsNone(butler2.run)
         self.assertIs(butler.datastore, butler2.datastore)
 
+        # Test that we can use an environment variable to find this
+        # repository.
+        butler_index = Config()
+        butler_index["label"] = self.tmpConfigFile
+        butler_index.dump
```
ktlim: Um, this looks useless.

tests/test_butler.py
```diff
@@ -482,6 +482,26 @@ def testConstructor(self):
...
+            with unittest.mock.patch.dict(os.environ, {"BUTLER_REPOSITORY_INDEX": str(temp_file)}):
+                uri = Butler.get_repo_uri("label")
+                butler = Butler(uri, writeable=False)
+                self.assertIsInstance(butler, Butler)
+                with self.assertRaises(KeyError):
+                    Butler.get_repo_uri("missing")
+            with unittest.mock.patch.dict(os.environ, {"BUTLER_REPOSITORY_INDEX": "file://not_found/x.yaml"}):
+                with self.assertRaises(FileNotFoundError):
+                    Butler.get_repo_uri("label")
+            with self.assertRaises(KeyError):
```
ktlim: This `KeyError` is different from the previous one, though.

ktlim (approved):

tests/test_butler.py
timj: indeed. Left over from a code reorganization.

python/lsst/daf/butler/_butler.py
timj: `Butler.list_repo_labels()` ?

tests/test_butler.py
timj: You want me to test the contents of the exception?

ktlim: If you wanted to be absolutely safe, then yes, but I think it's OK to treat "missing-label" and "no-labels-defined" as essentially the same.

python/lsst/daf/butler/_butler.py
timj: I've reworked things such that there is now a `ButlerRepoIndex` class that does the heavy lifting and the two `Butler` methods forward directly to that class.  The main issue is whether we bother with the `Butler` class methods or just tell people they have to use `ButlerRepoIndex` directly. The point in favor of `Butler` is that they already have that class imported. The point against is that it's new methods cluttering up the class that are only tangentially related to `Butler`. @mfisherlevine do you have an opinion on:

```python
from lsst.daf.butler import Butler
uri = Butler.get_repo_uri("latiss")
butler = Butler(uri, ...)
```
vs
```python
from lsst.daf.butler import Butler, ButlerRepoIndex
uri = ButlerRepoIndex.get_repo_uri("latiss")
butler = Butler(uri, ...)
```
?

python/lsst/daf/butler/_butlerRepoIndex.py
```diff
@@ -0,0 +1,173 @@
...
+    Config,
+)
+
+
+class ButlerRepoIndex:
+    """Index of all known butler repositories.
+
+    The index of butler repositories is found by looking for a
+    configuration file at the URI pointed at by the environment
+    variable ``$BUTLER_REPOSITORY_INDEX``. The configuration file
```
ktlim: Should this have an `LSST_` prefix to help avoid collisions, even if the Butler is being reused outside the project?

python/lsst/daf/butler/_butlerRepoIndex.py
```diff
@@ -0,0 +1,173 @@
...
+            URI to the repository index.
+
+        Raises
+        ------
+        KeyError
+            Raised if the location of the index could not be determined.
+        """
+        index_uri = os.environ.get(cls.index_env_var)
+        if index_uri is None:
+            raise KeyError(f"No repository index defined in enviroment variable {cls.index_env_var}")
```
ktlim, line 111:
```suggestion
            raise KeyError(f"No repository index defined in environment variable {cls.index_env_var}")
```

python/lsst/daf/butler/_butlerRepoIndex.py
```diff
@@ -0,0 +1,173 @@
...
+
+        Returns
+        -------
+        repos : `set` of `str`
+            All the known labels. Can be empty if no index can be found.
+        """
+        try:
+            repo_index = cls._read_repository_index_from_environment()
+        except (FileNotFoundError, KeyError):
+            return set()
```
ktlim, line 138: It's not entirely clear to me that it's nicer to return an empty set than to pass along the exception.  It means that there's no available reason for knowing why the set is empty.

tests/test_butler.py
```diff
@@ -482,6 +482,35 @@ def testConstructor(self):
...
+                    self.assertEqual(uri, ButlerURI(bad_label))
+                    uri = Butler.get_repo_uri("label")
+                    butler = Butler(uri, writeable=False)
+                    self.assertIsInstance(butler, Butler)
+                    with self.assertRaises(KeyError) as cm:
+                        Butler.get_repo_uri("missing")
+                    self.assertIn("not known to", str(cm.exception))
+        with unittest.mock.patch.dict(os.environ, {"BUTLER_REPOSITORY_INDEX": "file://not_found/x.yaml"}):
+            with self.assertRaises(FileNotFoundError):
+                Butler.get_repo_uri("label")
```
ktlim, line 508: Could easily check for empty set return from `get_known_repos()` here, if you keep that.

ktlim (approved):

python/lsst/daf/butler/_butlerRepoIndex.py
timj: Okay. So it looks like butler currently recognizes the following environment variables:

* DAF_BUTLER_PLUGINS
* DAF_BUTLER_CONFIG_PATH
* DAF_BUTLER_CACHE_DIRECTORY
* DAF_BUTLER_CACHE_EXPIRATION_MODE
* DAF_BUTLER_TEST_TMP (for test code)
* LSST_WEBDAV_*. (for ButlerURI webdav implementation)

So you make a strong case for DAF_BUTLER_REPOSITORY_INDEX...

python/lsst/daf/butler/_butlerRepoIndex.py
timj: I was wondering about that. On the one hand it's nice not to have to put it in a try block and it seems like no results is going to be a strong indicator that something odd is happening. On the other hand you can't then use that knowledge to work out what the actual problem is unless I add logging (or you do `get_repo_uri()` and look at that exception). Maybe some debug log messages wouldn't hurt anyhow.  If you really want it to raise I'll change it.

ktlim: It's a shame you can't return an empty set _and_ an error string :(.  And adding `get_known_repos_error()` doesn't seem reasonable.  I guess just leave it.

PR #604, opened 2021-11-22
DM-32687: add a butler remove-collections command

n8pease:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-32687.feature.rst
+++ b/doc/changes/DM-32687.feature.rst
@@ -1,0 +1,1 @@
+Add a butler command line command called ``butler remove-collections`` that can remove non-RUN collections.
--- a/python/lsst/daf/butler/cli/cmd/__init__.py
+++ b/python/lsst/daf/butler/cli/cmd/__init__.py
@@ -37,6 +37,7 @@ __all__ = ("associate",
            "query_dimension_records",
            "register_dataset_type",
            "retrieve_artifacts",
+           "remove_collections",
            "remove_dataset_type",
            "transfer_datasets",
 )
@@ -63,3 +64,4 @@ from .commands import (associate,
                        remove_dataset_type,
                        transfer_datasets,
 )
+from ._remove_collections import remove_collections
--- a/python/lsst/daf/butler/cli/cmd/_remove_collections.py
+++ b/python/lsst/daf/butler/cli/cmd/_remove_collections.py
@@ -1,0 +1,90 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+import click
+
+from ..utils import (
+    ButlerCommand,
+)
+from ..opt import (
+    collection_argument,
+    confirm_option,
+    options_file_option,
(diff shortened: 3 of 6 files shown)


n8pease:
ooh Codecov is a fancy new toy since I was here last. I'll fix the misses tomorrow am.

n8pease:
both the +2 codecov misses are in unit test files. One is the `if __name__ == "__main__"` statement, I don't think I can do anything about that?

timj:
Correct, the unittest main one can't ever be hit. You'll only get that report for a new test file.

python/lsst/daf/butler/script/removeCollection.py
```diff
@@ -0,0 +1,134 @@
...
+    -------
+    collectionInfo : `CollectionInfo`
+        Contains tables with run and non-run collection info.
+    """
+    butler = Butler(repo)
+    try:
+        names = list(
+            butler.registry.queryCollections(
+                collectionTypes=frozenset((
+                    CollectionType.RUN,
```
timj: See my comment below, but removing this RUN and not reporting RUNs would make things simpler wouldn't it?

python/lsst/daf/butler/script/removeCollection.py
```diff
@@ -0,0 +1,134 @@
...
+                includeChains=True,
+            )
+        )
+    except MissingCollectionError:
+        names = list()
+    collections = Table(names=("Collection", "Collection Type"), dtype=(str, str))
+    runCollections = Table(names=("Collection",), dtype=(str,))
+    for name in names:
+        collectionType = butler.registry.getCollectionType(name).name
+        if collectionType == "RUN":
```
timj: A lot of work on this ticket seems to be handling RUN collections but do we need to report them at all? Currently we get:

```
$ butler remove-collection ../ci_hsc_gen3/DATA HSC/runs/*

The following collections will be removed:
      Collection      Collection Type
--------------------- ---------------
HSC/runs/ci_hsc       CHAINED        
HSC/runs/ci_hsc_fakes CHAINED        
HSC/runs/ci_hsc_faro  CHAINED        

The following run collections were found but can NOT be removed by this command:
              Collection              
--------------------------------------
HSC/runs/ci_hsc/20211111T192427Z      
HSC/runs/ci_hsc_fakes/20211111T213258Z
HSC/runs/ci_hsc_faro/20211111T223428Z 

Continue? [y/N]: n

Aborted.
```
which does show that some RUN collections matched but is all the extra code worth it? Isn't printing what it will do enough given that the help goes out of its way to say that RUN collections won't be touched.

I'm happy for @TallJimbo to tell me that reporting matching RUN collections will be useful.

python/lsst/daf/butler/script/removeCollection.py
```diff
@@ -0,0 +1,134 @@
...
+    try:
+        names = list(
+            butler.registry.queryCollections(
+                collectionTypes=frozenset((
+                    CollectionType.RUN,
+                    CollectionType.TAGGED,
+                    CollectionType.CHAINED,
+                    CollectionType.CALIBRATION,
+                )),
+                expression=(collection,),
```
timj: Can't `expression` take a string? It seems to from my testing so I don't think the tuple is needed.

timj (approved):
Looks okay. One minor code comment and then a more general comment about whether we want the added complexity of reporting on RUNs that might have matched the collections regex.

python/lsst/daf/butler/script/removeCollection.py
TallJimbo: No strong opinions from me on this either way.

timj: Maybe I should ask people.

timj: Sounds like people like it as it is with the verbosity.

n8pease:
rewording TODO from slack: https://lsstc.slack.com/archives/C01FBUGM2CV/p1638222714262300?thread_ts=1638218403.251700&cid=C01FBUGM2CV

timj:
If you like you can leave the rewording to when the remove-run command exists. Otherwise we will be referencing a command that doesn't exist yet.

timj:
There has also been a comment on slack that it might be best to make the command:

```
butler remove-collections REPO TYPE COLLECTIONs
```
so that you must declare what type of collection you want to remove. This would lead to a single command for RUN and CHAINED (and CALIBRATION/TAGGED). Worth considering.

TallJimbo:
I think that's more complicated than it appears on first glance, because there are probably going to be some options relevant for removing RUN collections that aren't relevant for removing others.

PR #605, opened 2021-11-23
DM-32682: Support TaskMetadata storage class and serialization

timj:
## Checklist

- [ ] ran Jenkins


ktlim (approved):

python/lsst/daf/butler/configs/datastores/formatters.yaml
```diff
@@ -77,3 +77,4 @@ AstropyQTable: lsst.daf.butler.formatters.astropyTable.AstropyTableFormatter
 ExtendedPsf: lsst.obs.base.formatters.fitsGeneric.FitsGenericFormatter
 HealSparseMap: lsst.pipe.tasks.healSparseMapping.HealSparseMapFormatter
 ButlerLogRecords: lsst.daf.butler.formatters.logs.ButlerLogRecordsFormatter
+TaskMetadata: lsst.daf.butler.formatters.json.JsonFormatter
```
ktlim, line 80: Wait, doesn't the change from YAML to JSON mean that older metadata cannot be read into the new class?

timj: No. The formatter is burned in on put so anything written with the YAML formatter will be read by the YAML formatter. Those are a different storage class anyhow (PropertySet -- that's the problem that makes switching hard).

PR #608, opened 2021-11-29
DM-32144: Suppress warnings for cartesian join

andy-slac:
Some queries involving user expression can generate cartesian join,
it is harmless as it should normally be constrained to a single row in
a joined table. Instead of trying to avoid cartesian joins we simply
suppress sqlalchemy warnings for one type of query.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/dimensions/table.py
+++ b/python/lsst/daf/butler/registry/dimensions/table.py
@@ -38,6 +38,7 @@ from typing import (
     Set,
     Union,
 )
+import warnings
 
 import sqlalchemy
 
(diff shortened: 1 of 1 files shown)


python/lsst/daf/butler/registry/dimensions/table.py
```diff
@@ -205,11 +206,17 @@ def fetch(self, dataIds: DataCoordinateIterable) -> Iterable[DimensionRecord]:
...
         query.join(self._table)
         dataIds.constrain(query, lambda name: self._fetchColumns[name])
-        for row in self._db.query(query.combine()):
-            values = row._asdict()
-            if self.element.temporal is not None:
-                values[TimespanDatabaseRepresentation.NAME] = TimespanReprClass.extract(values)
-            yield RecordClass(**values)
+        with warnings.catch_warnings():
+            # Some of our generated queries may contain cartesian joins, this
+            # is not a serious issue as it is properly constained, so we want
```
TallJimbo:
```suggestion
            # is not a serious issue as it is properly constrained, so we want
```

TallJimbo (approved):

PR #613, opened 2021-12-06
DM-32820: Skip ButlerURI in table of contents

timj:
Aliasing `ButlerURI` to `ResourcePath` works fine for code but breaks sphinx documentation building. Workaround this by removing the table of content entry.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/doc/lsst.daf.butler/index.rst
+++ b/doc/lsst.daf.butler/index.rst
@@ -100,6 +100,12 @@ Python API reference
 
 .. automodapi:: lsst.daf.butler
    :no-main-docstr:
+   :skip: ButlerURI
+
+.. py:class:: lsst.daf.butler.ButlerURI(uri)
+
+   ``ButlerURI`` implementation. Exists for backwards compatibility.
+   New code should use `lsst.resources.ResourcePath`.
 
 .. automodapi:: lsst.daf.butler.registry
    :no-main-docstr:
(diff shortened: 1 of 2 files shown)


doc/lsst.daf.butler/index.rst
```diff
@@ -100,6 +100,12 @@ Python API reference
 
 .. automodapi:: lsst.daf.butler
    :no-main-docstr:
+   :skip: ButlerURI
+
+.. py:class:: lsst.daf.butler.ButlerURI(uri)
```
timj, line 105: @jonathansick This does work in the sense that some words appear, but it doesn't result in something that is clickable. Am I doing something wrong? The `:skip:` fixes the pipelines.lsst.io build problem so it's not important.

jonathansick: What automodapi does, once it detects the available symbols in a namespace, is create rst pages that in turn contain `py:class` directives and so on. Then it creates an automodsumm in place of the automodapi while creates the click-able table of contents. I'm assuming that this ButlerURI documentation in `lsst.daf.butler` is only temporary, so I don't think it's worth the trouble to fake the look of automodsumm.

Also note that if there are existing links in the documentation to methods of ButlerURI, you'd technically want to add those methods to your doc for ButlerURI. You'd do this by putting `py:method` directives inside the scope of the `py:class` directive. If there aren't such links, I wouldn't worry about it.

jonathansick (approved):

doc/lsst.daf.butler/index.rst
timj: There aren't any links to methods, just the class. The docs do build fine with this change so I'll leave it. The plan is to do a global replacement of `ButlerURI` with `ResourcePath` so I'll leave it as is and merge this to solve the immediate problem of the pipelines.lsst.io build not working.

PR #614, opened 2021-12-07
DM-32843: v23.0.0 release notes

timj:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/doc/changes/DM-13365.feature.rst
+++ b/doc/changes/DM-13365.feature.rst
@@ -1,2 +1,0 @@
-Add ability to cache datasets locally when using a remote file store.
-This can significantly improve performance when retrieving components from a dataset.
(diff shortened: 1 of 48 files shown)


timj:
@TallJimbo no need to look at every entry. My main request is to consider:

* Is it worth maintaining separate registry and datastore entries?
* Should we combine misc and other?

My intent is to backport this to main (which will remove all the v23 news fragments from main).

TallJimbo (approved):
I'd vote for merging `other` and `misc`, and I don't think the separation of registry and datastore entries is useful.

Any idea why some entries have ticket links and others do not?

timj:
I think they all have ticket links. Lines that don't are because the link only appears at the end of the entry and some news fragments are multi-line.

PR #615, opened 2021-12-10
DM-32883: Support StorageClass conversion on get.

timj:
This allows a converter to be registered with a storage class that will be run if the dataset read from the datastore has an incorrect type that can be converted to the expected one.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/storageClass.py
+++ b/python/lsst/daf/butler/core/storageClass.py
@@ -79,6 +79,9 @@ class StorageClass:
     delegate : `str`, optional
         Fully qualified name of class supporting assembly and disassembly
         of a `pytype` instance.
+    converters : `dict` [`str`, `str`], optional
+        Mapping of python type to function that can be called to convert
+        that python type to the valid type of this storage class.
     """
 
     _cls_name: str = "BaseStorageClass"
@@ -87,6 +90,7 @@ class StorageClass:
     _cls_parameters: Optional[Union[Set[str], Sequence[str]]] = None
     _cls_delegate: Optional[str] = None
     _cls_pytype: Optional[Union[Type, str]] = None
+    _cls_converters: Optional[Dict[str, str]] = None
     defaultDelegate: Type = StorageClassDelegate
     defaultDelegateName: str = get_full_type_name(defaultDelegate)
 
@@ -95,7 +99,8 @@ class StorageClass:
                  components: Optional[Dict[str, StorageClass]] = None,
                  derivedComponents: Optional[Dict[str, StorageClass]] = None,
                  parameters: Optional[Union[Sequence, Set]] = None,
-                 delegate: Optional[str] = None):
+                 delegate: Optional[str] = None,
+                 converters: Optional[Dict[str, str]] = None):
         if name is None:
             name = self._cls_name
         if pytype is None:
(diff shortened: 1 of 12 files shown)


python/lsst/daf/butler/core/storageClass.py
```diff
@@ -160,6 +177,41 @@ def derivedComponents(self) -> Dict[str, StorageClass]:
...
+                    try:
+                        candidate_type = doImportType(candidate_type_str)
+                    except ImportError as e:
+                        log.info("Unable to import type %s associated with storage class %s (%s)",
+                                 candidate_type_str, self.name, e)
+                        del self.converters[candidate_type_str]
+                        continue
+
+                try:
+                    converter = doImportType(converter_str)
```
andy-slac, line 205: Maybe add a check that imported object is indeed a function? 

` doImportType` raises TypeError if imported thing is a module, maybe catch that and re-raise with more specific error message?

python/lsst/daf/butler/core/storageClass.py
```diff
@@ -160,6 +177,41 @@ def derivedComponents(self) -> Dict[str, StorageClass]:
...
+
+            # Loop over list because the dict can be edited in loop.
+            for candidate_type_str, converter_str in list(self.converters.items()):
+                if hasattr(builtins, candidate_type_str):
+                    candidate_type = getattr(builtins, candidate_type_str)
+                else:
+                    try:
+                        candidate_type = doImportType(candidate_type_str)
+                    except ImportError as e:
+                        log.info("Unable to import type %s associated with storage class %s (%s)",
```
andy-slac: Should this be a warning, this looks like a potential configuration mistake?

python/lsst/daf/butler/formatters/json.py
```diff
@@ -136,32 +136,36 @@ def _toBytes(self, inMemoryDataset: Any) -> bytes:
...
+                # The content will match the written storage class.
                 # Pydantic models have their own scheme.
                 try:
-                    inMemoryDataset = pytype.parse_obj(inMemoryDataset)
+                    inMemoryDataset = writeStorageClass.pytype.parse_obj(inMemoryDataset)
                 except AttributeError:
                     # Hope that we can pass the arguments in directly
-                    inMemoryDataset = pytype(inMemoryDataset)
+                    inMemoryDataset = writeStorageClass.pytype(inMemoryDataset)
+                inMemoryDataset = readStorageClass.coerce_type(inMemoryDataset)
```
andy-slac, line 170: I would hope that you could go from JSON/YAML directly to `readStorageClass.pytype` without going through `writeStorageClass.pytype`, but it may be safer to do this intermediate step?

andy-slac (approved):

python/lsst/daf/butler/formatters/json.py
timj: I think the problem is that you've written something to JSON with class A but need it to be returned as a class B -- you can't guarantee that the constructor for B will take the serialized JSON structure written for class A. For PropertySet to TaskMetadata that definitely wouldn't work because TaskMetadata doesn't allow construction from a dict object.

python/lsst/daf/butler/core/storageClass.py
timj: I'm not overly enamored by the options here. Checking for `types.FunctionType` seems to be one approach but doesn't cover everything (maybe also check for `types.BuiltInFunctionType`). Seeing if `callable` also works for classes so that's not useful either. You'd find out soon enough when the function falls over later.

andy-slac: My worry here was that when exception happens at the call site due to misconfiguration, it's hard to track that exception to a specific `converter_str` in a config. Could be worth to extend diagnostics there with something that says "this was the converter name that you gave us in the config"?

timj: Maybe trap for `callable` in the doImportType (to filter out variables) and then also catch the converter call and put in the get_full_type_name in a log message.

andy-slac: That should work for most common mistakes. I can imagine more complicated cases, but it's not worth thinking about those at your vacation time before the holidays 🎅

timj: Okay. I just pushed a callable check and a `log.error` in the conversion itself to report the converter.

PR #617, opened 2021-12-13
DM-32950 Make Quantum unit test deterministic

natelust:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/tests/test_quantum.py
+++ b/tests/test_quantum.py
@@ -68,11 +68,9 @@ class QuantumTestCase(unittest.TestCase):
 
         visit = universe['visit']
         region = Circle()
-        hashed = hash(visit).to_bytes(30, 'little')
-        visit_record_42 = visit.RecordClass(id=42, instrument='test', name='test_visit', region=region,
-                                            hash=hashed)
-        visit_record_42 = visit.RecordClass(id=43, instrument='test', name='test_visit', region=region,
-                                            hash=hashed)
+        # create a synthetic value to mock as a visit hash
+        visit_record_42 = visit.RecordClass(id=42, instrument='test', name='test_visit', region=region,)
+        visit_record_42 = visit.RecordClass(id=43, instrument='test', name='test_visit', region=region,)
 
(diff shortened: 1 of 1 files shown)


tests/test_quantum.py
```diff
@@ -68,7 +68,8 @@ def _buildFullQuantum(self, taskName, addRecords=False) -> Tuple[Quantum, Iterab
 
         visit = universe['visit']
         region = Circle()
-        hashed = hash(visit).to_bytes(30, 'little')
+        # create a synthetic value to mock as a visit hash
+        hashed = hash(5786663976355850161).to_bytes(30, 'little')
```
TallJimbo: Did you mean to hash that big int, or just convert the big integer directly to bytes?  It's not obvious to me that hashing the big integer is deterministic, but maybe you know more about its guarantees than I do.

natelust: no, I didn't mean to do that at all, thank you for pointing that out, I was in too much of a rush trying to juggle to notice my highlight replace failed me. I will fix that

natelust: it was supposed to be `(5...1).to_bytes`

TallJimbo: Actually, @natelust and I discussed this out-of-band, and this line isn't actually testing anything.  There are two problems unrelated to this ticket that we should just punt on in order to get the main branch working again:

- passing `hash` below does nothing, because `DimensionRecord` accepts `**kwargs` and silently ignores those it doesn't recognize - it should raise instead;
- the code this is trying to test, at https://github.com/lsst/daf_butler/blob/ad64451f589f4f9d17a884c831cd3412a2483760/python/lsst/daf/butler/core/dimensions/_records.py#L383-L384, looks for an attribute named "hash" when it really should look for any attribute with _type_ `bytes`.  But our current dimension schema only has one such field (in the `skymap` dimension), and it _is_ named `hash`.

Short version: we should just delete this line, and create some new tickets.

TallJimbo (approved):

PR #618, opened 2021-12-14
DM-32831 add a butler remove-runs CLI command

n8pease:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -19,7 +19,8 @@ jobs:
           python-version: 3.8
 
       - name: Install
-        run: pip install mypy pydantic
+        # pin mypy to fix a compatibility issue between mypy 0.92 and pedantic
+        run: pip install "mypy==0.910" pydantic
 
       - name: Install third party stubs
         run: pip install types-requests types-PyYAML types-click types-pkg_resources
--- a/python/lsst/daf/butler/cli/cmd/_remove_runs.py
+++ b/python/lsst/daf/butler/cli/cmd/_remove_runs.py
@@ -1,0 +1,96 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+import click
+from typing import Mapping, Sequence
+
+from ..utils import (
+    ButlerCommand,
+)
+from ..opt import (
+    collection_argument,
+    confirm_option,
+    options_file_option,
+    repo_argument,
+)
+from ... import script
+
+
+# messages emitted by remove-runs, defined separately for use in unit
+# tests.
+noRunCollectionsMsg = "No RUN collections were found."
+willRemoveRunsMsg = "The following RUN collections will be removed:"
+willRemoveDatasetsMsg = "The following datasets will be removed:"
+didRemoveRunsMsg = "The following RUN collections were removed:"
+didRemoveDatasetsMsg = "The following datasets were removed:"
+removedRunsMsg = "Removed collections"
+abortedMsg = "Aborted."
+
+
+def _print_remove(will: bool, runs: Sequence[str], datasets: Mapping[str, int]):
+    """Print the formatted remove statement.
+
+    Parameters
+    ----------
+    will : bool
+        True if remove "will" happen, False if the remove "did" happen.
+    runs : Sequence[str]
+        The RUNs that will be or were removed.
+    datasets : Mapping[str, int]
+        The dataset types & count that will be or were removed.
+    """
+    print(willRemoveRunsMsg if will else didRemoveRunsMsg)
+    print(", ".join(runs))
+    print(willRemoveDatasetsMsg if will else didRemoveDatasetsMsg)
+    print(", ".join([f"{i[0]}({i[1]})" for i in datasets.items()]))
+
+
+@click.command(cls=ButlerCommand)
+@repo_argument(required=True)
+@collection_argument(
+    help="COLLECTION is a glob-style expression that identifies the RUN collection(s) to remove."
+)
+@confirm_option()
+@options_file_option()
+def remove_runs(**kwargs):
+    """Remove one or more RUN collections.
+
+    This command can be used to remove RUN collections and the datasets within
+    them.
+    """
+    confirm = kwargs.pop("confirm")
+    result = script.removeRuns(**kwargs)
+    canRemoveRuns = len(result.runs)
+    if not canRemoveRuns:
+        print(noRunCollectionsMsg)
+        return
+    if confirm:
+        _print_remove(True, result.runs, result.datasets)
+        doContinue = click.confirm("Continue?", default=False)
+        if doContinue:
+            result.onConfirmation()
+            print(removedRunsMsg)
+        else:
+            print(abortedMsg)
+    else:
+        result.onConfirmation()
+        _print_remove(False, result.runs, result.datasets)
--- a/python/lsst/daf/butler/script/removeRuns.py
+++ b/python/lsst/daf/butler/script/removeRuns.py
@@ -1,0 +1,124 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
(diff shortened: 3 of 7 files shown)


tests/test_cliCmdRemoveRuns.py
```diff
@@ -0,0 +1,96 @@
...
+    willRemoveRunsMsg,
+    willRemoveDatasetsMsg,
+    didRemoveRunsMsg,
+    didRemoveDatasetsMsg,
+    removedRunsMsg,
+    abortedMsg,
+    noRepoFoundMsg,
+)
+from lsst.daf.butler.cli.utils import clickResultMsg, LogCliRunner
+from lsst.daf.butler.tests.utils import MetricTestRepo;
```
timj: The linter is not wrong about that semicolon

.github/workflows/mypy.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
           python-version: 3.8
 
       - name: Install
-        run: pip install mypy pydantic
+        run: pip install "mypy==0.910" pydantic
```
n8pease, line 23: @timj does this want a comment, like "pin to address compatibility issue between mypy 0.92 and pedantic"?

python/lsst/daf/butler/script/removeRuns.py
```diff
@@ -0,0 +1,121 @@
...
+        )
+    except MissingCollectionError:
+        collectionNames = list()
+    runs = []
+    datasets: Dict[str, int] = defaultdict(int)
+    for collectionName in collectionNames:
+        if butler.registry.getCollectionType(collectionName).name == "RUN":
+            runs.append(collectionName)
+            all_results = butler.registry.queryDatasets(..., collections=collectionName)
+            for r in cast(DatasetQueryResults, all_results).byParentDatasetType():
```
n8pease: @TallJimbo thanks for the code you gave me to get the number of datasets (https://lsstc.slack.com/archives/C2JPT1KB7/p1638824513252900?thread_ts=1638823786.251200&cid=C2JPT1KB7)
I wasn't able to get this past mypy without a cast tho. Is there a better way to do this for the type checker?

TallJimbo: I think you're basically stuck with a cast or an assert, unless we annotate `queryDatasets` with `typing.overload`: the fact that you passed `...` guarantees a more-derived return type for which this code is valid, but passing a `DatasetType` instance returns a different derived type for which it isn't, and mypy doesn't know about any of that.

.github/workflows/mypy.yaml
timj: Yes. You can put comments in YAML. It should remind people to remove the pin once pydantic becomes compatible with the newer mypy.

python/lsst/daf/butler/cli/cmd/_remove_runs.py
```diff
@@ -0,0 +1,104 @@
...
+# messages emitted by remove-runs, defined separately for use in unit
+# tests.
+noRunCollectionsMsg = "No RUN collections were found."
+willRemoveRunsMsg = "The following RUN collections will be removed:"
+willRemoveDatasetsMsg = "The following datasets will be removed:"
+didRemoveRunsMsg = "The following RUN collections were removed:"
+didRemoveDatasetsMsg = "The following datasets were removed:"
+removedRunsMsg = "Removed collections"
+abortedMsg = "Aborted."
+noRepoFoundMsg = "No repo was found at '{root}'"
```
timj: Maybe say "No Butler repository was found at..."? What do we say in all the other command lines?  Shouldn't this be a standard message?

python/lsst/daf/butler/cli/cmd/_remove_runs.py
```diff
@@ -0,0 +1,104 @@
...
+def remove_runs(**kwargs):
+    """Remove one or more RUN collections.
+
+    This command can be used to remove RUN collections and the datasets within
+    them.
+    """
+    confirm = kwargs.pop("confirm")
+    try:
+        result = script.removeRuns(**kwargs)
+    except FileNotFoundError:
```
timj: This seems inconsistent with every other command where we let the FileNotFoundError continue to the caller. Why is this different?

python/lsst/daf/butler/cli/cmd/_remove_runs.py
```diff
@@ -0,0 +1,104 @@
...
+        print(f"\n{noRepoFoundMsg.format(root=kwargs['repo'])}.\n")
+        return
```
timj: Don't we want a non-zero exit status if this happens?

python/lsst/daf/butler/cli/cmd/_remove_runs.py
```diff
@@ -0,0 +1,104 @@
...
+    canRemoveRuns = len(result.runs)
+    if not canRemoveRuns:
+        print("\n" + noRunCollectionsMsg + "\n")
+        return
+    if confirm:
```
timj, line 86: Pondering reorganizing this if-statement as the following -- it is shorter and makes the no-confirm case more obvious but I'm not sure it's worth it.
```python
if confirm:
    _print_remove(True, ...)
    doContinue = click.confirm()
    if doContinue:
        result.onConfirmation()
        print(removedRunsMsg)
    else:
        print(abortedMsg)
else:
    result.onConfirmation()
    _print_remove(False, ...)
```

python/lsst/daf/butler/script/removeRuns.py
```diff
@@ -0,0 +1,121 @@
...
+from typing import cast, Callable, Dict, List, Mapping, Sequence, Tuple
+
+from .. import Butler
+from ..registry import CollectionType, MissingCollectionError
+from ..registry.queries import DatasetQueryResults
+
+
+@dataclass
+class RemoveRunsResult:
+    """Container to return to the cli command; contains the names of runs that
```
timj: Please put the content after the semi-colon on a new line with a blank line between so that the class summary is distinct from the explanation.

python/lsst/daf/butler/script/removeRuns.py
```diff
@@ -0,0 +1,121 @@
...
+    will be deleted, and a map of dataset type to how many of that dataset will
+    be deleted. Also contains the callback funciton to execute the remove upon
```
timj: Typo: funciton

python/lsst/daf/butler/script/removeRuns.py
```diff
@@ -0,0 +1,121 @@
...
+    ----------
+    repo : `str`
+        The URI to the repostiory.
+    collection : `str`
+        The collection string to search for. Same as the `expression`
+        argument to `registry.queryCollections`.
+
+    Returns
+    -------
+    runs, datasets : [List[str], Mapping[str, int]]
```
timj: This is not how numpydoc declares multiple returns. You need something like:

```python
Returns
-------
runs : `list` of `str`
    The runs...
datasets : `dict` [`str`, `int`]
    The datasets...
```

python/lsst/daf/butler/script/removeRuns.py
```diff
@@ -0,0 +1,121 @@
...
+    butler = Butler(repo)
+    try:
+        collectionNames = list(
+            butler.registry.queryCollections(
+                collectionTypes=frozenset((CollectionType.RUN,)),
+                expression=collection,
+                includeChains=False,
+            )
+        )
+    except MissingCollectionError:
```
timj, line 80: Wouldn't hurt to add a trivial test here for a missing run collection.

python/lsst/daf/butler/script/removeRuns.py
```diff
@@ -0,0 +1,121 @@
...
+        collectionNames = list()
+    runs = []
+    datasets: Dict[str, int] = defaultdict(int)
+    for collectionName in collectionNames:
+        if butler.registry.getCollectionType(collectionName).name == "RUN":
```
timj: Isn't it a bug if this is not a RUN since you filtered out none RUN collections in the `queryCollections` above? Maybe instead write this as:

```python
assert butler.registry.getCollectionType(collectionName).name == "RUN"
```
because it should never happen unless the logic a few lines above is changed? Or have I missed something with that query?

python/lsst/daf/butler/script/removeRuns.py
timj: This is all caused by remote registry not knowing about DatasetQueryResults -- in that case it might be better to do it with an instance check rather than a cast so that RemoteRegistry being used here will break early.

timj (approved):
I have some comments and some things that need fixing but over all it looks okay.

python/lsst/daf/butler/cli/cmd/_remove_runs.py
n8pease: We don't do this checking in other places (yet?) so there's no standard. Maybe we should just let it fall through. More in the comment at line 85...

python/lsst/daf/butler/cli/cmd/_remove_runs.py
n8pease: Just trying to be more complete. Maybe it's better to be consistent with other existing commands? Qserv folks were complaining when the CLI tools allowed errors to continue to the caller. I find it useful sometimes to have _all_ the info, but others seem to think it's messy, not useful. I'll remove the handling to be consistent, unless you think it should stay.

timj: If it stays you still have to fix the exit status. I'd be inclined to remove it for now. Andy already added some special code to limit the amount of stack trace that is reported from these commands.

PR #619, opened 2021-12-15
DM-32986: fix option handling in butler CLI presets files

n8pease:
accept option names in the preset file,
map them to argument names before applying
them to the click.context.default_map.

also fail if an option is named in the
preset file that does not have a matching
parameter in the current command's options
or arguments.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-32986.bugfix.rst
+++ b/doc/changes/DM-32986.bugfix.rst
@@ -1,0 +1,2 @@
+In the butler presets file, use option names that match the butler CLI command option names (without leading dashes).
+Fail if option names used in the presets file do not match options for the current butler command.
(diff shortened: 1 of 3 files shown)


timj:
There should really be a test that the `--something-something` options work (with or without the leading `--`) to ensure that people don't have to write `something_something`.

n8pease:
> There should really be a test that the --something-something options work (with or without the leading --) to ensure that people don't have to write something_something.

I didn't think it was needed, since the name change is based on lookups not transforming the inputs. But it's easy enough to add so I put one in.

timj (approved):
Looks good to me.

PR #620, opened 2021-12-17
DM-32437: format with black and isort

TallJimbo:
## Checklist

- [x] ran Jenkins


timj (approved):
This looks good. The docstring triple quote problem fix you did seems reasonable. It won't
rearrange the content and forcing it to a second line will always place the triple quotes on its own line.

PR #621, opened 2021-12-17
DM-33013: Add ipython display function to correctly print object

mfisherlevine:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/dimensions/_records.py
+++ b/python/lsst/daf/butler/core/dimensions/_records.py
@@ -287,6 +287,13 @@ class DimensionRecord:
         mapping = {name: getattr(self, name) for name in self.__slots__}
         return (_reconstructDimensionRecord, (self.definition, mapping))
 
+    def _repr_html_(self) -> str:
+        """Override the default representation in IPython/Jupyter notebooks.
+
+        This gives a more readable output that understands embedded newlines.
+        """
+        return f"<pre>{self}<pre>"
+
     def to_simple(self, minimal: bool = False) -> SerializedDimensionRecord:
         """Convert this class to a simple python type.
 
--- a/tests/test_simpleButler.py
+++ b/tests/test_simpleButler.py
@@ -596,7 +596,7 @@ class SimpleButlerTestCase(unittest.TestCase):
                     from_json = type(test_item).from_json(json_str, universe=butler.registry.dimensions)
                     self.assertEqual(from_json, test_item, msg=f"From JSON '{json_str}' using universe")
 
-    def testJsonDimensionRecords(self):
+    def testJsonDimensionRecordsAndHtmlRepresentation(self):
         # Dimension Records
         butler = self.makeButler(writeable=True)
         butler.import_(filename=os.path.join(TESTDIR, "data", "registry", "hsc-rc2-subset.yaml"))
@@ -614,6 +614,11 @@ class SimpleButlerTestCase(unittest.TestCase):
                     self.assertEqual(simple, fromDirect)
                     # Also check equality of each of the components as dicts
                     self.assertEqual(r_json.toDict(), r.toDict())
+
+                    # check the html representation of records
+                    r_html = r._repr_html_()
+                    self.assertTrue(isinstance(r_html, str))
+                    self.assertIn(dimension, r_html)
 
     def testWildcardQueries(self):
         """Test that different collection type queries work."""


python/lsst/daf/butler/core/dimensions/_records.py
```diff
@@ -287,6 +287,9 @@ def __reduce__(self) -> tuple:
         mapping = {name: getattr(self, name) for name in self.__slots__}
         return (_reconstructDimensionRecord, (self.definition, mapping))
 
+    def _ipython_display_(self) -> None:
```
timj: Is there a reason why this is being used and not `_repr_html_`?

timj (commented):
I'm not a huge fan of special-purpose display methods but this does need a docstring and it also needs a test to make sure it doesn't crash when it runs.

The test can capture stdout or something (with `_repr_html_` it's easier because that's a return value).

python/lsst/daf/butler/core/dimensions/_records.py
mfisherlevine: Yeah, I because I didn't want it to print in a different font or anything fancy at all, just to have it actually print the object correctly with all the nice newline formatting that y'all actually went to the trouble to include!

I totally agree on the docstring, but do you really think a test which is just calling `print()` (which just calls `str`) actually needs its own test?! Surely either `__str__` is already being tested, or it's not, and that was previously deemed OK, no?

timj: We need a test because how else do we know that the code isn't broken? Just because `__str__` is being tested doesn't mean that your new code doesn't have a problem. You are asserting by inspection that it must be fine.

As for `_repr_html_` wouldn't that be:

```python
def _repr_html_(self) -> str:
    return f"<pre>{self}<pre>"
```
?

mfisherlevine: Sure, that looks fine. Would you prefer that for some reason?

Also, I'm not familiar with the tests in this package, and couldn't, at a glance, see the `test_records.py` or equivalent. There must be a place with an instantiated record object where I can test this easily - could you point me to it please?

timj: The `_repr_html_` version has the advantage of being more testable. Does it work?

How about a new method after this one: https://github.com/lsst/daf_butler/blob/main/tests/test_simpleButler.py#L599?

mfisherlevine: Sure, looks good, thanks. Personally, creating a whole other butler just to test a print statement feels like a net-negative in terms of overkill and build-time slowdown, but it's your package, so I'm happy to include it.

mfisherlevine: I think I've dealt with everything now, take a look and see what you think.

tests/test_simpleButler.py
```diff
@@ -615,6 +615,19 @@ def testJsonDimensionRecords(self):
...
+        butler = self.makeButler(writeable=True)
+        butler.import_(filename=os.path.join(TESTDIR, "data", "registry", "hsc-rc2-subset.yaml"))
+
+        for dimension in ("detector", "visit"):
+            records = butler.registry.queryDimensionRecords(dimension, instrument="HSC")
+            for r in list(records)[0:1]:
+                for minimal in (True, False):
+                    ret = r._repr_html_()
+                    self.assertTrue(isinstance(ret, str))
+                    self.assertFalse(ret == "")
```
TallJimbo: I don't see `minimal` being used inside the loop body.

TallJimbo (approved):
Minor comment on the test code.  Otherwise looks good to me.

python/lsst/daf/butler/core/dimensions/_records.py
timj: I'm fine with you putting the ipython repr test in the JSON test loop.

tests/test_simpleButler.py
```diff
@@ -615,6 +615,19 @@ def testJsonDimensionRecords(self):
...
+        """Test the printing of things for ipython/notebooks doesn't crash."""
+        butler = self.makeButler(writeable=True)
+        butler.import_(filename=os.path.join(TESTDIR, "data", "registry", "hsc-rc2-subset.yaml"))
+
+        for dimension in ("detector", "visit"):
+            records = butler.registry.queryDimensionRecords(dimension, instrument="HSC")
+            for r in list(records)[0:1]:
+                for minimal in (True, False):
+                    ret = r._repr_html_()
+                    self.assertTrue(isinstance(ret, str))
```
timj: I'd prefer this test was a bit more specific. It could check something like:
```python
self.assertIn(dimension, ret)
```
couldn't it? Then the `assertFalse` below wouldn't be needed either (and that should have been written as `self.assertNotEqual(ret, "")` if it was to be kept).

python/lsst/daf/butler/core/dimensions/_records.py
```diff
@@ -287,6 +287,15 @@ def __reduce__(self) -> tuple:
         mapping = {name: getattr(self, name) for name in self.__slots__}
         return (_reconstructDimensionRecord, (self.definition, mapping))
 
+    def _repr_html_(self) -> str:
+        r"""Call (called automatically) when displaying in IPython/notebooks.
+
+        If one types the name of a record object and just hits return, the
```
timj: I think this text is trying to say what happens when this method doesn't exist -- I think I'd rather it said what this method does.

```
Overrides the default representation of these objects in IPython/Jupyter giving a more
readable output that understands embedded newlines.
```
or something.

timj (approved):

python/lsst/daf/butler/core/dimensions/_records.py
mfisherlevine: I dislike this as much as you do - this was my 5th or so attempt, because D401 meant I was getting a linter error for not putting it in the imperative, but wanting to get across that one doesn't actually call this method directly, but is called from by the ipython/notebook kernel automatically.

I guess I can use `Override` instead of `Override` and chop down to a single line to make something comply here, but it's going to be pretty gross whatever I do in order to comply with the rules.

timj: I was commenting on the second paragraph, not the summary line.

PR #622, opened 2021-12-29
DM-33049: Allow duplicate values in dataId

timj:
`detector` and `raft` and `day_obs` and `exposure` should be allowed so long as they are consistent.

## Checklist

- [x] ran Jenkins


mfisherlevine (approved):
Looks great!

PR #626, opened 2022-01-07
DM-33155: Also allow storage class conversion on butler.put

timj:
The infrastructure already existed to support it and there
is no particular reason why get should convert but put
should not. Doing this provides a way for newer classes
to be used in code but whilst still supporting old
dataset type definitions.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_butler.py
+++ b/tests/test_butler.py
@@ -154,10 +154,7 @@ class ButlerPutGetTests:
     def tearDown(self):
         removeTestTempDir(self.root)
 
-    def runPutGetTest(self, storageClass, datasetTypeName):
-        # New datasets will be added to run and tag, but we will only look in
-        # tag when looking up datasets.
-        run = "ingest"
+    def create_butler(self, run, storageClass, datasetTypeName):
         butler = Butler(self.tmpConfigFile, run=run)
 
         collections = set(butler.registry.queryCollections())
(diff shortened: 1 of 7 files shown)


tests/test_butler.py
```diff
@@ -1372,6 +1372,15 @@ def testPytypeCoercion(self):
         datasetType_ori = butler.registry.getDatasetType(datasetTypeName)
         self.assertEqual(datasetType_ori.storageClass.name, "StructuredDataNoComponents")
 
+        # Put a dict and this should coerce to a MetricsExample
+        test_dict = {"summary": {"a": 1}, "output": {"b": 2}}
+        metric_ref = butler.put(test_dict, datasetTypeName, dataId=dataId, visit=424)
+        test_metric = butler.getDirect(metric_ref)
+        self.assertEqual(test_metric.summary, test_dict["summary"])
+        self.assertEqual(test_metric.output, test_dict["output"])
+        # Remove the dataset because another will be stored later in the test.
```
TallJimbo: I think it might be better to just duplicate the setup code and put this in another test method; it's incidental duplication (not so bad at all) vs. starting the kind of test daisy-chaining we've both regretted in the past, over in `runPutGetTests`.

TallJimbo (approved):

tests/test_butler.py
timj: The test is called `testPytypeCoercion` so it's not a completely irrelevant thing to put in here. If I put it into a different run collection I wouldn't need to remove the dataset.

I do agree that we've made life hard for ourselves with `runPutGetTests` and then having every single datastore run every test.

TallJimbo: It's not the relevance of this to the test method that worries me; it's making one test's setup involve another test's cleanup - it makes it hard to modify the middle test without digging into the next one.

timj: I've separated it out and also factored out the butler creation code from runPutGetTest so we don't need to run all the tests again.

PR #629, opened 2022-01-10
DM-33204: Add deprecation warning for ButlerURI

timj:
The code is a bit hairy...

@TallJimbo let me know if you have any better ideas, especially regarding the subclasses trying to instantiate something.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -23,7 +23,7 @@ jobs:
         run: pip install "mypy==0.910" pydantic
 
       - name: Install third party stubs
-        run: pip install types-requests types-PyYAML types-click types-pkg_resources
+        run: pip install types-requests types-PyYAML types-click types-pkg_resources types-Deprecated
 
       - name: Install lsst.utils
         run: |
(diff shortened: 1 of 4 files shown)


TallJimbo (approved):
I don't really have an opinion about whether this is better or worse than no deprecation warning at all, but I don't have any better ideas.  I spent a few minutes researching import hooks in the hopes of finding a way to emit a warning when the symbol is imported or accessed as an attribute of a module, but came up empty-handed.

timj:
There can't possibly be many ButlerURI users out in the wild so this should be fine even if it's a bit convoluted.

timj:
Looks like I can't merge this until `pipe_base` loses `ButlerURI` -- mostly because one of the tests in `pipelines_check` is getting confused by the warnings come out of `pipe_base` (although I'll take a quick look at the test and see if it's a real problem).

TallJimbo:
`pipe_base` changes are pretty thoroughly backed up behind other changes now.  If you want to move faster feel free to convert it (by cherry-picking b014171b73c1271bfc3db8df568fcff7fa6d8931 or otherwise) and I'll rebase.

timj:
I'll take a quick look at that.

PR #632, opened 2022-01-14
DM-33278: Improve storage class conversion

timj:
* Add a dataset type compatibility method.
* Add support for specifying methods.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/lsst.daf.butler/formatters.rst
+++ b/doc/lsst.daf.butler/formatters.rst
@@ -88,6 +88,13 @@ It does not make sense for subsetting to be applied to the integer pixel count.
 
 For this reason read parameters for derived components are processed prior to calculating the derived component.
 
+Type Converters
+^^^^^^^^^^^^^^^
+
+It is sometimes convenient to be able to call `Butler.put` with a Python type that is not a match to the storage class defined for that dataset type in the registry.
+Storage classes can be defined with converters that declare which Python types can be coerced into the required type, and what functions or method should be used to perform that conversion.
+Butler can support this on `Butler.put` and `Butler.get`, the latter being required if the dataset type definition has been changed in registry after a dataset was stored.
+
 Defining a Storage Class
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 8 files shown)


doc/lsst.daf.butler/formatters.rst
```diff
@@ -139,6 +146,22 @@ You can do this using YAML anchors and references but the preferred approach is
...
+      pytype: dict
+      converters:
+        lsst.daf.base.PropertySet: toDict()
+    TaskMetadata:
+      pytype: lsst.pipe.base.TaskMetadata
+      converters:
+        lsst.daf.base.PropertySet: lsst.pipe.base.TaskMetadata.from_metadata
+
+In the first definition, the configuration says that if a ``PropertySet`` object is given then the ``toDict()`` method can be called on it and the returned value will be a `dict`.
+In the second definition a ``PropertySet`` is again specified but this time the ``from_metadata`` class method will be called with the ``PropertySet`` as the first parameter and a ``TaskMetadata`` will be returned.
```
andy-slac: This difference between `()` and no `()` confused me a lot when I first saw it and it was not obvious what it means. I had to read everything more than once to figure it out. Now I think that `toDict()` is basically the same as `lsst.daf.base.PropertySet.toDict` (non-bound method). Would it be better to drop `()` and use non-bound method syntax for symmetry?

python/lsst/daf/butler/core/datasets/type.py
```diff
@@ -240,25 +240,78 @@ def __repr__(self) -> str:
...
         if self._storageClass is not None and other._storageClass is not None:
             if self._storageClass != other._storageClass:
                 return False
         else:
             if self._storageClassName != other._storageClassName:
                 return False
-        if self._isCalibration != other._isCalibration:
+        return True
+
+    def is_compatible_with(self, other: "DatasetType") -> bool:
```
andy-slac: Do you need to quote `DatasetType` here?

andy-slac: Does "compatible" mean compatible for both `get` and `put`? Is it symmetrical (X.is_compatible_with(Y) === Y.is_compatible_with(X))?

andy-slac (approved):
Looks good, couple of minor comments/questions.

python/lsst/daf/butler/core/datasets/type.py
timj: It really means `self_sc.can_convert(other_sc)` -- You have made me realize that ctrl_mpexec is going to have to try to distinguish inputs (so can convert on get to the required type for the Task) and outputs (so can convert the type used in the Task to the required type in the butler).

timj: I think I'm confused between `from __future__ import annotations` being enabled and not being enabled. There was some other place where we did have to use strings if the type was used in the context of its own class but maybe that was to do with pydantic.

TallJimbo: I believe we cannot use `from __future__ import annotations` with pydantic (which is really a shame, and I think it means we should mostly put pydantic classes in their own modules, so we can use it everywhere else).  And without `__future__.annotations`, you do have to quote in places like this.

doc/lsst.daf.butler/formatters.rst
timj: I've been pondering this and the `()` syntax has the advantage that it doesn't lead to duplicating the full class name on both sides:

```yaml
lsst.daf.butler.tests.MetricsExample: exportAsDict()
```
vs
```yaml
lsst.daf.butler.tests.MetricsExample: lsst.daf.butler.tests.MetricsExample.exportAsDict
```
The latter does work and doesn't require any code changes to make it work without this ticket. Maybe that's a good enough argument for not trying to bother with the more succinct form.

PR #633, opened 2022-01-20
DM-30966: Make Spectractions Gen3-serializable

mfisherlevine:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -280,3 +280,7 @@ storageClasses:
     converters:
       # Compatible python types.
       lsst.daf.base.PropertySet: lsst.pipe.base.TaskMetadata.from_metadata
+  SpectractorSpectrum:
+    pytype: spectractor.extractor.spectrum.Spectrum
+  SpectractorImage:
+    pytype: spectractor.extractor.images.Image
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -278,3 +278,7 @@ storageClasses:
     converters:
       # Compatible python types.
       lsst.daf.base.PropertySet: lsst.pipe.base.TaskMetadata.from_metadata
+  SpectractorSpectrum:
+    pytype: lsst.atmospec.formatters.Spectrum
+  SpectractorImage:
+    pytype: lsst.atmospec.formatters.Image
```
timj: I understand this is a draft PR but these class names seem to imply that you are creating special classes specifically for I/O and they aren't actually Spectractor classes that you are serializing? (putting them in the same hierarchy as the formatter itself).

mfisherlevine: That was temporarily deliberate I think (I was brainstorming this with Jim), but yes, you're right, now that should be changed to the native type, I think. I've pushed the fix.

czwa (approved):
I think I'd be happier if these used the `FitsGenericFormatter`, but I think that would involve pushing changes into the `spectractor` code, and I'm not clear on how difficult that is.  The `Spectrum` and `Image` types would need to implement `readFits` and `writeFits` methods (likely as wrappers around the existing `load_X` and `save_X` methods).

PR #635, opened 2022-01-20
DM-33303: Allow DatasetType to differ to registry on get/put

timj:
If the DatasetTypes are compatible with each other.

## Checklist

- [ ] ran Jenkins


andy-slac (approved):
Looks good.

PR #636, opened 2022-01-21
DM-33347: Fix compatibility with moto 3.0

timj:
moto3 broke the s3 tests solely in test_butler.py. It seems
that something has changed in using the mock_s3 class decorator
such that the inherited methods (the ones that are doing all
the testing) are not actually using the mocks at all but are
using raw boto. Moving the start and stop calls explicitly
into setUp and tearDown fixes the problem.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/tests/test_butler.py
+++ b/tests/test_butler.py
@@ -1590,7 +1590,6 @@ class ButlerMakeRepoOutfileUriTestCase(ButlerMakeRepoOutfileTestCase):
 
 
 @unittest.skipIf(not boto3, "Warning: boto3 AWS SDK not found!")
-@mock_s3
 class S3DatastoreButlerTestCase(FileDatastoreButlerTests, unittest.TestCase):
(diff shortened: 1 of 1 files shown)


tests/test_butler.py
```diff
@@ -1590,7 +1590,6 @@ def setUp(self):
 
 
 @unittest.skipIf(not boto3, "Warning: boto3 AWS SDK not found!")
-@mock_s3
```
ktlim, line 1593: This looks like it follows the documentation, but it's a little unclear how inherited tests are handled.

Since this code works, I guess it's OK.

ktlim (approved):

tests/test_butler.py
timj: Yes, the exact same mocking works fine in the `resources` package but that doesn't inherit tests from another class.

PR #638, opened 2022-01-27
DM-33417: manual back port of DM-33360 — Improve performance of https as_local() method

airnandez:
## Checklist

- [x] ran Jenkins: done on macOS, ongoing on centOS
- [ ] added a release note for user-visible changes to `doc/changes` (n/a)

### DIFF
--- a/.github/workflows/mypy.yaml
+++ b/.github/workflows/mypy.yaml
@@ -19,7 +19,8 @@ jobs:
           python-version: 3.8
 
       - name: Install
-        run: pip install mypy pydantic
+        # pin mypy to fix a compatibility issue between mypy 0.92 and pedantic
+        run: pip install "mypy==0.910" pydantic
 
       - name: Install third party stubs
         run: pip install types-requests types-PyYAML types-click types-pkg_resources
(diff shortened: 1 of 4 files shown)


timj:
I think you will need to cherry pick 9afaeb3e  to fix the mypy problem.

Can you also add in that as_local test that was included in the original PR? That can go into tests/test_uri.py (at least one of the tests so that as_local can be shown to work). It will have to use a ButlerURI of course.

timj:
The test failures are from moto3. You will have to pin the moto version to <3 in requirements.txt (or we backport DM-33347 -- which isn't needed because we pin moto in the rubin-env that is used for v23).

timj:
Generally we don't want the branch name for the backport to differ from the original branch (so we want to reuse the same ticket) -- but I guess in this case it's probably okay because a completely different package is being modified.

airnandez:
> cherry pick 9afaeb3 to fix the mypy problem.

✅

> add in that as_local test ... into tests/test_uri.py

✅

> pin the moto version to <3 in requirements.txt

✅

> Generally we don't want the branch name for the backport to differ from the original branch...

Understood. I thought that since this was a different PR it required a different JIRA issue and therefore a different branch name.

Thanks for your guidance!

timj:
I merged it.  Yes, we add v23 to the ticket branch name for backports.

PR #639, opened 2022-01-27
DM-33414: Allow id to be used in file template

timj:
@gpdf / @TallJimbo this does the job -- I also restrict the id to having to be in the filename itself if it is specified anywhere.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/fileTemplates.py
+++ b/python/lsst/daf/butler/core/fileTemplates.py
@@ -459,10 +459,7 @@ class FileTemplate:
         if component is not None:
             fields["component"] = component
 
-        usedRun = False
         fields["run"] = ref.run
-
-        usedId = False
         fields["id"] = ref.id
 
         fmt = string.Formatter()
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/core/fileTemplates.py
```diff
@@ -552,9 +558,11 @@ def format(self, ref: DatasetRef) -> str:
...
             )
 
-        # Complain if there's no run
-        if not usedRun:
-            raise KeyError("Template does not include 'run'.")
+        # Complain if there's no run or id
+        if not usedRun and not usedId:
+            missing = ("run" if not usedRun else None, "id" if not usedId else None)
+            text = " or ".join(f"'{m}'" for m in missing if m is not None)
+            raise KeyError(f"Template does not include {text}.")
```
TallJimbo: Is there a similar check for the dataset type name that could also be relaxed?

timj: It doesn't look like we require the dataset type to be in the template. "run" was the only thing we required from what I could see (and I was able to use a file template of just `{id}`.

timj: The fact that this line has no coverage implies that we catch the no-run logic higher up and none of this is needed any more.

TallJimbo: Sounds like we should just remove it, then, and update the docs to say it's expected?

TallJimbo (approved):
Seems fine - functionality is clearly what we want, but I'm a little worried that the template validation code is getting a little hard to follow now that we have so many rules (especially if we don't really understand why some of the new lines aren't in the coverage report).

python/lsst/daf/butler/core/fileTemplates.py
timj: I think that this block of code was an early attempt at template validation but at some point we moved the validation logic earlier and never noticed that this block was irrelevant. I'll remove it.

PR #644, opened 2022-02-16
DM-33326: Fix round-tripping of spectractor spectra

mfisherlevine:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -284,3 +284,5 @@ storageClasses:
     pytype: spectractor.extractor.spectrum.Spectrum
   SpectractorImage:
     pytype: spectractor.extractor.images.Image
+  Spectraction:
+    pytype: lsst.atmospec.spectraction.Spectraction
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -284,3 +284,5 @@ storageClasses:
     pytype: spectractor.extractor.spectrum.Spectrum
   SpectractorImage:
     pytype: spectractor.extractor.images.Image
+  SpectractorPickle:
```
timj: I'd rather we avoid defining a storage class name after it's serialization format if at all possible. The storage class should hew closer to the python type rather than the output file format.

czwa (approved):
Can't this just be a `Spectraction`?

Don't forget to add the change message to the doc directory.

python/lsst/daf/butler/configs/storageClasses.yaml
mfisherlevine: Yeah, that was a deliberate choice to be honest, because it's so fragile and shouldn't be around to stay, so wanted people using it to be aware that they're touching a pickle. I can remove it if you prefer, but this was a conscious choice to try to mark it as dodgy.

PR #645, opened 2022-02-17
DM-33684: Support concurrent table creation

andy-slac:
The patch catches exceptions due to potential concurrent table creation
by different process and re-checks that table is indeed created.

## Checklist

- [X] ran Jenkins


TallJimbo (approved):
Looks good!  My only potential concern is that this does push us further in the direction of not permitting transactions around DDL, because I think we might get a rollback in this new `try`/`except` even when the exception is caught (unless maybe there was a SAVEPOINT).  But as you said in Jira, that's already not allowed, so it's not actually a change.

andy-slac:
Indeed, an exception in table.create() will abort a transaction in Postgres, if later we need to make `ensureTableExists()` transaction-safe, the logic will have to change.

PR #646, opened 2022-02-18
DM-33638: Check datastore cache for existence before checking remote datastore

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/datastoreCacheManager.py
+++ b/python/lsst/daf/butler/core/datastoreCacheManager.py
@@ -714,7 +714,7 @@ class DatastoreCacheManager(AbstractDatastoreCacheManager):
         """
         if self._cache_directory is None:
             return False
-        if self.cache_size == 0:
+        if self.file_count == 0:
             return False
 
         if extension is None:
--- a/python/lsst/daf/butler/core/datastoreCacheManager.py
+++ b/python/lsst/daf/butler/core/datastoreCacheManager.py
@@ -171,6 +171,10 @@ class CacheEntry(BaseModel):
         )
 
 
+class _MarkerEntry(CacheEntry):
+    pass
+
+
 class CacheRegistry(BaseModel):
     """Collection of cache entries."""
 
@@ -220,8 +224,21 @@ class CacheRegistry(BaseModel):
     def items(self) -> ItemsView[str, CacheEntry]:
         return self._entries.items()
 
-    def pop(self, key: str, default: Optional[CacheEntry] = None) -> Optional[CacheEntry]:
-        entry = self._entries.pop(key, default)
+    # An private marker to indicate that pop() should raise if no default
+    # is given.
+    __marker = _MarkerEntry(name="marker", size=0, ref=0, ctime=datetime.datetime.utcfromtimestamp(0))
+
+    def pop(self, key: str, default: Optional[CacheEntry] = __marker) -> Optional[CacheEntry]:
+        # The marker for dict.pop is not the same as our marker.
+        if default is self.__marker:
+            entry = self._entries.pop(key)
+        else:
+            entry = self._entries.pop(key, self.__marker)
+            # Should not attempt to correct for this entry being removed
+            # if we got the default value.
+            if entry is self.__marker:
+                return default
+
         self._decrement(entry)
         return entry
 
@@ -680,7 +697,7 @@ class DatastoreCacheManager(AbstractDatastoreCacheManager):
                 "Entries no longer on disk but thought to be in cache and so removed: %s", ",".join(missing)
             )
             for path_in_cache in missing:
-                self._cache_entries.pop(path_in_cache)
+                self._cache_entries.pop(path_in_cache, None)
 
     def known_to_cache(self, ref: DatasetRef, extension: Optional[str] = None) -> bool:
         """Report if the dataset is known to the cache.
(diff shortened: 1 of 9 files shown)


python/lsst/daf/butler/core/datastoreCacheManager.py
```diff
@@ -282,6 +282,35 @@ def should_be_cached(self, entity: Union[DatasetRef, DatasetType, StorageClass])
         """
         raise NotImplementedError()
 
+    @abstractmethod
+    def known_to_cache(self, ref: DatasetRef, extension: Optional[str] = None) -> bool:
```
timj, line 339: I do wonder whether I should put an underscore in front of this name. It's not a method that we should encourage people using. It's only here as a shortcut for the datastore checking.

andy-slac: Is `AbstractDatastoreCacheManager` exposed to "people"? If it is internal to datastores' implementation, then we probably don't care if it's public or not.

python/lsst/daf/butler/core/datastoreCacheManager.py
```diff
@@ -653,6 +682,57 @@ def scan_cache(self) -> None:
...
+                if ref.id == entry.ref:
+                    return True
+            return False
+
+        else:
+            # Extension is known so we can do an explicit look up for the
+            # cache entry.
+            cached_location = self._construct_cache_name(ref, extension)
+            path_in_cache = cached_location.relative_to(self.cache_directory)
+            assert path_in_cache is not None  # For mypy
```
andy-slac, line 784: These three lines feel like you want a separate method `_construct_cache_name_relative(ref, extension) -> str`?

python/lsst/daf/butler/core/datastoreCacheManager.py
```diff
@@ -653,6 +682,57 @@ def scan_cache(self) -> None:
...
+        in this specific instant and does not indicate whether the file
+        can be read from the cache later. `find_in_cache()` should be called
+        if the cached file is to be used.
+
+        This method does not force the cache to be re-scanned and so can miss
+        cached datasets that have recently been written by other processes.
+        """
+        if self._cache_directory is None:
+            return False
+        if self.cache_size == 0:
```
andy-slac: I guess we do not expect zero-size files in datastore?

python/lsst/daf/butler/core/datastoreCacheManager.py
```diff
@@ -653,6 +682,57 @@ def scan_cache(self) -> None:
...
+            return False
+
+        if extension is None:
+            # Look solely for matching dataset ref ID and not specific
+            # components.
+            for entry in self._cache_entries.values():
+                if ref.id == entry.ref:
+                    return True
```
andy-slac: If this method is used frequently maybe it makes sense to index cache by DatasetId?

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1455,6 +1471,15 @@ def mexists(
...
             Mapping from dataset to boolean indicating existence.
+
+        Notes
+        -----
+        To minimize potentially costly remote existence checks, the local
+        cache is checked as a proxy for existence. If a file for this
+        `DatasetRef` does exist no check is done for the actual URI. This
+        could result in possibly unexpected behavior if the dataset itself
+        has been removed from the datastore by another process whilst it is
+        still in the cache.
```
andy-slac, line 1482: Should the check in local cache be combined with the registry check for DatasetId existence, or is it still too expensive?

andy-slac (approved):

python/lsst/daf/butler/core/datastoreCacheManager.py
timj: Agreed. When I tried to remember how the cache key was formed I was a bit shocked to see that I repeated the logic in multiple places.

python/lsst/daf/butler/core/datastoreCacheManager.py
timj: Actually, I wanted that check to be "if there are no files in the cache" not "total size of cache"...

python/lsst/daf/butler/datastores/fileDatastore.py
timj: The cache is only checked if the datastore registry already thinks it might exist (or we are in execution butler mode where there might not be a registry entry anyhow).

python/lsst/daf/butler/core/datastoreCacheManager.py
timj: Yes, the assumption is that this is something that only datastores care about. Hopfully the "don't use this" caveat in the description is enough.

python/lsst/daf/butler/core/datastoreCacheManager.py
timj: Yes. With a few thousand files in the cache and 100k files to check existence for it could get annoying.

timj: We can't index by datasetId because there are disassembled composites so one datasetRef can be multiple files. The uniqueness is therefore the tuple of datasetId + component (which is effectively what is encoded in the file name).

python/lsst/daf/butler/core/datastoreCacheManager.py
timj: Although now that I look again I see that these two lines:

```python
cached_location = self._construct_cache_name(ref, extension)
path_in_cache = cached_location.relative_to(self.cache_directory)
```
are never consecutive in the code other than this new method so it's not going to be a straightforward tweak.

python/lsst/daf/butler/core/datastoreCacheManager.py
timj: I've added an extra dict into `CacheRegistry` to simplify this.

PR #647, opened 2022-02-19
DM-32250: minor tweaks to typing configuration

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -39,8 +39,23 @@ ignore_missing_imports = True
 [mypy-httpx.*]
 ignore_missing_imports = True
 
+# Don't check LSST packages generally or even try to import them, since most
+# don't have type annotations.
+
 [mypy-lsst.*]
 ignore_missing_imports = True
+ignore_errors = True
+
+# Do expect to import type annotations from LSST middleware dependencies that
+# should have them, but don't re-check them as they have their own mypy.ini
+# files with their own exceptions.
+
+[mypy-lsst.utils.*]
+ignore_missing_imports = False
+ignore_errors = True
+
+[mypy-lsst.resources.*]
+ignore_missing_imports = False
 ignore_errors = True
 
 # Check all of daf.butler...
--- a/python/lsst/daf/butler/core/storageClassDelegate.py
+++ b/python/lsst/daf/butler/core/storageClassDelegate.py
@@ -246,8 +246,8 @@ class StorageClassDelegate:
         return component
 
     def disassemble(
-        self, composite: Any, subset: Optional[Iterable] = None, override: bool = None
-    ) -> Dict[str, Any]:
+        self, composite: Any, subset: Optional[Iterable] = None, override: Optional[Any] = None
+    ) -> Optional[Dict[str, DatasetComponent]]:
         """Disassembler a composite.
 
         This is a generic implementation of a disassembler.
@@ -269,7 +269,7 @@ class StorageClassDelegate:
 
         Returns
         -------
-        components : `dict`
+        components : `dict` or `None`
             `dict` with keys matching the components defined in
             `StorageClassDelegate.storageClass`
             and values being `DatasetComponent` instances describing the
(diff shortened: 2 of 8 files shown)


TallJimbo:
@timj , do you have any guesses as to why (after my update to the `mypy.ini` here, which I think makes sense), the GitHub action mypy check can see the `utils` types, but not the `resources` ones?

timj:
No. There is definitely a `py.typed` file there. Maybe I need to see if `pip install` into a new local conda env does install the `py.typed` file.

mypy.ini
```diff
@@ -39,10 +39,25 @@ ignore_missing_imports = True
...
 ignore_missing_imports = True
 ignore_errors = True
 
+# Do expect to import type annotations from LSST middleware dependencies that
+# should have them, but don't re-check them as they have their own mypy.ini
+# files with their own exceptions.
+
+[mypy-lsst.utils.*]
+ignore_missing_imports = False
+ignore_errors = True
```
timj, line 55: Why didn't we need this before? Did the previous block mean that we weren't testing utils/resources at all before? That can't be right because when I factored out the utility code I did get mypy errors.

python/lsst/daf/butler/core/storageClassDelegate.py
```diff
@@ -246,8 +246,8 @@ def getComponent(self, composite: Any, componentName: str) -> Any:
         return component
 
     def disassemble(
-        self, composite: Any, subset: Optional[Iterable] = None, override: bool = None
-    ) -> Dict[str, Any]:
+        self, composite: Any, subset: Optional[Iterable] = None, override: Optional[Any] = None
+    ) -> Optional[Dict[str, DatasetComponent]]:
```
timj: I can't understand how this ever returns `None`. `components` is always a dict and is returned as a dict unless an exception is raised.

requirements.txt
```diff
@@ -4,7 +4,7 @@ click >7.0
 sqlalchemy >= 1.3
 git+https://github.com/lsst/sphgeom@main#egg=lsst_sphgeom
 git+https://github.com/lsst/utils@main#egg=lsst_utils
-git+https://github.com/lsst/resources@main#egg=lsst_resources
+git+https://github.com/lsst/resources@tickets/DM-32250#egg=lsst_resources
```
timj: Remember to back this out.

setup.cfg
```diff
@@ -25,7 +25,7 @@ install_requires =
   click >= 7.0
   lsst_sphgeom @ git+https://github.com/lsst/sphgeom@main
   lsst_utils @ git+https://github.com/lsst/utils@main
-  lsst_resources @ git+https://github.com/lsst/resources@main
+  lsst_resources @ git+https://github.com/lsst/resources@tickets/DM-32250
```
timj: And this.

timj (approved):
I am confused by the disassemble method returning `None` but I'm sure that's me not reading the code properly.

mypy.ini
TallJimbo: It means that when `daf_butler` imported types from `utils` or `resources` before, they were all effectively seen as `Any` by MyPy, so how much it could tell us about our usage of them was pretty limited.

python/lsst/daf/butler/core/storageClassDelegate.py
TallJimbo: I'll try to remove the `Optional` and report back if I encounter a problem.  I don't recall how it got here.

TallJimbo: I remembered after looking more closely at the code: the (preexisting) docstring says:
> Returns None if this is not a composite `StorageClassDelegate.storageClass`.

I assume I should remove that sentence from the docs, then?

timj: I guess it changed at some point. It will seemingly always return an empty dict.

PR #648, opened 2022-02-22
DM-33784: Make minimal ButlerURI test

timj:
Duplicating all the ResourcePath tests does not give us anything.
We only need to test that we get a ButlerURI instance back.

## Checklist

- [ ] ran Jenkins


tests/test_uri.py
```diff
@@ -19,474 +19,32 @@
...
-        with ButlerURI.temporary_uri(prefix=tmpdir, suffix=".yaml") as tmp:
-            # Use a specified tmpdir and check it is okay for the file
-            # to not be created.
-            self.assertFalse(tmp.exists(), f"uri: {tmp}")
-        self.assertTrue(tmpdir.exists(), f"uri: {tmpdir} still exists")
+    def test_uri(self):
+        netloc = "netloc"
+        path = "/a/b/c.txt"
+        for scheme in ("file", "s3", "https"):
+            uri_str = f"{scheme}://{netloc}{path}"
```
mxk62, line 38: Just a little nitpick. Only `scheme` varies here and `netloc` and `path` are not used anywhere else. So I'm not sure if defining them is worth the trouble. On the other hand having them doesn't hurt much either so if you have reasons to keep them I won't object.

mxk62 (approved):

tests/test_uri.py
timj: Good point. I changed my mind half way through.

timj:
Typo in the ticket branch so will close and re-open.

PR #651, opened 2022-02-25
DM-33600: Improve error handling in registry query methods

andy-slac:
Several related changes, some small and some big:
- Fix handling of binds in analysis of user expression for governor values.
- Check that governor values specified in user expression really exist, raise exception if they are not. This is to make it consistent with handling of keyword arguments.
- Add `explain_no_result` to the result class returned from `queryDimensionRecords`.
- Change types of exceptions generated by registry to use more structured exception types, to make it easier to handle these exceptions on client side.

## Checklist

- [X] ran Jenkins


TallJimbo (approved):
Looks great!

My only quibble is with one of the exception additions: at least most (maybe all) of the direct raises of `CollectionError` (as opposed to one of its derived classes) seem like they might be better remaining as `TypeError`, or maybe a new exception like `NoDefaultProvided` that inherits from it - they're cases where an argument isn't being provided to a method, and no default was provided at `Registry` initialization.  To me that's on the spectrum of function-signature-mismatch logic bugs (which are reliably `TypeError` in Python), and the kind of thing that a more strongly-typed implementation might even statically reject (e.g. if `RegistryWithDefaults` was not the same class as `RegistryWithoutDefaults`), and not terribly related to the `CollectionError` subclasses.

andy-slac:
I think I'll add `NoDefaultCollectionError` for the case of missing default collection. I'd like to avoid using very common exception types because handling those reliably is not really possible. Getting exception hierarchy right is a tough problem, lake many other things related to error handling.

I also realized that I probably need to expose some exception classes at the `butler` module level, right now they are only visible in `registry` module. As `Registry` API is already more or less public, I think we should make documented exception public as well.

andy-slac:
@TallJimbo, could you check 12540ac, I have added a small paragraph documenting exceptions and explain_no_results?

doc/lsst.daf.butler/queries.rst
```diff
@@ -410,3 +410,18 @@ Example of use of these two methods:
     # Print ten latest visit records in reverse time order
     for record in registry.queryDimensionRecords("visit").order_by("-timespan.begin").limit(10):
         print(record)
+
+.. _daf_butler_query_error_handling:
+
+Error handling with Registry methods
+------------------------------------
+
+`Registry` methods typically raise exceptions when they detect problems with input parameters. Documentation for these methods describes a set of exception classes and conditions in which exceptions are generated. In most cases, these exceptions belong to one of the special exception classes defined in `lsst.daf.butler.registry` module, e.g. `~lsst.daf.butler.registry.DataIdError`, which have `~lsst.daf.butler.registry.RegistryError` as a common base class. These exception classes are not exposed in `lsst.daf.butler` module interface, to use these classes they need to be imported explicitly, e.g.:
```
TallJimbo: I believe our recommended steal is to add newlines after each sentence:

```suggestion
`Registry` methods typically raise exceptions when they detect problems with input parameters.
Documentation for these methods describes a set of exception classes and conditions in which exceptions are generated.
In most cases, these exceptions belong to one of the special exception classes defined in `lsst.daf.butler.registry` module, e.g. `~lsst.daf.butler.registry.DataIdError`, which have `~lsst.daf.butler.registry.RegistryError` as a common base class.
These exception classes are not exposed by the `lsst.daf.butler` module interface; to use these classes they need to be imported explicitly, e.g.:
```

(also includes a few tiny grammatical fixes)

doc/lsst.daf.butler/queries.rst
```diff
@@ -410,3 +410,18 @@ Example of use of these two methods:
...
+
+.. code-block:: Python
+
+    from lsst.daf.butler.registry import DataIdError, UserExpressionError
+
+While class documentation should list most commonly produced exception, there may be other exceptions raised by its methods. Code that needs to handle all types of exceptions generated by `Registry` methods should be prepared to handle other types of exceptions as well.
```
TallJimbo:
```suggestion
While class documentation should list most commonly produced exceptions, there may be other exceptions raised by its methods.
Code that needs to handle all types of exceptions generated by `Registry` methods should be prepared to handle other types of exceptions as well.
```

doc/lsst.daf.butler/queries.rst
```diff
@@ -410,3 +410,18 @@ Example of use of these two methods:
...
+
+Few of the `Registry` query methods (`~Registry.queryDataIds`, `~Registry.queryDatasets`, and `~Registry.queryDimensionRecords`) return result objects. These objects are iterables of the corresponding record types and typically they represent a non-empty result set. In some cases these methods can return empty results without generating an exception, for example due to a combination of constraints excluding all existing records. Result classes implement ``explain_no_results()`` method which can be used to try to identify the reason for an empty result. It returns a list of strings, each string contains human-readable message describing the reason for an empty result. This method does not always work reliably and can return an empty list even when result is empty. In particular it cannot analyze user expression and identify which part of that expression is responsible for an empty result.
```
TallJimbo:
```suggestion
A few of the `Registry` query methods (`~Registry.queryDataIds`, `~Registry.queryDatasets`, and `~Registry.queryDimensionRecords`) return result objects.
These objects are iterables of the corresponding record types and typically they represent a non-empty result set.
In some cases these methods can return empty results without generating an exception, for example due to a combination of constraints excluding all existing records.
Result classes implement ``explain_no_results()`` method which can be used to try to identify the reason for an empty result.
It returns a list of strings, with each string a human-readable message describing the reason for an empty result.
This method does not always work reliably and can return an empty list even when result is empty.
In particular it cannot analyze user expression and identify which part of that expression is responsible for an empty result.
```

TallJimbo (approved):
I've made added some patch recommendations with newlines between sentences and very small grammatical fixes.

PR #655, opened 2022-03-02
DM-33890: Fix certifyCalibrations crash due to exception type change

andy-slac:
Replaced exception catching with a collection type cheeck.

## Checklist

- [ ] ran Jenkins


timj (approved):
Not spotting this tells us something about our test coverage of course.

andy-slac:
> Not spotting this tells us something about our test coverage of course.

It's hard to write tests for 100% coverage with that level of complexity in the code. I'm worried there may be other cases like this, I did some grepping but it's hard to prove things that way.

PR #657, opened 2022-03-03
DM-33904: Fix bug in Registry.getCollectionParentChains

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/interfaces/_collections.py
+++ b/python/lsst/daf/butler/registry/interfaces/_collections.py
@@ -206,7 +206,13 @@ class ChainedCollectionRecord(CollectionRecord):
             The object that manages this records instance and all records
             instances that may appear as its children.
         """
+        # Clear out the old reverse mapping (from child to parents).
+        for child in self._children:
+            manager._parents_by_child[manager.find(child).key].discard(self.key)
         self._children = self._load(manager)
+        # Update the reverse mapping (from child to parents) in the manager.
+        for child in self._children:
+            manager._parents_by_child[manager.find(child).key].add(self.key)
 
     @abstractmethod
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/registry/interfaces/_collections.py
```diff
@@ -207,6 +207,9 @@ def refresh(self, manager: CollectionManager) -> None:
             instances that may appear as its children.
         """
         self._children = self._load(manager)
+        # Update the reverse mapping (from child to parents) in the manager.
+        for child in self._children:
+            manager._parents_by_child[manager.find(child).key].add(self.key)
```
andy-slac, line 215: What happens if a child is removed from a chain, should this code remove all previously known children of this chain from reverse mapping first?

python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -114,8 +114,22 @@ def makeRegistryConfig(self) -> RegistryConfig:
         return config
 
     @abstractmethod
-    def makeRegistry(self) -> Registry:
-        """Return the Registry instance to be tested."""
+    def makeRegistry(self, share_repo_with: Optional[Registry] = None) -> Optional[Registry]:
+        """Return the Registry instance to be tested.]
```
andy-slac: Extra `]`

andy-slac (approved):
Looks good, one question.

python/lsst/daf/butler/registry/interfaces/_collections.py
TallJimbo: Good catch, fixed.

PR #660, opened 2022-03-10
DM-33999: Use unicode in run names for some tests

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/utils.py
+++ b/python/lsst/daf/butler/core/utils.py
@@ -103,7 +103,8 @@ def globToRegex(
     if not expressions or "*" in expressions:
         return Ellipsis
 
-    nomagic = re.compile(r"^[\w/\.\-@]+$")
+    # The character class range at the end adds support for emoji.
+    nomagic = re.compile(r"^[\w/\.\-@\u263a-\U0001f645]+$")
 
     # Try not to convert simple string to a regex.
     results: List[Union[str, Pattern]] = []
--- a/python/lsst/daf/butler/core/utils.py
+++ b/python/lsst/daf/butler/core/utils.py
@@ -103,7 +103,7 @@ def globToRegex(
     if not expressions or "*" in expressions:
         return Ellipsis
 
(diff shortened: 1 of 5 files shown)


python/lsst/daf/butler/core/utils.py
```diff
@@ -103,7 +103,8 @@ def globToRegex(
     if not expressions or "*" in expressions:
         return Ellipsis
 
-    nomagic = re.compile(r"^[\w/\.\-@]+$", re.ASCII)
+    # The character class range at the end adds support for emoji.
+    nomagic = re.compile(r"^[\w/\.\-@\u263a-\U0001f645]+$")
```
timj: This is a little brute force in terms of character ranges. There are more limited ranges that can be supported for emoticons and flags etc rather than allowing all emoji.

ktlim: Seems to me that you should give up on trying to identify legal characters and instead merely detect whether there are glob characters that need `fnmatch`:
```
nomagic = re.compile(r"^[^*?[\]]*$")
```

ktlim (approved):

python/lsst/daf/butler/core/utils.py
timj: Okay. That is probably better. This is specifically designed for `fnmatch` -- I wasn't entirely sure of the full list but I'll assume the above is correct.

ktlim: I took the list from https://docs.python.org/3/library/fnmatch.html

ktlim: In fact `fnmatch` doesn't seem to treat a `]` as special unless it's paired with `[`, so you could leave the `\]` out.

ktlim: Oh, and you could invert the logic then, making it just an RE to look for one of the three characters (`*?[`) anywhere in the string.

timj: Yes, I'm inverting.

PR #661, opened 2022-03-15
DM-33619: Add unlink option to butler remove-runs

n8pease:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/cli/cmd/_remove_runs.py
+++ b/python/lsst/daf/butler/cli/cmd/_remove_runs.py
@@ -27,7 +27,7 @@ import click
 from ... import script
 from ..opt import collection_argument, confirm_option, options_file_option, repo_argument
 from ..utils import ButlerCommand
-
+from .commands import existingRepoHelp
 
 # messages emitted by remove-runs, defined separately for use in unit
 # tests.
@@ -83,8 +83,11 @@ def _print_requires_confirmation(runs: Sequence[script.RemoveRun], datasets: Map
 
 
 @click.command(cls=ButlerCommand)
-@repo_argument(required=True)
 @click.pass_context
+@repo_argument(
+    help=existingRepoHelp,
+    required=True,
+)
 @collection_argument(
     help="COLLECTION is a glob-style expression that identifies the RUN collection(s) to remove."
 )
--- a/python/lsst/daf/butler/cli/cmd/_remove_runs.py
+++ b/python/lsst/daf/butler/cli/cmd/_remove_runs.py
@@ -20,13 +20,14 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 
-from typing import Mapping, Sequence
+from typing import Iterator, Mapping, Sequence
 
 import click
 
 from ... import script
 from ..opt import collection_argument, confirm_option, options_file_option, repo_argument
 from ..utils import ButlerCommand
+
 
 # messages emitted by remove-runs, defined separately for use in unit
 # tests.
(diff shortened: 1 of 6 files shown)


python/lsst/daf/butler/cli/cmd/_remove_runs.py
```diff
@@ -85,5 +118,11 @@ def remove_runs(**kwargs):
         else:
             print(abortedMsg)
     else:
+        # if the user opted out of confirmation but there are runs with
+        # parent collections then they must confirm; print a message
+        # and exit.
+        if any(run.parents for run in result.runs) and not force:
+            _print_requires_confirmation(result.runs, result.datasets)
+            sys.exit(1)
```
timj: Is Click okay with calling `sys.exit()`? I see there is a `click.Context.exit()`.

python/lsst/daf/butler/cli/utils.py
```diff
@@ -196,6 +196,9 @@ def addArgumentHelp(doc, helpText):
         # (Lines after the first blank line will be argument help.)
         while len(doclines) > 1 and doclines[1]:
             doclines[0] = " ".join((doclines[0], doclines.pop(1).strip()))
+        # Add standard indent to help text for proper alignment with command
+        # function documentation:
+        helpText = "    " + helpText
```
timj, line 201: Word wrapping in click does seem harder than it should be...

python/lsst/daf/butler/script/removeRuns.py
```diff
@@ -113,10 +124,15 @@ def removeRuns(
     """
     runs, datasets = _getCollectionInfo(repo, collection)
 
-    def doRemove(runs: Sequence[str]) -> None:
+    def doRemove(runs: Sequence[RemoveRun]) -> None:
         """Perform the remove step."""
         butler = Butler(repo, writeable=True)
-        butler.removeRuns(runs, unstore=True)
+        for run in runs:
```
timj: Should this entire loop be in a transaction in case something goes wrong before the removeRuns happens and you may not want some of the chains to have been modified and some unchanged?

tests/test_cliCmdRemoveRuns.py
```diff
@@ -83,8 +95,25 @@ def test_removeRuns(self):
             root = "repo1"
             MetricTestRepo(root, configFile=os.path.join(TESTDIR, "config/basic/butler.yaml"))
 
-            # Execute cmd but say no, check for expected outputs.
+            # add the run to a CHAINED collection
```
timj: As a general point comments should be sentences so this should be:

```python
# Add the run to a CHAINED collection.
```

python/lsst/daf/butler/cli/cmd/_remove_runs.py
```diff
@@ -52,25 +66,44 @@ def _print_remove(will: bool, runs: Sequence[str], datasets: Mapping[str, int]):
...
     """
     print(willRemoveRunsMsg if will else didRemoveRunsMsg)
-    print(", ".join(runs))
-    print(willRemoveDatasetsMsg if will else didRemoveDatasetsMsg)
+    unlinkMsg = willUnlinkMsg if will else didUnlinkMsg
+    for run in runs:
+        if run.parents:
+            print(unlinkMsg.format(run=run.name, parents=", ".join(_quoted(run.parents))))
+        else:
+            print(run.name)
```
timj, line 73: Does the coverage report indicating that this line never runs mean that we aren't testing the command without parents?

tests/test_cliCmdRemoveRuns.py
```diff
@@ -83,8 +95,25 @@ def test_removeRuns(self):
...
             result = self.runner.invoke(butler.cli, ["remove-runs", root, "ingest*", "--no-confirm"])
+            self.assertNotEqual(result.exit_code, 0, clickResultMsg(result))
+            self.assertIn(
+                mustBeUnlinkedMsg.format(run="ingest/run", parents=f'"{parentCollection}"'), result.output
+            )
+
+            # Execute cmd with --force
+            result = self.runner.invoke(
+                butler.cli, ["remove-runs", root, "ingest*", "--no-confirm", "--force"]
+            )
```
timj, line 147: Can we also run `butler query-collections` for the run to show it has disappeared?

timj (approved):
Looks good. Very minor comments.

python/lsst/daf/butler/script/removeRuns.py
n8pease: that seems like a good idea. I think to do that I put the loop inside a block that begins with `with butler.transaction():`, and that's all, is that right?

python/lsst/daf/butler/cli/utils.py
n8pease: This is partially my fault, for modifying how click.argument help is inserted; click says you should add it yourself in the command function doctoring, but I wanted to reuse the help text along with the argument declaration via the argument decorator. I'm kind of fighting the API. If it's a continuous PITA then probably I should back out what I've done & let the wookie win.

python/lsst/daf/butler/cli/cmd/_remove_runs.py
n8pease: yes, I changed the tests for CHAINED collections, and didn't consider this case. Will fix.

PR #664, opened 2022-03-21
DM-33490: Implement export_records/import_records methods

andy-slac:
These methods are not useful for the in-memory datastore and their
implementation does nothing. FileDatastore has a working implementation
of both methods. ChainedDatastore methods will work if it is backed by
FileDatastores.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/datastore.py
+++ b/python/lsst/daf/butler/core/datastore.py
@@ -62,6 +62,7 @@ if TYPE_CHECKING:
     from .configSupport import LookupKey
     from .datasets import DatasetRef, DatasetType
     from .storageClass import StorageClass
+    from .storedFileInfo import StoredDatastoreItemInfo
 
 
 class DatastoreConfig(ConfigSubset):
@@ -80,17 +81,18 @@ class DatastoreValidationError(ValidationError):
 
 @dataclasses.dataclass
 class DatastoreRecordData:
-    """A struct that represents a tabular data export from one or more
-    datastores.
+    """A struct that represents a tabular data export from a single
+    datastore.
     """
 
-    locations: Dict[str, List[DatasetIdRef]] = dataclasses.field(default_factory=lambda: defaultdict(list))
-    """Mapping from datastore name to the datasets in that datastore.
+    refs: List[DatasetIdRef] = dataclasses.field(default_factory=list)
+    """List of DatasetRefs known to this datastore.
     """
 
-    records: Dict[str, List[Dict[str, Any]]] = dataclasses.field(default_factory=lambda: defaultdict(list))
-    """Opaque table data that backs one or more datastores, grouped by
-    opaque table name.
+    records: Dict[str, List[StoredDatastoreItemInfo]] = dataclasses.field(
+        default_factory=lambda: defaultdict(list)
+    )
+    """Opaque table data, grouped by opaque table name.
     """
 
 
@@ -1035,20 +1037,20 @@ class Datastore(metaclass=ABCMeta):
         """
         return True
 
-    # TODO: make abstract, implement in all concrete datastores
+    @abstractmethod
     def import_records(
         self,
-        data: DatastoreRecordData,
+        data: Mapping[str, DatastoreRecordData],
     ) -> None:
         """Import datastore location and record data from an in-memory data
         structure.
 
         Parameters
         ----------
-        data : `DatastoreRecordData`
-            Data structure to load from.  May contain data for other
-            `Datastore` instances (generally because they are chained to this
-            one), which should be ignored.
+        data : `Mapping` [ `str`, `DatastoreRecordData` ]
+            Datastore records indexed by datastore name.  May contain data for
+            other `Datastore` instances (generally because they are chained to
+            this one), which should be ignored.
 
         Notes
         -----
@@ -1063,12 +1065,12 @@ class Datastore(metaclass=ABCMeta):
         """
         raise NotImplementedError()
 
-    # TODO: make abstract, implement in all concrete datastores
+    @abstractmethod
     def export_records(
         self,
         refs: Iterable[DatasetIdRef],
-    ) -> DatastoreRecordData:
-        """Export datastore records and locations from an in-memory data
+    ) -> Mapping[str, DatastoreRecordData]:
+        """Export datastore records and locations to an in-memory data
         structure.
 
         Parameters
@@ -1079,7 +1081,7 @@ class Datastore(metaclass=ABCMeta):
 
         Returns
         -------
-        data : `DatastoreRecordData`
-            Populated data structure.
+        data : `Mapping` [ `str`, `DatastoreRecordData` ]
+            Exported datastore records indexed by datastore name.
         """
         raise NotImplementedError()
--- a/python/lsst/daf/butler/core/storedFileInfo.py
+++ b/python/lsst/daf/butler/core/storedFileInfo.py
@@ -34,7 +34,7 @@ from .location import Location, LocationFactory
 from .storageClass import StorageClass, StorageClassFactory
 
 if TYPE_CHECKING:
-    from .datasets import DatasetRef
+    from .datasets import DatasetId, DatasetRef
 
 # String to use when a Python None is encountered
 NULLSTR = "__NULL_STRING__"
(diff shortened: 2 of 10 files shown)


python/lsst/daf/butler/core/datastore.py
```diff
@@ -80,17 +81,18 @@ class DatastoreValidationError(ValidationError):
...
 class DatastoreRecordData:
-    """A struct that represents a tabular data export from one or more
-    datastores.
+    """A struct that represents a tabular data export from a single
+    datastore.
     """
 
-    locations: Dict[str, List[DatasetIdRef]] = dataclasses.field(default_factory=lambda: defaultdict(list))
-    """Mapping from datastore name to the datasets in that datastore.
+    refs: List[DatasetRef] = dataclasses.field(default_factory=list)
```
TallJimbo: Using `DatasetIdRef` rather than `DatasetRef` here was intentional, because I don't think we need anything but the UUID, and it'd be wastefully duplicatively to store more than the UUID in the QuantumGraph or per-quantum files.

(I'd love to get rid of `DatasetIdRef` and just have `Datastore` methods that really only need the UUID only take the UUID, but I think that's out of scope for this ticket.)

python/lsst/daf/butler/core/storedFileInfo.py
```diff
@@ -97,6 +97,7 @@ def __init__(
         component: Optional[str],
         checksum: Optional[str],
         file_size: int,
+        dataset_id: Optional[DatasetId] = None,
```
TallJimbo: Mind opening a ticket to remove the `Optional` and the default here (assuming @timj agrees)?  I think it'd be better to always have it and not need it than having to worry in any particular context whether it's been provided.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -2689,3 +2695,33 @@ def needs_expanded_data_ids(
         # dataset type, but that's not necessary for correctness; it just
         # enables more optimizations (perhaps only in theory).
         return transfer not in ("direct", None)
+
+    def import_records(self, data: Mapping[str, DatastoreRecordData]) -> None:
+        # Docstring inherited from the base class.
+        record_data = data[self.name]
+        if record_data:
```
TallJimbo: Should this be `if record_data.refs`?  And if that's false, shouldn't we short-circuit quit (or maybe assert that `record_data.records` is empty, too)?

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -2689,3 +2695,33 @@ def needs_expanded_data_ids(
...
+            for info in records:
+                assert isinstance(info, StoredFileInfo), "Expecting StoredFileInfo records"
+                unpacked_records.append(info.to_record())
+            self._table.insert(*unpacked_records)
+
+    def export_records(self, refs: Iterable[DatasetRef]) -> Mapping[str, DatastoreRecordData]:
+        # Docstring inherited from the base class.
+        exported_refs = []
+        for ref in refs:
+            datastores = self._bridgeManager.findDatastores(ref)
```
TallJimbo: I hadn't appreciated that we'd need to call this per dataset in order to implement export; we should at least create a new ticket to add a bridge method that would let us vectorize this somehow.

An alternative might be to skip this check entirely, and rely on the `fetch` call later to only return rows whose IDs are present; I think that will yield the same results.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -2689,3 +2695,33 @@ def needs_expanded_data_ids(
...
+            if self.name in datastores:
+                exported_refs.append(ref)
+
+        id2ref = {ref.id: ref for ref in refs}
+        rows = self._table.fetch(dataset_id=list(id2ref.keys()))
```
TallJimbo, line 2722: I think this will result in rows for all components being exported when any component or the parent are referenced, since they all have the `dataset_id`.  I think that's fine, even if we only care about one component, and it'll be simpler this way, but we should keep in mind that that's how it works.

python/lsst/daf/butler/datastores/inMemoryDatastore.py
```diff
@@ -642,3 +642,13 @@ def needs_expanded_data_ids(
     ) -> bool:
         # Docstring inherited.
         return False
+
+    def import_records(self, data: Mapping[str, DatastoreRecordData]) -> None:
+        # Docstring inherited from the base class.
+        return
```
TallJimbo, line 653: I believe the commit message still says that this raises.

TallJimbo (approved):
Looks good!  Left a number of minor comments; the only big one is probably the `findDatastores` many-small-queries performance hit that's looming, but that could be punted to another ticket.

One other small aside is that we should probably think about the data structure here one more time with an eye towards exporting all records from a full QG at once, or maybe per-dataset-type, to maximize vectorization of queries, and then split the results up among quanta in Python.  I think that's at least _okay_ with what we have here - we'll have to reorganize all of the datasets to be in mappings keyed by dataset ID instead of by datastore name, but since we need to transform everything that way (rather than just pull out some records), that's fine.  Does that make sense to everyone?

python/lsst/daf/butler/core/storedFileInfo.py
timj: I suggested optional because I wasn't really sure what would break and need fixing if we made it mandatory. Datastore did not need it itself because of the way it's used but maybe it now always gets it so it no longer makes a difference?

python/lsst/daf/butler/core/datastore.py
andy-slac: I think there was some API that did not accept `DatasetIdRef`, but I will re-check.

python/lsst/daf/butler/core/storedFileInfo.py
andy-slac: I can do it here, should be a trivial change.
I'm still not sure what is the right thing to do with `to_record()` method. @timj, should we drop `ref` argument, or is it useful to override component from that ref?

python/lsst/daf/butler/datastores/fileDatastore.py
andy-slac: Hmm, I should probably change `data[]` to `data.get()` to allow a case when nothing is passed for a specific datastore.

python/lsst/daf/butler/core/storedFileInfo.py
timj: I guess we drop it from to_record now that it's inherently part of the record.

timj: I'm sure you will find out if the dataset_id  really has to be optional or not.

python/lsst/daf/butler/datastores/fileDatastore.py
andy-slac: I was worried about the case when a ref may be in dataset_locations table but have no records associated to its dataset_id. This is probably not the case today, though.

python/lsst/daf/butler/datastores/inMemoryDatastore.py
andy-slac: Not updated after re-implementing, will fix it.

python/lsst/daf/butler/core/datastore.py
andy-slac: Not quite that thing, but `BridgeManager.findDatastores()` only accepts true DatasetRef, which forced me to change `export_records` to accept DatasetRefs as well, and that led to changing this structure (though it was not really needed here). Maybe the right thing to do is to change `findDatastores()` to accept DatasetIdRef (or DatasetId)? Bridge API actually uses DatasetRefs extensively, not sure it will be an easy change.

andy-slac: I changed `findDatastores()` to accept `DatasetIdRef`, it did not trigger any changes elsewhere.

python/lsst/daf/butler/core/storedFileInfo.py
andy-slac: > I guess we drop it from to_record now that it's inherently part of the record.

I did not manage to remove argument from `to_record`, there is a code in ingest which reuses the same record instance with multiple component refs. I do not want to touch that code yet, so I'll keep `to_record` unchanged for now.

(sorry for slow progress, there are usual emergencies in my other project)

python/lsst/daf/butler/datastores/fileDatastore.py
andy-slac: I think I had a wrong idea about how this method is supposed to work, the correct thing to do here should be to use bridge `check()` method, and that is already vectorized.

andy-slac:
> I think that's at least _okay_ with what we have here - we'll have to reorganize all of the datasets to be in mappings keyed by dataset ID instead of by datastore name

I think the current structure is built for somewhat different purpose, but we still can transform it into something indexed by DatasetRef. For that sort of splitting it may be easier to start with a structure that is already indexed by dataset ID. It will probably become clearer when we start using this more extensively.

timj:
So long as we all remember that one DatasetRef can refer to multiple datastore records.

PR #665, opened 2022-03-22
DM-33902 Long chained collections print badly via CLI

n8pease:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/script/queryCollections.py
+++ b/python/lsst/daf/butler/script/queryCollections.py
@@ -22,14 +22,22 @@
 from __future__ import annotations
 
 import itertools
+from typing import Iterable, Union
 
 from astropy.table import Table
 from numpy import array
 
 from .._butler import Butler
+from ..registry import CollectionType
 
 
-def queryCollections(repo, glob, collection_type, chains):
+
+def queryCollections(
+    repo: str,
+    glob: Union[ellipsis, Iterable[str]],
+    collection_type: Iterable[CollectionType],
+    chains: str,
+) -> Table:
     """Get the collections whose names match an expression.
 
     Parameters
@@ -37,7 +45,7 @@ def queryCollections(repo, glob, collection_type, chains):
     repo : `str`
         URI to the location of the repo or URI to a config file describing the
         repo and its location.
-    glob : iterable [`str`]
+    glob : `Iterable` [`str`] or `union`
         A list of glob-style search string that fully or partially identify
         the dataset type names to search for.
     collection_type : `Iterable` [ `CollectionType` ], optional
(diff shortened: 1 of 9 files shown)


n8pease:
are there unit tests for lsst.daf.butler.remote? That codcov miss is not so great.

python/lsst/daf/butler/script/queryCollections.py
```diff
@@ -22,22 +22,30 @@
...
 
 from .._butler import Butler
+from ..registry import CollectionType
 
 
-def queryCollections(repo, glob, collection_type, chains):
+
+def queryCollections(
+    repo: str,
+    glob: Union[ellipsis, Iterable[str]],
```
TallJimbo: To get MyPy to handle `Ellipsis` properly I've always had to import the workaround [here](https://github.com/lsst/daf_butler/blob/2036ad53e56a84472a511b129f493549de86bdfd/python/lsst/daf/butler/registry/wildcards.py#L53-L75).  I suspect this means that this script isn't being type checked anyway; if you want to enable that, there's probably an ignore pattern in `mypy.ini` where an exception needs to be made.  In any case, I'd recommend just annotating this as `typing.Any`; the set of types accepted as collection expressions is actually large enough that I don't think it's worth spelling out for static checking.

python/lsst/daf/butler/script/queryCollections.py
```diff
@@ -21,20 +21,160 @@
...
+                    first = False
+            else:
+                table.add_row((name, type, ""))
+        # If none of the datasets has a parent dataset then remove the
+        # description column.
+        if not any(c for c in table[descriptionCol]):
+            del table[descriptionCol]
+    else:
+        for name in sorted(names):
+            type = butler.registry.getCollectionType(name).name
```
TallJimbo: Instead of converting the enum object to a string here, I'd recommend leaving it, using `if type is CollectionType.CHAINED` in the `if` block on the next line, and then using `.name` only when you need a string (e.g. in `table.add_row`).

python/lsst/daf/butler/script/queryCollections.py
```diff
@@ -21,20 +21,160 @@
...
+        Same as `queryCollections`
+    """
+    table = Table(
+        names=("Name", "Type"),
+        dtype=(str, str),
+    )
+    butler = Butler(repo)
+
+    def addCollection(name: str, level: int = 0) -> None:
+        collectionType = butler.registry.getCollectionType(name).name
```
TallJimbo: Same comment about `enum`->`str` conversion as above.

python/lsst/daf/butler/script/queryCollections.py
```diff
@@ -54,89 +194,19 @@ def queryCollections(
         Must be one of "FLATTEN", "TABLE", or "TREE" (case sensitive).
         Affects contents and formatting of results, see
         ``cli.commands.query_collections``.
+    inverse : `bool`
+        If true, show what CHAINED datasets the dataset is a member of
+        (instead of what datasets any CHAINED datasets contain)
```
TallJimbo: "datasets" -> "collections" in the two lines above.

TallJimbo (approved):

PR #666, opened 2022-03-23
DM-33769-v23: manual back port of lsst.resources tickets/DM-33769

airnandez:
This is v23 implementation of lsst/resources#7.

## Checklist

- [x] ran Jenkins

This is a manual back port of `tickets/DM-33769` of `lsst.resources` (see [pull request](https://github.com/lsst/resources/pull/7)).


airnandez:
Although all tests in my local environment pass (using `lsst-scipipe-0.8.0`), Jenkins is failing with any of the environments `lsst-scipipe-0.8.0`, `lsst-scipipe-0.2.0` and `lsst-scipipe-0.3.0`.

I am surely missing something and I don't manage to see what it may be. Any guidance would be appreciated.

timj:
I've run up a 23.0.1.rc4 session on the IDF on data-int and the S3 tests do pass. That seems to be using 0.8.0 rubin-env with moto (2.3.2), boto3 (1.20.46), botocore (1.23.46).

This compares with the versions in the action (2.3.2, 1.21.24, 1.24.24) [which fails] and in Jenkins (2.3.2, 1.21.24, 1.24.24) [which also fails].

At a guess we might want to try with `boto3<1.21` and `botocore<1.24` in the requirements for this branch. If that works we will have to change the rubin-env to include the pins.  cc/ @ktlim

ktlim:
I'm a bit more suspicious of a moto 2.3.2/responses 0.20.0 incompatibility, since all the failures seem to be in the latter package (or its use of jinja2), and responses does seem to deprecate in minor versions (since they have no major ones).  Current rubin-env uses moto 3.x, which seems fine with older or newer responses.

airnandez:
Thank you both for your inputs. With the constraint below in `requirements.txt` the tests run in GitHub actions pass:

```
responses >= 0.12.0, <= 0.18.0
```

(`responses 0.18.0` is the release included in `lsst-scipipe-0.8.0`).

However, even with this modification, tests executed by Jenkins (with env `0.8.0`) work on centOS but fail on macOS (see [run 36206](https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/36206/pipeline)).

airnandez:
Adding the constraint below to `requirements.txt`:

```
jinja2 <= 3.0.3
```

all the tests run by GitHub actions pass. Tests run by Jenkins (with env `0.8.0`) pass on centOS but fail on macOS (see [run 36207](https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/36207/pipeline/)).

(`jinja2 3.0.3` is the release included in `lsst-scipipe-0.8.0`).

ktlim:
Strange: those macOS runs both have responses=0.20.0, even though the CentOS runs have responses=0.18.0.

This appears to be because a generic rubin-env 0.8.0 environment is used, and the pre-existing environments were built at different times.  Changes to `requirements.txt` have no effect on Jenkins.

airnandez:
> Changes to requirements.txt have no effect on Jenkins.

Thanks for this clarification: that was not obvious to me.

I reverted the constraint `jinja2 <= 3.0.3` since it does not seem needed for tests in GitHub actions to pass.

timj:
@airnandez thanks for working out the constraints. Please squash the experimental commits and have only one commit that changes the version constraints. This all means that we will have to do a patch to the 0.8.0 rubin-env before we can do a new build but that should not stop the merging of this branch (s3 tests are already broken).

requirements.txt
```diff
@@ -16,7 +16,7 @@ pandas >= 1.0
 numpy >= 1.17
 matplotlib >= 3.0.3
 pyarrow >= 0.16
-responses >= 0.12.0
+responses >= 0.12.0, <= 0.18.0
```
timj, line 19: I think we may also need to change setup.cfg

airnandez: Commits squashed and contraints of `responses` included in `setup.cfg`.

All GitHub actions checks pass.

timj (approved):
Looks okay to me. Thanks.

PR #669, opened 2022-03-31
DM-34261: respond to click.Path api change

n8pease:
## Checklist

- [ ] ran Jenkins


doc/changes/DM-34261.bugfix.rst
```diff
@@ -0,0 +1 @@
+The `click.Path` API changed, change from ordered arguments to keyword arguments when calling it.
```
timj, line 1: I think they may say that they always were expecting kwargs to be used. It's just they did not enforce that interface.

timj (approved):

PR #670, opened 2022-03-31
DM-34247: simplify dataset subquery logic and fix edge-case bugs

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
+++ b/python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
@@ -405,8 +405,8 @@ class ByDimensionsDatasetRecordStorage(DatasetRecordStorage):
             # generate a valid SQL query that can't yield results.  This should
             # never get executed, but lots of downstream code will still try
             # to access the SQLAlchemy objects representing the columns in the
-            # subquery.  That's not idea, but it'd take a lot of refactoring to
-            # fix it.
+            # subquery.  That's not ideal, but it'd take a lot of refactoring
+            # to fix it (DM-31725).
             query.where.append(sqlalchemy.sql.literal(False))
         else:
             query.where.append(collection_col.in_([collection.key for collection in collections]))
--- a/python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
+++ b/python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 __all__ = ("ByDimensionsDatasetRecordStorage",)
 
 import uuid
-from typing import TYPE_CHECKING, Any, Dict, Iterable, Iterator, List, Optional, Set, Tuple
+from typing import TYPE_CHECKING, Any, Dict, Iterable, Iterator, List, Optional, Sequence, Set, Tuple
 
 import sqlalchemy
 from lsst.daf.butler import (
@@ -77,7 +77,6 @@ class ByDimensionsDatasetRecordStorage(DatasetRecordStorage):
         sql = self.select(
             collection, dataId=dataId, id=SimpleQuery.Select, run=SimpleQuery.Select, timespan=timespan
         )
-        sql = sql.combine()
         results = self._db.query(sql)
         row = results.fetchone()
         if row is None:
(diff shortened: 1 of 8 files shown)


python/lsst/daf/butler/registry/queries/_builder.py
```diff
@@ -224,7 +216,7 @@ def joinDataset(
                             f"collection '{collectionRecord.name}' is not yet supported."
                         )
                     else:
-                        calibration_collections.append(collectionRecord)
+                        collectionRecords.append(collectionRecord)
                 else:
                     # We can never find a non-calibration dataset in a
                     # CALIBRATION collection.
```
andy-slac, line 243: Re-reading pre-existing code - I think this comment is not quite right, in this `else` branch dataset type can still be calibration (if `collectionRecord.name` is not in `explicitCollections`).

python/lsst/daf/butler/registry/queries/_builder.py
```diff
@@ -233,141 +225,225 @@ def joinDataset(
...
+        return (
+            storage.select(
+                *collections,
                 dataId=SimpleQuery.Select,
-                id=SimpleQuery.Select if isResult else None,
-                run=SimpleQuery.Select if isResult else None,
-                ingestDate=SimpleQuery.Select if isResult else None,
+                # If this dataset type has no dimensions, we're in danger of
+                # generating an invalid subquery that has no columns in the
+                # SELECT clause.  Any easy fix is to just select some arbitrary
```
andy-slac: Any -> An?

andy-slac (approved):
Looks great, a couple of minor comments.

python/lsst/daf/butler/registry/queries/_builder.py
TallJimbo: Yes, I believe this was an edge-case bug that is now fixed, and the comment is now correct.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
```diff
@@ -387,15 +387,10 @@ def select(
             query.join(self._calibs, **kwargs)
             dataset_id_col = self._calibs.columns.dataset_id
             collection_col = self._calibs.columns[self._collections.getCollectionForeignKeyName()]
-        elif CollectionType.CALIBRATION not in collection_types:
+        elif CollectionType.CALIBRATION not in collection_types or timespan is None:
             query.join(self._tags, **kwargs)
             dataset_id_col = self._tags.columns.dataset_id
             collection_col = self._tags.columns[self._collections.getCollectionForeignKeyName()]
-        else:
```
andy-slac, line 394: So, the `else` case cannot happen at all?

python/lsst/daf/butler/registry/queries/_builder.py
```diff
@@ -362,8 +361,9 @@ def _build_dataset_search_subquery(
             collection key, and the ingest date.
         """
         # Query-simplification shortcut: if there is only one collection, a
-        # find-first search is just a regular result subquery.
-        if len(collections) == 1:
+        # find-first search is just a regular result subquery.  Same is true
+        # if this is a doomed query with no collectoins to search.
```
andy-slac: Typo: collectoins

andy-slac (approved):
Looks good, minor comments.

python/lsst/daf/butler/registry/queries/_builder.py
TallJimbo: Oh, nevermind; this wasn't fixed - I had started, and then decided to the refactor first, and then forgot to come back to it.  Fixed on another commit.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
TallJimbo: Looking more closely sent me down a rabbit hole - it's much worse, because this new commit doesn't work at all (because calibration membership is stored in a totally different table), and that leaves a problem (failure to generate queries that include both CALIBRATION and other collections, when a dataset type is present in both categories) introduced by the earlier refactor on this ticket still around.

So I've now backed out this commit and added a new one with a different approach.

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -511,6 +514,7 @@ def transaction(
             # `Connection.in_nested_transaction()` method.
             savepoint = savepoint or connection.info.get(_IN_SAVEPOINT_TRANSACTION, False)
             connection.info[_IN_SAVEPOINT_TRANSACTION] = savepoint
+            trans: Union[sqlalchemy.engine.Transaction, sqlalchemy.engine.NestedTransaction]
```
andy-slac: NestedTransaction should be a subclass of Transaction, do you need explicit NestedTransaction here?

python/lsst/daf/butler/registry/queries/_builder.py
```diff
@@ -202,19 +202,40 @@ def joinDataset(
...
-                        )
+                        if collectionRecord.name in explicitCollections:
+                            raise NotImplementedError(
+                                f"Temporal query for dataset type '{datasetType.name}' in CALIBRATION-type "
+                                f"collection '{collectionRecord.name}' is not yet supported."
+                            )
+                        else:
+                            rejections.append(
+                                f"Not searching for dataset {datasetType.name!r} in CALIBRATION collection "
+                                f"{collectionRecord.name!r} because temporal calibration queries are't "
```
andy-slac: Typo: `are't`, same below.

andy-slac (approved):

python/lsst/daf/butler/registry/interfaces/_database.py
TallJimbo: MyPy didn't seem to think they had a direct subclass relationship, but the docs claim they do, so given the uncertainty about whether the stubs package I used is official, I'll just replace this with `Transaction` until we're ready to start depending on sqlalchemy stubs.

PR #672, opened 2022-04-05
DM-33489: Implement unit tests for QuantumBackedButler

andy-slac:
This also needed few changes to QBB classes to support serialization of
the provenance data.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/__init__.py
+++ b/python/lsst/daf/butler/__init__.py
@@ -9,6 +9,7 @@ from ._butler import *
 from ._butlerConfig import *
 from ._butlerRepoIndex import *
 from ._deferredDatasetHandle import *
+from ._quantum_backed import *
 from .core import *
 
 # Import the registry subpackage directly for other symbols.
--- a/python/lsst/daf/butler/_quantum_backed.py
+++ b/python/lsst/daf/butler/_quantum_backed.py
@@ -21,15 +21,18 @@
 
 from __future__ import annotations
 
-__all__ = ("QuantumBackedButler",)
+__all__ = ("QuantumBackedButler", "QuantumProvenanceData")
 
+import dataclasses
+import functools
 import itertools
 import logging
 import uuid
 from collections import defaultdict
 from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Optional, Set, Type, Union
 
-import pydantic
+from lsst.utils import doImportType
+from lsst.utils.introspection import get_full_type_name
 
 from ._butlerConfig import ButlerConfig
 from ._deferredDatasetHandle import DeferredDatasetHandle
@@ -53,6 +56,7 @@ from .registry.opaque import ByNameOpaqueTableStorageManager
 
 if TYPE_CHECKING:
     from ._butler import Butler
+    from .registry import Registry
 
 _LOG = logging.getLogger(__name__)
 
@@ -223,6 +227,8 @@ class QuantumBackedButler(LimitedButler):
         # TODO: We need to inform `Datastore` here that it needs to support
         # predictive reads; right now that's a configuration option, but after
         # execution butler is retired it could just be a kwarg we pass here.
+        # For now just force this option as we cannot work without it.
+        butler_config["datastore", "trust_get_request"] = True
         datastore = Datastore.fromConfig(butler_config, bridge_manager, butler_root)
         datastore.import_records(quantum.datastore_records)
         storageClasses = StorageClassFactory()
@@ -347,12 +353,13 @@ class QuantumBackedButler(LimitedButler):
             actual_inputs=self._actual_inputs,
             predicted_outputs=self._predicted_outputs,
             actual_outputs={ref.getCheckedId() for ref in self._actual_output_refs},
-            locations=locations,
-            records=records,
+            locations=dict(locations),
+            records=dict(records),
         )
 
 
-class QuantumProvenanceData(pydantic.BaseModel):
+@dataclasses.dataclass(frozen=True)
+class QuantumProvenanceData:
     """A serializable struct for per-quantum provenance information and
     datastore records.
 
@@ -371,12 +378,12 @@ class QuantumProvenanceData(pydantic.BaseModel):
     # `~CollectionType.RUN` level, such as the compute node ID). but adding it
     # now is out of scope for this prototype.
 
-    predicted_inputs: Set[uuid.UUID]
+    predicted_inputs: Set[DatasetId]
     """Unique IDs of datasets that were predicted as inputs to this quantum
     when the QuantumGraph was built.
     """
 
-    available_inputs: Set[uuid.UUID]
+    available_inputs: Set[DatasetId]
     """Unique IDs of input datasets that were actually present in the datastore
     when this quantum was executed.
 
@@ -385,7 +392,7 @@ class QuantumProvenanceData(pydantic.BaseModel):
     task.
     """
 
-    actual_inputs: Set[uuid.UUID]
+    actual_inputs: Set[DatasetId]
     """Unique IDs of datasets that were actually used as inputs by this task.
 
     This is a subset of `available_inputs`.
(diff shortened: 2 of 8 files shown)


tests/test_quantumBackedButler.py
```diff
@@ -0,0 +1,333 @@
...
+            outputs=outputs,
+            initInputs=initInputs,
+            datastore_records=datastore_records,
+        )
+
+    def test_initialize(self):
+        """Test for initialize method"""
+
+        quantum = self.make_quantum()
+        qbb = QuantumBackedButler.initialize(config=self.config, quantum=quantum, dimensions=self.universe)
```
timj, line 127: I had naively assumed that people would still use `Butler(some_thing)` to instantiate a `QuantumBackedButler` -- point at a config file that includes a pointer to the graph file. Did we decide that won't be useful because no-one outside execution is going to be wanting to create one of these and we are in charge of all the code that would?

andy-slac: My impression was indeed that this is quite special and not generally useful outside execution. Besides, QBB is a LimitedButler, not a full Butler.

timj: Yes, there's a general comment in there as to whether people need to know the type of butler they are about to create or whether they just use the same API and point their code at a URI and it works out for you what type of `Butler` you actually need. That would probably be easier with a `LimitedButler.from_config()` class method that could return a `Butler` or `QuantumBackedButler` depending on the config. For now if QBB is just for internal use it doesn't matter.

andy-slac: `LimitedButler.from_config()` may make sense to avoid special casing in ctrl_mpexec (if it can be done at all). I'll figure it out when I start working on that.

TallJimbo: I had punted on the question of construction entirely when thinking about this before.  I'd be inclined to use a `QuantumBackedButler` constructor or classmethod directly instead of trying to come up with an interface that works for any `LimitedButler`, as they could have very different arguments that wouldn't necessarily coexist well in a single signature.

timj: In my world the calling interface for `LimitedButler.from_config()` would be a URI and that URI would be a config file that has a `cls` field that allows the correct butler class to be imported which would then be given the config as normal. Things like collections and dataId args would be the same.

TallJimbo: > Things like collections and dataId args would be the same.

I think this makes my point; a `QuantumBackedButler` doesn't get to pick its collections or its default data IDs, because those aren't meaningful concepts for a `LimitedButler`.  Instead, it needs to be passed a `Quantum` or a way to get read `Quantum`.

python/lsst/daf/butler/__init__.py
```diff
@@ -15,3 +15,7 @@
 from .registry import CollectionSearch, CollectionType, DatasetIdGenEnum, Registry, RegistryConfig
 from .transfers import YamlRepoExportBackend, YamlRepoImportBackend
 from .version import *
+
+# isort: off
+# _quantum_backed has to be imported after core and registry
+from ._quantum_backed import *
```
TallJimbo: Uh oh, does this mean we have an import cycle now?  If so, I guess we can live with it now, but we should refactor it away ASAP.

python/lsst/daf/butler/_quantum_backed.py
```diff
@@ -371,12 +378,12 @@ class QuantumProvenanceData(pydantic.BaseModel):
     # `~CollectionType.RUN` level, such as the compute node ID). but adding it
     # now is out of scope for this prototype.
 
-    predicted_inputs: Set[uuid.UUID]
+    predicted_inputs: Set[DatasetId]
```
TallJimbo, line 381: The `QuantumBackedButler` concept doesn't work if we don't have UUIDs (we need to generate dataset IDs in advance, without inserting them into the repo up front, and also without having clashes), so this really can be assumed to be UUID.  I'm opening to changing it later if this is primarily a way to satisfy MyPy for now, but we shouldn't be under any illusions about that.

tests/test_quantumBackedButler.py
timj: I was putting the URI of the quantum graph in the YAML file...

python/lsst/daf/butler/_quantum_backed.py
```diff
@@ -399,26 +406,112 @@ class QuantumProvenanceData(pydantic.BaseModel):
...
+        simple: Dict[str, Any],
+        universe: Optional[DimensionUniverse] = None,
+        registry: Optional[Registry] = None,
+    ) -> QuantumProvenanceData:
+        """Make an instance of this class from serialized data.
+
+        Parameters
+        ----------
+        data : `dict`
+            Serialized representation returned from `to_simple` method,
```
TallJimbo: Missing some params, but I also don't see them used.  Just remove them from the signature?  I know we've got a bit of a convention here for which optional arguments these methods should take, but I don't think they're actually called in any completely generic context where the caller can't know the exact signature.  I suppose the same applies to `minimal` in `to_simple`.  @timj, do you have opinions on whether to have these arguments when they aren't used?

tests/test_quantumBackedButler.py
TallJimbo: Ok, fair, so we _can_ clearly pass in anything any `LimitedButler` needs via a config file, but what do we gain from doing that?  I don't think we have a use case for a consistent construction interface, just a consistent get/put/etc. use case.  And it still seems like we lose by having arguments in the signature that end up getting ignored by most implementations.

TallJimbo (approved):
Looks good!  Only minor comments, though if we do have an import cycle I don't want to wait too long to get it fixed.

python/lsst/daf/butler/_quantum_backed.py
timj: We do have a convention that these to_simple flags can propagate down to places that might use them. This is because we wanted to be able to say "give me minimal stuff" and know that it will either be ignored or include just the minimum. I think this means we need to keep the unused parameters around and be consistent in all to_simple implementations.

tests/test_quantumBackedButler.py
timj: In this specific case where a general user is not going to be confused it's fine. I think we are currently fine with "remote butler" as well because that still has to be a full butler and it's the registry class that gets replaced not the butler class. If remote butler becomes something that is its own butler class then we will have to probably have Butler be able to do something clever based on the content of the configuration YAML so that it can switch itself into a RemoteButler vs LocalButler without requiring everyone to change their constructor calls from `Butler(config)` to `Butler.from_config(config)`.

python/lsst/daf/butler/__init__.py
andy-slac: It's not a cycle, I think it's because we use absolute imports (`from lsst.daf.butler import smthg`) in some of the modules. That needs correct ordering of imports, if I import them alphabetically then I'm getting an error:
```
$ python tests/test_quantumBackedButler.py
  File "tests/test_quantumBackedButler.py", line 26, in <module>
    from lsst.daf.butler import (
  File "/home/salnikov/gen3-middleware/daf_butler/python/lsst/daf/butler/__init__.py", line 12, in <module>
    from ._quantum_backed import *
  File "/home/salnikov/gen3-middleware/daf_butler/python/lsst/daf/butler/_quantum_backed.py", line 52, in <module>
    from .registry.bridge.monolithic import MonolithicDatastoreRegistryBridgeManager
  File "/home/salnikov/gen3-middleware/daf_butler/python/lsst/daf/butler/registry/bridge/monolithic.py", line 31, in <module>
    from lsst.daf.butler import NamedValueSet, StoredDatastoreItemInfo, ddl
ImportError: cannot import name 'NamedValueSet' from partially initialized module 'lsst.daf.butler' (most likely due to a circular import) (/home/salnikov/gen3-middleware/daf_butler/python/lsst/daf/butler/__init__.py)
```

I could try to replace some of the absolute imports with relative to resolve this.

python/lsst/daf/butler/_quantum_backed.py
andy-slac: This is indeed to make mypy happy. Clearly this cannot work with integers, but I think we'll be switching to UUIDs eventually and `DatasetId` will be an alias for UUID. I can also add runtime check for the type.

andy-slac: Thinking again, I don't want to check the type of every `DatasetRef.id`, there are too many of them. As currently implemented, this butler can only work with the registries that use UUID for `dataset_id` and it will most likely crash if integers are used. I think this should be sufficient protection.

python/lsst/daf/butler/__init__.py
andy-slac: I fixed it by switching couple of modules in `registry/bridge` to use relative imports. There are many other places where we use absolute imports, maybe we should update all of them on a separate ticket.

PR #673, opened 2022-04-05
DM-34328: fix collection sorting in query results and scripts

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/script/pruneCollection.py
+++ b/python/lsst/daf/butler/script/pruneCollection.py
@@ -116,6 +116,8 @@ def pruneCollection(
         for name in collectionNames:
             addCollection(name)
 
+        collections = {k: collections[k] for k in sorted(collections.keys())}
+
         queryDatasets = QueryDatasets(
             repo=repo,
             glob=None,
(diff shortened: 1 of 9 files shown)


python/lsst/daf/butler/script/pruneCollection.py
```diff
@@ -116,6 +116,8 @@ def addCollection(name: str) -> None:
         for name in collectionNames:
             addCollection(name)
 
+        collections = {k: collections[k] for k in sorted(collections.keys())}
+
```
n8pease, line 120: it looks like this line sorts the `collections` dict by its keys, but then never uses it sequentially (and also it's not an ordered dict). Am I misunderstanding it, or is it not needed?

TallJimbo: Usage is in the lists defined in lines 137-139.  `dict` itself now preserves insertion order in Python; `OrderedDict` is I think still around only for some edge case different behavior and backwards compatibility.

n8pease: oh, of course. I didn't expand the view down. Thx.

n8pease:
LGTM

PR #674, opened 2022-04-07
DM-33942: Schema changes

timj:
## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/core/dimensions/_schema.py
```diff
@@ -170,7 +170,7 @@ def __init__(self, element: DimensionElement):
         # don't want *recursive* implied dependencies.
         self.implied = NamedValueSet()
         for dimension in element.implied:
-            fieldSpec = addDimensionForeignKey(self._tableSpec, dimension, primaryKey=False, nullable=True)
+            fieldSpec = addDimensionForeignKey(self._tableSpec, dimension, primaryKey=False, nullable=False)
```
andy-slac, line 173: Could we also add explicit `nullable=False` to required elements above (line 155)? They are part of primary key so they are implicitly non-NULL, and it looks like sqlalchemy is OK with the column definition which is both PK and nullable, but this is not consistent with the schema when it is reflected back from database. This is not a schema change, of course, just some inconsistency that I discovered in looking at checksusm.

timj: Setting `primaryKey=True` does not sett `nullable=False` automatically? @TallJimbo made the above change. If he agrees I can add `nullable=False` to line 155.

TallJimbo: I had just assumed sqlalchemy would ignore `nullable` for primary keys at a lower level.  If that's not true at some level of interest to us as @andy-slac says, yeah, we should change it - though maybe the `FieldSpec` constructor or `__post_init__` (it's a dataclass) would be a more thorough way to do that.

andy-slac: It's a bit complicated, in sqlalchemy indeed setting `primaryKey=True` does not automatically set `nullable=False`, so when you inspect sqlalchemy `Column` you can see the definitions where both `PK` and `nullable` are true. When a table is created in the database then database of course does the right thing. And when you reflect schema from the database the PK columns always have `nullable=False`. I think it would better if reflected definitions and definitions that we construct ourselves were matching each other (but there are other things that don't match of course). And indeed, `FieldSpec` is probably a better place to enforce that consistently.

andy-slac: @TallJimbo - one more issue with this update - I tried to implement migration script for dimensions manager (from 6.0.1 to 6.0.2) by altering the columns that correspond to implied dimensions to be `NOT NULL` and using the copy of the `main_20210215` repository. My script failed because `visit` table has NULLs in `physical_filter` column for some records (all come from LATISS). Not sure what to do about it.

timj: Hmm. This is probably some very old LATISS data with a broken translator. It's all going to be unusable if it can't be linked to a physical filter. Maybe we need to recalculate the exposure records for them.

TallJimbo: LATISS has an "unknown~unknown" `physical_filter` we could update these rows with (same with `exposure`), and I imagine any other instrument for which this is a possibility has a similar sentinel value, and it might be best to just do that now to get us through the migration first.  But we should record the IDs of updated `visits` and `exposures`, too, so we can come back later and check them with the latest translator and visit-definition code.

andy-slac: `exposure` does not have any NULLs, they only appear in `visit`. There are 182 of them. I made a page with the CSV file attached which lists instrument and visit ID of those rows: https://confluence.lsstcorp.org/display/~salnikov/NULL+physical_filter+values+in+visit+table

I'll update things in my local copy to continue with my tests and we can update central databases later.

timj: That's a bit of a bug then because the vist filter should match the exposure one. Maybe we need to re-run define-visits in update mode for them.

timj: So FieldSpec defaults to nullable=True but  addDimensionForeignKey defaults to nullable=False and that forces the FieldSpec to be nullable=False regardless (since FieldSpec has no way of stopping someone changing nullable after it's constructed). I don't know if we want to add special logic to FieldSpec to force setting nullable=False if primaryKey==True but I can at least fix addDimensionForeignKey to make it use a dynamic default based on primaryKey.

edit: I got things the wrong way round before. nullable does have a different default in FieldSpec and addDimensionForeignKey, but the deafult is False in addDimensionForeignKey so I'm a bit confused why Andy is seeing an inconsistency.

andy-slac: I looked at the physical filters for exposures associated with visits that have NULL physical_philter. Those visits have multiple exposures with different physical filters, e.g.:
```
instrument|visit           |exposure     |visit_system|visit_filter|exposure_filter        
----------+----------------+-------------+------------+------------+-----------------------
LATISS    |1911254347800000|2021012000118|           1|            |empty~ronchi90lpmm     
LATISS    |1911254347800000|2021012000119|           1|            |quadnotch1~ronchi90lpmm
LATISS    |1911254347800000|2021012000120|           1|            |BG40~empty             
LATISS    |1911254347800000|2021012000121|           1|            |RG610~empty            
```

timj: I just ran the test suite where I added a RuntimeError if the fieldspec coming out of addDimensionForeignKey had nullable=True and primaryKey=True and all tests passed.

python/lsst/daf/butler/core/ddl.py
```diff
@@ -308,6 +309,11 @@ class FieldSpec:
     doc: Optional[str] = None
     """Documentation for this field."""
 
+    def __post_init__(self) -> None:
```
timj, line 312: @andy-slac I added this to force the value to False. It does not catch someone creating a `FieldSpec` and then later changing it to `fieldSpec.primaryKey=True` (which is what is seemingly going on in the `addDimensionForeignKey` call but with a default of `False` which is confusing me).

python/lsst/daf/butler/core/dimensions/_schema.py
timj: Ah. Okay. That's us suffering from the current visit_system=1 definition that uses the exposure.group. It means that they had one observing script but kept changing the filter. When define-visits runs, it drops the entry for filter because it changed during the visit. In the new schema change world the default visit definition will make a filter change impossible.

timj: Anyhow, adding `unknown~unknown` for the visit filters there is fine as a hack.

python/lsst/daf/butler/configs/dimensions.yaml
```diff
@@ -269,6 +273,18 @@ elements:
           Counter for the observation within a larger sequence. Context
           of the sequence number is observatory specific. Can be
           a global counter or counter within day_obs.
+      -
+        name: seq_start
+        type: int
+        doc: >
+          The sequence number of the first exposure of the visit that contains
+          this exposure.
```
TallJimbo: I'd like to find a way to reword this; I know what you mean, but "the visit that contains this exposure" is not a well-defined thing, because there are in general two such visits.  Maybe something like "the as-observed visit" or "the standard visit"?

python/lsst/daf/butler/transfers/_yaml.py
```diff
@@ -247,6 +247,16 @@ def __init__(self, stream: IO, registry: Registry):
...
         self.refsByFileId: Dict[DatasetId, DatasetRef] = {}
         self.registry: Registry = registry
+
+        # If this is data exported before the reorganization of visits
+        # and visit systems and that new schema is in use, some filtering
+        # will be needed. The entry in the visit dimension record will be
+        # silently dropped when visit is created but the
+        # visit_system_membership must be constructed.
+        migrate_visit_system = False
+        if fileVersion < VersionTuple(1, 0, 2) and "visit_system_membership" in self.registry.dimensions:
```
TallJimbo: I'm not sure the file version is relevant here; consider someone creating an export file after this branch merges from a repo that hasn't been migrated.  I think it's the dimensions configuration version we need to start stuffing in the export files and using for logic like this.

python/lsst/daf/butler/transfers/_yaml.py
```diff
@@ -262,6 +272,16 @@ def __init__(self, stream: IO, registry: Registry):
...
                 self.dimensions[element].extend(RecordClass(**r) for r in data["records"])
+
+                if data["element"] == "visit" and migrate_visit_system:
+                    # Must create the visit_system_membership records.
+                    element = self.registry.dimensions["visit_system_membership"]
+                    RecordClass = element.RecordClass
+                    self.dimensions[element].extend(
+                        RecordClass(instrument=r["instrument"], visit_system=r["visit_system"], visit=r["id"])
+                        for r in data["records"]
+                    )
```
TallJimbo, line 294: I think we need a migration here for `visit_definition`, too, right, since that had its columns change?

Maybe it's worked because `insertDimensionData` is ignoring columns it doesn't recognize?  (Either that or maybe we don't have test coverage.)

python/lsst/daf/butler/core/ddl.py
TallJimbo: It'd probably be cleaner to make `FieldSpec` a frozen dataclass (not just for this), but I bet there's at least one more place where we assume it isn't (in addition to the new `__post_init__`).  So a project for another day, unless you're inspired.

TallJimbo (approved):
Looks good, but I'm concerned the export/import fixes are necessary-but-not-sufficient.

python/lsst/daf/butler/configs/dimensions.yaml
timj: I don't really follow. This visit has a set of exposures in it. The seq_num is the lowest seq_num of all those exposures. (the data model allows it to be any seq_num we want in reality, but define-visits ensures that it's the lowest value). If there are two visits that both have the same seq_num that's not a problem (and that's deliberate) and is why we have the instrument.visit_system tie breaker.

How about:

> The lowest seq_num from the constituent exposures defined by this visit.

?

python/lsst/daf/butler/transfers/_yaml.py
timj: I'm not really sure what I was thinking. It's entirely true that <1.0.2 and modern registry does require a migration but you are right that we are still going to write the new version when exporting from the old registry.

Maybe the quick fix here is to make it write 1.0.1 as the version number if it's exporting an old registry?

The alternative is to store the dimensions version and namespace in the export YAML and read them on import. I'm starting to wonder now how we handle export/import for SPHEREx and PFS where they have a different namespace. Changing some dimension records is fine, but larger changes (as we did when adding visit_system_membership) are going to make this all blow up for other universes.

python/lsst/daf/butler/transfers/_yaml.py
timj: It drops unknown columns on the floor.

python/lsst/daf/butler/transfers/_yaml.py
timj: Oh, and exporting from modern and trying to import into old is not going to work. Should it? Does that mean these export files should be version 2?

python/lsst/daf/butler/configs/dimensions.yaml
TallJimbo: My point was that "this visit" isn't well defined in the sense that there are multiple visit systems, and an exposure can be in different visits for different visit systems; I think you mean "the visit as defined by the script queue or whatever writes this exposure's header", but it's confusing that we use "visit" to mean both:
- a concept defined by the script queue/header/exposure metadata
- a concept defined by middleware/DefineVisitsTAsk, which may or may not care what the header says.

python/lsst/daf/butler/core/ddl.py
timj: The code that started us looking at this is directly manipulating FieldSpec so I know that freezing is not going to work immediately. If we are lucky that's the only place.

python/lsst/daf/butler/configs/dimensions.yaml
timj: The only safe definition for the doc string is "A sequence number associated in some manner with this visit". Everything else involves some assumptions as to how define-visit, in our case, is setting that seq_num field. The seq_num definition shouldn't need to know whether it's the first exposure in the visit, the last one, or a random number not related to any exposure. At Rubin we require that the seq_num is the first exposure in the visit and both visit definitions agree with what that value is.

TallJimbo: Yeah, I think that the best description here is probably one that avoids saying "visit" at all; maybe something like "The first 'seq_num' value in the set of exposures this one is part of."  It'd be nicer to have a non-overloaded word for this particular kind of "set of exposures" (can't call it a visit, can't call it a sequence, can't call it a group!), but just not saying visit is probably fine.

(Note that this is in the definition for `exposure`, not `visit`; I'm not sure if that might have been part of where we were miscommunicating.)

python/lsst/daf/butler/transfers/_yaml.py
TallJimbo: I don't think there are any easy answers here - migration of YAML export files w.r.t. dimensions changes is just as hard in general as migration of databases, and we've been unclear on whether YAML export files are _more_ stable and migration-supported than databases (because many things that do affect databases don't affect them at all) or less (because they're _often_ more ephemeral and we don't have anything like alembic to lean on).

I do think we should start at least writing the dimensions namespace+version into export files, so we can tell when a migration may be needed.

And I think we should reserve the export file format version for changes to the import/export format itself - new kinds of fields it can have, changing some field from a `list` to `dict`, etc - and not use it as a stand-in for the dimensions config, _except_ as a way to guess the dimensions namespace+version when reading files that were written before we started saving the actual namespace+version.

Beyond that, maybe we need to start putting export YAML migration scripts for the transitions we care about somewhere, or just declaring that we don't support migrating those and they should really be transient things?

python/lsst/daf/butler/configs/dimensions.yaml
timj: You are indeed correct. I thought we were talking about `visit` since I was only looking at the four lines. Sorry.

If this is `seq_start` then... The definition is effectively instrument-specific but something like:

> Earliest sequence number that might be related to this exposure.

? And similarly for seq_end.

python/lsst/daf/butler/transfers/_yaml.py
timj: Okay. I'll write the dimension version and namespace to the file. I will up the patch number for the file format but I won't use it -- I'll be able to use the version/patch for that test instead.

But yes, in some sense we should not only be writing `butler migrate` commands for migrating registries but also for migrating the export YAML files.

python/lsst/daf/butler/configs/dimensions.yaml
andy-slac: The migration script, that I just added for v2, sets both `seq_start` and `seq_end` to the value of `seq_num` from the same record. Should it try to do something different and use min/max from related exposures (not sure if there is an easy way to do that)?

timj: The migration script is correct. Only LSST instruments (for now) have seq_num != seq_start. There are a few very new LATISS files that we will have to fix the seq_start/seq_end value for but you can't predict when you might need to do that and it needs to be done by looking at the actual file headers.

python/lsst/daf/butler/transfers/_yaml.py
timj: Now looks like:
```
description: Butler Data Repository Export
version: 1.0.2
universe_version: 0
universe_namespace: daf_butler
```

I added a `DimensionUniverse.getVersionInfo()` method (but now I'm thinking I should make .version and .namespace read-only properties instead.

timj: Went with .version and .namespace.

python/lsst/daf/butler/transfers/_yaml.py
```diff
@@ -247,6 +251,23 @@ def __init__(self, stream: IO, registry: Registry):
         self.calibAssociations: Dict[str, Dict[Timespan, List[DatasetId]]] = defaultdict(dict)
         self.refsByFileId: Dict[DatasetId, DatasetRef] = {}
         self.registry: Registry = registry
+
+        universe_version = wrapper.get("universe_version", 0)
```
timj, line 255: @TallJimbo would you mind taking a quick look at the changes here and in `_universe.py` to make sure they match your expectations before I merged?  I added version and namespace properties to universe and now check specifically for the old universe version when deciding to do a migration.

timj, line 255: @TallJimbo would you mind taking a quick look at the changes here and in `_universe.py` to make sure they match your expectations before I merged?  I added version and namespace properties to universe and now check specifically for the old universe version when deciding to do a migration.

TallJimbo (approved):
Yup, all looks good with the export/import dimensions changes, and I do think keeping the bit of hard-coded translation logic you have (dispatched off the right version, as it now is) is the pragmatic thing to do now, even if it doesn't scale.

PR #676, opened 2022-04-08
DM-34377: Add postgres testing for butler (including datastore)

timj:
This will hopefully catch the problem fixed by #675. We do not run postgres tests on Jenkins but we do run them in github actions. This test does fail if #675 is not applied.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_butler.py
+++ b/tests/test_butler.py
@@ -34,6 +34,8 @@ import string
 import tempfile
 import time
 import unittest
+from tempfile import gettempdir
+from threading import Thread
 
 try:
     import boto3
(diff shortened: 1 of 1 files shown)


andy-slac (approved):
Looks good. General comment - the server setup logic repeats some of the `test_postgresql.py`, would it be worth to move it to a common base class?

timj:
`test_postgresql.py` does the setup a slightly different way. I tried using that way but the tests got a bit upset. I therefore switched to the caching scheme recommended in the testing.postgresql docs. I also tried switching `test_postgresql.py` over to that approach but it took 10% longer to run the tests that way. If we don't mind that slowdown I can move the setUp/teadDown[class] code to a helper class.

andy-slac:
Is it 10% for `test_postgresql.py` or the whole test? I'm OK with leaving it as it is now, and let the next person who also needs postgres to refactor it.

timj:
The entire run of test_postgresql.py went from 60 seconds to 70 seconds when I used the postgre cache factory thing.

PR #677, opened 2022-04-12
DM-33493: Add serialization support for datastore records

andy-slac:
Adds a bunch of pydantic classes for datastore records and implements
serialization of the records in `Quantum` class. I switched back
`QuantumProvenanceData` to be a pydantic model to use the same pydantic
classes for datastore records. Moved `DatastoreRecordData` class (and its
pydantic friend) to a separate module, for some reason pydantic was not
happy when I added `SerializedDatastoreRecordData` class to the
`datastore` module.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/datastoreRecordData.py
+++ b/python/lsst/daf/butler/core/datastoreRecordData.py
@@ -1,0 +1,174 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Support for generic data stores."""
+
+from __future__ import annotations
+
+__all__ = ("DatastoreRecordData", "SerializedDatastoreRecordData")
+
+import dataclasses
+import uuid
+from collections import defaultdict
+from typing import TYPE_CHECKING, AbstractSet, Dict, List, Optional
+
+from pydantic import BaseModel
+
+from .datasets import DatasetId
+from .dimensions import DimensionUniverse
+from .storedFileInfo import SerializedStoredDatastoreItemInfo, StoredDatastoreItemInfo
+
+if TYPE_CHECKING:
+
+    from ..registry import Registry
+    from ..registry.interfaces import DatasetIdRef
+
+
+class SerializedDatastoreRecordData(BaseModel):
+    """Representation of a `DatastoreRecordData` suitable for serialization."""
+
+    ref_ids: List[uuid.UUID]
+    records: Dict[str, List[SerializedStoredDatastoreItemInfo]]
+
+    @classmethod
+    def direct(
+        cls,
+        *,
+        ref_ids: List[uuid.UUID],
+        records: Dict[str, List[Dict]],
+    ) -> SerializedDatastoreRecordData:
+        """Construct a `SerializedDatastoreRecordData` directly without
+        validators.
+
+        This differs from the pydantic "construct" method in that the
+        arguments are explicitly what the model requires, and it will recurse
+        through members, constructing them from their corresponding `direct`
+        methods.
+
+        This method should only be called when the inputs are trusted.
+        """
+        data = SerializedDatastoreRecordData.__new__(cls)
+        setter = object.__setattr__
+        setter(data, "ref_ids", ref_ids)
+        setter(
+            data,
+            "records",
+            {k: [SerializedStoredDatastoreItemInfo.direct(**item) for item in v] for k, v in records.items()},
+        )
+        return data
+
+
+@dataclasses.dataclass
+class DatastoreRecordData:
+    """A struct that represents a tabular data export from a single
+    datastore.
+    """
+
+    refs: List["DatasetIdRef"] = dataclasses.field(default_factory=list)
+    """List of DatasetRefs known to this datastore.
+    """
+
+    records: Dict[str, List[StoredDatastoreItemInfo]] = dataclasses.field(
+        default_factory=lambda: defaultdict(list)
+    )
+    """Opaque table data, grouped by opaque table name.
+    """
+
+    def subset(self, dataset_ids: AbstractSet[DatasetId]) -> Optional[DatastoreRecordData]:
+        """Extract a subset of the records that match given dataset IDs.
+
+        Parameters
+        ----------
+        dataset_ids : `set` [ `DatasetId` ]
+            Dataset IDs to match.
+
+        Returns
+        -------
+        record_data : `DatastoreRecordData` or `None`
+            `None` is returned if there are no matching refs.
+        """
+        matching_refs = [ref for ref in self.refs if ref.id in dataset_ids]
+        records: Dict[str, List[StoredDatastoreItemInfo]] = {}
+        for table_name, record_data in self.records.items():
+            matching_records = [record for record in record_data if record.dataset_id in dataset_ids]
+            if matching_records:
+                records[table_name] = matching_records
+        if matching_refs or matching_records:
+            return DatastoreRecordData(refs=matching_refs, records=records)
+        else:
+            return None
+
+    def to_simple(self, minimal: bool = False) -> SerializedDatastoreRecordData:
+        """Make representation of the object for serialization.
+
+        Implements `~lsst.daf.butler.core.json.SupportsSimple` protocol.
+
+        Parameters
+        ----------
+        minimal : `bool`, optional
+            If True produce minimal representation, not used by this method.
+
+        Returns
+        -------
+        simple : `dict`
+            Representation of this instance as a simple dictionary.
+        """
+        ref_ids = [ref.id for ref in self.refs]
+        record_data = {
+            table_name: [record.to_simple() for record in records]
+            for table_name, records in self.records.items()
+        }
+        return SerializedDatastoreRecordData(ref_ids=ref_ids, records=record_data)
+
+    @classmethod
+    def from_simple(
+        cls,
+        simple: SerializedDatastoreRecordData,
+        universe: Optional[DimensionUniverse] = None,
+        registry: Optional[Registry] = None,
+    ) -> DatastoreRecordData:
+        """Make an instance of this class from serialized data.
+
+        Implements `~lsst.daf.butler.core.json.SupportsSimple` protocol.
+
+        Parameters
+        ----------
+        data : `dict`
+            Serialized representation returned from `to_simple` method.
+        universe : `DimensionUniverse`, optional
+            Dimension universe, not used by this method.
+        registry : `Registry`, optional
+            Registry instance, not used by this method.
+
+        Returns
+        -------
+        item_info : `StoredDatastoreItemInfo`
+            De-serialized instance of `StoredDatastoreItemInfo`.
+        """
+        # this causes dependency cycle if imported at top level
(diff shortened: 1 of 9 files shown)


python/lsst/daf/butler/core/quantum.py
```diff
@@ -147,6 +156,7 @@ def direct(
                 "inputs",
                 "outputs",
                 "dimensionRecords",
+                "datastore_records",
```
TallJimbo, line 159: I _think_ we tend to use camelCase in the "simple" representations since that usually maps to JSON and JSON tends to use camelCase, even though we're trying to use snake_case more in new Python code.  @timj, is that your understanding, too?

(I'm not commenting on other cases of the same for now)

python/lsst/daf/butler/core/storedFileInfo.py
```diff
@@ -21,25 +21,70 @@
...
+    """Representation of `StoredDatastoreItemInfo` used for serialization.
+
+    Note that this class supports polymorphism by storing actual class name of
+    the items together with all other data members. It depends on a correct
+    implementation of ``to_record`` and ``from_record`` methods in each
+    sub-class.
+    """
+
+    class_name: str
+    """Actual fully-qualified name of the item class."""
```
TallJimbo: How hard would it be to normalize this field out into a per-datastore value?  I think it'll have the same for all instances corresponding to a particular datastore.  And if we could do so, it looks like this class could just be replaced by using a `dict`; I'm thinking of something like redefining `SerializedDatastoreRecordData.records` as
```
records: Dict[str, List[Dict[str, Any]]
```
where the first item in each list could be `{"className": <value>}`.

TallJimbo (approved):

python/lsst/daf/butler/core/quantum.py
timj: There doesn't seem to be a JSON standard. From what I can tell people use whatever convention their language layer is using.  I'm obviously trying to switch us to underscores but camel case would use less space in the file. If we stick with underscores here we should probably change the line above to be `dimension_records` (assuming we don't have a backwards compatibility problem)

andy-slac: > If we stick with underscores here we should probably change the line above to be `dimension_records` (assuming we don't have a backwards compatibility problem)

I think there is a potential backward compatibility issue with renaming an existing field, I prefer to change the new one to camel case and save one character.

python/lsst/daf/butler/core/storedFileInfo.py
andy-slac: This may be more complicated (if done correctly). The `SerializedDatastoreRecordData.records` is now indexed by the table name, and potentially the same table name could have different schema and record type for different datastores. We now have only FileDatastore, so this may not matter, but if we ever to add a different datastore type it may be worth thinking about how to avoid future complications. I guess one way is to solve this to do something like:

    records: Dict[str, Dict[str, List[Dict[str, Any]]]

or 

    records: Dict[Tuple[str, str], List[Dict[str, Any]]


with the first index being the table name and second is the record class name. But I also need to think about your proposed change (on pipe_base PR) to index it by dataset ID instead, I'll figure something out.

andy-slac: Another option that looked attractive but had other complications is to let datastore handle deserialization. Record type is specific to datastore so it's a natural place to do that. Complication is that I would need to get an access to a datastore by its name, which is probably not trivial.

timj: The name is a public property of an instantiated datastore. `butler.datastore` is all you have but there doesn't seem to be a problem with passing records to a method of `butler.datastore` and then allowing the datastore to handle it. It knows its own name and a chained datastore can iterate over each datastore in the chain and only pass records that have a matching name.

andy-slac: Unfortunately it's harder than I hoped, `butler.datastore` is not currently accessible anywhere in the chain of `from_simple()` calls. QuantumGraph loading in general does not require butler to even be present today, and I'm reluctant to add it as an (optional) dependency for `load` method. Another approach to try is a sort of singleton registry for deserializers.

timj: I should have read the code that was being discussed. No, adding datastore to from_simple seems like a bad idea. I thought we were talking about something downstream of from_simple that would be handed to a chained datastore (but I imagine that was the code you wrote a couple of weeks back).

python/lsst/daf/butler/core/datastoreRecordData.py
```diff
@@ -28,32 +28,68 @@
...
+
+def _str_to_uuid(dataset_id: Any) -> uuid.UUID:
+    """Convert dataset_id from string back to UUID, e.g. after JSON
+    round-trip.
+    """
+    if isinstance(dataset_id, str):
+        return uuid.UUID(dataset_id)
+    else:
+        assert isinstance(dataset_id, uuid.UUID), "Expecting UUIDs"
+        return dataset_id
```
TallJimbo: If one of these code paths can be identified as the usual one, my intuition is that it could be a lot more performant to do this with `try`/`except` if that avoids an `isinstance` check most of the time.

python/lsst/daf/butler/core/datastoreRecordData.py
```diff
@@ -82,15 +120,29 @@ class DatastoreRecordData:
...
 
-    refs: List["DatasetIdRef"] = dataclasses.field(default_factory=list)
-    """List of DatasetRefs known to this datastore.
+    dataset_ids: Set[DatasetId] = dataclasses.field(default_factory=set)
+    """List of DatasetIds known to this datastore.
     """
 
-    records: Dict[str, List[StoredDatastoreItemInfo]] = dataclasses.field(
-        default_factory=lambda: defaultdict(list)
+    records: Dict[DatasetId, Dict[str, List[StoredDatastoreItemInfo]]] = dataclasses.field(
```
TallJimbo: I think it'd be better to annotate this as:
```
defaultdict[DatasetId, defaultdict[str, list[StoredDatastoreItemInfo]]]
```
i.e. explicitly point out that these are `defautldict` objects, since (despite the formal inheritance relationship) the differing behavior of `__getitem__` can be confusing when you're expecting a vanilla `dict`.  Even more importantly, (as I think another comment here demonstrates) if someone did put a vanilla `dict` here and other code was expecting to use `defaultdict` for the usual missing-insertion pattern it provides, that code would break, and we want MyPy to catch that.

python/lsst/daf/butler/core/datastoreRecordData.py
```diff
@@ -82,15 +120,29 @@ class DatastoreRecordData:
     datastore.
     """
 
-    refs: List["DatasetIdRef"] = dataclasses.field(default_factory=list)
-    """List of DatasetRefs known to this datastore.
+    dataset_ids: Set[DatasetId] = dataclasses.field(default_factory=set)
+    """List of DatasetIds known to this datastore.
```
TallJimbo: I'm surprised this field is useful, given that its contents seem like they should be the same as `records.keys()`.  Is that not the case, or perhaps just not _guaranteed_ to be the case?

python/lsst/daf/butler/core/datastoreRecordData.py
```diff
@@ -82,15 +120,29 @@ class DatastoreRecordData:
...
+        Parameters
+        ----------
+        other : `DatastoreRecordData`
+            Records tho merge into this instance.
+        """
+        self.dataset_ids.update(other.dataset_ids)
+        for dataset_id, table_records in other.records.items():
+            this_table_records = self.records.setdefault(dataset_id, {})
+            for table_name, records in table_records.items():
+                this_table_records.setdefault(table_name, []).extend(records)
```
TallJimbo: No need for `setdefault` here or two lines above since you're using `defaultdict`, and actually using `setdefault` with a plain `dict` will mess up other code that presumably does assume the nested thing is a `defaultdict`.  I think you need to stick to either `setdefault` or `defaultdict.__getitem__` across the board.

python/lsst/daf/butler/core/datastoreRecordData.py
```diff
@@ -105,14 +157,13 @@ def subset(self, dataset_ids: AbstractSet[DatasetId]) -> Optional[DatastoreRecor
...
-            matching_records = [record for record in record_data if record.dataset_id in dataset_ids]
-            if matching_records:
-                records[table_name] = matching_records
-        if matching_refs or matching_records:
-            return DatastoreRecordData(refs=matching_refs, records=records)
+        matching_ids = self.dataset_ids & dataset_ids
+        matching_records: Dict[DatasetId, Dict[str, List[StoredDatastoreItemInfo]]] = {}
+        for dataset_id in matching_ids:
+            if (id_records := self.records.get(dataset_id)) is not None:
+                matching_records[dataset_id] = id_records
```
TallJimbo, line 142: Should we be worried about the aliasing of mutable data structures here?

I suspect not in practice, but it should probably be called out in docs.

python/lsst/daf/butler/core/datastoreRecordData.py
```diff
@@ -131,12 +182,24 @@ def to_simple(self, minimal: bool = False) -> SerializedDatastoreRecordData:
...
-        return SerializedDatastoreRecordData(ref_ids=ref_ids, records=record_data)
+
+        def _class_name(records: list[StoredDatastoreItemInfo]) -> str:
+            """Get fully qualified class name for the records. Empty string
+            returned if list is empty. Exception is raised if records are of
+            different classes.
+            """
+            if not records:
+                return ""
+            classes = set(record.__class__ for record in records)
```
TallJimbo, line 171: Probably better to use `lsst.utils.introspection.get_full_type_name`, as that tries to determine a "less private" name by dropping leading-underscore modules that we don't want to bake into files.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -2702,25 +2703,30 @@ def import_records(self, data: Mapping[str, DatastoreRecordData]) -> None:
...
+                    assert isinstance(info, StoredFileInfo), "Expecting StoredFileInfo records"
+                    unpacked_records.append(info.to_record())
+        if unpacked_records:
             self._table.insert(*unpacked_records)
 
     def export_records(self, refs: Iterable[DatasetIdRef]) -> Mapping[str, DatastoreRecordData]:
         # Docstring inherited from the base class.
         exported_refs = list(self._bridge.check(refs))
+        ids = {ref.getCheckedId() for ref in exported_refs}
+        records: Dict[DatasetId, Dict[str, List[StoredDatastoreItemInfo]]] = defaultdict(
```
TallJimbo: Also recommend `defaultdict` explicitly in this type annotation.

TallJimbo (approved):
Looks good!  I have some minor comments that I think should be addressed, but the big picture looks great.

python/lsst/daf/butler/core/datastoreRecordData.py
andy-slac: The issue here is JSON which stores UUIDs as strings and reads string back, and regular construction from datastore should always use true UUIDs. The problem is that both of those cases use regular constructor. We also have a `direct()` method that we use when restoring quantum graph from saved data. If we could say that JSON deserialization have to always use `direct()` instead of `__init__` than we can skip the conversion in `__init__` but always do it in `direct()` (but then we loose validation when reading from JSON).

python/lsst/daf/butler/core/datastoreRecordData.py
andy-slac: My thinking was that there may not be a guarantee for some future implementation. Thinking about it again, it is already possible to model that case by using an empty dict in `records` for that key. I'll drop `dataset_ids`.

python/lsst/daf/butler/core/datastoreRecordData.py
andy-slac: I don't think anyone would need to update `DatastoreRecordData` after construction, so its content should be immutable. I'll add a comment to docstring.

python/lsst/daf/butler/core/datastoreRecordData.py
andy-slac: This line is to only make sure that there is a single class for all records, I think it does not matter here what is the exact name and I tried to avoid building a complete class name for each record. `get_full_type_name` is used below to return the name of that single class.

python/lsst/daf/butler/core/datastoreRecordData.py
andy-slac: I removed validators, constructor now accepts UUIDs only, and `direct()` method converts strings to UUIDs, similarly to how it's done for `DatasetRef` class.

PR #680, opened 2022-05-02
DM-34616: prepare for NOT NULL constraints on implied dimensions

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/tests/_testRepo.py
+++ b/python/lsst/daf/butler/tests/_testRepo.py
@@ -19,6 +19,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from __future__ import annotations
 
 __all__ = [
     "makeTestRepo",
@@ -30,7 +31,7 @@ __all__ = [
 ]
 
 import random
-from typing import Any, Iterable, Mapping, Optional, Set, Tuple
+from typing import Any, Iterable, Mapping, Optional, Set, Tuple, Union
 from unittest.mock import MagicMock
 
 import sqlalchemy
@@ -43,7 +44,6 @@ from lsst.daf.butler import (
     Dimension,
     DimensionUniverse,
     FileDataset,
-    Registry,
 )
 
 
@@ -57,17 +57,18 @@ def makeTestRepo(
     root : `str`
         The location of the root directory for the repository.
     dataIds : `~collections.abc.Mapping` [`str`, `iterable`], optional
-        A mapping keyed by the dimensions used in the test. Each value
-        is an iterable of names for that dimension (e.g., detector IDs for
-        `"detector"`). Related dimensions (e.g., instruments and detectors)
-        are linked arbitrarily. This parameter is provided for compatibility
-        with old code; newer code should make the repository, then call
+        A mapping keyed by the dimensions used in the test. Each value is an
+        iterable of names for that dimension (e.g., detector IDs for
+        `"detector"`). Related dimensions (e.g., instruments and detectors) are
+        linked arbitrarily, with values created for implied dimensions only
+        when needed. This parameter is provided for compatibility with old
+        code; newer code should make the repository, then call
         `~lsst.daf.butler.tests.addDataIdValue`.
     config : `lsst.daf.butler.Config`, optional
         A configuration for the repository (for details, see
         `lsst.daf.butler.Butler.makeRepo`). If omitted, creates a repository
-        with default dataset and storage types, but optimized for speed.
-        The defaults set ``.datastore.cls``, ``.datastore.checksum`` and
+        with default dataset and storage types, but optimized for speed.  The
+        defaults set ``.datastore.cls``, ``.datastore.checksum`` and
         ``.registry.db``.  If a supplied config does not specify these values
         the internal defaults will be used to ensure that we have a usable
         configuration.
@@ -79,15 +80,15 @@ def makeTestRepo(
     butler : `lsst.daf.butler.Butler`
         A Butler referring to the new repository. This Butler is provided only
         for additional setup; to keep test cases isolated, it is highly
-        recommended that each test create its own Butler with a
-        unique run/collection. See `makeTestCollection`.
+        recommended that each test create its own Butler with a unique
+        run/collection. See `makeTestCollection`.
 
     Notes
     -----
-    This function provides a "quick and dirty" repository for simple unit
-    tests that don't depend on complex data relationships. It is ill-suited
-    for tests where the structure of the data matters. If you need such a
-    dataset, create it directly or use a saved test dataset.
+    This function provides a "quick and dirty" repository for simple unit tests
+    that don't depend on complex data relationships. It is ill-suited for tests
+    where the structure of the data matters. If you need such a dataset, create
+    it directly or use a saved test dataset.
     """
     defaults = Config()
     defaults["datastore", "cls"] = "lsst.daf.butler.datastores.inMemoryDatastore.InMemoryDatastore"
@@ -165,36 +166,41 @@ def _makeRecords(dataIds: Mapping[str, Iterable], universe: DimensionUniverse) -
         `~lsst.daf.butler.DimensionRecord` for each input name. Related
         dimensions (e.g., instruments and detectors) are linked arbitrarily.
     """
-    expandedIds = {}
-    # Provide alternate keys like detector names
-    for name, values in dataIds.items():
-        expandedIds[name] = []
+
+    # Create values for all dimensions that are (recursive) required or implied
+    # dependencies of the given ones.
+    complete_data_id_values = {}
+    for dimension in universe.extract(dataIds.keys()):
+        if dimension.name in dataIds:
+            complete_data_id_values[dimension.name] = list(dataIds[dimension.name])
+        if dimension.name not in complete_data_id_values:
+            complete_data_id_values[dimension.name] = [_makeRandomDataIdValue(dimension)]
+
+    # Start populating dicts that will become DimensionRecords by providing
+    # alternate keys like detector names
+    record_dicts_by_dimension_name = {}
+    for name, values in complete_data_id_values.items():
+        record_dicts_by_dimension_name[name] = []
         dimension = universe[name]
         for value in values:
-            expandedIds[name].append(_fillAllKeys(dimension, value))
+            record_dicts_by_dimension_name[name].append(_fillAllKeys(dimension, value))
 
     # Pick cross-relationships arbitrarily
-    for name, values in expandedIds.items():
+    for name, record_dicts in record_dicts_by_dimension_name.items():
         dimension = universe[name]
-        for value in values:
-            for other in dimension.required:
+        for record_dict in record_dicts:
+            for other in dimension.dimensions:
                 if other != dimension:
-                    relation = expandedIds[other.name][0]
-                    value[other.name] = relation[other.primaryKey.name]
-            # Do not recurse, to keep the user from having to provide
-            # irrelevant dimensions
-            for other in dimension.implied:
-                if other.name in expandedIds:
-                    relation = expandedIds[other.name][0]
-                    value[other.name] = relation[other.primaryKey.name]
+                    relation = record_dicts_by_dimension_name[other.name][0]
+                    record_dict[other.name] = relation[other.primaryKey.name]
 
     return {
-        dimension: [universe[dimension].RecordClass(**value) for value in values]
-        for dimension, values in expandedIds.items()
+        dimension: [universe[dimension].RecordClass(**record_dict) for record_dict in record_dicts]
+        for dimension, record_dicts in record_dicts_by_dimension_name.items()
     }
 
 
-def _fillAllKeys(dimension: Dimension, value: Any) -> Mapping[str, Any]:
+def _fillAllKeys(dimension: Dimension, value: Union[str, int]) -> Mapping[str, Union[str, int]]:
     """Create an arbitrary mapping of all required keys for a given dimension
     that do not refer to other dimensions.
 
@@ -215,7 +221,15 @@ def _fillAllKeys(dimension: Dimension, value: Any) -> Mapping[str, Any]:
     expandedValue = {}
     for key in dimension.uniqueKeys:
         if key.nbytes:
-            castType = bytes
+            # For `bytes` fields, we want something that casts at least `str`
+            # and `int` values to bytes and yields b'' when called with no
+            # arguments (as in the except block below).  Unfortunately, the
+            # `bytes` type itself fails for both `str` and `int`, but this
+            # lambda does what we need.  This particularly important for the
+            # skymap dimensions' bytes 'hash' field, which has a unique
+            # constraint; without this, all skymaps would get a hash of b''
+            # and end up conflicting.
+            castType = lambda *args: str(*args).encode()  # noqa: E731
         else:
             castType = key.dtype().python_type
         try:
@@ -229,72 +243,23 @@ def _fillAllKeys(dimension: Dimension, value: Any) -> Mapping[str, Any]:
     return expandedValue
 
 
-def _matchAnyDataId(record: Mapping[str, Any], registry: Registry, dimension: Dimension):
-    """Matches a partial dimension record to an existing record along a
-    specific dimension.
+def _makeRandomDataIdValue(dimension: Dimension) -> Union[int, str]:
+    """Generate a random value of the appropriate type for a data ID key.
 
     Parameters
     ----------
-    record : `dict` [`str`]
-        A mapping representing the record to be matched.
-    registry : `lsst.daf.butler.Registry`
-        The registry with all known dimension records.
-    dimension : `lsst.daf.butler.Dimension`
-        The dimension on which to find a match for ``record``.
-
-    Raises
-    ------
-    RuntimeError
-        Raised if there are no existing records for ``dimension``.
-    """
-    matches = list(registry.queryDimensionRecords(dimension.name))
-    if matches:
-        record[dimension.name] = matches[0].dataId[dimension.name]
-    else:
-        raise RuntimeError(f"No matching values for {dimension.name} found.")
-
-
-def _fillRelationships(
-    dimension: Dimension, dimensionInfo: Mapping[str, Any], existing: Registry
-) -> Mapping[str, Any]:
-    """Create arbitrary mappings from one dimension to all dimensions it
-    depends on.
-
-    Parameters
-    ----------
-    dimension : `lsst.daf.butler.Dimension`
-        The dimension for which to generate relationships.
-    dimensionInfo : `dict` [`str`]
-        A mapping of dimension keys to values.
-    existing : `lsst.daf.butler.Registry`
-        The registry with all previously registered dimensions.
+    dimension : `Dimension`
+        Dimension the value corresponds to.
 
     Returns
     -------
-    filledInfo : `dict` [`str`]
-        A version of ``dimensionInfo`` with extra mappings for any
-        relationships required by ``dimension``. Any relationships already
-        defined in ``dimensionInfo`` are preserved.
-
-    Raises
-    ------
-    ValueError
-        Raised if ``dimension`` depends on a dimension for which no values
-        exist yet.
+    value : `int` or `str`
+        Random value.
     """
-    filledInfo = dimensionInfo.copy()
-    for other in dimension.required:
-        if other != dimension and other.name not in filledInfo:
-            _matchAnyDataId(filledInfo, existing, other)
-    # Do not recurse, to keep the user from having to provide
-    # irrelevant dimensions.
-    for other in dimension.implied:
-        toUpdate = other != dimension and other.name not in filledInfo
-        updatable = other.viewOf is None
-        # Do not run query if either toUpdate or updatable is false
-        if toUpdate and updatable and list(existing.queryDimensionRecords(other)):
-            _matchAnyDataId(filledInfo, existing, other)
-    return filledInfo
+    if dimension.primaryKey.getPythonType() is str:
+        return str(random.randrange(1000))
+    else:
+        return random.randrange(1000)
 
 
 def expandUniqueId(butler: Butler, partialId: Mapping[str, Any]) -> DataCoordinate:
@@ -350,14 +315,80 @@ def expandUniqueId(butler: Butler, partialId: Mapping[str, Any]) -> DataCoordina
         raise ValueError(f"Found {len(dataId)} matches for {partialId}, expected 1.")
 
 
-def addDataIdValue(butler: Butler, dimension: str, value: Any, **related: Any):
-    """Add a new data ID to a repository.
+def _findOrInventDataIdValue(
+    butler: Butler, data_id: dict[str, Union[str, int]], dimension: Dimension
+) -> tuple[Union[str, int], bool]:
+    """Look up an arbitrary value for a dimension that is consistent with a
+    partial data ID that does not specify that dimension, or invent one if no
+    such value exists.
 
-    Related dimensions (e.g., the instrument associated with a detector) may
-    be specified using ``related``. While these keywords are sometimes needed
-    to get self-consistent repositories, you do not need to define
-    relationships you do not use. Any unspecified dimensions will be
-    linked arbitrarily.
+    Parameters
+    ----------
+    butler : `Butler`
+        Butler to use to look up data ID values.
+    data_id : `dict` [ `str`, `str` or `int` ]
+        Dictionary of possibly-related data ID values.
+    dimension : `Dimension`
+        Dimension to obtain a value for.
+
+    Returns
+    -------
+    value : `int` or `str`
+        Value for this dimension.
+    invented : `bool`
+        `True` if the value had to be invented, `False` if a compatible value
+        already existed.
+    """
+    # No values given by caller for this dimension.  See if any exist
+    # in the registry that are consistent with the values of dimensions
+    # we do have:
+    match_data_id = {key: data_id[key] for key in data_id.keys() & dimension.dimensions.names}
+    matches = list(butler.registry.queryDimensionRecords(dimension, dataId=match_data_id).limit(1))
+    if not matches:
+        # Nothing in the registry matches: invent a data ID value
+        # with the right type (actual value does not matter).
+        # We may or may not actually make a record with this; that's
+        # easier to check later.
+        dimension_value = _makeRandomDataIdValue(dimension)
+        return dimension_value, True
+    else:
+        # A record does exist in the registry.  Use its data ID value.
+        return matches[0].dataId[dimension.name], False
+
+
+def _makeDimensionRecordDict(data_id: dict[str, Union[str, int]], dimension: Dimension) -> dict[str, Any]:
+    """Create a dictionary that can be used to build a `DimensionRecord` that
+    is consistent with the given data ID.
+
+    Parameters
+    ----------
+    data_id : `dict` [ `str`, `str` or `int` ]
+        Dictionary that contains values for at least all of
+        ``dimension.dimensions.names`` (the main dimension, its recursive
+        required dependencies, and its non-recursive implied dependencies).
+    dimension : `Dimension`
+        Dimension to build a record dictionary for.
+
+    Returns
+    -------
+    record_dict : `dict` [ `str`, `object` ]
+        Dictionary that can be passed as ``**kwargs`` to this dimensions
+        record class constructor.
+    """
+    # Add the primary key field for this dimension.
+    record_dict: dict[str, Any] = {dimension.primaryKey.name: data_id[dimension.name]}
+    # Define secondary keys (e.g., detector name given detector id)
+    record_dict.update(_fillAllKeys(dimension, data_id[dimension.name]))
+    # Set the foreign key values for any related dimensions that should
+    # appear in the record.
+    for related_dimension in dimension.dimensions:
+        if related_dimension.name != dimension.name:
+            record_dict[related_dimension.name] = data_id[related_dimension.name]
+    return record_dict
+
+
+def addDataIdValue(butler: Butler, dimension: str, value: Union[str, int], **related: Union[str, int]):
+    """Add the records that back a new data ID to a repository.
 
     Parameters
     ----------
@@ -372,13 +403,21 @@ def addDataIdValue(butler: Butler, dimension: str, value: Any, **related: Any):
 
     Notes
     -----
+    Related dimensions (e.g., the instrument associated with a detector) may be
+    specified using ``related``, which requires a value for those dimensions to
+    have been added to the repository already (generally with a previous call
+    to `addDataIdValue`.  Any dependencies of the given dimension that are not
+    included in ``related`` will be linked to existing values arbitrarily, and
+    (for implied dependencies only) created and also inserted into the registry
+    if they do not exist.  Values for required dimensions and those given in
+    ``related`` are never created.
+
     Because this function creates filler data, it is only suitable for test
     repositories. It should not be used for repositories intended for real data
     processing or analysis, which have known dimension values.
 
     Examples
     --------
-
     See the guide on :ref:`using-butler-in-tests-make-repo` for usage examples.
     """
     # Example is not doctest, because it's probably unsafe to create even an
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -107,7 +107,8 @@ def makeTestRepo(
     butler = Butler(newConfig, writeable=True)
     dimensionRecords = _makeRecords(dataIds, butler.registry.dimensions)
     for dimension, records in dimensionRecords.items():
-        butler.registry.insertDimensionData(dimension, *records)
+        if butler.registry.dimensions[dimension].viewOf is None:
+            butler.registry.insertDimensionData(dimension, *records)
```
kfindeisen, line 112: Can you explain why this change is needed? I thought I'd already included support for registering bands.

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -183,7 +184,7 @@ def _makeRecords(dataIds: Mapping[str, Iterable], universe: DimensionUniverse) -
             # Do not recurse, to keep the user from having to provide
             # irrelevant dimensions
             for other in dimension.implied:
-                if other != dimension and other.name in expandedIds and other.viewOf is None:
+                if other.name in expandedIds:
```
kfindeisen: If you're doing what I think you're doing, then the comment is no longer correct.

kfindeisen: Part of the reason `_testRepo.py` exists in the first place is to allow `PipelineTask` developers to set up a test environment without any deep knowledge of how Middleware works, including the sometimes unintuitive relationships between dimensions and the special treatment of `band`.

Would it be possible to generate a value (e.g., `band = physical_filter`) analogously to what `_fillAllKeys` does, instead of forcing the user to figure out why their test setup is giving cryptic errors both when they try to create a band and when they don't?

kfindeisen: Do similar changes need to be made to `addDataIdValue`? It's a bit surprising that all the code changes are in the old system.

tests/test_testRepo.py
```diff
@@ -53,7 +53,7 @@ def setUpClass(cls):
         addDataIdValue(cls.creatorButler, "instrument", "notACam")
         addDataIdValue(cls.creatorButler, "instrument", "dummyCam")
         addDataIdValue(cls.creatorButler, "physical_filter", "k2020", band="k", instrument="notACam")
-        addDataIdValue(cls.creatorButler, "physical_filter", "l2019", instrument="dummyCam")
+        addDataIdValue(cls.creatorButler, "physical_filter", "l2019", band="l", instrument="dummyCam")
```
kfindeisen: I think this line is actually redundant now, since its purpose was to test that you could omit `band`. Does the test suite break if you delete this line?

kfindeisen (commented):
I think I understand the problem, but this feels more like a workaround than a solution. The test support code previously *made up* values as necessary to get valid data IDs and dimension records; why can't the same approach work for implied dimensions?

kfindeisen:
One other comment: if the system can't be improved over what's here, then https://pipelines.lsst.io/v/daily/modules/lsst.daf.butler/use-in-tests.html#creating-and-populating-a-repository probably needs to be updated with a sentence saying that you need to define at least one value for most dimensions (and the examples need vetting to see if they're still valid).

python/lsst/daf/butler/tests/_testRepo.py
TallJimbo: It may be that it was already possible via some other method, but I made this change because passing band key in the `dataIds` argument did result in an attempt to insert it, which failed.  I suspect that I missed a way to make it unnecessary to pass a band key here, even after the NOT NULL constraint lands - and I'll discuss that in a thread for another comment - but even if that's the case it seems like it might still be good to permit a `band` key to be passed here, and allow its values to be associated with other dimensions like `physical_filter`, in the same way other relationships are generated arbitrarily.

python/lsst/daf/butler/tests/_testRepo.py
TallJimbo: I agree the comment is incorrect, and I'll at least fix that (if I can't make it remain correct).

I think there may indeed be an opportunity to generate values for the missing dimensions automatically, and avoid the other branches for this ticket, and I just didn't understand where to put it.  My guess would be to have `_makeRecords` start by ensuring that its input mapping has keys for all dimensions implied by any of the dimensions it is given, using that `_fillAllKeys`-like logic.  I'll attempt that later today, but if you can confirm or reject that as a reasonable approach first I'd appreciate it.

> Do similar changes need to be made to `addDataIdValue`? It's a bit surprising that all the code changes are in the old system.

If I can fix the other code path as you've requested, then I think that means I should fix this one, too.  I suspect the reason I didn't have to was because it already made it possible (but after the NOT NULL constraint lands, necessary) to pass implied values by including them in the data ID key, so I just went around making changes in the tests like:
```
        -addDataIdValue(cls.creatorButler, "physical_filter", "l2019", instrument="dummyCam")
        +addDataIdValue(cls.creatorButler, "physical_filter", "l2019", band="l", instrument="dummyCam")
```
And I think the right place for that fix is in `_matchAnyDataId`, which could create (with `_findAllKeys`-like logic) and insert a new dimension record when there isn't one, instead of raising.  Once again, confirmation that that approach makes sense would be appreciated.

tests/test_testRepo.py
TallJimbo: Yes, and it sounds like the preferred approach is to revert this change and try to make the test pass without it (as per the other thread).

kfindeisen: If the test breaks, then the line is needed either way.

python/lsst/daf/butler/tests/_testRepo.py
kfindeisen: >  My guess would be to have `_makeRecords` start by ensuring that its input mapping has keys for all dimensions implied by any of the dimensions it is given, using that `_fillAllKeys`-like logic. I'll attempt that later today, but if you can confirm or reject that as a reasonable approach first I'd appreciate it.

Yes, I think so. Unfortunately, I haven't looked at this code since last year, and I wrote it with a very vague idea of how the dimensions system works, so I can't offer any special insights.

> And I think the right place for that fix is in `_matchAnyDataId`, which could create (with `_findAllKeys`-like logic) and insert a new dimension record when there isn't one, instead of raising. Once again, confirmation that that approach makes sense would be appreciated.

I'm a little nervous about having `addDataIdValue` create dimension records for other dimensions (looks like it would be `physical_filter` in practice), but I think it's the least of the available evils. Please add a note to the docstring that it may create values for dimensions other than `dimension`.

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -394,6 +395,11 @@ def addDataIdValue(butler: Butler, dimension: str, value: Any, **related: Any):
             f"in {fullDimension.required | fullDimension.implied}"
         )
 
+    if fullDimension.viewOf:
+        # Nothing to do; this dimension's records aren't actually saved in the
+        # database directly anyway (and caller shouldn't have to know this).
+        return
```
kfindeisen: Clear and concise! 👍

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -57,11 +57,12 @@ def makeTestRepo(
...
     dataIds : `~collections.abc.Mapping` [`str`, `iterable`], optional
-        A mapping keyed by the dimensions used in the test. Each value
-        is an iterable of names for that dimension (e.g., detector IDs for
-        `"detector"`). Related dimensions (e.g., instruments and detectors)
-        are linked arbitrarily. This parameter is provided for compatibility
-        with old code; newer code should make the repository, then call
+        A mapping keyed by the dimensions used in the test. Each value is an
+        iterable of names for that dimension (e.g., detector IDs for
+        `"detector"`). Related dimensions (e.g., instruments and detectors) are
+        linked arbitrarily, with values created randomly when needed. This
```
kfindeisen: I would omit "randomly" from this statement; I would prefer that the user make no assumptions whatsoever about what the values do or don't look like.

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -165,36 +166,41 @@ def _makeRecords(dataIds: Mapping[str, Iterable], universe: DimensionUniverse) -
...
-        expandedIds[name] = []
+
+    # Create values for all dimensions that are (recursive) required or implied
+    # dependencies of the given ones.
+    complete_data_id_values = {}
+    for dimension in universe.extract(dataIds.keys()):
+        if dimension.name in dataIds:
+            complete_data_id_values[dimension.name] = list(dataIds[dimension.name])
+        if dimension.name not in complete_data_id_values:
+            complete_data_id_values[dimension.name] = [_makeRandomDataIdValue(dimension)]
```
kfindeisen, line 177: I find it very strange that Codecov is claiming that neither of these blocks is entered in tests. That shouldn't be the case if there's even a single test with non-empty `dataIds`.

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -165,36 +166,41 @@ def _makeRecords(dataIds: Mapping[str, Iterable], universe: DimensionUniverse) -
...
     return {
-        dimension: [universe[dimension].RecordClass(**value) for value in values]
-        for dimension, values in expandedIds.items()
+        dimension: [universe[dimension].RecordClass(**record_dict) for record_dict in record_dicts]
+        for dimension, record_dicts in record_dicts_by_dimension_name.items()
     }
 
 
-def _fillAllKeys(dimension: Dimension, value: Any) -> Mapping[str, Any]:
+def _fillAllKeys(dimension: Dimension, value: Union[str, int]) -> Mapping[str, Union[str, int]]:
```
kfindeisen, line 203: Can you actually guarantee that all dimensions (including any added in the future) will be either strings or integers?

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -215,7 +221,11 @@ def _fillAllKeys(dimension: Dimension, value: Any) -> Mapping[str, Any]:
     expandedValue = {}
     for key in dimension.uniqueKeys:
         if key.nbytes:
-            castType = bytes
+            # bytes(str(...)) would normally fail, bringing us to the
+            # except clause below, and that in particular is a problem because
+            # of the skymap dimensions' bytes 'hash' field, which has a
+            # unique constraint.
+            castType = lambda x: str(x).encode()  # noqa: E731
```
kfindeisen: Can you explain why this is needed? Making `castType` be anything other than a type makes the subsequent code somewhat misleading; I'm not sure if the "default-construction" fallback still works, either.

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -165,36 +166,41 @@ def _makeRecords(dataIds: Mapping[str, Iterable], universe: DimensionUniverse) -
...
-                    relation = expandedIds[other.name][0]
-                    value[other.name] = relation[other.primaryKey.name]
-            # Do not recurse, to keep the user from having to provide
-            # irrelevant dimensions
-            for other in dimension.implied:
-                if other.name in expandedIds:
-                    relation = expandedIds[other.name][0]
-                    value[other.name] = relation[other.primaryKey.name]
+                    relation = record_dicts_by_dimension_name[other.name][0]
+                    record_dict[other.name] = relation[other.primaryKey.name]
```
kfindeisen: I think I'm a bit confused about the scope of the change. The old code made a distinction between required dimensions (assumed to be provided by the user) and implied dimensions (hidden from the user). The new code tries to hide everything. Is there a reason continuing to require required dimensions is now impractical?

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -350,14 +311,14 @@ def expandUniqueId(butler: Butler, partialId: Mapping[str, Any]) -> DataCoordina
         raise ValueError(f"Found {len(dataId)} matches for {partialId}, expected 1.")
 
 
-def addDataIdValue(butler: Butler, dimension: str, value: Any, **related: Any):
+def addDataIdValue(butler: Butler, dimension: str, value: Union[str, int], **related: Union[str, int]):
```
kfindeisen: Same question about how safe it is to assume it's only these two types.

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -350,14 +311,14 @@ def expandUniqueId(butler: Butler, partialId: Mapping[str, Any]) -> DataCoordina
...
-    Related dimensions (e.g., the instrument associated with a detector) may
-    be specified using ``related``. While these keywords are sometimes needed
-    to get self-consistent repositories, you do not need to define
-    relationships you do not use. Any unspecified dimensions will be
-    linked arbitrarily.
+    Related dimensions (e.g., the instrument associated with a detector) may be
+    specified using ``related``. While these keywords are sometimes needed to
+    get self-consistent repositories, you do not need to define relationships
+    you do not use. Any unspecified dimensions will be linked arbitrarily,
+    withe values created randomly when needed.
```
kfindeisen: Typo, "withe". I'd prefer not to say *how* the values are created.

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -388,32 +349,71 @@ def addDataIdValue(butler: Butler, dimension: str, value: Any, **related: Any):
     except KeyError as e:
         raise ValueError from e
     # Bad keys ignored by registry code
-    extraKeys = related.keys() - (fullDimension.required | fullDimension.implied)
+    extraKeys = related.keys() - fullDimension.graph.dimensions.names
     if extraKeys:
         raise ValueError(
-            f"Unexpected keywords {extraKeys} not found "
-            f"in {fullDimension.required | fullDimension.implied}"
+            f"Unexpected keywords {extraKeys} not found " f"in {fullDimension.graph.dimensions.names}"
```
kfindeisen: Weird formatting -- please make this either a single string, or on two lines.

python/lsst/daf/butler/tests/_testRepo.py
kfindeisen: I'm worried the existing text may not make the consequences clear enough. Can the "Notes" section say something like "values for dimensions other than `dimension` may be created if necessary to create a valid instance of `dimension`".

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -388,32 +349,71 @@ def addDataIdValue(butler: Butler, dimension: str, value: Any, **related: Any):
...
+    data_id.update(related)
+
+    # Compute the set of all dimensions that these recursively depend on.
+    all_dimensions = butler.registry.dimensions.extract(data_id.keys())
+
+    # Create dicts that will become DimensionRecords for all of these data IDs.
+    # This iteration is guaranteed to be in topological order, so we can count
+    # on new data ID values being invented before they are needed.
+    record_dicts_by_dimension: dict[Dimension, dict[str, Any]] = {}
+    for dimension_obj in all_dimensions:
```
kfindeisen, line 448: I'm finding this loop very hard to follow because of the lack of distinction between the dimension we are actually trying to create (`data_id`) and the dimensions that are incidental to that and probably already exist (`dimension_obj`, `dimension_value`, etc.). I especially don't understand why you are carefully constructing dimension records for existing dimensions.

Would it be possible to factor out some of these blocks into their own functions (making the code a bit more self-documenting in the process)? It looks like the body of `if dimension_value is None` and the code below the view check are each trying to do one thing. This should reduce the number of variables the human needs to keep track of at a time.

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -388,32 +349,71 @@ def addDataIdValue(butler: Butler, dimension: str, value: Any, **related: Any):
...
+        record_dict: dict[str, Any] = {dimension_obj.primaryKey.name: dimension_value}
+        # Define secondary keys (e.g., detector name given detector id)
+        record_dict.update(_fillAllKeys(dimension_obj, dimension_value))
+        # Set the foreign key values for any related dimensions that should
+        # appear in the record.
+        for related_dimension in dimension_obj.dimensions:
+            if related_dimension.name != dimension_obj.name:
+                record_dict[related_dimension.name] = data_id[related_dimension.name]
+
+        record_dicts_by_dimension[dimension_obj] = record_dict
```
kfindeisen: Would it be simpler to just call `addDataIdValue` recursively for missing required dimensions? I'm worried about `syncDimensionData` either failing, or worse, changing, records for dimensions that existed before the call.

tests/test_testRepo.py
```diff
@@ -126,15 +126,16 @@ def testAddDataIdValue(self):
         with self.assertRaises(ValueError):
             addDataIdValue(self.butler, "detector", 101, nonsenseField="string")
 
-        # Keywords imply different instruments
-        with self.assertRaises(RuntimeError):
-            addDataIdValue(self.butler, "exposure", 101, instrument="dummyCam", physical_filter="k2020")
```
kfindeisen: I don't understand why this test case has been removed. The extra dimensions are inconsistent (`"k2020"` implies `"notACam"`, not `"dummyCam"`), so it should still fail.

tests/test_testRepo.py
```diff
@@ -126,15 +126,16 @@ def testAddDataIdValue(self):
...
-            addDataIdValue(self.butler, "tract", 42)
-        with self.assertRaises(RuntimeError):
-            addDataIdValue(self.butler, "tract", 43, skymap="map")
+        # No skymap defined; should be created automatically.
+        addDataIdValue(self.butler, "tract", 42)
+        (tract_data_id,) = set(self.butler.registry.queryDataIds(["tract"]).limit(1))
+        self.assertEqual(tract_data_id["tract"], 42)
+        # And then a new one created here, since the now-existing one should
+        # have a random name that isn't 'map'.
+        addDataIdValue(self.butler, "tract", 43, skymap="map")
```
kfindeisen: Please do not allow this behavior; it will be too surprising for users. A user who specified a specific skymap is clearly assuming it exists.

kfindeisen (changes requested):
Apologies for dragging this out, but while the code itself looks reasonable (if hard to understand), the test cases imply behavior that sharply deviates from [what's been documented](https://pipelines.lsst.io/v/daily/modules/lsst.daf.butler/use-in-tests.html#creating-and-populating-a-repository).

I'm particularly concerned about new values being created for dimensions where the user was almost certainly expecting to use existing ones and/or overwriting dimensions the user previously created. The behavioral changes turn the purpose of `addDataIdValue` from "create one dimension, and sweep the implementation details under the rug" to "create multiple dimensions anchored at a particular node".

(Also, in the future, please put formatting and variable renames on different commits from changes of substance. The blend made it very hard for me to reconstruct how the new code did or did not relate to the old code.)

python/lsst/daf/butler/tests/_testRepo.py
timj: That's black realizing that it can now fit it in one line, but it does not attempt to merge the strings. It's a black oddity that will happen and we need to keep an eye out for.

python/lsst/daf/butler/tests/_testRepo.py
timj: All current ones are. Floating point dimensions are problematic and we've had discussion about what to do if we have spectrographs that are tunable -- what's the equivalent of physical filter for a heterodyne receiver that can tune to arbitrary frequency and have configurable bandwidth? Should the central tuning frequency be used in the dataId? If it's 230.23 GHz should it be converted to an integer of 230230 for that instrument? Does every possible floating point value have to have a registered dimension record or can it be more flexible?

TallJimbo: Yes, we can guarantee that - or at least it's an assumption also made elsewhere in daf_butler, and encoded in other type annotations like this one.

python/lsst/daf/butler/tests/_testRepo.py
TallJimbo: Without this, the `_findAllKeys` logic for `bytes` fields always falls through to the `except` clause below, because `bytes(int)` `bytes(str)` both raise, and hence the value is always just `b''`.  When there is a unique constraint on a bytes field - as there is for the `skymap` dimension, and when more than one record in inserted (which is more likely now that dimension values are sometimes invented as needed, and one of the tests now does it), we get a constraint violation.

I think you're right that the fallback no longer works with this, though; I'll see about fixing that.

python/lsst/daf/butler/tests/_testRepo.py
TallJimbo: A version that invents required dimensions as needed as well was a bit easier to implement, and I figured it'd actually be desirable behavior if the overall goal is keeping tests from caring about the details of the dimension data model.

python/lsst/daf/butler/tests/_testRepo.py
TallJimbo: In a system where any missing dependencies are created as needed, I don't think there is a distinction between the original dimension identified and any others anymore; it's unfortunate that the interface continues to suggest that there is, but I don't want to go through the deprecation-cycle effort of changing that now, either.

>  especially don't understand why you are carefully constructing dimension records for existing dimensions.

I don't think I am; I am fetching any records that exist that are compatible with the given data IDs.  Calling `syncDimensionData` on those later should be a no-op (it's an expensive enough one that I would try to avoid it in non-test code, but here I figured that was better than having another local variable to track which records are new and which are existing).

> Would it be possible to factor out some of these blocks into their own functions

:+1:

python/lsst/daf/butler/tests/_testRepo.py
TallJimbo: `syncDimensionData` definitely will not change existing records, and I think if it does fail it does still say something important.  That said, I'm not sure I've left it possible for it to fail - I have not been able to trigger the previous expected failures in the tests without breaking something else (more on that in other threads).

python/lsst/daf/butler/tests/_testRepo.py
kfindeisen: Got it, so it's a bug in my code. However, can you reword the comment to explicitly say that this is, logically, a conversion to `bytes`? That won't be clear once the diff is gone.

tests/test_testRepo.py
TallJimbo: I played around with a number of adjustments to try to make this work, and eventually I convinced myself that if the system is allowed to create new dimension values, there's no way to distinguish between this case and missing implied dimension records more generally, at least not without the kind of exhaustive search that the butler deliberately disables because in real-world contexts it always indicates a mistake (if someone queries for data IDs with `detector=4` in a real repo without specifying the instrument, they're not asking for to cartesian-product iterate over all instruments that have a detector 4).

Switching to only allowing implied dimension records to be created might mitigate this somewhat, but it wouldn't save this test (that's one of the things I tried), because `physical_filter` is an implied dimension of `exposure`, so creating one that matches the undesired `instrument` with the given filter name would still be permitted.

So I think the choice is between this behavior and reverting back to the original version of this branch, where downstream tests were still required to provide implied dimension value explicitly.

tests/test_testRepo.py
TallJimbo: I think this is a case that would be addressed by permitting only implied dimension records to be created.

tests/test_testRepo.py
TallJimbo: Ah!  Nevermind - the thread just below this one made me realize the distinction you want made: you do not want records created for any dimension that is referenced explicitly in the call to `addDataIdValue`; those are asserted to already exist, and only those not mentioned should be created automatically.  It seems obvious now; I was too deep in the weeds (having basically dropped the original set of dimensions given on the floor in favor of its expansion) to see it before.  I'll be back with another of this shortly.

python/lsst/daf/butler/tests/_testRepo.py
TallJimbo: It looks like there just are no tests of `dataIds` in daf_butler; the ones here all use `addDataIdValue` instead, which fits much better with the structure of the test, since it tries to use only one repo created in `setUpClass` (for performance reasons).  There are tests in downstream packages that use `dataIds`.

python/lsst/daf/butler/tests/_testRepo.py
TallJimbo: I did look a bit into calling `addDataIdValue` recursively just now, and I think it'd a bit messier in the end, because we'd have to add new arguments to distinguish between the original contents of `**related` and later derived ones.

tests/test_testRepo.py
TallJimbo: Ok, done!  This let me restore the original test to exactly what it was before.

TallJimbo (commented):
Ok, I've just pushed a change to address the last round of comments (will squash later), and I'm pretty happy with it.  I also hit "resolved" on a few threads here where the resolution was obvious.

The big good news is that the original test now passes, so this is really the minimal change needed to address the upcoming problem.

The code coverage is apparently not great:
- the existing coverage wasn't as good as either of us expected, though there is coverage in downstream packages, and what's missing here is clearly documented as the "old interface";
- the new logic that adds implied dimension records isn't tested here _yet_, but it will be covered by the existing tests as soon as the new constraint lands.

I think that's enough for me, even if it's not ideal.

Finally, apologies for mixing up the renames with the behavior changes - separating those is usually very much my policy as well, but in this case I ended up doing the renames as I worked out what the code was doing, and by the time I was done I figured the functions I touched were pretty much completely rewritten so the quality of the diff wasn't as important (though it seems the PR diff pretty thoroughly obfuscated that).

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -60,9 +60,9 @@ def makeTestRepo(
         A mapping keyed by the dimensions used in the test. Each value is an
         iterable of names for that dimension (e.g., detector IDs for
         `"detector"`). Related dimensions (e.g., instruments and detectors) are
-        linked arbitrarily, with values created randomly when needed. This
-        parameter is provided for compatibility with old code; newer code
-        should make the repository, then call
+        linked arbitrarily, with values created for implie dimensions only when
```
kfindeisen:
```suggestion
        linked arbitrarily, with values created for implied dimensions only when
```

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -221,11 +221,13 @@ def _fillAllKeys(dimension: Dimension, value: Union[str, int]) -> Mapping[str, U
...
-            # except clause below, and that in particular is a problem because
-            # of the skymap dimensions' bytes 'hash' field, which has a
-            # unique constraint.
-            castType = lambda x: str(x).encode()  # noqa: E731
+            # bytes(str(...)) would normally fail, bringing us to the except
+            # clause below, and that in particular is a problem because of the
+            # skymap dimensions' bytes 'hash' field, which has a unique
+            # constraint.
+            # This lambda casts `str` and `int` values to bytes and yields b''
+            # when called with no arguments (as in the except block below).
```
kfindeisen: I think it would make more sense to say this first, and *then* justify why you're using a lambda to do it.

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -311,14 +313,88 @@ def expandUniqueId(butler: Butler, partialId: Mapping[str, Any]) -> DataCoordina
...
+
 def addDataIdValue(butler: Butler, dimension: str, value: Union[str, int], **related: Union[str, int]):
     """Add a new data ID to a repository.
 
     Related dimensions (e.g., the instrument associated with a detector) may be
     specified using ``related``. While these keywords are sometimes needed to
     get self-consistent repositories, you do not need to define relationships
-    you do not use. Any unspecified dimensions will be linked arbitrarily,
-    withe values created randomly when needed.
+    you do not use.  Any unspecified dimensions implied will be linked
```
kfindeisen: "Any unspecified implied dimensions"? I'm having trouble parsing the sentence as-is.

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -337,6 +413,13 @@ def addDataIdValue(butler: Butler, dimension: str, value: Union[str, int], **rel
     repositories. It should not be used for repositories intended for real data
     processing or analysis, which have known dimension values.
 
+    Values for dimensions other than ``dimension`` may be created if necessary,
+    in order to construct records for ``dimension`` that satisfy database
+    foreign key constraints.  This is never done for required dimensions (those
+    that are also needed disambiguate any value for ``dimension``) or those
+    given as keys in ``related``; records fo hose dimensions must already
```
kfindeisen:
```suggestion
    given as keys in ``related``; records for those dimensions must already
```

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -311,14 +313,88 @@ def expandUniqueId(butler: Butler, partialId: Mapping[str, Any]) -> DataCoordina
...
+    if not matches:
+        # Nothing in the registry matches: invent a data ID value
+        # with the right type (actual value does not matter).
+        # We may or may not actually make a record with this; that's
+        # easier to check later.
+        dimension_value = _makeRandomDataIdValue(dimension)
+        return dimension_value, True
+    else:
+        # A record does exist in the registry.  Use its data ID value,
+        # and skip the rest of the record-fabrication logic.
```
kfindeisen: The "skip" part of this statement is no longer applicable, since it's outside the scope of this function.

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -311,14 +313,88 @@ def expandUniqueId(butler: Butler, partialId: Mapping[str, Any]) -> DataCoordina
...
+def _makeDimensionRecordDict(data_id: dict[str, Union[str, int]], dimension: Dimension) -> dict[str, Any]:
+    """Create a dictionary that can be used to build a `DimensionRecord` that
+    is consistent with the given data ID.
+
+    Parameters
+    ----------
+    data_id : `dict` [ `str`, `str` or `int` ]
+        Dictionary that contains values for all of
+        ``dimension.dimensions.names`` (the main dimension, its recursive
+        required dependencies, and its non-recursive implied dependencies).
```
kfindeisen, line 368: Are the implied dependencies non-recursive? I thought satisfying dependencies recursively was the point of this PR.

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -370,45 +453,33 @@ def addDataIdValue(butler: Butler, dimension: str, value: Union[str, int], **rel
...
+            # Caller passed in a value of this dimension explicitly, but it
+            # isn't the primary dimension they asked to have a record created
+            # for.  That means they expect this record to already exist.
+            continue
+        if dimension_obj != fullDimension and dimension_obj in all_dimensions.required:
+            # We also don't want to automatically create new dimension records
+            # for required dimensions (except for the main dimension the caller
+            # asked for); those are also asserted by the caller to already
+            # exist.
+            continue
```
kfindeisen, line 465: This was the original behavior, but I thought (based on your answer to https://github.com/lsst/daf_butler/pull/680#discussion_r871800711) that this wouldn't work anymore. Given the incoming not-null constraints, what "non-obvious" dimensions will people need to specify that they didn't before? `physical_filter` for `exposure`?

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -370,45 +453,33 @@ def addDataIdValue(butler: Butler, dimension: str, value: Union[str, int], **rel
...
-        record_dicts_by_dimension[dimension_obj] = record_dict
+        record_dicts_by_dimension[dimension_obj] = _makeDimensionRecordDict(data_id, dimension_obj)
 
     # Sync those dimension record dictionaries with the database.
     for dimension_obj, record_dict in record_dicts_by_dimension.items():
         record = dimension_obj.RecordClass(**record_dict)
         try:
             butler.registry.syncDimensionData(dimension_obj, record)
-        except ConflictingDefinitionError as e:
+        except sqlalchemy.exc.IntegrityError as e:
```
kfindeisen, line 477: I don't understand this change. Was it a mis-specification of `syncDimensionData`?

tests/test_testRepo.py
```diff
@@ -126,16 +126,15 @@ def testAddDataIdValue(self):
...
-        )
+        # Keywords imply different instruments
+        with self.assertRaises(RuntimeError):
+            addDataIdValue(self.butler, "exposure", 101, instrument="dummyCam", physical_filter="k2020")
+
+        # No skymap defined
+        with self.assertRaises(RuntimeError):
+            addDataIdValue(self.butler, "tract", 42)
+        with self.assertRaises(RuntimeError):
+            addDataIdValue(self.butler, "tract", 43, skymap="map")
```
kfindeisen: Maybe this should have its own comment:
```suggestion
        # Didn't create skymap "map" first.
        with self.assertRaises(RuntimeError):
            addDataIdValue(self.butler, "tract", 43, skymap="map")
```

kfindeisen (approved):
Looks good, mostly documentation comments left. Thank you for your patience!

My biggest outstanding question is about exactly which dependencies do  or don't get invented. I just want to make sure the current solution satisfies both our requirements (that existing tests will not break when the new registry constraints come in, and that pipeline task developers don't need more knowledge of the dimensions system than they needed to write the task in the first place).

python/lsst/daf/butler/tests/_testRepo.py
TallJimbo: The record for a particular dimension only includes direct (non-recursive) implied dimension values (e.g. `visit` has a column for `physical_filter`, but not `band), which is why that's all that's needed in this method.  But the records for those implied dependencies will then include columns for their implied dependencies, so we do need to recurse elsewhere.

python/lsst/daf/butler/tests/_testRepo.py
TallJimbo: I've rewritten this block and merged it into to the `Notes` section, as it was getting pretty long.  Here's the new complete `Notes` text (the second paragraph is unchanged):

> Related dimensions (e.g., the instrument associated with a detector) may be
    specified using ``related``, which requires a value for those dimensions to
    have been added to the repository already (generally with a previous call
    to `addDataIdValue`.  Any dependencies of the given dimension that are not
    included in ``related`` will be linked to existing values arbitrarily, and
    (for implied dependencies only) created and also inserted into the registry
    if they do not exist.  Values for required dimensions and those given in
    ``related`` are never created.
>
>    Because this function creates filler data, it is only suitable for test
    repositories. It should not be used for repositories intended for real data
    processing or analysis, which have known dimension values.

python/lsst/daf/butler/tests/_testRepo.py
TallJimbo: The original behavior was that users did not need to provide values for implied dependencies, and the foreign key columns with the values for those dimensions were set to `None` / `NULL`.

The new behavior is that users still do not need to provide values for implied dependencies, but this is now because those missing values are generated randomly and records for them are inserted into the registry, allowing the foreign key columns that point to them to be not-`NULL` and have their constraints satisfied.

python/lsst/daf/butler/tests/_testRepo.py
TallJimbo: Short version: this is reverting the exception type back to `sqlalchemy.exc.IntegrityError`, which it was originally and probably should have been throughout my work on the branch, but in the middle I got confused about what the likely failure mode was.

Long version: `IntegrityError` is the right type for low-level foreign key and unique constraints (in this context, foreign key violations are much more likely, and what the re-raise here anticipates, but sqlalchemy makes it really hard to tell which, which is the main reason we don't re-raise at a lower level).  The changes on this branch that have not been dropped made that kind of constraint violation impossible, but I wasn't able to convince myself that a different error condition could be ruled out, in which the record already exists but it has different content, and `ConflictingDefinitionError` is raised.  So I switched the except block to that.  With the final version of the logic here that's more like the original, foreign key violations (due to missing required dependencies) are once again possible, and the original re-raise makes sense again.

python/lsst/daf/butler/tests/_testRepo.py
kfindeisen: My mistake, I'd forgotten the meaning of "required" (as opposed to "implied") in the context of dimensions.

PR #681, opened 2022-05-02
DM-33148: avoid conflicts due to existing dimension records in import

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/transfers/_yaml.py
+++ b/python/lsst/daf/butler/transfers/_yaml.py
@@ -372,7 +372,12 @@ class YamlRepoImportBackend(RepoImportBackend):
         for element, dimensionRecords in self.dimensions.items():
             if skip_dimensions and element in skip_dimensions:
                 continue
-            self.registry.insertDimensionData(element, *dimensionRecords)
+            # Using skip_existing=True here assumes that the records in the
+            # database are either equivalent or at least preferable to the ones
+            # being imported.  It'd be ideal to check that, but that would mean
+            # using syncDimensionData, which is not vectorized and is hence
+            # unacceptably slo.
+            self.registry.insertDimensionData(element, *dimensionRecords, skip_existing=True)
         # FileDatasets to ingest into the datastore (in bulk):
         fileDatasets = []
         for (datasetTypeName, run), records in self.datasets.items():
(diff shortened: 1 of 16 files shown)


python/lsst/daf/butler/transfers/_yaml.py
```diff
@@ -372,7 +372,7 @@ def load(
         for element, dimensionRecords in self.dimensions.items():
             if skip_dimensions and element in skip_dimensions:
                 continue
-            self.registry.insertDimensionData(element, *dimensionRecords)
+            self.registry.insertDimensionData(element, *dimensionRecords, skip_existing=True)
```
timj, line 380: Can there be a comment here to say that we are using skip_existing because we are assuming that records that are being imported from an export from another registry are assumed to be trustworthy and not to have been manually altered and we are doing this for speed reasons because sync would be too slow?

tests/test_simpleButler.py
```diff
@@ -204,11 +204,8 @@ def testDatasetImportTwice(self):
             butler2 = self.makeButler(writeable=True)
             # Import it once.
             butler2.import_(filename=file.name)
-            # Import it again, but ignore all dimensions
-            dimensions = set(
-                dimension.name for dimension in butler2.registry.dimensions.getStaticDimensions()
-            )
-            butler2.import_(filename=file.name, skip_dimensions=dimensions)
+            # Import it again
+            butler2.import_(filename=file.name)
```
timj, line 208: I wonder why the code coverage tool does not think this line ran...

timj (approved):
Looks good to me. Only a couple of minor comments.

tests/test_simpleButler.py
TallJimbo: Good catch!  It's right; I tried sabotaging it and it seems to always be skipped.

At the top of the base test case in this file, we have:
```
datasetsIdType = int
```
and then this test method has:
```
if self.datasetsIdType is not uuid.UUID:
    self.skipTest("This test can only work for UUIDs")
```
but at the bottom we have two derived test cases that both set
```
datasetsIdType = uuid.UUID
```
so it's not clear to me what's going on.  I'll see if I can figure it out, but ideas are most welcome.

TallJimbo: Ha, nope, nothing so complicated as that.  I had accidentally renamed the test method to `tesImportTwice`.

timj: 😄 On the plus side code coverage caught it.

PR #682, opened 2022-05-05
DM-34483: Fix DataCoordinate.timespan method

andy-slac:
Add protection for a case when there are no timespans found.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/dimensions/_coordinate.py
+++ b/python/lsst/daf/butler/core/dimensions/_coordinate.py
@@ -623,7 +623,12 @@ class DataCoordinate(NamedKeyMapping[Dimension, DataIdValue]):
                 return None
             else:
                 timespans.append(record.timespan)
-        return Timespan.intersection(*timespans)
+        if not timespans:
+            return None
+        elif len(timespans) == 1:
+            return timespans[0]
+        else:
+            return Timespan.intersection(*timespans)
 
     def pack(self, name: str, *, returnMaxBits: bool = False) -> Union[Tuple[int, int], int]:
         """Pack this data ID into an integer.
--- a/tests/test_dimensions.py
+++ b/tests/test_dimensions.py
@@ -658,6 +658,11 @@ class DataCoordinateTestCase(unittest.TestCase):
             self.assertIsNotNone(dataId.timespan)
             self.assertEqual(dataId.graph.temporal.names, {"observation_timespans"})
             self.assertEqual(dataId.timespan, dataId.records["visit"].timespan)
+        # Also test the case for non-temporal DataIds.
+        for dataId in self.randomDataIds(n=4).subset(
+            DimensionGraph(self.allDataIds.universe, names=["patch"])
+        ):
+            self.assertIsNone(dataId.timespan)
 
     def testIterableStatusFlags(self):
         """Test that DataCoordinateSet and DataCoordinateSequence compute


python/lsst/daf/butler/core/dimensions/_coordinate.py
```diff
@@ -623,7 +623,12 @@ def timespan(self) -> Optional[Timespan]:
                 return None
             else:
                 timespans.append(record.timespan)
-        return Timespan.intersection(*timespans)
+        if not timespans:
+            return None
+        elif len(timespans) == 1:
+            return timespans[0]
+        else:
+            return Timespan.intersection(*timespans)
```
timj, line 631: I am intrigued as to why it is that this line is not tested given that this is the line that was used before. That implies to me that (1) we are not testing this code with two times and (2) `Timespan.intersection` works with one time.

timj (approved):
I am confused by the test coverage. Can we easily add some tests for the new options?

python/lsst/daf/butler/core/dimensions/_coordinate.py
andy-slac: I think both are true, I do not know where this is being tested, but obviously we are only testing it with a single timespan. I'll try to figure out how it can be tested. `Timespan.intersection` should definitely work with one timespan or more than one, as this is unbound method it treats first argument as `self`. I actually wanted to rewrite it as 

    timespans[0].intersection(*timespans[1:])

but it looks uglier.

TallJimbo: Writing these multi-argument operators so they can be used this way was definitely intentional, as I like not having to special-case all of the calling code for when there's only one operand.

In fact, I recently discovered that you can go a step further with e.g.:
```
class Thing:
    def intersection(*args: Thing) -> Thing:
        ...
```
and make it work even with zero arguments (if that's a reasonable limit) via the type while still permitting `a.intersection(b)` bound-method calls on instances.  I've started doing that in some similar cases on my current branch, though flake8 has to be shushed about the first argument not being called `self`.

andy-slac: @TallJimbo, should I update `Timespan.intersection` instead of fixing this method in `_coordinate.py`?

TallJimbo: Oh, that's an interesting idea - I do think that's slightly better, but if you've already run Jenkins and are hoping to make only smaller changes now, feel free not to, as I have another commit on my current branch where I could easily do the same.

andy-slac: I want to check if there is a unit test that can be extended to include cases for 0 or more than one timespan. If I manage to do that I can also replace this code with updated `intersection` method.

andy-slac: I tried to update `intersection` to not use `self` but I do not quite like that. The issue is that I then need to change its return type to `Optional[Timestamp]` and that breaks `mypy` in bunch of other places which do not expect `None` to be returned from that method now. I don't think it's possible to tell `mypy` that `None` is never returned from a bound method. I think will stick with a workaround that I have for `DataCoordinate.timespan` and try to add a unit test for case of zero timespans (more than one is much harder to construct).

PR #684, opened 2022-05-16
DM-34812 MWCommand argument capture does not split values separated by an equal sign

n8pease:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/cli/utils.py
+++ b/python/lsst/daf/butler/cli/utils.py
@@ -19,6 +19,36 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+
+__all__ = (
+    "astropyTablesToStr",
+    "printAstropyTables",
+    "textTypeStr",
+    "LogCliRunner",
+    "clickResultMsg",
+    "command_test_env",
+    "addArgumentHelp",
+    "split_commas",
+    "split_kv",
+    "to_upper",
+    "unwrap",
+    "option_section",
+    "MWPath",
+    "MWOption",
+    "MWArgument",
+    "OptionSection",
+    "MWOptionDecorator",
+    "MWArgumentDecorator",
+    "MWCommand",
+    "ButlerCommand",
+    "OptionGroup",
+    "MWCtxObj",
+    "yaml_presets",
+    "sortAstropyTable",
+    "catch_and_exit",
+)
+
+
 import itertools
 import logging
 import os
--- a/python/lsst/daf/butler/cli/utils.py
+++ b/python/lsst/daf/butler/cli/utils.py
@@ -19,7 +19,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import copy
 import itertools
 import logging
 import os
@@ -27,6 +26,7 @@ import sys
 import textwrap
 import traceback
 import uuid
+from collections import Counter
 from contextlib import contextmanager
 from functools import partial, wraps
 from unittest.mock import patch
@@ -689,8 +689,86 @@ class MWCommand(click.Command):
             kwargs["callback"] = catch_and_exit(callback)
         super().__init__(*args, **kwargs)
 
+    def _capture_args(self, ctx, args):
+        """Capture the command line options and arguments.
+
+        See details about what is captured and the order in which it is stored
+        in the documentation of `MWCtxObj`.
+
+        Parameters
+        ----------
+        ctx : `click.Context`
+            The current Context.
+        args : `list` [`str`]
+            The list of arguments from the command line, split at spaces but
+            not at separators (like "=").
+        """
+        parser = self.make_parser(ctx)
+        opts, _, param_order = parser.parse_args(args=list(args))
+        # `param_order` is a list of click.Option and click.Argument, there is
+        # one item for each time the Option or Argument was used on the
+        # command line. Options will precede Arguments, within each sublist
+        # they are in the order they were used on the command line. Note that
+        # click.Option and click.Argument do not contain the value from the
+        # command line; values are in `opts`.
+        #
+        # `opts` is a dict where the key is the argument name to the
+        # click.Command function, this name matches the `click.Option.name` or
+        # `click.Argument.name`. For Options, an item will only be present if
+        # the Option was used on the command line. For Arguments, an item will
+        # always be present and if no value was provided on the command line
+        # the value will be `None`. If the option accepts multiple values, the
+        # value in `opts` is a tuple, otherwise it is a single item.
+        next_idx = Counter()
+        captured_args = []
+        for param in param_order:
+            if isinstance(param, click.Option):
+                if param.multiple:
+                    val = opts[param.name][next_idx[param.name]]
+                    next_idx[param.name] += 1
+                else:
+                    val = opts[param.name]
+                if param.is_flag:
+                    # Bool options store their True flags in opts and their
+                    # False flags in secondary_opts.
+                    if val:
+                        flag = max(param.opts, key=len)
+                    else:
+                        flag = max(param.secondary_opts, key=len)
+                    captured_args.append(flag)
+                else:
+                    captured_args.append(max(param.opts, key=len))
+                    captured_args.append(val)
+            elif isinstance(param, click.Argument):
+                if (opt := opts[param.name]) is not None:
+                    captured_args.append(opt)
+            else:
+                assert False  # All parameters should be an Option or an Argument.
+        MWCtxObj.getFrom(ctx).args = captured_args
+
     def parse_args(self, ctx, args):
-        MWCtxObj.getFrom(ctx).args = copy.copy(args)
+        """Given a context and a list of arguments this creates the parser and
+        parses the arguments, then modifies the context as necessary. This is
+        automatically invoked by make_context().
+
+        This function overrides `click.Command.parse_args`.
+
+        The call to `_capture_args` in this override stores the arguments
+        (option names, option value, and argument values) that were used by the
+        caller on the command line in the context object. These stored
+        arugments can be used by the command function, e.g. to process options
+        in the order they appeared on the command line (pipetask uses this
+        feature to create pipeline actions in an order from different options).
+
+        Parameters
+        ----------
+        ctx : `click.core.Context`
+            The current Context.ß
+        args : `list` [`str`]
+            The list of arguments from the command line, split at spaces but
+            not at separators (like "=").
+        """
+        self._capture_args(ctx, args)
         super().parse_args(ctx, args)
 
     @property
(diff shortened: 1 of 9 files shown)


python/lsst/daf/butler/cli/utils.py
```diff
@@ -689,8 +689,64 @@ def __init__(self, *args, **kwargs):
             kwargs["callback"] = catch_and_exit(callback)
         super().__init__(*args, **kwargs)
 
+    def _capture_args(self, ctx, args):
+        """Catpure the command line options and arguments.
```
timj: Typo: catpure

python/lsst/daf/butler/cli/utils.py
```diff
@@ -689,8 +689,64 @@ def __init__(self, *args, **kwargs):
...
+
+        See details about what is caputured and the order in which it is stored
```
timj: Typo: caputured

python/lsst/daf/butler/cli/utils.py
```diff
@@ -689,8 +689,64 @@ def __init__(self, *args, **kwargs):
...
+        in the documentation of MWCtxObj.
```
timj: Back ticks so it will link properly. I don't think it hurts to say "this function reorders the given argument list to ensure that options come before arguments". I'm not sure what "capture" means in this sense.

python/lsst/daf/butler/cli/utils.py
```diff
@@ -689,8 +689,64 @@ def __init__(self, *args, **kwargs):
...
+
+        Parameters
+        ----------
+        ctx : click.Context
```
timj: back ticks around `click.Context`

python/lsst/daf/butler/cli/utils.py
```diff
@@ -689,8 +689,64 @@ def __init__(self, *args, **kwargs):
...
+            The current Context.
+        args : `list` [`Any`]
+        """
+        parser = self.make_parser(ctx)
+        opts, _, param_order = parser.parse_args(args=list(args))
+        # `param_order` is a list of click.Option and click.Argument, there is
+        # one item for each time the Option or Argument was used on the
+        # command line. Options will preceed Arguments, within each sublist
+        # they are in the order they were used on the command line. Note that
+        # click.Option and click.Argumet do not contain the value from the
```
timj: Typo: Argumet

python/lsst/daf/butler/cli/utils.py
```diff
@@ -689,8 +689,64 @@ def __init__(self, *args, **kwargs):
...
+        # command line; values are in `opts`.
+        #
+        # `opts` is a dict where the key is the argument name to the
+        # click.Command function, this name matches the `click.Option.name` or
+        # `click.Argument.name`. For Options, an item will only be present if
+        # the Option was used on the command line. For Arguments, an item will
+        # always be present and if no value was provided on the command line
+        # the value will be `None`. If the option accepts multiple values, the
+        # value in `opts` is a tuple, otherwise it is a single item.
+        next_idx = defaultdict(lambda: 0)
```
timj: I think you can use `collections.Counter` here instead.

```python
>>> from collections import Counter
>>> c = Counter()
>>> c["a"] += 1
>>> c["d"] += 1
>>> c["d"] += 1
>>> c
Counter({'d': 2, 'a': 1})
>>> c["c"]
0
```

python/lsst/daf/butler/cli/utils.py
```diff
@@ -689,8 +689,64 @@ def __init__(self, *args, **kwargs):
...
+                    captured_args.append(flag)
+                else:
+                    captured_args.append(max(param.opts, key=len))
+                    captured_args.append(val)
+            elif isinstance(param, click.Argument):
+                if (opt := opts[param.name]) is not None:
+                    captured_args.append(opt)
+            else:
+                # All parameters should be an Option or an Argument.
+                assert False
```
timj: Put the comment on the end of the assert otherwise if somehow it ever triggers it's not going to tell the user anything.

```python
assert False, "All parameters should be an Option or an Argument"
```

python/lsst/daf/butler/cli/utils.py
```diff
@@ -689,8 +689,64 @@ def __init__(self, *args, **kwargs):
...
+        """Catpure the command line options and arguments.
+
+        See details about what is caputured and the order in which it is stored
+        in the documentation of MWCtxObj.
+
+        Parameters
+        ----------
+        ctx : click.Context
+            The current Context.
+        args : `list` [`Any`]
```
timj: Add a one sentence description.

python/lsst/daf/butler/cli/utils.py
```diff
@@ -736,9 +792,34 @@ class MWCtxObj:
...
+        Similar to the list of arguments and options that were passed in on the
+        command line, mostly in the order they were in on the command line,
+        excepting that where on the command line arguments may come before or
+        be interleaved with options, in this list options will be first,
+        followed by arguments.
+
+        This list is generated from command line options and arguments thathave
+        been parsed. Differences between this list and what may have appeared
+        on the command line are:
+        * For long options, which begin with two dashes, values that were
```
timj: Does this render okay without a blank line above?

python/lsst/daf/butler/cli/utils.py
```diff
@@ -736,9 +792,34 @@ class MWCtxObj:
...
-        The list of arguments (argument values, option flags, and option
-        values), split using whitespace, that were passed in on the command
-        line for the subcommand represented by the parent context object.
+        Similar to the list of arguments and options that were passed in on the
+        command line, mostly in the order they were in on the command line,
+        excepting that where on the command line arguments may come before or
+        be interleaved with options, in this list options will be first,
+        followed by arguments.
+
+        This list is generated from command line options and arguments thathave
```
timj: Typo: thathve

python/lsst/daf/butler/cli/utils.py
```diff
@@ -689,8 +689,64 @@ def __init__(self, *args, **kwargs):
             kwargs["callback"] = catch_and_exit(callback)
         super().__init__(*args, **kwargs)
 
+    def _capture_args(self, ctx, args):
```
timj, line 722: The commit message doesn't really say why this method is needed at all. What are we capturing and why isn't click doing it right in the first place?

python/lsst/daf/butler/cli/utils.py
```diff
@@ -689,8 +689,64 @@ def __init__(self, *args, **kwargs):
...
+        ----------
+        ctx : click.Context
+            The current Context.
+        args : `list` [`Any`]
+        """
+        parser = self.make_parser(ctx)
+        opts, _, param_order = parser.parse_args(args=list(args))
+        # `param_order` is a list of click.Option and click.Argument, there is
+        # one item for each time the Option or Argument was used on the
+        # command line. Options will preceed Arguments, within each sublist
```
timj: Typo: preceed

tests/test_cliUtils.py
```diff
@@ -366,5 +368,60 @@ def test_exist(self):
...
+        @click.option("-f", is_flag=True)
+        @click.option("-d/-n", "--do/--do-not", "do_or_do_not")
+        @click.option("--multi", multiple=True)
+        @click.pass_context
+        def cmd(ctx, arg_a, arg_b, an_option, second_option, f, do_or_do_not, multi):
+            self.assertIsNotNone(ctx)
+            self.ctx = ctx
+
+        for args, expected in (
+            ("--an-option foo --second-option bar", None),
```
timj, line 396: I'm not sure what is going on here. Why does this one seem to expand to `None` but a few lines below the same option does turn up in the expected list when it includes an `=` sign? I think I need a bit more explanation as to what is happening.

tests/test_cliUtils.py
```diff
@@ -366,5 +368,60 @@ def test_exist(self):
...
+            ("--second-option bar --an-option foo", None),
+            ("--an-option foo", None),
+            ("--second-option bar", None),
+            ("--an-option foo -f --second-option bar", None),
+            ("-o foo -s bar", "--an-option foo --second-option bar"),
+            ("--an-option foo -f -s bar", "--an-option foo -f --second-option bar"),
+            ("--an-option=foo", "--an-option foo"),
+            # NB when using a short flag everything that follows, including an
+            # equals sign, is part of the value!
+            ("-o=foo", "--an-option =foo"),
```
timj, line 406: I thought the point of this ticket was that the `=` would be handled so why isn't this `--an-option foo` or else a warning issued to indicate that the `=` was likely not intended?

timj (commented):
I've left some comments but the more I read the more confused I got. There is a lot of code here but I haven't worked out where the warning to the user about the spurious `=` is issued or even where we do anything special with the discovery of an `=`.

tests/test_cliUtils.py
n8pease: I thought the fix for this ticket was to emulate "normal" unix behavior which AFAIK is for an option name with one leading dash is to accept everything after the option name, stripping whitespace. (so `-o=foo` gets the value "=foo" for the `-o` option). 

This not working properly was the underlying issue that caused the reported failure https://lsstc.slack.com/archives/C01FBUGM2CV/p1652380003550509

If the normal behavior was working correctly, the pipetask execution would have reported an error, something to the effect of "no task named =characterizeImage".

tests/test_cliUtils.py
n8pease: `None` in the second item of the tuple means that the expected value is exactly the same as the input value (see line 418). I will add a comment to make that more clear.

python/lsst/daf/butler/cli/utils.py
n8pease: No, it didn't. Actually none of the CLI code had been added to the sphinx output. I figured out how to add it, should I add that to this ticket?

timj: Sure. Add it (I'm assuming it's a couple of lines to the index file -- make it clear that these are developer docs and not expected to be for general usage).

python/lsst/daf/butler/cli/utils.py
n8pease: Capturing the args is not new behavior; captured args are used by pipetask to turn CLI options into pipeline actions. There are 5 different options that result in a pipeline action (--task, --delete, --config, --config-file, --instrument), and the order these options are used on the command line specifies the order of the actions to be run. 

Click doesn't include any facility get the order of different options applied, it's a non-standard use case I think.

The problem with the way we were capturing arguments before was that they were not parsed enough; they were separated by whitespace, but not by separators ("="). And there's conditions where the separator should be included or excluded from the value. So I wrote this helper function to parse the list of CLI arguments, and reconstruct the list of option names & values, and argument values from the result.

I'll add some comments to the code and to the commit message.

timj: Thanks. That explanation helps a lot because it wasn't clear that this was specifically a pipetask-needing-oddly-specific-behavior problem versus a more general click-parsing-options problem.

PR #685, opened 2022-05-17
DM-34589: Add a way to mark which dimensions populate others.

TallJimbo:
In the future, this will be used to make it easier to export dimension data without knowing the details of the dimension combinations that represent many-to-main joins (DM-34838).

There's no real gain to updating the dimensions configuration before that happens, but merging these changes early is useful because it will make software versions with just this change much better able to handle future data repositories that use them in their dimensions configuration, even if the associated functionality isn't available as a result.

## Checklist

- [x] ran ~Jenkins~ local lsstsw

### DIFF
--- a/doc/changes/DM-34589.misc.md
+++ b/doc/changes/DM-34589.misc.md
@@ -1,0 +1,3 @@
+Add dimensions config entries that declare that the `visit` dimension "populates" various dimension elements that define many-to-many relationships.
+
+In the future, this will be used to ensure the correct records are included in exports of dimension records.
--- a/python/lsst/daf/butler/configs/dimensions.yaml
+++ b/python/lsst/daf/butler/configs/dimensions.yaml
@@ -1,4 +1,4 @@
-version: 3
+version: 4
 namespace: daf_butler
 skypix:
   # 'common' is the skypix system and level used to relate all other spatial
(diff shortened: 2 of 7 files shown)


timj (commented):
I don't understand why this can't have any tests? Why can't we have a small test dimensions universe that has a couple of elements in where one populates the other. We have special test dimensions.yaml content in pipe_base and ctrl_mpexec.  Why can't there be one call to get_elements_populated_by in one of the dimensions tests?

TallJimbo:
I can add some tests with a custom dimensions universe later in the week.

TallJimbo:
I think I'm changing my mind (a bit) again on how far to go on this ticket - I'm thinking now that we _should_ update the default dimensions config to use `populated_by`, even if it doesn't do anything yet, because that will hopefully save repos created between now and some date in the future from needing a migration.  But I still don't think we want to migrate any existing repos to use it.

So, since I think the rest of the migration changes are pretty close to merging over on DM-33942, let's hold this ticket up until that merges, then update the default dimensions config on this ticket after rebasing (and add a test based on that default config then).

timj:
#674 (DM-33942) was merged.

doc/changes/DM-34589.misc.md
```diff
@@ -0,0 +1,3 @@
+Add dimensions config entries that declare that the `visit` dimension "populates" various dimension elements that define many-to-main relationships.
```
timj:
```suggestion
Add dimensions config entries that declare that the `visit` dimension "populates" various dimension elements that define many-to-many relationships.
```
?

timj (approved):

PR #686, opened 2022-05-19
DM-34884: Disable checks of schema checksums

andy-slac:
The code that calculates checksums is still there, will be removed when
we are sure that we never need it.

## Checklist

- [X] ran Jenkins


timj (approved):
This matches what we agreed in the meeting today.

PR #687, opened 2022-05-25
DM-34919: Ensure that a component dataset type inherits the calibration flag

timj:
And that a composite dataset type also inherits the calibration flag from a component.

Other changes triggered by action failures:

* Switch to Python 3.10 to match rubin-env 4.0.0
* Fix a problem in a test with astropy 5.1
* Fix mypy problem where py3.8 picked the wrong line to ignore.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -20,7 +20,7 @@ jobs:
 
       - uses: conda-incubator/setup-miniconda@v2
         with:
-          python-version: 3.8
+          python-version: '3.10'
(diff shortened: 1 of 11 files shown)


timj:
@TallJimbo this is a two line change that was needed to make a test in obs_lsst work. It turned out that you could not get `bias.metadata` because the component was not inheriting the calibration flag which meant that the special temporal/spatial lookup was not triggering.

TallJimbo (approved):
All looks good!

PR #692, opened 2022-06-07
DM-34855: Add vectorized way of getting URIs for many datasets

n8pease:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/datastore.py
+++ b/python/lsst/daf/butler/core/datastore.py
@@ -757,10 +757,60 @@ class Datastore(metaclass=ABCMeta):
 
         raise NotImplementedError(f"Datastore {type(self)} must implement a transfer_from method.")
 
+    def getManyURIs(
+        self,
+        refs: Iterable[DatasetRef],
+        predict: bool = False,
+        allow_missing: bool = False,
+    ) -> Dict[DatasetRef, DatasetRefURIs]:
+        """Return URIs associated with many datasets.
+
+        Parameters
+        ----------
+        refs : iterable of `DatasetIdRef`
+            References to the required datasets.
+        predict : `bool`, optional
+            If the datastore does not know about a dataset, should it
+            return a predicted URI or not?
+        allow_missing : `bool`
+            If `False`, and `predict` is `False`, will raise if a `DatasetRef`
+            does not exist.
+
+        Returns
+        -------
+        URIs : `dict` of [`DatasetRef`, `DatasetRefUris`]
+            A dict of primary and component URIs, indexed by the passed-in
+            refs.
+
+        Raises
+        ------
+        FileNotFoundError
+            A URI has been requested for a dataset that does not exist and
+            guessing is not allowed.
+
+        Notes
+        -----
+        In file-based datastores, getManuURIs does not check that the file is
+        really there, it's assuming it is if datastore is aware of the file
+        then it actually exists.
+        """
+        uris: Dict[DatasetRef, DatasetRefURIs] = {}
+        missing_refs = []
+        for ref in refs:
+            try:
+                uris[ref] = self.getURIs(ref, predict=predict)
+            except FileNotFoundError:
+                missing_refs.append(ref)
+        if missing_refs and not allow_missing:
+            raise FileNotFoundError(
+                "Missing {} refs from datastore out of {} and predict=False.".format(
+                    num_missing := len(missing_refs), num_missing + len(uris)
+                )
+            )
+        return uris
+
     @abstractmethod
-    def getURIs(
-        self, datasetRef: DatasetRef, predict: bool = False
-    ) -> DatasetRefURIs:
+    def getURIs(self, datasetRef: DatasetRef, predict: bool = False) -> DatasetRefURIs:
         """Return URIs associated with dataset.
 
         Parameters
--- a/python/lsst/daf/butler/core/datastore.py
+++ b/python/lsst/daf/butler/core/datastore.py
@@ -23,13 +23,13 @@
 
 from __future__ import annotations
 
-__all__ = ("DatastoreConfig", "Datastore", "DatastoreValidationError")
+__all__ = ("DatastoreConfig", "Datastore", "DatastoreValidationError", "DatasetRefURIs")
 
 import contextlib
 import dataclasses
 import logging
 from abc import ABCMeta, abstractmethod
-from collections import defaultdict
+from collections import abc, defaultdict
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -192,6 +192,56 @@ class DatastoreTransaction:
             # We may still want to events from this transaction as part of
             # the parent.
             self.parent._log.extend(self._log)
+
+
+@dataclasses.dataclass
+class DatasetRefURIs(abc.Sequence):
+    """Represents the primary and component ResourcePath(s) associated with a
+    DatasetRef.
+
+    This is used in places where its members used to be represented as a tuple
+    `(primaryURI, componentURIs)`. To maintain backward compatibility this
+    inherits from Sequence and so instances can be treated as a two-item
+    tuple.
+    """
+
+    def __init__(
+        self,
+        primaryURI: Optional[ResourcePath] = None,
+        componentURIs: Optional[Dict[str, ResourcePath]] = None,
+    ):
+
+        self.primaryURI = primaryURI
+        """The URI to the primary artifact associated with this dataset. If the
+        dataset was disassembled within the datastore this may be `None`.
+        """
+
+        self.componentURIs = componentURIs or {}
+        """The URIs to any components associated with the dataset artifact
+        indexed by component name. This can be empty if there are no
+        components.
+        """
+
+    def __getitem__(self, index: Any) -> Any:
+        """Get primaryURI and componentURIs by index.
+
+        Provides support for tuple-like access.
+        """
+        if index == 0:
+            return self.primaryURI
+        elif index == 1:
+            return self.componentURIs
+        raise IndexError("list index out of range")
+
+    def __len__(self) -> int:
+        """Get the number of data members.
+
+        Provides support for tuple-like access.
+        """
+        return 2
+
+    def __repr__(self) -> str:
+        return f"DatasetRefURIs({repr(self.primaryURI)}, {repr(self.componentURIs)})"
 
 
 class Datastore(metaclass=ABCMeta):
@@ -710,7 +760,7 @@ class Datastore(metaclass=ABCMeta):
     @abstractmethod
     def getURIs(
         self, datasetRef: DatasetRef, predict: bool = False
-    ) -> Tuple[Optional[ResourcePath], Dict[str, ResourcePath]]:
+    ) -> DatasetRefURIs:
         """Return URIs associated with dataset.
 
         Parameters
@@ -723,13 +773,11 @@ class Datastore(metaclass=ABCMeta):
 
         Returns
         -------
-        primary : `lsst.resources.ResourcePath`
-            The URI to the primary artifact associated with this dataset.
-            If the dataset was disassembled within the datastore this
-            may be `None`.
-        components : `dict`
-            URIs to any components associated with the dataset artifact.
-            Can be empty if there are no components.
+        uris : `DatasetRefURIs`
+            The URI to the primary artifact associated with this dataset (if
+            the dataset was disassembled within the datastore this may be
+            `None`), and the URIs to any components associated with the dataset
+            artifact. (can be empty if there are no components).
         """
         raise NotImplementedError()
 
--- a/python/lsst/daf/butler/datastores/chainedDatastore.py
+++ b/python/lsst/daf/butler/datastores/chainedDatastore.py
@@ -34,6 +34,7 @@ from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Mapping, Optional, 
 from lsst.daf.butler import (
     Constraints,
     DatasetRef,
+    DatasetRefURIs,
     DatasetTypeNotSupportedError,
     Datastore,
     DatastoreConfig,
@@ -503,7 +504,7 @@ class ChainedDatastore(Datastore):
 
     def getURIs(
         self, ref: DatasetRef, predict: bool = False
-    ) -> Tuple[Optional[ResourcePath], Dict[str, ResourcePath]]:
+    ) -> DatasetRefURIs:
         """Return URIs associated with dataset.
 
         Parameters
@@ -516,13 +517,11 @@ class ChainedDatastore(Datastore):
 
         Returns
         -------
-        primary : `lsst.resources.ResourcePath`
-            The URI to the primary artifact associated with this dataset.
-            If the dataset was disassembled within the datastore this
-            may be `None`.
-        components : `dict`
-            URIs to any components associated with the dataset artifact.
-            Can be empty if there are no components.
+        uris : `DatasetRefURIs`
+            The URI to the primary artifact associated with this dataset (if
+            the dataset was disassembled within the datastore this may be
+            `None`), and the URIs to any components associated with the dataset
+            artifact. (can be empty if there are no components).
 
         Notes
         -----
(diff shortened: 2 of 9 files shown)


python/lsst/daf/butler/core/datastore.py
```diff
@@ -707,6 +724,34 @@ def transfer_from(
 
         raise NotImplementedError(f"Datastore {type(self)} must implement a transfer_from method.")
 
+    def getManyURIs(
+        self, refs: Iterable[DatasetRef], predict: bool = False, throwIfNotFound: bool = True
```
timj: I'm trying to switch to things like `throw_if_not_found` instead. How about a more positive spin: `allow_missing`, defaulting to `False`?

I initially didn't understand why we needed it but I see now that it's critical for ChainedDatastore to be able to function.

python/lsst/daf/butler/core/datastore.py
```diff
@@ -707,6 +724,34 @@ def transfer_from(
...
+        predict : `bool`, optional
+            If the datastore does not know about a dataset, should it
+            return a predicted URI or not?
+
+        Returns
+        -------
+        URIs : `dict` of [`DatasetRef`, `DatasetRefUris`]
+            A dict of primary and component URIs, indexed by the passed-in
+            refs.
+        """
```
timj, line 796: Add a `Raises` section.

python/lsst/daf/butler/core/datastore.py
```diff
@@ -194,6 +194,23 @@ def commit(self) -> None:
             self.parent._log.extend(self._log)
 
 
+@dataclasses.dataclass
+class DatasetRefURIs:
```
timj: I'm not sure how easy it would be for us to change `getURIs` at this point, unless we had a way for this dataclass to take 0 and 1 index or unpack as a tuple for existing callers.

python/lsst/daf/butler/core/datastore.py
```diff
@@ -707,6 +724,34 @@ def transfer_from(
...
+    ) -> Dict[DatasetRef, DatasetRefURIs]:
+        """Return URIs associated with many datasets.
+
+        Parameters
+        ----------
+        refs : iterable of `DatasetIdRef`
+            References to the required datasets.
+        predict : `bool`, optional
+            If the datastore does not know about a dataset, should it
+            return a predicted URI or not?
```
timj, line 774: Missing documentation of the boolean parameter. Presumably it's only relevant if `predict=False`?

python/lsst/daf/butler/core/datastore.py
```diff
@@ -707,6 +724,34 @@ def transfer_from(
...
+            A dict of primary and component URIs, indexed by the passed-in
+            refs.
+        """
+        uris: Dict[DatasetRef, DatasetRefURIs] = {}
+        for ref in refs:
+            try:
+                uris[ref] = DatasetRefURIs(*self.getURIs(ref, predict=predict))
+            except FileNotFoundError:
+                if throwIfNotFound:
+                    raise
```
timj: I think I would prefer it if the code accumulated the missing refs and then reported them all at the end rather than only reporting the first one. In theory the list could get large but we should be able to say something like:

> Missing 5 DatasetRefs from datastore out of 400 and predict=False

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -501,6 +502,34 @@ def _finishIngest(
...
+        # datastore if it actually exists in a later datastore, so in that
+        # case check all datastores with predict=False first, and then try
+        # again with predict=True.
+        for p in (False, True) if predict else (False,):
+            for datastore in self.datastores:
+                gotUris = datastore.getManyURIs(missingRefs, p, throwIfNotFound=False)
+                missingRefs -= gotUris.keys()
+                uris.update(gotUris)
+                if not missingRefs:
+                    break
```
timj, line 528: This breaks out the first loop but it will still cause the outer loop to keep running so the first datastore will be called again with no refs.

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -501,6 +502,34 @@ def _finishIngest(
...
+        # Docstring inherited
+
+        uris: Dict[DatasetRef, DatasetRefURIs] = {}
+        missingRefs = set(refs)
+
+        # If predict is True we don't want to predict a dataset in the first
+        # datastore if it actually exists in a later datastore, so in that
+        # case check all datastores with predict=False first, and then try
+        # again with predict=True.
+        for p in (False, True) if predict else (False,):
```
timj, line 520: I've not seen this style before. Interesting.

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -501,6 +502,34 @@ def _finishIngest(
...
+            for datastore in self.datastores:
+                gotUris = datastore.getManyURIs(missingRefs, p, throwIfNotFound=False)
+                missingRefs -= gotUris.keys()
+                uris.update(gotUris)
+                if not missingRefs:
+                    break
+
+        if missingRefs and throwIfNotFound:
+            raise FileNotFoundError("Dataset(s) {} not in this datastore.".format(missingRefs))
```
timj: Should test the missing refs branches.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1762,6 +1763,105 @@ def getURI(self, ref: DatasetRef, predict: bool = False) -> ResourcePath:
...
+    ) -> Dict[DatasetRef, DatasetRefURIs]:
+        # Docstring inherited
+
+        PrimaryType = Optional[ResourcePath]
+        ComponentsType = Dict[str, ResourcePath]
+        uris: Dict[DatasetRef, DatasetRefURIs] = {}
+
+        records = self._get_stored_records_associated_with_refs(refs)
+
+        missingRefs = [ref for ref in refs if ref.id not in records.keys()]
```
timj: Is there an overhead to calling `records.keys()` in a loop? I'm assuming it's some kind of internal state that has low overhead each call so doesn't need to be stored in a variable first.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1762,6 +1763,105 @@ def getURI(self, ref: DatasetRef, predict: bool = False) -> ResourcePath:
...
+
+        for ref in missingRefs:
+            # if this has never been written then we have to guess
```
timj, line 1769: This code has all been copied from `getURIs()`. We should not be doing that. Better to call `self.getURIs()` directly. The one reason not to do that is that we already know that `self.exists(ref)` is going to fail and so it would be more efficient to not call that at all.

The quick fix is to delete this code and live with doing an extra call to exists() that we know is going to fail.

The efficient thing to do is to factor out the URI prediction code so that getURIs can call it and getManyURIs can call it with getManyURIs calling it without having to do the self.exists check.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1762,6 +1763,105 @@ def getURI(self, ref: DatasetRef, predict: bool = False) -> ResourcePath:
...
+            if not self.exists(ref) and not predict:
```
timj: We know by definition that `self.exists(ref)` is False (because it will query the datastore records first and find that there is no entry) so this check tells us nothing. We can raise immediately `if missingRefs and not predict`.

The exists() call in getURIs is checking that the datastore knows about the file and that the file is there. getManyURIs is not checking that the file is really there, it's assuming it is if datastore thinks it is. We probably should document that distinction.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1762,6 +1763,105 @@ def getURI(self, ref: DatasetRef, predict: bool = False) -> ResourcePath:
...
+            if ref.id is None:
+                log.warning("Can't get information about a ref when ref.id is None: %s", ref)
+                continue
+            storedFileInfos = records.get(ref.id, None)
+            if storedFileInfos is None:
+                continue
+
+            fileLocations = [(i.file_location(self.locationFactory), i) for i in storedFileInfos]
+
+            guessing = False
```
timj: Isn't this all duplicating the logic from the second half of getURIs? We can't be duplicating all this. Can you factor getURIs into a prediction part and "must exist" part so they can both share the same logic? The first new method takes a DatasetRef that is known not to exist, the second can take a StoredFileInfo.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1762,6 +1763,105 @@ def getURI(self, ref: DatasetRef, predict: bool = False) -> ResourcePath:
...
+                        raise
+                    else:
+                        continue
+
+                # Add the "#predicted" URI fragment to indicate this is a guess
+                primary = ResourcePath(location.uri.geturl() + "#predicted")
+
+            uris[ref] = DatasetRefURIs(primary, components)
+
+        for ref in refs:
```
timj: This loop is going over all the input refs, including the ones we already dealt with. Is it worth partitioning `refs` at the start into ones that were found and ones that weren't?

timj (changes requested):
I've not looked at the test code. My main comment is that there is too much code duplication of getURIs. getURIs needs to be split into two helper methods that getManyURIs can use. The logic in getURIs is to convoluted to clone it.

python/lsst/daf/butler/datastores/fileDatastore.py
n8pease: `records` is just a `dict` so the result of `records.keys()` is a `dict_keys`, which is designed to be iterated over and behaves something like a generator if I understand it correctly.

For that matter if this was going to be unchanged I probably should have changed the whole statement from a list comprehension to a generator IE
```
missingRefs = (ref for ref in refs if ref.id not in records.keys())
```
because there's no need to materialize the whole `missingRefs` list (I did think it was helpful to name the list though, instead of using it directly in line 1782).

But, as you note on line 1818 maybe we shouldn't iterate over the whole list of refs again, so this will want some rewriting to keep track of the not-missing refs.

timj: I think `records.keys()` is set-like so `in` should be fast. I was mainly wondering whether there is a cost to python to create it or if it's a read-only view of something that is free to return each time.

n8pease: oh, I see. Yes, it's probably reasonable & best to move it out of the comprehension & cache it.

n8pease:
Ok, I think I've addressed the code review comments. I'm still working on getting the last of the CI steps to pass, I guess the unit test is failing using python 3.8(?), so I'll have to figure out how to run that version of python on devl-01 (any suggestions?). And then I can look at the codecov report and try to cover missing lines.

timj:
The unit test is failing using python 3.9 and seemingly that causes the rest of the matrix to be killed early. The python 3.9 failure doesn't seem to be caused by python 3.9 vs 3.10 but you can edit the action work flow to leave only 3.10 in the matrix to check that it does pass with 3.10.

n8pease:
@timj tomorrow is my last planned full day of work. I've been working on test coverage, but it's slow going. I can spend some more time on it tomorrow. But, this might be ready for re-review, or if you want we can discuss how to hand it off? And, I'll plan to dial into the MW stand up on Thursday.

python/lsst/daf/butler/core/datastore.py
```diff
@@ -707,10 +757,60 @@ def transfer_from(
...
+
+        Parameters
+        ----------
+        refs : iterable of `DatasetIdRef`
+            References to the required datasets.
+        predict : `bool`, optional
+            If the datastore does not know about a dataset, should it
+            return a predicted URI or not?
+        allow_missing : `bool`
+            If `False`, and `predct` is `False`, will raise if a `DatasetRef`
```
timj: Typo `predct`

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -501,9 +502,37 @@ def _finishIngest(
...
+                break
+            for datastore in self.datastores:
+                got_uris = datastore.getManyURIs(missing_refs, p, allow_missing=True)
+                missing_refs -= got_uris.keys()
+                uris.update(got_uris)
+                if not missing_refs:
+                    break
+
+        if missing_refs and not allow_missing:
+            raise FileNotFoundError("Dataset(s) {} not in this datastore.".format(missing_refs))
```
timj: You can use an f-string here if you like.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1762,6 +1710,157 @@ def getURI(self, ref: DatasetRef, predict: bool = False) -> ResourcePath:
             )
         return primary
 
+    def _predict_URIs(
+        self,
+        ref: DatasetRef,
+        allow_missing: bool,
+    ) -> DatasetRefURIs:
+        """Predict a dataset ref's URIs.
```
timj: If you don't want to be possessive you can say something like

> Predict the URIs of a `DatasetRef`

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1762,6 +1710,157 @@ def getURI(self, ref: DatasetRef, predict: bool = False) -> ResourcePath:
...
+
+        Parameters
+        ----------
+        ref : `DatasetRef`
+            Reference to the required Dataset.
+        allow_missing: `bool`
+            If `False` and a ref is not able to be predicted then throw.
```
timj: then raise? rather than throw?

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1762,6 +1710,157 @@ def getURI(self, ref: DatasetRef, predict: bool = False) -> ResourcePath:
...
+        ref: DatasetRef,
+        allow_missing: bool,
+    ) -> DatasetRefURIs:
+        """Predict a dataset ref's URIs.
+
+        Parameters
+        ----------
+        ref : `DatasetRef`
+            Reference to the required Dataset.
+        allow_missing: `bool`
```
timj: Were you encountering problems with prediction that required this parameter? If someone is wanting a bunch of predicted URIs and then encounters a `DatasetTypeNotSupportedError` that would be surprising. Maybe this scenario is so rare we should at least also log (maybe DEBUG for now).

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1762,6 +1710,157 @@ def getURI(self, ref: DatasetRef, predict: bool = False) -> ResourcePath:
...
+            if not predict:
+                if not allow_missing:
+                    raise FileNotFoundError("Dataset {} not in this datastore.".format(ref))
+            else:
+                uris[ref] = self._predict_URIs(ref, allow_missing)
+
+        for ref in existing_refs:
+            if ref.id is None:
+                raise RuntimeError("Unexpected DatasetRef.id == None for existing refs.")
+            file_infos = records[ref.id]
```
timj: If you use `ref.getCheckedId()` here you can remove the mypy `if ref.id is None` check above because `getCheckedId` will ensure you have a valid dataset ref.

tests/test_cliCmdQueryDatasets.py
```diff
@@ -44,141 +143,39 @@ def _queryDatasets(repo, glob=(), collections=(), where="", find_first=False, sh
...
 
-        self.assertAstropyTablesEqual(tables, expectedTables, filterColumns=True)
+        self.assertAstropyTablesEqual(
+            tables, expectedFilesystemDatastoreTables(testRepo.butler.datastore.root), filterColumns=True
+        )
 
     def testNoShowURI(self):
         """Test for expected output without show_uri (default is False)."""
-        tables = self._queryDatasets(repo=self.root)
+        _ = MetricTestRepo(self.repoDir, configFile=self.configFile)
```
timj, line 176: Does something complain if the `_ =` is missing?

tests/test_datastore.py
```diff
@@ -57,6 +60,57 @@
 TESTDIR = os.path.dirname(__file__)
 
 
+class DataIdForTest(UserDict):
+
+    """A dict-like class that can be used for a DataId dict that is hashable.
+
+    By default the class is immutable ("frozen"). The `frozen`
+    attribue can be set to `False` to change values (but note that
```
timj: typo: attribue

timj (approved):
This is great. Thanks.

tests/test_cliCmdQueryDatasets.py
n8pease: If I make it a named variable then at least my editor complains that the var is never used. I guess I could try not assigning the instance to any var, though I would think that would look like an error to a linter as well, and would probably merit a comment for humans. I was thinking that assigning to `_` is the most expressive way to say "yes, I'm not using this"

python/lsst/daf/butler/datastores/fileDatastore.py
n8pease: I tried taking it out all unit tests, Jenkins, and the demo query in the notebook work fine without it, so I'm assuming I was confused about something when I added it here. I'll leave it out.

PR #696, opened 2022-06-09
DM-33488: Add pruneDataset method to QuntumBackedButler (DM-33488

andy-slac:
There are few changes to the `pruneDatasets` method:

- `run` parameter has been removed completely, it was not used in the
  standard Butler implementation.
- Handling of the execution butler-specific logic is now performed in
  `Butler.pruneDataset` method, clients are not supposed to know the
  exact type of butler (and can use LimitedButler where possible).
- `QuantumBackedButler.pruneDataset` does not support removal of datasets
  from tagged collections (which do not exist in QBB).

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_quantum_backed.py
+++ b/python/lsst/daf/butler/_quantum_backed.py
@@ -318,10 +318,7 @@ class QuantumBackedButler(LimitedButler):
         if purge:
             for ref in refs:
                 # We only care about removing them from actual output refs,
-                try:
-                    self._actual_output_refs.remove(ref)
-                except KeyError:
-                    pass
+                self._actual_output_refs.discard(ref)
 
         if unstore:
             # Point of no return for removing artifacts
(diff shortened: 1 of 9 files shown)


python/lsst/daf/butler/_quantum_backed.py
```diff
@@ -286,6 +286,47 @@ def putDirect(self, obj: Any, ref: DatasetRef) -> DatasetRef:
...
+            if ref.datasetType.component():
+                raise ValueError(f"Can not prune a component of a dataset (ref={ref})")
+
+        if unstore:
+            self.datastore.trash(refs)
+        if purge:
+            for ref in refs:
+                # We only care about removing them from actual output refs,
+                try:
+                    self._actual_output_refs.remove(ref)
```
TallJimbo: Could just use `discard` instead of `remove`, without the `try/except`.

TallJimbo (approved):
Looks good; one minor line comment, and there's a typo (`Quntum`) in a commit message.

PR #699, opened 2022-06-10
DM-35191: Add test matrix for python versions

mwittgen:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -11,6 +11,10 @@ on:
 jobs:
   build_and_test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


timj (approved):
This looks good to me. I see that it changed the name of the build_and_test action to include the python version number so it looks like I need to tweak the branch protections before you can merge this. It would be nice if the protection could use a wildcard or something.

mwittgen:
> This looks good to me. I see that it changed the name of the build_and_test action to include the python version number so it looks like I need to tweak the branch protections before you can merge this. It would be nice if the protection could use a wildcard or something.

Correct. I didn't want to mess with the protected branches before this gets merged.

PR #701, opened 2022-06-28
fix requirement for sqlalchemy

PaulPrice:
We use sqlalchemy.sql.visitors.InternalTraversal, which requires
sqlalchemy 1.4.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 pyyaml >= 5.1
 astropy >= 4.0
 click >7.0
-sqlalchemy >= 1.3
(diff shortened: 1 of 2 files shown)


PaulPrice:
https://docs.sqlalchemy.org/en/14/core/visitors.html#sqlalchemy.sql.visitors.InternalTraversal

timj (approved):
Thanks for this. Can you also please fix the entry in setup.cfg? (which is the one that pip install will use) Then I'm fine with you merging.

PR #704, opened 2022-07-01
DM-35391: Improve query-data-ids command

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/script/queryDataIds.py
+++ b/python/lsst/daf/butler/script/queryDataIds.py
@@ -19,11 +19,15 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import logging
+
 import numpy as np
 from astropy.table import Table as AstropyTable
 
 from .._butler import Butler
 from ..cli.utils import sortAstropyTable
+
+_LOG = logging.getLogger(__name__)
 
 
 class _Table:
@@ -78,6 +82,34 @@ def queryDataIds(repo, dimensions, datasets, where, collections, order_by, limit
     # Docstring for supported parameters is the same as Registry.queryDataIds
 
     butler = Butler(repo)
+
+    if datasets and collections and not dimensions:
+        # Determine the dimensions relevant to all given dataset types.
+        # Since we are going to AND together all dimensions, we can not
+        # seed the result with an empty set.
+        graph = None
+        dataset_types = list(butler.registry.queryDatasetTypes(datasets))
+        for dataset_type in dataset_types:
+            if graph is None:
+                # Seed with dimensions of first dataset type.
+                graph = dataset_type.dimensions
+            else:
+                # Only retain dimensions that are in the current
+                # set AND the set from this dataset type.
+                graph = graph.intersection(dataset_type.dimensions)
+            _LOG.debug("Dimensions now %s from %s", set(graph.names), dataset_type.name)
+
+            # Break out of the loop early. No additional dimensions
+            # can be added to an empty set when using AND.
+            if not graph:
+                break
+
+        if not graph:
+            names = [d.name for d in dataset_types]
+            return None, f"No dimensions in common for specified dataset types ({names})"
+        dimensions = set(graph.names)
+        _LOG.info("Determined dimensions %s from datasets option %s", dimensions, datasets)
+
     results = butler.registry.queryDataIds(
         dimensions, datasets=datasets, where=where, collections=collections
     )
@@ -91,6 +123,6 @@ def queryDataIds(repo, dimensions, datasets, where, collections, order_by, limit
 
     if len(results.graph) > 0:
         table = _Table(results)
-        return table.getAstropyTable(not order_by)
+        return table.getAstropyTable(not order_by), None
     else:
-        return None
+        return None, None
--- a/python/lsst/daf/butler/script/queryDataIds.py
+++ b/python/lsst/daf/butler/script/queryDataIds.py
@@ -121,8 +121,8 @@ def queryDataIds(repo, dimensions, datasets, where, collections, order_by, limit
             offset = None
         results.limit(limit, offset)
 
-    if len(results.graph) > 0:
+    if results.count() > 0 and len(results.graph) > 0:
         table = _Table(results)
         return table.getAstropyTable(not order_by), None
     else:
-        return None, None
+        return None, "\n".join(results.explain_no_results())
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/script/queryDataIds.py
```diff
@@ -78,6 +82,26 @@ def queryDataIds(repo, dimensions, datasets, where, collections, order_by, limit
     # Docstring for supported parameters is the same as Registry.queryDataIds
 
     butler = Butler(repo)
+
+    if not dimensions and datasets and collections:
```
parejkoj: I'd rather have parentheses here: I never get the ordering of not/or/and.

python/lsst/daf/butler/script/queryDataIds.py
```diff
@@ -89,8 +113,8 @@ def queryDataIds(repo, dimensions, datasets, where, collections, order_by, limit
...
         results.limit(limit, offset)
 
-    if len(results.graph) > 0:
+    if results.count() > 0 and len(results.graph) > 0:
         table = _Table(results)
-        return table.getAstropyTable(not order_by)
+        return table.getAstropyTable(not order_by), None
     else:
-        return None
+        return None, "\n".join(results.explain_no_results())
```
parejkoj, line 128: Is this tuple the convention for returning no results explanations elsewhere in the butler? I feel like raising `NoResults` (or some other custom exception) with the explanation in the payload would be more pythonic. Looking at the tests, it certainly would simplify this interface.

tests/test_cliCmdQueryDataIds.py
```diff
@@ -63,28 +63,31 @@ def tearDown(self):
...
+        self.assertFalse(msg)
         self.assertAstropyTablesEqual(res, expected)
 
     def testNull(self):
         "Test asking for nothing."
-        res = self._queryDataIds(self.root)
-        self.assertEqual(res, None)
+        res, msg = self._queryDataIds(self.root)
+        self.assertEqual(res, None, msg)
+        self.assertFalse(msg)
```
parejkoj: `msg` here is `None`, so use `assertIsNone(msg)`. (and more below)

tests/test_cliCmdQueryDataIds.py
```diff
@@ -63,28 +63,31 @@ def tearDown(self):
 
     def testDimensions(self):
         """Test getting a dimension."""
-        res = self._queryDataIds(self.root, dimensions=("visit",))
+        res, msg = self._queryDataIds(self.root, dimensions=("visit",))
         expected = AstropyTable(
             array((("R", "DummyCamComp", "d-r", 423), ("R", "DummyCamComp", "d-r", 424))),
             names=("band", "instrument", "physical_filter", "visit"),
         )
+        self.assertFalse(msg)
```
parejkoj, line 71: `msg` here is `None`, so use `assertIsNone(msg)`.

python/lsst/daf/butler/script/queryDataIds.py
```diff
@@ -78,6 +82,26 @@ def queryDataIds(repo, dimensions, datasets, where, collections, order_by, limit
...
+        graph = None
+        dataset_types = list(butler.registry.queryDatasetTypes(datasets))
+        for dataset_type in dataset_types:
+            if graph is None:
+                graph = dataset_type.dimensions
+            else:
+                graph = graph & dataset_type.dimensions
+            _LOG.debug("Dimensions now %s from %s", set(graph.names), dataset_type.name)
+            if not graph:
+                break
```
parejkoj, line 105: It took me a bit to work out the logic here (using the other `if not graph` below: maybe an explanatory comment just above? `graph = graph & dimensions` did not obviously parse in my head as something that could make `graph` an empty set (it's a set, right?).

python/lsst/daf/butler/script/queryDataIds.py
```diff
@@ -78,6 +82,26 @@ def queryDataIds(repo, dimensions, datasets, where, collections, order_by, limit
     # Docstring for supported parameters is the same as Registry.queryDataIds
 
     butler = Butler(repo)
+
+    if not dimensions and datasets and collections:
+        # Determine the dimensions relevant to all given dataset types.
+        graph = None
+        dataset_types = list(butler.registry.queryDatasetTypes(datasets))
+        for dataset_type in dataset_types:
+            if graph is None:
```
parejkoj, line 93: `graph` here is a `set`, no? Maybe assign it to an empty set above, as that would make the logic of this loop more obvious.

python/lsst/daf/butler/script/queryDataIds.py
```diff
@@ -78,6 +82,26 @@ def queryDataIds(repo, dimensions, datasets, where, collections, order_by, limit
...
+                graph = dataset_type.dimensions
+            else:
+                graph = graph & dataset_type.dimensions
+            _LOG.debug("Dimensions now %s from %s", set(graph.names), dataset_type.name)
+            if not graph:
+                break
+
+        if not graph:
+            names = [d.name for d in dataset_types]
+            return None, f"No dimensions in common for specified dataset types ({names})"
```
parejkoj, line 109: As I comment below on the other `return` statements, I think this makes more sense raising a custom exception or `NoResults`.

parejkoj (changes requested):
Some suggestions on clarifying the logic, and I'd rather it raised an exception than returned a tuple that always had one element be `None`.

python/lsst/daf/butler/script/queryDataIds.py
timj: `black` controls the formatting and decides whether parentheses are required or not. I can move the `not dimensions` to the end though.

python/lsst/daf/butler/script/queryDataIds.py
timj: This is not a butler convention -- it's trying to bridge the click to script implementation boundary. A specialist exception for this one function seems overkill to me.

python/lsst/daf/butler/script/queryDataIds.py
timj: Okay. This line is a short circuit given that once you have an empty set the logical AND can never result in content.

python/lsst/daf/butler/script/queryDataIds.py
timj: If I start with an empty set then ANDing anything will result in an empty set. I use `None` to indicate that nothing has been looked at yet. The alternative would be to build a default set with all possible dimensions in it. `None` seemed easier.

parejkoj (approved):
Thanks, those added comments help a lot with understanding the logic.

I'd still prefer a custom exception to a "one is always None" return tuple, but this works.

PR #705, opened 2022-07-05
DM-35458: Always write local datastore files to temporaries

timj:
This allows for an atomic rename in case something goes wrong
in the file writing.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-35458.misc.rst
+++ b/doc/changes/DM-35458.misc.rst
@@ -1,0 +1,2 @@
+File datastore now always writes a temporary file and renames it even for local file system datastores.
+This minimizes the risk of a corrupt file being written if the process writing the file is killed at the wrong time.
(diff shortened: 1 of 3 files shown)


ktlim (approved):
Looks OK.  Any chance to test the new path?  (Maybe start writing a large file, then use inotify to see that the temp file has shown up?)

timj:
There is not a special path for local files any more. Everything uses a temporary and I can confirm that the log messages are showing that the temporary is in use and is then moved over to the final location.

ktlim:
But you don't check those log messages in a test case, do you?  In this case the side effect of the implementation choice is significant.

timj:
I don't check the log message. What I can do is add a check that the resources logger is reporting a move transfer for local writes. The only code choice there is is whether a move or copy is happening to the temporary file (and whether the temp file is inside datastore or not).

PR #706, opened 2022-07-06
DM-35469: Make the storage class redefinition error message clearer

timj:
Now uses repr() so that the differences are reported (str reported
just the name, which was already known to match)

## Checklist

- [x] ran Jenkins


TallJimbo (approved):

timj:
@TallJimbo sorry, can you please take a second look at this. I cracked and decided to add the file name (where available) to the error message so we know which file is causing the problem (although I'm not yet retaining a list of all previous config files that have been read to provide further information).

TallJimbo (approved):

PR #707, opened 2022-07-07
DM-35347: Switch to pyproject.toml

mwittgen:
## Checklist

- [ ] ran Jenkins


pyproject.toml
```diff
@@ -1,7 +1,83 @@
 [build-system]
-requires = ["setuptools", "lsst-versions"]
+requires = ["setuptools",
+    "lsst-versions",
+    "astropy >=4.0",
```
timj: Are you sure this is needed? It wasn't needed for `resources`. I thought build system was to list the dependencies to perform the build itself, not general dependencies for the package. Those are later on.

mwittgen: python -m build fails interactively and in GHA. 
It works when manually installing the dependencies and not using a venv for oython-build.
Let me investigate this further and finish the rest of the repos in this ticket.

timj: Is it because python build is being told to skip the dependency checks? We don't install dependencies when running `python -m build` (but build-system does). I will say that we did upload a resources release this week to PyPI, including a wheel. Maybe that was plain luck because resources does not import all the files immediately, only during tests, and those aren't run (_resourcePath.py does not depend on any PyPI packages).

It can't be right that you need to specify the same dependencies in both build-system and dependencies. I can imagine that we have to `pip install -r requirements.txt` before building if `build` imports the code at some point (I assume the error message you are getting is showing you that it's doing an import at some point).

mwittgen: `pip install -r requirements.txt` and not using a `venv` for `python-build` would work
I agree that it should not be specified twice and I need to check what the best solution is.

pyproject.toml
```diff
@@ -1,7 +1,70 @@
...
+    "lsst-resources",
+    "deprecated >= 1.2",
+    "pydantic",
+]
+
+dynamic = ["version"]
+[project.urls]
+"Homepage" = "https://github.com/lsst/daf_butler"
+[project.optional-dependencies]
+test = [
```
timj, line 40: In theory you can do DM-35272 (add a postgres option here) since you are editing the file but doing it on a separate ticket is fine.

timj (approved):
Looks great.

timj:
What's going on with the conda test env though?

mwittgen:
Network glitch, re-running jobs.

pyproject.toml
mwittgen: Implementing this in  DM-35272

PR #708, opened 2022-07-07
DM-34887: Transfer dimension records with butler.transfer_from()

timj:
This allows completely unrelated datasets to be copied into
a new butler.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -78,6 +78,8 @@ from .core import (
     Datastore,
     Dimension,
     DimensionConfig,
+    DimensionElement,
+    DimensionRecord,
     DimensionUniverse,
     FileDataset,
     Progress,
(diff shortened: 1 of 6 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -2235,6 +2241,27 @@ def transfer_from(
         else:
             log.log(VERBOSE, "All required dataset types are known to the target Butler")
 
+        dimension_records: Dict[DimensionElement, Dict[DataCoordinate, DimensionRecord]] = defaultdict(dict)
+        if transfer_dimension_data:
+            # Collect all the dimension records for these refs.
+            # All dimensions are to be copied but querying this butler
+            # universe.
```
TallJimbo: Something wrong with this sentence.

TallJimbo (approved):

python/lsst/daf/butler/_butler.py
timj: I think I'm trying to say that even though I'm copying records from the source butler I'm limiting the dimensions to those understood by the destination butler...

PR #710, opened 2022-07-12
DM-32412: Add storageClass and formatter for cell-based coadds

arunkannawadi:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -287,3 +287,5 @@ storageClasses:
     pytype: lsst.meas.extensions.scarlet.ScarletModelData
   MetricMeasurementBundle:
     pytype: lsst.analysis.tools.analysisMetrics.MetricMeasurementBundle
+  MultipleCellCoadd:
+    pytype: lsst.cell_coadds.MultipleCellCoadd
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -282,3 +282,5 @@ storageClasses:
     pytype: lsst.atmospec.spectraction.Spectraction
   ScarletModelData:
     pytype: lsst.meas.extensions.scarlet.ScarletModelData
+  MultipleCellCoadd:
+    pytytpe: lsst.cell_coadds.MultipleCellCoadd
```
timj: Typo.

arunkannawadi: Thanks for catching it, Tim

arunkannawadi:
I'm not sure why `mypy` is failing, given that I did not even touch python code.

timj:
It looks like a new release of mypy came out today that is clearly a bit cleverer and so that ignore is no longer needed. You can remove the ignore.

arunkannawadi:
mypy workflow is failing on main itself. I could add a commit may be removing the `type: ignore`

timj:
That's what I was trying to say. A new version of mypy came out today and so everything is going to break until we fix the code. Please fix the warning by removing the ignore.

TallJimbo (approved):

PR #711, opened 2022-07-12
Dm-35082: Add Universe compatibility check

natelust:
Add a method on DimensionUniverse that checks if another universe
is compatible with itself.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/dimensions/_universe.py
+++ b/python/lsst/daf/butler/core/dimensions/_universe.py
@@ -23,6 +23,7 @@ from __future__ import annotations
 
 __all__ = ["DimensionUniverse"]
 
+import logging
 import math
 import pickle
 from typing import (
@@ -60,6 +61,7 @@ if TYPE_CHECKING:  # Imports needed only for type annotations; may be circular.
 
 
 E = TypeVar("E", bound=DimensionElement)
+_LOG = logging.getLogger(__name__)
 
 
 @immutable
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/core/dimensions/_universe.py
```diff
@@ -206,6 +206,22 @@ def namespace(self) -> str:
         """
         return self._namespace
 
+    def checkCompatibility(self, other: DimensionUniverse) -> bool:
```
timj: Can we have `is_compatible_with`? We have such a named method in `StorageClass`.

python/lsst/daf/butler/core/dimensions/_universe.py
```diff
@@ -206,6 +206,22 @@ def namespace(self) -> str:
...
+        other : `DimensionUniverse`
+            The other `DimensionUniverse` to check for compatibility
+
+        Returns
+        -------
+        results : `bool`
+            If the other `DimensionUniverse` is compatible with this one return
+            `True`, else `False`
+        """
+        return self.namespace == other.namespace
```
timj: We agreed to issue a log message if the version number did not match.

timj (commented):
I know you are in a rush but this seems like a trivial thing to add a test for (even if we are passing in the same universe).

I worry that If they aren't compatible it's not clear why because the logic here is not visible to the caller and we have no way to report a reason.

python/lsst/daf/butler/core/dimensions/_universe.py
natelust: This may be a distinction without a difference, but we are not asking if this Universe is consistent with other, we are asking if other is consistent with this, which does not seem to be what I read from your name. And arn't methods supposed to start with verb?

PR #720, opened 2022-08-03
DM-35803: Add DataFrame delegate to enable InMemoryDatasets of DataFrames.

erykoff:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-35803.feature.rst
+++ b/doc/changes/DM-35803.feature.rst
@@ -1,0 +1,1 @@
+Added ``DataFrameDelegate`` to allow DataFrames to be used with ``lsst.pipe.base.InMemoryDatasetHandle``.
--- a/python/lsst/daf/butler/delegates/dataframe.py
+++ b/python/lsst/daf/butler/delegates/dataframe.py
@@ -1,0 +1,133 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Support for reading DataFrames."""
+from __future__ import annotations
+
+import collections.abc
+from typing import Any, Mapping, Optional
+
+import pandas
+from lsst.daf.butler import StorageClassDelegate
+from lsst.utils.introspection import get_full_type_name
+from lsst.utils.iteration import ensure_iterable
+
+__all__ = ["DataFrameDelegate"]
+
+
+class DataFrameDelegate(StorageClassDelegate):
+    def getComponent(self, composite: pandas.DataFrame, componentName: str) -> Any:
+        """Get a component from a DataFrame.
+
+        Parameters
+        ----------
+        composite : `~pandas.DataFrame`
+            ``DataFrame`` to access component.
+        componentName : `str`
+            Name of component to retrieve.
+
+        Returns
+        -------
+        component : `object`
+            The component.
+
+        Raises
+        ------
+        AttributeError
+            The component can not be found.
+        """
+        if componentName == "columns":
+            return pandas.Index(self._getAllColumns(composite))
+        else:
+            raise AttributeError(
+                f"Do not know how to retrieve component {componentName} from {get_full_type_name(composite)}"
+            )
+
+    def handleParameters(
+        self, inMemoryDataset: pandas.DataFrame, parameters: Optional[Mapping[str, Any]] = None
+    ) -> Any:
+        """Return possibly new in-memory dataset using the supplied parameters.
+
+        Parameters
+        ----------
+        inMemoryDataset : `object`
+            Object to modify based on the parameters.
+        parameters : `dict`, optional
+            Parameters to apply. Values are specific to the parameter.
+            Supported parameters are defined in the associated
+            `StorageClass`.  If no relevant parameters are specified the
+            ``inMemoryDataset`` will be return unchanged.
+
+        Returns
+        -------
+        inMemoryDataset : `object`
+            Original in-memory dataset, or updated form after parameters
+            have been used.
+        """
+        if not isinstance(inMemoryDataset, pandas.DataFrame):
+            raise ValueError(
+                "handleParameters for a DataFrame must get a DataFrame, "
+                f"not {get_full_type_name(inMemoryDataset)}."
+            )
+
+        if parameters is None:
+            return inMemoryDataset
+
+        if "columns" in parameters:
+            allColumns = self._getAllColumns(inMemoryDataset)
+
+            if not isinstance(parameters["columns"], collections.abc.Iterable):
+                raise NotImplementedError(
+                    "InMemoryDataset of a DataFrame only supports list/tuple of string column names"
+                )
+
+            for column in ensure_iterable(parameters["columns"]):
+                if not isinstance(column, str):
+                    raise NotImplementedError(
+                        "InMemoryDataset of a DataFrame only supports string column names."
+                    )
+                if column not in allColumns:
+                    raise ValueError(f"Unrecognized column name {column!r}.")
+
+            # Exclude index columns from the subset.
+            readColumns = [
+                name
+                for name in ensure_iterable(parameters["columns"])
+                if name not in inMemoryDataset.index.names
+            ]
+
+            return inMemoryDataset[readColumns]
+        else:
+            return inMemoryDataset
+
+    def _getAllColumns(self, inMemoryDataset: pandas.DataFrame) -> list[str]:
+        """Get all columns, including index columns.
+
+        Returns
+        -------
+        columns : `list` [`str`]
+            List of all columns.
+        """
+        allColumns = list(inMemoryDataset.columns)
+        if inMemoryDataset.index.names[0] is not None:
+            allColumns.extend(inMemoryDataset.index.names)
+
+        return allColumns
--- a/tests/test_parquet.py
+++ b/tests/test_parquet.py
@@ -60,23 +60,35 @@ class ParquetFormatterTestCase(unittest.TestCase):
         removeTestTempDir(self.root)
 
     def testSingleIndexDataFrame(self):
-        columns1 = pd.Index(["a", "b", "c"])
-        df1 = pd.DataFrame(np.random.randn(5, 3), index=np.arange(5, dtype=int), columns=columns1)
+        data = np.zeros(5, dtype=[("index", "i4"), ("a", "f8"), ("b", "f8"), ("c", "f8"), ("ddd", "f8")])
+        data["index"][:] = np.arange(5)
+        data["a"] = np.random.randn(5)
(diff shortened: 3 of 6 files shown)


python/lsst/daf/butler/delegates/__init__.py
```diff
@@ -0,0 +1 @@
+from .dataframe import *
```
timj: We should leave this file empty -- for delegates we aren't entirely sure what their dependencies are and we never want them to be imported by default somewhere. Leaving this blank means we don't have to put try blocks around the imports in each delegate.

doc/changes/DM-35803.feature.rst
```diff
@@ -0,0 +1,2 @@
+Added ``DataFrameDelegate`` to allow DataFrames to be used with ``lsst.pipe.base.InMemoryDatasetHandle``.
+Additionally, fixed a bug in the parquet reader where a single string column name would be interpreted as an iterable.
```
timj: In theory you can have a second file `DM-35803.bugfix.rst` for the bug fix message. That would make it turn up in the bug fix section of the release notes.

python/lsst/daf/butler/delegates/dataframe.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class DataFrameDelegate(StorageClassDelegate):
+    def getComponent(self, composite: pandas.DataFrame, componentName: str) -> Any:
+        """Get a component from a DataFrame.
+
+        Parameters
+        ----------
+        composite : `pandas.DataFrame`
+            `DataFrame` to access component.
```
timj: Not sure if backticks are going to work there without `~pandas.DataFrame`.

```
`~pandas.DataFrame` from which to extract the component.
```

python/lsst/daf/butler/delegates/dataframe.py
```diff
@@ -0,0 +1,128 @@
...
+        Raises
+        ------
+        AttributeError
+            The component can not be found.
+        """
+        if componentName == "columns":
+            return pandas.Index(self._getAllColumns(composite))
+        else:
+            raise AttributeError(
+                "Do not now how to retrieve component {} from {}".format(componentName, type(composite))
```
timj: Typo: now

You can use an f-string. Consider using `get_full_type_name` rather than `type`.

python/lsst/daf/butler/delegates/dataframe.py
```diff
@@ -0,0 +1,128 @@
...
+            )
+
+    def handleParameters(
+        self, inMemoryDataset: pandas.DataFrame, parameters: Optional[Mapping[str, Any]] = None
+    ) -> Any:
+        """Modify the in-memory dataset using the supplied parameters,
```
timj: This is my fault I think, but "Modify the in-memory dataset" does sound scary.

> Return possibly new in-memory dataset using the supplied parameters.

?

python/lsst/daf/butler/delegates/dataframe.py
```diff
@@ -0,0 +1,128 @@
...
+
+        Parameters
+        ----------
+        inMemoryDataset : `object`
+            Object to modify based on the parameters.
+        parameters : `dict`, optional
+            Parameters to apply. Values are specific to the parameter.
+            Supported parameters are defined in the associated
+            `StorageClass`.  If no relevant parameters are specified the
+            inMemoryDataset will be return unchanged.
```
timj: Double back ticks around `inMemoryDataset` since it's referring to a parameter.

python/lsst/daf/butler/delegates/dataframe.py
```diff
@@ -0,0 +1,128 @@
...
+
+        Returns
+        -------
+        inMemoryDataset : `object`
+            Updated form of supplied in-memory dataset, after parameters
```
timj: Probably should say here that it might be the original object.

python/lsst/daf/butler/delegates/dataframe.py
```diff
@@ -0,0 +1,128 @@
...
+            have been used.
+        """
+        if not isinstance(inMemoryDataset, pandas.DataFrame):
+            raise ValueError("handleParameters for a DataFrame must get a DataFrame.")
```
timj: Consider using get_full_type_name in this error message (and adding test for this scenario).

python/lsst/daf/butler/delegates/dataframe.py
```diff
@@ -0,0 +1,128 @@
...
+            # Exclude index columns from the subset
+            readColumns = [
+                name
+                for name in ensure_iterable(parameters["columns"])
+                if name not in inMemoryDataset.index.names
+            ]
+
+            return inMemoryDataset[readColumns]
+        else:
+            return inMemoryDataset
```
timj, line 119: A trivial test that passes in unknown parameters that are then ignored seems worth doing.

python/lsst/daf/butler/delegates/dataframe.py
```diff
@@ -0,0 +1,128 @@
...
+
+            if not isinstance(parameters["columns"], (list, tuple, str)):
+                raise NotImplementedError(
+                    "InMemoryDataset of a DataFrame only supports list/tuple of string column names"
+                )
+
+            for column in ensure_iterable(parameters["columns"]):
+                if not isinstance(column, str):
+                    raise NotImplementedError(
+                        "InMemoryDataset of a DataFrame only supports string column names."
```
timj, line 105: Say the type of the column name (or include the repr of it). (and add trivial test?)

python/lsst/daf/butler/delegates/dataframe.py
```diff
@@ -0,0 +1,128 @@
...
+                    )
+                if column not in allColumns:
+                    raise ValueError(f"Unrecognized column name {column}.")
```
timj: Using `column!r` would automatically include quotes around the column name.

python/lsst/daf/butler/delegates/dataframe.py
```diff
@@ -0,0 +1,128 @@
...
+
+            # Exclude index columns from the subset
```
timj: Pedantically, comments should be full sentences so should include a period at the end.

python/lsst/daf/butler/delegates/dataframe.py
```diff
@@ -0,0 +1,128 @@
...
+        if not isinstance(inMemoryDataset, pandas.DataFrame):
+            raise ValueError("handleParameters for a DataFrame must get a DataFrame.")
+
+        if parameters is None:
+            return inMemoryDataset
+
+        if "columns" in parameters:
+            allColumns = self._getAllColumns(inMemoryDataset)
+
+            if not isinstance(parameters["columns"], (list, tuple, str)):
```
timj: I'm not sure this test gives us much since the test below is checking the values we iterate over are `str`. The only relevant test seems to be that the value for the columns parameter is an iterable.  I think:

```python
if not isinstance(parameters["columns"], collections.abc.Iterable):
    raise ...
```
is sufficient isn't it?

python/lsst/daf/butler/delegates/dataframe.py
```diff
@@ -0,0 +1,128 @@
...
+                name
+                for name in ensure_iterable(parameters["columns"])
+                if name not in inMemoryDataset.index.names
+            ]
+
+            return inMemoryDataset[readColumns]
+        else:
+            return inMemoryDataset
+
+    def _getAllColumns(self, inMemoryDataset: pandas.DataFrame) -> List[str]:
```
timj: Add `from __future__ import annotations` at the top and then you can use `list[str]` for the type annotation and not `List[str]`.  We are trying to switch over to using native types.

python/lsst/daf/butler/delegates/dataframe.py
```diff
@@ -0,0 +1,128 @@
...
+        """Get all columns, including index columns.
+
+        Returns
+        -------
+        columns : `list` [`str`]
+            List of all columns.
+        """
+        allColumns = list(inMemoryDataset.columns)
+        if inMemoryDataset.index.names[0] is not None:
+            allColumns.extend(list(inMemoryDataset.index.names))
```
timj: Is the `list` needed here? `list.extend()` works with any iterable.

tests/test_parquet.py
```diff
@@ -60,23 +66,35 @@ def tearDown(self):
         removeTestTempDir(self.root)
 
     def testSingleIndexDataFrame(self):
-        columns1 = pd.Index(["a", "b", "c"])
-        df1 = pd.DataFrame(np.random.randn(5, 3), index=np.arange(5, dtype=int), columns=columns1)
+        data = np.zeros(5, dtype=[("index", "i4"), ("a", "f8"), ("b", "f8"), ("c", "f8"), ("ddd", "f8")])
```
timj: Not important but maybe have a `n=5` and use `n` everywhere?

tests/test_parquet.py
```diff
@@ -30,16 +30,22 @@
...
 from lsst.daf.butler import Butler, DatasetType
+from lsst.daf.butler.delegates import DataFrameDelegate
 from lsst.daf.butler.tests.utils import makeTestTempDir, removeTestTempDir
 
 TESTDIR = os.path.abspath(os.path.dirname(__file__))
 
 
+@unittest.skipUnless(pd is not None, "Cannot test ParquetFormatter without pandas.")
 @unittest.skipUnless(pyarrow is not None, "Cannot test ParquetFormatter without pyarrow.")
 class ParquetFormatterTestCase(unittest.TestCase):
```
timj, line 98: This test case uses a file datastore. If you were to have a subclass of this test that was configured in its setUp to use an in-memory datastore then you would automatically be testing the new delegate with the exact same tests that were used for the formatter.

Using something like:

```python
configFile = None

def setUp(self):
    self.root = makeTestTempDir(TESTDIR)
    kwargs = {}
    if self.configFile is not None:
        kwargs["config"] = Config(self.configFile)
        kwargs["forceConfigRoot"] = True
    Butler.makeRepo(self.root, **kwargs)
```

and then:

```python
class InMemoryDatastoreParquetTestCase(ParquetFormatterTestCase):
    configFile = os.path.join(TESTDIR, "config/basic/butler-chained.yaml")
```
might be sufficient.

tests/test_parquet.py
```diff
@@ -113,5 +131,76 @@ def testMultiIndexDataFrame(self):
             self.butler.get(self.datasetType, dataId={}, parameters={"columns": ["d"]})
 
 
+@unittest.skipUnless(pd is not None, "Cannot test DataFrameDelegate without pandas.")
+class DataFrameDelegateTestCase(unittest.TestCase):
```
timj: If you have an in-memory datastore test case you might be able to remove a lot of the code duplication here and reserve this test case class for dealing with all the places that trigger exceptions.

tests/test_parquet.py
```diff
@@ -113,5 +131,76 @@ def testMultiIndexDataFrame(self):
...
+    """Tests for DataFrameDelegate, for use with InMemoryDatastore."""
+
+    def testSingleIndexDataFrame(self):
+        data = np.zeros(5, dtype=[("index", "i4"), ("a", "f8"), ("b", "f8"), ("c", "f8"), ("ddd", "f8")])
```
timj: If the two datasets are remaining identical to the ones in the other test case consider having a helper function to generate them.

timj (approved):
Thank you for adding this. Lots of minor comments to consider but no big issues.

python/lsst/daf/butler/delegates/dataframe.py
erykoff: I thought I had tried this, but I must have done something wrong because this does work.

tests/test_parquet.py
erykoff: Okay, I got this working.

python/lsst/daf/butler/delegates/dataframe.py
erykoff: Can you remind me how to do this in a raise?

timj:
```python
with self.assertRaises(NotImplementedError) as cm:
    ...
self.assertIn("only supports string column names", str(cm.exception))
```
?

PR #721, opened 2022-08-03
DM-35815: Add StorageClassFactory.findStorageClass

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/storageClass.py
+++ b/python/lsst/daf/butler/core/storageClass.py
@@ -439,7 +439,7 @@ class StorageClass:
         """
         return isinstance(instance, self.pytype)
 
-    def is_type(self, other: Type) -> bool:
+    def is_type(self, other: Type, compare_types: bool = False) -> bool:
         """Return Boolean indicating whether the supplied type matches
         the type in this `StorageClass`.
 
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/core/storageClass.py
```diff
@@ -842,6 +856,55 @@ def getStorageClass(self, storageClassName: str) -> StorageClass:
...
+        if compare_types:
+            # The fast comparison failed and we were asked to try the
+            # variant that might involve code imports.
+            result = self._find_storage_class(pytype, True)
+            if result:
+                return result
+
+        raise KeyError(f"Unable to find a StorageClass associated with type {get_full_type_name(pytype)}")
+
+    def _find_storage_class(self, pytype: Type, compare_types: bool) -> Optional[StorageClass]:
```
erykoff, line 901: Does this need a docstring?

erykoff (approved):

python/lsst/daf/butler/core/storageClass.py
timj: It's really meant to be a super secret helper method. I toyed with having it be inside the other method but seemed better to be outside. I don't think we are required to fully document these kinds of methods but I will do so.

PR #724, opened 2022-08-17
DM-35937: Get the composite dataset type from the component dataset type

timj:
There is no need to ask registry for this (assuming the component
dataset type has been properly configured with a parent storage
class). This also prevents look ups into registry for non-existent
dataset types which do not occur in the else branch (which always
uses the given dataset type).

@kfindeisen can you see if this works for you?

## Checklist

- [x] ran Jenkins


kfindeisen:
Yes, it works, both in the demo and in our original `prompt_prototype` case. Thanks for the quick response! 🙂

TallJimbo (approved):

PR #725, opened 2022-08-22
DM-35917: Update split_commas to understand [] list syntax

timj:
"a[1,2]" will no longer be split.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_cliUtilSplitCommas.py
+++ b/tests/test_cliUtilSplitCommas.py
@@ -84,6 +84,28 @@ class SplitCommasTestCase(unittest.TestCase):
         self.assertEqual(result.exit_code, 0, msg=clickResultMsg(result))
         mock.assert_called_with(())
 
+    def test_parens(self):
+        """Test that split commas understands [a,b]"""
+        for test, expected in (
+            ("single", ("single",)),
+            ("a,b", ("a", "b")),
+            ("[a,b]", ("[a,b]",)),
+            ("a[1,2],b", ("a[1,2]", "b")),
+        ):
+            result = split_commas(None, None, test)
+            self.assertEqual(result, expected)
+
+        # These should warn because it's likely a typo.
+        for test, expected in (
+            ("a[1,b[2,3],c", ("a[1,b[2,3]", "c")),
+            ("a[1,b,c", ("a[1,b,c",)),
+            ("a[1,b", ("a[1,b",)),
+            ("a1,b]", ("a1", "b]")),
+        ):
+            with self.assertWarns(UserWarning, msg=f"Testing {test!r}"):
+                result = split_commas(None, None, test)
+            self.assertEqual(result, expected)
+
 
 if __name__ == "__main__":
     unittest.main()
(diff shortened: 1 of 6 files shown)


tests/test_cliUtilSplitCommas.py
```diff
@@ -84,6 +84,30 @@ def test_none(self):
...
+            ),
+            ("a1,b]", ("a1", "b]")),
+        ):
+            result = split_commas(None, None, test)
+            self.assertEqual(result, expected)
+
+        # This one should warn because it's likely that this is a typo.
+        with self.assertWarns(UserWarning):
+            result = split_commas(None, None, "a[1,b[2,3],c")
+        self.assertEqual(result, ("a[1,b[2,3]", "c")),
```
andy-slac: Does `a[1,b` make a warning too? Sort of inconsistent if `a[1,b` is OK, but `a[1,b[2,3]` is a typo, should not all unbalanced brackets make this warning?

andy-slac (approved):
Looks good, one minor question.

tests/test_cliUtilSplitCommas.py
timj: Ah yes, I don't check `in_parens` when I finish the loop... Will add.

timj: Oh, it won't even trigger the code path because it doesn't match the regex looking for `[.*,.*]`. Should I relax the regex to look for `[.*,` ?

andy-slac: How useful is that warning? Maybe just allow people do crazy unbalanced things? 🙂

timj: Can you think of cases where unbalanced bracketing is planned?

andy-slac: No, I have no idea actually for what are use cases for this syntax. If all brackets are supposed to be balanced then the warning should be produced for any unbalanced combination. I think you probably need to do slow loop for either `[` or `]` present in a string.

timj: If you look in lsst/ctrl_mpexec#199 you will see that the motivation for this is `--config a=5,b=[2,5]`

andy-slac: So, I gather the purpose is to handle `--config` option where the value is a list. But I can also imagine that a value of some parameter can be a string which accepts arbitrary characters, including unbalanced brackets? I suspect that allowing comma as a separator in `--config` is what causes us all this grief. Commas could also appear in string values, right?

timj: Yes, allowing `--config` to support commas to allow `--config a=b,c=2` is what I want to support, but also we have to allow `--config a=b[2,3,4]` because we say that should work (and by extension `--config a=b[2,3,4],c=2`. It could be that the actual fix is to tell `--config` that it should not do any splitting at all and if you want to specify two configs you say `--config` twice.  If someone wants to do `--config a=b,c` and is not wanting to assign to a list config item but a string item that happens to include a comma then indeed this is going to be problematic. I don't know if we have config items that have commas in strings but if we do I should turn off comma splitting everywhere for `--config`. cc/ @TallJimbo

TallJimbo: I don't know of any such config options, but as Tim can guess based on a previous conversation about this I'm generally on the side of limiting functionality in CLI config-handling whenever it's ambiguous and/or hard to maintain.

PR #731, opened 2022-09-06
DM-36111: miscellaneous improvements to registry support classes.

TallJimbo:
## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/registry/queries/_structs.py
```diff
@@ -223,10 +224,10 @@ class QueryWhereClause:
     (`lsst.sphgeom.Region` or `None`).
     """
 
-    restriction: GovernorDimensionRestriction
+    governor_constraints: Mapping[str, AbstractSet[str]]
     """Restrictions on the values governor dimensions can take in this query,
-    imposed by the string expression or data ID
-    (`GovernorDimensionRestriction`).
+    imposed by the string expression and/or data ID
+    (`Mapping` [ ``set`,  `~collections.abc.Set` ] ]).
```
andy-slac: Double backtick, and `set` should be `str`?

python/lsst/daf/butler/registry/datasets/byDimensions/summaries.py
```diff
@@ -294,20 +291,16 @@ def refresh(self, get_dataset_type: Callable[[int], DatasetType]) -> None:
...
-                summary.datasetTypes.add(datasetType)
             # Update the dimensions with the values in this row that aren't
             # None/NULL (many will be in general, because these enter the query
             # via LEFT OUTER JOIN).
+            summary.dataset_types.add(datasetType)
             for dimension in self._tables.dimensions:
                 value = row[dimension.name]
                 if value is not None:
-                    summary.dimensions.add(dimension, value)
+                    summary.governors.setdefault(dimension.name, set()).add(value)
```
andy-slac, line 303: Should be this using `summary.add_data_ids()` instead of modifying data members directly?

python/lsst/daf/butler/registry/queries/expressions/check.py
```diff
@@ -336,33 +333,33 @@ def visitInner(self, branches: Sequence[TreeSummary], form: NormalForm) -> Inner
...
-            # (or equivalent; categorizeIdentifier is smart enough to see that
-            # e.g. 'detector.id=4' is equivalent to 'detector=4').
-            # If so, and it's a governor dimension, remember that we've
-            # constrained it on this branch, and make sure it's consistent
-            # with any other constraints on any other branches its AND'd with.
-            if isinstance(branch.dataIdKey, GovernorDimension) and branch.dataIdValue is not None:
-                governor = branch.dataIdKey
-                value = summary.governors.setdefault(governor, branch.dataIdValue)
-                if value != branch.dataIdValue:
+            # Test whether this branch has a form like '<dimension>=<value' (or
```
andy-slac: Angle bracket missing after `<value`.

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -1724,6 +1725,67 @@ def query(
...
+        `constant_rows` for this backend.
+
+        Returns
+        -------
+        max : `int`
+            Maximum number of rows.
+
+        Notes
+        -----
+        This is used by `Session.upload` to decide when `constant_rows` can be
```
andy-slac: I don't see `Session.upload`, is it in one of your other branches?

python/lsst/daf/butler/core/timespan.py
```diff
@@ -647,6 +719,110 @@ def fromLiteral(cls: Type[_S], timespan: Timespan) -> _S:
...
+            A dictionary representing a database row containing a `Timespan`
+            in this representation.  Should have key(s) equal to the return
+            value of `getFieldNames`.
+        name : `str`, optional
+            Name for the logical column; a part of the name for multi-column
+            representations.  Defaults to ``cls.NAME``.
+
+        Returns
+        -------
+        timespan
```
andy-slac: Add `: Timespan, optional`?

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
```diff
@@ -396,6 +397,14 @@ def select(
             # create a dummy subquery that we know will fail.
             tags_query = SimpleQuery()
             tags_query.join(self._tags, **kwargs)
+            # If the timespan is requested or contrained, simulate a
```
andy-slac: typo: contrained

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -265,6 +265,31 @@ def dropTemporaryTable(self, table: sqlalchemy.schema.Table) -> None:
         else:
             raise TypeError(f"Table {table.key} was not created by makeTemporaryTable.")
 
+    @contextmanager
+    def temporary_table(self, spec: ddl.TableSpec, name: Optional[str]) -> Iterator[sqlalchemy.schema.Table]:
```
andy-slac: I don't see this used anywhere, I guess it's also for one of the other branches?

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -263,6 +265,31 @@ def dropTemporaryTable(self, table: sqlalchemy.schema.Table) -> None:
...
+        Returns
+        -------
+        table : `sqlalchemy.schema.Table`
+            SQLAlchemy representation of the table.
+        """
+        table = self.makeTemporaryTable(spec=spec, name=name)
+        try:
+            yield table
+        finally:
+            self.dropTemporaryTable(table)
```
andy-slac, line 293: I think Coverity tries to say that there is no unit test for the whole `temporary_table` but somehow it only points to a single line.

andy-slac (approved):

python/lsst/daf/butler/core/timespan.py
TallJimbo: Actually want it to be "`Timespan` or `None`" in this case, but yes.

python/lsst/daf/butler/registry/datasets/byDimensions/summaries.py
TallJimbo: No, I left the data member as public (with a public annotation as a mutable type) because I wasn't bothered by external code doing its own updates to it, and trying to make this block produce data IDs just so they can be unpacked into a different structure didn't seem worthwhile; I prefer the transparency of seeing exactly what's going on over encapsulation in this case.

python/lsst/daf/butler/registry/interfaces/_database.py
TallJimbo: Correct.  An upcoming branch will use it.  I suppose I could also use it to simplify the existing `[Something]QueryResult.materialize` context managers, but I'll be replacing them completely in one of those upcoming branches, too.

python/lsst/daf/butler/registry/interfaces/_database.py
TallJimbo: It's something I ended up removing, because I found that calling code always wanted to manually inspect this value to decide how whether to make a temporary table or use `constant_rows` itself (since one involves a resource whose lifetime should be managed, and the other does not).  I've dropped this first sentence from the docs and reworded the next accordingly.

TallJimbo:
@andy-slac , I'm sending this back to you for another look because Jenkins revealed a problem that led to a non-trivial change, and then I decided to throw in a couple more miscellaneous fixes.  In particular:

- 465ebeb8572c0e7a7032d5917f50d2a6198fc47a has been modified to define the timespan for a non-CALIBRATION collection to a fully-unbounded one, instead of NULL.  That's logically the behavior we want in the query system anyway, and it avoids the need for special IS NULL handling in queries to address that downstream problem Jenkins discovered.

- That change led to d6df482bddf77552db46b7f58ae49aed206bf664, fixing a preexisting bug in how SQLAlchemy converts literals in queries (see also https://github.com/sqlalchemy/sqlalchemy/discussions/8540).

- 5aa83fa7ef2a04e75e2befb7a682dc4251d7f61d is a fix for a bug reported on Slack a couple of weeks ago that I was going to put on this branch but forgot about as I disappeared for vacation.

- 9cb3aec1111283e95904b80b741d7efe09345fb0, 38c22a6c622a56693adabbb323e8fb47ec593258, and 3909cdf3ea2cc199cc197e6f03f6993df814aedf are tiny fixed for things I noticed while I was working through review comments.

andy-slac (approved):
Looks good, no new comments.

PR #732, opened 2022-09-09
DM-35947: Implement live obscore table updates in daf_butler

andy-slac:
Adds new registry manager class responsible for obscore table updates. 

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/interfaces/_obscore.py
+++ b/python/lsst/daf/butler/registry/interfaces/_obscore.py
@@ -27,7 +27,7 @@ __all__ = ["ObsCoreTableManager"]
 
 from abc import abstractmethod
 from collections.abc import Mapping
-from typing import TYPE_CHECKING, Iterable, Type
+from typing import TYPE_CHECKING, Iterable, Optional, Type
 
 from ._versioning import VersionedExtension
 
@@ -90,5 +90,29 @@ class ObsCoreTableManager(VersionedExtension):
         """
         raise NotImplementedError()
 
-    def add_datasets(self, refs: Iterable[DatasetRef]) -> None:
+    def add_datasets(self, refs: Iterable[DatasetRef], collection: Optional[str] = None) -> None:
+        """Possibly add datasets to the obscore table.
+
+        Parameters
+        ----------
+        refs : `iterable` [ `DatasetRef` ]
+            Dataset refs to add. Dataset refs have to be completely expanded.
+            If a record with the same dataset ID is already in obscore table,
+            the dataset is ignored.
+        collection : `str`, optional
+            Name of a collection. This should be set to `None` when datasets
+            are added to their run collection, in which case datasets' own run
+            collection name is used. When existing dataset are being associated
+            with TAGGED or CALIBRATION collections, this parameter should be
+            set to the name of the associated collection.
+
+        Notes
+        -----
+        Dataset data types and collection names are checked against configured
+        list of collections and dataset types, non-matching datasets are
+        ignored and not added to the obscore table. When ``collection`` is
+        specified it is used instead of dataset run name for filtering
+        purposes, but obscore record will still store original dataset run
+        name (when configuration defines the column for it).
+        """
         raise NotImplementedError()
--- a/python/lsst/daf/butler/registry/interfaces/_obscore.py
+++ b/python/lsst/daf/butler/registry/interfaces/_obscore.py
@@ -27,13 +27,13 @@ __all__ = ["ObsCoreTableManager"]
 
 from abc import abstractmethod
 from collections.abc import Mapping
-from typing import TYPE_CHECKING, Iterable, Optional, Type
+from typing import TYPE_CHECKING, Iterable, Type
 
 from ._versioning import VersionedExtension
 
 if TYPE_CHECKING:
     from ...core import DatasetRef, DimensionUniverse
-    from ._collections import CollectionManager
+    from ._collections import CollectionRecord
     from ._database import Database, StaticTablesContext
     from ._datasets import DatasetRecordStorageManager
     from ._dimensions import DimensionRecordStorageManager
@@ -52,7 +52,6 @@ class ObsCoreTableManager(VersionedExtension):
         universe: DimensionUniverse,
         config: Mapping,
         datasets: Type[DatasetRecordStorageManager],
-        collections: CollectionManager,
         dimensions: DimensionRecordStorageManager,
     ) -> ObsCoreTableManager:
         """Construct an instance of the manager.
@@ -71,8 +70,6 @@ class ObsCoreTableManager(VersionedExtension):
             Configuration of the obscore manager.
         datasets : `type`
             Type of dataset manager.
-        collections : `CollectionManager`
-            Manager of Registry collections.
         dimensions: `DimensionRecordStorageManager`
             Manager for Registry dimensions.
 
@@ -94,8 +91,11 @@ class ObsCoreTableManager(VersionedExtension):
         """
         raise NotImplementedError()
 
-    def add_datasets(self, refs: Iterable[DatasetRef], collection: Optional[str] = None) -> None:
+    def add_datasets(self, refs: Iterable[DatasetRef]) -> None:
         """Possibly add datasets to the obscore table.
+
+        This method should be called when new datasets are added to a RUN
+        collection.
 
         Parameters
         ----------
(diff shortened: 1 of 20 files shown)


python/lsst/daf/butler/registry/obscore/_records.py
```diff
@@ -0,0 +1,247 @@
...
+
+        Parameters
+        ----------
+        ref : `DatasetRef`
+            Dataset ref, its DataId must be in expanded form.
+
+        Returns
+        -------
+        record : `dict` [ `str`, `Any` ]
+            ObsCore record represented a dictionary.
```
TallJimbo: Type annotation and code say this could also return `None`.

python/lsst/daf/butler/registry/obscore/_records.py
```diff
@@ -0,0 +1,247 @@
...
+    def _check_collections(self, collection_name: str) -> bool:
+        """Check that specified collection (usually RUN or TAGGED) or any
+        parent CHAINED collection is in the configured collection list.
+        """
+
+        def _all_collections(key: Any) -> Iterator[CollectionRecord]:
+            """Return records of all chained collections."""
+            for collection_record in self.collections.getParentChains(key):
+                yield collection_record
+                yield from _all_collections(collection_record.key)
```
TallJimbo: I'd certainly prefer not to have logic like this run per-row, as it seems to be doing.  I think whenever we insert new datasets they're grouped by RUN in advance, and of course we also get that kind of grouping when associating into TAGGED collections.  It'd be nice to restructure to take advantage of that.

TallJimbo (approved):

python/lsst/daf/butler/registry/obscore/_records.py
andy-slac: I have updated that logic (it has actually moved to _manager.py in later commit) to reduce the number of checks.

python/lsst/daf/butler/registry/interfaces/_obscore.py
```diff
@@ -0,0 +1,163 @@
...
+        Returns
+        -------
+        json : `str`
+            Configuration serialized in JSON format.
+        """
+        raise NotImplementedError()
+
+    def add_datasets(self, refs: Iterable[DatasetRef]) -> None:
+        """Possibly add datasets to the obscore table, this method should be
+        called when new datasets are added to a RUN collection.
```
TallJimbo: Probably best to make the second clause here a new sentence on a separate line.  Same for similar docstrings of other methods in this class.

python/lsst/daf/butler/registry/interfaces/_obscore.py
```diff
@@ -0,0 +1,163 @@
...
+        ignored and not added to the obscore table.
+
+        When configuration parameter ``collection_type`` is not "TAGGED", this
+        method should return immediately.
+        """
+        raise NotImplementedError()
+
+    def disassociate(self, refs: Iterable[DatasetRef], collection: Optional[str] = None) -> None:
+        """Possibly remove datasets from the obscore table, this method should
+        be called when datasets are disassociated from a TAGGED collection.
```
TallJimbo: (Putting a big-picture comment here just because it's a semi-related place to start a thread.)

It occurs to me that we might be able to handle the TAGGED-collection removal problem more robustly with `ON DELETE CASCADE` _if_ we were to create a separate ObsCore table for each `dataset_tags_*` table, and then defined the full ObsCore "table" as an actual SQL `UNION` `VIEW`.  We could also have separate ObsCore tables for RUN-watching vs. TAGGED-watching, and the `UNION` would take care of the potential for duplicates there and let them coexist in the same repo.  I think that might be a simple enough view in practice (there aren't _that_ many distinct sets of dimensions in the dataset types of interest to ObsCore, I think) to avoid confusing the query optimizer too much, in contrast to the rejected "make it all a view" approach.

But it'd of course be a big change from what you have here, and not something I'd advise doing on this ticket, even if you like the idea enough to try it before turning this on in a major repo.  I'm pretty sure it'd also require some new `DatasetRecordStorageManager` methods to provide access to those foreign keys.

python/lsst/daf/butler/registry/interfaces/_obscore.py
```diff
@@ -0,0 +1,163 @@
...
+        When configuration parameter ``collection_type`` is not "RUN", this
+        method should return immediately.
+
+        Note that there is no matching method to remove datasets from obscore
+        table, we assume that removal happens via foreign key constraint to
+        dataset table with "ON DELETE CASCADE" option.
+        """
+        raise NotImplementedError()
+
+    def associate(self, refs: Iterable[DatasetRef], collection: Optional[str] = None) -> None:
```
TallJimbo: I think it's a bit more common for `Manager` interfaces to take `CollectionRecord` objects directly, and leave it to the calling `Registry` code to lookup records from names; that tends to avoid duplicate lookups better.  And it can sometimes save one manager from having to hold an instance of another, which I think is good when it works.

python/lsst/daf/butler/registry/managers.py
```diff
@@ -128,7 +140,37 @@ def fromConfig(cls, config: RegistryConfig) -> RegistryManagerTypes:
...
+        manager_types: Dict[str, Type] = {}
+        for manager in managers:
+            manager_config = config["managers"].get(manager)
+            if isinstance(manager_config, Config):
+                # Expect "cls" and optional "config" sub-keys.
+                manager_config_dict = manager_config.toDict()
+                try:
+                    class_name = manager_config_dict.pop("cls")
+                except KeyError:
+                    raise KeyError(f"'cls' key is not defined in {manager!r} manager configuration")
```
TallJimbo:
```suggestion
                    raise KeyError(f"'cls' key is not defined in {manager!r} manager configuration") from None
```
This will avoid those noisy "while one exception was being handled..." extra tracebacks.

python/lsst/daf/butler/registry/managers.py
```diff
@@ -204,6 +249,16 @@ def loadRepo(self, database: Database) -> RegistryManagerInstances:
         else:
             raise LookupError(f"Registry attribute {_DIMENSIONS_ATTR} is missing from database")
         universe = DimensionUniverse(dimensionConfig)
+        if self.obscore is not None:
+            # Get ObsCore configuration from attributes table, this silently
+            # overrides whatever may come from config file. Idea is that we do
+            # not want to carry around the whole thing, and butler config will
+            # have empty obscore configuration after initialization.
```
TallJimbo, line 261: If someone creates a repo with non-empty obscore configuration, do we actually have logic to make that result in empty configuration in the repo's `butler.yaml` file?

python/lsst/daf/butler/registry/obscore/_manager.py
```diff
@@ -0,0 +1,358 @@
...
+                return None
+            visitDataId = DataCoordinate.standardize(
+                {
+                    "instrument": dataId["instrument"],
+                    "visit": visit,
+                    "detector": dataId["detector"],
+                },
+                universe=self.universe,
+            )
+            records = visit_detector_region_storage.fetch(DataCoordinateIterable.fromScalar(visitDataId))
```
TallJimbo, line 98: It's really not ideal to be doing these fetches one data ID at a time.  Any chance you could at least rework the internal interfaces to support vectorized lookups, even if the implementation remains per-data ID fetches?  I think some changes coming on DM-31725 might make it easier to actually provide a vectorized implementation, so I'm fine with deferring that.

python/lsst/daf/butler/registry/obscore/_manager.py
```diff
@@ -0,0 +1,358 @@
...
+    ):
+        super().__init__(
+            db=db,
+            table=table,
+            schema=schema,
+            universe=universe,
+            config=config,
+            dimensions=dimensions,
+        )
+        self.tagged_collection = tagged_collection
```
TallJimbo, line 232: I wonder if it might be better to have one concrete `ObsCoreLiveTableManager` implementation with a `collection_type` parameter.  The penalty for splitting it into three is just boilerplate, which isn't so bad, but it is a lot of boilerplate compared to what seems like it'd just be a few `if` statements.  And if we did someday work out a way to make RUN-watching and TAGGED-watching coexist in a single repo, a single manager class would be more future-proof.

tests/test_obscore.py
```diff
@@ -0,0 +1,468 @@
...
+        registry.disassociate("tagged", (ref for ref in refs if ref.run == "run2"))
+        rows = self._obscore_select(registry)
+        self.assertEqual(len(rows), 1)
+
+        registry.disassociate("tagged", (ref for ref in refs if ref.run == "run1"))
+        rows = self._obscore_select(registry)
+        self.assertEqual(len(rows), 0)
+
+
+class SQliteObsCoreTest(ObsCoreTests, unittest.TestCase):
```
TallJimbo:
```suggestion
class SqliteObsCoreTest(ObsCoreTests, unittest.TestCase):
```
or `SQLite`, if you prefer.  Just not `SQlite` :slightly_smiling_face: .

TallJimbo (approved):

python/lsst/daf/butler/registry/obscore/_manager.py
TallJimbo: Actually, it looks like there's now a totally different approach to this in `dax_obscore`, and that's already able to handle lots of datasets at once.  Is this just dead code here?

python/lsst/daf/butler/registry/interfaces/_obscore.py
andy-slac: I thought about it but decided to keep things simple. Having multiple table also means that obscore manager would need an additional hook for when a new dataset type is created to also create those dynamic tables for it, and also re-define a view to include this new table. Optimizer is also a concern, I do not have much experience with views and it is hard to guess for me whether a simple view with multiple UNIONs will work OK for all queries. I suspect there maybe cases when optimization breaks for some non-trivial queries. I can try to experiment, but I'd need a working spatial indexing first.

python/lsst/daf/butler/registry/interfaces/_obscore.py
andy-slac: For these methods I only need the name of a collection to match it against the configured tagged collection name. But I can make them to accept a CollectionRecord, just in case it will need to do something more complicated.

python/lsst/daf/butler/registry/managers.py
andy-slac: 👍 Old habits, I still keep forgetting that I need `from` for exceptions.

python/lsst/daf/butler/registry/managers.py
andy-slac: We don't have that logic yet, I'll try to find a place where to add it.

python/lsst/daf/butler/registry/obscore/_manager.py
andy-slac: Would be nice indeed to have a vectorized `fetch`, I'll look at it when DM-31725 is done.

`dax_obscore` fetches too many records as an optimization in the assumption that it is doing bulk conversion and will need a large fraction of those records (which may not be true). That comes with a significant memory overhead which is probably fine for a standalone tool, but I would not want to do the same thing in a generic case. In the case of a regular ingest of raws (this is probably the only dataset that is going to use this logic) there won't be visit defined yet, so this should normally result in 0 records. In this case it will return early after trying to find a visit definition few lines above.

python/lsst/daf/butler/registry/obscore/_manager.py
andy-slac: I started with that idea, but then decided to split it into two classes to avoid that mixed/exclusive state and bunch of `if`s and annoying `mypy` asserts. I can put it back together, should not be too hard.

TallJimbo: If you've already tried it, please stick with this, because you already know better than I do how the alternative would look.

python/lsst/daf/butler/registry/interfaces/_obscore.py
TallJimbo: I'm happy with deferring this indefinitely if you don't think it's an obvious win, and what we have here meets our needs for now.

python/lsst/daf/butler/registry/interfaces/_obscore.py
TallJimbo: Ah, if you don't actually need it for anything else then I don't have a strong opinion on this.

python/lsst/daf/butler/registry/managers.py
andy-slac: I implemented stripping of obscore config before writing `butler.yaml` on [8305317](https://github.com/lsst/daf_butler/pull/732/commits/8305317fbc57c5fb4d15c975f087fcd92fa7ebc1). I'm not super happy about it because it adds some knowledge of obscore to Butler class, but I could not figure out a better place for that logic.

PR #733, opened 2022-09-09
DM-36172: Add explicit transaction around moveToTrash calls

timj:
Without this the in-memory datastore can not rollback a `datastore.trash()` call.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1636,13 +1636,14 @@ class Butler(LimitedButler):
             if collectionType is not CollectionType.RUN:
                 raise TypeError(f"The collection type of '{name}' is {collectionType.name}, not RUN.")
             refs.extend(self.registry.queryDatasets(..., collections=name, findFirst=True))
-        with self.registry.transaction():
-            if unstore:
-                self.datastore.trash(refs)
-            else:
-                self.datastore.forget(refs)
-            for name in names:
-                self.registry.removeCollection(name)
+        with self.datastore.transaction():
+            with self.registry.transaction():
+                if unstore:
+                    self.datastore.trash(refs)
+                else:
+                    self.datastore.forget(refs)
+                for name in names:
+                    self.registry.removeCollection(name)
         if unstore:
             # Point of no return for removing artifacts
             self.datastore.emptyTrash()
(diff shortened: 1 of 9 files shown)


python/lsst/daf/butler/registry/bridge/ephemeral.py
```diff
@@ -62,9 +63,20 @@ def insert(self, refs: Iterable[DatasetIdRef]) -> None:
...
+        for ref in refs:
+            self._trashedIds.remove(ref.getCheckedId())
+
+    def moveToTrash(
+        self, refs: Iterable[DatasetIdRef], transaction: Optional[DatastoreTransaction] = None
+    ) -> None:
         # Docstring inherited from DatastoreRegistryBridge
-        self._trashedIds.update(ref.getCheckedId() for ref in refs)
+        if transaction is None:
+            raise RuntimeError("A transaction must be provided to this method.")
```
TallJimbo: Making this an optional argument that really isn't optional is sort of cheating on the typing; it'd be better for the ABC and the ephemeral implementation to make it non-optional, and then have only the "monolithic" implementation make it optional.

tests/test_butler.py
```diff
@@ -800,9 +800,9 @@ def testPruneCollections(self):
         # rollback because they're still referenced by the CHAINED
         # collection.
         with self.assertRaises(Exception):
-            butler.pruneCollection(run1, pruge=True, unstore=True)
+            butler.pruneCollection(run1, purge=True, unstore=True)
```
TallJimbo, line 803: If you want to go a bit further in improving this test, the exception we're expecting is `sqlalchemy.exc.IntegrityError`.

python/lsst/daf/butler/_butler.py
```diff
@@ -1606,6 +1606,7 @@ def datasetExists(
         ref = self._findDatasetRef(datasetRefOrType, dataId, collections=collections, **kwargs)
         return self.datastore.exists(ref)
 
+    @transactional
```
TallJimbo: Instead of putting this transaction up here, I think we want to start add a `with self.datastore.transaction()` block just before the registry transaction block, so the `emptyTrash` call happens with it but still after the registry transaction closes.  That will keep us from doing the (slow) querying inside the transaction, and it'll make sure we actually try to commit the registry transaction before starting to empty the trash, which might be necessary to detect some errors there.  I think with the outer `@transactional` the inner registry transaction might not do anything, so we might not see errors until after the trash has been emptied.

TallJimbo (approved):
I've requested some adjustments that I do think need to be made, but I think they're straightforward (and I'm about to disappear), so I'm happy to let you take it from here.

python/lsst/daf/butler/_butler.py
timj: `emptyTrash` is traditionally not part of the transaction -- we designed it to try to run regardless and it only deletes the rows if the file deletes happen and we can't rollback the file deletes. I'll put the datastore transaction around the registry transaction.

PR #734, opened 2022-09-21
DM-36174: more miscellaneous query system changes, now requiring Python 3.10

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.10"]
 
     steps:
       - uses: actions/checkout@v3
@@ -95,7 +95,6 @@ jobs:
           file: ./coverage.xml
 
   pypi:
-
     runs-on: ubuntu-latest
     needs: [build_and_test]
     if: startsWith(github.ref, 'refs/tags/')
@@ -109,7 +108,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: "3.10"
 
       - name: Install dependencies
         run: |
--- a/python/lsst/daf/butler/registries/sql.py
+++ b/python/lsst/daf/butler/registries/sql.py
@@ -26,7 +26,20 @@ __all__ = ("SqlRegistry",)
 import contextlib
 import logging
 from collections import defaultdict
-from typing import TYPE_CHECKING, Any, Dict, Iterable, Iterator, List, Mapping, Optional, Set, Tuple, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    Iterable,
+    Iterator,
+    List,
+    Literal,
+    Mapping,
+    Optional,
+    Set,
+    Tuple,
+    Union,
+)
 
 import sqlalchemy
 from lsst.resources import ResourcePathExpression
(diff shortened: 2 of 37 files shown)


timj:
Regarding the news fragment question. I think I'm edging towards noting major internal reorganisations as news worthy but as a misc or performance category. This will at least mean that we can understand where a lot of story points ended up and also let interested parties know that there was a change. Maybe the template should say user visible or major internal rework. 

With the change to 3.10 this will need to have a tweak made to pipe base to stop it trying to use older pythons.

.github/workflows/build.yaml
```diff
@@ -109,7 +108,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: 3.10
```
timj: Needs quotes.

python/lsst/daf/butler/registries/sql.py
```diff
@@ -924,15 +936,88 @@ def _makeQueryBuilder(
...
+              collection wildcards and prohibits ``...`` as a dataset type
+              wildcard.
+
+        doomed_by : `list` [ `str` ]
+            List to append messages indicating why the query is doomed to
+            yield no results.
+
+        Returns
+        -------
+        composition : `defaultdict` [ `DatasetType`, `list` ] ]
```
timj: Extra square bracket.

python/lsst/daf/butler/registries/sql.py
```diff
@@ -924,15 +936,88 @@ def _makeQueryBuilder(
...
+                if not self.defaults.collections:
+                    raise NoDefaultCollectionError("No collections, and no registry default collections.")
+                collections = self.defaults.collections
+            elif mode == "find_first":
+                collections = CollectionSearch.fromExpression(collections)
+            else:
+                collections = CollectionQuery.fromExpression(collections)
+            missing: list[str] = []
+            if mode == "constrain" and datasets is Ellipsis:
+                raise ValueError(
```
timj: Would it be easy to add a test that this exception raises in this scenario?

python/lsst/daf/butler/registry/managers.py
```diff
@@ -229,6 +230,11 @@ class RegistryManagerInstances(
     `Registry`.
     """
 
+    column_types: ColumnTypeInfo
+    """Information about column types that can differ between data repositories
+    and registry instances, including the dimension universe.
+    """
+
```
TallJimbo, line 237: FYI, following some initial work on [DM-36326](https://jira.lsstcorp.org/browse/DM-36326), I'm planning to pare this commit back significantly to just adding the `column_types` attribute here without passing it to all of the manager `initialize` methods.  I'm now pretty confident I can avoid the chain of events that made me need it on downstream branches, and if I do need it after all I'll bring back that part of the commit on a later ticket.  I'm not planning to rebase this branch until you're done with your review, though, unless you'd like me to (I have the updated branch in hand, just not pushed).

timj: Push your updated branch. I had started looking but hadn't made much progress because of JSR distractions.

timj (commented):
Quick comments so I can respond to your new comment. You can push your updated branch.

python/lsst/daf/butler/registries/sql.py
TallJimbo: Yup, will do.

python/lsst/daf/butler/registries/sql.py
```diff
@@ -924,15 +936,86 @@ def _makeQueryBuilder(
...
+              but permits dataset type wildcards.
+
+            - "find_all": this is a query for all datasets in all matched
+              collections.  Permits collection and dataset type wildcards.
+
+            - "constrain": this is a query for something other than datasets,
+              with results constrained by dataset existence.  Permits
+              collection wildcards and prohibits ``...`` as a dataset type
+              wildcard.
+
```
timj: Don't need this blank line.

python/lsst/daf/butler/registries/sql.py
```diff
@@ -948,105 +1031,58 @@ def queryDatasets(
...
+            mode="find_first" if findFirst else "find_all",
+            doomed_by=doomed_by,
+        )
+        parent_results: list[queries.ParentDatasetQueryResults] = []
+        for parent_dataset_type, components_for_parent in dataset_composition.items():
+            # The full set of dimensions in the query is the combination of
+            # those needed for the DatasetType and those explicitly requested,
+            # if any.
+            dimension_names = set(parent_dataset_type.dimensions.names)
+            if dimensions is not None:
```
timj, line 1048: We never call this with a `dimensions` parameter?

python/lsst/daf/butler/registries/sql.py
```diff
@@ -1063,58 +1099,31 @@ def queryDataIds(
...
-                summary, doomed_by=[f"Dataset type {name} is not registered." for name in missing]
-            )
-            for datasetType in standardizedDatasetTypes:
-                builder.joinDataset(
-                    datasetType,
-                    collections,
-                    isResult=False,
-                )
+            builder = self._makeQueryBuilder(summary, doomed_by=doomed_by)
+            for datasetType in dataset_composition.keys():
```
timj: Is `.keys()` needed here?

timj (approved):

python/lsst/daf/butler/registries/sql.py
TallJimbo: Looks like we don't.  I don't remember the history, and I can't recall a use case.  I thought we might have used it in QG generation, but I just checked, and we don't.  So maybe it's worth an RFC to drop it, but I'd rather not spend any energy on it right now.

python/lsst/daf/butler/registries/sql.py
TallJimbo: Just my usual preference for being explicit when iterating over mappings, but I think you feel more strongly than me so I'll change it.

timj: Okay. I think we've had this conversation before 😄 . Keep it as it is.

timj: Ah. I see you changed it. Okay.

timj:
That is a very odd mypy error:

> /opt/hostedtoolcache/Python/3.10.7/x64/lib/python3.10/site-packages/numpy/__init__.pyi:636: error: Positional-only parameters are only supported in Python 3.8 and greater

TallJimbo:
I'm pretty baffled by that mypy error, too.  The way out of this might be that I don't think we want mypy looking at `numpy` at all, and historically I think it probably wasn't (after all, daf_butler _barely_ uses numpy at all, and we certainly don't depend on numpy's type annotations for correct static typing of our own code).  So I'll try to tell mypy just to ignore numpy and hope that makes it go away.

timj:
I was convinced we only used numpy in tests as an optional dependency but I see that somehow we've pulled it in as a dependency of the scripting layer.

TallJimbo:
I'm sure it's the astropy.table formatting that brings it in at the script layer.

TallJimbo:
Well, my fix to our mypy.ini didn't help but I found this:

https://github.com/python/mypy/issues/13627

Apparently it's a mypy bug that will be fixed in the next release.  Not sure if we should just ignore it for now or try to patch `rubin_workflows`; I don't think there's anything I can do here.

timj:
What sort of patch to rubin workflows are you thinking of?

PR #736, opened 2022-09-21
DM-36312: Deprecate support for component datasets in Registry

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registries/sql.py
+++ b/python/lsst/daf/butler/registries/sql.py
@@ -929,6 +929,10 @@ class SqlRegistry(Registry):
             parent datasets were not matched by the expression.
             Fully-specified component datasets (`str` or `DatasetType`
             instances) are always included.
+
+            Values other than `False` are deprecated, and only `False` will be
+            supported after v26.  After v27 this argument will be removed
+            entirely.
         mode : `str`, optional
             The way in which datasets are being used in this query; one of:
 
--- a/python/lsst/daf/butler/registries/sql.py
+++ b/python/lsst/daf/butler/registries/sql.py
@@ -418,7 +418,12 @@ class SqlRegistry(Registry):
 
     def getDatasetType(self, name: str) -> DatasetType:
         # Docstring inherited from lsst.daf.butler.registry.Registry
-        return self._managers.datasets[name].datasetType
+        parent_name, component = DatasetType.splitDatasetTypeName(name)
+        storage = self._managers.datasets[parent_name]
+        if component is None:
+            return storage.datasetType
+        else:
+            return storage.datasetType.makeComponentDatasetType(component)
 
     def supportsIdGenerationMode(self, mode: DatasetIdGenEnum) -> bool:
         # Docstring inherited from lsst.daf.butler.registry.Registry
@@ -435,9 +440,10 @@ class SqlRegistry(Registry):
     ) -> Optional[DatasetRef]:
         # Docstring inherited from lsst.daf.butler.registry.Registry
         if isinstance(datasetType, DatasetType):
-            storage = self._managers.datasets[datasetType.name]
+            parent_name, component = datasetType.nameAndComponent()
         else:
-            storage = self._managers.datasets[datasetType]
+            parent_name, component = DatasetType.splitDatasetTypeName(datasetType)
+        storage = self._managers.datasets[parent_name]
         dataId = DataCoordinate.standardize(
             dataId,
             graph=storage.datasetType.dimensions,
@@ -460,6 +466,8 @@ class SqlRegistry(Registry):
                 continue
             result = storage.find(collectionRecord, dataId, timespan=timespan)
             if result is not None:
+                if component is not None:
+                    return result.makeComponentRef(component)
                 return result
 
         return None
--- a/python/lsst/daf/butler/registries/sql.py
+++ b/python/lsst/daf/butler/registries/sql.py
@@ -1038,7 +1038,7 @@ class SqlRegistry(Registry):
                 check=check,
                 datasets=[parent_dataset_type],
             )
-            builder = self._makeQueryBuilder(summary, doomed_by=doomed_by)
+            builder = self._makeQueryBuilder(summary)
             # Add the dataset subquery to the query, telling the QueryBuilder
             # to include the rank of the selected collection in the results
             # only if we need to findFirst.  Note that if any of the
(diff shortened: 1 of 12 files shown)


python/lsst/daf/butler/registries/sql.py
```diff
@@ -435,9 +440,14 @@ def findDataset(
     ) -> Optional[DatasetRef]:
         # Docstring inherited from lsst.daf.butler.registry.Registry
         if isinstance(datasetType, DatasetType):
-            storage = self._managers.datasets[datasetType.name]
+            parent_name, component = datasetType.nameAndComponent()
+            storage = self._managers.datasets[parent_name]
         else:
-            storage = self._managers.datasets[datasetType]
+            parent_name, component = DatasetType.splitDatasetTypeName(datasetType)
```
timj, line 445: It's a shame this logic is duplicated from `getDatasetType` but it looks like you need `storage.find` to work. How about:

```python
if not isinstance(datasetType, DatasetType):  # or is a str
    registryDatasetType = self.getDatasetType(datasetType)
else:
    registryDatasetType = datasetType
parent_name, component = registryDatasetType.nameAndComponent()
storage = self._managers.datasets[parent_name]
```
?

python/lsst/daf/butler/registries/sql.py
```diff
@@ -435,9 +440,14 @@ def findDataset(
...
+            storage = self._managers.datasets[parent_name]
+            datasetType = storage.datasetType
+            if component is not None:
+                datasetType.makeComponentDatasetType(component)
```
timj: This line seems like we should be capturing the return value.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -342,8 +334,15 @@ def resolve_wildcard(
     ) -> dict[DatasetType, list[str | None]]:
         wildcard = DatasetTypeWildcard.from_expression(expression)
         result: defaultdict[DatasetType, set[str | None]] = defaultdict(set)
+        deprecation_message = (
+            "Querying for component datasets via Registry query methods is deprecated in favor of using "
+            "DatasetRef and DatasetType methods on parent datasets. Only components=False will be supported "
+            "after v26."
```
timj: Maybe clarify that the parameter will be removed completely after v27 (which is a long time from now). We will need a jira ticket linked to a release somehow.

tests/test_simpleButler.py
```diff
@@ -166,10 +166,10 @@ def testDatasetTransfers(self):
...
         """Test exporting component datasets and then importing them.
 
-        This test intentionally does not depend on whether just the component
-        is exported and then imported vs. the full composite dataset, because
-        I don't want it to assume more than it needs to about the
-        implementation.
+        This test exports component datasets and then checks via the parent
+        dataset type, since this is the only valid way to handle component
+        exports (since Registry cannot record just a component without a
+        parent, even if Datastore can).
```
timj: I'm effectively disabling the ability of datastore to store a bare component in #737.

timj (approved):
Looks good.

python/lsst/daf/butler/registries/sql.py
TallJimbo: Aha!  We don't actually use `datasetType` after this, just `storage.datasetType` and `component`.  So I can remove this check, and the line above it, and I can move the `storage = ` line outside the `if`.  And then there's actually very little duplicated between this method and `getDatasetType`, addressing your other comment.

That does mean that `findDataset` does not respect a storage class override that's passed to it, but that seems to have been the behavior before (that's handled in `Butler` instead).  As it turns out, the `Registry.query` methods _do_ respect storage class overrides, but that was largely an accident.  I'm formalizing that on DM-31725, and it'll be easier to make `findDataset` respect overrides then, too.

python/lsst/daf/butler/registries/sql.py
TallJimbo: See next comment; thanks to that, duplication is much reduced.

tests/test_simpleButler.py
TallJimbo: Where does that leave this test?  It seems to still be working.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
TallJimbo: I've created DM-36457 for the post-v27 removals, but haven't tried to create release tickets for v26 or v27 (neither exist yet).

tests/test_simpleButler.py
timj: I removed the test in datastore, although I didn't make datastore.put reject a component dataset type. Maybe that was a mistake, I hadn't realized we had other tests that did disassembly outside datastore.

TallJimbo: Sounds like I should probably keep it around for now, then, until you get a chance to figure out the code paths involved and disable them more fully?

TallJimbo:
@timj , I've addressed all review comments (in fixup commits for now), and added a new commit for the problem @kfindeisen reported on Slack.  Can you take a look at that?

tests/test_simpleButler.py
timj: I'm a bit confused what is happening in this test since registry doesn't do composite disassembly.

timj (approved):

tests/test_simpleButler.py
TallJimbo: Ah, it looks like the doctring had just gotten out of date, and I took it at face-value when I updated it.  It just exports the composite and checks that the components are accessible after import.  But this ticket makes that check for components impossible to do directly, and doing it indirectly via the parent datasets just makes this test the same as some others, so I'll delete it.

python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -2666,6 +2666,16 @@ def testQueryResultSummaries(self):
         self.assertEqual(query5.count(exact=True), 0)
         messages = list(query5.explain_no_results())
         self.assertFalse(messages)
+        # This query should yield results from one dataset type but not the
+        # other, which is not registered.
+        query5 = registry.queryDatasets(["bias", "nonexistent"], collections=["biases"])
```
kfindeisen, line 2671: Perfect, thanks!

PR #737, opened 2022-09-28
DM-4551: Allow override of storage class on butler.get()

timj:
Support `data = butler.getDirect(ref, readStorageClass="AstropyTable")` style overrides.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1235,7 +1235,7 @@ class Butler(LimitedButler):
         ref: DatasetRef,
         *,
         parameters: Optional[Dict[str, Any]] = None,
-        readStorageClass: Optional[Union[StorageClass, str]] = None,
+        storageClass: Optional[Union[StorageClass, str]] = None,
     ) -> Any:
         """Retrieve a stored dataset.
 
@@ -1250,13 +1250,19 @@ class Butler(LimitedButler):
         parameters : `dict`
             Additional StorageClass-defined options to control reading,
             typically used to efficiently read only a subset of the dataset.
+        storageClass : `StorageClass` or `str`, optional
+            The storage class to be used to override the Python type
+            returned by this method. By default the returned type matches
+            the dataset type definition for this dataset. Specifying a
+            read `StorageClass` can force a different type to be returned.
+            This type must be compatible with the original type.
 
         Returns
         -------
         obj : `object`
             The dataset.
         """
-        return self.datastore.get(ref, parameters=parameters, readStorageClass=readStorageClass)
+        return self.datastore.get(ref, parameters=parameters, storageClass=storageClass)
 
     def getDirectDeferred(
         self, ref: DatasetRef, *, parameters: Union[dict, None] = None
@@ -1346,7 +1352,7 @@ class Butler(LimitedButler):
         *,
         parameters: Optional[Dict[str, Any]] = None,
         collections: Any = None,
-        readStorageClass: Optional[Union[StorageClass, str]] = None,
+        storageClass: Optional[Union[StorageClass, str]] = None,
         **kwargs: Any,
     ) -> Any:
         """Retrieve a stored dataset.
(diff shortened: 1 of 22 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -1229,7 +1230,13 @@ def put(
 
         return ref
 
-    def getDirect(self, ref: DatasetRef, *, parameters: Optional[Dict[str, Any]] = None) -> Any:
+    def getDirect(
+        self,
+        ref: DatasetRef,
+        *,
+        parameters: Optional[Dict[str, Any]] = None,
+        readStorageClass: Optional[Union[StorageClass, str]] = None,
```
timj: As a general question, if this override facility is going to be used a lot, should we consider a shorter parameter name to make it easier to use? `read_sc`?

cc/ @erykoff

python/lsst/daf/butler/core/datasets/type.py
```diff
@@ -532,6 +532,35 @@ def makeAllComponentDatasetTypes(self) -> List[DatasetType]:
...
+        -------
+        modified : `DatasetType`
+            A new dataset type that is the same as the current one but with
+            a different storage class.
+
+        Notes
+        -----
+        If this is a component dataset type, the parent storage class
+        will be retained.
+        """
```
TallJimbo, line 554: Would it be worthwhile to add a `is_compatible` check of some kind here?  (Genuine question; I can also imagine that it'd mostly just be a wasteful repeated check.)

python/lsst/daf/butler/_butler.py
TallJimbo: I'd at least just say `storageClass` or `storage_class`, since everything to do with `Butler.get` is "read" from the user perspective.

TallJimbo (approved):

python/lsst/daf/butler/core/datasets/type.py
timj: We could. That would force the storage class and python type to be loaded early. That might be a good idea, although that would lead to different error reporting if someone passes in an override vs someone passing in an externally-created dataset type (which is how pipelines work). This override will trigger the incompatibility error later on. I'm a bit torn -- raising early is generally better than raising late (and the stack trace from formatters is a nightmare until python 3.11 lets us easily add an extra context message to the existing stack trace). I think we would have to be a bit looser on the compatibility check here because we don't know if this is an override for write or override for read so would only complain if it's not compatible with either direction, whereas the later test is explicitly for reading. I think I've talked myself into doing the basic test.

python/lsst/daf/butler/_butler.py
```diff
@@ -1359,6 +1373,12 @@ def get(
             Collections to be searched, overriding ``self.collections``.
             Can be any of the types supported by the ``collections`` argument
             to butler construction.
+        readStorageClass : `StorageClass` or `str`, optional
```
erykoff: This line needs an update.

python/lsst/daf/butler/_butler.py
```diff
@@ -1243,13 +1250,19 @@ def getDirect(self, ref: DatasetRef, *, parameters: Optional[Dict[str, Any]] = N
...
 
         Returns
         -------
         obj : `object`
             The dataset.
         """
-        return self.datastore.get(ref, parameters=parameters)
+        return self.datastore.get(ref, parameters=parameters, storageClass=storageClass)
 
     def getDirectDeferred(
```
timj, line 1267: I had forgotten about the deferred versions. Given that this takes a parameters parameter, I imagine we would like the ability to force a new storage class to be used later for the get...

TallJimbo: Up to you.  In the primary `PipelineTask` use case for `DeferredDatasetHandle`, I'd expect the storage class override to be declared in the connections so we wouldn't need that.

TallJimbo (approved):

python/lsst/daf/butler/_butler.py
timj: I added it, although I decided not to add it to the InMemoryDatasetHandle since it seemed a bit odd to create one of those with a python object and immediately declare you wanted it to be returned as a completely different type.

PR #738, opened 2022-10-04
DM-36412: Add new env var DAF_BUTLER_CACHE_DIRECTORY_IF_UNSET

timj:
This only overrides the cache directory if nothing has set it explicitly from config or the other env var.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/datastoreCacheManager.py
+++ b/python/lsst/daf/butler/core/datastoreCacheManager.py
@@ -40,6 +40,7 @@ import shutil
 import tempfile
 from abc import ABC, abstractmethod
 from collections import defaultdict
+from random import Random
 from typing import (
     TYPE_CHECKING,
     Dict,
@@ -455,9 +456,15 @@ class DatastoreCacheManager(AbstractDatastoreCacheManager):
     The expiration mode should take the form ``mode=threshold`` so for
     example to configure expiration to limit the cache directory to 5 datasets
     the value would be ``datasets=5``.
+
+    Additionally the ``$DAF_BUTLER_CACHE_DIRECTORY_IF_UNSET`` environment
+    variable can be used to indicate that this directory should be used
+    if no explicit directory has been specified from configuration or from
+    the ``$DAF_BUTLER_CACHE_DIRECTORY`` environment variable.
     """
 
     _temp_exemption_prefix = "exempt/"
+    _tmpdir_prefix = "butler-cache-dir-"
 
     def __init__(self, config: Union[str, DatastoreCacheManagerConfig], universe: DimensionUniverse):
         super().__init__(config, universe)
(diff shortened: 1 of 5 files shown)


python/lsst/daf/butler/core/datastoreCacheManager.py
```diff
@@ -595,6 +601,14 @@ def move_to_cache(self, uri: ResourcePath, ref: DatasetRef) -> Optional[Resource
         # item.
         self._expire_cache()
 
+        # The above reset the in-memory cache status. It's entirely possible
+        # that another process has just cached this file (if multiple
+        # processes are caching on read), so check our in-memory cache
+        # before attempting to cache the dataset.
+        path_in_cache = cached_location.relative_to(self.cache_directory)
+        if path_in_cache and path_in_cache in self._cache_entries:
+            return cached_location
```
andy-slac, line 674: A concurrency question - is `cached_location.transfer_from()` an atomic operation, meaning that if file path exists then it is a complete and usable file?

tests/test_datastore.py
```diff
@@ -1352,6 +1353,54 @@ def testNoCacheDirReversed(self):
...
+    metric1: false
+        """
+        cache_manager = self._make_cache_manager(config_str)
+
+        # This environment variable should override the config value.
+        with unittest.mock.patch.dict(os.environ, {"DAF_BUTLER_CACHE_DIRECTORY_IF_UNSET": env_dir.ospath}):
+            cache_manager = self._make_cache_manager(config_str)
+        self.assertEqual(cache_manager.cache_directory, env_dir)
+
+        # If both environment variables are set the not IF_UNSET should win.
```
andy-slac: :master-yoda: "are set the not" -> "are not set the"?

andy-slac (approved):
Looks good, one easy question.

python/lsst/daf/butler/core/datastoreCacheManager.py
timj: Yes. I was also worrying about the transfer from but trying not to think about it. The design of the cache was mainly worrying about *output* datasets which we can be fairly sure are not going to be duplicates. The problem I was encountering here was that we are caching on read and all three of the quanta are wanting the same dark to be read and each quantum tries to cache it for later.

You are correct that `transfer_from` in `copy` mode is not in fact atomic in that it does a copy to the final destination and does not do a copy to a temp file followed by a rename.

I'm pretty sure I can't do atomic writes on object store but I can change the internal implementation of the local file system implementation. I'm not sure whether that should always be done or if it should be controlled by a parameter that would be a hint to the transfer command but could be ignored by URI schemes that don't support it.

timj: I'm addressing the transfer_from problem in lsst/resources#33

PR #742, opened 2022-10-08
DM-36410: Support "move" ingest into chained datastore

timj:
Treat it as a copy and then clean up.

## Checklist

- [x] ran Jenkins


andy-slac (approved):
Looks good.

PR #743, opened 2022-10-11
DM-36517: Add support for ChainedDatastore.export()

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastores/chainedDatastore.py
+++ b/python/lsst/daf/butler/datastores/chainedDatastore.py
@@ -979,16 +979,10 @@ class ChainedDatastore(Datastore):
         # better than exporting files from the first datastore and then finding
         # that one is missing but is not in the second datastore either.
         known = [datastore.knows_these(refs) for datastore in self.datastores]
-        combined_known: dict[DatasetRef, bool] = {}
+        refs_known: set[DatasetRef] = set()
         for known_to_this in known:
-            for ref, knows_this in known_to_this.items():
-                # If we have not known it yet, store the newer value.
-                if not combined_known.get(ref):
-                    combined_known[ref] = knows_this
-        missing_count = 0
-        for knows_this in combined_known.values():
-            if not knows_this:
-                missing_count += 1
+            refs_known.update({ref for ref, knows_this in known_to_this.items() if knows_this})
+        missing_count = len(refs) - len(refs_known)
         if missing_count:
             raise FileNotFoundError(f"Not all datasets known to this datastore. Missing {missing_count}")
 
--- a/python/lsst/daf/butler/datastores/chainedDatastore.py
+++ b/python/lsst/daf/butler/datastores/chainedDatastore.py
@@ -48,6 +48,7 @@ from lsst.utils import doImportType
 if TYPE_CHECKING:
     from lsst.daf.butler import Config, DatasetType, LookupKey, StorageClass
     from lsst.daf.butler.registry.interfaces import DatasetIdRef, DatastoreRegistryBridgeManager
+    from lsst.resources import ResourcePathExpression
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -262,6 +263,18 @@ def knows(self, ref: DatasetRef) -> bool:
...
 
+    def knows_these(self, refs: Iterable[DatasetRef]) -> dict[DatasetRef, bool]:
+        # Docstring inherited from the base class.
+        known: dict[DatasetRef, bool] = {}
+        for datastore in self.datastores:
+            known.update(datastore.knows_these(refs))
+
+            # No need to check in next datastore for refs that are known.
+            # We only update entries that were initially False.
+            refs = [ref for ref, known in known.items() if not known]
```
andy-slac: `known in known.items() if not known` - I know 😉 that this should work OK, but maybe rename loop variable into something else to reduce potential confusion?

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -934,3 +947,75 @@ def export_records(self, refs: Iterable[DatasetIdRef]) -> Mapping[str, Datastore
                 all_records[name] = record_data
 
         return all_records
+
+    def export(
+        self,
+        refs: Iterable[DatasetRef],
+        *,
+        directory: Optional[ResourcePathExpression] = None,
+        transfer: Optional[str] = "auto",
```
andy-slac, line 956: Is it OK to change the default value of `transfer` from the base class?

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -934,3 +947,75 @@ def export_records(self, refs: Iterable[DatasetIdRef]) -> Mapping[str, Datastore
...
+        combined_known: dict[DatasetRef, bool] = {}
+        for known_to_this in known:
+            for ref, knows_this in known_to_this.items():
+                # If we have not known it yet, store the newer value.
+                if not combined_known.get(ref):
+                    combined_known[ref] = knows_this
+        missing_count = 0
+        for knows_this in combined_known.values():
+            if not knows_this:
+                missing_count += 1
```
andy-slac: Maybe something like 

    missing_count = list(combined_known.values()).count(False)

going to be shorter/faster?

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -934,3 +947,75 @@ def export_records(self, refs: Iterable[DatasetIdRef]) -> Mapping[str, Datastore
...
+        # before we try to run the export. This can be a little messy but is
+        # better than exporting files from the first datastore and then finding
+        # that one is missing but is not in the second datastore either.
+        known = [datastore.knows_these(refs) for datastore in self.datastores]
+        combined_known: dict[DatasetRef, bool] = {}
+        for known_to_this in known:
+            for ref, knows_this in known_to_this.items():
+                # If we have not known it yet, store the newer value.
+                if not combined_known.get(ref):
+                    combined_known[ref] = knows_this
```
andy-slac: Maybe use `collections.Counter` for `combined_known`, you can just sum 'True' values with `update()` instead of doing an inner loop (but internally it's probably the same loop).

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -934,3 +947,75 @@ def export_records(self, refs: Iterable[DatasetIdRef]) -> Mapping[str, Datastore
...
+                # Try the next datastore.
+                continue
+
+            for ref, export in zip(filtered, this_export):
+                # Get the position and also delete it from the list.
+                exported[ref_positions.pop(ref)] = export
+
+        for i, dataset in enumerate(exported):
+            if dataset is None:
+                raise FileNotFoundError(f"Failed to export dataset {refs[i]}.")
```
andy-slac, line 1019: Would it make sense to check that all datasets were exported before yielding any of them?

andy-slac (approved):

python/lsst/daf/butler/datastores/chainedDatastore.py
timj: Wow. Not sure what I was thinking with that variable naming...

python/lsst/daf/butler/datastores/chainedDatastore.py
timj: Why not? I think the base class was written before I came up with the "auto" meaning (when `transfer=None` means "don't transfer" and we determined that was generally a bad default but the base class doesn't really have a choice because we haven't used an enum for the transfer options.

python/lsst/daf/butler/datastores/chainedDatastore.py
timj: I was trying to avoid scanning through the list of refs an additional time but I can do that.

python/lsst/daf/butler/datastores/chainedDatastore.py
andy-slac: I was thinking that the client can call the method assuming that default for `transfer` is `None`, as declared in the base class, but in reality `transfer` will be "auto" (not sure if there is a difference between the two).

timj: For a chained datastore `transfer=None` could be a bit tricky. `None` means "do not export any files to the given directory, leave them in place" which is fine if you have a single datastore but in a chained datastore you could end up with some entries from one datastore and also entries from another. I suppose I could switch to "auto" if `directory != None` rather than raise below in that scenario.

python/lsst/daf/butler/datastores/chainedDatastore.py
timj: I'm not really sure I understand how to use `collections.Counter` here. It did make me consider this alternative though:

```python
refs_known: set[DatasetRef] = set()
for known_to_this in known:
    refs_known.update({ref for ref, knows_this in known_to_this.items() if knows_this})
missing_count = len(refs) - len(refs_known)
```
which is much shorter and seems to work.

python/lsst/daf/butler/datastores/chainedDatastore.py
timj: Thinking about it a bit more, this check is mostly there for mypy so it knows the method can't yield `None`.  The missing check earlier in the method should be enough to know that every dataset is known to at least one datastore. I can add a note that this is for mypy.

PR #744, opened 2022-10-12
DM-36489: Add pgSphere support to obscore manager

andy-slac:
Obscore manager adds generic plugin mechanisms to support spatial columns
and indices. A specific implementation using pgSphere extension adds
two columns and corresponding indices that contain pgSphere polygon and
position (center of bounding circle) to obscore table.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/obscore/pgsphere.py
+++ b/python/lsst/daf/butler/registry/obscore/pgsphere.py
@@ -1,0 +1,188 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import annotations
+
+__all__ = ["PgSphereObsCorePlugin"]
+
+import warnings
+from collections.abc import Callable, Mapping, Sequence
+from typing import TYPE_CHECKING, Any, Optional, Tuple
+
+import sqlalchemy
+from lsst.daf.butler import DatasetId
+from lsst.sphgeom import ConvexPolygon, LonLat, Region
+from sqlalchemy.dialects.postgresql.base import ischema_names
+from sqlalchemy.types import UserDefinedType
+
+from ...core import ddl
+from ._spatial import MissingDatabaseError, RegionTypeWarning, SpatialObsCorePlugin
+
+if TYPE_CHECKING:
+    from ..interfaces import Database, StaticTablesContext
+    from ._records import Record
+    from ._schema import ObsCoreSchema
+
+
+class PgSpherePoint(UserDefinedType):
+    """SQLAlchemy type representing pgSphere point (spoint) type.
+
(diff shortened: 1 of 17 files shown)


python/lsst/daf/butler/registry/obscore/pgsphere.py
```diff
@@ -0,0 +1,182 @@
...
+        cls, *, name: str, config: Mapping[str, Any], db: Optional[Database]
+    ) -> SpatialObsCorePlugin:
+        # docstring inherited.
+
+        if db is None:
+            raise MissingDatabaseError("Database access is required for pgSphere plugin")
+
+        # Check that engine is Postgres and pgSphere extension is enabled.
+        if db.dialect.name != "postgresql":
+            raise RuntimeError("PgSphere spatial plugin for obscore requires requires PostgreSQL database.")
```
TallJimbo:
```suggestion
            raise RuntimeError("PgSphere spatial plugin for obscore requires PostgreSQL database.")
```

python/lsst/daf/butler/registry/obscore/pgsphere.py
```diff
@@ -0,0 +1,182 @@
...
+                ),
+            )
+        )
+        # Spatial columns need GIST index type
+        table_spec.indexes.add(ddl.IndexSpec(self._region_column_name, postgresql_using="gist"))
+        table_spec.indexes.add(ddl.IndexSpec(self._position_column_name, postgresql_using="gist"))
+
+    def make_extra_tables(self, schema: ObsCoreSchema, context: StaticTablesContext) -> None:
+        # docstring inherited.
+        return
```
TallJimbo: Do you have a future plugin in mind that will actually use the `make_extra_tables` (etc) hook?  If not, I wonder if it would be better to remove it all for now.

python/lsst/daf/butler/registry/obscore/pgsphere.py
```diff
@@ -0,0 +1,182 @@
...
+            return None, None
+
+        record: Record = {}
+        circle = region.getBoundingCircle()
+        record[self._position_column_name] = LonLat(circle.getCenter())
+
+        # Presently we can only handle polygons
+        if isinstance(region, ConvexPolygon):
+            poly_points = [LonLat(vertex) for vertex in region.getVertices()]
+            record[self._region_column_name] = poly_points
```
TallJimbo, line 174: Should we warn if we see something that isn't a polygon?

TallJimbo (approved):
Looks good; just a few minor comments.

But codecov seems pretty convinced that your tests aren't actually exercising any of the pgSphere stuff, despite the tests' clear intent to do that.  I suspect that means something's gone wrong in either the installation or the test-skipping logic.

python/lsst/daf/butler/registry/obscore/pgsphere.py
andy-slac: I wanted to add a plugin which just stores dataset region's pixel indices in a separate table, but decided to postpone that. I'll drop that method, it should simplify things a little.

python/lsst/daf/butler/registry/obscore/pgsphere.py
andy-slac: I was worried that if non-polygon regions become common we'll get many annoying warnings. I guess we should worry about it when we see it, I'll add a warning here.

andy-slac: Do we prefer `warnings.warn()` or `logging.warning()` for such cases?

TallJimbo: I'm afraid I have no idea of the pros and cons.  I tend to go by whether there's already a logger available nearby, but that doesn't mean that's a good approach.

andy-slac: OK, I'll add `warnings.warn()` and hope it never fires. 🙂

andy-slac:
> But codecov seems pretty convinced that your tests aren't actually exercising any of the pgSphere stuff

I'm not sure that this is true. I checked `build_and_test` log and every `PostgresPgSphereObsCoreTest` is tagged as PASSED. Also this codecov page: https://app.codecov.io/gh/lsst/daf_butler/pull/744 shows that coverage is ~85% for `pgsphere.py`.

PR #751, opened 2022-11-08
DM-36808: Add citation information

timj:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -27,6 +27,9 @@ authors:
   orcid: "https://orcid.org/0000-0002-8324-0880"
 title: "lsst-daf-butler: Abstracted Data Access of Astronomical Data"
 repository-code: "https://github.com/lsst/daf_butler"
+identifiers:
+- type: swh
+  value: "swh:1:dir:ef572b4bb431e898d279d4de786bc462bd5ca466"
 abstract: |
   The Rubin Observatory's Data Butler is designed to allow data file location
   and file formats to be abstracted away from the people writing the science
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,6 +1,7 @@
 cff-version: 1.2.0
 message: "Please cite the following when using this software: https://doi.org/10.1117/12.2629569"
 authors:
+- name: "Rubin Observatory Data Management Team"
 - family-names: Jenness
   given-names: Tim
   orcid: "https://orcid.org/0000-0001-5982-167X"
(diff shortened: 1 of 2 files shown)


CITATION.cff
```diff
@@ -0,0 +1,71 @@
+cff-version: 1.2.0
+message: "Please cite the following when using this software: https://doi.org/10.1117/12.2629569"
```
ktlim, line 2: I think the commonly accepted paradigm is to say something like "If you use this software, please cite it using this metadata." so that you don't have to update the link twice (here in the `message` and later in the `preferred-citation`).

CITATION.cff
```diff
@@ -0,0 +1,71 @@
...
+authors:
+- family-names: Jenness
```
ktlim, line 5: Should the software authors be only the paper authors?  That seems kind of restrictive.  But having all code contributors is hard to maintain.  I was going to use something like:
```
authors:
  - name: "Rubin Observatory Data Management Team"
```

timj: I was listing people who contributed to the software itself, not the paper authors. I can't find a CITATION.cff file in the astropy package so I'm not sure what others do for projects where it's more than a couple of contributors. Looking at the CITATION.cff docs it allows what you have and additional people so maybe I should add `name` (DM Team)  as first author and then core team after that.

natelust: maybe say "Rubin Observatory Data Management Team, including Tim J., Jim B., ....., and many other contributers"

timj: I think:

```yaml
authors:
  - name: "Rubin Observatory Data Management Team"
  - family-names: Jenness
    given-names: Tim
  - family-names: Bosch
    given-names: James
  ...
```
works within the standard.

CITATION.cff
timj: Yes, although I was worried that might lead to an attempt to cite the software itself which doesn't have a DOI or ASCL ID and so thought I should be explicit.

timj:
For the first attempt GitHub seems to generate a bibtex entry of:

```bib
@article{Jenness_The_Vera_C,
author = {Jenness, Tim and Bosch, James F. and Salnikov, Andrei and Lust, Nate B. and Pease, Nathan M. and Gower, Michelle and Kowalik, Mikolaj and Dubois-Felsmann, Gregory P. and Mueller, Fritz and Schellart, Pim},
title = {{The Vera C. Rubin Observatory Data Butler and pipeline execution system}}
}
```

which unfortunately lacks the DOI information so I must have got that wrong somewhere.

PR #753, opened 2022-11-11
DM-36974: Clean up mypy annotations in lsst.daf.butler.script

timj:
We were using annotations in many of the files but allowed mypy to ignore them if they were not in typed functions. This fixes the entire directory so that everything is annotated. Mostly annotations and some very minor code rearrangements to let the annotations work (and some new variables created for the same reason)

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/script/_associate.py
+++ b/python/lsst/daf/butler/script/_associate.py
@@ -18,13 +18,23 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+from collections.abc import Iterable
 
 from .._butler import Butler
 from ..registry import CollectionType
 from .queryDatasets import QueryDatasets
 
 
-def associate(repo, collection, dataset_type, collections, where, find_first):
+def associate(
+    repo: str,
+    collection: str,
+    dataset_type: Iterable[str],
+    collections: Iterable[str],
+    where: str | None,
+    find_first: bool,
+) -> None:
     """Add existing datasets to a CHAINED collection."""
 
     butler = Butler(repo, writeable=True)
--- a/python/lsst/daf/butler/script/_pruneDatasets.py
+++ b/python/lsst/daf/butler/script/_pruneDatasets.py
@@ -18,13 +18,18 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
 
-
+from collections.abc import Callable, Iterable
 from enum import Enum, auto
+from typing import TYPE_CHECKING, Any
 
 from .._butler import Butler
 from ..registry import CollectionType
 from .queryDatasets import QueryDatasets
+
+if TYPE_CHECKING:
+    from astropy.table import Table
 
 
 class PruneDatasetsResult:
@@ -35,9 +40,9 @@ class PruneDatasetsResult:
 
     Parameters
     ----------
-    tables : `list` [``astropy.table.table``], optional
+    tables : `list` [`astropy.table.Table`], optional
         The astropy tables that will be or were deleted, by default None.
-    state : ``PruneDatasetsResult.State``, optional
+    state : `PruneDatasetsResult.State`, optional
         The initial state of execution of the action, if `None` the result
         state is ``INIT``, by default None.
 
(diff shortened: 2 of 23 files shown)


python/lsst/daf/butler/script/_associate.py
```diff
@@ -18,13 +18,23 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+
+from typing import Iterable
```
andy-slac: Should it be `collections.abc` instead?

python/lsst/daf/butler/script/_pruneDatasets.py
```diff
@@ -18,14 +18,18 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
-
+from __future__ import annotations
 
 from enum import Enum, auto
+from typing import TYPE_CHECKING, Any, Callable, Iterable
```
andy-slac: Same here, Callable and Iterable should come from `collections.abc`. Check other files too, I won't comment on them.

python/lsst/daf/butler/script/configValidate.py
```diff
@@ -19,11 +19,13 @@
...
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from __future__ import annotations
+
 from .._butler import Butler
 from ..core import ValidationError
 
 
-def configValidate(repo, quiet, dataset_type, ignore):
+def configValidate(repo: str, quiet: bool, dataset_type: str, ignore: str) -> bool:
```
andy-slac: I think `dataset_type` and `ignore` should be `list[str]`, doctring needs some fixing too.

python/lsst/daf/butler/script/queryDatasetTypes.py
```diff
@@ -51,15 +51,14 @@ def queryDatasetTypes(repo, verbose, glob, components):
...
         collection names.
     """
     butler = Butler(repo)
-    if not glob:
-        glob = ...
-    datasetTypes = butler.registry.queryDatasetTypes(components=components, expression=glob)
-    info: List[Any]
+    expression = glob if glob else ...
+    datasetTypes = butler.registry.queryDatasetTypes(components=components, expression=expression)
+    info: list[Any]
```
andy-slac: `info` does not seem to be used?

python/lsst/daf/butler/script/queryDimensionRecords.py
```diff
@@ -19,31 +19,44 @@
...
     butler = Butler(repo)
 
-    records = butler.registry.queryDimensionRecords(
+    query_results = butler.registry.queryDimensionRecords(
         element, datasets=datasets, collections=collections, where=where, check=not no_check
     )
 
     if order_by:
-        records.order_by(*order_by)
+        query_results.order_by(*order_by)
```
andy-slac: Hmm, not related to your changes, but I thought that we are supposed to do `query_results = query_results.order_by(*order_by)`, the fact that we modify iterator itself should be internal detail. Looking at documentation I think it's contradictory: https://github.com/lsst/daf_butler/blob/main/python/lsst/daf/butler/registry/queries/_results.py#L963-L972 
Same for `limit`, @TallJimbo, do you have an opinion on this?

andy-slac (approved):

python/lsst/daf/butler/script/queryDimensionRecords.py
TallJimbo: Definitely definitely return a new object.  DM-31725 is removing the modify-in-place, and making all of these objects immutable.

python/lsst/daf/butler/script/_associate.py
timj: Thanks. I forgot that I'm supposed to be using those directly now rather than `typing.Iterable`.

python/lsst/daf/butler/script/queryDatasetTypes.py
timj: I didn't write any of this code but I can try to take a quick look to see what's going on.

python/lsst/daf/butler/script/queryDimensionRecords.py
timj: Okay. Will fix.

python/lsst/daf/butler/script/configValidate.py
timj: Indeed. The problem being that we declare the API this calls to accept `Iterable[str]` and `str` itself matches that type. Grr.

PR #754, opened 2022-11-14
DM-36984: Fix astropy table test code

timj:
## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/tests/utils.py
```diff
@@ -142,8 +139,16 @@ def assertAstropyTablesEqual(self, tables, expectedTables, filterColumns=False,
                 table.sort(table.colnames)
                 expected = expected.copy()
                 expected.sort(expected.colnames)
-            # Assert that they match:
-            self.assertTrue(report_diff_values(table, expected, fileobj=diff), msg="\n" + diff.getvalue())
+            # Assert that they match.
+            # Recommendation is to format the table into lines for comparison.
+            table1 = table.pformat_all()
+            expected1 = expected.pformat_all()
```
erykoff, line 146: I never would have expected this method, that's for sure!  But I note that it doesn't compare datatypes for the columns.  However, in this case that's fine because they're all string columns and the string lengths aren't the same between `table` and `expected` so they don't actually compare equal.  

In any case, I would suggest adding in a comment that this compares column names and values, but does not compare data types.

erykoff (approved):
Looks good, with the small caveat to add to the code comments.

PR #755, opened 2022-11-16
DM-37025: Enable type checking for click commands

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/cli/butler.py
+++ b/python/lsst/daf/butler/cli/butler.py
@@ -18,7 +18,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
-
+from __future__ import annotations
 
 __all__ = (
     "LoaderCLI",
@@ -33,7 +33,9 @@ import functools
 import logging
 import os
 import traceback
+import types
 from collections import defaultdict
+from typing import Any
 
 import click
 import yaml
@@ -47,7 +49,7 @@ log = logging.getLogger(__name__)
 
 
 @functools.lru_cache
-def _importPlugin(pluginName):
+def _importPlugin(pluginName: str) -> types.ModuleType | type | None | click.Command:
     """Import a plugin that contains Click commands.
 
     Parameters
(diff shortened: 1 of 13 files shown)


python/lsst/daf/butler/cli/progress.py
```diff
@@ -74,7 +74,7 @@ def option(cls, cmd: Any) -> Any:
         )(cmd)
 
     def get_progress_bar(
-        self, iterable: Optional[Iterable[_T]], desc: Optional[str], total: Optional[int], level: int
+        self, iterable: Iterable[_T] | None, desc: str | None, total: int | None, level: int
     ) -> ContextManager[ProgressBar[_T]]:
         # Docstring inherited.
-        return click.progressbar(iterable=iterable, length=total, label=desc, **self._kwargs)
+        return click.progressbar(iterable=iterable, length=total, label=desc, **self._kwargs)  # type: ignore
```
timj, line 80: I wasn't really sure what to do with this one. The error was:

```
Incompatible return value type (got "Union[click._termui_impl.ProgressBar[_T], click._termui_impl.ProgressBar[int]]", expected "AbstractContextManager[lsst.daf.butler.core.progress.ProgressBar[_T]]")  [return-value]
```

python/lsst/daf/butler/cli/butler.py
```diff
@@ -100,7 +102,7 @@ def localCmdPkg(self):
         """
         pass
```
andy-slac: This should probably raise NotImplementedError instead.

python/lsst/daf/butler/cli/butler.py
```diff
@@ -158,9 +161,20 @@ def get_command(self, ctx, name):
         if name not in commands:
             return None
         self._raiseIfDuplicateCommands(commands)
-        return _importPlugin(commands[name][0] + "." + self._cmdNameToFuncName(name))
+        module_str = commands[name][0] + "." + self._cmdNameToFuncName(name)
+        # The click.command decorator returns an instance of a class, which
+        # is something that doImport is not expecting. We add it in as an
+        # option here to appease mypy.
+        plugin = cast(types.ModuleType | type | click.Command | None, _importPlugin(module_str))
```
andy-slac: Should `click.Command` be added to `_importPlugin` return type annotation instead?

python/lsst/daf/butler/cli/progress.py
andy-slac: I'm not sure whether it complains about `click.ProgressBar` not matching `daf.butler.ProgressBar` protocol, or that `click.ProgressBar` does not inherit from `ContextManager`.

andy-slac (approved):

python/lsst/daf/butler/cli/butler.py
timj: Could do, although it will probably trigger other places to ignore because `doImport` is never expecting to return an instance, only types and modules. I could change `doImport` but that probably has huge ramifications to typing elsewhere. I can try seeing what the damage is if I force `_importPlugin` to change what it can return.

python/lsst/daf/butler/cli/progress.py
timj: The docs for `click.progressbar` say it returns `ProgressBar[V]` and then says "This function creates an iterable context manager", so for all I know the type annotation is wrong in `click`.

PR #756, opened 2022-11-17
DM-37036: Fix type annotations in lsst.daf.butler.tests

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/logging.py
+++ b/python/lsst/daf/butler/core/logging.py
@@ -26,7 +26,20 @@ import logging
 import traceback
 from contextlib import contextmanager
 from logging import Formatter, LogRecord, StreamHandler
-from typing import IO, Any, Callable, ClassVar, Dict, Generator, Iterable, Iterator, List, Optional, Union
+from typing import (
+    IO,
+    Any,
+    Callable,
+    ClassVar,
+    Dict,
+    Generator,
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Union,
+    overload,
+)
 
 from lsst.utils.introspection import get_full_type_name
 from lsst.utils.iteration import isplit
@@ -466,6 +479,14 @@ class ButlerLogRecords(BaseModel):
 
     def __setitem__(self, index: int, value: Record) -> None:
         self.__root__[index] = self._validate_record(value)
+
+    @overload
+    def __getitem__(self, index: int) -> ButlerLogRecord:
+        ...
+
+    @overload
+    def __getitem__(self, index: slice) -> "ButlerLogRecords":
+        ...
 
     def __getitem__(self, index: Union[slice, int]) -> "Union[ButlerLogRecords, ButlerLogRecord]":
         # Handles slices and returns a new collection in that
--- a/python/lsst/daf/butler/tests/_datasetsHelper.py
+++ b/python/lsst/daf/butler/tests/_datasetsHelper.py
@@ -19,6 +19,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from __future__ import annotations
+
 __all__ = (
     "DatasetTestHelper",
     "DatastoreTestHelper",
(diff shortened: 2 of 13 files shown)


python/lsst/daf/butler/core/logging.py
```diff
@@ -26,7 +26,20 @@
...
+from typing import (
+    IO,
+    Any,
+    Callable,
+    ClassVar,
+    Dict,
+    Generator,
+    Iterable,
+    Iterator,
+    List,
```
andy-slac, line 38: Some of these should come from `collections.abc`?

python/lsst/daf/butler/tests/_datasetsHelper.py
```diff
@@ -126,21 +167,22 @@ def _writeFile(self, inMemoryDataset):
...
         raise RuntimeError("Did not writing anything at all")
 
 
 class MultiDetectorFormatter(YamlFormatter):
-    def _writeFile(self, inMemoryDataset):
+    def _writeFile(self, inMemoryDataset: Any) -> None:
         raise NotImplementedError("Can not write")
 
-    def _fromBytes(self, serializedDataset, pytype=None):
+    def _fromBytes(self, serializedDataset: Any, pytype: type[Any] | None = None) -> Any:
```
andy-slac: Base class has `serializedDataset: bytes`, should it be the same here?

python/lsst/daf/butler/tests/_examplePythonTypes.py
```diff
@@ -186,30 +193,37 @@ class MetricsExample:
...
+        data: list[Any] | None = None,
+    ) -> None:
         self.summary = summary
         self.output = output
         self.data = data
 
-    def __eq__(self, other):
+    def __eq__(self, other: Any) -> bool:
+        if not (hasattr(other, "summary") and hasattr(other, "output") and hasattr(other, "data")):
+            return NotImplemented
```
andy-slac: May be more efficient to catch `AttributeError` than do explicit check, though I'm sure it does not matter here.

python/lsst/daf/butler/tests/_examplePythonTypes.py
```diff
@@ -237,18 +251,20 @@ def makeFromDict(cls, exportDict):
         newobject : `MetricsExample`
             New `MetricsExample` object.
         """
-        data = None
-        if "data" in exportDict:
-            data = exportDict["data"]
+        data = exportDict["data"] if "data" in exportDict else None
+        assert data is None or isinstance(data, list)
```
andy-slac: These days you should be able to write:

    assert isinstance(data, list | None)

andy-slac (approved):

python/lsst/daf/butler/tests/_examplePythonTypes.py
timj: I had no idea that syntax worked (and thought at least I needed `NoneType` for `isinstance`).

python/lsst/daf/butler/tests/_testRepo.py
```diff
@@ -179,18 +183,22 @@ def _makeRecords(dataIds: Mapping[str, Iterable], universe: DimensionUniverse) -
 
     # Start populating dicts that will become DimensionRecords by providing
     # alternate keys like detector names
-    record_dicts_by_dimension_name = {}
+    record_dicts_by_dimension_name: dict[str, list[dict[str, str | int | bytes]]] = {}
     for name, values in complete_data_id_values.items():
         record_dicts_by_dimension_name[name] = []
-        dimension = universe[name]
+        dimension_el = universe[name]
```
timj, line 189: This change breaks obs_base so I think mypy is getting a bit confused as to what the return type of `universe[name]` is. I need to add a test in daf_butler like the one in obs_base so I can catch this early.

python/lsst/daf/butler/tests/_examplePythonTypes.py
andy-slac: PEP 604 says it's OK to use just None.

TallJimbo: I think there's (unfortunately) a difference between what's allowed in type annotations and what's allowed in runtime code.  I'd be very surprised (but delighted) if `list | None` worked here.

timj:
```python
>>> x = None
>>> isinstance(x, None)
  File "<stdin>", line 1, in <module>
TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
>>> isinstance(x, types.NoneType)
True
>>> isinstance("str", types.NoneType)
False
>>> isinstance(x, str | None)
True
>>> isinstance("x", str | None)
True
>>> isinstance(5, str | None)
False
```

So @andy-slac is correct and I am very surprised indeed. It turns out it only works if you have two types.

timj: A very funny development is that whilst python is absolutely fine with the syntax, mypy 0.991 requires me to use `types.NoneType` (but does still allow the `str | types.NoneType` syntax).

andy-slac: Not sure what you mean by "only works if you have two types." It works ok for `list | str | None`. mypy needs a fix apparently, my understanding that `None` in UnionType is just a special shortcut for `NoneType`.

timj: Sorry. I was unclear. I meant `isinstance(None, None)` does not work but `isinstance(None, None | a | b | ...)` does work (so 2 or more).

TallJimbo: :exploding_head:

python/lsst/daf/butler/tests/_examplePythonTypes.py
timj: My recollection is that mypy doesn't work with try/catch but does work with hasattr.

timj: Maybe mypy  0.991 handles `__eq__` differently because now mypy doesn't complain at all about `other: Any` and lets everything through. I've used try/except now.

python/lsst/daf/butler/core/logging.py
timj: Yes, although I wasn't fixing up old syntax in core files (black reformatted it because I added overload).

PR #760, opened 2022-12-05
DM-37173: Rewrite getURIs to use getManyURIs

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -1537,6 +1537,10 @@ class FileDatastore(GenericBaseDatastore):
         ----------
         refs : iterable of `DatasetRef`
             The datasets to be checked.
+        artifact_existence : `dict` [`lsst.resources.ResourcePath`, `bool`]
+            Optional mapping of datastore artifact to existence. Updated by
+            this method with details of all artifacts tested. Can be `None`
+            if the caller is not interested.
 
         Returns
         -------
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -1529,7 +1529,7 @@ class FileDatastore(GenericBaseDatastore):
         return dataset_existence
 
     def _mexists(
-        self, refs: Iterable[DatasetRef], artifact_existence: Optional[Dict[ResourcePath, bool]] = None
+        self, refs: Sequence[DatasetRef], artifact_existence: Optional[Dict[ResourcePath, bool]] = None
     ) -> Dict[DatasetRef, bool]:
         """Check the existence of multiple datasets at once.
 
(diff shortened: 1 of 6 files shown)


python/lsst/daf/butler/registry/interfaces/_opaque.py
```diff
@@ -48,14 +51,17 @@ def __init__(self, name: str):
         self.name = name
 
     @abstractmethod
-    def insert(self, *data: dict) -> None:
+    def insert(self, *data: dict, transaction: DatastoreTransaction | None = None) -> None:
```
TallJimbo, line 54: I'm not thrilled with this change - it's giving `OpaqueTableStorage` a dependency on `DatastoreTransaction`, while the opposite relationship already exists.  That's something I could live with if it made a real-world implementation work, but I think this is just to make a mock interface work.

I think there's a good chance it's needed to work around existing defects in the Registry/Datastore boundary, so maybe we have to live with it.  But I'd like the docs to make it very clear under what conditions the transaction needs to be passed by callers, and under what conditions an implementation of `OpaqueTableStorage` needs to do something with it if it is given.  I gather it's something about whether `insert` is satisfies strong exception guarantees without it.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1566,30 +1566,63 @@ def _mexists(
...
+        return dataset_existence
 
-                dataset_existence.update(
-                    self._process_mexists_records(
-                        id_to_ref, records, False, artifact_existence=artifact_existence
-                    )
+    def _mexists_check_expected(
+        self, refs: Sequence[DatasetRef], artifact_existence: Optional[Dict[ResourcePath, bool]] = None
+    ) -> Dict[DatasetRef, bool]:
+        """Check existence of refs that are not known to datastore.
```
TallJimbo, line 1580: Add newline.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1809,11 +1809,27 @@ def getManyURIs(
...
-        missing_refs = (ref for ref in refs if ref.id not in records_keys)
+        existing_refs = tuple(ref for ref in refs if ref.id in records_keys)
+        missing_refs = tuple(ref for ref in refs if ref.id not in records_keys)
+
+        # Have to handle trustGetRequest mode by checking for the existence
+        # of the missing refs on disk.
+        if missing_refs:
+            dataset_existence = self._mexists_check_expected(missing_refs, None)
+            really_missing = set()
+            not_missing = set()
```
TallJimbo, line 1811: I'm surprised mypy isn't demanding an annotation for these.  Might be worth double-checking that is' looking at this file and function.  Or maybe it's just gotten better at type inference since the last time I saw a complaint about lines like these.

TallJimbo (approved):
I'm a little uneasy about the `OpaqueTableStorage` interface change, but I'd be content with more complete documentation to make it feel less like a carve-out for a particular testing class.

Otherwise looks good.

python/lsst/daf/butler/registry/interfaces/_opaque.py
timj: In some sense we need a `Transaction` abstract base class for this rather than `DatastoreTransaction`. The problem with a lot of our storage classes is that they are assuming they are not really responsible for transaction handling and that the database will deal with it all. Whenever something turns up that doesn't have a SQL database backing it we run into a problem. Datastore has a transaction object but it has no way to pass that on to registry without having an API for it. Where else should I put the undo if Datastore doesn't know whether the records table is going to sort itself out because of a database transaction or needs help to manually undo?

We had to make the same sort of change in `DatastoreRegistryBridge` in #733 but that was at least a datastore bridge getting a datastore transaction.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: I was under the impression that new mypy attaches the type on first assignment. If I change that to declaring it to be a `set[int]` it really does complain about the later assignment:

```python
python/lsst/daf/butler/datastores/fileDatastore.py:1815: error: Argument 1 to "add" of "set" has incompatible type "DatasetRef"; expected "int"  [arg-type]
python/lsst/daf/butler/datastores/fileDatastore.py:1820: error: Argument 1 to "tuple" has incompatible type "Set[int]"; expected "Iterable[DatasetRef]"  [arg-type]
```

python/lsst/daf/butler/registry/interfaces/_opaque.py
timj: After discussion I added more detailed docs to the insert method explaining the transaction.

PR #761, opened 2022-12-07
DM-37214: Fix problem with JSON formatter when Pydantic is used but Dict requested

timj:
JSON natively reads `dict` and so was never triggering the conversion code to the proper internal type before converting that to a dict. For `TaskMetadata` this could lead to oddities because the dict view of `TaskMetadata` does not match the internal model layout.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/formatters/file.py
+++ b/python/lsst/daf/butler/formatters/file.py
@@ -25,7 +25,6 @@ from __future__ import annotations
 
 __all__ = ("FileFormatter",)
 
-import builtins
 import dataclasses
 from abc import abstractmethod
 from typing import TYPE_CHECKING, Any, Optional, Type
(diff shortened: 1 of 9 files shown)


python/lsst/daf/butler/formatters/file.py
```diff
@@ -104,35 +105,84 @@ def _assembleDataset(self, data: Any, component: Optional[str] = None) -> Any:
...
+        -------
+        inMemoryDataset : `object`
+            Object of expected type ``writeStorageClass.pytype``.
+        """
+        # Only do anything if this is a built-in type and we did not originally
+        # store it as a builtin type.
+        if (
+            inMemoryDataset is not None
+            and not hasattr(builtins, writeStorageClass.pytype.__name__)
+            and hasattr(builtins, type(inMemoryDataset).__name__)
```
TallJimbo: These builtin checks are mostly only for `YamlFormatter`, right, since `JsonFormatter` always returns a `dict` at first?  If so, I wonder if we'd actually be better off having these share _less_ code, because we tend to register converters for YAML but call pydantic constructors for JSON when the `pytype` is something other than `dict`.  Probably worth considering for this ticket, but something to think about.

TallJimbo (approved):
Looks good.  Only comment is probably a case of idle musing about future possibilities.

python/lsst/daf/butler/formatters/file.py
timj: JSON can return a list as well as a dict (or even a float or int on its own). The YamlFormatter can return a normal python type if the converter was not registered (and there is code in the YamlFormatter to try to convert the type to a dict).

It's possible I could change that third part of the `if` statement to be a limited set of `list|dict`. The logic is really "if we didn't start with a simple type but ended up reading one, try to convert it to the type that we started with"

timj: I've expanded the documentation for the method and cleaned up the logic further. The code now runs if:

* The written type does not match the type read from the file.
* The type read from the file looks like a built-in.

In theory that second check could be changed to be an explicit check for `list|dict|int|float|str` (types you can read from a JSON or YAML file without cleverness) but I'm not sure that really gives us anything.

TallJimbo: Could that be changed to:

- the written type does not match the type read from the file;
- `component=None`?

Or do we need the builtin checks to deal with storage class conversions that also have component extraction in play?

timj: Checking that the types don't match is not sufficient because then we can end up being very confused by pre-empting storage class conversions that happen immediately after these checks. These checks are here specifically for the case where you get a generic python type but want a more specialized type. Storage class conversions have to run after you've gone to the more specialized type.

component handling is handled above this method and if a component is in play I update the write storage class to match that of the component.

PR #762, opened 2022-12-08
DM-35063: continue deprecation and removal of integer dataset IDs

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_butler.py
+++ b/tests/test_butler.py
@@ -2083,23 +2083,6 @@ class PosixDatastoreTransfers(unittest.TestCase):
         # Setting id_gen_map should have no effect here
         self.assertButlerTransfers(id_gen_map={"random_data_2": DatasetIdGenEnum.DATAID_TYPE})
 
-    def testTransferIntToInt(self):
-        with self.assertWarns(FutureWarning):
-            self.create_butlers(
-                "lsst.daf.butler.registry.datasets.byDimensions.ByDimensionsDatasetRecordStorageManager",
-                "lsst.daf.butler.registry.datasets.byDimensions.ByDimensionsDatasetRecordStorageManager",
-            )
-        # int dataset ID only allows UNIQUE
-        self.assertButlerTransfers()
-
-    def testTransferIntToUuid(self):
-        with self.assertWarns(FutureWarning):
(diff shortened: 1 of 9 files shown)


tests/test_butler.py
```diff
@@ -2083,23 +2083,6 @@ def testTransferUuidToUuid(self):
...
-    def testTransferIntToInt(self):
-        with self.assertWarns(FutureWarning):
-            self.create_butlers(
-                "lsst.daf.butler.registry.datasets.byDimensions.ByDimensionsDatasetRecordStorageManager",
-                "lsst.daf.butler.registry.datasets.byDimensions.ByDimensionsDatasetRecordStorageManager",
-            )
-        # int dataset ID only allows UNIQUE
-        self.assertButlerTransfers()
-
-    def testTransferIntToUuid(self):
```
timj, line 2095: If we are allowing migration for the moment, shouldn't we still be testing int to UUID? Presumably the problem is that we can't any longer create a test repo from scratch? The only alternative is to have a pre-created repo for the test but it's likely not worth it.

tests/test_simpleButler.py
```diff
@@ -49,9 +49,11 @@ class SimpleButlerTestCase(unittest.TestCase):
     can instead utilize an in-memory SQLite Registry and a mocked Datastore.
     """
 
-    datasetsManager = "lsst.daf.butler.registry.datasets.byDimensions.ByDimensionsDatasetRecordStorageManager"
-    datasetsImportFile = "datasets.yaml"
```
timj, line 53: Does this mean that this file is no longer used?

timj (approved):
Looks okay.

tests/test_butler.py
TallJimbo: You've replicated my train of thought on this exactly.

tests/test_simpleButler.py
TallJimbo: I've left in place the test that imports this file (with integer IDs) into a UUID data repository.  I don't know if we want to support that indefinitely, but I haven't tried to deprecate it (I didn't think about it at all when writing the RFC).

PR #763, opened 2022-12-09
DM-37249: make connection usage compatible with pgbouncer transaction-level pooling

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/databases/postgresql.py
+++ b/python/lsst/daf/butler/registry/databases/postgresql.py
@@ -123,6 +123,7 @@ class PostgresqlDatabase(Database):
             if not self.isWriteable():
                 with closing(self._session_connection.connection.cursor()) as cursor:
                     cursor.execute("SET TRANSACTION READ ONLY")
+                    cursor.execute("SET TIME ZONE 0")
             else:
                 with closing(self._session_connection.connection.cursor()) as cursor:
                     # Make timestamps UTC, because we didn't use TIMESTAMPZ for
--- a/python/lsst/daf/butler/registry/databases/postgresql.py
+++ b/python/lsst/daf/butler/registry/databases/postgresql.py
@@ -160,6 +160,15 @@ class PostgresqlDatabase(Database):
                         cursor.execute("SET TIME ZONE 0")
             yield is_new
 
+    @contextmanager
+    def temporary_table(
+        self, spec: ddl.TableSpec, name: Optional[str] = None
+    ) -> Iterator[sqlalchemy.schema.Table]:
+        # Docstring inherited.
+        with self.transaction(for_temp_tables=True):
+            with super().temporary_table(spec, name) as table:
+                yield table
+
     def _lockTables(
         self, connection: sqlalchemy.engine.Connection, tables: Iterable[sqlalchemy.schema.Table] = ()
     ) -> None:
--- a/python/lsst/daf/butler/registry/databases/postgresql.py
+++ b/python/lsst/daf/butler/registry/databases/postgresql.py
@@ -97,7 +97,7 @@ class PostgresqlDatabase(Database):
 
     @classmethod
     def makeEngine(cls, uri: str, *, writeable: bool = True) -> sqlalchemy.engine.Engine:
-        return sqlalchemy.engine.create_engine(uri)
+        return sqlalchemy.engine.create_engine(uri, pool_size=1)
 
     @classmethod
     def fromEngine(
--- a/python/lsst/daf/butler/registry/databases/postgresql.py
+++ b/python/lsst/daf/butler/registry/databases/postgresql.py
@@ -117,41 +117,48 @@ class PostgresqlDatabase(Database):
         interrupting: bool = False,
         savepoint: bool = False,
         lock: Iterable[sqlalchemy.schema.Table] = (),
-    ) -> Iterator[None]:
-        with super().transaction(interrupting=interrupting, savepoint=savepoint, lock=lock):
-            assert self._session_connection is not None, "Guaranteed to have a connection in transaction"
-            # pgbouncer with transaction-level pooling (which we aim to
-            # support) says that SET cannot be used, except for a list of
-            # "Startup parameters" that includes "timezone" (see
-            # https://www.pgbouncer.org/features.html#fnref:0).  But I don't
-            # see "timezone" in PostgreSQL's list of parameters passed when
-            # creating a new connection
-            # (https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS).
-            # Given that the pgbouncer docs say, "PgBouncer detects their
-            # changes and so it can guarantee they remain consistent for the
-            # client", I assume we can use "SET TIMESPAN" and pgbouncer will
-            # take care of clients that share connections being set
-            # consistently.  And if that assumption is wrong, we should still
-            # probably be okay, since all clients should be Butler clients, and
-            # they'll all be setting the same thing.
-            #
-            # The "SET TRANSACTION READ ONLY" should also be safe, because it
-            # only ever acts on the current transaction; I think it's not
-            # included in pgbouncer's declaration that SET is incompatible with
-            # transaction-level pooling because PostgreSQL actually considers
-            # SET TRANSACTION to be a fundamentally different statement from
-            # SET (they have their own distinct doc pages, at least).
-            if not self.isWriteable():
-                with closing(self._session_connection.connection.cursor()) as cursor:
-                    cursor.execute("SET TRANSACTION READ ONLY")
-                    cursor.execute("SET TIME ZONE 0")
-            else:
-                with closing(self._session_connection.connection.cursor()) as cursor:
-                    # Make timestamps UTC, because we didn't use TIMESTAMPZ for
-                    # the column type.  When we can tolerate a schema change,
-                    # we should change that type and remove this line.
-                    cursor.execute("SET TIME ZONE 0")
-            yield
+        for_temp_tables: bool = False,
+    ) -> Iterator[bool]:
+        with super().transaction(interrupting=interrupting, savepoint=savepoint, lock=lock) as is_new:
+            if is_new:
+                assert self._session_connection is not None, "Guaranteed to have a connection in transaction"
+                # pgbouncer with transaction-level pooling (which we aim to
+                # support) says that SET cannot be used, except for a list of
+                # "Startup parameters" that includes "timezone" (see
+                # https://www.pgbouncer.org/features.html#fnref:0).  But I
+                # don't see "timezone" in PostgreSQL's list of parameters
+                # passed when creating a new connection
+                # (https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS).
+                # Given that the pgbouncer docs say, "PgBouncer detects their
+                # changes and so it can guarantee they remain consistent for
+                # the client", I assume we can use "SET TIMESPAN" and pgbouncer
+                # will take care of clients that share connections being set
+                # consistently.  And if that assumption is wrong, we should
+                # still probably be okay, since all clients should be Butler
+                # clients, and they'll all be setting the same thing.
+                #
+                # The "SET TRANSACTION READ ONLY" should also be safe, because
+                # it only ever acts on the current transaction; I think it's
+                # not included in pgbouncer's declaration that SET is
+                # incompatible with transaction-level pooling because
+                # PostgreSQL actually considers SET TRANSACTION to be a
+                # fundamentally different statement from SET (they have their
+                # own distinct doc pages, at least).
+                if not (self.isWriteable() or for_temp_tables):
+                    # PostgreSQL permits writing to temporary tables inside
+                    # read-only transactions, but it doesn't permit creating
+                    # them.
+                    with closing(self._session_connection.connection.cursor()) as cursor:
+                        cursor.execute("SET TRANSACTION READ ONLY")
+                        cursor.execute("SET TIME ZONE 0")
+                else:
+                    with closing(self._session_connection.connection.cursor()) as cursor:
+                        # Make timestamps UTC, because we didn't use TIMESTAMPZ
+                        # for the column type.  When we can tolerate a schema
+                        # change, we should change that type and remove this
+                        # line.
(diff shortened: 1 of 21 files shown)


python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
```diff
@@ -767,25 +772,19 @@ def import_(
...
-                dict(protoTagsRow, dataset_id=dataset_id, **dataId.byName())
-                for dataset_id, dataId in dataIds.items()
-            ]
-
-            with self._db.transaction():
+        # We'll insert all new rows into a temporary table
+        tableSpec = makeTagTableSpec(self.datasetType, type(self._collections), ddl.GUID, constraints=False)
+        with self._db.transaction(for_temp_tables=True):
+            with self._db.temporary_table(tableSpec) as tmp_tags:
+                collFkName = self._collections.getCollectionForeignKeyName()
```
ktlim: Can't `collFkName`, `protoTagsRow`, and `tmpRows` be hoisted outside the transaction?

python/lsst/daf/butler/registry/dimensions/table.py
```diff
@@ -185,11 +185,12 @@ def fetch(self, dataIds: DataCoordinateIterable) -> Iterable[DimensionRecord]:
...
-                values = row._asdict()
-                if self.element.temporal is not None:
-                    values[TimespanDatabaseRepresentation.NAME] = TimespanReprClass.extract(values)
-                yield RecordClass(**values)
+            with self._db.query(query.combine()) as sql_result:
+                for row in sql_result:
+                    values = row._asdict()
+                    if self.element.temporal is not None:
+                        values[TimespanDatabaseRepresentation.NAME] = TimespanReprClass.extract(values)
+                    yield RecordClass(**values)
```
ktlim, line 193: This one is still yielding within a query, unlike most others where you now capture all the results first.  Was that intended?

python/lsst/daf/butler/registry/databases/postgresql.py
```diff
@@ -117,12 +117,37 @@ def transaction(
...
+            # SET (they have their own distinct doc pages, at least).
+            if not (self.isWriteable() or for_temp_tables):
+                # PostgreSQL permits writing to temporary tables inside
+                # read-only transactions, but it doesn't permit creating them.
                 with closing(self._session_connection.connection.cursor()) as cursor:
                     cursor.execute("SET TRANSACTION READ ONLY")
+                    cursor.execute("SET TIME ZONE 0")
             else:
                 with closing(self._session_connection.connection.cursor()) as cursor:
                     # Make timestamps UTC, because we didn't use TIMESTAMPZ for
```
ktlim: Should we have included this with the longer-observation-reason schema change?  Or (because it possibly requires modifying column values) would that have been riskier?

python/lsst/daf/butler/registry/databases/sqlite.py
```diff
@@ -190,6 +180,16 @@ def creator() -> sqlite3.Connection:
 
         engine = sqlalchemy.engine.create_engine(uri, creator=creator)
 
+        def _onSqlite3Begin(connection: sqlalchemy.engine.Connection) -> sqlalchemy.engine.Connection:
+            assert connection.dialect.name == "sqlite"
+            # Replace pysqlite's buggy transaction handling that never BEGINs
+            # with our own that does, and tell SQLite to try to acquire a lock
+            # as soon as we start a transaction (this should lead to more
+            # blocking and fewer deadlocks).
+            if writeable:
```
ktlim, line 191: Can't there still be transactions on read-only databases to ensure consistency between multiple SELECTs?  Or is this not a problem?

python/lsst/daf/butler/registry/databases/postgresql.py
```diff
@@ -123,6 +123,7 @@ def transaction(
             if not self.isWriteable():
                 with closing(self._session_connection.connection.cursor()) as cursor:
                     cursor.execute("SET TRANSACTION READ ONLY")
+                    cursor.execute("SET TIME ZONE 0")
```
andy-slac: At USDF default server timezone is `UTC` so this probably does nothing. I do not remember why do we need to set time zone explicitly?

python/lsst/daf/butler/registry/databases/postgresql.py
```diff
@@ -120,6 +120,27 @@ def transaction(
     ) -> Iterator[None]:
         with super().transaction(interrupting=interrupting, savepoint=savepoint, lock=lock):
             assert self._session_connection is not None, "Guaranteed to have a connection in transaction"
+            # pgbouncer with transaction-level pooling (which we aim to
+            # support) says that SET cannot be used, except for a list of
+            # "Startup parameters" that includes "timespan" (see
```
andy-slac: timespan -> timezone?

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -533,26 +533,19 @@ def transaction(
...
-            # the connection object's 'info' dict, which is explicitly for user
-            # information like this.  This is safer than a regular `Database`
-            # instance attribute, because it guards against multiple `Database`
-            # instances sharing the same connection.  The need to use our own
-            # flag here to track whether we're in a nested transaction should
-            # go away in SQLAlchemy 1.4, which seems to have a
-            # `Connection.in_nested_transaction()` method.
-            savepoint = savepoint or connection.info.get(_IN_SAVEPOINT_TRANSACTION, False)
+            savepoint = savepoint or connection.in_nested_transaction()
             connection.info[_IN_SAVEPOINT_TRANSACTION] = savepoint
```
andy-slac: Why do we need to keep this line (and `_IN_SAVEPOINT_TRANSACTION`)?

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -534,7 +407,6 @@ def transaction(
                 "interrupt the active transaction context has been requested."
             )
             savepoint = savepoint or connection.in_nested_transaction()
-            connection.info[_IN_SAVEPOINT_TRANSACTION] = savepoint
```
andy-slac, line 545: OK, I see this dropped now, but `_IN_SAVEPOINT_TRANSACTION` is no longer needed too.

python/lsst/daf/butler/registry/databases/postgresql.py
```diff
@@ -194,6 +197,22 @@ def _convertExclusionConstraintSpec(
...
+        self,
+        connection: sqlalchemy.engine.Connection,
+        spec: ddl.TableSpec,
+        name: Optional[str] = None,
+        **kwargs: Any,
+    ) -> sqlalchemy.schema.Table:
+        # Docstring inherited
+        # Adding ON COMMIT DROP here is really quite defensive: we already
+        # manually drop the table at the end of the temporary_table context
+        # manager, and that will usually happen first.  But this will guarantee
```
andy-slac, line 225: Hmm, I'm not sure where do we drop temporary tables, I don't think that new `temporary_table` does that?

andy-slac (approved):
Looks good, small bunch of questions/comments.

ktlim (commented):
Haven't read `interfaces/_database.py`, which is arguably the most important part, but some comments on the rest.

python/lsst/daf/butler/registry/databases/postgresql.py
TallJimbo: I'm afraid I don't recall either, but I also don't want to assume all PostgreSQL instances we work with are configured like that.

python/lsst/daf/butler/registry/databases/postgresql.py
TallJimbo: I think we probably could have included it.  But I'd defer to @andy-slac on that.  The real reason it wasn't included was because this had fallen off of everyone's radar completely (it predates even the big suite of schema changes we made to address visit-exposure relationships a while back).  I've created [DM-37322](https://jira.lsstcorp.org/browse/DM-37322) to try to keep that from happening again.

python/lsst/daf/butler/registry/databases/sqlite.py
TallJimbo: It's not something we've ever needed, at least not outside the context of doing a write as well in the same transaction.

python/lsst/daf/butler/registry/dimensions/table.py
TallJimbo: Good catch; I think I meant to add a `.fetchall()` to the `for` loop target.

python/lsst/daf/butler/registry/databases/postgresql.py
TallJimbo: Oops, it did originally and then I think I lost that line in some later refactoring.  Fixed.

python/lsst/daf/butler/registry/interfaces/_database.py
TallJimbo: I think I amended the wrong commit here; will fix when I rebase and squash after review is complete.

TallJimbo (commented):
Just pushed two new commits and a number of fixups to address review comments.  The new changes were spurred my one-off testing revealing that we're both using more connections than we need per `Butler` instance and interacting with the SQLAlchemy pool an awful lot during Butler startup (resulting in those ubiquitous rollbacks in the logs).

python/lsst/daf/butler/registry/databases/postgresql.py
TallJimbo: Actually, nevermind, for now - if I attempt the `drop` in the `finally`, it's apparently easy to trigger an error in which we end up trying a DROP TABLE in a read-only transaction. Will investigate tomorrow.

TallJimbo:
> Just pushed two new commits ...

... and already reverted one of them, because apparently it can cause deadlocks in the tests when those are run in parallel (I suspect due to the artificial way we create tables in PostgreSQL for those tests, but it's still a problem).  I'll look into a better fix tomorrow.

python/lsst/daf/butler/registry/databases/postgresql.py
TallJimbo: Fixed, in a new fixup commit that reworks the `transaction` method inheritance a bit; I was previously running the `SET TRANSACTION READ ONLY` bit within do-nothing `transaction` context manager calls, not just the outermost "real" transaction.

TallJimbo:
> >  Just pushed two new commits ...

> ... and already reverted one of them, because apparently it can cause deadlocks in the tests when those are run in parallel (I suspect due to the artificial way we create tables in PostgreSQL for those tests, but it's still a problem). I'll look into a better fix tomorrow.

Revert has now been replaced with a fixup commit that solves the test problem by switching from a transaction to just a connection session around table inspection.

python/lsst/daf/butler/registry/databases/sqlite.py
TallJimbo: Last night I discovered that we actually do have a use case for read-only transactions for consistency: at butler startup we fetch a lot of information about dataset types and collections, and it's possible for those queries to be made inconsistent by a concurrent writer (and on `main` we don't have a transaction around that at all).  I discovered this while running ci_cpp a lot to try to reproduce an unrelated race condition; I think it's unusual in that it tends to initialize `Butler` instances (for running `PipelineTasks`) while also registering dataset types and collections (when starting to run `PipelineTasks` from a different pipeline), all with SQLite.

So I've pushed a modified version of this commit that just emits `BEGIN` for read-only databases, rather than `BEGIN IMMEDIATE`; that should start a read-only transaction when followed (as it will be) by a `SELECT` query.  The overall behavior should then be:
 - Multiple transactions in read-only databases can occur at the same time, as they only acquire a shared lock on the database file.
 - Any read-write transactions will block until any concurrent read-only or read-write transactions complete.
 - An ongoing read-write transaction will block any read-only transactions from starting.

And I've also wrapped those butler startup fetches in a transaction context.

TallJimbo:
@andy-slac , could you take a look at the commits I've added since your last review (i.e. the ones after "Add changelog entry.")?

python/lsst/daf/butler/registry/databases/postgresql.py
```diff
@@ -97,7 +97,7 @@ def __init__(
 
     @classmethod
     def makeEngine(cls, uri: str, *, writeable: bool = True) -> sqlalchemy.engine.Engine:
-        return sqlalchemy.engine.create_engine(uri)
+        return sqlalchemy.engine.create_engine(uri, pool_size=1)
```
andy-slac, line 100: Would be interesting to detect situation when there is actually more than one connection (in case we break something in the future) on some future ticket. Maybe Pool events could be used for this.

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -468,29 +455,126 @@ def temporary_table(
...
+        Returns
+        -------
+        context : `AbstractContextManager` [ `tuple` [ `bool`,
+                `sqlalchemy.engine.Connection` ] ]
+            A context manager that commits the transaction when it is exited
+            without error and rolls back the transactoin when it is exited via
+            an exception.  When entered, it returns a tuple of:
+
+            - `is_new` (`bool`) whether this is a new (outermost) transaction;
+            - `connection` (`sqlalchemy.engine.Connection`) the connection.
```
andy-slac: Double backticks for `is_new` and `connection`?

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -468,29 +455,126 @@ def temporary_table(
...
+        """
+        with self._session() as connection:
+            already_in_transaction = connection.in_transaction()
+            assert not (interrupting and already_in_transaction), (
+                "Logic error in transaction nesting: an operation that would "
+                "interrupt the active transaction context has been requested."
+            )
+            savepoint = savepoint or connection.in_nested_transaction()
+            trans: sqlalchemy.engine.Transaction
```
andy-slac: Does this need `| None`?

andy-slac (approved):

python/lsst/daf/butler/registry/interfaces/_database.py
TallJimbo: Yup, good catch.  I'm sure MyPy is not complaining only because it sees all the SQLAlchemy types as `Any` (for now).

PR #765, opened 2022-12-15
DM-37245: Add support for array columns and multi-dimensional columns.

erykoff:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/formatters/parquet.py
+++ b/python/lsst/daf/butler/formatters/parquet.py
@@ -43,7 +43,7 @@ import collections.abc
 import itertools
 import json
 import re
-from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Optional, Sequence, Union
+from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Optional, Sequence, Union, cast
 
 import pyarrow as pa
 import pyarrow.parquet as pq
@@ -343,6 +343,7 @@ def numpy_dict_to_arrow(numpy_dict: Dict[str, np.ndarray]) -> pa.Table:
 
     arrays = []
     for name, col in numpy_dict.items():
+        val: Any
         if len(dtype[name].shape) > 0:
             if rowcount > 0:
                 val = np.split(col.ravel(), rowcount)
@@ -999,7 +1000,7 @@ def _schema_to_dtype_list(schema: pa.Schema) -> list[tuple[str, tuple[Any] | str
     """
     metadata = schema.metadata if schema.metadata is not None else {}
 
-    dtype = []
+    dtype: list[Any] = []
     for name in schema.names:
         t = schema.field(name).type
         if isinstance(t, pa.FixedSizeListType):
@@ -1027,14 +1028,20 @@ def _numpy_dtype_to_arrow_types(dtype: np.dtype) -> list[Any]:
     """
     from math import prod
 
+    import numpy as np
+
     type_list: list[Any] = []
     if dtype.names is None:
         return type_list
 
     for name in dtype.names:
         dt = dtype[name]
+        arrow_type: Any
         if len(dt.shape) > 0:
-            arrow_type = pa.list_(pa.from_numpy_dtype(dt.subdtype[0].type), prod(dt.shape))
+            arrow_type = pa.list_(
+                pa.from_numpy_dtype(cast(tuple[np.dtype, tuple[int, ...]], dt.subdtype)[0].type),
+                prod(dt.shape),
+            )
         else:
             arrow_type = pa.from_numpy_dtype(dt.type)
         type_list.append((name, arrow_type))
--- a/python/lsst/daf/butler/formatters/parquet.py
+++ b/python/lsst/daf/butler/formatters/parquet.py
@@ -334,9 +334,10 @@ def numpy_dict_to_arrow(numpy_dict: Dict[str, np.ndarray]) -> pa.Table:
     md = {}
     md[b"lsst::arrow::rowcount"] = str(rowcount)
 
-    for name in dtype.names:
-        _append_numpy_string_metadata(md, name, dtype[name])
-        _append_numpy_multidim_metadata(md, name, dtype[name])
+    if dtype.names is not None:
+        for name in dtype.names:
+            _append_numpy_string_metadata(md, name, dtype[name])
+            _append_numpy_multidim_metadata(md, name, dtype[name])
 
     schema = pa.schema(type_list, metadata=md)
 
@@ -1026,13 +1027,18 @@ def _numpy_dtype_to_arrow_types(dtype: np.dtype) -> list[Any]:
     """
     from math import prod
 
-    type_list = []
+    type_list: list[Any] = []
+    if dtype.names is None:
+        return type_list
+
     for name in dtype.names:
         dt = dtype[name]
         if len(dt.shape) > 0:
-            type_list.append((name, pa.list_(pa.from_numpy_dtype(dt.subdtype[0].type), prod(dt.shape))))
+            arrow_type = pa.list_(pa.from_numpy_dtype(dt.subdtype[0].type), prod(dt.shape))
         else:
-            type_list.append((name, pa.from_numpy_dtype(dt.type)))
+            arrow_type = pa.from_numpy_dtype(dt.type)
+        type_list.append((name, arrow_type))
+
     return type_list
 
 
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -238,15 +241,28 @@ def arrow_to_numpy_dict(arrow_table: pa.Table) -> Dict[str, np.ndarray]:
     numpy_dict : `dict` [`str`, `numpy.ndarray`]
         Dict with keys as the column names, values as the arrays.
     """
+    import numpy as np
```
TallJimbo, line 248: Should be safe to do this at module scope.

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -238,15 +241,28 @@ def arrow_to_numpy_dict(arrow_table: pa.Table) -> Dict[str, np.ndarray]:
...
             col = col.astype(_arrow_string_to_numpy_dtype(schema, name, col))
+        elif isinstance(t, pa.FixedSizeListType):
+            if len(col) > 0:
+                col = np.stack(col)
+            else:
+                # this is an empty column, and needs to be coerced to type.
+                col = col.astype(t.value_type.to_pandas_dtype())
+
+            shape = _multidim_shape_from_metadata(metadata, t.list_size, name)
+            col = col.reshape((len(arrow_table), *shape))
```
TallJimbo, line 269: If you think it should be possible to always do this without making a copy, you might want to assign to `col.shape` instead, as per https://numpy.org/doc/stable/reference/generated/numpy.reshape.html (I didn't know that was how to express that intent until just now).

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -264,17 +280,31 @@ def numpy_to_arrow(np_array: np.ndarray) -> pa.Table:
...
 
     schema = pa.schema(type_list, metadata=md)
 
-    arrays = [pa.array(np_array[col]) for col in np_array.dtype.names]
+    arrays = []
+    for name in np_array.dtype.names:
+        dt = np_array.dtype[name]
+        if len(dt.shape) > 0:
+            if len(np_array) > 0:
+                val = np.split(np_array[name].ravel(), len(np_array))
```
TallJimbo: FWIW, I thought about whether you really want `ravel` here vs. something else that might copy less but doesn't guarantee row-major order, like `reshape(-1)` - but I think you do want `ravel`, since row-major order is necessary to have the `reshape` on the other end round-trip.

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -291,18 +321,36 @@ def numpy_dict_to_arrow(numpy_dict: Dict[str, np.ndarray]) -> pa.Table:
     Returns
     -------
     arrow_table : `pyarrow.Table`
+
+    Raises
+    ------
+    ValueError if columns in numpy_dict have unequal numbers of rows.
     """
-    type_list = [(name, pa.from_numpy_dtype(col.dtype.type)) for name, col in numpy_dict.items()]
+    import numpy as np
```
TallJimbo: Another import that looks like it could go at module scope.

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -319,25 +367,36 @@ def astropy_to_arrow(astropy_table: atable.Table) -> pa.Table:
...
+    for name in astropy_table.dtype.names:
+        dt = astropy_table.dtype[name]
+        if len(dt.shape) > 0:
+            if len(astropy_table) > 0:
+                val = np.split(astropy_table[name].ravel(), len(astropy_table))
+            else:
+                val = []
+        else:
+            val = astropy_table[name]
+        arrays.append(pa.array(val, type=schema.field(name).type))
```
TallJimbo: Lots of blocks very similar to this one above.  If you haven't tried to factor it out into a free function, please do, but I can also believe that it's the kind of thing where scattered tiny differences would torpedo that.

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -883,3 +929,145 @@ def _append_numpy_string_metadata(metadata: Dict[bytes, str], name: str, dtype:
         metadata[f"lsst::arrow::len::{name}".encode("UTF-8")] = str(dtype.itemsize // 4)
     elif dtype.type is np.bytes_:
         metadata[f"lsst::arrow::len::{name}".encode("UTF-8")] = str(dtype.itemsize)
+
+
+def _append_numpy_multidim_metadata(metadata: Dict[bytes, str], name: str, dtype: np.dtype) -> None:
```
TallJimbo: Can write the `Dict` in this type annotation as `dict` and drop the import from `typing` (this new in Python 3.10, which is why you might have seen the old form a lot more).

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -883,3 +929,145 @@ def _append_numpy_string_metadata(metadata: Dict[bytes, str], name: str, dtype:
...
+    ------
+    RuntimeError
+        Raised if metadata is found but has incorrect format.
+    """
+    md_name = f"lsst::arrow::shape::{name}"
+    if (encoded := md_name.encode("UTF-8")) in metadata:
+        groups = re.search(r"\((.*)\)", metadata[encoded].decode("UTF-8"))
+        if groups is None:
+            raise RuntimeError("Illegal value found in metadata.")
+        shape = tuple(int(x) for x in groups[1].split(",") if x != "")
```
TallJimbo, line 997: Using a regex instead of `startswith("(")` and `endswith(")")` but not instead of `split(",")` probably puts you in a sparsely-populated region of regex love-hate space, but I won't argue with it (especially since I'm not sure how much you have to worry about whitespace or something outside those parens).  If you'd like a regex for the whole thing I'm sure I could cook one up.

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -883,3 +929,145 @@ def _append_numpy_string_metadata(metadata: Dict[bytes, str], name: str, dtype:
...
+    """Convert a numpy dtype to a list of arrow types.
+
+    Parameters
+    ----------
+    dtype : `numpy.dtype`
+        Numpy dtype to convert.
+
+    Returns
+    -------
+    type_list : `list` [`object`]
```
TallJimbo, line 1043: I think Sphinx really wants a description after this, even if it's redundant with the initial line of the docstring.

TallJimbo (approved):
Looks good; only comments are very minor ones.

python/lsst/daf/butler/formatters/parquet.py
timj: Formatters are imported on-demand so in general it doesn't matter where the imports go. I think deferred imports are fine if there is optional formatter functionality. Butler doesn't itself explicitly depend on numpy although it does depend on astropy which depends on numpy.

erykoff: I originally put the numpy imports not at module scope because of Tim's suggestions.  I think I want to leave them as-is.  I hope that's okay Jim?

python/lsst/daf/butler/formatters/parquet.py
erykoff: The numpy docs https://numpy.org/doc/stable/reference/generated/numpy.ndarray.shape.html#numpy.ndarray.shape say "Setting `arr.shape` is discouraged and may be deprecated in the future.  Using `ndarray.reshape` is the preferred approach".  So I'm going to leave this as-is.

python/lsst/daf/butler/formatters/parquet.py
erykoff: I believe that `ravel` and `reshape` are the no-copy, maintain-order operations.  But it's true that I don't have tests for anything but the default ordering.  I will add a multi-dimensional column with fortran ordering to the tests to see what happens.

python/lsst/daf/butler/formatters/parquet.py
erykoff: (Same as above)

python/lsst/daf/butler/formatters/parquet.py
erykoff: This was a bit annoying, but I think ultimately fruitful.  I even figured out how to appease mypy while doing it!

python/lsst/daf/butler/formatters/parquet.py
erykoff: I think I'll leave it as is, but interestingly this is essentially the same solution that chatgpt recommended when I just asked.  So it must be common online?

erykoff: To be clear, I don't use chatgpt for programming advice, I was just curious what it would say.

PR #769, opened 2023-01-12
DM-30439: Refactor DataCoordinate

timj:
## Checklist

- [ ] ran Jenkins


timj:
I know this needs work to deal with conflicts, created as a draft PR to make it easier to examine the changes.

TallJimbo:
:+1:  Note that I definitely want to revisit some of the code here before merging it - my thinking on the custom containers has evolved quite a bit since I wrote those here.  [RFC-834](https://jira.lsstcorp.org/browse/RFC-834) is in play for the `DataCoordinate` containers, and I think storage class conversions sort of blow up the whole concept of name-keyed `DatasetRef` containers, since we cannot assume that dataset type name equality means dataset type equality in many contexts anymore.

timj:
Rebasing to w.2021.46 was relatively easy (although I'm sure tests don't pass). The first major problem comes with #601 since that is when @andy-slac reorganizes how dimension record query results are returned such that order_by/limit can be supported and that provides a completely different return type with different parameters to the new, on this ticket, `HomogeneousDimensionRecordIterable`.

Not sure where to take this now.

PR #774, opened 2023-01-18
DM-37609: Resurrect prototype client/server and add tests

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/repoRelocation.py
+++ b/python/lsst/daf/butler/core/repoRelocation.py
@@ -64,8 +64,12 @@ def replaceRoot(configRoot: str, butlerRoot: Optional[ResourcePathExpression]) -
     ValueError
         Raised if ``butlerRoot`` is not set but a value is required.
     """
-    # Do nothing if there is nothing to be done
-    if BUTLER_ROOT_TAG not in configRoot:
+    # Do nothing if there is nothing to be done or if there is a type
+    # mismatch.
+    try:
+        if BUTLER_ROOT_TAG not in configRoot:
+            return configRoot
+    except TypeError:
         return configRoot
 
     # None or empty string indicate a problem
--- a/python/lsst/daf/butler/registries/remote.py
+++ b/python/lsst/daf/butler/registries/remote.py
@@ -218,7 +218,12 @@ class RemoteRegistry(Registry):
 
         # Need to determine what to refresh.
         # Might need a server method to return all the DatasetTypes up front.
-        pass
+        # How do we know which server should be refreshed?
+        # Should there be caches in the client?
+        response = self._client.put(self._get_url("v1/registry/refresh"))
+        response.raise_for_status()
+
+        return
 
     @contextlib.contextmanager
     def transaction(self, *, savepoint: bool = False) -> Iterator[None]:
@@ -458,6 +463,7 @@ class RemoteRegistry(Registry):
         missing: list[str] | None = None,
     ) -> Iterable[DatasetType]:
         # Docstring inherited from lsst.daf.butler.registry.Registry
+        # Note no caching implemented in client.
         if missing is not None:
             raise NotImplementedError("RemoteRegistry does not support the 'missing' parameter.")
 
--- a/python/lsst/daf/butler/registries/remote.py
+++ b/python/lsst/daf/butler/registries/remote.py
@@ -412,7 +412,10 @@ class RemoteRegistry(Registry):
 
     def getDatasetLocations(self, ref: DatasetRef) -> Iterable[str]:
         # Docstring inherited from lsst.daf.butler.registry.Registry
-        raise NotImplementedError()
+        path = f"v1/registry/datasetLocations/{ref.id}"
+        response = self._client.get(self._get_url(path))
+        response.raise_for_status()
+        return response.json()
 
     def expandDataId(
         self,
--- a/python/lsst/daf/butler/registries/remote.py
+++ b/python/lsst/daf/butler/registries/remote.py
@@ -119,10 +119,21 @@ class RemoteRegistry(Registry):
         if defaults is None:
             defaults = RegistryDefaults()
 
-        server_uri = ResourcePath(config["db"])
-        return cls(server_uri, defaults, writeable)
+        if isinstance(config["db"], httpx.Client):
+            client = config["db"]
+            server_uri = ResourcePath("/")
+        else:
+            client = None
+            server_uri = ResourcePath(config["db"])
+        return cls(server_uri, defaults, writeable, client=client)
 
-    def __init__(self, server_uri: ResourcePath, defaults: RegistryDefaults, writeable: bool):
+    def __init__(
+        self,
+        server_uri: ResourcePath,
+        defaults: RegistryDefaults,
+        writeable: bool,
+        client: httpx.Client | None = None,
+    ):
         self._db = server_uri
         self._defaults = defaults
 
@@ -135,8 +146,13 @@ class RemoteRegistry(Registry):
 
         self._dimensions: DimensionUniverse | None = None
 
-        headers = {"user-agent": f"{get_full_type_name(self)}/{__version__}"}
-        self._client = httpx.Client(headers=headers)
+        if client is not None:
+            # We have injected a client explicitly in to the class.
+            # This is generally done for testing.
+            self._client = client
+        else:
+            headers = {"user-agent": f"{get_full_type_name(self)}/{__version__}"}
+            self._client = httpx.Client(headers=headers)
 
         # Does each API need to be sent the defaults so that the server
         # can use specific defaults each time?
@@ -163,6 +179,26 @@ class RemoteRegistry(Registry):
             defaults = self.defaults
         return type(self)(self._db, defaults, self.isWriteable())
 
+    def _get_url(self, path: str) -> str:
+        """Form the full URL to the server given the path on server.
+
+        Parameters
+        ----------
+        path : `str`
+            The path to the server endpoint. Should not include the "/butler"
+            prefix.
+
+        Returns
+        -------
+        url : `str`
+            The full URL to the service.
+        """
+        prefix = "butler"
+        if self._db.scheme == "file":
+            # Not a server, assume a test server and so prepend a /.
+            return f"/{prefix}/{path}"
+        return str(self._db.join(prefix).join(path))
+
     @property
     def dimensions(self) -> DimensionUniverse:
         # Docstring inherited from lsst.daf.butler.registry.Registry
@@ -170,7 +206,7 @@ class RemoteRegistry(Registry):
             return self._dimensions
 
         # Access /dimensions.json on server and cache it locally.
-        response = self._client.get(str(self._db.join("universe")))
+        response = self._client.get(self._get_url("universe"))
         response.raise_for_status()
 
         config = DimensionConfig.fromString(response.text, format="json")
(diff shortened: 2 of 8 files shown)


python/lsst/daf/butler/core/repoRelocation.py
```diff
@@ -64,8 +64,12 @@ def replaceRoot(configRoot: str, butlerRoot: Optional[ResourcePathExpression]) -
...
         Raised if ``butlerRoot`` is not set but a value is required.
     """
-    # Do nothing if there is nothing to be done
-    if BUTLER_ROOT_TAG not in configRoot:
+    # Do nothing if there is nothing to be done or if there is a type
+    # mismatch.
+    try:
+        if BUTLER_ROOT_TAG not in configRoot:
+            return configRoot
+    except TypeError:
```
andy-slac, line 72: I'm not sure how it can raise TypeError if `configRoot` is a `str` (unless type annotation is wrong).

python/lsst/daf/butler/registries/remote.py
```diff
@@ -182,7 +218,12 @@ def refresh(self) -> None:
 
         # Need to determine what to refresh.
         # Might need a server method to return all the DatasetTypes up front.
-        pass
+        # How do we know which server should be refreshed?
+        # Should there be caches in the client?
+        response = self._client.put(self._get_url("v1/registry/refresh"))
```
andy-slac: Hmm, it feels a bit odd that server-side refresh should depend on one of the clients asking for it explicitly. 🤔 There could be multiple server instances too, and which one of them receives this request is not known to this client. Maybe the client should only refresh its own caches, and assume that server-side state is always current (I guess that probably means we'd have to disable caching in the Registry on server side).

python/lsst/daf/butler/registries/remote.py
```diff
@@ -264,32 +305,55 @@ def removeDatasetType(self, name: str) -> None:
 
     def getDatasetType(self, name: str) -> DatasetType:
         # Docstring inherited from lsst.daf.butler.registry.Registry
-        raise NotImplementedError()
+        path = f"v1/registry/datasetType/{name}"
```
andy-slac: Maybe avoid hardcoding `v1` in every resource, e.g. do it in `_get_url`?

python/lsst/daf/butler/registries/remote.py
```diff
@@ -264,32 +305,55 @@ def removeDatasetType(self, name: str) -> None:
...
         # Docstring inherited from lsst.daf.butler.registry.Registry
         raise NotImplementedError()
 
+    def _simplify_dataId(self, dataId: DataId | None) -> SerializedDataCoordinate | None:
+        """Take a generic Data ID and convert it to a serializable form.
+
+        Parameters
+        ----------
+        dataId : `dict`, `None`, `DataCoordinate`
+            The data ID to normalize.
```
andy-slac: Serialize instead of normalize?

python/lsst/daf/butler/server.py
```diff
@@ -0,0 +1,441 @@
...
+    QueryDataIdsModel,
+    QueryDatasetsModel,
+    QueryDimensionRecordsModel,
+)
+from lsst.daf.butler.registry import CollectionType
+from pydantic import BaseModel
+
+BUTLER_ROOT = "ci_hsc_gen3/DATA"
+
+log = logging.getLogger("excalibur")
```
andy-slac, line 55: ⚔️ ?

python/lsst/daf/butler/server.py
```diff
@@ -0,0 +1,441 @@
...
+from lsst.daf.butler.core.serverModels import (
+    ExpressionQueryParameter,
+    QueryDataIdsModel,
+    QueryDatasetsModel,
+    QueryDimensionRecordsModel,
+)
+from lsst.daf.butler.registry import CollectionType
+from pydantic import BaseModel
+
+BUTLER_ROOT = "ci_hsc_gen3/DATA"
```
andy-slac, line 53: I guess this is just for testing and will disappear later?

python/lsst/daf/butler/server.py
```diff
@@ -0,0 +1,441 @@
...
+log = logging.getLogger("excalibur")
+
+
+class MaximalDataId(BaseModel):
+    """Something that looks like a DataId but isn't."""
+
+    detector: int | None = None
+    instrument: str | None = None
+    physical_filter: str | None = None
+    exposure: int | None = None
```
andy-slac: Is this class used anywhere?

python/lsst/daf/butler/server.py
```diff
@@ -0,0 +1,441 @@
...
+
+
+def unpack_dataId(butler: Butler, data_id: SerializedDataCoordinate | None) -> DataCoordinate | None:
+    """Convert the serialized dataId back to full DataCoordinate.
+
+    Parameters
+    ----------
+    butler : `lsst.daf.butler.Butler`
+        The butler to use for registry and universe.
+    data_id : `SerializedDatacoordinate` or `None`
```
andy-slac: Coordinate with capital C

python/lsst/daf/butler/server.py
```diff
@@ -0,0 +1,441 @@
...
+    root: {BUTLER_ROOT}
+registry:
+    cls: lsst.daf.butler.registries.remote.RemoteRegistry
+    db: <butlerRoot>
+"""
+    config = Config.fromString(config_str, format="yaml")
+    return config
+
+
+@app.get("/butler/universe", response_model=dict[str, Any])
```
andy-slac: Why no `v1` for this one?

python/lsst/daf/butler/server.py
```diff
@@ -0,0 +1,441 @@
...
+    return str(uri)
+
+
+@app.put("/butler/v1/registry/refresh")
+def refresh(butler: Butler = Depends(butler_readonly_dependency)) -> None:
+    # Unclear whether this should exist. Which butler is really being
+    # refreshed? How do we know the server we are refreshing is used later?
+    # For testing at the moment it is important if a test adds a dataset type
+    # directly in the server since the test client will not see it.
+    butler.registry.refresh()
```
andy-slac, line 161: Should it refresh both read-only and writable butlers?

python/lsst/daf/butler/server.py
```diff
@@ -0,0 +1,441 @@
...
+        # SQL getDatasetLocations looks at ID in datastore and does not
+        # check it is in registry. Follow that example and return without
+        # error.
+        return []
+
+    return list(butler.registry.getDatasetLocations(ref))
+
+
+# TimeSpan not yet a pydantic model
+@app.post(
```
andy-slac, line 307: Should it be `GET`?

python/lsst/daf/butler/tests/utils.py
```diff
@@ -289,4 +289,7 @@ def addDataset(
         if run:
             self.butler.registry.registerCollection(run, type=CollectionType.RUN)
         metric = self._makeExampleMetrics()
-        self.butler.put(metric, self.datasetType if datasetType is None else datasetType, dataId, run=run)
+        ref = self.butler.put(
+            metric, self.datasetType if datasetType is None else datasetType, dataId, run=run
+        )
+        print("Put dataset: ", ref)
```
andy-slac: Leftover from debugging?

tests/test_server.py
```diff
@@ -0,0 +1,167 @@
...
+
+try:
+    import lsst.daf.butler.server
+except ImportError:
+    pass  # Import below will set skip variable.
+
+try:
+    from fastapi.testclient import TestClient
+except ImportError:
+    TestClient = None
```
andy-slac: Can you do all optional imports in one `try` block?

andy-slac (approved):
Looks good, few comments. I'm not very familiar yet with httpx/fastapi, will need to learn more. My biggest worry now is server-side state and caching of the registry data in memory, it's clear that we need to think about consistency, probably disable all caching in registry. A related issue is potential caching of responses by proxy servers (if we ever get to that), some of our GET requests are not idempotent, so we may need to disable caching for those too.

python/lsst/daf/butler/core/repoRelocation.py
timj: Funny you should say that. I had to add the try/except because I'm injecting the `TestClient` into the `Config` to override the db field after I have read it and the `Config` classes are not typed at all so eventually what is normally a string ends up being a `TestClient` and the code blows up. Ordinarily the db would be a URL to the server but in the test code we don't run up a server and so this work around has to be used.

python/lsst/daf/butler/registries/remote.py
timj: Yes, my comment in the code there was meant to be me worrying a lot about refresh but in this test code I really need refresh to run in the server because I inject the collections after the initial cached butler is created. You make a later comment on that.

python/lsst/daf/butler/registries/remote.py
timj: Yes. Version handling has to be thought about. The `v1` was a suggestion from @rra to make it explicit when we change things in the future. I guess the get_url method could take an optional version override string as well.

python/lsst/daf/butler/server.py
timj: Two years ago I was having fun with branding and I was fitting in with the SQuaRE naming scheme involving King Arthur names.

python/lsst/daf/butler/server.py
timj: Yes, I override it in the tests anyhow. This will have to come from a configuration file for the real code deployment.

python/lsst/daf/butler/server.py
timj: No. It was used but during testing I ended up deleting the code that used it.

python/lsst/daf/butler/server.py
timj: My feeling was that we would never need to change this and it's a fundamental part of the API that always has to work for all clients (no arguments, returns a dict). Maybe I'm wrong and I should add in a v1 as well here just in case.

python/lsst/daf/butler/server.py
timj: Probably, but I'm having an existential crisis over the refreshing in general. As you say in the overall comment, caching the dataset types and collections in the long-lived server butlers is going to cause a lot of problems since we never really know when something has changed in the database backend that would need a refresh. Globally disabling the cache wold be safest but might be a performance disaster when 10,000 people connect. cc/ @TallJimbo

python/lsst/daf/butler/core/repoRelocation.py
andy-slac: Mypy should have complained about it, because `configRoot` is annotated as `str` and that cannot raise TypeError 🙂

tests/test_server.py
timj: Yes, probably, since the failure mode in both cases is not being able to import fastapi and/or httpx.

python/lsst/daf/butler/server.py
timj: I think it has to be a POST because of the Query parameter in the collections. It's been a while since I had to think about this. If you run the server up it has some very nice automatically created documentation that gets built from knowledge of all the types and models.

python/lsst/daf/butler/server.py
timj: Added a v1 here as well.

PR #779, opened 2023-02-06
DM-37873: Fix code to remove obscore manager in makeRepo

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -961,8 +961,7 @@ class Butler(LimitedButler):
                         if (
                             dimensionName == "visit"
                             and "visit_system_membership" in self.registry.dimensions
-                            and "visit_system"
-                            in self.registry.dimensions["instrument"].metadata  # type: ignore
+                            and "visit_system" in self.registry.dimensions["instrument"].metadata
                         ):
                             instrument_records = list(
                                 self.registry.queryDimensionRecords(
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -489,7 +489,7 @@ def makeRepo(
         config_to_write = config
         if ("registry", "managers", "obscore") in config:
             config_to_write = config.copy()
-            del config_to_write["registry", "managers", "obscore", "config"]
+            del config_to_write["registry", "managers", "obscore"]
```
timj: Maybe I could change this to:

```python
        if (obscore_manager := ("registry", "managers", "obscore")) in config:
             config_to_write = config.copy()
             del config_to_write[obscore_manager]
```
to make it impossible for them not to match.

andy-slac: Hmm, the intent here is to delete "config" sub-key of the "obscore", but the code assumes that "config" exists which I guess is not true. Does this take embargo config? I was assuming that the makeRepo takes a "full" config where obscore is configured with:

    managers:
        obscore:
            cls: manager class
            config: complete obscore config

We want to keep "obscore" in `config_to_write`, but we want to drop its "config" as it's not needed. Maybe change that to:

        if ("registry", "managers", "obscore", "config") in config:
            config_to_write = config.copy()
            del config_to_write["registry", "managers", "obscore", "config"]

But then for execution butler I'm sure we do not need "oscore" at all, maybe we should drop it but not in makeRepo. OTOH if configuration is missing then obscore should be ignored (unless something is also broken in `managers.py`).

andy-slac: I looked at `managers.py` and I believe it should work fine if obscore manager is specified but config is missing. So dropping "config" if it exists should be enough to fix this.

andy-slac (approved):
Looks good.

PR #781, opened 2023-02-07
DM-37868: fix more fallout from DM-31725

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/tests/_registry.py
+++ b/python/lsst/daf/butler/registry/tests/_registry.py
@@ -475,23 +475,20 @@ class RegistryTests(ABC):
                 timespan=timespan,
             ),
         )
-        # If we try to search those same collections without a timespan, the
-        # first one works, since the CALIBRATION collection is irrelevant after
-        # the datast is found in the first collection.  But the second one
-        # should raise.
+        # If we try to search those same collections without a timespan, it
+        # should still work, since the CALIBRATION collection is ignored.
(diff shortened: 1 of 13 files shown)


python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -3165,3 +3165,33 @@ def testExposureQueries(self):
...
         )
+
+    def test_long_query_names(self) -> None:
+        """Test that queries involving very long names are handled correctly.
+
+        This is especially important for PostgreSQL, which truncates symbols
+        longer than 64 chars, but it's worth testing for all DBs.
+        """
+        registry = self.makeRegistry()
+        name = "this_is_a_very_long_name_that_exceeds_the_postgresql_sixty_four_char_limit"
```
andy-slac: I'm going to believe you, but it's not easy to verify that this is actually longer than 64 chars.

andy-slac (approved):
Looks good. I can't parse what `login` means in `Remove problematic collection-type login in findDatasets` commit message.

python/lsst/daf/butler/registry/tests/_registry.py
TallJimbo: I'll replace it.

TallJimbo:
> I can't parse what `login` means

Should have been "logic"; will fix.

PR #782, opened 2023-02-08
DM-37913: Check for big-endian arrays when serializing to Parquet.

erykoff:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/formatters/parquet.py
+++ b/python/lsst/daf/butler/formatters/parquet.py
@@ -1163,6 +1163,20 @@ def _numpy_style_arrays_to_arrow_arrays(
                 val = []
         else:
             val = np_style_arrays[name]
-        arrow_arrays.append(pa.array(val, type=schema.field(name).type))
+
+        try:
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -1163,6 +1163,13 @@ def _numpy_style_arrays_to_arrow_arrays(
                 val = []
         else:
             val = np_style_arrays[name]
-        arrow_arrays.append(pa.array(val, type=schema.field(name).type))
+
+        try:
+            arrow_arrays.append(pa.array(val, type=schema.field(name).type))
+        except pa.ArrowNotImplementedError:
```
leeskelvin: This looks like it will send all `ArrowNotImplementedError` exceptions down to try  byte swap - is that what you want here? Or would some conditional checking work well, e.g.:
```
except pa.ArrowNotImplementedError as err:
    if "Byte-swapped arrays not supported" not in str(e):
        # try byte swapping
    else:
        # raise
```

leeskelvin (approved):

PR #784, opened 2023-02-09
DM-37930: fix bugs in spatial query constraits

TallJimbo:
## Checklist

- [ ] ran Jenkins


timj (approved):
I don't really understand the details of the fix, but given that the tests fail before the fix and they don't fail afterwards then this seems good to me.

PR #787, opened 2023-02-10
DM-37938: additional spatial constraint query fixes and testing

TallJimbo:
See also lsst/daf_relation#6

## Checklist

- [x] ran Jenkins


TallJimbo:
The `build_and_test` failure is just because it needs the `daf_relation` branch; I'll just re-run after that lands instead of adding a temporary pin this time.

timj (approved):
Thanks for the extensive tests.

PR #788, opened 2023-02-10
DM-37744: Make daf_butler compatible with sqlalchemy 2

andy-slac:
Lots of changes everywhere to make it run with sqlalchemy 2. Many mypy fixes. I enabled sqlalchemy 2 in requirements, so GA should build with that.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/ddl.py
+++ b/python/lsst/daf/butler/core/ddl.py
@@ -224,7 +224,7 @@ class GUID(sqlalchemy.TypeDecorator):
 
     cache_ok = True
 
-    def load_dialect_impl(self, dialect: sqlalchemy.Dialect) -> sqlalchemy.TypeEngine:
+    def load_dialect_impl(self, dialect: sqlalchemy.Dialect) -> sqlalchemy.types.TypeEngine:
         if dialect.name == "postgresql":
             return dialect.type_descriptor(UUID())
         else:
@@ -412,7 +412,7 @@ class FieldSpec:
                 return True
         return False
 
-    def getSizedColumnType(self) -> sqlalchemy.types.TypeEngine:
+    def getSizedColumnType(self) -> sqlalchemy.types.TypeEngine | type:
         """Return a sized version of the column type.
 
         Utilizes either (or neither) of ``self.length`` and ``self.nbytes``.
--- a/python/lsst/daf/butler/core/ddl.py
+++ b/python/lsst/daf/butler/core/ddl.py
@@ -211,6 +211,8 @@ class AstropyTimeNsecTai(sqlalchemy.TypeDecorator):
         return value
 
 
+# TODO: sqlalchemy 2 has internal support for UUID:
+# https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Uuid
 class GUID(sqlalchemy.TypeDecorator):
     """Platform-independent GUID type.
 
@@ -249,8 +251,13 @@ class GUID(sqlalchemy.TypeDecorator):
         else:
             return "%.32x" % value.int
 
-    def process_result_value(self, value: Optional[str], dialect: sqlalchemy.Dialect) -> Optional[uuid.UUID]:
+    def process_result_value(
+        self, value: str | uuid.UUID | None, dialect: sqlalchemy.Dialect
+    ) -> Optional[uuid.UUID]:
         if value is None:
+            return value
+        elif isinstance(value, uuid.UUID):
+            # sqlalchemy 2 converts to UUID internally
             return value
         else:
             return uuid.UUID(hex=value)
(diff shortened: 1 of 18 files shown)


python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -1763,7 +1774,9 @@ def query(
             connection = self._engine.connect()
         else:
             connection = self._session_connection
-        result = connection.execute(sql, *args, **kwargs)
+        # TODO: SelectBase is not good for execute(), but it used everywhere,
+        # e.g. in daf_relation. We should switch to Executable at some point.
```
timj, line 1778: Can you make a ticket for this if daf_relation is using some old syntax? cc/ @TallJimbo

python/lsst/daf/butler/core/ddl.py
```diff
@@ -249,9 +251,14 @@ def process_bind_param(self, value: Any, dialect: sqlalchemy.Dialect) -> Optiona
...
             return "%.32x" % value.int
 
-    def process_result_value(self, value: Optional[str], dialect: sqlalchemy.Dialect) -> Optional[uuid.UUID]:
+    def process_result_value(
+        self, value: str | uuid.UUID | None, dialect: sqlalchemy.Dialect
+    ) -> Optional[uuid.UUID]:
         if value is None:
             return value
+        elif isinstance(value, uuid.UUID):
+            # sqlalchemy 2 converts to UUID internally
```
timj, line 260: Does that mean in sqlalchemy 1.4 we can't do this?

python/lsst/daf/butler/registry/databases/postgresql.py
```diff
@@ -197,7 +200,7 @@ def _convertExclusionConstraintSpec(
         metadata: sqlalchemy.MetaData,
     ) -> sqlalchemy.schema.Constraint:
         # Docstring inherited.
-        args = []
+        args: list[tuple[sqlalchemy.schema.Column, str]] = []
```
timj, line 203: Are you surprised this method never gets called?

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -57,7 +58,9 @@
 from .._exceptions import ConflictingDefinitionError
 
 
-def _checkExistingTableDefinition(name: str, spec: ddl.TableSpec, inspection: List[Dict[str, Any]]) -> None:
+# TODO: method is called with list[ReflectedColumn] in SA 2, and
```
timj, line 61: So when we get SA2 we can make the type annotation explicit about that list contents?

timj (approved):
Thanks. Looks great. I can confirm that the tests no longer have any sqlalchemy warnings (and I believe we still have that env var turned on)

python/lsst/daf/butler/registry/databases/postgresql.py
TallJimbo: I'm surprised.  I'd expect this to be called whenever we register a calibration dataset type with a new set of dimensions.

python/lsst/daf/butler/registry/interfaces/_database.py
andy-slac: I made DM-37971 for @TallJimbo, but you can reassign it to me if you don't have time to work on that.

python/lsst/daf/butler/core/ddl.py
andy-slac: In 1.4 strings are returned for UUIDs for both Postgres and sqlite, strings are still handled in `else` clause below.

python/lsst/daf/butler/registry/databases/postgresql.py
andy-slac: Grepping through the code it looks like that it is only called when `TableSpec.exclusion` is not empty, and that could only happen when timespan representation class `hasExclusionConstraint()` returns `True`. I found only one implementation of `hasExclusionConstraint()` which always returns `False`.

python/lsst/daf/butler/registry/interfaces/_database.py
andy-slac: Right,  we should clean it up when we decide to drop 1.4 support.

python/lsst/daf/butler/registry/databases/postgresql.py
TallJimbo: Oh, interesting.  The PostgreSQL-specific timespan representation should have `hasExclusionConstraint()` return `True`, as that's one of the things that `btree_gist` brings, and would avoid a table lock (IIRC) in `certify` (which I thought we were already avoiding for PostgreSQL).

But I'm not sure if we can just fix that without a migration now that we've made the mistake.

PR #789, opened 2023-02-11
DM-34568: Add tuple parameter support to removeDatasetType

Fireye04:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-34568.feature.md
+++ b/doc/changes/DM-34568.feature.md
@@ -1,0 +1,1 @@
+Added the ability to remove multiple dataset types at once, including expansion of wildcards, with `Registry.removeDatasetType` and `butler remove-dataset-type`.
--- a/python/lsst/daf/butler/cli/cmd/commands.py
+++ b/python/lsst/daf/butler/cli/cmd/commands.py
@@ -451,9 +451,9 @@ def query_dataset_types(*args: Any, **kwargs: Any) -> None:
 
 @click.command(cls=ButlerCommand)
 @repo_argument(required=True)
-@click.argument("dataset-type-name", nargs=1)
+@click.argument("dataset-type-name", nargs=-1)
 def remove_dataset_type(*args: Any, **kwargs: Any) -> None:
-    """Remove a dataset type definition from a repository."""
+    """Remove the dataset type definitions from a repository."""
     script.removeDatasetType(*args, **kwargs)
 
 
--- a/python/lsst/daf/butler/registries/sql.py
+++ b/python/lsst/daf/butler/registries/sql.py
@@ -415,7 +415,7 @@ class SqlRegistry(Registry):
         _, inserted = self._managers.datasets.register(datasetType)
         return inserted
 
-    def removeDatasetType(self, name: str | tuple([str, ...])) -> None:
+    def removeDatasetType(self, name: str | tuple[str, ...]) -> None:
         # Docstring inherited from lsst.daf.butler.registry.Registry
 
         for datasetTypeExpression in ensure_iterable(name):
@@ -425,7 +425,7 @@ class SqlRegistry(Registry):
             else:
                 for datasetType in datasetTypes:
                     self._managers.datasets.remove(datasetType.name)
-                    _LOG.info("Removed %r", datasetType.name)
+                    _LOG.info("Removed dataset type %r", datasetType.name)
 
     def getDatasetType(self, name: str) -> DatasetType:
         # Docstring inherited from lsst.daf.butler.registry.Registry
--- a/python/lsst/daf/butler/registries/sql.py
+++ b/python/lsst/daf/butler/registries/sql.py
@@ -415,9 +415,17 @@ class SqlRegistry(Registry):
         _, inserted = self._managers.datasets.register(datasetType)
         return inserted
 
-    def removeDatasetType(self, name: str) -> None:
+    def removeDatasetType(self, name: str | tuple([str, ...])) -> None:
         # Docstring inherited from lsst.daf.butler.registry.Registry
-        self._managers.datasets.remove(name)
+
+        for datasetTypeExpression in ensure_iterable(name):
+            datasetTypes = list(self.queryDatasetTypes(datasetTypeExpression))
+            if not datasetTypes:
+                _LOG.info("Dataset type %r not defined", datasetTypeExpression)
+            else:
+                for datasetType in datasetTypes:
+                    self._managers.datasets.remove(datasetType.name)
+                    _LOG.info("Removed %r", datasetType.name)
 
     def getDatasetType(self, name: str) -> DatasetType:
         # Docstring inherited from lsst.daf.butler.registry.Registry
--- a/python/lsst/daf/butler/registry/_registry.py
+++ b/python/lsst/daf/butler/registry/_registry.py
@@ -595,7 +595,7 @@ class Registry(ABC):
         raise NotImplementedError()
 
     @abstractmethod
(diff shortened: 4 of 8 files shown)


tests/test_butler.py
```diff
@@ -976,6 +976,9 @@ def testPruneCollections(self):
         # dataset type
         butler.registry.removeDatasetType(datasetType.name)
 
+        with self.assertLogs("lsst.daf.butler.registries.sql", "INFO"):
```
timj: We deleted this entire test in #775 so we probably need to move this code and the 3 lines above to the end of the `testRemoveRuns()` method.

python/lsst/daf/butler/registries/sql.py
```diff
@@ -415,9 +415,21 @@ def registerDatasetType(self, datasetType: DatasetType) -> bool:
...
+        for datasetTypeExpression in ensure_iterable(name):
+            datasetTypes = list(self.queryDatasetTypes(datasetTypeExpression))
+            if not datasetTypes:
+                _LOG.info(f"Dataset type {datasetTypeExpression} not defined")
+            else:
+                for datasetType in datasetTypes:
+                    self._managers.datasets.remove(datasetType.name)
+                    _LOG.info(f"Removed {datasetType.name}")
+
+    # def removeDatasetType(self, name: str) -> None:
```
timj: We will have to delete this commented out code. git history will always be around to tell us how things used to be.

python/lsst/daf/butler/registries/sql.py
```diff
@@ -415,9 +415,21 @@ def registerDatasetType(self, datasetType: DatasetType) -> bool:
         _, inserted = self._managers.datasets.register(datasetType)
         return inserted
 
-    def removeDatasetType(self, name: str) -> None:
+    def removeDatasetType(self, name: str | tuple([str, ...])) -> None:
```
timj: You will have to also update the base class (python/lsst/daf/butler/registry/_registry.py) to have the same call signature and to document the `name` parameter. You will likely also need to update `remote.py` in this directory.

python/lsst/daf/butler/registries/sql.py
```diff
@@ -415,9 +415,21 @@ def registerDatasetType(self, datasetType: DatasetType) -> bool:
...
         # Docstring inherited from lsst.daf.butler.registry.Registry
-        self._managers.datasets.remove(name)
+
+        for datasetTypeExpression in ensure_iterable(name):
+            datasetTypes = list(self.queryDatasetTypes(datasetTypeExpression))
+            if not datasetTypes:
+                _LOG.info(f"Dataset type {datasetTypeExpression} not defined")
```
timj: The general comment is that all python logging should use deferred variable expansion. Should be something like:

```python
_LOG.info("Dataset type %r not defined", datasetTypeExpression)
```
where I use `%r` so that the string will be quoted in the log message.

python/lsst/daf/butler/registries/sql.py
```diff
@@ -415,9 +415,21 @@ def registerDatasetType(self, datasetType: DatasetType) -> bool:
...
+            else:
+                for datasetType in datasetTypes:
+                    self._managers.datasets.remove(datasetType.name)
+                    _LOG.info(f"Removed {datasetType.name}")
```
timj: Similarly remove f-string.

tests/test_butler.py
timj: We should include a call to removeDatasetType *before* the calls to `removeRuns` -- that should result in an error (so protect with `with self.assertRaises(the_error):`

tests/test_butler.py
```diff
@@ -976,6 +976,9 @@ def testPruneCollections(self):
...
+            butler.registry.removeDatasetType(tuple(["test", "test2"]))
```
timj: Maybe use a wildcard "test*"? (or use that in the version that does delete correctly above).

tests/test_butler.py
timj: If you look at other uses of `self.assertLogs` you will see that it's possible to check the log message that is issued so you can see that you are getting the expected log message. You might want to use `lsst.daf.butler.registries` as the logger to check rather than being super specific.

timj (commented):
Quick comments on the draft. Once you've dealt with these you will also need to change the click interface in commands.py so that it can accept multiple values for the dataset type name since it currently says `nargs=1`.

Fireye04:
Note to future self: update release notes before merging

python/lsst/daf/butler/cli/cmd/commands.py
```diff
@@ -451,9 +451,11 @@ def query_dataset_types(*args: Any, **kwargs: Any) -> None:
 
 @click.command(cls=ButlerCommand)
 @repo_argument(required=True)
-@click.argument("dataset-type-name", nargs=1)
+@click.argument("dataset-type-name", nargs=-1)
 def remove_dataset_type(*args: Any, **kwargs: Any) -> None:
-    """Remove a dataset type definition from a repository."""
+    """Remove the dataset type definitions from a repository."""
+    print(args)
```
timj: Remember to remove these two prints (and then squash the commits).

python/lsst/daf/butler/registries/sql.py
```diff
@@ -415,9 +415,17 @@ def registerDatasetType(self, datasetType: DatasetType) -> bool:
...
-        self._managers.datasets.remove(name)
+
+        for datasetTypeExpression in ensure_iterable(name):
+            datasetTypes = list(self.queryDatasetTypes(datasetTypeExpression))
+            if not datasetTypes:
+                _LOG.info("Dataset type %r not defined", datasetTypeExpression)
+            else:
+                for datasetType in datasetTypes:
+                    self._managers.datasets.remove(datasetType.name)
+                    _LOG.info("Removed %r", datasetType.name)
```
timj: We should probably say `"Removed dataset type %r"` here.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -607,8 +607,9 @@ def removeDatasetType(self, name: str) -> None:
 
         Parameters
         ----------
-        name : `str`
-            Name of the type to be removed.
+        name : `str` or `tuple([str, ...])`
+            Name of the type to be removed or tuple containing a list of type
+            names to be removed.
```
timj: Add "Wildcards are allowed."

tests/test_cliCmdQueryDatasetTypes.py
```diff
@@ -149,6 +149,99 @@ def testQueryDatasetTypes(self):
...
+                        "register-dataset-type",
+                        "here",
+                        name,
+                        storageClassName,
+                        instrumentDimension,
+                        visitDimension,
+                    ],
+                )
+
+            # checks wildcard / literal combination
```
timj: Python prefers full sentences for code comments. This would be something like:

```python
# Check wildcard / literal combination.
```

Same comment for the comments below.

timj (approved):
Looks great. Very minor comments. Try to squash the fixes into the relevant earlier commit.

doc/changes/DM-34568.feature.md
```diff
@@ -0,0 +1,2 @@
+Add support for tuple parameters in the removeDatasetType command.
```
timj: How about something like:

> Added the ability to remove multiple dataset types at once, including expansion of wildcards, with ``Registry.removeDatasetType`` and ``butler remove-dataset-type``.

?

Fireye04: updated & squashed :+1:

PR #790, opened 2023-02-17
DM-38062: Allow test to pass if `--log-level` is used for pytest

timj:
There is still a confusing problem if `pytest --log-cli-level` is used.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/tests/cliLogTestBase.py
+++ b/python/lsst/daf/butler/tests/cliLogTestBase.py
@@ -151,7 +151,9 @@ class CliLogTestBase:
             result = cmd()
         self.assertEqual(result.exit_code, 0, clickResultMsg(result))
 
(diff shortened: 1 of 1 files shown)


timj:
For the record, if I use `--log-cli-level=DEBUG` tests involving the Click runner mysteriously fail with:

```python
>               stdout = outstreams[0].getvalue()
E               ValueError: I/O operation on closed file.

../../miniconda/envs/lsst-scipipe-5.0.0/lib/python3.10/site-packages/click/testing.py:434: ValueError
```

If I modify the click testing package to put a try/except around that line then all tests pass. I have no idea what is causing the problem (`--log-cli-level=16` works but `=15` fails).

Seems to be a known problem: pallets/click#824

ktlim (approved):

PR #791, opened 2023-02-21
DM-37995: Improve storage class override logic

andy-slac:
In trusted mode the datastore has to use original registry storage class
to search for artifacts. Because datastore has no access to registry
a new method was added to Datastore class to specify a way to retrieve
registry dataset type based on its name (dependency inversion).

Butler instances use this new method to provide access to to dataset
type mapping. In case of QBB, when registry does not exist this mapping
has to be provided when constructing QBB.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_quantum_backed.py
+++ b/python/lsst/daf/butler/_quantum_backed.py
@@ -207,7 +207,7 @@ class QuantumBackedButler(LimitedButler):
             location records.  Default is a SQL-backed implementation.
         search_paths : `list` of `str`, optional
             Additional search paths for butler configuration.
-        dataset_types: `Mapping` [`str`, `DatasetType`]
+        dataset_types: `Mapping` [`str`, `DatasetType`], optional
             Mapping of the dataset type name to its registry definition.
         """
         predicted_inputs = [
@@ -272,7 +272,7 @@ class QuantumBackedButler(LimitedButler):
             location records.  Default is a SQL-backed implementation.
         search_paths : `list` of `str`, optional
             Additional search paths for butler configuration.
-        dataset_types: `Mapping` [`str`, `DatasetType`]
+        dataset_types: `Mapping` [`str`, `DatasetType`], optional
             Mapping of the dataset type name to its registry definition.
         """
         return cls._initialize(
--- a/python/lsst/daf/butler/_quantum_backed.py
+++ b/python/lsst/daf/butler/_quantum_backed.py
@@ -38,6 +38,7 @@ from .core import (
     Config,
     DatasetId,
     DatasetRef,
+    DatasetType,
     Datastore,
     DatastoreRecordData,
     DimensionUniverse,
@@ -153,6 +154,7 @@ class QuantumBackedButler(LimitedButler):
         dimensions: DimensionUniverse,
         datastore: Datastore,
         storageClasses: StorageClassFactory,
+        dataset_types: Mapping[str, DatasetType] | None = None,
     ):
         self._dimensions = dimensions
         self._predicted_inputs = set(predicted_inputs)
@@ -163,6 +165,10 @@ class QuantumBackedButler(LimitedButler):
         self._actual_output_refs: Set[DatasetRef] = set()
         self.datastore = datastore
         self.storageClasses = storageClasses
+        self._dataset_types: Mapping[str, DatasetType] = {}
+        if dataset_types is not None:
+            self._dataset_types = dataset_types
+        self.datastore.set_retrieve_dataset_type_method(self._retrieve_dataset_type)
 
     @classmethod
     def initialize(
@@ -174,6 +180,7 @@ class QuantumBackedButler(LimitedButler):
         OpaqueManagerClass: Type[OpaqueTableStorageManager] = ByNameOpaqueTableStorageManager,
         BridgeManagerClass: Type[DatastoreRegistryBridgeManager] = MonolithicDatastoreRegistryBridgeManager,
         search_paths: Optional[List[str]] = None,
+        dataset_types: Mapping[str, DatasetType] | None = None,
     ) -> QuantumBackedButler:
         """Construct a new `QuantumBackedButler` from repository configuration
         and helper types.
@@ -200,6 +207,8 @@ class QuantumBackedButler(LimitedButler):
             location records.  Default is a SQL-backed implementation.
         search_paths : `list` of `str`, optional
             Additional search paths for butler configuration.
+        dataset_types: `Mapping` [`str`, `DatasetType`]
+            Mapping of the dataset type name to its registry definition.
         """
         predicted_inputs = [
             ref.getCheckedId() for ref in itertools.chain.from_iterable(quantum.inputs.values())
@@ -218,6 +227,7 @@ class QuantumBackedButler(LimitedButler):
             OpaqueManagerClass=OpaqueManagerClass,
             BridgeManagerClass=BridgeManagerClass,
             search_paths=search_paths,
+            dataset_types=dataset_types,
         )
 
     @classmethod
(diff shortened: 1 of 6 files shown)


python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -2890,3 +2900,26 @@ def export_records(self, refs: Iterable[DatasetIdRef]) -> Mapping[str, Datastore
...
+        be called beforehand.
+        """
+        if self.trustGetRequest:
+            if self._retrieve_dataset_method is None:
+                # We could raise an exception here but unit tests do not define
+                # this method.
+                return ref
+            dataset_type = self._retrieve_dataset_method(ref.datasetType.name)
+            if dataset_type is not None:
+                if ref.datasetType.storageClass.name != dataset_type.storageClass.name:
```
timj: This test is done inside the override method and it can take the storage class directly.

```python
if dataset_type is not None:
    ref = ref.overrideStorageClass(dataset_type.storageClass)
```
should be fine.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -2890,3 +2900,26 @@ def export_records(self, refs: Iterable[DatasetIdRef]) -> Mapping[str, Datastore
...
         return {self.name: record_data}
+
+    def set_retrieve_dataset_type_method(self, method: Callable[[str], DatasetType | None] | None) -> None:
+        # Docstring inherited from the base class.
+        self._retrieve_dataset_method = method
+
+    def _cast_storage_class(self, ref: DatasetRef) -> DatasetRef:
+        """Update dataset reference to use the storage class from registry.
+
+        This does nothing for regular datasetores, and is only enabled for
```
timj: typo: datasetores

tests/test_datastore.py
```diff
@@ -438,17 +440,44 @@ def testTrustGetRequest(self):
...
+
+                # Make new dataset ref with compatible storage class.
+                ref_comp = ref.overrideStorageClass("StructuredDataDictJson")
+
+                # get() should return a dict now
+                metrics_as_dict = datastore.get(ref_comp)
+                self.assertIsInstance(metrics_as_dict, dict)
+
+                # exists() should work as well
+                self.assertTrue(datastore.exists(ref_comp))
```
timj, line 485: Is it possible for there also to be a test that fails if the set_retrieve_dataset_type_method had not been overridden?

python/lsst/daf/butler/core/datastore.py
```diff
@@ -1201,3 +1201,22 @@ def export_records(
...
+        Parameters
+        ----------
+        method : `~collections.abc.Callable` | `None`
+            Method that takes a name of the dataset type and returns a
+            corresponding `DatasetType` instance as defined in Registry. If
+            dataset type name is not known to registry `None` is returned.
+
+        Notes
+        -----
+        This method si only needed for a Datastore supporting a "trusted" mode
```
timj:
```suggestion
        This method is only needed for a Datastore supporting a "trusted" mode
```

python/lsst/daf/butler/_quantum_backed.py
```diff
@@ -261,6 +272,8 @@ def from_predicted(
             location records.  Default is a SQL-backed implementation.
         search_paths : `list` of `str`, optional
             Additional search paths for butler configuration.
+        dataset_types: `Mapping` [`str`, `DatasetType`]
```
timj:
```suggestion
        dataset_types: `Mapping` [`str`, `DatasetType`], optional
```

python/lsst/daf/butler/_quantum_backed.py
```diff
@@ -200,6 +207,8 @@ def initialize(
...
```
timj:
```suggestion
        dataset_types: `Mapping` [`str`, `DatasetType`], optional
```

tests/test_datastore.py
```diff
@@ -438,17 +440,44 @@ def testTrustGetRequest(self):
...
 
             # Get the URI -- if we trust this should work even without
             # enabling prediction.
             primaryURI2, componentURIs2 = datastore.getURIs(ref)
             self.assertEqual(primaryURI2, primaryURI)
             self.assertEqual(componentURIs2, componentURIs)
 
+            # Check for compatible storage class
+            if sc_name in ("StructuredDataNoComponents", "StructuredData"):
+                # need a special method to generate stored dataset type
```
timj:
```suggestion
                # Need a special method to generate stored dataset type.
```

tests/test_datastore.py
```diff
@@ -438,17 +440,44 @@ def testTrustGetRequest(self):
...
+                output = datastore.get(compRef)
+                self.assertEqual(output, getattr(metrics, comp))
 
             # Get the URI -- if we trust this should work even without
             # enabling prediction.
             primaryURI2, componentURIs2 = datastore.getURIs(ref)
             self.assertEqual(primaryURI2, primaryURI)
             self.assertEqual(componentURIs2, componentURIs)
 
+            # Check for compatible storage class
```
timj:
```suggestion
            # Check for compatible storage class.
```

tests/test_datastore.py
```diff
@@ -438,17 +440,44 @@ def testTrustGetRequest(self):
...
+            if sc_name in ("StructuredDataNoComponents", "StructuredData"):
+                # need a special method to generate stored dataset type
+                def _stored_dataset_type(name: str) -> DatasetType:
+                    if name == ref.datasetType.name:
+                        return ref.datasetType
+                    raise ValueError(f"Unexpected dataset type name {ref.datasetType.name}")
+
+                datastore.set_retrieve_dataset_type_method(_stored_dataset_type)
+
+                # Storage class override with original dataset ref
```
timj:
```suggestion
                # Storage class override with original dataset ref.
```

tests/test_datastore.py
```diff
@@ -438,17 +440,44 @@ def testTrustGetRequest(self):
...
+                metrics_as_dict = datastore.get(ref, storageClass="StructuredDataDictJson")
+                self.assertIsInstance(metrics_as_dict, dict)
+
+                # Make new dataset ref with compatible storage class.
+                ref_comp = ref.overrideStorageClass("StructuredDataDictJson")
+
+                # get() should return a dict now
```
timj:
```suggestion
                # get() should return a dict now.
```

tests/test_datastore.py
```diff
@@ -438,17 +440,44 @@ def testTrustGetRequest(self):
...
+                metrics_as_dict = datastore.get(ref_comp)
+                self.assertIsInstance(metrics_as_dict, dict)
+
+                # exists() should work as well
```
timj:
```suggestion
                # exists() should work as well.
```

tests/test_quantumBackedButler.py
```diff
@@ -128,7 +135,9 @@ def test_initialize(self) -> None:
         """Test for initialize factory method"""
 
         quantum = self.make_quantum()
-        qbb = QuantumBackedButler.initialize(config=self.config, quantum=quantum, dimensions=self.universe)
+        qbb = QuantumBackedButler.initialize(
```
timj, line 138: Do any of these tests fail without the dataset types? I assume without it all falls back to the graph dataset type which will mostly work fine.

timj (approved):
I'm not overly happy but this is the pragmatic solution :-)

tests/test_datastore.py
andy-slac: I have added failing tests with set_retrieve_dataset_type_method unset.

tests/test_quantumBackedButler.py
andy-slac: We do not have tests here which override storage classes, so in this case it's not strictly necessary to pass that parameter here, tests will work OK even without it.

PR #792, opened 2023-02-22
DM-38054: Add DataCoordinate attribute access to DimensionRecords.

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/dimensions/_coordinate.py
+++ b/python/lsst/daf/butler/core/dimensions/_coordinate.py
@@ -156,6 +156,10 @@ class DataCoordinate(NamedKeyMapping[Dimension, DataIdValue]):
        between simple `dict` data IDs work, and hence using a `DataCoordinate`
        instance for at least one operand in any data ID comparison is strongly
        recommended.
+
+    See also
+    --------
+    :ref:`lsst.daf.butler-dimensions_data_ids`
     """
 
     __slots__ = ()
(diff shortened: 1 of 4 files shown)


timj (approved):
Looks good to me. Can we also have an explicit test for the common usage? In test_dimensions.py there is a line:

```python
self.assertEqual(dataId.timespan, dataId.records["visit"].timespan)
```

can we add a line underneath it as:

```python
self.assertEqual(dataId.timespan, dataId.visit.timespan)
```
?

python/lsst/daf/butler/core/dimensions/_coordinate.py
```diff
@@ -1045,3 +1049,14 @@ def hasRecords(self) -> bool:
     def _record(self, name: str) -> Optional[DimensionRecord]:
         # Docstring inherited from DataCoordinate.
         return self._records[name]
+
+    def __getattr__(self, name: str) -> Any:
+        try:
+            return self._record(name)
+        except KeyError:
+            raise AttributeError(name) from None
```
timj, line 1070: It wouldn't hurt to also have a test on an unexpanded dataId to make sure it raises AttributeError? Then I imagine the follow up will be that people will want to distinguish between an attribute that is failing because it's not part of the dataId vs one that is failing because the records were not available (so maybe the error message should mention the distinction).

python/lsst/daf/butler/core/dimensions/_coordinate.py
```diff
@@ -156,6 +156,10 @@ class DataCoordinate(NamedKeyMapping[Dimension, DataIdValue]):
        between simple `dict` data IDs work, and hence using a `DataCoordinate`
        instance for at least one operand in any data ID comparison is strongly
        recommended.
+
+    See also
+    --------
+    :ref:`lsst.daf.butler-dimensions_data_ids`
```
timj, line 162: I tried building the docs but I didn't see this section included. May want to consider embedding it into the Notes section instead.

timj (approved):

python/lsst/daf/butler/core/dimensions/_coordinate.py
TallJimbo: It's there (or at least it is for me) - it's a little call-out box _above_ the Notes section, unlike the way it appears in the source.

timj: Ah. Yes. I see it now I know what I'm looking for. Sorry.

PR #794, opened 2023-02-27
DM-38065: V24.1 release notes

timj:
## Checklist

- [ ] ran Jenkins


timj:
The mypy failure is not important (we have no means of pinning the mypy version to the one that was current when we made the v24 release). The build has to pass though since this is required for PyPI publication.

PR #797, opened 2023-03-03
DM-38210: Make butler.get act like getDirect and deprecate getDirect

timj:
We now want a resolved DatasetRef to work with butler.get without doing registry lookup. Also change the LimitedButler to add a .get(). .getDirect and deferred form are now deprecated.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -27,10 +27,6 @@ from __future__ import annotations
 __all__ = (
     "Butler",
     "ButlerValidationError",
-    "PruneCollectionsArgsError",
-    "PurgeWithoutUnstorePruneCollectionsError",
-    "RunWithoutPurgePruneCollectionsError",
-    "PurgeUnsupportedPruneCollectionsError",
 )
 
 import collections.abc
@@ -112,42 +108,6 @@ class ButlerValidationError(ValidationError):
     """There is a problem with the Butler configuration."""
 
     pass
-
-
-class PruneCollectionsArgsError(TypeError):
-    """Base class for errors relating to Butler.pruneCollections input
-    arguments.
-    """
-
-    pass
-
-
-class PurgeWithoutUnstorePruneCollectionsError(PruneCollectionsArgsError):
-    """Raised when purge and unstore are both required to be True, and
-    purge is True but unstore is False.
-    """
-
-    def __init__(self) -> None:
-        super().__init__("Cannot pass purge=True without unstore=True.")
-
-
-class RunWithoutPurgePruneCollectionsError(PruneCollectionsArgsError):
-    """Raised when pruning a RUN collection but purge is False."""
-
-    def __init__(self, collectionType: CollectionType):
-        self.collectionType = collectionType
-        super().__init__(f"Cannot prune RUN collection {self.collectionType.name} without purge=True.")
-
-
-class PurgeUnsupportedPruneCollectionsError(PruneCollectionsArgsError):
-    """Raised when purge is True but is not supported for the given
-    collection."""
-
-    def __init__(self, collectionType: CollectionType):
-        self.collectionType = collectionType
-        super().__init__(
-            f"Cannot prune {self.collectionType} collection {self.collectionType.name} with purge=True."
-        )
 
 
 class Butler(LimitedButler):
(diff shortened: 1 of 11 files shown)


timj:
@TallJimbo can you take a quick look at this to see if it matches your expectations? Obviously a lot of the checks for `.id is not None` will go away when the other half of the deprecation is done but it seems like getting this deprecation for `getDirect` out the way early would be a good idea. I need to fix the other lsst_distrib packages.

python/lsst/daf/butler/_limited_butler.py
```diff
@@ -91,6 +95,135 @@ def putDirect(self, obj: Any, ref: DatasetRef) -> DatasetRef:
...
+
+    def get(
+        self,
+        datasetRefOrType: DatasetRef | DatasetType | str,
+        dataId: DataId | None = None,
+        *,
+        parameters: dict[str, Any] | None = None,
+        collections: Any = None,
+        storageClass: StorageClass | str | None = None,
+        **kwargs: Any,
```
TallJimbo: I was thinking we'd just make the `LimitedButler` signature
```
def get(self, ref: DatasetRef, /, *, parameters: dict[str, Any] | None = None) -> Any:
```
Maybe it should have `storageClass`, too, but since `ref` has one in it, and QG execution can easily make sure `ref` has the right one in it, it doesn't seem necessary (and as we've seen, we can't assume the `ref` storage class is the Registry one).

That would also let us shrink the `_findDatasetRef` implementation (or just merge it into other methods), and it'd keep the `LimitedButler` interface from being misleading.

To make that work with MyPy we might have to use `typing.overload` in full `Butler` and make `datasetRefOrType` a positional-only parameter, to show that it does satisfy the original signature as well as adding a new one.

Same comment for `getDeferred`.

TallJimbo (approved):
Looks good in terms of behavior, but I'd like to handle the signatures in `LimitedButler` slightly differently.

`datasetExistsDirect` should probably also go away, and the `getURI[s]` are probably worth a look to see if they need modifications (my recollection is that they already have behavior consistent with the new `get` behavior, but I'm not sure).

python/lsst/daf/butler/_limited_butler.py
timj: If `LimitedButler.get()` has a different signature to `Butler.get()` that seems a bit odd to me since it implies that the user needs to understand that `Butler` is special and is the only subclass that is allowed to pass in something other than a `DatasetRef`. Don't we generally frown upon subclasses having completely different parameters to other subclasses?

TallJimbo: The user needs to understand that `Butler` is special regardless; I think the only thing that's actually up for debate here is whether _MyPy_ should understand that that `LimitedButler` requires a `DatasetRef` and hence `Butler` is special.

> Don't we generally frown upon subclasses having completely different parameters to other subclasses?

Yes, in the absence of overloading.  And overloading is maybe sort of frowned upon, but it's what users said they wanted, and once you accept overloading, we're not changing the signature; we're allowing a new one, and that is more analogous to adding a new method.

timj: @TallJimbo as you surmised, mypy has no problem with the subclass having additional parameters so long as the signature includes a `/` to force it not to support a keyword value usage.

timj:
>  the getURI[s] are probably worth a look 

`getURI` uses `_findDatasetRef` so will now work directly with a resolved ref.

> datasetExistsDirect should probably also go away

This begs the question of whether we are "fixing" `datasetExists` such that it only returns a boolean rather than raising if registry doesn't know about the dataset. In [DM-32940](https://jira.lsstcorp.org/browse/DM-32940) you suggest we change that to return an enum that can evaluate as a boolean. Then `datasetExistsDirect` could go away. We have known to registry, known to datastore, actually exists on "disk" as options but only the last one is True.

I assume `putDirect` should disappear as well.

TallJimbo:
:+1: to removing `putDirect`.

What to do about `datasetExists` is trickier.  I tried an enum on the DMTN-249 prototyping and while I think it might have worked fine for full `Butler` alone, it didn't make sense for `LimitedButler` where there's no concept of registry existence.  And if you have something that checks for datastore existence only in `LimitedButler`, that doesn't make sense in full `Butler`, where Datastore-only existence checks are designed to almost never be necessary.

So I think our best bet is to give the `LimitedButler` datastore-only method a somewhat obscure, very precise name, and reserve `exists` or `datastoreExists` as a full-`Butler` method that returns an enum, but we might need to experiment on that a bit, or gather feedback from users.

python/lsst/daf/butler/_limited_butler.py
```diff
@@ -91,6 +132,56 @@ def putDirect(self, obj: Any, ref: DatasetRef) -> DatasetRef:
...
+            A resolved `DatasetRef` directly associated with a dataset.
+        parameters : `dict`
+            Additional StorageClass-defined options to control reading,
+            typically used to efficiently read only a subset of the dataset.
+        storageClass : `StorageClass` or `str`, optional
+            The storage class to be used to override the Python type
+            returned by this method. By default the returned type matches
+            the dataset type definition for this dataset. Specifying a
+            read `StorageClass` can force a different type to be returned.
+            This type must be compatible with the original type.
```
TallJimbo, line 156: I'd rather not have a `storageClass` argument here, since the caller can override the registry storage class via `ref.datasetType.storageClass` already and I think that makes it more confusing than convenient (especially since right now `LimitedButler` mostly exists just for task execution and that doesn't use this parameter at all).  I think that argument is even more applicable to `Datastore.get`, since that really isn't in need of convenience functionality since only butler should call it, but that may be out of scope for this ticket.

To be clear, I _do_ think it's worth having a `storageClass` parameter to full `Butler.get`, since there it's very convenient when not passing a `DatasetRef`, and then since it'll present anyway it might as well further override what's in the ref when a `DatasetRef` is passed.

python/lsst/daf/butler/_butler.py
```diff
@@ -1660,7 +1655,15 @@ def datasetExists(
         TypeError
             Raised if no collections were provided.
         """
-        ref = self._findDatasetRef(datasetRefOrType, dataId, collections=collections, **kwargs)
+        # A resolved ref may be given that is not known to this butler.
+        if isinstance(datasetRefOrType, DatasetRef) and datasetRefOrType.id is not None:
+            ref = self.registry.getDataset(datasetRefOrType.id)
+            if ref is None:
+                raise LookupError(
+                    f"Resolved DatasetRef with id {datasetRefOrType.id} " "is not known to registry."
```
```suggestion
                    f"Resolved DatasetRef with id {datasetRefOrType.id} is not known to registry."
```

TallJimbo (approved):

python/lsst/daf/butler/_limited_butler.py
timj: My main argument for keeping it is that it's forwarded directly to datastore and datastore will use it and treat it as the priority. It seems odd to declare that the base class can't support the parameter even when it can.

PR #798, opened 2023-03-06
DM-38240: Add ChainedDatastore.transfer_from support

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -2415,13 +2415,22 @@ class Butler(LimitedButler):
 
             # Ask the datastore to transfer. The datastore has to check that
             # the source datastore is compatible with the target datastore.
-            self.datastore.transfer_from(
+            accepted, rejected = self.datastore.transfer_from(
                 source_butler.datastore,
                 source_refs,
                 local_refs=transferred_refs,
                 transfer=transfer,
                 artifact_existence=artifact_existence,
             )
+            if rejected:
+                # For now, accept the registry entries but not the files.
+                log.warning(
+                    "%d datasets were rejected and %d accepted for dataset type %s in run %r.",
+                    len(rejected),
+                    len(accepted),
+                    datasetType,
+                    run,
+                )
 
         return transferred_refs
 
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -2260,9 +2260,11 @@ class Butler(LimitedButler):
             )
             source_refs = [ref for ref, exists in dataset_existence.items() if exists]
             filtered_count = len(source_refs)
+            n_missing = original_count - filtered_count
             log.verbose(
-                "%d datasets removed because the artifact does not exist. Now have %d.",
-                original_count - filtered_count,
+                "%d dataset%s removed because the artifact does not exist. Now have %d.",
+                n_missing,
+                "" if n_missing == 1 else "s",
                 filtered_count,
             )
 
--- a/python/lsst/daf/butler/core/datastore.py
+++ b/python/lsst/daf/butler/core/datastore.py
@@ -742,7 +742,7 @@ class Datastore(metaclass=ABCMeta):
         local_refs: Optional[Iterable[DatasetRef]] = None,
         transfer: str = "auto",
         artifact_existence: Optional[Dict[ResourcePath, bool]] = None,
-    ) -> None:
+    ) -> tuple[set[DatasetRef], set[DatasetRef]]:
         """Transfer dataset artifacts from another datastore to this one.
 
         Parameters
@@ -772,6 +772,13 @@ class Datastore(metaclass=ABCMeta):
             Optional mapping of datastore artifact to existence. Updated by
             this method with details of all artifacts tested. Can be `None`
             if the caller is not interested.
+
+        Returns
+        -------
+        accepted : `set` [`DatasetRef`]
+            The datasets that were transferred.
+        rejected : `set` [`DatasetRef`]
+            The datasets that were rejected due to a constraints violation.
 
         Raises
         ------
--- a/python/lsst/daf/butler/datastores/chainedDatastore.py
+++ b/python/lsst/daf/butler/datastores/chainedDatastore.py
@@ -1047,26 +1047,40 @@ class ChainedDatastore(Datastore):
         all_accepted = set()
         nsuccess = 0
         for source_child in source_datastores:
+            # If we are reading from a chained datastore, it's possible that
+            # only a subset of the datastores know about the dataset. We can't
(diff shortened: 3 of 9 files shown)


python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -1018,3 +1018,84 @@ def export(
...
+        self,
+        source_datastore: Datastore,
+        refs: Iterable[DatasetRef],
+        local_refs: Optional[Iterable[DatasetRef]] = None,
+        transfer: str = "auto",
+        artifact_existence: Optional[Dict[ResourcePath, bool]] = None,
+    ) -> tuple[set[DatasetRef], set[DatasetRef]]:
+        # Docstring inherited
+        # mypy does not understand "type(self) is not type(source)"
+        if isinstance(source_datastore, type(self)):
```
andy-slac: I guess we do not subclass ChainedDatastore so this should work OK, but if some day we add ChainedDatastore subclass then this test may fail.

python/lsst/daf/butler/core/datastore.py
```diff
@@ -773,6 +773,13 @@ def transfer_from(
             this method with details of all artifacts tested. Can be `None`
             if the caller is not interested.
 
+        Returns
+        -------
+        accepted : `set` [`DatasetRef`]
+            The datasets that were transferred.
+        rejected : `set` [`DatasetRef`]
+            The datasets that were rejected due to a constraints violation.
```
andy-slac, line 781: Are these refs from `refs` or `local_refs` (when it's not None)? I would assume that these two sets include all input refs, but maybe mention it explicitly here?

tests/test_butler.py
```diff
@@ -1798,6 +1798,14 @@ def testTransferUuidToUuid(self):
         # Setting id_gen_map should have no effect here
         self.assertButlerTransfers(id_gen_map={"random_data_2": DatasetIdGenEnum.DATAID_TYPE})
 
+    def _enable_trust(self, datastore) -> None:
+        if hasattr(datastore, "trustGetRequest"):
+            datastore.trustGetRequest = True
+        elif hasattr(datastore, "datastores"):
+            for datastore in datastore.datastores:
+                if hasattr(datastore, "trustGetRequest"):
+                    datastore.trustGetRequest = True
```
andy-slac, line 1807: Could chained datastore include other chained datastores?

tests/test_butler.py
```diff
@@ -1914,17 +1922,31 @@ def assertButlerTransfers(self, id_gen_map=None, purge=False, storageClassName="
...
-                            primary.remove()
-                        for uri in uris.values():
-                            uri.remove()
+                        # For a chained datastore we need to remove
+                        # files in each chain.
+                        for datastore in datastores:
+                            primary, uris = datastore.getURIs(ref)
+                            if primary:
+                                if primary.scheme != "mem":
+                                    primary.remove()
```
andy-slac, line 1944: Maybe the check for "mem" is better done inside `remove()` so that you need to know less about this special case?

python/lsst/daf/butler/_butler.py
```diff
@@ -2415,13 +2415,22 @@ def transfer_from(
...
-            self.datastore.transfer_from(
+            accepted, rejected = self.datastore.transfer_from(
                 source_butler.datastore,
                 source_refs,
                 local_refs=transferred_refs,
                 transfer=transfer,
                 artifact_existence=artifact_existence,
             )
+            if rejected:
+                # For now, accept the registry entries but not the files.
```
andy-slac, line 2426: Not sure what this comment means 🤔

andy-slac (approved):

python/lsst/daf/butler/_butler.py
timj: It means I'm not now going to try to remove the entries from the destination registry for all the times that datastore couldn't receive the actual file. We could go back and remove the entries from Registry (maybe we should). @TallJimbo do you have an opinion on what happens if you transfer from A to B but datastore in B rejects doesn't accept the files because of constraints? We can (a) raise an exception because you asked us to transfer something and it didn't work (b) have registry entries but no datastore records (c) remove the registry entries as well. Also, should we raise if this means that nothing was transferred?

tests/test_butler.py
timj: I hope not. I don't think anything stops it though.

python/lsst/daf/butler/datastores/chainedDatastore.py
timj: If it's subclassed I assume there is some assumption that the API we use will still work.

tests/test_butler.py
timj: Hmm. Yes, I can probably make `resource.remove()` a no-op for `mem://` -- I think I resisted doing that because you could conceive of people thinking that would delete the in-memory resource when it can't actually do that.

python/lsst/daf/butler/core/datastore.py
timj: Good question. It seems like I'm currently rejecting the target ref and not the source ref but since we are using UUID they are the same thing. The real answer is that I need to change things a bit to now understand that there is no difference between source and target ref.

1. Remove the id_gen_map parameter from Butler.transfer_from (they are always UUID)
2. Remove the local_refs parameter from Datastore.transfer_from.

This will make things a lot simpler. I will see if I can quickly remove the relevant code -- if it's just removing code I won't send it for a new review.

timj: Cleaning up to assume UUID is quite involved. Will put on a new ticket.

python/lsst/daf/butler/datastores/chainedDatastore.py
andy-slac: I mean that if `self` is a subclass instance and `source_datastore` is just a `ChaninedDatastore` then `isinstance` will return False. Maybe it's safer to just do `isinstance(source_datastore, ChanedDatastore)`?

tests/test_butler.py
andy-slac: It maybe safer to make this method to recursively call itself? But for tests it probably does not matter.

timj: Right, if a test fails because somehow I've got a chained datastore inside a chained datastore I can fix the test.

python/lsst/daf/butler/_butler.py
TallJimbo: Transferring registry content and dropping the datastore content seems like the best course of action, but I'm also not opposed to raising (and we might be able to do that up front, just based on the dataset types being transferred and the constraints).

timj: Ok. Will leave it as is.

PR #799, opened 2023-03-07
DM-37279: Add ArrowNumpyDict storage class to parquet formatter

erykoff:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -113,6 +113,7 @@ storageClasses:
       pyarrow.Table: lsst.daf.butler.formatters.parquet.arrow_to_pandas
       astropy.table.Table: astropy.table.Table.to_pandas
       numpy.ndarray: pandas.DataFrame.from_records
+      dict: pandas.DataFrame.from_records
     delegate: lsst.daf.butler.delegates.dataframe.DataFrameDelegate
     derivedComponents:
       columns: DataFrameIndex
@@ -138,6 +139,7 @@ storageClasses:
       pandas.core.frame.DataFrame: lsst.daf.butler.formatters.parquet.pandas_to_arrow
       astropy.table.Table: lsst.daf.butler.formatters.parquet.astropy_to_arrow
       numpy.ndarray: lsst.daf.butler.formatters.parquet.numpy_to_arrow
+      dict: lsst.daf.butler.formatters.parquet.numpy_dict_to_arrow
     delegate: lsst.daf.butler.delegates.arrowtable.ArrowTableDelegate
     derivedComponents:
       columns: ArrowColumnList
@@ -163,6 +165,7 @@ storageClasses:
       pyarrow.Table: lsst.daf.butler.formatters.parquet.arrow_to_astropy
       pandas.core.frame.DataFrame: lsst.daf.butler.formatters.parquet.pandas_to_astropy
       numpy.ndarray: lsst.daf.butler.formatters.parquet.numpy_to_astropy
+      dict: astropy.table.Table
     delegate: lsst.daf.butler.delegates.arrowastropy.ArrowAstropyDelegate
     derivedComponents:
       columns: ArrowColumnList
@@ -183,6 +186,7 @@ storageClasses:
       pyarrow.Table: lsst.daf.butler.formatters.parquet.arrow_to_numpy
       pandas.core.frame.DataFrame: pandas.DataFrame.to_records
       astropy.table.Table: astropy.table.Table.as_array
+      dict: lsst.daf.butler.formatters.parquet._numpy_dict_to_numpy
     delegate: lsst.daf.butler.delegates.arrownumpy.ArrowNumpyDelegate
     derivedComponents:
       columns: ArrowColumnList
@@ -197,6 +201,21 @@ storageClasses:
       pyarrow.Schema: lsst.daf.butler.formatters.parquet.ArrowNumpySchema.from_arrow
       lsst.daf.butler.formatters.parquet.DataFrameSchema: lsst.daf.butler.formatters.parquet.DataFrameSchema.to_arrow_numpy_schema
       lsst.daf.butler.formatters.parquet.ArrowAstropySchema: lsst.daf.butler.formatters.parquet.ArrowAstropySchema.to_arrow_numpy_schema
+  ArrowNumpyDict:
+    pytype: dict
+    converters:
+      pyarrow.Table: lsst.daf.butler.formatters.parquet.arrow_to_numpy_dict
+      pandas.core.frame.DataFrame: lsst.daf.butler.formatters.parquet._pandas_to_numpy_dict
+      astropy.table.Table: lsst.daf.butler.formatters.parquet._astropy_to_numpy_dict
+      numpy.ndarray: lsst.daf.butler.formatters.parquet._numpy_to_numpy_dict
+    delegate: lsst.daf.butler.delegates.arrownumpydict.ArrowNumpyDictDelegate
+    derivedComponents:
+      columns: ArrowColumnList
+      rowcount: int
+      schema: ArrowNumpySchema
+    parameters:
+      - columns
+      - filters
   SkyMap:
     pytype: lsst.skymap.BaseSkyMap
   PropertySet:
--- a/python/lsst/daf/butler/formatters/parquet.py
+++ b/python/lsst/daf/butler/formatters/parquet.py
@@ -136,6 +136,13 @@ class ParquetFormatter(Formatter):
             arrow_table = astropy_to_arrow(inMemoryDataset)
         elif isinstance(inMemoryDataset, np.ndarray):
             arrow_table = numpy_to_arrow(inMemoryDataset)
+        elif isinstance(inMemoryDataset, dict):
+            try:
+                arrow_table = numpy_dict_to_arrow(inMemoryDataset)
+            except (TypeError, AttributeError) as e:
+                raise ValueError(
+                    "Input dict for inMemoryDataset does not appear to be a dict of numpy arrays."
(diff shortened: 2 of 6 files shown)


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -163,6 +165,8 @@ storageClasses:
       pyarrow.Table: lsst.daf.butler.formatters.parquet.arrow_to_astropy
       pandas.core.frame.DataFrame: lsst.daf.butler.formatters.parquet.pandas_to_astropy
       numpy.ndarray: lsst.daf.butler.formatters.parquet.numpy_to_astropy
+      # dict: lsst.daf.butler.formatters.parquet._numpy_dict_to_astropy
```
cmsaunders: Was this supposed to be removed?

python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -163,6 +165,8 @@ storageClasses:
...
+      dict: astropy.table.Table
```
cmsaunders, line 168: Does this need to be `astropy.table.Table.<some converting function>`, or does it work as is because `astropy.table.Table(dict)` is a way to construct an astropy Table?

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -136,6 +136,8 @@ def write(self, inMemoryDataset: Any) -> None:
             arrow_table = astropy_to_arrow(inMemoryDataset)
         elif isinstance(inMemoryDataset, np.ndarray):
             arrow_table = numpy_to_arrow(inMemoryDataset)
+        elif isinstance(inMemoryDataset, dict):
+            arrow_table = numpy_dict_to_arrow(inMemoryDataset)
```
cmsaunders: Do you need to do anything to check that this is a dict of numpy arrays, not just a generic dict?

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -290,6 +292,38 @@ def arrow_to_numpy_dict(arrow_table: pa.Table) -> dict[str, np.ndarray]:
...
+    ----------
+    np_array : `numpy.ndarray`
+        Input numpy array with multiple fields.
+
+    Returns
+    -------
+    numpy_dict : `dict` [`str`, `numpy.ndarray`]
+        Dict with keys as the column names, values as the arrays.
+    """
+    return arrow_to_numpy_dict(numpy_to_arrow(np_array))
```
cmsaunders, line 329: I get that you want to recycle existing functions as much as possible, but are you sure there's no lost information here? Are there no data-type conversion issues?

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -368,6 +402,22 @@ def numpy_dict_to_arrow(numpy_dict: dict[str, np.ndarray]) -> pa.Table:
     return arrow_table
 
 
+def _DONOTUSE_numpy_dict_to_pandas(numpy_dict: dict[str, np.ndarray]) -> pd.DataFrame:
```
cmsaunders: Can you add some documentation to explain this?

tests/test_parquet.py
```diff
@@ -1511,5 +1572,142 @@ def testStorageClass(self):
...
+        dict1 = _numpy_to_numpy_dict(tab1)
+
+        self.butler.put(dict1, self.datasetType, dataId={})
+        # Read the whole table.
+        dict2 = self.butler.get(self.datasetType, dataId={})
+        self._checkNumpyDictEquality(dict1, dict2)
+        # Read the columns.
+        columns2 = self.butler.get(self.datasetType.componentTypeName("columns"), dataId={})
+        self.assertEqual(len(columns2), len(dict1.keys()))
+        for i, name in enumerate(dict1.keys()):
```
cmsaunders, line 1608: Are these actually guaranteed to be in the same order?

tests/test_parquet.py
```diff
@@ -1511,5 +1572,142 @@ def testStorageClass(self):
...
+        self._checkNumpyDictEquality(subdict, tab3)
+        tab4 = self.butler.get(self.datasetType, dataId={}, parameters={"columns": "a"})
+        subdict = {key: dict1[key] for key in ["a"]}
+        self._checkNumpyDictEquality(subdict, tab4)
+        tab5 = self.butler.get(self.datasetType, dataId={}, parameters={"columns": ["index", "a"]})
+        subdict = {key: dict1[key] for key in ["index", "a"]}
+        self._checkNumpyDictEquality(subdict, tab5)
+        tab6 = self.butler.get(self.datasetType, dataId={}, parameters={"columns": "ddd"})
+        subdict = {key: dict1[key] for key in ["ddd"]}
+        self._checkNumpyDictEquality(subdict, tab6)
```
cmsaunders, line 1628: Is this actually different from the `tab4` check?

tests/test_parquet.py
```diff
@@ -1511,5 +1572,142 @@ def testStorageClass(self):
...
+            self.assertTrue(np.all(dict1[name] == dict2[name]))
+
+
+@unittest.skipUnless(np is not None, "Cannot test ParquetFormatterArrowNumpy without numpy.")
+@unittest.skipUnless(pa is not None, "Cannot test ParquetFormatterArrowNumpy without pyarrow.")
+class InMemoryNumpyDictDelegateTestCase(ParquetFormatterArrowNumpyDictTestCase):
+    """Tests for InMemoryDatastore, using ArrowNumpyDictDelegate."""
+
+    configFile = os.path.join(TESTDIR, "config/basic/butler-inmemory.yaml")
+
```
cmsaunders, line 1723: Any testing needed here? It looks like the other parallel classes all have their own tests.

tests/test_parquet.py
```diff
@@ -914,6 +943,18 @@ def testWriteAstropyReadAsNumpyTable(self):
...
 
+    @unittest.skipUnless(np is not None, "Cannot test reading as numpy without numpy.")
+    def testWriteAstropyReadAsNumpyDict(self):
+        tab1 = _makeSimpleAstropyTable()
+        self.butler.put(tab1, self.datasetType, dataId={})
+
+        tab2 = self.butler.get(self.datasetType, dataId={}, storageClass="ArrowNumpyDict")
+
+        # This is tricky because it loses the units.
+        tab2_astropy = atable.Table(tab2)
```
cmsaunders, line 954: Why not use your `_numpy_dict_to_astropy` function here? Or if not, why have that function?

cmsaunders (approved):
This mostly looks good to me, except that the logic is not clear to me regarding which combinations of <datatype>_to_<datatype> are public vs internal vs not implemented. Why for example, is there `astropy_to_arrow`, but `_astropy_to_numpy_dict`? I assume for the not-implemented combinations that there are external versions that are already sufficient.

I tried running my task that needs this, and successfully did a `butlerQC.put(<numpy dict>)` with `ArrowNumpyDict` as the storage class. All worked as expected.

tests/test_parquet.py
timj: This test inherits from the test above but uses the in-memory datastore. You can see from the coverage report that test coverage is pretty good except that `_numpy_dict_to_astropy` is never called.

python/lsst/daf/butler/configs/storageClasses.yaml
erykoff: That's exactly right.  So it just calls the object constructor and it works.

python/lsst/daf/butler/formatters/parquet.py
erykoff: Interesting question.  If it isn't it will crash, but maybe with a less helpful error than the unsupported type `ValueError`.  I'd like to know @timj's thoughts on this.

timj: One downside of using a generic python type is that butler can't help you very much since it's not looking inside it. You could put a try block around this if you wanted to reraise a ValueError...

python/lsst/daf/butler/formatters/parquet.py
erykoff: Well, there aren't any caught by the tests!  😄  But the conversions are totally reversible (except to/from pandas because pandas) so it should be fine.  There's also a memory penalty going from a row-based numpy array to a column-based arrow table OR astropy table OR numpy dict.  But there is no copying of the underlying data when converting between these column-based formats.

python/lsst/daf/butler/formatters/parquet.py
erykoff: That should not have been there, oops.  😞

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -489,6 +571,22 @@ def numpy_to_astropy(np_array: np.ndarray) -> atable.Table:
     return Table(data=np_array, copy=False)
 
 
+def _numpy_dict_to_astropy(numpy_dict: dict[str, np.ndarray]) -> atable.Table:
```
timj: Are you deliberately not calling this function?

tests/test_parquet.py
erykoff: Because I thought I needed that function before I realized it was redundant.  Now removed.

tests/test_parquet.py
erykoff: These tests were put in because of a previous bug where a single column was being read as a string array.  So `"ddd"` checks that it gets the `"ddd"` column back and not the `"d"` column (de-duplicated).  And I just copied that test everywhere even though it may only need to be in one place.

python/lsst/daf/butler/formatters/parquet.py
erykoff: This function should not be here, and is being removed.

PR #801, opened 2023-03-08
DM-38280: Remove integer option for dataset ID

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/datasets/ref.py
+++ b/python/lsst/daf/butler/core/datasets/ref.py
@@ -56,7 +56,7 @@ class SerializedDatasetRef(BaseModel):
     """Simplified model of a `DatasetRef` suitable for serialization."""
 
     # DO NOT change order in the Union, pydantic is sensitive to that!
-    id: Optional[Union[uuid.UUID, PositiveInt]] = None
+    id: uuid.UUID | None = None
     datasetType: Optional[SerializedDatasetType] = None
     dataId: Optional[SerializedDataCoordinate] = None
     run: Optional[StrictStr] = None
@@ -114,9 +114,8 @@ class SerializedDatasetRef(BaseModel):
         return node
 
 
-DatasetId = Union[int, uuid.UUID]
-"""A type-annotation alias for dataset ID which could be either integer or
-UUID.
+DatasetId = uuid.UUID
+"""A type-annotation alias for dataset ID providing typing flexibility.
 """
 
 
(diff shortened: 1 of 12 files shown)


python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
```diff
@@ -549,119 +548,6 @@ def getDataId(self, id: DatasetId) -> DataCoordinate:
...
-
-        for dataId, datasetId in zip(dataIdList, datasetIdList):
-            yield DatasetRef(
-                datasetType=self.datasetType,
-                dataId=dataId,
-                id=datasetId,
-                run=run.name,
-            )
-
-
```
TallJimbo, line 664: It'd be a nice simplification to drop the intermediate base class and merge it with the remaining derived class, but I'm not sure that's worth holding up the ticket.

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -114,9 +114,8 @@ def direct(
         return node
 
 
-DatasetId = Union[int, uuid.UUID]
-"""A type-annotation alias for dataset ID which could be either integer or
-UUID.
+DatasetId = uuid.UUID
+"""A type-annotation alias for dataset ID providing typing flexibility.
```
TallJimbo, line 118: Someday I'd like to get rid of this type alias entirely, as I'm pretty confident we'll always use UUID and `DatasetId` is too similar to `DataId` for my comfort.  But not on this ticket.

python/lsst/daf/butler/core/quantum.py
```diff
@@ -77,7 +77,7 @@ def _reconstructDatasetRef(
 class SerializedQuantum(BaseModel):
     """Simplified model of a `Quantum` suitable for serialization."""
 
-    taskName: str
+    taskName: str | None
```
TallJimbo, line 80: I'm a little worried that something wants to set this to `None`, but that's probably also something to punt.

TallJimbo (approved):

python/lsst/daf/butler/core/quantum.py
timj: The error is:

```
python/lsst/daf/butler/core/quantum.py:380: error: Argument "taskName" to "SerializedQuantum" has incompatible type "Optional[str]"; expected "str"  [arg-type]
```

and it's because `Quantum` is allowed to accept an optional task name.

PR #803, opened 2023-03-13
DM-38205: Add update_exposure_regions method to obscore manager

andy-slac:
This new method will be used by DefineVisitsTask to update existing exposure
records in obscore table. A minor refactoring in spatial plugin interface and
RecordFactory class to support this new method. Unit test updated to check
that it works OK, plus a mypy cleanup in obscore unit test.

Also adds obscore manager property to Registry interface.
Some clients, e.g. DefineVisitsTask, will need access to ObsCore update
methods. I do not want to expose those methods in Registry interface because
they are only relevant for a small set of clients and a small number of repos.
Instead, an `obsCoreTableManager` property is added which will be `None` if
obscore is not configured for a repo.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/doc/changes/DM-38205.misc.md
+++ b/doc/changes/DM-38205.misc.md
@@ -1,0 +1,3 @@
+`Registry` adds an `obsCoreTableManager` property for access to ObsCore table manager.
+This will be set to `None` when repository lacks ObsCore table.
+It should only be used by a limited number of clients, e.g. DefineVisitsTask, which need to update the table.
--- a/doc/changes/DM-38205.misc.md
+++ b/doc/changes/DM-38205.misc.md
@@ -1,3 +1,3 @@
 `Registry` adds an `obsCoreTableManager` property for access to ObsCore table manager.
 This will be set to `None` when repository lacks ObsCore table.
-It should only be used by a limited number of clients, e.g. DefineVisitsTask, which need to update the table.
+It should only be used by a limited number of clients, e.g. `DefineVisitsTask`, which need to update the table.
--- a/python/lsst/daf/butler/registry/interfaces/_obscore.py
+++ b/python/lsst/daf/butler/registry/interfaces/_obscore.py
@@ -26,8 +26,11 @@ from __future__ import annotations
 __all__ = ["ObsCoreTableManager"]
 
 from abc import abstractmethod
-from collections.abc import Iterable, Mapping
-from typing import TYPE_CHECKING, Type
+from collections.abc import Iterable, Iterator, Mapping
+from contextlib import contextmanager
+from typing import TYPE_CHECKING, Any, Type
+
+import sqlalchemy
 
 from ._versioning import VersionedExtension
 
@@ -94,6 +97,7 @@ class ObsCoreTableManager(VersionedExtension):
         """
         raise NotImplementedError()
 
+    @abstractmethod
     def add_datasets(self, refs: Iterable[DatasetRef], context: SqlQueryContext) -> int:
         """Possibly add datasets to the obscore table.
 
@@ -129,6 +133,7 @@ class ObsCoreTableManager(VersionedExtension):
         """
         raise NotImplementedError()
 
+    @abstractmethod
     def associate(
         self, refs: Iterable[DatasetRef], collection: CollectionRecord, context: SqlQueryContext
     ) -> int:
@@ -164,6 +169,7 @@ class ObsCoreTableManager(VersionedExtension):
         """
         raise NotImplementedError()
 
+    @abstractmethod
     def disassociate(self, refs: Iterable[DatasetRef], collection: CollectionRecord) -> int:
         """Possibly remove datasets from the obscore table.
 
@@ -193,6 +199,7 @@ class ObsCoreTableManager(VersionedExtension):
         """
         raise NotImplementedError()
 
+    @abstractmethod
     def update_exposure_regions(self, instrument: str, region_data: Iterable[tuple[int, int, Region]]) -> int:
         """Update existing exposure records with spatial region data.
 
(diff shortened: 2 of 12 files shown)


doc/changes/DM-38205.misc.md
```diff
@@ -0,0 +1,3 @@
+`Registry` adds an `obsCoreTableManager` property for access to ObsCore table manager.
+This will be set to `None` when repository lacks ObsCore table.
+It should only be used by a limited number of clients, e.g. DefineVisitsTask, which need to update the table.
```
timj:
```suggestion
It should only be used by a limited number of clients, e.g. `DefineVisitsTask`, which need to update the table.
```

python/lsst/daf/butler/registry/interfaces/_obscore.py
```diff
@@ -190,3 +198,54 @@ def disassociate(self, refs: Iterable[DatasetRef], collection: CollectionRecord)
...
+        Returns
+        -------
+        count : `int`
+            Actual number of records updated.
+
+        Notes
+        -----
+        This method is needed to update obscore records for raw exposures which
+        are ingested before their corresponding visits are defined. Exposure
+        records added when visit are already defined will get their regions
```
timj:
```suggestion
        records added when visit is already defined will get their regions
```

python/lsst/daf/butler/registry/interfaces/_obscore.py
```diff
@@ -190,3 +198,54 @@ def disassociate(self, refs: Iterable[DatasetRef], collection: CollectionRecord)
...
+        from their matching visits automatically.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    @contextmanager
+    def query(self, **kwargs: Any) -> Iterator[sqlalchemy.engine.CursorResult]:
+        """Run a SELECT query against obscore table and return reslut rows.
```
timj:
```suggestion
        """Run a SELECT query against obscore table and return result rows.
```

python/lsst/daf/butler/registry/obscore/_manager.py
```diff
@@ -294,3 +296,62 @@ def _check_dataset_run(self, run: str) -> bool:
...
+            exposure_column: "exposure_column",
+            detector_column: "detector_column",
+        }
+
+        count = self.db.update(self.table, where_dict, *update_rows)
+        return count
+
+    @contextmanager
+    def query(self, **kwargs: Any) -> Iterator[sqlalchemy.engine.CursorResult]:
+        """Run a SELECT query against obscore table and return reslut rows.
```
timj:
```suggestion
        """Run a SELECT query against obscore table and return result rows.
```

python/lsst/daf/butler/registry/obscore/_records.py
```diff
@@ -189,11 +191,15 @@ def __call__(self, ref: DatasetRef, context: SqlQueryContext) -> Optional[Record
...
         # ask each plugin for its values to add to a record.
-        for plugin in self.spatial_plugins:
-            assert ref.id is not None, "Dataset ID must be defined"
-            plugin_record = plugin.make_records(ref.id, region)
-            if plugin_record is not None:
-                record.update(plugin_record)
+        try:
+            plugin_records = self.make_spatial_records(region)
+        except RegionTypeError as exc:
+            warnings.warn(
```
timj, line 197: Philosophically do we want this to be a warning or a fatal/critical log message? (also above)

tests/test_obscore.py
```diff
@@ -357,80 +363,150 @@ def test_insert_existing_collection(self):
...
+
+        registry = self.make_registry(["run1"])
+        obscore = registry.obsCoreTableManager
+        assert obscore is not None
+
+        # Exposure 4 is not associated with any visit.
+        for detector in (1, 2, 3, 4):
+            self._insert_dataset(registry, "run1", "raw", detector=detector, exposure=4)
+
+        # All spatial columns should be None
```
timj:
```suggestion
        # All spatial columns should be None.
```

tests/test_obscore.py
```diff
@@ -357,80 +363,150 @@ def test_insert_existing_collection(self):
...
+        with obscore.query() as result:
+            rows = list(result)
+            self.assertEqual(len(rows), 4)
+            for row in rows:
+                self.assertIsNone(row.s_ra)
+                self.assertIsNone(row.s_dec)
+                self.assertIsNone(row.s_region)
+
+        # Assign Region from visit 4
```
timj:
```suggestion
        # Assign Region from visit 4.
```

tests/test_obscore.py
```diff
@@ -357,80 +363,150 @@ def test_insert_existing_collection(self):
...
+                if row.lsst_detector in (1, 2):
+                    self.assertIsNotNone(row.s_ra)
+                    self.assertIsNotNone(row.s_dec)
+                    self.assertIsNotNone(row.s_region)
+                else:
+                    self.assertIsNone(row.s_ra)
+                    self.assertIsNone(row.s_dec)
+                    self.assertIsNone(row.s_region)
+
+    if TYPE_CHECKING:
```
timj, line 495: I didn't even know we were running mypy on the tests directory.

timj (approved):

python/lsst/daf/butler/registry/obscore/_records.py
andy-slac: I prefer warning just in case in the future someone wants to define visits with odd region types, they could at least suppress the messages. Fatal message would probably mean to some people that things are failing badly, while we are only ignoring that region data, but we fill the rest of the obscore record correctly. And I hope we'll never actually see this message.

tests/test_obscore.py
andy-slac: We don't :slightly_smiling_face: I just run it manually on this file. It would be nice to run mypy on everything, but many files in tests/ are not ready for that.

PR #805, opened 2023-03-21
DM-37339: Fix typing on Config class

timj:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/config.py
+++ b/python/lsst/daf/butler/core/config.py
@@ -35,7 +35,7 @@ import sys
 from collections import defaultdict
 from collections.abc import Iterable, Mapping, MutableMapping, Sequence
 from pathlib import Path
-from typing import IO, TYPE_CHECKING, Any, ClassVar
+from typing import IO, TYPE_CHECKING, Any, ClassVar, Iterator
 
 import yaml
 from lsst.resources import ResourcePath, ResourcePathExpression
@@ -62,7 +62,7 @@ else:
         yamlLoader = yaml.SafeLoader
 
 
-def _doUpdate(d, u):
+def _doUpdate(d: Mapping[str, Any], u: Mapping[str, Any]) -> Mapping[str, Any]:
     if not isinstance(u, Mapping) or not isinstance(d, MutableMapping):
         raise RuntimeError("Only call update with Mapping, not {}".format(type(d)))
     for k, v in u.items():
@@ -73,7 +73,7 @@ def _doUpdate(d, u):
     return d
 
 
-def _checkNextItem(k, d, create, must_be_dict):
+def _checkNextItem(k: str, d: Any, create: bool, must_be_dict: bool) -> tuple[Any, bool]:
     """See if k is in d and if it is return the new child."""
     nextVal = None
     isThere = False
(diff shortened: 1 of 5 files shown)


python/lsst/daf/butler/core/config.py
```diff
@@ -458,17 +464,11 @@ def _processExplicitIncludes(self) -> None:
...
-                                log.warning(
-                                    "Do not understand search path entry '%s' of type %s",
-                                    dir,
-                                    type(dir).__name__,
-                                )
+                            specific = dir.join(fileName.path)
+                            # Remote resource check might be expensive
+                            if specific.exists():
+                                found = specific
+                                # TODO: do we need `break` here?
```
andy-slac: @timj, I think that we might want a `break` here, don't we?

timj: Yes, looks like it should break out the loop once it has found something in the search path...

python/lsst/daf/butler/core/config.py
```diff
@@ -540,16 +540,15 @@ def _splitIntoKeys(key):
...
         elif isinstance(key, Iterable):
-            return list(key)
+            return [k for k in key]
         else:
-            # Not sure what this is so try it anyway
-            return [
-                key,
-            ]
+            # Do not try to guess.
+            raise TypeError(f"Provided key [{key}] neither str nor iterable.")
```
timj, line 549: I'm assuming this is okay. mypy didn't like the fallback code and so it was easier for me to change it to a raise. If jenkins passes it's likely okay.

python/lsst/daf/butler/core/config.py
```diff
@@ -540,16 +540,15 @@ def _splitIntoKeys(key):
             hierarchy = key.split(d)
             if temp:
                 hierarchy = [h.replace(temp, d) for h in hierarchy]
-            return hierarchy
+            # Copy the list to keep mypy quiet.
+            return list(hierarchy)
         elif isinstance(key, Iterable):
-            return list(key)
+            return [k for k in key]
```
timj: I was unsure about this one as well. It may be that it can go back to being `list(key)` (which I assume is more efficient?)

python/lsst/daf/butler/core/config.py
```diff
@@ -564,15 +563,14 @@ def _getKeyHierarchy(self, name):
             as a key in the Config it will be used regardless of the presence
             of any nominal delimiter.
         """
+        keys: list[str | int]
         if name in self._data:
-            keys = [
-                name,
-            ]
+            keys = [cast(str, name)]
```
timj, line 568: This is probably the right thing to do but in theory `name` could be a tuple -- the trick is that if `in` works directly there is no reason to try anything clever.

python/lsst/daf/butler/core/config.py
```diff
@@ -623,8 +621,9 @@ def __getitem__(self, name):
         # all further cleverness.
         found_directly = False
         try:
-            data = self._data[name]
-            found_directly = True
+            if isinstance(name, str):
```
timj, line 624: It does seem painful to have to do an `isinstance` check even though the dict lookup would fail if the item wasn't a string but that's the price we pay for mypy. Especially when this short circuit was there as an optimization to let the fast thing be fast.

timj (commented):
Thanks. I can't approve it because I made the PR...

python/lsst/daf/butler/core/config.py
andy-slac: OK, I'll add a `break` and re-run Jenkins.

python/lsst/daf/butler/core/config.py
andy-slac: `list` is likely more efficient, and it should work OK, I'll put it back.

python/lsst/daf/butler/core/config.py
andy-slac: But `self._data` keys are strings, right? Or at least this is how it is annotated.

python/lsst/daf/butler/core/config.py
andy-slac: Well, it is still an optimization, but for non-string case. I think checking the type first should be as quick or quicker than checking the non-string key in a dict.

python/lsst/daf/butler/core/config.py
timj: Okay. Yes, `name` is anything but in this case we are looking in the internal dict and so yes it's only going to work if it's a string.

PR #806, opened 2023-03-22
DM-38235: Remove unused code related to schema digest calculation

andy-slac:
Schema digests will no longer be written to butler_attributes table when registry database is initialized. The version of the DefaultButlerAttributeManager switched to 1.0.1, which is fully compatible with previous 1.0.0. daf_butler_migrate will add a migration script for 1.0.1.

## Checklist

- [X] ran Jenkins


timj (approved):
Can you add a "removal" news fragment for this please?

PR #807, opened 2023-03-22
DM-38409: Remove support for int dataset IDs in Butler.transfer_from

timj:
This significantly cleans up the code since there is no longer a need to collect the source and target refs separately and make Datastore.transfer_from have to keep track of them. With UUIDs the source ref and the target ref are the same.

## Checklist

- [x] ran Jenkins


andy-slac (approved):
Looks good. Maybe add a news fragment?

timj:
> Maybe add a news fragment?

Yes. Meant to do that and then got distracted.

PR #808, opened 2023-03-23
DM-38402: Move projections to SQL when dropping Python-side region filtering.

TallJimbo:
## Checklist

- [x] ran Jenkins


andy-slac (approved):
Looks good. Codecov seems to be unhappy that new code is not tested, is it difficult to add a unit test for this?

TallJimbo:
A test would have been hard to write without poking a bit at internal attributes to get at the relation tree, but with that it wasn't hard at all.

PR #809, opened 2023-03-27
DM-38447: Fix config file used in chained datastore transfer test

timj:
We need the butler config not just the datastore config because we need to ensure that the storage classes used for the test are defined.

## Checklist

- [ ] ran Jenkins


andy-slac (approved):
Looks good.

PR #810, opened 2023-03-29
DM-38412: Improve schema version handling in Registry

andy-slac:
This commit enables support of multiple schema versions by a single manager. Managers now have an access to the exact schema version as defined by the registry. Added unit test to check new functionality.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/interfaces/_attributes.py
+++ b/python/lsst/daf/butler/registry/interfaces/_attributes.py
@@ -28,7 +28,7 @@ __all__ = [
 from abc import abstractmethod
 from typing import TYPE_CHECKING, Iterable, Optional, Tuple
 
-from ._versioning import VersionedExtension
+from ._versioning import VersionedExtension, VersionTuple
 
 if TYPE_CHECKING:
     from ._database import Database, StaticTablesContext
@@ -58,9 +58,14 @@ class ButlerAttributeManager(VersionedExtension):
     database table with a stable schema.
     """
 
+    def __init__(self, *, registry_schema_version: VersionTuple | None = None) -> None:
+        super().__init__(registry_schema_version=registry_schema_version)
+
     @classmethod
     @abstractmethod
-    def initialize(cls, db: Database, context: StaticTablesContext) -> ButlerAttributeManager:
+    def initialize(
+        cls, db: Database, context: StaticTablesContext, registry_schema_version: VersionTuple | None = None
+    ) -> ButlerAttributeManager:
         """Construct an instance of the manager.
 
         Parameters
@@ -71,6 +76,8 @@ class ButlerAttributeManager(VersionedExtension):
             Context object obtained from `Database.declareStaticTables`; used
             to declare any tables that should always be present in a layer
             implemented with this manager.
+        registry_schema_version : `VersionTuple` or `None`
+            Schema version of this extension as defined in registry.
 
         Returns
         -------
--- a/python/lsst/daf/butler/registry/interfaces/_versioning.py
+++ b/python/lsst/daf/butler/registry/interfaces/_versioning.py
@@ -22,12 +22,21 @@
 from __future__ import annotations
 
 __all__ = [
+    "IncompatibleVersionError",
     "VersionTuple",
     "VersionedExtension",
 ]
 
 from abc import ABC, abstractmethod
-from typing import NamedTuple, Optional
+from typing import NamedTuple
+
+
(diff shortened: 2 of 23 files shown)


python/lsst/daf/butler/registry/interfaces/_attributes.py
```diff
@@ -58,9 +58,14 @@ class ButlerAttributeManager(VersionedExtension):
     database table with a stable schema.
     """
 
+    def __init__(self, *, registry_schema_version: VersionTuple | None = None) -> None:
+        super().__init__(registry_schema_version=registry_schema_version)
```
timj, line 62: If this just calls `super()` why is the method needed at all? Are you worried that someone will add their own `__init__` later and forget they are supposed to call the parent?

python/lsst/daf/butler/registry/interfaces/_versioning.py
```diff
@@ -108,3 +155,114 @@ def extensionName(cls) -> str:
...
+
+        Returns
+        -------
+        version : `VersionTuple`, `list` [`VersionTuple`], or `None`
+            Schema version for this extension. `None` is returned if an
+            extension does not require its version to be saved or checked.
+            Extensions that support multiple schema versions return the list
+            of versions.
+        """
+        raise NotImplementedError()
```
timj, line 170: I am reminded that we can teach the coverage reporting not to worry about lines like this. Could you copy over the content from 

https://github.com/lsst-dm/cm-service/blob/main/pyproject.toml#L46-L58

(although not the debug bits or the `__repr__` ignore since testing should enable debug logging and it's not a hardship to test `repr()`)? Type checking and not implemented error seem like good ones to include.

python/lsst/daf/butler/registry/interfaces/_versioning.py
```diff
@@ -108,3 +155,114 @@ def extensionName(cls) -> str:
...
+        support different compatibility model will overwrite this method.
+        """
+        # Extensions that do not define current versions are compatible with
+        # anything.
+        if (my_version := cls.currentVersion()) is not None:
+            if isinstance(my_version, VersionTuple):
+                my_version = [my_version]
+            # If there are multiple version supported one of them should be
+            # compatible to succeed.
+            for version in my_version:
```
timj: It's a shame `ensure_iterable` won't work because `VersionTuple` is already something that can iterate.

python/lsst/daf/butler/registry/interfaces/_versioning.py
```diff
@@ -108,3 +155,114 @@ def extensionName(cls) -> str:
...
+        Notes
+        -----
+        Default implementation uses `VersionTuple.checkCompatibility` on
+        the version(s) returned from `currentVersion` method. Subclasses that
+        support different compatibility model will overwrite this method.
+        """
+        # Extensions that do not define current versions are compatible with
+        # anything.
+        if (my_version := cls.currentVersion()) is not None:
+            if isinstance(my_version, VersionTuple):
```
timj: are we sure that `currentVersion` can't always return a list so that we could remove this check?

python/lsst/daf/butler/registry/managers.py
```diff
@@ -197,6 +208,13 @@ def makeRepo(self, database: Database, dimensionConfig: DimensionConfig) -> Regi
             Struct containing instances of the types contained by ``self``,
             pointing to the new repository and backed by ``database``.
         """
+        # If schema versions were specified in the config, check that they are
+        # compatible with their mananagers.
```
timj:
```suggestion
        # compatible with their managers.
```

timj (approved):
Looks good

python/lsst/daf/butler/registry/interfaces/_attributes.py
andy-slac: Right, my worry was that someone later needs to pass a parameter to a constructor and adds the `__init__` but forgets about this optional keyword argument.

python/lsst/daf/butler/registry/interfaces/_versioning.py
andy-slac: Well, it certainly can, I just did not want to touch all existing implementations of `currentVersion`. Probably worth doing that though, to avoid checking for type (and then rename it to `currentVersions`).

andy-slac: I changed its name to `currentVersions` and it now returns `list[VersionTuple]`.

PR #812, opened 2023-04-03
DM-37322: Add support for nanosecond ingest_date to datasets manager

andy-slac:
Schema version management in daf_butler is extended to support multiple versions per manager type.
Datasets manager class now supports both `TIMESTAMP` and nanosecond integer (for `astropy.time.Time`) as `ingest_date` column type.
Default `ingest_date` for new registries is still `TIMESTAMP`, it can be changed by specifying `schema_version` in `butler.yaml` for datasets manager.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
+++ b/python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
@@ -162,7 +162,7 @@ class ByDimensionsDatasetRecordStorageManagerBase(DatasetRecordStorageManager):
         universe : `DimensionUniverse`
             Universe graph containing all dimensions known to this `Registry`.
         schema_version : `VersionTuple` or `None`
-            Version of the schema that should be crteated, if `None` then
+            Version of the schema that should be created, if `None` then
             default schema should be used.
 
         Returns
--- a/python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
+++ b/python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
@@ -39,6 +39,10 @@ if TYPE_CHECKING:
 
 # This has to be updated on every schema change
 _VERSION_UUID = VersionTuple(1, 0, 0)
+# Starting with 2.0.0 the `ingest_date` column type uses nanoseconds instead
+# of TIMESTAMP. The code supports both 1.0.0 and 2.0.0 for the duration of
+# client migration period.
+_VERSION_UUID_NS = VersionTuple(2, 0, 0)
 
 _LOG = logging.getLogger(__name__)
 
@@ -115,7 +119,9 @@ class ByDimensionsDatasetRecordStorageManagerBase(DatasetRecordStorageManager):
         registry_schema_version: VersionTuple | None = None,
     ) -> DatasetRecordStorageManager:
         # Docstring inherited from DatasetRecordStorageManager.
-        specs = cls.makeStaticTableSpecs(type(collections), universe=dimensions.universe)
+        specs = cls.makeStaticTableSpecs(
+            type(collections), universe=dimensions.universe, schema_version=registry_schema_version
+        )
         static: StaticDatasetTablesTuple = context.addTableTuple(specs)  # type: ignore
         summaries = CollectionSummaryManager.initialize(
             db,
@@ -135,11 +141,14 @@ class ByDimensionsDatasetRecordStorageManagerBase(DatasetRecordStorageManager):
     @classmethod
     def currentVersions(cls) -> list[VersionTuple]:
         # Docstring inherited from VersionedExtension.
-        return [cls._version]
+        return cls._versions
 
     @classmethod
     def makeStaticTableSpecs(
-        cls, collections: type[CollectionManager], universe: DimensionUniverse
+        cls,
+        collections: type[CollectionManager],
+        universe: DimensionUniverse,
+        schema_version: VersionTuple | None,
     ) -> StaticDatasetTablesTuple:
         """Construct all static tables used by the classes in this package.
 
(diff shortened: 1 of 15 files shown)


python/lsst/daf/butler/registry/interfaces/_versioning.py
```diff
@@ -180,19 +180,62 @@ def newSchemaVersion(self) -> VersionTuple | None:
...
+        Returns
+        -------
+        version : `VersionTuple` or `None`
+            Schema version created by this extension. `None` is returned if an
+            extension does not require its version to be saved or checked.
+
+        Notes
+        -----
+        Default implementation of this method can work in simple cases. If
+        the extension only supports single schema version than that versions is
```
TallJimbo:
```suggestion
        the extension only supports single schema version than that version is
```

python/lsst/daf/butler/registry/interfaces/_versioning.py
```diff
@@ -180,19 +180,62 @@ def newSchemaVersion(self) -> VersionTuple | None:
 
         Notes
         -----
-        Default implementation only forks for extensions that support single
-        schema version and it returns version obtained from `currentVersions`.
-        If `currentVersions` returns multiple version then default
-        implementation will raise an exception and the method has to be
-        reimplemented by a subclass.
+        Extension classes which support multiple schema versions need to
```
TallJimbo:
```suggestion
        Extension classes that support multiple schema versions need to
```

python/lsst/daf/butler/registry/interfaces/_versioning.py
```diff
@@ -180,19 +180,62 @@ def newSchemaVersion(self) -> VersionTuple | None:
...
+        override `_newDefaultSchemaVersion` method.
         """
-        my_versions = self.currentVersions()
+        return self.clsNewSchemaVersion(self._registry_schema_version)
+
+    @classmethod
+    def clsNewSchemaVersion(cls, schema_version: VersionTuple | None) -> VersionTuple | None:
+        """Class method which returns schema version for new registry.
```
TallJimbo: Could you clarify whether this means a new registry database vs. just a new registry client interface?

python/lsst/daf/butler/registry/interfaces/_versioning.py
```diff
@@ -180,19 +180,62 @@ def newSchemaVersion(self) -> VersionTuple | None:
...
+        method which needs to be reimplemented in a subsclass.
+        """
+        my_versions = cls.currentVersions()
         if not my_versions:
             return None
         elif len(my_versions) == 1:
             return my_versions[0]
+        else:
+            if schema_version is not None:
+                assert schema_version in my_versions, "Schema version must be compatible."
```
TallJimbo, line 223: Do we get to this assertion if somebody tries to load a Butler whose software version is incompatible with the repo version?  If so, it probably shouldn't be an assert.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -152,14 +161,23 @@ def makeStaticTableSpecs(
             Manager object for the collections in this `Registry`.
         universe : `DimensionUniverse`
             Universe graph containing all dimensions known to this `Registry`.
+        schema_version : `VersionTuple` or `None`
+            Version of the schema that should be crteated, if `None` then
```
TallJimbo:
```suggestion
            Version of the schema that should be created, if `None` then
```

python/lsst/daf/butler/registry/queries/expressions/_predicate.py
```diff
@@ -241,6 +247,23 @@ def visitBinaryOp(
             ]:
                 rhs = ColumnLiteral(self.to_datetime(rhs.value), datetime.datetime)
                 return lhs.predicate_method(self.OPERATOR_MAP[operator], rhs)
+            # Allow comparisons between astropy.time.Time expressions and
+            # datetime literals/binds, by coercing the
+            # datetime literals to astropy.time.Time (in UTC scale).
```
TallJimbo, line 252: Someday I'd like to deprecate this new code and some of the old code around here and say we only support `astropy.time.Time`. I'm not in a rush to do that now, but we should keep it in mind as we make RFCs to transition the usage of TIMESTAMP in the schema.

python/lsst/daf/butler/registry/queries/expressions/_predicate.py
```diff
@@ -78,6 +79,7 @@ def make_string_expression_predicate(
     string: str,
     dimensions: DimensionGraph,
     *,
+    column_types: ColumnTypeInfo,
```
TallJimbo, line 82: It's a bit of a shame that we have to pass `column_types` here (and through all the classes in `_structs.py`), since `QueryBuilder` already has access to that and that's enough in all other cases.  But I don't see a way out of it, at least not until some distant-future date when we've gotten rid of all `datetime` usage.

TallJimbo (approved):

python/lsst/daf/butler/registry/interfaces/_versioning.py
andy-slac: This method is only used when creating new database, the non-None version that is passed as argument can come from `butler.yaml`, and there is a separate piece of code that verifies that the version is already compatible (and raises a better exception). I just added `assert` here as an additional protection in case someone decides to do something different.

python/lsst/daf/butler/registry/queries/expressions/_predicate.py
andy-slac: I added this only to keep existing unit tests happy, It should indeed disappear once we remove datetime support.

PR #813, opened 2023-04-03
DM-38492: Support URIs in some additional APIs

timj:
It is convenient in some cases if URIs can be supported rather than local files.

## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/_butler.py
```diff
@@ -2037,16 +2044,17 @@ def import_(
...
             A stream or name of file that contains database information
             associated with the exported datasets, typically generated by
-            `~lsst.daf.butler.Butler.export`.  If this a string (name) and
-            is not an absolute path, does not exist in the current working
-            directory, and ``directory`` is not `None`, it is assumed to be in
+            `~lsst.daf.butler.Butler.export`.  If this a string (name) or
+            `~lsst.resources.ResourcePath` and is not an absolute path,
+            does not exist in the current working directory, and ``directory``
+            is not `None`, it is assumed to be in
             ``directory``. Defaults to "export.{format}".
```
andy-slac: Just a general though, not related to the changes - this handling of the `directory` resembles a search path which always includes two entries - `cwd` and `directory`. This may cause confusion if an export file can be found in both places but user really wants it to come from `directory`, as there is no way to remove `cwd` from a search path. Would it make sense to reverse search path to look in `directory` first, or have an option to disable search in `cwd`?

python/lsst/daf/butler/_butler.py
```diff
@@ -2080,12 +2088,16 @@ def import_(
             else:
                 _, format = os.path.splitext(filename)  # type: ignore
         elif filename is None:
-            filename = f"export.{format}"
-        if isinstance(filename, str) and directory is not None and not os.path.exists(filename):
-            filename = os.path.join(directory, filename)
+            filename = ResourcePath(f"export.{format}", forceAbsolute=False)
+        if directory is not None:
+            directory = ResourcePath(directory, forceDirectory=True)
+        if isinstance(filename, (str, ResourcePath, pathlib.Path, urllib.parse.ParseResult)):
```
andy-slac: I guess this is the same as `ResourcePathExpression`, what happens if `ResourcePathExpression` definition changes?

python/lsst/daf/butler/transfers/_context.py
```diff
@@ -60,7 +63,7 @@ class RepoExportContext:
         Datastore to export from.
     backend : `RepoExportBackend`
         Implementation class for a particular export file format.
-    directory : `str`, optional
+    directory : `str` or `~lsst.resources.ResourcePath`, optional
```
andy-slac: Should it be `ResourcePathExpression`?

andy-slac (approved):
Looks good, a couple of minor questions.

python/lsst/daf/butler/_butler.py
timj: Yes. I was worrying about that. I couldn't initially think of a way to avoid the instance check and appease mypy. Unfortunately I can't say `isinstance(filename, ResourcePathExpression)`.

timj: One option is to add a class method to ResourcePath along the lines of ResourcePath.can_accept.  What do you think?

python/lsst/daf/butler/_butler.py
timj: I think @TallJimbo designed this API. I'm happy to change it to prefer the given directory over cwd if he agrees.

python/lsst/daf/butler/_butler.py
andy-slac: Would it work if you define:

    ResourcePathExpression = str | urllib.parse.ParseResult | ResourcePath | Path

(after ResourcePath class is declared)? You can then say `if isinstance(filename, ResourcePathExpression)`.

python/lsst/daf/butler/_butler.py
TallJimbo: I have no problem changing it.  Andy's argument makes sense.

python/lsst/daf/butler/_butler.py
TallJimbo: If Andy's last post works, I'm all for it.  I keep forgetting that `isinstance` now does unions.

timj: A quick test indicates that moving it to the end (so that we no longer need to quote the ResourcePath entry) does indeed work.

python/lsst/daf/butler/_butler.py
timj: I've changed it to use directory if both cwd and directory have the file. I also issue a warning message in that scenario just in case.

andy-slac (approved):
Re-reviewed once again, looks good.

PR #817, opened 2023-04-11
DM-38642: Add support for multi-index dataframes with DataFrameDelegate/InMemoryDatastore

erykoff:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/formatters/parquet.py
+++ b/python/lsst/daf/butler/formatters/parquet.py
@@ -43,7 +43,7 @@ import collections.abc
 import itertools
 import json
 import re
-from typing import TYPE_CHECKING, Any, Iterable, List, Optional, Sequence, Union, cast
+from typing import TYPE_CHECKING, Any, Iterable, List, Optional, Sequence, cast
 
 import pyarrow as pa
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -915,7 +919,7 @@ def _standardize_multi_index_columns(
     pd_index = arrow_schema_to_pandas_index(schema)
     index_level_names = tuple(pd_index.names)
 
-    names = []
+    names: List[Union[str, Sequence[Any]]] = []
```
timj: You can now say:

```python
names: list[str | Sequence[Any]] = []
```
(ie use `|` instead of `Union` and `list` instead of `List`).

timj (approved):
I can confirm that this fixes my problem in lsst/pipe_tasks#758.

I'm not sure I understand the commit message that removes testMultiIndexDataFrame. Was it never meant to be there or something?

erykoff:
I can amend the commit message.  The old test was overriding the default test to check for the correct NotImplemented raises.  Those have been removed because this is now implemented.

PR #819, opened 2023-04-13
DM-38694: Add StorageClassDelegate.copy() method

timj:
This allows for a future where in-memory datastore can decide whether to do copying when returning an object.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/storageClassDelegate.py
+++ b/python/lsst/daf/butler/core/storageClassDelegate.py
@@ -26,9 +26,12 @@ from __future__ import annotations
 __all__ = ("DatasetComponent", "StorageClassDelegate")
 
 import collections.abc
+import copy
 import logging
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Dict, Iterable, Mapping, Optional, Set, Tuple, Type
+
+from lsst.utils.introspection import get_full_type_name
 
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/core/storageClassDelegate.py
```diff
@@ -396,3 +399,38 @@ def selectResponsibleComponent(cls, derivedComponent: str, fromComponents: Set[O
...
+        The default implementation uses `copy.deepcopy()`.
+        It is generally expected that this method is the equivalent of a deep
+        copy. Subclasses can override this method if they already know the
+        optimal approach for deep copying.
+        """
+
+        try:
+            return copy.deepcopy(inMemoryDataset)
+        except Exception as e:
+            raise NotImplementedError(
```
erykoff, line 433: Can we add a test to exercise this failure?

erykoff (changes requested):
I think we want to test the alternate branch (failure) as well as the success branch.

python/lsst/daf/butler/core/storageClassDelegate.py
timj: I meant to do that but I utterly failed to think of a class that failed. Will think again.

erykoff (approved):

PR #820, opened 2023-04-14
DM-31924: deprecate aspects of the data ID packing system

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/dimensions/_config.py
+++ b/python/lsst/daf/butler/core/dimensions/_config.py
@@ -23,6 +23,7 @@ from __future__ import annotations
 
 __all__ = ("DimensionConfig",)
 
+import warnings
 from collections.abc import Iterator, Mapping, Sequence
 from typing import Any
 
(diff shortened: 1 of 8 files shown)


TallJimbo:
Codecov is sad only because it can't tell that the code that lost coverage is now deprecated (and I don't think it's worth keeping the tests alive until the removal happens).

parejkoj (commented):
Is the "remove" ticket already filed? I've found it very handy to mark the methods that need to be removed post-deprecation with `# DM-XXXXX: remove`, especially if their deprecation is implicit, like some of these appear to be.

TallJimbo:
> Is the "remove" ticket already filed?

It is ([DM-38687](https://jira.lsstcorp.org/browse/DM-38687)), but it hadn't been when I made most of these changes.  I've gone back and added references to it.

python/lsst/daf/butler/core/dimensions/_config.py
```diff
@@ -247,13 +248,16 @@ def _extractPackerVisitors(self) -> Iterator[DimensionConstructionVisitor]:
...
         """
-        for name, subconfig in self["packers"].items():
-            yield DimensionPackerConstructionVisitor(
-                name=name,
-                clsName=subconfig["cls"],
-                fixed=subconfig["fixed"],
-                dimensions=subconfig["dimensions"],
-            )
+        with warnings.catch_warnings():
+            # Don't warn when deprecated code calls other deprecated code.
```
parejkoj, line 255: Does this method (`_extractPackerVisitors`) need a `# TODO: Remove this class on DM-38687.` comment?

PR #821, opened 2023-04-14
DM-37703: Deprecate unresolved refs

timj:
Warnings do come from tests -- the test code is testing unresolved refs for a reason. I have not gone through every test to decide if something can be replaced with a resolved ref. I have changed the `DatasetRef` constructor so that it now creates an id if called with just a run.

## Checklist

- [x] ran Jenkins


doc/changes/DM-37703.api.md
```diff
@@ -0,0 +1 @@
+Creating an unresolved dataset reference now issues an ``UnresolvedRefWarning`` and is deprecated.
```
andy-slac: Maybe mention new `id_generation_mode` parameter for `DatasetRef`?

python/lsst/daf/butler/_butler.py
```diff
@@ -2409,7 +2415,11 @@ def validateConfiguration(
             if "instrument" in datasetType.dimensions:
                 for instrument in instruments:
                     datasetRef = DatasetRef(
-                        datasetType, {"instrument": instrument}, conform=False  # type: ignore
+                        datasetType,
+                        {"instrument": instrument},  # type: ignore
+                        conform=False,
+                        run="validate",
+                        id=uuid.uuid4(),
```
andy-slac: Do you need this, I thought IDs are now generated automatically if run is provided?

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -52,6 +68,157 @@ class PositiveInt(ConstrainedInt):
...
+            del s
+            continue
+        if not module.__name__.startswith("lsst.daf.butler"):
+            # 0 will be this function.
+            # 1 will be the caller
+            # and so does not need adjustment.
+            stacklevel = i
+            break
+        # Stack frames sometimes hang around so explicitly delete.
+        del s
```
andy-slac: Maybe move `del s` right after `inspect.getmodule` so you don't need to do it twice?

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -52,6 +68,157 @@ class PositiveInt(ConstrainedInt):
...
+
+    return stacklevel
+
+
+class DatasetIdGenEnum(enum.Enum):
+    """This enum is used to specify dataset ID generation options for
+    ``insert()`` method.
```
andy-slac: I guess `insert()` method was somewhere in the old location, maybe update docstring?

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -172,24 +347,36 @@ def __init__(
...
+                    self.run, self.datasetType, self.dataId, id_generation_mode
+                )
+            else:
+                self.id = None
+                self.run = None
+                warnings.warn(
+                    "Support for creating unresolved refs will soon be removed. Please contact the "
+                    "middleware team for advice on modifying your code to use resolved refs.",
+                    category=UnresolvedRefWarning,
+                    stacklevel=2,  # _find_outside_stacklevel(),
```
andy-slac: Remove comment?

andy-slac (approved):

python/lsst/daf/butler/core/datasets/ref.py
timj: Actually, I'm supposed to remove the "2". I used the 2 for testing daf_butler so I could see whether things were coming from but failed to put it back to use the right code to find the outside level -- that explains why my ci_hsc_gen3 testing is full of reports about using `ref.py` and I was surprised it wasn't telling me the real calling location outside of butler.

python/lsst/daf/butler/core/datasets/ref.py
timj: I probably should fix `utils` as well. Which is where this code came from. We can remove it when we remove the deprecation message.

python/lsst/daf/butler/_butler.py
timj: Correct. After my rewrite I failed to spot that I had added this in a previous commit.

PR #822, opened 2023-04-18
DM-38063: Ensure that all Parquet files are written with row groups.

erykoff:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/formatters/parquet.py
+++ b/python/lsst/daf/butler/formatters/parquet.py
@@ -1274,10 +1274,10 @@ def compute_row_group_size(schema: pa.Schema, target_size: int = TARGET_ROW_GROU
     """Compute approximate row group size for a given arrow schema.
 
     Given a schema, this routine will compute the number of rows in a row group
-    that will yield persisted size on disk of the target size or smaller.  The
-    exact size on disk depends on the compression settings and ratios; typical
-    binary data tables will have around 15-20% compression with the pyarrow
-    default ``snappy`` compression.
+    that targets the persisted size on disk (or smaller).  The exact size on
+    disk depends on the compression settings and ratios; typical binary data
+    tables will have around 15-20% compression with the pyarrow default
+    ``snappy`` compression algorithm.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -1261,3 +1266,57 @@ def _numpy_style_arrays_to_arrow_arrays(
...
+        elif isinstance(t, pa.FixedSizeListType):
+            t_width = t.list_size * t.value_type.bit_width
+        elif t == pa.null():
+            t_width = 0
+        else:
+            t_width = t.bit_width
+
+        bit_width += t_width
+
+    byte_width = bit_width // 8
```
ktlim, line 1326: Round up?  Also, is there a pathological case where `bit_width < 8` (in particular, `== 0`)?

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -1261,3 +1266,57 @@ def _numpy_style_arrays_to_arrow_arrays(
                 raise err
 
     return arrow_arrays
+
+
+def compute_row_group_size(schema: pa.Schema, target_size: int = TARGET_ROW_GROUP_BYTES) -> int:
+    """Compute approximate row group size for a given arrow schema.
+
+    Given a schema, this routine will compute the number of rows in a row group
+    that will yield persisted size on disk of the target size or smaller.  The
```
ktlim: "will yield" is a little strong given things like the guesses and heuristics.

tests/test_parquet.py
```diff
@@ -1708,5 +1710,39 @@ def testWriteNumpyDictBad(self):
...
+            numpyTable.dtype,
+            len(numpyTable),
+            numpyTable,
+            schema,
+        )
+        arrowTable = pa.Table.from_arrays(arrays, schema=schema)
+
+        row_group_size = compute_row_group_size(arrowTable.schema)
+
+        self.assertGreater(row_group_size, 1_000_000)
```
ktlim, line 1735: This doesn't seem like much of a check; is there any way to tighten it?

ktlim (approved):

python/lsst/daf/butler/formatters/parquet.py
erykoff: Great point, I can check for that.  I'm not too worried about rounding, these are all approximate anyway.

tests/test_parquet.py
erykoff: So I was struggling with how to test this except just empirically with the table that I'm already making.  What I didn't want to do was make a 1Gb temporary file just to test that it has more than one row group, since I thought that would be excessive.  But if you have any more stringent test ideas...

PR #823, opened 2023-04-19
DM-38689: Update _standardize_muli_index_columns to be faster for dataframe del…

erykoff:
…egates.

## Checklist

- [x] ran Jenkins


timj (approved):
I don't really follow all of it but it does indeed seem better to use the columns you already knew about.

erykoff:
All of this code is about dealing with the way you can specify multi-index columns of dataframes as tuples, and how this looks slightly different for arrow tables and dataframes.  But I didn't write the original code.
This update here is just making the utility function use a MultiIndex which saves unnecessary overhead when using it in the delegate where we have an explicit dataframe.

PR #824, opened 2023-04-19
DM-38444 Add Sasquatch datastore

natelust:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastores/chainedDatastore.py
+++ b/python/lsst/daf/butler/datastores/chainedDatastore.py
@@ -411,7 +411,9 @@ class ChainedDatastore(Datastore):
         npermanent = 0
         nephemeral = 0
         for datastore, constraints in zip(self.datastores, self.datastoreConstraints):
-            if constraints is not None and not constraints.isAcceptable(ref):
+            if (
+                constraints is not None and not constraints.isAcceptable(ref)
+            ) or not datastore.constraints.isAcceptable(ref):
                 log.debug("Datastore %s skipping put via configuration for ref %s", datastore.name, ref)
                 continue
 
@@ -570,7 +572,11 @@ class ChainedDatastore(Datastore):
             if not missing_refs:
                 break
             for datastore in self.datastores:
-                got_uris = datastore.getManyURIs(missing_refs, p, allow_missing=True)
+                try:
+                    got_uris = datastore.getManyURIs(missing_refs, p, allow_missing=True)
+                except NotImplementedError:
+                    # some datastores may not implement generating URIs
+                    continue
                 missing_refs -= got_uris.keys()
                 uris.update(got_uris)
                 if not missing_refs:
@@ -745,7 +751,11 @@ class ChainedDatastore(Datastore):
                 # caching datastore since using an on-disk local
                 # cache is exactly what we should be doing.
                 continue
-            datastore_refs = {ref for ref in pending if datastore.exists(ref)}
+            try:
+                datastore_refs = {ref for ref in pending if datastore.exists(ref)}
+            except NotImplementedError:
+                # Some datastores may not support retrieving artifacts
+                continue
 
             if datastore_refs:
                 grouped_by_datastore[number] = datastore_refs
--- a/doc/lsst.daf.butler/datastores.rst
+++ b/doc/lsst.daf.butler/datastores.rst
@@ -112,3 +112,20 @@ The only constraint is that all the datasets in registry are associated with at 
 
 `~datastores.chainedDatastore.ChainedDatastore` has a ``datastores`` key that contains a list of datastore configurations that can match the ``datastore`` contents from other datastores.
 Additionally, a `~datastores.chainedDatastore.ChainedDatastore` can also support ``constraints`` definitions.
+
+Sasquatch Datastore
+===================
+
+The `lsst.analysis.tools` package implements a special kind of datastore to facilitate  uploading `~lsst.analysis.tools.interface.MetricMeasurementBundle`\ s to a Sasquatch instance.
+This `~lsst.analysis.tools.interface.datastore.SasquatchDatastore` is currently write only and is meant to aid dispatching `~lsst.analysis.tools.interface.MetricMeasurementBundle`\ s anytime such a dataset is put with the butler.
+Often times this datastore will be used in conjunction with both a `datastores.chainedDatastore.ChainedDatastore` and a `datastores.fileDatastore.FileDatastore`.
+In such a setup, the `~lsst.analysis.tools.interface.MetricMeasurementBundle` will be uploaded to the Sasquatch database, and then persisted to a file based location the butler can retrieve from.
+
+The supported configurations are
+
+**restProxyUrl**
+    The url where a http rest api based kafka proxy to the Sasquatch database can be found.
+**accessToken**
+    An access token that is used to authenticate to the rest api server.
+**namespace**
+    The namespace within a Sasquatch database where metrics will be uploaded.


doc/changes/DM-38444.md
```diff
@@ -0,0 +1 @@
+Introduced a special datastore to upload metric measurements to a Sasquatch instance when a MetricMeasurementBundle is stored with a butler.put call.
```
timj:
```suggestion
Introduced a special datastore to upload metric measurements to a Sasquatch instance when a `MetricMeasurementBundle` is stored with a `butler.put()` call.
```

Also can the full name of the datastore be included so people know where to find it?

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+    def put(self, inMemoryDataset: Any, ref: DatasetRef) -> None:
+        if self.constraints.isAcceptable(ref):
+            try:
+                self._dispatcher.dispatchRef(inMemoryDataset, ref)
+            except SasquatchDispatchPartialFailure:
+                raise DatasetPutError(
+                    "One or more records may have failed to upload, or only partially succeeded"
+                )
+        else:
+            log.debug("Could not put dataset type %s with Sasquatch datastore", ref.datasetType)
```
timj: This is supposed to raise `DatasetTypeNotSupportedError`. That is the convention when a dataset fails the constraints.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+    def bridge(self) -> DatastoreRegistryBridge:
+        return self._bridge
+
+    def put(self, inMemoryDataset: Any, ref: DatasetRef) -> None:
+        if self.constraints.isAcceptable(ref):
+            try:
+                self._dispatcher.dispatchRef(inMemoryDataset, ref)
+            except SasquatchDispatchPartialFailure:
+                raise DatasetPutError(
+                    "One or more records may have failed to upload, or only partially succeeded"
```
timj: In a chained datastore this message is going to be swallowed. Do you want a log entry to be written somewhere? Should you include the sasquatch error text? Should you raise `from e`? (I'm looking forward to python 3.11 where we can add this message to the other exception)

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+    def __init__(
+        self,
+        config: Config | str,
+        bridgeManager: DatastoreRegistryBridgeManager,
+        butlerRoot: str | None = None,
+    ):
+        super().__init__(config, bridgeManager)
+
+        # Name ourselves either using an explicit name or a name
+        # derived from the (unexpanded) root
```
timj:
```suggestion
        # derived from the (unexpanded) root.
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        self.name = self.config.get("name", "{}@{}".format(type(self).__name__, self.config["root"]))
```
timj: I don't see an example yaml configuration in this PR so I'm not sure what `root` means here. Are you always explicitly naming with the `name` or should the URL be used?

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        log.debug("Creating datastore %s", self.name)
+
+        self._bridge = bridgeManager.register(self.name, ephemeral=False)
+
+        self.restProxyUrl = self.config["restProxyUrl"]
+
+        self.accessToken = self.config.get("accessToken", "na")
```
timj: Does "na" mean something magic? Are you assuming the access token will be read from a local `sasquatchDatastore.yaml` (via `$DAF_BUTLER_CONFIG_PATH`) that will merge with this one and have the token in it? (we all agreed the token won't be in the public file didn't we?)

python/lsst/daf/butler/configs/datastores/sasquatchDatastore.yaml
```diff
@@ -0,0 +1,3 @@
+constraints:
```
timj: This file needs to have the `cls` entry included in it. At the very least that will ensure that this reference sasquatch configuration file tells people where to look for the datastore.  Please look at `fileDatastore.yaml` as an example.

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -411,7 +412,9 @@ def put(self, inMemoryDataset: Any, ref: DatasetRef) -> None:
         npermanent = 0
         nephemeral = 0
         for datastore, constraints in zip(self.datastores, self.datastoreConstraints):
-            if constraints is not None and not constraints.isAcceptable(ref):
+            if (
+                constraints is not None and not constraints.isAcceptable(ref)
+            ) or not datastore.constraints.isAcceptable(ref):
```
timj, line 416: This might be a nice short circuit but we do still expect a child datastore to raise `DatasetTypeNotSupportedError` so that if someone uses the datastore directly they get a consistent answer.

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -424,7 +427,7 @@ def put(self, inMemoryDataset: Any, ref: DatasetRef) -> None:
                 nsuccess += 1
                 if not datastore.isEphemeral:
                     isPermanent = True
-            except DatasetTypeNotSupportedError:
+            except (DatasetTypeNotSupportedError, DatasetPutError):
```
timj: We have to be a bit careful here because `DatasetPutError` is documented as being the exception that is raised when a `put` fails and so one would expect that we should change `FileDatastore` to also use it. If we do that though the semantics of `ChainedDatastore.put()` change completely. Previous to this, a failure of any kind from a child datastore that was not due to constraints failure caused the entire `put` to fail. Maybe that was wrong all along and that we should have caught any failure in put and only raised later on if `nsuccess` was 0 (and really if none worked we'd probably want to cache the most recent error and raise that if `nsuccess` was 0 and possibly include the other errors if there was more than one failure). If we are catching multiple exception types we can't raise `DatasetTypeNotSupportedError` on the line below.

If we aren't raising it at all we probably should be issuing some kind of log entry because otherwise it implies that we can be in a situation where nothing gets put to sasquatch and we can only tell by noticing that everything is missing.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+    """Url of the Sasquatch proxy server"""
+
+    token: str
+    """Authentication token used in communicating with the proxy server"""
+
+    namespace: str = "lsst.debug"
+    """The namespace in Sasquatch in which to write the uploaded metrics"""
+
+    def __post_init__(self) -> None:
+        parsedUrl = urllib.parse.urlparse(self.url)
```
timj: You could use `ResourcePath` here.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+    @property
+    def bridge(self) -> DatastoreRegistryBridge:
+        return self._bridge
+
+    def put(self, inMemoryDataset: Any, ref: DatasetRef) -> None:
+        if self.constraints.isAcceptable(ref):
+            try:
+                self._dispatcher.dispatchRef(inMemoryDataset, ref)
+            except SasquatchDispatchPartialFailure:
+                raise DatasetPutError(
```
timj: If you get a sasquatch error that is not due to a partial failure, does that end up being propagated up directly and won't be a `DatasetPutError`?

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+]
+
+
+class SasquatchDispatchPartialFailure(RuntimeError):
+    """This indicates that a Sasquatch dispatch was partially successful."""
+
+    pass
+
+
+def _tag2VersionTime(productStr: str) -> tuple[str, float]:
```
timj: This is a new file. You can use `_tag_to_version_time` and use underscores in the variables.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+    """Determine versions and dates from the string returned from
+    getEnvironmentPackages.
+
+    The genEnvironmentPackages function returns the setup version associated
```
timj:
```suggestion
    The `~lsst.utils.packages.getEnvironmentPackages` function returns the setup version associated
```
(will need to wrap the sentence)

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+    ----------
+    productStr : `str`
+        The product string returned from a lookup on the result of a call to
+        `~lsst.utils.packages.getEnvironmentPackages`.
+
+    Returns
+    -------
+    result : `tuple` of `str`, `datetime.datetime`
+        The first `str` is the version of the package, and the second is the
+        datetime object associates with that released version.
```
timj:
```suggestion
        datetime object associated with that released version.
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+    for tag in tags.split():
+        match tag.split("_"):
+            # Daily tag branch.
+            case ("d", year, month, day):
+                dt = datetime.datetime(year=int(year), month=int(month), day=int(day))
+            # Weekly tag branch.
+            case ("w", year, week):
+                iyear = int(year)
+                iweek = int(week)
+                if (diff := iweek - 52) > 0:
```
timj: Maybe add a comment explaining that we can go all the way to `w_2022_53`. (which was made on 2022-12-29 but refers to 2022-12-28 since it seems like there was a delay in getting it out).

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+                    iyear += (diff // 52) + 1
+                    iweek = diff % 52
+                # Use 3 as the day because we cut releases on Wednesday.
+                dt = datetime.datetime.fromisocalendar(iyear, iweek, 3)
```
timj: I am a tad concerned with our weekly tag calculations. 2020 had 53 weeks but 2022 does not (according to iso) so how do we end up with w_2022_53 which `fromisocalendar` correctly complains about -- it should be impossible regardless of which day of the week we calculated it at the end of the year.

@athornton or @ktlim do you know where we calculate the tag?

If I plug `w_2022_53` into this algorithm I get back 2023-01-04 and not 2022-12-28 (obviously we get the same date as `w_2023_01`. Additionally, the valid ISO tag of `w_2020_53` gives the invalid 2021-01-06 when it should be reporting 2020-12-30. Also, w.2022.01 was issued on 2021-12-30 (and not 2022-01-05) but I'm not sure if that is because somehow the weekly is calculated for the end of the week or something?

It seems to me that the options here are:

* Look at the code that makes the tags and try to reverse engineer it here.
* Abandon the whole idea and generate an integer 202253 (but then hope that it doesn't have a big gap when it displays 202301)
* Change sconsUtils so that we store the git tag date in the `version.py` and then we use that date directly.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+    Raises
+    ------
+    ValueError
+        Raised if there are no tags which correspond to dates.
+    """
+    times: list[datetime.datetime] = []
+    version = productStr.split()[0]
+    tags: str = re.findall("[(](.*)[)]", productStr)[0]
+    for tag in tags.split():
+        match tag.split("_"):
```
timj: Can we also try to support `w.2022.12` style git tags as well as EUPS tags?

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+            self._populateClusterId()
+        return cast(str, self._cluster_id)
+
+    def _populateClusterId(self) -> None:
+        """Get Sasquatch kafka cluster ID."""
+
+        headers = {"content-type": "application/json"}
+        r = requests.get(f"{self.url}/v3/clusters", headers=headers)
+
+        if r.status_code == 200:
```
timj:
```suggestion
        if r.status_code == requests.codes.ok:
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        }
+
+        r = requests.post(
+            f"{self.url}/v3/clusters/{self.clusterId}/topics", json=topic_config, headers=headers
+        )
+
+        if r.status_code == 201:
+            log.debug("Created topic %s.%s", self.namespace, topic_name)
+            return True
+        elif r.status_code == 400:
```
timj:
```suggestion
        elif r.status_code == requests.codes.bad_request:
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+            "topic_name": f"{self.namespace}.{topic_name}",
+            "partitions_count": PARTITIONS,
+            "replication_factor": REPLICATION_FACTOR,
+        }
+
+        r = requests.post(
+            f"{self.url}/v3/clusters/{self.clusterId}/topics", json=topic_config, headers=headers
+        )
+
+        if r.status_code == 201:
```
timj:
```suggestion
        if r.status_code == requests.codes.created:
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+
+        return json.dumps(schema), resultsTrimmed
+
+    def _python2Avro(self, value: Any) -> Mapping:
+        """Map python type to avro schema
+
+        Parameters
+        ----------
+        value : `Any`
+            Any python parameter
```
timj:
```suggestion
            Any python parameter.
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+                return {"type": "int", "default": 0}
+            case Sequence():
+                tmp = set()
+                for item in value:
+                    tmp.add(self._python2Avro(item)["type"])
+                if len(tmp) == 0:
+                    return {}
+                if len(tmp) > 1:
+                    log.error(
+                        "Sequence contains mixed types: %s, must be heterogeneous for avro conversion "
```
timj: Do you mean homogeneous here?

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+            case float() | None:
+                return {"type": "float", "default": 0.0}
+            case str():
+                return {"type": "string", "default": ""}
+            case int():
+                return {"type": "int", "default": 0}
+            case Sequence():
+                tmp = set()
+                for item in value:
+                    tmp.add(self._python2Avro(item)["type"])
```
timj:
```python
tmp = {self._python2Avro(item)["type"] for item in value}
```
?

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+                        tmp,
+                    )
+                    return {"error_in_conversion": True}
+                return {"type": "array", "items": tmp.pop()}
+            case _:
+                log.error("Unsupported type %s, skipping record", type(value))
+                return {}
+
+    def _handleReferencePackage(self, meta: MutableMapping, bundle: MetricMeasurementBundle) -> None:
+        """Check to see if there is a reference package, if there is determine
```
timj: This is too long as a summary line.

```python
"""Check to see if there is a reference package.

If there is a reference package determine the datetime associated 
with this reference package. Save the package...
"""
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        Parameters
+        ----------
+        meta : `MutableMapping`
+            A mapping which corresponds to fields which should be encoded in
+            all records.
+        bundle : `MetricMeasurementBundle`
+            The bundled metrics
+        """
+        ref_package, package_version, package_timestamp = "", "", 0.0
+        if hasattr(bundle, "reference_package") and bundle.reference_package:
```
timj:
```python
if ref_package := getattr(bundle, "reference_package", ""):
```
? and then no need for the line below.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        preferred time. It so it sets that time as the main time stamp to be
+        used for this record.
+
+        Parameters
+        ----------
+        meta : `MutableMapping`
+            A mapping which corresponds to fields which should be encoded in
+            all records.
+        bundle : `MetricMeasurementBundle`
+            The bundled metrics
```
timj: `run` parameter is not documented.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+                        log.error("Reference package timestamp is empty, using run_timestamp")
+                        meta["timestamp"] = meta["run_timestamp"]
+                    else:
+                        meta["timestamp"] = meta["reference_package_timestamp"]
+                case "run_timestamp":
+                    meta["timestamp"] = meta["run_timestamp"]
+                case "current_timestamp":
+                    timeStamp = datetime.datetime.now()
+                    meta["timestamp"] = timeStamp.timestamp()
+                case "dataset_timestamp":
```
timj: This is the one where we get the date from the exposure/visit record?

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        ----------
+        meta : `MutableMapping`
+            A mapping which corresponds to fields which should be encoded in
+            all records.
+        bundle : `MetricMeasurementBundle`
+            The bundled metrics
+        """
+        identifier: str = "Generic"
+        if hasattr(bundle, "datasetIdentifier") and bundle.datasetIdentifier is not None:
+            identifier = bundle.datasetIdentifier
```
timj: It's a shame this doesn't work:
```python
identifier = getattr(bundle, "datasetIdentifier", "Generic")
```
since it would all be in one line, but then you would need to make it a bit more annoying to handle the `is not None`

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        datasetIdentifier parameter is specified, that is used instead of
+        anything specified by the bundle.
+
+        This will also add any identifier fields supplied to the meta record
+        mapping.
+
+        Together these values (in addition to the timestamp and topic) should
+        uniquely identify an upload to the Sasquatch system.
+
+        Parameters
```
timj: Missing half the parameters.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+            all records.
+        bundle : `MetricMeasurementBundle`
+            The bundled metrics
+        """
+        identifier: str = "Generic"
+        if hasattr(bundle, "datasetIdentifier") and bundle.datasetIdentifier is not None:
+            identifier = bundle.datasetIdentifier
+
+        # override with any user supplied identifer
+        if datasetIdentifier is not None:
```
timj: If this always overrides the bundle identifier should we reverse the checks?

```python
if datasetIdentifier is not None:
    identifer = datasetIdentifier
elif hasattr(bundle, "datasetIdentifier") and bundle.datasetIdentifier is not None:
    identifier = bundle.datasetIdentifier
else:
    identifier = "Generic"
```
?

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        datasetIdentifier: str | None = None,
+        extraFields: Mapping | None = None,
+    ) -> tuple[Mapping[str, list[Any]], bool]:
+        """Encode all of the inputs into a format that can be sent to the
+        kafka proxy server.
+
+        Parameters
+        ----------
+        bundle : `~lsst.analysis.tools.MetricMeasurementBundle`
+            The bundle containing metric values to upload
```
timj:
```suggestion
            The bundle containing metric values to upload.
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+            is also stored in the butler, this should be the butler run
+            collection the bundle is stored in the butler.
+        datasetType : `str`
+            The dataset type name associated with this MetricMeasurementBundle
+        timestamp : `str` or `None`
+            The timestamp to be associated with the measurements in the ingress
+            database. If this value is None, timestamp will be set by the run
+            time or current time.
+        id : `UUID` or `None`
+            The UUID of the MetricMeasurementBundle within the butler. If None,
```
timj:
```suggestion
            The UUID of the `MetricMeasurementBundle` within the butler. If `None`,
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        Parameters
+        ----------
+        bundle : `~lsst.analysis.tools.MetricMeasurementBundle`
+            The bundle containing metric values to upload
+        run : `str`
+            The run name to associate with these metric values. If this bundle
+            is also stored in the butler, this should be the butler run
+            collection the bundle is stored in the butler.
+        datasetType : `str`
+            The dataset type name associated with this MetricMeasurementBundle
```
timj:
```suggestion
            The dataset type name associated with this `MetricMeasurementBundle`.
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+            dataset type. I.e. the timestamp, and the tag will be unique, and
+            each record will belong to one combination of such.
+        extraFields: `Mapping`
+            Extra mapping keys and values that will be added as fields to the
+            dispatched record.
+
+        Returns
+        -------
+        result : `Mapping` of `str` to `list`
+            A mapping of metric name of list of metric measurement records
```
timj:
```suggestion
            A mapping of metric name of list of metric measurement records.
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        status : `bool`
+            A status bool indicating if some records had to be skipped due to
```
timj:
```suggestion
            A status boolean indicating if some records had to be skipped due to
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        # Record if any records needed skipped
+        resultsTrimmed = False
+
+        # Look at each of the metrics in the bundle (name, values)
+        for metric, measurements in bundle.items():
+            # create a list which will contain the records for each measurement
+            # associated with metric
+            metricRecordList = metricRecords.setdefault(metric, list())
+
+            record: dict[str, Any] = dict()
```
timj: Is it more efficient to do this as:

```python
record = meta.copy()
```
rather than using `update()` immediately?

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+            record.update(meta)
+
+            # loop over each metric measurement within the metric
+            for measurement in measurements:
+                # need to extract any tags, package info, etc
+                notes = dict(measurement.notes.items())
```
timj:
```python
notes = measurement.notes.copy()
```
? But I'm not sure why `notes` exists at all because it's only used in the line below.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+                record["tags"] = notes.get("metric_tags", list())
+
+                # Missing values are replaced by 0 in sasquatch, see RFC-763
```
timj:
```suggestion
                # Missing values are replaced by 0 in sasquatch, see RFC-763.
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+                name = ""
+                value = 0.0
+                match measurement.json:
```
timj: `.json` returns a `dict` and not a `str`? Pydantic has `.dict()` return a dict and `.json()` return the json string. It seems like this `json` property doesn't return JSON and I'm now confused.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+                    case {"metric": name, "value": None}:
+                        pass
+                    case {"metric": name, "value": value}:
+                        if value != value:
```
timj: Should we be using the more explicit:
```python
if math.isnan(value):
```
rather than being this abstract?

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+
+        metricRecords: dict[str, list[Any]] = dict()
+
+        # Record if any records needed skipped
+        resultsTrimmed = False
+
+        # Look at each of the metrics in the bundle (name, values)
+        for metric, measurements in bundle.items():
+            # create a list which will contain the records for each measurement
+            # associated with metric
```
timj:
```suggestion
            # associated with metric.
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+            meta.update(extraFields)
+
+        metricRecords: dict[str, list[Any]] = dict()
+
+        # Record if any records needed skipped
+        resultsTrimmed = False
+
+        # Look at each of the metrics in the bundle (name, values)
+        for metric, measurements in bundle.items():
+            # create a list which will contain the records for each measurement
```
timj:
```suggestion
            # Create a list which will contain the records for each measurement
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        datasetIdentifier: str | None = None,
+        identifierFields: Mapping | None = None,
+        extraFields: Mapping | None = None,
+    ) -> None:
+        """Dispatch a `MetricMeasurementBundle` to Sasquatch.
+
+        Parameters
+        ----------
+        bundle : `~lsst.analysis.tools.MetricMeasurementBundle`
+            The bundle containing metric values to upload
```
timj:
```suggestion
            The bundle containing metric values to upload.
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        run : `str`
+            The run name to associate with these metric values. If this bundle
+            is also stored in the butler, this should be the butler run
+            collection the bundle is stored in the butler. This will be used
+            in generating uniqueness constraints in Sasquatch.
+        datasetType : `str`
+            The dataset type name associated with this MetricMeasurementBundle
```
timj:
```suggestion
            The dataset type name associated with this `MetricMeasurementBundle`.
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        timestamp : `str` or `None`
+            The timestamp to be associated with the measurements in the ingress
+            database. If this value is None, timestamp will be set by the run
+            time or current time.
+        id : `UUID` or `None`
+            The UUID of the MetricMeasurementBundle within the butler. If None,
```
timj:
```suggestion
            The UUID of the `MetricMeasurementBundle` within the Butler. If `None`,
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+            a new random UUID will be generated so that each record in
+            Sasquatch will have a unique value.
+        datasetIdentifier : `str` or None
+            A string which will be used in creating unique identifier tags. If
+            None, a default value will be inserted.
```
timj:
```suggestion
            `None`, a default value will be inserted.
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+            # Generate schemas for each record
+            data["value_schema"], schemaTrimmed = self._generateAvroSchema(metric, recordValue)
+            data["records"] = record
+
+            if schemaTrimmed:
+                partialUpload = True
+
+            r = requests.post(f"{self.url}/topics/{self.namespace}.{metric}", json=data, headers=headers)
+
+            if r.status_code == 200:
```
timj:
```suggestion
            if r.status_code == requests.codes.ok:
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        extraFields: Mapping | None = None,
+        datasetIdentifier: str | None = None,
+    ) -> None:
+        """Dispatch a `MetricMeasurementBundle` to Sasquatch with a known
+        `DatasetRef`.
+
+        Parameters
+        ----------
+        bundle : `~lsst.analysis.tools.MetricMeasurementBundle`
+            The bundle containing metric values to upload
```
timj:
```suggestion
            The bundle containing metric values to upload.
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        ref : `DatasetRef`
+            The `Butler` dataset ref corresponding to the input
+            `MetricMeasurementBundle`.
+        timestamp : `str` or `None`
+            The timestamp to be associated with the measurements in the ingress
+            database. If this value is None, timestamp will be set by the run
+            time or current time.
+        extraFields: `Mapping` or None
```
timj:
```suggestion
        extraFields: `Mapping` or `None`
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+            Extra mapping keys and values that will be added as fields to the
+            dispatched record if not None.
+        datasetIdentifier : `str` or None
```
timj:
```suggestion
        datasetIdentifier : `str` or `None`
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        Raises
+        ------
+        SasquatchDispatchPartialFailure
+            Raised if there were any errors in dispatching a bundle.
+        """
+        # Parse the relevant info out of the dataset ref.
+        serializedRef = ref.to_simple()
+        id = serializedRef.id
+        if serializedRef.run is None:
+            run = ""
```
timj:
```suggestion
            run = "<unknown>"
```
? Since this should be impossible (and soon mypy will tell you this).

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        else:
+            run = serializedRef.run
+        dstype = serializedRef.datasetType
+        datasetType = dstype.name if dstype is not None else ""
```
timj: Is it possible for `SerializedDatasetRef` to have undefined dataset type and dataId?

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+            identifierFields=dataRefMapping,
+            extraFields=extraFields,
+            datasetIdentifier=datasetIdentifier,
+        )
+
+
+class SasquatchDatastore(GenericBaseDatastore):
+    """Basic Datastore for writing to an in Sasquatch instance.
+
+    This Datastore is write only, meaning that it can dispatch data to a
```
timj:
```suggestion
    This Datastore is currently write only, meaning that it can dispatch data to a
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+    sasquatch instance, but at the present can not be used to retrieve values.
```
timj:
```suggestion
    Sasquatch instance, but at the present can not be used to retrieve values.
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+            All the problems are reported in a single exception.
+
+        Notes
+        -----
+        This method is a no-op.
+        """
+        return
+
+    def validateKey(self, lookupKey: LookupKey, entity: DatasetRef | DatasetType | StorageClass) -> None:
+        # Docstring is inherited from base class
```
timj:
```suggestion
        # Docstring is inherited from base class.
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        return
+
+    def getLookupKeys(self) -> set[LookupKey]:
+        # Docstring is inherited from base class
```
timj:
```suggestion
        # Docstring is inherited from base class.
```

python/lsst/daf/butler/datastores/sasquatchDatastore.py
```diff
@@ -0,0 +1,885 @@
...
+        return False
+
+    def import_records(self, data: Mapping[str, DatastoreRecordData]) -> None:
+        # Docstring inherited from the base class.
+        return
+
+    def export_records(self, refs: Iterable[DatasetIdRef]) -> Mapping[str, DatastoreRecordData]:
+        # Docstring inherited from the base class.
+
+        # In-memory Datastore records cannot be exported or imported
```
timj:
```suggestion
        # Sasquatch Datastore records cannot be exported or imported.
```

timj (changes requested):
Thanks for all this. I've made lots of comments and will take another look.

There needs to be some text added to `doc/lsst.daf.butler/datastores.rst` explaining the datastore and, importantly, the configuration options.

timj:
An additional point I've just noticed, you need to make sure that `self.bridge.insert(ref)` is called at some point during your put.  This populates the `dataset_locations` table to indicate that the ref is associated with the named datastore. In theory this can be used externally to find out which datastores know about a given dataset. If you do this you will also have to include a trash method and emptyTrash that moves the row out of dataset_locations even if sasquatch still has the dataset (which is also what the `datastore.forget` method can do).

python/lsst/daf/butler/datastores/sasquatchDatastore.py
natelust: I cribbed this from another datastore prior to fully understanding how it all worked and fit together, now that I do I think I need to re-visit that. I think the URL is reasonably unique

python/lsst/daf/butler/datastores/sasquatchDatastore.py
natelust: There is no magic there. And in fact there is no "there" there. The token does nothing at this point. The token will not be public in the file, and will be specific to each user. The details are still being worked out, but it might even give the ability to scope what namespace a user is allowed to write into. I was thinking either that config path variable, or some other shell expansion if needed.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
timj: @ktlim tells me that the weekly tag is calculated here: https://github.com/lsst-dm/jenkins-dm-jobs/blob/main/pipelines/release/weekly_release_cron.groovy#L27-L31

and somehow that is generating week numbers that are not ISO compliant.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
natelust: Actually I am not sure what to do about this yet, it seems like no matter what we do, some times data is going to be overwritten. I think I might have to translate the extra weeks to some time after week 52 or something

timj: Maybe the simplest hack is to declare that if you see week 53 then the date is 202X-12-31T23:59 or something. It should never roll over to the next year.

natelust: Agreed, except if we end up with 54. I think I am going to increment the date. That will give us 4 extra "weeklies" in case.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
natelust: Can you imagine a case where we would have them mixed?

natelust: I put in a work around.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
timj: You don't have to worry about 54 -- that is impossible. 53 is a valid week -- `fromisocalendar` will accept it, but only for years that actually have 53 weeks.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
natelust: Yes, and working all that out is being saved for a different ticket.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
natelust: I could go with `identifier: str = x if (x := getattr(bundle, 'dataset_identifier', None)) is not None else "Generic"`

python/lsst/daf/butler/datastores/sasquatchDatastore.py
natelust: old iteration of the code, i'll change

python/lsst/daf/butler/datastores/sasquatchDatastore.py
natelust: yes, a `lsst.verify.Measurement` property `.json` returns a python dict that is compatible with `json.dumps`

python/lsst/daf/butler/datastores/sasquatchDatastore.py
natelust: There was some edge case where something was not considering something nan, mixing type implementations or something like that, I cant remember what now though. We can switch to `math` and see if the edge case comes up again, I dont have strong feelings on it, outside this is exactly the definition of nan by the IEEE, and so I think it is a reasonable choice

timj: I do feel that `math.isnan` is more explicit and it is implemented in C so it shouldn't be slower than what you have. If you still want to do the `x != x` code please add a comment -- not every python programmer will understand what you mean there.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
natelust: According to the typing yes, but this handles those cases.

natelust:
Do we want to insert a record into a table about who knows about this data? I would prefer not to at this moment because a) we cant fetch anything from Sasquatch anyway b) we cant do anything about the fact that someone might delete the record on the sasquatch side. Have a table that is wrong seems worse than not having a table entry c) the worst that will happen if someone runs something again is the same value will just be uploaded again.

timj:
Regarding recording that the datastore has accepted it, I'm fine with us punting for now but we do need to address it at some point.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
natelust: I have modified the date algorithm and tested it against the dates as reported on `https://eups.lsst.codes/stack/src/tags/`

python/lsst/daf/butler/datastores/sasquatchDatastore.py
natelust: I changed around how exceptions are being handled

python/lsst/daf/butler/datastores/chainedDatastore.py
natelust: I am uploading a change, lets talk about it when you see it.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
timj: You are right that in PyPI versions you would get versions like `25.2023.1700` for the weekly.

python/lsst/daf/butler/datastores/sasquatchDatastore.py
natelust: Actually this is going to need some discussion

python/lsst/daf/butler/datastores/sasquatchDatastore.py
natelust: I removed DatasetPutError, but  this needs to be part of the larger discussion on Exceptions

python/lsst/daf/butler/datastores/sasquatchDatastore.py
natelust: Personally, I prefer camel case, though part of that may be the muscle memory from the years of doing it.

python/lsst/daf/butler/core/exceptions.py
```diff
@@ -37,3 +37,13 @@ class ValidationError(RuntimeError):
     """Some sort of validation error has occurred."""
 
     pass
+
+
+class DatasetPutError(RuntimeError):
```
timj: I don't think this is used anywhere now so can be deleted.

timj (approved):
Thanks for the clean ups.

PR #829, opened 2023-04-24
DM-38779: Change Butler.ingest() to use resolved DatasetRef

timj:
Previously only unresolved DatasetRef were accepted. Now only resolved DatasetRef can be used.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -75,6 +75,7 @@ from .core import (
     DataCoordinate,
     DataId,
     DataIdValue,
+    DatasetIdFactory,
     DatasetIdGenEnum,
     DatasetRef,
     DatasetRefURIs,
@@ -1862,12 +1863,10 @@ class Butler(LimitedButler):
         """
         if not self.isWriteable():
             raise TypeError("Butler is read-only.")
-        if run is not None:
-            warnings.warn(
-                "All DatasetRefs to be ingested must have resolved dataset IDs. "
-                "The run from the ref will be used for ingest and the provided run will be ignored."
-            )
+
         log.verbose("Ingesting %d file dataset%s.", len(datasets), "" if len(datasets) == 1 else "s")
+        if not datasets:
+            return
 
         progress = Progress("lsst.daf.butler.Butler.ingest", level=logging.DEBUG)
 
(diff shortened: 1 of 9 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -1862,44 +1862,83 @@ def ingest(
...
-            resolvedRefs: List[DatasetRef] = []
-
             # Somewhere to store pre-existing refs if we have an
             # execution butler.
             existingRefs: List[DatasetRef] = []
 
+            dataset_run: str | None = None
             for ref in dataset.refs:
-                if ref.dataId in groupedData[ref.datasetType]:
+                if ref.id is None:
```
timj, line 1900: Open question: If this is an unresolved ref, should I resolve it using `self.run` or should it be an error?

TallJimbo: Since this question is only relevant during the deprecation period, I think for now we resolve it using the now-deprecated `run` argument, and fall back to `self.run` when that isn't provided.

timj: Okay. I now do resolve the refs -- I was forced into this a bit because the resource gathering graph doesn't use resolved refs and so pipeline execution fails because the `_log` files can't be ingested.

python/lsst/daf/butler/_butler.py
```diff
@@ -1807,22 +1809,21 @@ def ingest(
...
-            used for `put` is assumed.  On successful return, all
-            `FileDataset.ref` attributes will have their `DatasetRef.id`
-            attribute populated and all `FileDataset.formatter` attributes will
-            be set to the formatter class used.  `FileDataset.path` attributes
-            may be modified to put paths in whatever the datastore considers a
-            standardized form.
+            relative to the datastore root, if applicable), a resolved
+            `DatasetRef`, and optionally a formatter class or its
+            fully-qualified string name.  If a formatter is not provided, the
+            formatter that would be used for `put` is assumed.  On successful
```
andy-slac, line 1815: On successful _ingest_?

python/lsst/daf/butler/_butler.py
```diff
@@ -1862,44 +1863,97 @@ def ingest(
...
-        # it's safe to swap them into FileDataset.refs.
-        # Some type annotation aliases to make that clearer:
-        GroupForType = Dict[DataCoordinate, Tuple[FileDataset, List[DatasetRef]]]
-        GroupedData = MutableMapping[DatasetType, GroupForType]
-        # The actual data structure:
-        groupedData: GroupedData = defaultdict(dict)
+
+        # We need to reorganize all the inputs so that they are grouped
+        # by dataset type and run. Multiple refs in a single FileDataset
+        # are required to share the run and dataset type.
```
andy-slac, line 1875: Does `FileDataset` enforce that in general (I don't see changes to `fileDataset.py` on this PR)? Or is this only relevant for ingest?

python/lsst/daf/butler/_butler.py
```diff
@@ -1921,35 +1975,65 @@ def ingest(
...
+            # Import the refs and expand the DataCoordinates since we can't
+            # guarantee that they are expanded and Datastore will need
+            # the records.
+            imported_refs = self.registry._importDatasets(refs_to_import, expand=True)
+            assert set(imported_refs) == set(refs_to_import)
+
+            # Replace all the refs in the FileDataset with expanded versions.
+            for dataset in grouped_datasets:
+                new_refs = [imported_refs.pop(0) for _ in dataset.refs]
+                dataset.refs = new_refs
```
andy-slac, line 2027: Made me think more than few instants what `pop` does here. Maybe something more obvious like:

    n_dataset_refs = len(dataset.refs)
    dataset.refs = imported_refs[:n_dataset_refs]
    del imported_refs[:n_dataset_refs]

?

andy-slac (approved):
Looks good

python/lsst/daf/butler/_butler.py
timj: `FileDataset` does not check that all the refs share a run. Before this they were all unresolved. Didn't cross my mind for the FileDataset constructor to check this (but maybe it should -- future ticket).

python/lsst/daf/butler/_butler.py
timj: Somehow I merged without seeing this comment.  I can see how your "read the first N" from the list might be clearer than doing N pops... 😄

python/lsst/daf/butler/_butler.py
timj: Will be resolved in #836.

PR #831, opened 2023-04-25
DM-38870: Change how file datastore transfer_from handles absolute URIs

timj:
Previously the code stored the absolute URI in the target datastore. Now it only does that in "auto" transfer mode. In other modes it will copy/link etc.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/storedFileInfo.py
+++ b/python/lsst/daf/butler/core/storedFileInfo.py
@@ -89,6 +89,11 @@ class StoredDatastoreItemInfo:
         """Dataset ID associated with this record (`DatasetId`)"""
         raise NotImplementedError()
 
+    def update(self, **kwargs: Any) -> StoredDatastoreItemInfo:
+        """Create a new class with everything retained apart from the
+        specified values."""
+        raise NotImplementedError()
+
 
 @dataclass(frozen=True)
 class StoredFileInfo(StoredDatastoreItemInfo):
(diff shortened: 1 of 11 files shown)


python/lsst/daf/butler/core/storedFileInfo.py
```diff
@@ -244,3 +241,12 @@ def from_record(cls: Type[StoredFileInfo], record: Mapping[str, Any]) -> StoredF
             dataset_id=record["dataset_id"],
         )
         return info
+
+    def update(self, **kwargs: Any) -> StoredFileInfo:
```
andy-slac, line 245: Unrecognized keywords are just silently ignored?

python/lsst/daf/butler/registry/bridge/monolithic.py
```diff
@@ -237,7 +237,13 @@ def join_records(
         # Run query, transform results into a list of dicts that we can later
         # use to delete.
         with self._db.query(info_in_trash) as sql_result:
-            rows = [dict(**row, datastore_name=self.datastoreName) for row in sql_result.mappings()]
+            # can not use dict(**row, datastore_name=) because mypy knows that
```
andy-slac: Can you do `dict(row, datastore=...)` instead?

andy-slac (approved):
Looks good, couple of minor questions.

python/lsst/daf/butler/registry/bridge/monolithic.py
timj: Very good. That does work.

PR #832, opened 2023-04-26
DM-37918: Change MetricMeasurementBundle import path

natelust:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -391,6 +391,6 @@ storageClasses:
   ScarletModelData:
     pytype: lsst.meas.extensions.scarlet.ScarletModelData
   MetricMeasurementBundle:
-    pytype: lsst.analysis.tools.analysisMetrics.MetricMeasurementBundle
+    pytype: lsst.analysis.tools.interfaces.MetricMeasurementBundle
   MultipleCellCoadd:
     pytype: lsst.cell_coadds.MultipleCellCoadd


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -391,6 +391,6 @@ storageClasses:
   ScarletModelData:
     pytype: lsst.meas.extensions.scarlet.ScarletModelData
   MetricMeasurementBundle:
-    pytype: lsst.analysis.tools.analysisMetrics.MetricMeasurementBundle
+    pytype: lsst.analysis.tools.interfaces._metricMeasurementBundle.MetricMeasurementBundle
```
timj:
```suggestion
    pytype: lsst.analysis.tools.interfaces.MetricMeasurementBundle
```

we don't include underscores in the name -- `get_full_type_name` elides them from the hierarchy (that's deliberate because the underscore is meant to be an internal implementation details)

timj (approved):
This change is fine if you remove the underscore part of the name.

python/lsst/daf/butler/configs/storageClasses.yaml
natelust: I must not be understanding something in how this works then:
```
In [1]: import lsst.analysis.tools.interfaces.MetricMeasurementBundle
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[1], line 1
----> 1 import lsst.analysis.tools.interfaces.MetricMeasurementBundle

ModuleNotFoundError: No module named 'lsst.analysis.tools.interfaces.MetricMeasurementBundle'
```

natelust: This was resolved on slack by Tim pointing out I was being dense.

PR #835, opened 2023-05-03
DM-38915: Fix handling of empty collection lists

andy-slac:
Empty collection iterable is now handled as it was supposed to be handled -
producing empty result (with "doomed by" messages). I added a proper type
annotation for `collections` parameter in Registry methods to help me
verify that it is used correctly (maybe need to extend that to Butler
methods too). Also needed to update couple of CLI scripts to fix their
handling of the default collection values. Unit tests was added to check
that it works correctly now.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/wildcards.py
+++ b/python/lsst/daf/butler/registry/wildcards.py
@@ -490,6 +490,11 @@ class CollectionWildcard:
             )
         return self.strings
 
+    def empty(self) -> bool:
+        """Return true if both ``strings`` and ``patterns`` are empty."""
+        # bool(Ellipsis) is True
+        return not self.strings and not self.patterns
+
     def __str__(self) -> str:
         if self.patterns is Ellipsis:
             return "..."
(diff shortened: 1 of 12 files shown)


python/lsst/daf/butler/registry/wildcards.py
```diff
@@ -490,6 +490,11 @@ def require_ordered(self) -> tuple[str, ...]:
             )
         return self.strings
 
+    def empty(self) -> bool:
+        """Return true if both ``strings`` and ``patterns`` are empty."""
+        # bool(Ellipsis) is True
+        return not self.strings and not self.patterns
+
```
TallJimbo, line 497: Would it make sense to implement `__bool__` instead, with the opposite sign?

TallJimbo (approved):
Looks good.

I don't think it's necessary to modify the butler methods right now unless you feel like it.  It does seem like we should try to get to that point eventually.  I had originally thought just using `Any` was fine here since we always conformed it with the wildcard class, but if it less this inconsistency slip through it's probably better to have this type alias.

python/lsst/daf/butler/registry/wildcards.py
andy-slac: I thought about `__bool__`, but I decided that explicit method is better, it's too easy to mix check for True/False with check for `None`.

PR #838, opened 2023-05-16
DM-39198: fix storage class handling in datastore, as used by QuantumBackedButler

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-39198.bugfix.md
+++ b/doc/changes/DM-39198.bugfix.md
@@ -1,0 +1,1 @@
+Fix bugs in storage class conversion in `FileDatastore`, as used by `QuantumBackedButler`.
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -1108,7 +1108,9 @@ class FileDatastore(GenericBaseDatastore):
             Information describing the artifact written to the datastore.
         """
         # May need to coerce the in memory dataset to the correct
-        # python type.
+        # python type, but first we need to make sure the storage class
+        # reflects the one defined in the data repository.
+        ref = self._cast_storage_class(ref)
         inMemoryDataset = ref.datasetType.storageClass.coerce_type(inMemoryDataset)
 
         location, formatter = self._prepare_for_put(inMemoryDataset, ref)
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -1563,15 +1563,20 @@ class FileDatastore(GenericBaseDatastore):
         existence : `dict` of [`DatasetRef`, `bool`]
             Mapping from dataset to boolean indicating existence.
         """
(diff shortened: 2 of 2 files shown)


doc/changes/DM-39198.bugfix.md
```diff
@@ -0,0 +1 @@
+Fix bugs in storage class conversion in FileDatastore, as used by QuantumBackedButler.
```
timj:
```suggestion
Fix bugs in storage class conversion in `FileDatastore`, as used by `QuantumBackedButler`.
```

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1561,15 +1563,20 @@ def _mexists(
         existence : `dict` of [`DatasetRef`, `bool`]
             Mapping from dataset to boolean indicating existence.
         """
-        # Need a mapping of dataset_id to dataset ref since the API
-        # works with dataset_id
-        id_to_ref = {ref.getCheckedId(): ref for ref in refs}
+        # Make a mapping from refs with the internal storage class to the given
+        # refs that may have a different one.  We'll use the internal refs
+        # throughout this method and convert back at the very end.
+        internal_ref_to_input_ref = {self._cast_storage_class(ref): ref for ref in refs}
```
timj, line 1569: This is needed because the graph doesn't include datastore records and so we are in "trustGetRequest" mode in QBB execution and so we have to determine the output file name and therefore need the correct storage class to find the correct template/file extension?

timj (approved):
Took me a while to work out that this is needed because the graph has no datastore records and so has to determine the file name every time.

python/lsst/daf/butler/datastores/fileDatastore.py
TallJimbo: In the particular QBB context, yes, and I think in execution butler too (which is weird, but maybe historical and irrelevant).  But I can't claim I've made sense of it all; it seems like sometimes we're expecting butler to guarantee that certain refs passed to Datastore already use the registry storage class, and in other cases we're expecting Datastore to fix them itself via the callback passed to `set_retrieve_dataset_type_method`.  What I've tried to do here is just be super defensive so the Datastore can handle what's thrown at it, and I bet that means we've got a few code paths where we ensure we've got the registry storage class on the same ref multiple times.

I really think the right fix is to make `Butler.datastore` private (which means giving `ctrl_mpexec` an alternative to calling `butler.datastore.mexists`) and then rework the Datastore API so it always gets registry storage classes in one argument and overrides in another argument, with the butler (either full or QBB) making sure those method are called correctly.  It'd also help a lot to modify Datastore APIs that really only need the dataset ID or ID + component (not the full ref) to take just that.

timj: It didn't used to matter because this was always getting refs that came from Registry because graphs used unresolved refs and butler.put used unresolved refs. Now that everything is resolved and there is no guarantee Registry even knows about it things are a bit more exciting.

PR #840, opened 2023-05-22
DM-32940: Add new Butler methods for checking existence.

timj:
These replace `Butler.datasetExists()`.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -27,7 +27,6 @@ from __future__ import annotations
 __all__ = (
     "Butler",
     "ButlerValidationError",
-    "DatasetExistence",
 )
 
 import collections.abc
@@ -38,7 +37,6 @@ import numbers
 import os
 import warnings
 from collections import defaultdict
-from enum import Flag, auto
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -67,6 +65,7 @@ from sqlalchemy.exc import IntegrityError
 
 from ._butlerConfig import ButlerConfig
 from ._butlerRepoIndex import ButlerRepoIndex
+from ._dataset_existence import DatasetExistence
 from ._deferredDatasetHandle import DeferredDatasetHandle
 from ._limited_butler import LimitedButler
 from .core import (
@@ -112,19 +111,6 @@ if TYPE_CHECKING:
     from .transfers import RepoImportBackend
 
 log = getLogger(__name__)
-
-
-class DatasetExistence(Flag):
-    Unknown = 0
-    RegistryKnows = auto()
-    DatastoreKnows = auto()
-    ArtifactExists = auto()
-    ArtifactNotChecked = auto()
-    Exists = RegistryKnows | DatastoreKnows | ArtifactExists
-    PartialCheck = RegistryKnows | DatastoreKnows | ArtifactNotChecked
-
-    def __bool__(self) -> bool:
-        return self.value == self.Exists.value or self.value == self.PartialCheck.value
 
 
 class ButlerValidationError(ValidationError):
@@ -1657,29 +1643,29 @@ class Butler(LimitedButler):
             datastore. Evaluates to `True` if the dataset is present and known
             to both.
         """
-        existence = DatasetExistence.Unknown
+        existence = DatasetExistence.UNRECOGNIZED
 
         if isinstance(datasetRefOrType, DatasetRef):
             registry_ref = self.registry.getDataset(datasetRefOrType.id)
             if registry_ref is not None:
-                existence |= DatasetExistence.RegistryKnows
+                existence |= DatasetExistence.RECORDED
             ref = datasetRefOrType
         else:
             try:
                 ref = self._findDatasetRef(datasetRefOrType, dataId, collections=collections, **kwargs)
             except (LookupError, TypeError, NoDefaultCollectionError):
                 return existence
-            existence |= DatasetExistence.RegistryKnows
+            existence |= DatasetExistence.RECORDED
 
         if self.datastore.knows(ref):
-            existence |= DatasetExistence.DatastoreKnows
+            existence |= DatasetExistence.DATASTORE
 
         if full_check:
             if self.datastore.exists(ref):
-                existence |= DatasetExistence.ArtifactExists
-        elif existence != DatasetExistence.Unknown:
+                existence |= DatasetExistence._ARTIFACT
+        elif existence != DatasetExistence.UNRECOGNIZED:
             # Do not add this flag if we have no other idea about a dataset.
-            existence |= DatasetExistence.ArtifactNotChecked
+            existence |= DatasetExistence._ASSUMED
 
         return existence
 
@@ -1712,30 +1698,30 @@ class Butler(LimitedButler):
             Each value evaluates to `True` if the dataset is present and known
             to both.
         """
-        existence = {ref: DatasetExistence.Unknown for ref in refs}
+        existence = {ref: DatasetExistence.UNRECOGNIZED for ref in refs}
 
         # Registry does not have a bulk API to check for a ref.
         for ref in refs:
             registry_ref = self.registry.getDataset(ref.id)
             if registry_ref is not None:
-                existence[ref] |= DatasetExistence.RegistryKnows
+                existence[ref] |= DatasetExistence.RECORDED
 
         # Ask datastore if it knows about these refs.
         knows = self.datastore.knows_these(refs)
         for ref, known in knows.items():
             if known:
-                existence[ref] |= DatasetExistence.DatastoreKnows
+                existence[ref] |= DatasetExistence.DATASTORE
 
         if full_check:
             mexists = self.datastore.mexists(refs)
             for ref, exists in mexists.items():
                 if exists:
-                    existence[ref] |= DatasetExistence.ArtifactExists
+                    existence[ref] |= DatasetExistence._ARTIFACT
         else:
             # Do not set this flag if nothing is known about the dataset.
             for ref in existence.keys():
-                if existence[ref] != DatasetExistence.Unknown:
-                    existence[ref] |= DatasetExistence.ArtifactNotChecked
+                if existence[ref] != DatasetExistence.UNRECOGNIZED:
+                    existence[ref] |= DatasetExistence._ASSUMED
 
         return existence
 
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1601,9 +1601,9 @@ class Butler(LimitedButler):
 
     def exists(
         self,
-        datasetRefOrType: DatasetRef | DatasetType | str,
+        dataset_ref_or_type: DatasetRef | DatasetType | str,
         /,
-        dataId: DataId | None = None,
+        data_id: DataId | None = None,
         *,
         full_check: bool = True,
         collections: Any = None,
@@ -1614,10 +1614,10 @@ class Butler(LimitedButler):
 
         Parameters
         ----------
-        datasetRefOrType : `DatasetRef`, `DatasetType`, or `str`
+        dataset_ref_or_type : `DatasetRef`, `DatasetType`, or `str`
             When `DatasetRef` the `dataId` should be `None`.
             Otherwise the `DatasetType` or name thereof.
-        dataId : `dict` or `DataCoordinate`
+        data_id : `dict` or `DataCoordinate`
             A `dict` of `Dimension` link name, value pairs that label the
             `DatasetRef` within a Collection. When `None`, a `DatasetRef`
             should be provided as the first argument.
@@ -1645,14 +1645,14 @@ class Butler(LimitedButler):
         """
         existence = DatasetExistence.UNRECOGNIZED
 
-        if isinstance(datasetRefOrType, DatasetRef):
-            registry_ref = self.registry.getDataset(datasetRefOrType.id)
+        if isinstance(dataset_ref_or_type, DatasetRef):
+            registry_ref = self.registry.getDataset(dataset_ref_or_type.id)
             if registry_ref is not None:
                 existence |= DatasetExistence.RECORDED
-            ref = datasetRefOrType
+            ref = dataset_ref_or_type
         else:
             try:
-                ref = self._findDatasetRef(datasetRefOrType, dataId, collections=collections, **kwargs)
+                ref = self._findDatasetRef(dataset_ref_or_type, data_id, collections=collections, **kwargs)
             except (LookupError, TypeError, NoDefaultCollectionError):
                 return existence
             existence |= DatasetExistence.RECORDED
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1680,6 +1680,65 @@ class Butler(LimitedButler):
 
         return existence
 
+    def exists_many(
+        self,
+        refs: Iterable[DatasetRef],
+        /,
+        *,
+        full_check: bool = True,
+    ) -> dict[DatasetRef, DatasetExistence]:
+        """Indicate whether multiple datasets are known to Butler registry and
+        datastore.
+
+        Parameters
+        ----------
+        refs : iterable of `DatasetRef`
+            The datasets to be checked.
+        full_check : `bool`, optional
+            If `True`, an additional check will be made for dataset artifact
+            existence. This will involve additional overhead due to the need
+            to query an external system. If `False` registry and datastore
+            will solely be asked if they know about the dataset but no
+            check for the artifact will be performed.
+
+        Returns
+        -------
+        existence : dict of [`DatasetRef`, `DatasetExistence`]
+            Mapping from the original dataset to an enum indicating the
+            status of the dataset in registry and datastore.
+            Each value evaluates to `True` if the dataset is present and known
+            to both.
+        """
+        existence = {ref: DatasetExistence.Unknown for ref in refs}
+
+        # Registry does not have a bulk API to check for a ref.
+        for ref in refs:
+            registry_ref = self.registry.getDataset(ref.getCheckedId())
+            if registry_ref is not None:
+                existence[ref] |= DatasetExistence.RegistryKnows
+
+        # Ask datastore if it knows about these refs.
+        knows = self.datastore.knows_these(refs)
+        for ref, known in knows.items():
+            if known:
+                existence[ref] |= DatasetExistence.DatastoreKnows
+
+        if full_check:
+            mexists = self.datastore.mexists(refs)
+            for ref, exists in mexists.items():
+                if exists:
+                    existence[ref] |= DatasetExistence.ArtifactExists
+        else:
+            for ref in existence.keys():
+                existence[ref] |= DatasetExistence.ArtifactNotChecked
+
+        return existence
+
+    @deprecated(
+        reason="Butler.datasetExists() has been replaced by Butler.exists(). Will be removed after v27.0.",
+        version="v26.0",
+        category=FutureWarning,
+    )
     def datasetExists(
         self,
         datasetRefOrType: Union[DatasetRef, DatasetType, str],
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1084,7 +1084,7 @@ class Butler(LimitedButler):
         datasetType, dataId = self._standardizeArgs(datasetRefOrType, dataId, for_put=False, **kwargs)
         if isinstance(datasetRefOrType, DatasetRef):
             if collections is not None:
-                warnings.warn("Collections should not be specified with DatasetRef", stacklevel=2)
+                warnings.warn("Collections should not be specified with DatasetRef", stacklevel=3)
             return datasetRefOrType
         timespan: Optional[Timespan] = None
 
@@ -1197,7 +1197,7 @@ class Butler(LimitedButler):
             # This is a direct put of predefined DatasetRef.
             log.debug("Butler put direct: %s", datasetRefOrType)
             if run is not None:
-                warnings.warn("Run collection is not used for DatasetRef")
+                warnings.warn("Run collection is not used for DatasetRef", stacklevel=3)
             # If registry already has a dataset with the same dataset ID,
             # dataset type and DataId, then _importDatasets will do nothing and
             # just return an original ref. We have to raise in this case, there
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -37,6 +37,7 @@ import numbers
 import os
 import warnings
 from collections import defaultdict
+from enum import Flag, auto
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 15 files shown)


timj:
@TallJimbo can you please take a look at these changes ahead of the formal Jira review? I have to change a bunch of other packages to use the new API but if you think I've done the API incorrectly I'd rather know ahead of time. This PR is essentially complete with a caveat that it seems to me that `QuantumBackedButler` could implement `exists` where `RegistryKnows` would be true if the ref is in the graph at all.

python/lsst/daf/butler/_butler.py
```diff
@@ -111,6 +114,19 @@
 log = getLogger(__name__)
 
 
+class DatasetExistence(Flag):
```
timj: This will likely have to move to a separate file (`_enums.py`?) if we are going to add `QuantumBackedButler.exists()`.

python/lsst/daf/butler/_butler.py
```diff
@@ -111,6 +114,19 @@
...
+    Unknown = 0
+    RegistryKnows = auto()
```
timj: I am not attached to these names if people have opinions.

python/lsst/daf/butler/_butler.py
TallJimbo: :+1:, but probably better to just make it `_dataset_existence.py`, unless we already have other enums with approximately the same downstream consumers we want to bundle up.

python/lsst/daf/butler/_limited_butler.py
```diff
@@ -273,6 +276,52 @@ def getDeferred(
...
+        -------
+        stored : `bool`
+            Whether the dataset exists in the datastore and can be retrieved.
+        """
+        return self.datastore.exists(ref)
+
+    def stored_many(
+        self,
+        refs: Iterable[DatasetRef],
+        artifact_existence: Optional[dict[ResourcePath, bool]] = None,
```
timj: I am torn as to whether `artifact_existence` should be a parameter here. It's useful as an opaque cache if people know they are going to call it many times but only if they also know that nothing is actively writing files between calls to it (which `transfer_from` does assume). I think I'm talking myself out of having it here as being too dangerous.

python/lsst/daf/butler/_butler.py
```diff
@@ -1597,6 +1613,142 @@ def retrieveArtifacts(
             overwrite=overwrite,
         )
 
+    def exists(
+        self,
+        datasetRefOrType: DatasetRef | DatasetType | str,
+        /,
+        dataId: DataId | None = None,
```
TallJimbo: How about `data_id` here instead?  I know the analogous methods have `dataId`, but I think new ones are an opportunity to move us along the path to snake_case consistency, such that some day we'll just just have a few camelCase ones left and then it'll be worth it to go after changing them directly.

python/lsst/daf/butler/_butler.py
```diff
@@ -1597,6 +1613,142 @@ def retrieveArtifacts(
...
+        **kwargs
+            Additional keyword arguments used to augment or construct a
+            `DataCoordinate`.  See `DataCoordinate.standardize`
+            parameters.
+
+        Returns
+        -------
+        existence : `DatasetExistence`
+            Object indicating whether the dataset is known to registry and
+            datastore. Evaluates to `True` if the dataset is present is known
```
```suggestion
            datastore. Evaluates to `True` if the dataset is present and known
```

python/lsst/daf/butler/_butler.py
```diff
@@ -1597,6 +1613,142 @@ def retrieveArtifacts(
...
+            to both.
+        """
+        existence = DatasetExistence.Unknown
+
+        if isinstance(datasetRefOrType, DatasetRef):
+            registry_ref = self.registry.getDataset(datasetRefOrType.id)
+            if registry_ref is not None:
```
TallJimbo, line 1655: We probably want a check here to see if the data ID and dataset type we got are the same (or compatible with) what we were given.  I think that not being the case should be an exception, but we also could make "present in registry, but not as defined" another enum flag.

python/lsst/daf/butler/_butler.py
```diff
@@ -1597,6 +1613,142 @@ def retrieveArtifacts(
...
+        """Indicate whether a dataset is known to Butler registry and
+        datastore.
+
+        Parameters
+        ----------
+        datasetRefOrType : `DatasetRef`, `DatasetType`, or `str`
+            When `DatasetRef` the `dataId` should be `None`.
+            Otherwise the `DatasetType` or name thereof.
+            If a resolved `DatasetRef`, the associated dataset
+            is returned directly without additional querying.
```
TallJimbo: I'd consider the `getDataset` call "additional querying", since it still involves hitting the database (same comment for `exists_many`).

python/lsst/daf/butler/_butler.py
```diff
@@ -1597,6 +1613,142 @@ def retrieveArtifacts(
...
+            will solely be asked if they know about the dataset but no
+            check for the artifact will be performed.
+
+        Returns
+        -------
+        existence : dict of [`DatasetRef`, `DatasetExistence`]
+            Mapping from the original dataset to an enum indicating the
+            status of the dataset in registry and datastore.
+            Each value evaluates to `True` if the dataset is present is known
+            to both.
```
TallJimbo, line 1718: Should state whether it's guaranteed that all given refs will be present as keys, and if so, whether they'll be in the same order.

We also need to be really careful that all returned refs are identical to the given ones, not e.g. be mostly the same but have a compatible storage class.  I think this method is safe in that regard only if the datastore methods it calls are safe in that regard.

python/lsst/daf/butler/_limited_butler.py
```diff
@@ -273,6 +276,52 @@ def getDeferred(
         """
         return DeferredDatasetHandle(butler=self, ref=ref, parameters=parameters, storageClass=storageClass)
 
+    def stored(self, ref: DatasetRef) -> bool:
+        """Indicate whether the dataset is present in the Datastore.
```
TallJimbo: Should be more precise about what this means, e.g. "whether the dataset's artifacts are all present in the Datastore".

python/lsst/daf/butler/_limited_butler.py
```diff
@@ -273,6 +276,52 @@ def getDeferred(
...
+        Parameters
+        ----------
+        refs : iterable of `DatasetRef`
+            The datasets to be checked.
+        artifact_existence : `dict` [`lsst.resources.ResourcePath`, `bool`]
+            Optional mapping of datastore artifact to existence. Updated by
+            this method with details of all artifacts tested. Can be `None`
+            if the caller is not interested but can be an efficiency gain
+            if you know that this API is going to called repeatedly with
+            possibly duplicate refs.
```
TallJimbo: Is the idea that anything in this `dict` already is not checked?  Docs may need some clarification.

python/lsst/daf/butler/_butler.py
TallJimbo: I think they should be ALL_CAPS, right?  Or have we already been inconsistent in that sense about enum values?  I know `CollectionType` is all caps.

python/lsst/daf/butler/_butler.py
```diff
@@ -111,6 +114,19 @@
...
 
 
+class DatasetExistence(Flag):
+    Unknown = 0
+    RegistryKnows = auto()
+    DatastoreKnows = auto()
+    ArtifactExists = auto()
+    ArtifactNotChecked = auto()
+    Exists = RegistryKnows | DatastoreKnows | ArtifactExists
+    PartialCheck = RegistryKnows | DatastoreKnows | ArtifactNotChecked
```
TallJimbo: I'd like to handle `ArtifactNotChecked` differently somehow; we also don't check artifact existence when `~RegistryKnows` or `~DatastoreKnows`, so we should at least rename it to something like `ArtifactAssumed`.  But I also think it might be better to make `DatastoreKnows & ~ArtifactExists` back into an error, since that's really not a state the repository should ever be in (and that'll be even more seriously the case after DMTN-249).  We could then drop the `ArtifactNotChecked` flag, with the default being that artifact existence is assumed if `DatastoreKnows` and verified if `ArtifactExists`.

We could also consider dropping `ArtifactExists`, too, since its value would then be wholly deterministic from `DatastoreKnows & full_check`, and the caller could be expected to know what they passed as `full_check`, but I have no objection to keeping it.

If we keep `ArtifactExists`, I think we should also rename `PartialCheck` (` = RegistryKnows | DatastoreKnows`) to something like `FastExists` and rename `Exists` to `ExhaustiveExists` so the most default "does exist" result doesn't imply that the user should have done more.

I think we could accomplish all that in terms of flags, but here's a regular enum version that may be worth considering:
```
class DatasetExistence(Enum):
    UNRECOGNIZED = auto()   # No matching dataset in Registry.  Datastore not checked.
    RECORDED = auto()   # In Registry, but not in any Datastore (per datastore records)
    COMPLETE = auto()  # In Registry and Datastore (per datastore records)
    EXHAUSTIVE = auto()  # In Registry and Datastore, artifact verified

    def __bool__(self) -> bool:
        return self is self.COMPLETE or self is self.EXHAUSTIVE
```
The advantage of this is that it can only represent states that make sense (e.g. no `~RegistryKnows & DatastoreKnows` or `ArtifactExists & ArtifactNotChecked`).  The disadvantage is that you can't just OR in new flags to "upgrade" the state.

TallJimbo (commented):
My preferred interpretation of the `RegistryKnows` state is that it really means "in the central database for this data repository", not "in whatever is providing metadata about datasets for this butler client".  I suppose we could extend `exists` to return something about both, and then `LimitedButler` could have a version of `exists` that can takes `DatasetRef` only) and can say yes to the latter while always saying "I don't know" for the former.  But that's a bunch of of new `DatastoreExistence` states that I'm not sure carry their weight.

python/lsst/daf/butler/_butler.py
timj: All caps seemed overkill and shouty for `ARTIFACT_NOT_CHECKED` but that's fine but if that's the policy for enums.

python/lsst/daf/butler/_butler.py
timj: I used `ArtifactNotChecked` because it seemed to simplify the bool logic (and I had to check `.value` because otherwise I ended up with recursion errors) and allowed the "truth" checks to work consistently -- I wasn't expecting it to be a public flag that people would check against but was an internal assertion that everything was okay and had been checked.

timj: I also wasn't expecting PartialCheck to be public. It was also there to simplify the boolean. I was assuming that people would either just use truthiness or be asking for RegistryKnows or DatastoreKnows explicitly.

python/lsst/daf/butler/_butler.py
timj: I was a bit worried that people are so used to say `dataId=` in `butler.get` and `butler.getURI` that they will do the same thing for `butler.exists`.

python/lsst/daf/butler/_butler.py
timj: Yes. That's a copy and paste problem from `butler.get`

python/lsst/daf/butler/_butler.py
timj: I was not expecting a situation where a UUID could match but somehow be associated with a completely different ref. Doesn't that seem unlikely? I can add the check of dataId and datasetType if you think it's possible. I really need a `DatasetRef.is_compatible_with` call or something (since `__eq__` does not check for storage class compatibility)

python/lsst/daf/butler/_butler.py
timj: My other comment is that I much prefer compound enum states. RECORDED meaning in registry but not in datastore makes it harder to code because it means that you need to keep track of the registry state and then later decide whether to override with datastore state. Whereas OR-ing together as you go and then using combination flags for the public information is easier to calculate at the end.

timj: I've rewritten and renamed the Flags and hidden flag names for the artifact existence work well.

python/lsst/daf/butler/_limited_butler.py
timj: I think I've talked myself out of forwarding `artifact_existence` for the above reasons.

python/lsst/daf/butler/_butler.py
timj: Okay. I switched over.

python/lsst/daf/butler/_butler.py
timj: `datastore.mexists()` does guarantee that the keys are the refs you gave it even if the refs had different storage classes. I've updated the docs.

python/lsst/daf/butler/registry/queries/_query_backend.py
```diff
@@ -624,8 +623,7 @@ def resolve_governor_constraints(
 
         Returns
         -------
-        resolved : `Mapping` [ `str`,  [ `~collections.abc.Set`
-                [ `str` ] ] ]
+        resolved : `Mapping` [ `str`,  [ `~collections.abc.Set` [ `str` ] ] ]
```
kfindeisen: FTR, the [Python style guide](https://developer.lsst.io/python/numpydoc.html#py-docstring-parameter-types-dict) only requires spaces _before_ brackets, not inside them. The extra space adds up for highly nested types, and AFAIK it has no effect on the final HTML.

timj: I'm dreaming of a time when I don't need to put any of this in because the type annotations will be used directly.

kfindeisen: Personally, I hope that doesn't happen. Some of the broken links on #846 were the result of trying to use the exact same notation as the annotations, even when Sphinx couldn't understand them. In a few cases (e.g., "any collection expression"), the human-readable type is _much_ simpler than its machine-readable equivalent (`str | Pattern | Iterable[str | Pattern] | ellipsis | CollectionWildcard | None`).

kfindeisen: While revisiting this line I realized that `resolved : mapping [ str,  [ set [ str ] ] ]` is not valid notation to begin with. Is this supposed to be a mapping from strings to sets, or a mapping from strings to lists of sets?

python/lsst/daf/butler/_butler.py
```diff
@@ -1597,6 +1599,161 @@ def retrieveArtifacts(
...
+                    existence[ref] |= DatasetExistence._ARTIFACT
+        else:
+            # Do not set this flag if nothing is known about the dataset.
+            for ref in existence.keys():
+                if existence[ref] != DatasetExistence.UNRECOGNIZED:
+                    existence[ref] |= DatasetExistence._ASSUMED
+
+        return existence
+
+    @deprecated(
```
timj, line 1756: I think I will have to remove this message for the moment since we have not filed an RFC formally deprecating the interface.

python/lsst/daf/butler/_dataset_existence.py
```diff
@@ -0,0 +1,82 @@
...
+from __future__ import annotations
+
+__all__ = ["DatasetExistence"]
+
+from enum import Flag, auto
+
+
+class DatasetExistence(Flag):
+    """Flags representing the different states that a dataset can have
+    in a Butler repository."""
```
andy-slac: Maybe expand documentation a little bit to help people who are not very familiar with the combination of flags. Generated docs do not show that `KNOWN = RECORDED | DATASTORE | _ASSUMED`, it just shows that `KNOWN = 11` which could be confusing. And maybe explain logic behind boolean conversion, docs for `__bool__` method do not appear in the generated output.

python/lsst/daf/butler/_dataset_existence.py
```diff
@@ -0,0 +1,82 @@
...
+    def __bool__(self) -> bool:
+        """Indicate whether the dataset exists.
+
+        Returns
+        -------
+        exists : `bool
+            Evaluates to `True` if the flags evaluate to either ``KNOWN``
+            or ``VERIFIED``.
+        """
+        return self.value == self.KNOWN.value or self.value == self.VERIFIED.value
```
andy-slac, line 98: You could probably write `self is self.KNOWN or self is self.VERIFIED`?

python/lsst/daf/butler/_butler.py
```diff
@@ -1597,6 +1599,161 @@ def retrieveArtifacts(
...
+
+        if self.datastore.knows(ref):
+            existence |= DatasetExistence.DATASTORE
+
+        if full_check:
+            if self.datastore.exists(ref):
+                existence |= DatasetExistence._ARTIFACT
+        elif existence != DatasetExistence.UNRECOGNIZED:
+            # Do not add this flag if we have no other idea about a dataset.
+            existence |= DatasetExistence._ASSUMED
```
andy-slac, line 1685: Should this check be for `existence | DatasetExistence.DATASTORE`, does it make sense to have `_ASSUMED` without `DATASTORE`? Maybe whole `if full_check:` wants to be nested in `if self.datastore.knows(ref):`?

python/lsst/daf/butler/_butler.py
```diff
@@ -1597,6 +1599,161 @@ def retrieveArtifacts(
...
+
+        return existence
+
+    def _exists_many(
+        self,
+        refs: Iterable[DatasetRef],
+        /,
+        *,
```
andy-slac, line 1693: Do you need `/, *,` if there is nothing in between?

python/lsst/daf/butler/_butler.py
andy-slac: Probably makes sense to add it on an implementation ticket for that RFC.

python/lsst/daf/butler/_quantum_backed.py
```diff
@@ -447,15 +447,40 @@ def getDeferred(
...
+
+    @deprecated(
+        reason="Butler.datasetExistsDirect() has been replaced by Butler.stored(). "
+        "Will be removed after v27.0.",
+        version="v26.0",
+        category=FutureWarning,
+    )
+    def datasetExistsDirect(self, ref: DatasetRef) -> bool:
+        # Docstring inherited.
+        return self.stored(ref)
```
andy-slac: It looks the same as base class implementation, could be dropped?

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -628,6 +628,56 @@ def overrideStorageClass(self, storageClass: str | StorageClass) -> DatasetRef:
...
+
+            assert ref.is_compatible_with(new_ref) is True
+
+        since we know that the python type associated with the new ref can
+        be converted to the original python type. The reverse is not guaranteed
+        and depends on whether bidirectional converters have been registered.
+        """
+        if not isinstance(ref, DatasetRef):
+            # mypy thinks this is impossible but we need to prevent
+            # an AttributeError if possible.
```
andy-slac: Not clear where the AttributeError comes from?

tests/test_butler.py
```diff
@@ -342,8 +343,7 @@ def runPutGetTest(self, storageClass: StorageClass, datasetTypeName: str) -> But
                 # Now remove the dataset completely.
                 butler.pruneDatasets([ref], purge=True, unstore=True)
                 # Lookup with original args should still fail.
-                with self.assertRaises(LookupError):
-                    butler.datasetExists(*args, collections=this_run)
```
andy-slac, line 346: Maybe keep tests for `datasetExists` for now until it's actually deprecated?

andy-slac (approved):

python/lsst/daf/butler/_dataset_existence.py
timj: I will try that again. When I did `self == self.KNOWN` I got a recursion exception but I had not tried using `is`.

python/lsst/daf/butler/_butler.py
timj: It is possible for the file to exist but for datastore not to know about it. That's the reason we have to two independent states. This is how execution butler runs and how `trustGetRequest` works. `full_check` therefore can't be nested inside `knows`. In a "proper" datastore the `datastore.exists` will check if it knows first and return False regardless.

python/lsst/daf/butler/_butler.py
timj: I'm not entirely certain. I was giving myself flexibility but this is a private API at the moment so we can do whatever here.

python/lsst/daf/butler/core/datasets/ref.py
timj: It will come from the `ref.id` below if you pass the wrong thing here. mypy never expects you to pass the wrong thing here so is convinced it can't happen.

python/lsst/daf/butler/_butler.py
```diff
@@ -1597,6 +1599,161 @@ def retrieveArtifacts(
...
+        full_check : `bool`, optional
+            If `True`, an additional check will be made for dataset artifact
+            existence. This will involve additional overhead due to the need
+            to query an external system. If `False` registry and datastore
+            will solely be asked if they know about the dataset but no
+            check for the artifact will be performed.
+        collections : Any, optional
+            Collections to be searched, overriding ``self.collections``.
+            Can be any of the types supported by the ``collections`` argument
+            to butler construction.
```
andy-slac, line 1633: One more thought - it looks like `collections` is not used when `ref` is given, meaning that we'll find a ref even if it's not in any of the `collections`. Maybe clarify that in the docstring to avoid surprises for users who may think that `collections` could be used as restriction even if dataset ref is provided?

timj: Right. We do warn in `Butler.get()` now if you pass ref and collections.

python/lsst/daf/butler/core/datasets/ref.py
andy-slac: But this is a completely new method, don't we control what is being passed to it? And if someone passes and incorrect type then they probably deserve AttributeError, not? Otherwise if we allow other types maybe just declare argument as `ref: object`?

timj: I was being defensive since it felt to me that this was a lot like `__eq__` which does have to worry about any random thing being given to it. If you feel that AttributeError is fine if someone passes in a non-ref I'll change it.

andy-slac: I think that checking a non-ref to be compatible with a ref is a meaningless operation and no one is going to ask those questions, so an exception should be fine here. And Python 4 should probably raise TypeError if an actual argument type is different from annotation 🙂

python/lsst/daf/butler/_butler.py
TallJimbo: I may have misunderstood what happens in trust mode.  Does that never return DATASET_KNOWS or ARTIFACT_ASSUMED?

I guess that makes sense but I worry that it exposes an implementation detail (whether the datastore is in trust mode).

Of course, that may not matter if we can get to the point where trust mode is only ever paired with QBB, not a full Butler, after execution butler is gone.  I'm hoping we can.

timj: `_ASSUMED` is a private flag entirely there to make things obvious internally. It's not a public state. In execution butler it will return False because it really doesn't know about the ref and full check will indicate that the artifact does exist. You will see extensive tests exploring all the different states. If `datastore.trustGetRequest` is something that `Butler.exists` is aware of then we could make it look on the file system regardless of `full_check` and add `DATASTORE` flag if the artifact exists.

TallJimbo: I think I've convinced myself I don't care what happens with a full Butler in `trustGetRequest` mode, since I just want to make that combination go away entirely instead.

python/lsst/daf/butler/_butler.py
timj: [RFC-933](https://jira.lsstcorp.org/browse/RFC-933) has been approved.

python/lsst/daf/butler/_dataset_existence.py
timj: The doc strings do explain which flags evaluate to true and which evaluate to false but I've added some text.

python/lsst/daf/butler/core/datasets/ref.py
timj: I've removed it, it did remind me that `__eq__` does not check that the `run` matches, only the ID, dataset type, and dataID...

PR #841, opened 2023-05-23
DM-37704: Make all DatasetRefs resolved.

andy-slac:
DatasetRefs are now required to have defined dataset ID and run. This also
removes methods that made unresolved refs or resolved unresolved refs.
Many changes everywhere to avoid checking for unresolved refs.

`Butler.put()` method is updated to handle duplicate put attempts with the same
dataset ID. The behavior is still slightly different between resolved ref and
unresolved (DatasetType, DataId) in case dataset only exists in Registry but not in
datastore. Unit test reflects this difference in behavior.

## Checklist

- [X] ran Jenkins


python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -172,60 +145,12 @@ def makeDatasetId(
...
-        with warnings.catch_warnings():
-            warnings.simplefilter("ignore", category=UnresolvedRefWarning)
-            resolved = ref.resolved(datasetId, run)
-        return resolved
-
 
 class SerializedDatasetRef(BaseModel):
     """Simplified model of a `DatasetRef` suitable for serialization."""
 
     # DO NOT change order in the Union, pydantic is sensitive to that!
```
timj: This comment is irrelevant now (has been since we stopped allowing `int`).

python/lsst/daf/butler/_butler.py
```diff
@@ -243,7 +240,6 @@ def __init__(
                 else:
                     aliases = "(no known aliases)"
                 raise FileNotFoundError(f"{e} {aliases}") from e
-            self._config = ButlerConfig(config, searchPaths=searchPaths)
```
timj, line 246: Well spotted. I'm sure it will help not creating the config twice.

python/lsst/daf/butler/_butler.py
```diff
@@ -1052,7 +1050,7 @@ def _findDatasetRef(
         ------
         LookupError
             Raised if no matching dataset exists in the `Registry` (and
-            ``allowUnresolved is False``).
+            ``predict is False``).
```
timj:
```suggestion
            ``predict`` is `False`).
```
?

python/lsst/daf/butler/_butler.py
```diff
@@ -1179,62 +1167,41 @@ def put(
         TypeError
             Raised if the butler is read-only or if no run has been provided.
         """
-        if isinstance(datasetRefOrType, DatasetRef) and datasetRefOrType.id is not None:
+        if isinstance(datasetRefOrType, DatasetRef):
             # This is a direct put of predefined DatasetRef.
             log.debug("Butler put direct: %s", datasetRefOrType)
+            # _importDatasets ignores existing dataset ref and just returns an
+            # original ref.
```
timj: Do you mean it returns a new `DatasetRef` but one which should match the given `DatasetRef` if the given ref exists in registry already? I can't remember what happens if you try to import a ref that already exists in the same collection but which has a different UUID. Does it ignore the UUID and give you the existing registry version back instead? Maybe we should do the UUID check in `_importDatasets` and then not return anything given that all refs are resolved and so will either be in the registry already or be new additions, but nothing will change the ref.

python/lsst/daf/butler/_butler.py
```diff
@@ -1309,16 +1276,7 @@ def getDirectDeferred(
...
-        Raises
-        ------
-        AmbiguousDatasetError
-            Raised if ``ref.id is None``, i.e. the reference is unresolved.
         """
-        if ref.id is None:
-            raise AmbiguousDatasetError(
-                f"Dataset of type {ref.datasetType.name} with data ID {ref.dataId} is not resolved."
-            )
         return DeferredDatasetHandle(butler=self, ref=ref, parameters=parameters, storageClass=storageClass)
```
timj, line 1289: Part of me wonders if we should now be checking if `ref` is in `butler` when we construct a `DeferredDatasetHandle` now that it's entirely possible for refs to be magicked up from nowhere.

python/lsst/daf/butler/_butler.py
```diff
@@ -1897,20 +1836,6 @@ def ingest(
             resolvedRefs: list[DatasetRef] = []
```
timj: I think this can now be removed and the check for resolvedRefs also removed later on.

python/lsst/daf/butler/_butler.py
```diff
@@ -1921,40 +1846,6 @@ def ingest(
                         f" {groupedDataIds[group_key][ref.dataId].path} "
                         f" ({ref.dataId})"
                     )
-                if self._allow_put_of_predefined_dataset:
-                    existing_ref = self.registry.findDataset(
```
timj, line 1925: So in execution butler we do call ingest for the logs datasets. I suppose this works now because we now create execution butler from the graph and so we guarantee that the UUID matches.  Although we now rely on `_importDatasets` doing the right thing if it sees a mismatch in UUID (which is a lot like my comment above).

python/lsst/daf/butler/_butler.py
```diff
@@ -2149,16 +2038,6 @@ def import_(
             Transfer mode passed to `~lsst.daf.butler.Datastore.ingest`.
         skip_dimensions : `set`, optional
             Names of dimensions that should be skipped and not imported.
-        idGenerationMode : `DatasetIdGenEnum`, optional
-            Specifies option for generating dataset IDs when IDs are not
-            provided or their type does not match backend type. By default
-            unique IDs are generated for each inserted dataset.
-        reuseIds : `bool`, optional
```
timj, line 2156: Did you spot the use of `reuseIds` in `daf_butler_migrate`?

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -254,10 +179,10 @@ def _check_component(cls, v: Any, values: Dict[str, Any]) -> Any:  # noqa: N805
     def direct(
         cls,
         *,
-        id: Optional[Union[str, int]] = None,
+        id: str,
```
timj, line 181: Can we add a comment that `direct` always uses `str` even though the pydantic model would normally automatically convert it to UUID itself?

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -302,11 +227,11 @@ class DatasetRef:
     dataId : `DataCoordinate`
         A mapping of dimensions that labels the Dataset within a Collection.
     id : `DatasetId`, optional
-        The unique identifier assigned when the dataset is created. If ``run``
-        is specified and ``id`` is not specified, an ID will be created.
-    run : `str`, optional
+        The unique identifier assigned when the dataset is created. If ``id``
+        is not specified, a new unique ID will be created.
+    run : `str`
```
timj, line 233: You may want to switch the order to match the order below.

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -346,8 +271,8 @@ def __init__(
         datasetType: DatasetType,
         dataId: DataCoordinate,
         *,
+        run: str,
```
timj, line 277: Move it in front of the `*`?

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -399,15 +308,13 @@ def __repr__(self) -> str:
         # DataCoordinate's __repr__ - while adhering to the guidelines for
         # __repr__ - is much harder to users to read, while its __str__ just
         # produces a dict that can also be passed to DatasetRef's constructor.
-        if self.id is not None:
-            return f"DatasetRef({self.datasetType!r}, {self.dataId!s}, id={self.id}, run={self.run!r})"
-        else:
-            return f"DatasetRef({self.datasetType!r}, {self.dataId!s})"
+        return f"DatasetRef({self.datasetType!r}, {self.dataId!s}, id={self.id}, run={self.run!r})"
```
timj: Switch the run and id round to match the call signature?

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -399,15 +308,13 @@ def __repr__(self) -> str:
...
 
     def __str__(self) -> str:
-        s = f"{self.datasetType.name}@{self.dataId!s}, sc={self.datasetType.storageClass_name}]"
-        if self.id is not None:
-            s += f" (id={self.id})"
+        s = (
+            f"{self.datasetType.name}@{self.dataId!s}, sc={self.datasetType.storageClass_name}]"
```
timj: I do continually wonder why this doesn't include the run. I print these all the time and then end up remembering I have to use `repr()`.

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -441,7 +348,7 @@ def to_simple(self, minimal: bool = False) -> SerializedDatasetRef:
         simple : `dict` or `int`
             The object converted to a dictionary.
         """
-        if minimal and self.id is not None:
+        if minimal:
             # The only thing needed to uniquely define a DatasetRef
             # is its id so that can be used directly if it is
             # resolved and if it is not a component DatasetRef.
```
timj: Maybe clean up the comment to remove the "if it is resolved" bit. Don't know how many places that phrase exists in a comment still.

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -462,9 +369,8 @@ def to_simple(self, minimal: bool = False) -> SerializedDatasetRef:
         }
 
         # Only include the id entry if it is defined
```
timj: It's always defined so the comment can go, and you could consider removing `as_dict` completely since it was only there to deal with the optional ID/run.

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -539,21 +445,17 @@ def from_simple(
             raise ValueError("The DataId must be specified to construct a DatasetRef")
         dataId = DataCoordinate.from_simple(simple.dataId, universe=universe)
 
-        # Issue our own warning that could be more explicit.
-        if simple.id is None and simple.run is None:
+        # Check that simple ref is resolved.
+        if simple.id is None or simple.run is None:
```
timj: Doesn't the `SerializedDatasetRef` constructor disallow this anyhow? I'm surprised that mypy doesn't say this is a pointless test. Is this some edge case that you can get to via `direct()`?

python/lsst/daf/butler/script/butlerImport.py
```diff
@@ -56,6 +57,12 @@ def butlerImport(
     reuse_ids : `bool`
         If `True` forces re-use of imported dataset IDs for integer IDs.
     """
+    if reuse_ids:
+        warnings.warn(
+            "--reuse-ids option is deprecated and will be prohibited after v26.",
```
timj: I think you can remove the parameter from the script interface completely and add the warning directly to the click command itself.

python/lsst/daf/butler/tests/utils.py
```diff
@@ -38,8 +38,22 @@
 from ..tests import MetricsExample, addDatasetType
 
 if TYPE_CHECKING:
+    import unittest
+
     from lsst.daf.butler import DatasetType
 
+    class TestCaseMixin(unittest.TestCase):
```
timj, line 45: Hmm. Clever. I usually end up attaching things like `assertTrue: Callable` in a TYPE_CHECKING block but this is more efficient.

tests/test_butler.py
```diff
@@ -417,40 +435,44 @@ def runPutGetTest(self, storageClass, datasetTypeName):
         # already had a component removed
         butler.pruneDatasets([ref], unstore=True, purge=True)
 
-        # Check that we can configure a butler to accept a put even
-        # if it already has the dataset in registry.
-        ref = butler.put(metric, refIn)
+        # Check that duplicate put fails.
+        ref = butler.put(metric, datasetType, dataId)
```
timj, line 439: The comment says that it fails but the code is not in a try block. I think it's saying that the put after this put will fail but it's not clear until you get to the next statement.

tests/test_butler.py
```diff
@@ -1365,35 +1400,39 @@ def testPathConstructor(self):
...
             self.assertIsInstance(butler, Butler)
 
-    def testExportTransferCopy(self):
+    def testExportTransferCopy(self) -> None:
         """Test local export using all transfer modes"""
         storageClass = self.storageClassFactory.getStorageClass("StructuredDataNoComponents")
         exportButler = self.runPutGetTest(storageClass, "test_metric")
         # Test that the repo actually has at least one dataset.
-        datasets = list(exportButler.registry.queryDatasets(..., collections=...))
+        datasets = list(exportButler.registry.queryDatasets(..., collections=Ellipsis))
```
timj, line 1407: Why? People use `...` all the time don't they? Are you saying that people need to use `Ellipsis` explicitly for mypy to work?

tests/test_butler.py
```diff
@@ -1419,9 +1458,9 @@ def testPruneDatasets(self):
             butler.datasetExists(ref1.datasetType, ref1.dataId, collections=run1)
 
         # Put data back.
-        ref1 = butler.put(metric, ref1.unresolved(), run=run1)
-        ref2 = butler.put(metric, ref2.unresolved(), run=run2)
-        ref3 = butler.put(metric, ref3.unresolved(), run=run1)
+        ref1 = butler.put(metric, ref1, run=run1)
```
timj, line 1460: If people put a pre-existing ref but override the run that works and you get a new ref back?

tests/test_butler.py
```diff
@@ -2031,7 +2071,7 @@ def assertButlerTransfers(self, id_gen_map=None, purge=False, storageClassName="
             if index < 2:
                 source_refs.append(ref)
             if ref not in deleted:
-                new_metric = butler.get(ref.unresolved(), collections=run)
+                new_metric = butler.get(ref, collections=run)
```
timj: Is `collections=run` needed if you have a resolved ref that has the run in it? Should we warn if you try to do that in case people are expecting the collections parameter to do anything.

tests/test_butler.py
```diff
@@ -2105,9 +2145,8 @@ def assertButlerTransfers(self, id_gen_map=None, purge=False, storageClassName="
         # Now try to get the same refs from the new butler.
         for ref in source_refs:
             if ref not in deleted:
-                unresolved_ref = ref.unresolved()
-                new_metric = self.target_butler.get(unresolved_ref, collections=ref.run)
-                old_metric = self.source_butler.get(unresolved_ref, collections=ref.run)
+                new_metric = self.target_butler.get(ref, collections=ref.run)
```
timj: Again the collections parameter.

timj (approved):
Looks great. I have some minor comments. Main ones are:

* Should `_importDatasets` check that the UUIDs it is returning match those it's given? We seem to be testing that outside of the import and that seems unreliable.
* Maybe check that a DeferredDatasetHandle really can get the ref from the butler it's given?

tests/test_butler.py
TallJimbo: I think mypy might be able to handle `...` now if we removed the workaround for it (in utils, maybe?).  It couldn't previously.

andy-slac:
> Should `_importDatasets` check that the UUIDs it is returning match those it's given? We seem to be testing that outside of the import and that seems unreliable.

I have added a check: cfa6ae38

> Maybe check that a DeferredDatasetHandle really can get the ref from the butler it's given?

Is not the most common case for deferred get is when the ref comes from a quantum graph and supposed to exist? There is also a potential race here, checking that ref exists now does not mean it will exist later?

python/lsst/daf/butler/core/datasets/ref.py
andy-slac: I'll drop it.

python/lsst/daf/butler/_butler.py
andy-slac: The comment is confusing, I'll update it. What I meant is that if the dataset with the same UUID, dataset type and dataId is already in the registry then `_importDatasets` simply returns it and does not raise (so I need to check datastore and raise exception myself below). If a dataset with the same dataset type and dataId but different UUID is already in the registry, then `_importDatasets` will raise an exception (same for matching UUID but different other items).

I'm not very comfortable changing how `_importDatasets` works, there may be places that rely on current behavior?

python/lsst/daf/butler/_butler.py
andy-slac: Well, I think "nowhere" is going to be mostly QuantumGraph, and that already uses Registry contents?

python/lsst/daf/butler/_butler.py
andy-slac: I started removing that, but then saw something that confuses me and I don't know why `mypy` does not complain about it. Could you look at this whole [block](https://github.com/lsst/daf_butler/blob/tickets/DM-37704/python/lsst/daf/butler/_butler.py#L1838-L1875) to check that it looks OK? What worries me there is that `else` at the bottom uses `group_key` which is defined in the `for` loop above. If loop is empty then `group_key` will be undefined. I think this code implies that loop is never empty and that all `group_key` values are identical in that loop, is that indeed so? Code on `main` looks similar in structure.

python/lsst/daf/butler/_butler.py
andy-slac: Exactly.

python/lsst/daf/butler/_butler.py
andy-slac: Hmm, I forgot about that. I searched github but only in `lsst` org, not `lsst-dm`, will check that too.

python/lsst/daf/butler/_butler.py
timj: If `_importDatasets` already raises, does that mean the test a couple of lines down was always irrelevant?

python/lsst/daf/butler/_butler.py
andy-slac: PR for daf_butler_migrate: https://github.com/lsst-dm/daf_butler_migrate/pull/22. Thanks for approving!

python/lsst/daf/butler/core/datasets/ref.py
andy-slac: I think it's also missing an opening square bracket [my guess `sc=...` was supposed to be in square brackets]. I'll add `run=...` before the `id=...`, it feels reasonable to have it here. Though I do not see this format frequently on output, I think `repr()` is used more often.

python/lsst/daf/butler/_butler.py
timj: I think the point is that you can now create a resolved `DatasetRef` any time (before they were rare and the only way you got one was from a butler query). It seems prudent to do a quick registry check in butler that the ref really is known to that butler and not some other random butler (or is orphaned).

python/lsst/daf/butler/core/datasets/ref.py
andy-slac: I was worried that someone could try to read QuantumGraph made by previous release. I guess in that case there will be an exception raised in some other location, so I can probably drop `simple.id is None` check (but check for `run` is still needed for `mypy`).

timj: I'm pretty sure we are still saying that old graphs are not supported. We are going to say that right up until we start defining graphs as provenance.

python/lsst/daf/butler/_butler.py
timj: It's not meant to be the case that a FileDataset has no refs in it so the `group_key` should be set and it should be identical for every ref in `dataset.refs`. A single `FileDataset` is not meant to have different runs in it (I think we end up checking that further down but we should be checking it up front). This loop is meant to be (1) pre-emptively catching the case where someone has associated the same dataId with different files (2) checking that if there are existing refs known to registry we overwrite the refs with expanded refs (since datastore needs those) (3) if the refs are not known to registry, making sure they are registered (and I think importDatasets returns expanded refs).

tests/test_butler.py
andy-slac: I made DM-39410 to check if it can be removed.

tests/test_butler.py
andy-slac: I need to drop collection here, but in general I agree that passing collection with a resolved ref does not make sense and can cause confusion. I'll add a warning.

python/lsst/daf/butler/_butler.py
andy-slac: It probably is now, it could have been relevant for some corner case before (e.g. directPut for integer dataset ID into UUID registry)? I think we can remove it now.

python/lsst/daf/butler/_butler.py
andy-slac: OK, I'll add a check for `datasore.exists()` and raise `LookupError` immediately (in both `getDeferred` and `getDirectDeferred`).

python/lsst/daf/butler/_butler.py
andy-slac: ✔️ I'm mostly surprised that `mypy` does not complain. VSCode actually tells me that `"group_key" is possibly unbound` on that line. I'll leave it as it is.

tests/test_butler.py
andy-slac: I'll fix the comment.

tests/test_butler.py
andy-slac: Run is not actually used with resolved refs, I'll drop it here.

andy-slac:
I think I resolved all comments, two new commits were added: 408ec68d and 75ed3cb6 (plus cfa6ae38 I mentioned already).

PR #843, opened 2023-05-26
DM-39429: Enable repository index in QBB and `butler config-dump`

andy-slac:
Extended `ButlerRepoIndex.get_repo_uri` method with a `default` parameter, if label is not found in an index then default is returned. Updated QBB and `configDump` to use that method to try to resolve repo in the index. Added a unit test for QBB, `config-dump` was tested manually. Also changed `Butler` use of that method to pass a default to simplify its logic.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -230,8 +230,10 @@ class Butler(LimitedButler):
             self._config: ButlerConfig = butler._config
         else:
             # Can only look for strings in the known repos list.
-            if isinstance(config, str) and config in self.get_known_repos():
-                config = str(self.get_repo_uri(config))
+            if isinstance(config, str):
+                # Somehow ButlerConfig fails in some cases if config is a
+                # ResourcePath, force it back to string here.
+                config = str(self.get_repo_uri(config, config))
             try:
                 self._config = ButlerConfig(config, searchPaths=searchPaths)
             except FileNotFoundError as e:
@@ -286,31 +288,37 @@ class Butler(LimitedButler):
             return None
 
     @classmethod
-    def get_repo_uri(cls, label: str) -> ResourcePath:
+    def get_repo_uri(cls, label: str, default: ResourcePath | str | None = None) -> ResourcePath:
         """Look up the label in a butler repository index.
 
         Parameters
         ----------
         label : `str`
             Label of the Butler repository to look up.
+        default : `lsst.resources.ResourcePath` or `str`, optional
+            If ``label`` cannot be found in the repository index (either
+            because index is not defined or ``label`` is not in the index) and
+            ``default`` is not `None` then return its value. If not specified
+            then an exception will be raised instead.
 
         Returns
         -------
         uri : `lsst.resources.ResourcePath`
-            URI to the Butler repository associated with the given label.
+            URI to the Butler repository associated with the given label or
+            default value if it is provided.
 
         Raises
         ------
         KeyError
             Raised if the label is not found in the index, or if an index
-            can not be found at all.
+            is not defined, and ``default`` is not specified.
 
         Notes
         -----
         See `~lsst.daf.butler.ButlerRepoIndex` for details on how the
         information is discovered.
         """
-        return ButlerRepoIndex.get_repo_uri(label)
+        return ButlerRepoIndex.get_repo_uri(label, default)
 
     @classmethod
     def get_known_repos(cls) -> Set[str]:
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -233,7 +233,7 @@ class Butler(LimitedButler):
             if isinstance(config, str):
                 # Somehow ButlerConfig fails in some cases if config is a
                 # ResourcePath, force it back to string here.
-                config = str(self.get_repo_uri(config, config))
+                config = str(self.get_repo_uri(config, True))
             try:
                 self._config = ButlerConfig(config, searchPaths=searchPaths)
             except FileNotFoundError as e:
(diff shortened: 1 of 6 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -230,8 +230,10 @@ def __init__(
             self._config: ButlerConfig = butler._config
         else:
             # Can only look for strings in the known repos list.
-            if isinstance(config, str) and config in self.get_known_repos():
-                config = str(self.get_repo_uri(config))
+            if isinstance(config, str):
+                # Somehow ButlerConfig fails in some cases if config is a
```
timj, line 234: Hmm. I wonder what that's about. I clearly knew at some point given I did force it back to str previously.

python/lsst/daf/butler/_butler.py
```diff
@@ -286,31 +288,37 @@ def _retrieve_dataset_type(self, name: str) -> DatasetType | None:
...
         """Look up the label in a butler repository index.
 
         Parameters
         ----------
         label : `str`
             Label of the Butler repository to look up.
+        default : `lsst.resources.ResourcePath` or `str`, optional
+            If ``label`` cannot be found in the repository index (either
+            because index is not defined or ``label`` is not in the index) and
+            ``default`` is not `None` then return its value. If not specified
```
timj: It doesn't return the default. It returns `ResourcePath(default)`

python/lsst/daf/butler/_butler.py
```diff
@@ -286,31 +288,37 @@ def _retrieve_dataset_type(self, name: str) -> DatasetType | None:
             return None
 
     @classmethod
-    def get_repo_uri(cls, label: str) -> ResourcePath:
+    def get_repo_uri(cls, label: str, default: ResourcePath | str | None = None) -> ResourcePath:
```
timj: `ResourcePathExpression` for default?

python/lsst/daf/butler/_butler.py
```diff
@@ -286,31 +288,37 @@ def _retrieve_dataset_type(self, name: str) -> DatasetType | None:
...
         """Look up the label in a butler repository index.
 
         Parameters
         ----------
         label : `str`
             Label of the Butler repository to look up.
+        default : `lsst.resources.ResourcePath` or `str`, optional
```
timj: I do wonder whether there is any use case for the second parameter not being identical to the first. Why would you ever have a default that differed? In which case I'm not sure we need this parameter and should either have a boolean that indicates whether to default to the original or not, or always do the defaulting.

python/lsst/daf/butler/_butlerRepoIndex.py
```diff
@@ -137,31 +137,45 @@ def get_known_repos(cls) -> Set[str]:
         return set(repo_index)
 
     @classmethod
-    def get_repo_uri(cls, label: str) -> ResourcePath:
+    def get_repo_uri(cls, label: str, default: ResourcePath | str | None = None) -> ResourcePath:
```
timj: Same comment for this as above in `Butler.get_repo_uri`.

tests/test_quantumBackedButler.py
```diff
@@ -133,6 +135,27 @@ def test_initialize(self) -> None:
         )
         self._test_factory(qbb)
 
+    def test_initialize_repo_index(self) -> None:
+        """Test for initialize using config file and repo index."""
+
+        # store config to a file
```
timj:
```suggestion
        # Store config to a file.
```

timj (approved):
Looks great but I'm torn as to the default parameter in get_repo_uri since I can't understand a use case for why you would ever default to anything other than the value you were testing against.

python/lsst/daf/butler/_butler.py
andy-slac: If i strip `str()` then one unit test fails with this exception:
```
ERROR: testConstructor (__main__.S3DatastoreButlerTestCase)
Independent test of constructor.
----------------------------------------------------------------------
  File "/home/salnikov/lsst/gen3/daf_butler/tests/test_butler.py", line 548, in testConstructor
    butler = Butler(config_dir, run=self.default_run)
  File "/home/salnikov/lsst/gen3/daf_butler/python/lsst/daf/butler/_butler.py", line 238, in __init__
    self._config = ButlerConfig(config, searchPaths=searchPaths)
  File "/home/salnikov/lsst/gen3/daf_butler/python/lsst/daf/butler/_butlerConfig.py", line 95, in __init__
    uri = ResourcePath(other, forceDirectory=True)
  File "/u1/lsst/lsst_stack-6.0.0-new/conda/envs/lsst-scipipe-6.0.0/share/eups/Linux64/resources/geaf6fda20e+37bc81c514/python/lsst/resources/_resourcePath.py", line 214, in __new__
    raise RuntimeError(
RuntimeError: s3://anybucketname/6A6JK5LH67CLP5WFLC3U is already a file-like ResourcePath; cannot force it to directory.
```

python/lsst/daf/butler/_butler.py
andy-slac: I do not think there is any other use case now, but I wanted to keep it general enough for any future use cases and also similar to other Python patterns where getters accept a default value.

andy-slac: But I am also OK with the boolean flag instead of default. I already tried to always-default, but that breaks some unit tests in a bit of non-trivial way.

timj: Okay. I think I prefer boolean.

python/lsst/daf/butler/_butler.py
timj: Ah. Right. The old problem with an arbitrary S3 URI not being able to tell the difference between being a file or a directory. Whereas downstream the defaulting code can say "does not end in butler.yaml therefore it must be a directory" and force it to be a directory -- which it can't easily do if it's already a ResourcePath which doesn't look like a directory (although maybe I could add some logic for that to work).

andy-slac:
OK, I replaced default vaue with a boolean flag: [5e4a279](https://github.com/lsst/daf_butler/pull/843/commits/5e4a2791e49a2d8ac059c5f4b4044eafa36ff9c2)

PR #844, opened 2023-05-28
DM-39434: Avoid defaultdict with lambda in pickled dataclasses

andy-slac:
Pickle does not like lambdas or local/nested methods, using a `dataclass.field`
with defaultdict and default_factory as a lambda breaks pickle. Instead of defining
a module-level factory method I decided to use a plain dict and its `setdefault` method.
Also had to add explicit pickle method to StoredFileInfo, being a frozen
dataclass it does not work by default with pickle.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/datastoreRecordData.py
+++ b/python/lsst/daf/butler/core/datastoreRecordData.py
@@ -27,8 +27,8 @@ __all__ = ("DatastoreRecordData", "SerializedDatastoreRecordData")
 
 import dataclasses
 import uuid
-from collections import defaultdict
-from typing import TYPE_CHECKING, AbstractSet, Any, Dict, List, Optional, Union
+from collections.abc import Mapping
+from typing import TYPE_CHECKING, Any
 
 from lsst.utils import doImportType
 from lsst.utils.introspection import get_full_type_name
@@ -41,24 +41,24 @@ from .storedFileInfo import StoredDatastoreItemInfo
 if TYPE_CHECKING:
     from ..registry import Registry
 
-_Record = Dict[str, Any]
+_Record = dict[str, Any]
 
 
 class SerializedDatastoreRecordData(BaseModel):
     """Representation of a `DatastoreRecordData` suitable for serialization."""
 
-    dataset_ids: List[uuid.UUID]
+    dataset_ids: list[uuid.UUID]
     """List of dataset IDs"""
 
-    records: Dict[str, Dict[str, List[_Record]]]
+    records: Mapping[str, Mapping[str, list[_Record]]]
     """List of records indexed by record class name and table name."""
 
     @classmethod
     def direct(
         cls,
         *,
-        dataset_ids: List[Union[str, uuid.UUID]],
-        records: Dict[str, Dict[str, List[_Record]]],
+        dataset_ids: list[str | uuid.UUID],
+        records: dict[str, dict[str, list[_Record]]],
     ) -> SerializedDatastoreRecordData:
         """Construct a `SerializedDatastoreRecordData` directly without
         validators.
@@ -92,8 +92,8 @@ class DatastoreRecordData:
     datastore.
     """
 
-    records: defaultdict[DatasetId, defaultdict[str, List[StoredDatastoreItemInfo]]] = dataclasses.field(
-        default_factory=lambda: defaultdict(lambda: defaultdict(list))
+    records: dict[DatasetId, dict[str, list[StoredDatastoreItemInfo]]] = dataclasses.field(
+        default_factory=dict
     )
     """Opaque table data, indexed by dataset ID and grouped by opaque table
     name."""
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/core/datastoreRecordData.py
```diff
@@ -41,24 +41,24 @@
...
 
 class SerializedDatastoreRecordData(BaseModel):
     """Representation of a `DatastoreRecordData` suitable for serialization."""
 
-    dataset_ids: List[uuid.UUID]
+    dataset_ids: list[uuid.UUID]
     """List of dataset IDs"""
 
-    records: Dict[str, Dict[str, List[_Record]]]
+    records: Mapping[str, Mapping[str, list[_Record]]]
```
timj, line 53: Isn't it actually a dict though? What's the gain from using Mapping here?

python/lsst/daf/butler/core/datastoreRecordData.py
```diff
@@ -133,9 +133,7 @@ def subset(self, dataset_ids: AbstractSet[DatasetId]) -> Optional[DatastoreRecor
         Records in the returned instance are shared with this instance, clients
         should not update or extend records in the returned instance.
         """
-        matching_records: defaultdict[
-            DatasetId, defaultdict[str, List[StoredDatastoreItemInfo]]
-        ] = defaultdict(lambda: defaultdict(list))
+        matching_records: dict[DatasetId, dict[str, list[StoredDatastoreItemInfo]]] = {}
```
timj, line 136: Missing test?

python/lsst/daf/butler/core/storedFileInfo.py
```diff
@@ -252,3 +252,6 @@ def update(self, **kwargs: Any) -> StoredFileInfo:
         if kwargs:
             raise ValueError(f"Unexpected keyword arguments for update: {', '.join(kwargs)}")
         return type(self)(**new_args)
+
+    def __reduce__(self) -> str | tuple[Any, ...]:
+        return (self.from_record, (self.to_record(),))
```
timj, line 257: We probably should have a small round trip pickle test.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -2934,12 +2934,11 @@ def export_records(self, refs: Iterable[DatasetIdRef]) -> Mapping[str, Datastore
         # Docstring inherited from the base class.
         exported_refs = list(self._bridge.check(refs))
         ids = {ref.id for ref in exported_refs}
-        records: defaultdict[DatasetId, defaultdict[str, List[StoredDatastoreItemInfo]]] = defaultdict(
-            lambda: defaultdict(list), {id: defaultdict(list) for id in ids}
-        )
+        records: dict[DatasetId, dict[str, list[StoredDatastoreItemInfo]]] = {id: {} for id in ids}
```
timj, line 2937: Isn't it all a bit annoying that we can't use defaultdict specifically for this class?

timj (approved):
I forget why we can't use a single pydantic model for this and instead have to have a pydantic model and a dataclass. Is it for performance reasons? Presumably the pydantic model would pickle just fine.

Can you add a bit more detail about having to drop defaultdict in the commit message?

python/lsst/daf/butler/core/datastoreRecordData.py
andy-slac: Indeed, the implementation uses `dict` but the clients do not need/want to modify it, so for everyone else an immutable mapping should be OK. I just miss `const` from C++ 🙂

timj: Aha. I understand. We are using the annotation to make sure that mypy complains if someone tries to write to the dict.

python/lsst/daf/butler/core/datastoreRecordData.py
andy-slac: I can't see that this method is actually used by anything. I can add a test to `test_datastore.py` together with pickle test.

python/lsst/daf/butler/datastores/fileDatastore.py
andy-slac: I think we can use `defaultdict` but we cannot use lambdas or nested methods. Defining a module-level factory method just for this feels more complicated than just sticking to good old `setdefault`.

python/lsst/daf/butler/core/storedFileInfo.py
andy-slac: I added tests for both pickling of StoredFileInfo and DatastoreRecordData.subset().

andy-slac:
> I forget why we can't use a single pydantic model for this and instead have to have a pydantic model and a dataclass. Is it for performance reasons? Presumably the pydantic model would pickle just fine.

I'm not sure what the exact reason was, but I see that pydantic and dataclass have different structure, maybe pydantic is more YAML/JSON-friedly than the in-memory dataclass?

> Can you add a bit more detail about having to drop defaultdict in the commit message?

I updated the commit message.

PR #846, opened 2023-06-02
DM-38916: Link to bind documentation from query methods

kfindeisen:
This PR adds cross-links from the `bind` parameter of `Registry.query*` methods to the general query guide. Most of the diff is fixing more general documentation issues found in the process.

### DIFF
--- a/python/lsst/daf/butler/registry/_registry.py
+++ b/python/lsst/daf/butler/registry/_registry.py
@@ -93,8 +93,9 @@ class Registry(ABC):
     The base class will look for a ``cls`` entry and call that specific
     `fromConfig()` method.
 
-    All subclasses should store `RegistryDefaults` in a ``_defaults``
-    property. No other properties are assumed shared between implementations.
+    All subclasses should store `~lsst.daf.butler.registry.RegistryDefaults` in
+    a ``_defaults`` property. No other properties are assumed shared between
+    implementations.
     """
 
     defaultConfigFile: Optional[str] = None
@@ -203,7 +204,7 @@ class Registry(ABC):
         writeable: bool = True,
         defaults: Optional[RegistryDefaults] = None,
     ) -> Registry:
-        """Create `Registry` subclass instance from `config`.
+        """Create `Registry` subclass instance from ``config``.
 
         Registry database must be initialized prior to calling this method.
 
@@ -211,11 +212,11 @@ class Registry(ABC):
         ----------
         config : `ButlerConfig`, `RegistryConfig`, `Config` or `str`
             Registry configuration
-        butlerRoot : `lsst.resources.ResourcePathExpression`, optional
+        butlerRoot : convertible to `lsst.resources.ResourcePath`, optional
             Path to the repository root this `Registry` will manage.
         writeable : `bool`, optional
             If `True` (default) create a read-write connection to the database.
-        defaults : `RegistryDefaults`, optional
+        defaults : `~lsst.daf.butler.registry.RegistryDefaults`, optional
             Default collection search path and/or output `~CollectionType.RUN`
             collection.
 
@@ -250,7 +251,7 @@ class Registry(ABC):
 
         Parameters
         ----------
-        defaults : `RegistryDefaults`, optional
+        defaults : `~lsst.daf.butler.registry.RegistryDefaults`, optional
             Default collections and data ID values for the new registry.  If
             not provided, ``self.defaults`` will be used (but future changes
             to either registry's defaults will not affect the other).
@@ -280,7 +281,7 @@ class Registry(ABC):
     @property
     def defaults(self) -> RegistryDefaults:
         """Default collection search path and/or output `~CollectionType.RUN`
-        collection (`RegistryDefaults`).
+        collection (`~lsst.daf.butler.registry.RegistryDefaults`).
 
         This is an immutable struct whose components may not be set
         individually, but the entire struct can be set by assigning to this
@@ -367,7 +368,7 @@ class Registry(ABC):
 
         Raises
         ------
-        MissingCollectionError
+        lsst.daf.butler.registry.MissingCollectionError
             Raised if no collection with the given name exists.
         """
         raise NotImplementedError()
@@ -423,9 +424,9 @@ class Registry(ABC):
 
         Raises
         ------
-        MissingCollectionError
+        lsst.daf.butler.registry.MissingCollectionError
             Raised if no collection with the given name exists.
-        sqlalchemy.IntegrityError
+        sqlalchemy.exc.IntegrityError
             Raised if the database rows associated with the collection are
             still referenced by some other table, such as a dataset in a
             datastore (for `~CollectionType.RUN` collections only) or a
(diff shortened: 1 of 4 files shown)


timj:
See also the same fixes for the sphinx errors on #840. I guess it depends which PR merges first.

kfindeisen:
Oh, sorry. I thought you said that you only got duplication warnings (which I've ignored here).

timj:
Sorry. I said that I fixed all the non-duplication warnings yesterday. Your PR has many more fixes than those warnings but there is some duplication.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -203,19 +204,19 @@ def fromConfig(
...
+        """Create `Registry` subclass instance from ``config``.
 
         Registry database must be initialized prior to calling this method.
 
         Parameters
         ----------
         config : `ButlerConfig`, `RegistryConfig`, `Config` or `str`
             Registry configuration
-        butlerRoot : `lsst.resources.ResourcePathExpression`, optional
+        butlerRoot : convertible to `lsst.resources.ResourcePath`, optional
```
timj, line 215: There are many places where we refer to `ResourcePathExpression` so I wonder about this change but I see there is precedent in this file.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -614,7 +614,7 @@ def removeDatasetType(self, name: str | tuple[str, ...]) -> None:
 
         Parameters
         ----------
-        name : `str` or `tuple[str, ...]`
+        name : `str` or `tuple` [`str`]
```
timj, line 617: The ellipsis is important. It says a tuple with any number of strings, versus a tuple of a single string.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -870,7 +870,7 @@ def getDataset(self, id: DatasetId) -> Optional[DatasetRef]:
 
         Parameters
         ----------
-        id : `DatasetId`
+        id : `uuid.UUID`
```
timj: Formally this really should be a `lsst.daf.butler.DatasetId` which is a type alias for UUID. It seems that Sphinx handles the aliasing just fine in the formal type annotations but gets confused in the numpydoc and doesn't link it properly.

@TallJimbo are we formally deprecating `DatasetId` or do we want to continue to use it as a public-facing alias?

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -1385,27 +1388,27 @@ def queryDatasets(
...
         -------
-        refs : `queries.DatasetQueryResults`
+        refs : `registry.queries.DatasetQueryResults`
             Dataset references matching the given query criteria.  Nested data
             IDs are guaranteed to include values for all implied dimensions
             (i.e. `DataCoordinate.hasFull` will return `True`), but will not
             include dimension records (`DataCoordinate.hasRecords` will be
-            `False`) unless `~queries.DatasetQueryResults.expanded` is called
-            on the result object (which returns a new one).
+            `False`) unless `~registry.queries.DatasetQueryResults.expanded` is
```
timj:
```suggestion
            `False`) unless `~.queries.DatasetQueryResults.expanded` is
```

This seems to work for me without having to put the `registry` in the path. Similarly above on line 1391.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -1500,30 +1506,31 @@ def queryDataIds(
 
         Returns
         -------
-        dataIds : `queries.DataCoordinateQueryResults`
+        dataIds : `registry.queries.DataCoordinateQueryResults`
```
timj:
```suggestion
        dataIds : `.queries.DataCoordinateQueryResults`
```
and other places.

python/lsst/daf/butler/registry/_registry.py
```diff
@@ -1658,17 +1668,18 @@ def queryDatasetAssociations(
...
+        lsst.daf.butler.registry.CollectionExpressionError
             Raised when ``collections`` expression is invalid.
         """
         raise NotImplementedError()
 
     @property
     def obsCoreTableManager(self) -> ObsCoreTableManager | None:
-        """ObsCore manager instance for this registry (`ObsCoreTableManager`
+        """ObsCore manager instance for this registry
+        (`~lsst.daf.butler.registry.interfaces.ObsCoreTableManager`
```
timj:
```suggestion
        (`~.interfaces.ObsCoreTableManager`
```

timj (approved):
Thanks for doing this. I'd like @TallJimbo or @andy-slac to comment on the `DatasetId` issue as I'm not completely sure we are ready to drop that from the public APIs and replace it with an explicit UUID.

python/lsst/daf/butler/registry/_registry.py
kfindeisen: Yes, there are, but [the link can't work](https://pipelines.lsst.io/v/d_2023_06_05/py-api/lsst.daf.butler.registry.Registry.html#lsst.daf.butler.registry.Registry.fromConfig) because `ResourcePathExpression` is not a documented type. The same goes for any other Mypy-helper types.

It's not obvious to me, at least, what does or doesn't count as a "ResourcePath expression", so this is one case where I think it makes sense for the human-readable description to deviate from the machine-readable one. I'll admit `DataId` is a case where the meaning should be clear without explanation.

~EDIT: hmm, looks like I missed the `ButlerConfig` and `Config` links~ 😕

python/lsst/daf/butler/registry/_registry.py
kfindeisen: Except that the "tuple" is logically being used as a list, not as a tuple, and [the convention for lists](https://developer.lsst.io/python/numpydoc.html#sequence-types) is pretty clear.

(IMO this method should take any iterable, but that's a code issue, not a documentation formatting issue.)

python/lsst/daf/butler/registry/_registry.py
andy-slac: Personally I'd prefer to have an abstraction with the name `DatasetId`. And I think that having an actual type for it would be even better, we can then have all useful methods (e.g. factories) in that class. UUID will likely to stay as the only implementation for it, but our clients do not need to care about that detail.

kfindeisen: FTR, https://pipelines.lsst.io/v/d_2023_06_05/modules/lsst.daf.butler/index.html#classes includes `DatasetId` in its class list, but doesn't have a valid anchor for it (Sphinx warns about this as well; I'd assumed it couldn't be fixed short of removing all references to `DatasetId`). If the documentation of `DatasetId` could be fixed so that it's linkable, then I'd be fine with leaving these references as-is.

timj: Ok. Leave it as it was originally and we can try to fix DatasetId as @andy-slac describes above. There is a doc string attached to `DatasetId` in `ref.py` but clearly that doesn't do anything useful.

python/lsst/daf/butler/registry/_registry.py
kfindeisen: Interesting: it looks like there are *two* `Registry` pages, https://pipelines.lsst.io/v/d_2023_06_05/py-api/lsst.daf.butler.registry.Registry.html and https://pipelines.lsst.io/v/d_2023_06_05/py-api/lsst.daf.butler.Registry.html, and the links work on one and are broken on the other. A good argument for avoiding unqualified links. 😬

timj: Yes, this is one of the things that Sphinx complains about.

> docstring of lsst.daf.butler.registry._registry.Registry:1: WARNING: duplicate object description of lsst.daf.butler.registry._registry.Registry, other instance in py-api/lsst.daf.butler.Registry, use :noindex: for one of them

python/lsst/daf/butler/registry/_registry.py
TallJimbo: FWIW, I'm more in the "UUID is going to the only implementation forever and we should just use that type directly" camp.  But I certainly agree this isn't the ticket for it.

PR #848, opened 2023-06-06
DM-37499: Improve log message for datastore.mexists

timj:
* Use different log message for single file vs single chunk vs multiple chunks.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -1525,14 +1525,54 @@ class FileDatastore(GenericBaseDatastore):
         n_chunks = 0
         for chunk in chunk_iterable(refs, chunk_size=chunk_size):
             chunk_result = self._mexists(chunk, artifact_existence)
-            if log.isEnabledFor(VERBOSE):
-                n_results = len(chunk_result)
-                n_checked += n_results
+
+            # The log message level and content depend on how many
+            # datasets we are processing.
+            n_results = len(chunk_result)
+
+            # Use verbose logging to ensure that messages can be seen
+            # easily if many refs are being checked.
+            log_threshold = VERBOSE
+            n_checked += n_results
+
+            # This sum can take some time so only do it if we know the
+            # result is going to be used.
+            n_found = 0
+            if log.isEnabledFor(log_threshold):
                 # Can treat the booleans as 0, 1 integers and sum them.
                 n_found = sum(chunk_result.values())
                 n_found_total += n_found
-                log.verbose(
-                    "Number of datasets found in datastore for chunk %d = %d/%d (running total: %d/%d)",
+
+            # We are deliberately not trying to count the number of refs
+            # provided in case it's in the millions. This means there is a
+            # situation where the number of refs exactly matches the chunk
+            # size and we will switch to the multi-chunk path even though
+            # we only have a single chunk.
+            if n_results < chunk_size and n_chunks == 0:
+                # Single chunk will be processed so we can provide more detail.
+                if n_results == 1:
+                    ref = list(chunk_result)[0]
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1525,14 +1525,52 @@ def mexists(
...
+            # provided in case it's in the millions. This means there is a
+            # situation where the number of refs exactly matches the chunk
+            # size and we will switch to the multi-chunk path even though
+            # we only have a single chunk.
+            if n_results < chunk_size and n_chunks == 0:
+                # Single chunk will be processed so we can provide more detail.
+                if n_results == 1:
+                    ref = list(chunk_result)[0]
+                    log.log(
+                        log_threshold,
```
timj: It's possible this should be a DEBUG message since we don't log anything for the simple `datastore.exists()` API. The  multi-chunk case if VERBOSE because it takes a significant amount of time to run and it's good to have feedback. The middle ground is the single chunk and whether the verbose vs debug depends on the amount of time that it took to do the existence check (9999 datasets being checked will obviously take much longer than 2 datasets).

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1525,14 +1525,52 @@ def mexists(
...
+                        "Calling mexists() with single ref that does%s exist (%s).",
+                        "" if chunk_result[ref] else " not",
+                        ref,
+                    )
+                else:
+                    # Single chunk but multiple files. Summarize.
+                    log.log(
+                        log_threshold,
```
timj, line 1564: Maybe use VERBOSE here if the chunk took more than 10 seconds, else DEBUG? cc/ @TallJimbo

andy-slac: Changing logging level depending on how long the code runs may be confusing, people may be surprised if messages start appearing/disappearing randomly.

timj: So leave it so that you get a debug log message if you have one dataset but verbose log message if you have more than one.

python/lsst/daf/butler/datastores/fileDatastore.py
andy-slac: I don't have strong opinion, but I'm a big fan of trying to keep things simple. 🙂 Logging everything at VERBOSE may be OK, as I think most people will run it with INFO.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: I should add that we use `mexists()` in `SingleQuantumExecutor` and also turn on VERBOSE logging in workflows, so we are going to get this log message from there every time a quantum has more than one input (and most of the time there won't be 10,000 inputs). The 10,000 path comes up with `butler transfer-datasets` at the end of the workflow.

andy-slac (approved):
Looks good, it took me some time to verify that logic works fine, but I hope it's not goint to be updated frequently.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: We always run with VERBOSE during workflow execution (mainly because we need to issue log messages when things take a while in order to let people know that the task is still running).

timj:
I didn't have an easy way to test the chunking (and no programmatic way to shrink the chunk size) so that code path doesn't trigger in the tests.

python/lsst/daf/butler/datastores/fileDatastore.py
andy-slac: OK, if VERBOSE is the default for many cases then it should be OK to avoid logging for single ref case.

PR #849, opened 2023-06-07
DM-39563: Make clearer error message when repo alias has failed to find something

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butlerRepoIndex.py
+++ b/python/lsst/daf/butler/_butlerRepoIndex.py
@@ -191,13 +191,13 @@ class ButlerRepoIndex:
             repo_index = cls._read_repository_index_from_environment()
         except KeyError:
             if return_label:
-                return ResourcePath(label)
+                return ResourcePath(label, forceAbsolute=False)
             raise
 
         repo_uri = repo_index.get(label)
         if repo_uri is None:
             if return_label:
-                return ResourcePath(label)
+                return ResourcePath(label, forceAbsolute=False)
             # This should not raise since it worked earlier.
             try:
                 index_uri = str(cls._get_index_uri())
--- a/python/lsst/daf/butler/_butlerRepoIndex.py
+++ b/python/lsst/daf/butler/_butlerRepoIndex.py
@@ -85,7 +85,11 @@ class ButlerRepoIndex:
         if index_uri in cls._cache:
             return cls._cache[uri]
 
-        repo_index = Config(uri)
+        try:
+            repo_index = Config(uri)
+        except FileNotFoundError as e:
+            # More explicit error message.
+            raise FileNotFoundError(f"Butler repository index file not found at {uri}.") from e
         cls._cache[uri] = repo_index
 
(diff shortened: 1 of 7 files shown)


python/lsst/daf/butler/_butlerRepoIndex.py
```diff
@@ -136,6 +140,23 @@ def get_known_repos(cls) -> Set[str]:
...
+
+        Returns
+        -------
+        reason : `str`
+            If there is a problem reading the repository index, this will
+            contain a string with an explanation. Empty string if everything
+            worked.
+        """
+        try:
+            cls._read_repository_index_from_environment()
```
andy-slac: Hmm, I guess this assumes that it is supposed to fail in exactly the same way every time? With transient filesystem issues it may not be true, would it be better to save the original exception message and return it here?

tests/test_butler.py
```diff
@@ -101,6 +102,18 @@ def mock_s3(cls):
...
+def clean_environment() -> None:
+    """Remove external environment variables that affect the tests."""
+    for k in (
+        "DAF_BUTLER_REPOSITORY_INDEX",
+        "S3_ENDPOINT_URL",
+        "AWS_ACCESS_KEY_ID",
+        "AWS_SECRET_ACCESS_KEY",
+        "AWS_SHARED_CREDENTIALS_FILE",
+    ):
+        os.environ.pop(k, None)
```
andy-slac, line 115: Would it make sense to make this more reusable and move to `python/lsst/daf/butler/tests/utils.py`?

andy-slac (approved):
Looks good, a couple of minor questions. I agree that you want to catch a wider set of exceptions when reading index.

tests/test_butler.py
timj: I'm half inclined to remove the S3 code from here completely but we need one remote datastore for testing and we don't use the webdav one by default. There is `lsst.resources.s3utils` that tries to have something but it's not quite what we need. I probably should be trying to make that work for the S3 use case at USDF (where pytest doesn't work without this change). The DAF_BUTLER_REPOSITORY_INDEX clean up is really only needed for this one test of `testConstructor` and a general utility for the S3 tests in butler is only relevant for this one file.

python/lsst/daf/butler/_butlerRepoIndex.py
timj: Hmm. There is a cache for the contents of the URI so I probably could cache the contents of the exception in a different cache in the class. That should be okay for a transient failure assuming people don't call this out of sequence.

timj:
I did just push a tweak which catches `Exception` and rewrites it to make it a bit clearer where it's coming from.

python/lsst/daf/butler/_butlerRepoIndex.py
timj: Okay. I've changed it to a simple cache that resets on each call and that seems to work okay.

PR #850, opened 2023-06-08
DM-39605: Use butler.dimensions rather than butler.registry.dimensions

timj:
This is the preferred interface and is part of the long-term goal of lowering the visibility of butler.registry.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -118,7 +118,7 @@ class Butler(LimitedButler):
         datastore as the given one, but with the given collection and run.
         Incompatible with the ``config``, ``searchPaths``, and ``writeable``
         arguments.
-    collections : `str` or `Iterable` [ `str` ], optional
+    collections : `str` or `~collections.abc.Iterable` [ `str` ], optional
         An expression specifying the collections to be searched (in order) when
         reading datasets.
         This may be a `str` collection name or an iterable thereof.
@@ -1785,7 +1785,7 @@ class Butler(LimitedButler):
 
         Parameters
         ----------
-        names : `Iterable` [ `str` ]
+        names : `~collections.abc.Iterable` [ `str` ]
             The names of the collections to remove.
         unstore : `bool`, optional
             If `True` (default), delete datasets from all datastores in which
@@ -2596,7 +2596,7 @@ class Butler(LimitedButler):
     @property
     def collections(self) -> Sequence[str]:
         """The collections to search by default, in order
-        (`Sequence` [ `str` ]).
+        (`~collections.abc.Sequence` [ `str` ]).
 
         This is an alias for ``self.registry.defaults.collections``.  It cannot
         be set directly in isolation, but all defaults may be changed together
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -729,7 +729,7 @@ class Butler(LimitedButler):
             for dimensionName in list(dataIdDict):
                 value = dataIdDict[dimensionName]
                 try:
-                    dimension = self.registry.dimensions.getStaticDimensions()[dimensionName]
+                    dimension = self.dimensions.getStaticDimensions()[dimensionName]
                 except KeyError:
                     # This is not a real dimension
                     not_dimensions[dimensionName] = value
@@ -801,7 +801,7 @@ class Butler(LimitedButler):
             # fail but they are going to fail anyway because of the
             # ambiguousness of the dataId...
             if datasetType.isCalibration():
-                for dim in self.registry.dimensions.getStaticDimensions():
+                for dim in self.dimensions.getStaticDimensions():
                     if dim.temporal:
                         candidateDimensions.add(str(dim))
 
@@ -817,7 +817,7 @@ class Butler(LimitedButler):
             # given names with records within those dimensions
             matched_dims = set()
             for dimensionName in candidateDimensions:
-                dimension = self.registry.dimensions.getStaticDimensions()[dimensionName]
+                dimension = self.dimensions.getStaticDimensions()[dimensionName]
                 fields = dimension.metadata.names | dimension.uniqueKeys.names
                 for field in not_dimensions:
                     if field in fields:
@@ -940,8 +940,8 @@ class Butler(LimitedButler):
                         # by the instrument.
                         if (
                             dimensionName == "visit"
-                            and "visit_system_membership" in self.registry.dimensions
-                            and "visit_system" in self.registry.dimensions["instrument"].metadata
+                            and "visit_system_membership" in self.dimensions
+                            and "visit_system" in self.dimensions["instrument"].metadata
                         ):
                             instrument_records = list(
                                 self.registry.queryDimensionRecords(
@@ -992,7 +992,7 @@ class Butler(LimitedButler):
                         )
 
                 # Get the primary key from the real dimension object
-                dimension = self.registry.dimensions.getStaticDimensions()[dimensionName]
+                dimension = self.dimensions.getStaticDimensions()[dimensionName]
                 if not isinstance(dimension, Dimension):
                     raise RuntimeError(
                         f"{dimension.name} is not a true dimension, and cannot be used in data IDs."
@@ -1072,7 +1072,7 @@ class Butler(LimitedButler):
             # dimensions that provide temporal information for a validity-range
             # lookup.
             dataId = DataCoordinate.standardize(
-                dataId, universe=self.registry.dimensions, defaults=self.registry.defaults.dataId, **kwargs
+                dataId, universe=self.dimensions, defaults=self.registry.defaults.dataId, **kwargs
             )
             if dataId.graph.temporal:
                 dataId = self.registry.expandDataId(dataId)
@@ -2150,7 +2150,7 @@ class Butler(LimitedButler):
             raise ValueError(f"Unknown export format {format!r}, allowed: {','.join(formats.keys())}")
         BackendClass = get_class_of(formats[format, "export"])
         with open(filename, "w") as stream:
-            backend = BackendClass(stream, universe=self.registry.dimensions)
+            backend = BackendClass(stream, universe=self.dimensions)
             try:
                 helper = RepoExportContext(
                     self.registry, self.datastore, backend=backend, directory=directory, transfer=transfer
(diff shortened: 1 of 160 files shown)


timj:
This turned into a gigantic PR. The type annotations and f-string and set changes are distinct commits if that helps. mypy and flake8 are happy and all tests pass.

timj:
I've also replaced `typing.AbstractSet` with `collections.abc.Set` and fixed the pydocstyle problems.

andy-slac:
@TallJimbo, should we split this review in two? I could look at `python/lsst/daf/butler/core/...` and `python/lsst/daf/butler/registry/...` and leave the rest to you?

TallJimbo:
@andy-slac , that works for me.

python/lsst/daf/butler/core/configSupport.py
```diff
@@ -67,11 +67,11 @@ class LookupKey:
 
     def __init__(
         self,
-        name: Optional[str] = None,
-        dimensions: Optional[Iterable[Union[str, Dimension]]] = None,
-        dataId: Optional[Dict[str, Any]] = None,
+        name: str | None = None,
+        dimensions: Iterable[str | Dimension] | None = None,
```
andy-slac, line 71: I think code also expects `DimensionGraph` as parameter here.

python/lsst/daf/butler/core/timespan.py
```diff
@@ -28,20 +28,8 @@
...
-    List,
-    Mapping,
-    Optional,
-    Tuple,
-    Type,
-    TypeVar,
-    Union,
-)
+from collections.abc import Generator, Mapping
+from typing import TYPE_CHECKING, Any, ClassVar, TypeVar, Union
```
andy-slac, line 32: Can't get rid of this `Union`?

python/lsst/daf/butler/registry/queries/_structs.py
```diff
@@ -148,7 +148,7 @@ def combine(
     governor_constraints: Mapping[str, Set[str]]
     """Restrictions on the values governor dimensions can take in this query,
     imposed by the string expression and/or data ID
-    (`Mapping` [ `set`,  `~collections.abc.Set` [ `str` ] ]).
+    (`~collections.abc.Mapping` [ `set`,  `~collections.abc.Set` [ `str` ] ]).
```
andy-slac: `set` should be `str`.

andy-slac (approved):
Checked `butler/core`, `butler/registry`, and `butler/registries`, looks good, a couple of minor comments. (@TallJimbo, you can skip `registries/...`)

python/lsst/daf/butler/core/configSupport.py
TallJimbo: `DimensionGraph` should already be matched by `Iterable[Dimension]`.

python/lsst/daf/butler/_butler.py
```diff
@@ -330,12 +314,12 @@ def get_known_repos(cls) -> Set[str]:
     @staticmethod
     def makeRepo(
         root: ResourcePathExpression,
-        config: Union[Config, str, None] = None,
-        dimensionConfig: Union[Config, str, None] = None,
+        config: Config | str | None = None,
+        dimensionConfig: Config | str | None = None,
         standalone: bool = False,
-        searchPaths: Optional[List[str]] = None,
+        searchPaths: list[str] | None = None,
```
TallJimbo, line 320: I'm guessing many of these `str` values should be `ResourcePathExpression`, and for public APIs taking `Sequence`  (or `Iterable`, if possible) is much nicer than taking `list`.  In fact, I think if you try to make `searchPath` take a list of `ResourcePathExpression` it may not work because `list` is not covariant.

python/lsst/daf/butler/_butler.py
```diff
@@ -2081,10 +2065,10 @@ def ingest(
     def export(
         self,
         *,
-        directory: Optional[str] = None,
-        filename: Optional[str] = None,
-        format: Optional[str] = None,
-        transfer: Optional[str] = None,
+        directory: str | None = None,
+        filename: str | None = None,
```
TallJimbo, line 2069: I wonder if either of these could be `ResourcePathExpression`, too.

python/lsst/daf/butler/_limited_butler.py
```diff
@@ -208,7 +209,7 @@ def getDirectDeferred(
         self,
         ref: DatasetRef,
         *,
-        parameters: Union[dict, None] = None,
+        parameters: dict | None = None,
```
TallJimbo:
```suggestion
        parameters: dict[str, Any] | None = None,
```
for consistency with the rest of this file.

python/lsst/daf/butler/_quantum_backed.py
```diff
@@ -423,7 +426,7 @@ def getDirectDeferred(
         self,
         ref: DatasetRef,
         *,
-        parameters: Union[dict, None] = None,
+        parameters: dict | None = None,
```
TallJimbo:
```suggestion
        parameters: dict[str, Any] | None = None,
```

python/lsst/daf/butler/cli/butler.py
```diff
@@ -268,7 +268,7 @@ def _getPluginCommands(cls) -> defaultdict[str, list[str]]:
         commands: defaultdict[str, list[str]] = defaultdict(list)
         for pluginName in cls.getPluginList():
             try:
-                with open(pluginName, "r") as resourceFile:
+                with open(pluginName) as resourceFile:
```
TallJimbo, line 271: I'm not a fan of this kind of this change.  Explicit is better than implicit.

But then, I like to iterate over `.keys()` for the same reason, and I know you don't, so :man_shrugging: .

python/lsst/daf/butler/core/composites.py
```diff
@@ -77,7 +77,7 @@ class CompositesMap:
         in lookup keys.
     """
 
-    def __init__(self, config: Union[str, ButlerConfig, CompositesConfig], *, universe: DimensionUniverse):
+    def __init__(self, config: str | ButlerConfig | CompositesConfig, *, universe: DimensionUniverse):
```
TallJimbo:
```suggestion
    def __init__(self, config: ResourcePathExpression | ButlerConfig | CompositesConfig, *, universe: DimensionUniverse):
```
?

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -178,7 +179,7 @@ def setConfigRoot(cls, root: str, config: Config, full: Config, overwrite: bool
 
     def __init__(
         self,
-        config: Union[Config, str],
+        config: Config | str,
```
TallJimbo:
```suggestion
        config: Config | ResourcePathExpression,
```

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -104,7 +90,7 @@ class _IngestPrepData(Datastore.IngestPrepData):
         Files to be ingested by this datastore.
     """
 
-    def __init__(self, datasets: List[FileDataset]):
+    def __init__(self, datasets: list[FileDataset]):
```
TallJimbo:
```suggestion
    def __init__(self, datasets: Iterable[FileDataset]):
```
(+ associated doc change)

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -248,7 +234,7 @@ def makeTableSpec(cls, datasetIdColumnType: type) -> ddl.TableSpec:
 
     def __init__(
         self,
-        config: Union[DatastoreConfig, str],
+        config: DatastoreConfig | str,
```
TallJimbo:
```suggestion
        config: DatastoreConfig | ResourcePathExpression,
```

TallJimbo (approved):

python/lsst/daf/butler/core/timespan.py
timj: mypy hates it if I do that:

```
$ mypy python
python/lsst/daf/butler/core/timespan.py:136: error: Variable "lsst.daf.butler.core.timespan.TimespanBound" is not valid as a type  [valid-type]
python/lsst/daf/butler/core/timespan.py:136: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
python/lsst/daf/butler/core/timespan.py:137: error: Variable "lsst.daf.butler.core.timespan.TimespanBound" is not valid as a type  [valid-type]
python/lsst/daf/butler/core/timespan.py:137: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
python/lsst/daf/butler/core/timespan.py:146: error: Statement is unreachable  [unreachable]
python/lsst/daf/butler/core/timespan.py:157: error: Statement is unreachable  [unreachable]
python/lsst/daf/butler/core/timespan.py:169: error: Right operand of "and" is never evaluated  [unreachable]
python/lsst/daf/butler/core/timespan.py:171: error: Right operand of "and" is never evaluated  [unreachable]
python/lsst/daf/butler/core/timespan.py:239: error: Variable "lsst.daf.butler.core.timespan.TimespanBound" is not valid as a type  [valid-type]
python/lsst/daf/butler/core/timespan.py:239: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
python/lsst/daf/butler/core/timespan.py:255: error: Variable "lsst.daf.butler.core.timespan.TimespanBound" is not valid as a type  [valid-type]
python/lsst/daf/butler/core/timespan.py:255: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
python/lsst/daf/butler/core/timespan.py:285: error: Statement is unreachable  [unreachable]
python/lsst/daf/butler/core/timespan.py:290: error: Statement is unreachable  [unreachable]
python/lsst/daf/butler/registry/obscore/_records.py:177: error: Statement is unreachable  [unreachable]
python/lsst/daf/butler/registry/obscore/_records.py:180: error: Statement is unreachable  [unreachable]
Found 12 errors in 2 files (checked 223 source files)
```

python/lsst/daf/butler/_butler.py
timj: Yes, they could be and `RepoExportContext` already supports that. This export function does not though and would need to be modified a bit to make it work which is more than I'd want to do on this ticket which is not really meant to be dealing with those sorts of changes.

python/lsst/daf/butler/cli/butler.py
timj: If you feel strongly about it I can revert the commit. Each time we rerun `pyupgrade` it will try to "fix" it again...

TallJimbo: I don't care that much if it's not just some kind of "ignore this thing" configuration option.

PR #852, opened 2023-06-15
DM-39665: Remove faked DataCoordinate from all tests

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -2512,12 +2512,13 @@ class Butler(LimitedButler):
         # a DatasetRef for each defined instrument
         datasetRefs = []
 
+        dimensions = self.dimensions.extract(("instrument",))
         for datasetType in datasetTypes:
             if "instrument" in datasetType.dimensions:
                 for instrument in instruments:
                     datasetRef = DatasetRef(
                         datasetType,
-                        {"instrument": instrument},  # type: ignore
+                        DataCoordinate.standardize(instrument=instrument, graph=dimensions),
                         conform=False,
                         run="validate",
                     )
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -2505,24 +2505,26 @@ class Butler(LimitedButler):
         else:
             ignore = set()
 
-        # Find all the registered instruments
-        instruments = {record.name for record in self.registry.queryDimensionRecords("instrument")}
-
         # For each datasetType that has an instrument dimension, create
         # a DatasetRef for each defined instrument
         datasetRefs = []
 
-        dimensions = self.dimensions.extract(("instrument",))
-        for datasetType in datasetTypes:
-            if "instrument" in datasetType.dimensions:
-                for instrument in instruments:
-                    datasetRef = DatasetRef(
-                        datasetType,
-                        DataCoordinate.standardize(instrument=instrument, graph=dimensions),
-                        conform=False,
-                        run="validate",
-                    )
-                    datasetRefs.append(datasetRef)
+        # Find all the registered instruments (if "instrument" is in the
+        # universe).
+        if "instrument" in self.dimensions:
+            instruments = {record.name for record in self.registry.queryDimensionRecords("instrument")}
+
+            dimensions = self.dimensions.extract(("instrument",))
+            for datasetType in datasetTypes:
+                if "instrument" in datasetType.dimensions:
+                    for instrument in instruments:
+                        datasetRef = DatasetRef(
+                            datasetType,
+                            DataCoordinate.standardize(instrument=instrument, graph=dimensions),
+                            conform=False,
+                            run="validate",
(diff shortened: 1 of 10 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -2512,12 +2512,13 @@ def validateConfiguration(
         # a DatasetRef for each defined instrument
         datasetRefs = []
 
+        dimensions = self.dimensions.extract(("instrument",))
```
timj: It's possible this is a bad idea and I should move it into the loop. I don't want code to break that doesn't have instrument in the universe.

python/lsst/daf/butler/_butler.py
```diff
@@ -2512,12 +2512,13 @@ def validateConfiguration(
...
         for datasetType in datasetTypes:
             if "instrument" in datasetType.dimensions:
                 for instrument in instruments:
                     datasetRef = DatasetRef(
                         datasetType,
-                        {"instrument": instrument},  # type: ignore
+                        DataCoordinate.standardize(instrument=instrument, graph=dimensions),
                         conform=False,
```
timj: We still need the `conform=False` because by definition the dataId only has `instrument` in it and no values for the other dimensions. We can't change the DatasetType since that would mess with the matching system in the LookupKey.

python/lsst/daf/butler/tests/_datasetsHelper.py
```diff
@@ -91,26 +88,21 @@ def _makeDatasetRef(
 
         if run is None:
             run = "dummy"
+        dataId = DataCoordinate.standardize(dataId, graph=datasetType.dimensions)
```
timj: Now that everything conforms in the code calling this I think I can delete this line (I had to add it whilst I was testing things). We don't even annotate this method as taking Mapping for dataId.

timj: I had to leave it in but I only call it for a `dict` parameter since we do pass in dicts from the tests.

python/lsst/daf/butler/_butler.py
TallJimbo: I'm similarly ambivalent.  I suppose you could try/except and skip the whole block if instrument isn't in the dimensions.

python/lsst/daf/butler/_butler.py
TallJimbo: It might make more sense to invent values for other data ID keys here, as we do in some unit tests.  I'm a bit worried that by not conforming here we're putting constraints on the implementation (e.g. you can't compute a filename from a template as a part of whatever is invoked in the validation, since that will look for missing data ID keys).  I suppose by that argument we need to attach dimension records too.

Since inventing all of that is a bit of a pain, it might be reasonable to instead do a limit=1 query for them, and if you don't get anything, just warn that the validation was incomplete for certain dataset types.

TallJimbo (approved):

python/lsst/daf/butler/_butler.py
timj: Validation doesn't use dataId content other than instrument. Instrument is special because at the moment it's the only dataId value that can be used to select variants for things like templating and formatter selection. Validation does not need the dataId for anything else. In theory we could assign random content to the other dimensions since nothing is going to look at them.

TallJimbo: It doesn't use anything other than instrument now, but given the docs, I might have expected it to check that it can construct a valid template for each dataset type, too, since that's another part of the configuration that could go wrong.  I'm not saying we should change that now - I'm just saying that in addition to this being the last place (I think?) we need `conform=False`, it's also preventing us from making improvements like that in the future.

timj: Validation doesn't involve creating the final resolved template -- it involves looking at the template to make sure that there aren't any dimensions mentioned that are not present in the matching DatasetType. We can not select templates based on any DataId value other than the special "instrument". In the future you could conceive of extending that support to other dimensions but that's not done at the moment (hence this special block for "instrument").

TallJimbo: Ok, fair enough.  I suppose the other well-motivated way to get rid of `conform` would be inventing a `FakeDatasetRef` just for the validation script, but I'm ambivalent about whether that's actually better than just keeping it.

timj: I'm fine with trying to insert fake values for the missing dimensions since nothing cares about the value.

python/lsst/daf/butler/_butler.py
timj: I rearranged it a bit and fixed the related problem a couple of lines above where we were querying for the instrument dimension without checking the universe first.

python/lsst/daf/butler/_butler.py
timj: Creating fake missing dataId values was easy and I've now removed the `conform=False` flag.

PR #857, opened 2023-06-26
Fix typo in test docs

parejkoj:


timj (approved):
Ignore the mypy problem. We are discussing the right fix.

PR #858, opened 2023-06-27
DM-39582: Add caching for some butler primitives during deserialization

natelust:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-39582.api.md
+++ b/doc/changes/DM-39582.api.md
@@ -1,0 +1,1 @@
+Deprecate reconstituteDimensions argument from Quantum.from_simple
--- a/doc/changes/DM-39582.api.md
+++ b/doc/changes/DM-39582.api.md
@@ -1,1 +1,0 @@
-Deprecate reconstituteDimensions argument from Quantum.from_simple
--- a/python/lsst/daf/butler/core/datasets/ref.py
+++ b/python/lsst/daf/butler/core/datasets/ref.py
@@ -30,6 +30,7 @@ __all__ = [
 ]
 
 import enum
+import sys
 import uuid
 from collections.abc import Iterable
 from typing import TYPE_CHECKING, Any, ClassVar
@@ -41,6 +42,7 @@ from ..configSupport import LookupKey
 from ..dimensions import DataCoordinate, DimensionGraph, DimensionUniverse, SerializedDataCoordinate
 from ..json import from_json_pydantic, to_json_pydantic
 from ..named import NamedKeyDict
+from ..persistenceContext import PersistenceContextVars
 from .type import DatasetType, SerializedDatasetType
 
 if TYPE_CHECKING:
@@ -142,6 +144,10 @@ class DatasetIdFactory:
             return uuid.uuid5(self.NS_UUID, data)
 
 
+# This is constant, so don't recreate a set for each instance
+_serializedDatasetRefFieldsSet = {"id", "datasetType", "dataId", "run", "component"}
+
+
 class SerializedDatasetRef(BaseModel):
     """Simplified model of a `DatasetRef` suitable for serialization."""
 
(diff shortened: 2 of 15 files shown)


doc/changes/DM-39582.api.md
```diff
@@ -0,0 +1 @@
+Deprecate reconstituteDimensions argument from Quantum.from_simple
```
andy-slac: I think file name should be `removal`, not `api`, according to README.

Add backticks around `reconstituteDimensions` and `Quantum.from_simple` and period at the end.

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -396,9 +408,18 @@ def from_simple(
         ref : `DatasetRef`
             Newly-constructed object.
         """
+        cache = PersistenceContextVars.datasetRefs.get()
+        localName = sys.intern(
+            datasetType.name
+            if datasetType is not None
+            else (x.name if (x := simple.datasetType) is not None else "")
+        )
+        key = (simple.id.int, localName)
```
andy-slac, line 421: Should `simple.component` be a part of the key?

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -396,9 +408,18 @@ def from_simple(
...
+        if cache is not None and (cachedRef := cache.get(key, None)) is not None:
+            return cachedRef
         # Minimalist component will just specify component and id and
         # require registry to reconstruct
-        if set(simple.dict(exclude_unset=True, exclude_defaults=True)).issubset({"id", "component"}):
+        if not (simple.datasetType is not None or simple.dataId is not None or simple.run is not None):
```
andy-slac, line 426: Could you rewrite this as `simple.datasetType is None and simple.dataId is None and simple.run is None`, I think it makes it easier to read?

python/lsst/daf/butler/core/datastoreRecordData.py
```diff
@@ -70,6 +71,11 @@ def direct(
 
         This method should only be called when the inputs are trusted.
         """
+        key = frozenset(dataset_ids)
+        cache = PersistenceContextVars.serializedDatastoreRecordMapping.get()
+        if cache is not None and (value := cache.get(key)) is not None:
+            return value
```
andy-slac: I would not expect many (or maybe any) cache hits for this. `DatastoreRecordData` is per-quantum structure, I do not think any two quanta can have the same set of input datasets?

python/lsst/daf/butler/transfers/_yaml.py
```diff
@@ -64,6 +64,8 @@
 this version of the code.
 """
 
+_refIntId2UUID = defaultdict[int, uuid.UUID](uuid.uuid4)
```
andy-slac: Is this a workaround for input YAML files that still have integer dataset IDs in them? Should we instead fix those YAML files? It may also be better to generate reproducible UUIDs in that case if you want to keep this map.

andy-slac (approved):
Looks OK, one concern is about cache key for DatasetRef maybe needing `component`. I did not comment on docstring style, I think `ruff` should not take care of that.

python/lsst/daf/butler/core/datasets/ref.py
natelust: I ran across this too, by construction `simple.datasetType` is a string that will either just be the type itself, or if it is a component be a string of the form `type.component`

python/lsst/daf/butler/core/datasets/ref.py
natelust: That is not logically the same thing, We only want to run this when they are all `False`. But and is greedy, so False will always gobble up anything.

python/lsst/daf/butler/transfers/_yaml.py
natelust: I could not be sure all the places where this might be happening, so I opted to fix it here. Good point on the reproducibility I will go with that.

PR #859, opened 2023-07-03
DM-39876: Add bandaid for pydantic v2

timj:
Pydantic v2 breaks everything. Rather than pinning to <2 instead use the facility provided by v2 to import v1 APIs. This will give us some flexibility on timeline for fixing things to work only with v2.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/logging.py
+++ b/python/lsst/daf/butler/core/logging.py
@@ -31,7 +31,11 @@ from typing import IO, Any, ClassVar, Union, overload
 
 from lsst.utils.introspection import get_full_type_name
 from lsst.utils.iteration import isplit
-from pydantic import BaseModel, PrivateAttr
+
+try:
+    from pydantic.v1 import BaseModel, PrivateAttr
+except ModuleNotFoundError:
+    from pydantic import BaseModel, PrivateAttr
 
 _LONG_LOG_FORMAT = "{levelname} {asctime} {name} {filename}:{lineno} - {message}"
 """Default format for log records."""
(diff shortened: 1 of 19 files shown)


python/lsst/daf/butler/core/logging.py
```diff
@@ -175,10 +179,10 @@ class ButlerLogRecord(BaseModel):
     filename: str
     pathname: str
     lineno: int
-    funcName: str | None
+    funcName: str | None = None
     process: int
     processName: str
-    exc_info: str | None
+    exc_info: str | None = None
     MDC: dict[str, str]
```
andy-slac, line 186: Does this work with v1? I thought that pydantic (v1) does not allow non-default attributes after default ones (but I may be thinking of dataclasses)?

andy-slac (approved):
Looks good, one minor question.

python/lsst/daf/butler/core/logging.py
timj: I made that change first with the auto updater and it seemed to be okay. I am running Jenkins over night on everything so I'll see whether everything works in the morning.

timj:
I'm not sure what to do about the mypy failures. Mypy doesn't know about pydantic.v1 and with pydantic v1 it complains about thta type not existing because it never falls back to the except block. With pydantic v2 installed it gets even more confused.

timj:
In theory changing mypy.ini to use `pydantic.v1` with pydantic v2 and then adding `type: ignore` on the fallback imports does work. Unfortunately it doesn't work at the moment because fastapi pins pydantic to v1 so when we install that pydantic reverts to v1 and then mypy gets confused about the `pydantic.v1` import (which we have to do first), since mypy always picks up the first import it sees.

PR #862, opened 2023-07-06
DM-39915: Make butler.datastore an internal property

timj:
## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/script/queryDatasets.py
```diff
@@ -195,7 +195,7 @@ def getTables(self) -> list[AstropyTable]:
                 tables[dataset_ref.datasetType.name].add(dataset_ref)
         else:
             d = list(self.datasets)
-            ref_uris = self.butler.datastore.getManyURIs(d, predict=True)
+            ref_uris = self.butler._datastore.getManyURIs(d, predict=True)
```
timj: I imagine we need to create a `butler.get_many_uris()` method and switch to that here.

python/lsst/daf/butler/_butler.py
```diff
@@ -1265,8 +1265,8 @@ def getDirectDeferred(
         LookupError
             Raised if no matching dataset exists in the `Registry`.
         """
-        # Check thad dataset actuall exists.
-        if not self.datastore.exists(ref):
+        # Check thad dataset actually exists.
```
andy-slac:
```suggestion
        # Check that dataset actually exists.
```

python/lsst/daf/butler/_limited_butler.py
```diff
@@ -274,6 +275,137 @@ def getDeferred(
...
+            If `True`, allow URIs to be returned of datasets that have not
+            been written.
+
+        Returns
+        -------
+        uris : `DatasetRefURIs`
+            The URI to the primary artifact associated with this dataset (if
+            the dataset was disassembled within the datastore this may be
+            `None`), and the URIs to any components associated with the dataset
+            artifact. (can be empty if there are no components).
```
andy-slac:
```suggestion
            artifact (can be empty if there are no components).
```

python/lsst/daf/butler/_limited_butler.py
```diff
@@ -274,6 +275,137 @@ def getDeferred(
...
+
+        Parameters
+        ----------
+        refs : iterable of `DatasetIdRef`
+            References to the required datasets.
+        predict : `bool`, optional
+            If the datastore does not know about a dataset, should it
+            return a predicted URI or not?
+        allow_missing : `bool`
+            If `False`, and `predict` is `False`, will raise if a `DatasetRef`
```
andy-slac:
```suggestion
            If `False`, and ``predict`` is `False`, will raise if a `DatasetRef`
```

python/lsst/daf/butler/_limited_butler.py
```diff
@@ -274,6 +275,137 @@ def getDeferred(
...
+    ) -> dict[DatasetRef, DatasetRefURIs]:
+        """Return URIs associated with many datasets.
+
+        Parameters
+        ----------
+        refs : iterable of `DatasetIdRef`
+            References to the required datasets.
+        predict : `bool`, optional
+            If the datastore does not know about a dataset, should it
+            return a predicted URI or not?
```
andy-slac: Using a question for a description can raise some questions 🙂 Maybe reuse more assertive description as for `getURI`?

python/lsst/daf/butler/_limited_butler.py
```diff
@@ -274,6 +275,137 @@ def getDeferred(
...
+        allow_missing : `bool`
+            If `False`, and `predict` is `False`, will raise if a `DatasetRef`
+            does not exist.
+
+        Returns
+        -------
+        URIs : `dict` of [`DatasetRef`, `DatasetRefUris`]
```
andy-slac:
```suggestion
        URIs : `dict` of [`DatasetRef`, `DatasetRefURIs`]
```

python/lsst/daf/butler/_quantum_backed.py
```diff
@@ -227,7 +227,7 @@ def initialize(
             predicted_outputs=predicted_outputs,
             dimensions=dimensions,
             filename=filename,
-            datastore_records=quantum.datastore_records,
+            datastore_records=quantum._datastore_records,
```
andy-slac: I think we should keep `datastore_records` here.

python/lsst/daf/butler/core/datastore.py
```diff
@@ -372,6 +372,18 @@ def names(self) -> tuple[str, ...]:
         # Default implementation returns solely the name itself
         return (self.name,)
 
+    @property
+    def roots(self) -> dict[str, ResourcePath | None]:
+        """Return the root URIs for each named datastore.
+
+        Returns
```
andy-slac: Do we use `Returns` for `@property`? I thought we just write property type in parens.

python/lsst/daf/butler/script/queryDatasets.py
andy-slac: You have added that method, may use it here as well?

python/lsst/daf/butler/_limited_butler.py
```diff
@@ -274,6 +275,137 @@ def getDeferred(
...
+
+    def get_datastore_roots(self) -> dict[str, ResourcePath | None]:
+        """Return the defined root URIs for all registered datastores.
+
+        Returns
+        -------
+        roots : `dict` [`str`, `~lsst.resources.ResourcePath` | `None`]
+            A mapping from datastore name to datastore root URI. The root
+            can be `None` if the datastore does not have any concept of a root
+            URI.
```
andy-slac, line 296: Abstract question - I suppose it does not make sense to have a single datastore with multiple roots?

andy-slac (approved):
Looks good, few minor comments. I guess you need to figure out why build fails and mypy shows some new errors.

python/lsst/daf/butler/_quantum_backed.py
timj: Hmm. It seems I fluked the internal name. That explains why the tests worked...

python/lsst/daf/butler/core/datastore.py
timj: I wrote it as a getter first and then changed my mind.

python/lsst/daf/butler/_limited_butler.py
timj: Not the way we've currently set things up. Multiple roots means chained datastore at the moment.

timj:
The build failing is from fastapi supporting pydantic v2. The mypy failure is all pydantic and we have started a discussion on Slack about it -- it won't block the merge.

andy-slac:
> The mypy failure is all pydantic and we have started a discussion on Slack about it -- it won't block the merge.

I see GA mypy failures in `cli/utils.py` and `cli/cmd/commands.py` which I did not see before, maybe new `click` version?

andy-slac:
> I see GA mypy failures in `cli/utils.py` and `cli/cmd/commands.py` which I did not see before, maybe new `click` version?

Indeed, I see the same errors locally if I upgrade click from 8.1.3 to 8.1.4.

timj:
> I see the same errors locally if I upgrade click from 8.1.3 to 8.1.4.

which shows the dangers of having a failing action that we then ignore. I hadn't noticed the click issue.

PR #863, opened 2023-07-08
DM-39939: use query-results' grouping when processing iterables of DatasetRefs

TallJimbo:
## Checklist

- [x] ran [Jenkins](https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/39151/pipeline)

### DIFF
--- a/python/lsst/daf/butler/core/datasets/ref.py
+++ b/python/lsst/daf/butler/core/datasets/ref.py
@@ -33,7 +33,7 @@ import enum
 import sys
 import uuid
 from collections.abc import Iterable
-from typing import TYPE_CHECKING, Any, ClassVar
+from typing import TYPE_CHECKING, Any, ClassVar, Protocol, runtime_checkable
 
 from lsst.utils.classes import immutable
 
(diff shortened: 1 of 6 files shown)


python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -580,12 +580,41 @@ def groupByType(refs: Iterable[DatasetRef]) -> NamedKeyDict[DatasetType, list[Da
...
+            arguments and the result reutrnd.
+
+        Returns
+        -------
+        grouped : `~collections.abc.Iterable` [ `tuple` [ `DatasetType`, \
+                `Iterable` [ `DatasetRef` ] ]]
+            Grouped `DatasetRef` instances.
+        """
+        if hasattr(refs, "_iter_by_dataset_type"):
+            return refs._iter_by_dataset_type()
```
andy-slac: I'm not super-happy with this dynamic approach, particularly because our favorite mypy cannot type-check this. Would it be possible to add an abstraction (or maybe `Protocol` with runtime check) so that `isinstance` can be used instead?

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -580,12 +580,41 @@ def groupByType(refs: Iterable[DatasetRef]) -> NamedKeyDict[DatasetType, list[Da
...
+    ) -> Iterable[tuple[DatasetType, Iterable[DatasetRef]]]:
+        """Group an iterable of `DatasetRef` by `DatasetType` with special
+        hooks for custom iterables that can do this efficiently.
+
+        Parameters
+        ----------
+        refs : `~collections.abc.Iterable` [ `DatasetRef` ]
+            `DatasetRef` instances to group.  If this has a
+            ``_iter_by_dataset_type`` method, it will be called with no
+            arguments and the result reutrnd.
```
andy-slac:
```suggestion
            arguments and the result returned.
```

tests/test_progress.py
```diff
@@ -180,23 +183,101 @@ def test_bar_update(self):
...
+        mapping = {"x": iter(range(2)), "y": iter(range(3))}
+        seen = {}
+        for key, chunk in self.progress.iter_item_chunks(mapping.items(), total=False):
+            seen[key] = list(chunk)
+        self.assertEqual(seen, {"x": list(range(2)), "y": list(range(3))})
+        self.assertEqual(MockProgressBar.last.reported, [0, 1, 2])
+        self.assertEqual(MockProgressBar.last.total, 2)
+
+    def test_iter_item_chunks_not_sized(self):
+        """Test using `Progress.iter_item_chunks` with an unsized iterable
```
andy-slac:
```suggestion
        """Test using `Progress.iter_item_chunks` with an unsized iterable of
```

andy-slac (approved):
Looks good, one suggestion to make it a bit more type-checkable.

python/lsst/daf/butler/core/datasets/ref.py
TallJimbo: I've switched to a runtime-checkable protocol, but I've given it a leading underscore (and documented it as "package private") since I don't want external code to start using this.

PR #864, opened 2023-07-12
DM-39944: Replace Butler.registry with registry shim

andy-slac:
`Registry` ABC is split into three classes:
- `Registry` is now only contains the methods that are used by regular clients via `Butler.registry` attribute
- `ButlerRegistry` ABC extends `Registry` with few methods used by Butler internally, SQL and remote implementations now inherit `ButlerRegistry`
- `ButlerFactory` class has methods to create/instantiate registry from configuration. There is still `Registry.createFromConfig`, this is only for the tests that currently use this method, I do not want to update all those as I'm not sure if `ButlerFactory` survives all further updates (or this review).

`Butler.registry` now returns `ButlerShim` instance which is a subclass of Registry. Shim for now just forwards all calls to `Butler._registry`, as we do updates to Butler/Registry implementations the shim will probably do more complicated things, and eventually it should disappear after all clients migrate to new Butler API.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_registry_shim.py
+++ b/python/lsst/daf/butler/_registry_shim.py
@@ -1,0 +1,392 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import annotations
+
+__all__ = ("Registry",)
+
+import contextlib
+from collections.abc import Iterable, Iterator, Mapping, Sequence
+from typing import TYPE_CHECKING, Any
+
+from .core import (
+    DataCoordinate,
+    DataId,
+    DatasetAssociation,
+    DatasetId,
+    DatasetIdGenEnum,
+    DatasetRef,
+    DatasetType,
+    Dimension,
+    DimensionElement,
+    DimensionGraph,
+    DimensionRecord,
+    DimensionUniverse,
+    NameLookupMapping,
+    Timespan,
+)
+from .registry import Registry
+from .registry._collection_summary import CollectionSummary
+from .registry._collectionType import CollectionType
+from .registry._defaults import RegistryDefaults
+from .registry.queries import DataCoordinateQueryResults, DatasetQueryResults, DimensionRecordQueryResults
+
+if TYPE_CHECKING:
+    from ._butler import Butler
+    from .registry._registry import CollectionArgType
+    from .registry.interfaces import ObsCoreTableManager
+
+
+class RegistryShim(Registry):
+    """Implementation of `Registry` interface exposed to clients by `Butler`.
+
+    Parameters
+    ----------
+    butler : `Butler`
+        Data butler instance.
+
+    Notes
+    -----
+    This shim implementation of `Registry` forwards all methods to an actual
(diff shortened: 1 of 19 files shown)


timj:
It looks like I missed a butler.datastore fix in the tests since we get 22 deprecation warnings. It's interesting that python 3.11 is now significantly faster than 3.10.

timj:
😄 the warnings about `Butler.datastore` are coming specifically from this branch. Nothing to do with `main` at all.

andy-slac:
> 😄 the warnings about `Butler.datastore` are coming specifically from this branch. Nothing to do with `main` at all.

I likely messed it up when rebased after your changes, will check everything again.

python/lsst/daf/butler/_registry_shim.py
```diff
@@ -0,0 +1,394 @@
...
+    @property
+    def dimensions(self) -> DimensionUniverse:
+        # Docstring inherited from a base class.
+        return self._registry.dimensions
+
+    @property
+    def defaults(self) -> RegistryDefaults:
+        # Docstring inherited from a base class.
+
+        # Note that we do not allow changing defaults via this class
```
TallJimbo: Looks like we do, given the setter below.

python/lsst/daf/butler/registry/_registry_factory.py
```diff
@@ -0,0 +1,185 @@
...
+    Each registry implementation can have its own constructor parameters.
+    The assumption is that an instance of a specific subclass will be
+    constructed from configuration using ``RegistryClass.fromConfig()`` or
+    ``RegistryClass.createFromConfig()``.
+
+    This class will look for a ``cls`` entry in registry configuration object
+    (defaulting to ``SqlRegistry``), import that class, and call one of the
+    above methods on the imported class.
+    """
+
```
TallJimbo, line 60: Would it make sense to give this class a constructor that does the work of `force_registry_config` and holds the config as an attribute, and then make the other methods regular instance methods (probably abstract ones)?

TallJimbo (approved):
Looks good!  This is all consistent with how I see things going, at least at first.

I think we're in agreement that the roles of some of the new classes will change significantly before this is all over, and they may well disappear.  So how about we add a leading underscore to `ButlerRegistry` and `RegistryFactory` to make it extra clear that these are not symbols anyone should be depending on.

Did you see any usage of `createFromConfig` in tests downstream of daf_butler, or was it just a lot of tests in daf_butler itself?  That's currently at the top of my list of `Registry` methods we should actually deprecate (in favor of `ButlerRegistry`, for now), with `transaction` and `resetConnectionPool` close behind.  There are of course a ton of other methods we want to move to `Butler`, maybe adjust, and ultimately remove from `Registry` in the distant future, but we should probably be more aggressive about removing those three and maybe a few others ASAP since we don't expect to even give them `Butler`-method replacements.

python/lsst/daf/butler/_registry_shim.py
andy-slac: Indeed, I wanted to disable setter but then had to re-enable it because there are some legit clients, but forgot to remove the comment, will drop it.

python/lsst/daf/butler/registry/_registry_factory.py
andy-slac: Adding a constructor looks reasonable. If factory methods are abstract we'd need them implemented somewhere? Do you think that this should be another base class for ButlerRegistry subclasses?

andy-slac:
> So how about we add a leading underscore to `ButlerRegistry` and `RegistryFactory` to make it extra clear that these are not symbols anyone should be depending on.

I renamed those two classes.

> Did you see any usage of `createFromConfig` in tests downstream of daf_butler, or was it just a lot of tests in daf_butler itself? 

There are bout five other packages that use `createFromConfig` in their tests. I actually started replacing that with `RegistryFactory` but then decided that I want to wait until things become more stable. 

> That's currently at the top of my list of `Registry` methods we should actually deprecate (in favor of `ButlerRegistry`, for now), with `transaction` and `resetConnectionPool` close behind. 

I looked at `resetConnectionPool` and it's of course only needed by `ctrl_mpexec`. I'd like to remove it from registry interface too, but having it in `Butller` is not much better in my opinion. Need to think again how to handle it better.

python/lsst/daf/butler/registry/_registry_factory.py
TallJimbo: Ah, I think I got confused about the current state of things while reviewing commit-by-commit - I thought this still had some methods with docs that said they needed to be reimplemented by derived classes, and that made me think there were _already_ subclass implementations and I didn't step back and notice that there weren't any.

So just ignore the bit about abstract methods; this is just a regular concrete (probably final) class, and that's fine.

PR #866, opened 2023-07-14
DM-40002: Try to support pydantic v1 and v2

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_compat.py
+++ b/python/lsst/daf/butler/_compat.py
@@ -1,0 +1,112 @@
+# This file is part of pipe_base.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Code to support backwards compatibility."""
+
+__all__ = ["PYDANTIC_V2", "_BaseModelCompat"]
+
+from collections.abc import Callable, Mapping
+from typing import Any, Self
+
+from pydantic import BaseModel
+from pydantic.version import VERSION as PYDANTIC_VERSION
+
+PYDANTIC_V2 = PYDANTIC_VERSION.startswith("2.")
+
+
+if PYDANTIC_V2:
+
+    class _BaseModelCompat(BaseModel):
+        """Methods from pydantic v1 that we want to emulate in v2.
+
+        Some of these methods are provided by v2 but issue deprecation
+        warnings.  We need to decide whether we are also okay with deprecating
+        them or want to support them without the deprecation message.
+        """
+
+        def json(
+            self,
+            *,
+            include: set[int | str] | Mapping[int | str, Any] | None = None,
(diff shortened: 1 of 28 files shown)


tests/test_butler.py
```diff
@@ -42,13 +42,14 @@
 try:
     import boto3
     import botocore
+    from lsst.resources.s3utils import setAwsEnvCredentials, unsetAwsEnvCredentials
     from moto import mock_s3  # type: ignore[import]
 except ImportError:
     boto3 = None
 
-    def mock_s3(cls):  # type: ignore[no-untyped-def]
+    def mock_s3(*args: Any, **kwargs: Any) -> Any:  # type: ignore[no-untyped-def]
```
andy-slac, line 50: I'd expect you can drop `type: ignore` here?

python/lsst/daf/butler/_compat.py
```diff
@@ -0,0 +1,162 @@
...
+
+if sys.version_info >= (3, 11, 0):
+    from typing import Self
+else:
+    from typing import TypeVar
+
+    Self = TypeVar("Self", bound="_BaseModelCompat")  # type: ignore
+
+
+PYDANTIC_V2 = PYDANTIC_VERSION.startswith("2.")
```
andy-slac, line 42: Pydantic 3.0 will be the same as 1.x? 🙂

python/lsst/daf/butler/_compat.py
```diff
@@ -0,0 +1,162 @@
...
+        them or want to support them without the deprecation message.
+        """
+
+        def json(
+            self,
+            *,
+            include: set[int] | set[str] | dict[int, Any] | dict[str, Any] | None = None,  # type: ignore
+            exclude: set[int] | set[str] | dict[int, Any] | dict[str, Any] | None = None,  # type: ignore
+            by_alias: bool = False,
+            skip_defaults: bool | None = None,
```
andy-slac, line 61: `skip_defaults` is completely unused?

python/lsst/daf/butler/_quantum_backed.py
```diff
@@ -749,19 +745,40 @@ def _to_uuid_set(uuids: Iterable[str | uuid.UUID]) -> set[uuid.UUID]:
...
+            setter(data, "predicted_outputs", _to_uuid_set(predicted_outputs))
+            setter(data, "actual_outputs", _to_uuid_set(actual_outputs))
+            setter(
+                data,
+                "datastore_records",
+                {
+                    key: SerializedDatastoreRecordData.direct(**records)
+                    for key, records in datastore_records.items()
+                },
+            )
```
andy-slac: Would it be possible to implement V1 `model_construct` to do the same thing as we do here to avoid if/else in every model class?

andy-slac (approved):
Looks OK after a quick look, one question is whether `_BaseModelCompat.model_construct` could be changed to avoid if/else in many model classes.

python/lsst/daf/butler/_compat.py
timj: I'm hoping that we've removed the v1 code by then...

python/lsst/daf/butler/_quantum_backed.py
timj: There is v1 `BaseModel.construct()` which does the same thing as `model_construct`. We need to understand why @natelust decided not to use it. If we know why and need to change we would need to go back to the explicit setters above and change to using the `__pydantic_fields_set__` property instead.

python/lsst/daf/butler/_compat.py
timj: Yes. `model_dump_json` does not have it. I can't really say what the difference is between `exclude_defaults` and `skip_defaults` and in fact its use has been deprecated even in pydantic v1:

```python
        if skip_defaults is not None:
            warnings.warn(
                f'{self.__class__.__name__}.json(): "skip_defaults" is deprecated and replaced by "exclude_unset"',
                DeprecationWarning,
            )
            exclude_unset = skip_defaults
```

timj:
@andy-slac you are right that I could implement v1 model_construct in _BaseModelCompat and just use it in the same way for both pydantic. Then I could even make v2 do the explicit setter approach if we decided we didn't want to use the base v2 base class model_construct. I will change.

PR #867, opened 2023-07-20
DM-40120: Add without_datastore flag to Butler

timj:
This allows `butler = Butler(uri, skip_datastore=True)` to return a Butler instance that has no connection to a real datastore. I've also modified script implementations to add the flag where appropriate.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -222,7 +222,7 @@ class Butler(LimitedButler):
             self.storageClasses = butler.storageClasses
             self._config: ButlerConfig = butler._config
         else:
-            self._config = ButlerConfig(config, searchPaths=searchPaths)
+            self._config = ButlerConfig(config, searchPaths=searchPaths, skip_datastore=skip_datastore)
             try:
                 if "root" in self._config:
                     butlerRoot = self._config["root"]
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -69,6 +69,7 @@ from .core import (
     DimensionRecord,
     DimensionUniverse,
     FileDataset,
+    NullDatastore,
     Progress,
     StorageClass,
     StorageClassFactory,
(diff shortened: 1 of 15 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -149,6 +150,9 @@ class Butler(LimitedButler):
         the default for that dimension.  Nonexistent collections are ignored.
         If a default value is provided explicitly for a governor dimension via
         ``**kwargs``, no default will be inferred for that dimension.
+    skip_datastore : `bool`, optional
+        If `True` do not attach a datastore to this butler. Any attempts
+        to use a datastore will fail.
```
TallJimbo, line 155: How about "without" or "no" instead of "skip" in the name for this argument?

python/lsst/daf/butler/core/datastore.py
```diff
@@ -1205,3 +1208,122 @@ def set_retrieve_dataset_type_method(self, method: Callable[[str], DatasetType |
...
+
+    def __init__(
+        self,
+        config: Config | ResourcePathExpression | None,
+        bridgeManager: DatastoreRegistryBridgeManager | None,
+        butlerRoot: ResourcePathExpression | None = None,
+    ):
+        # Name ourselves with the timestamp the datastore
+        # was created.
+        self.name = f"{type(self).__name__}@{time.time()}"
```
TallJimbo, line 1231: Why do we need the timestamp?  Two `NullDatastore` instances should be identical (it could even be a singleton, though I don't think it needs to be), and I'd _hope_ that means we don't need to care about conflicts.

python/lsst/daf/butler/core/datastore.py
```diff
@@ -1205,3 +1208,122 @@ def set_retrieve_dataset_type_method(self, method: Callable[[str], DatasetType |
...
+    def exists(self, datasetRef: DatasetRef) -> bool:
+        return False
+
+    def get(
+        self,
+        datasetRef: DatasetRef,
+        parameters: Mapping[str, Any] | None = None,
+        storageClass: StorageClass | str | None = None,
+    ) -> Any:
+        raise NotImplementedError("This is a no-op datastore that can not access a real datastore")
```
TallJimbo: I'm a little torn on whether this exception type should instead match what some other datastore would raise when the dataset does not exist.  Does the Datastore ABC specify what the exception should be?

TallJimbo (approved):

python/lsst/daf/butler/core/datastore.py
timj: Good point, although since this datastore should never result in a dataset being stored anywhere there's no requirement for it to be a fixed name. The time is really there to let you know when you created that butler and it will appear in stringified output of `Butler`. Whether anyone cares is a question.

python/lsst/daf/butler/core/datastore.py
timj: Yes. It's a tricky one as to how much this datastore should act like a datastore that doesn't have anything in it versus a datastore that reports that it can't do anything. It looks like I should raise `FileNotFoundError`.

PR #868, opened 2023-07-21
DM-40057: shorten deprecation timeline and add ticket TODOs

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1109,6 +1109,7 @@ class Butler(LimitedButler):
 
         return ref
 
+    # TODO: remove on DM-40067.
     @transactional
     @deprecated(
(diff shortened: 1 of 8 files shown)


timj (approved):
Looks good. I think you also want `quantum.py` which has a "v 27" in it.

TallJimbo:
Thanks for catching `quantum.py`.  Do you know if we have a removal ticket for that deprecation?

timj:
If you use "fixup" in your commit messages to indicate they will be squashed later you can add this string to thr do-not-merge action.

TallJimbo:
> If you use "fixup" in your commit messages to indicate they will be squashed later you can add this string to the do-not-merge action.

That's a good idea, because that's actually coming from `git commit --fixup`, so it is a bit of convention.

PR #869, opened 2023-07-21
DM-40156: Some code cleanups

timj:
## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/core/dimensions/_coordinate.py
```diff
@@ -957,7 +956,7 @@ def hasRecords(self) -> bool:
 
     def _record(self, name: str) -> DimensionRecord | None:
         # Docstring inherited from DataCoordinate.
-        assert False
```
timj, line 960: This looks wrong to me and `AssertionError` doesn't seem any better. Is it meant to be `NotImplementedError`?

python/lsst/daf/butler/core/ddl.py
```diff
@@ -401,10 +399,9 @@ def isStringType(self) -> bool:
             string type if it has been decided that it should be implemented
             as a `sqlalchemy.Text` type.
         """
-        if self.dtype == sqlalchemy.String:
-            # For short strings retain them as strings
-            if self.dtype == sqlalchemy.String and self.length and self.length <= 32:
-                return True
+        # For short strings retain them as strings
+        if self.dtype == sqlalchemy.String and self.length and self.length <= 32:
```
andy-slac: Should it say `self.dtype is sqlalchemy.String`? I think `is` is preferred for checking type equality.

python/lsst/daf/butler/core/dimensions/_coordinate.py
andy-slac: Hmm, I guess motivation here is that this should never happen, so `AssertionError ` is probably fine.

python/lsst/daf/butler/registry/databases/sqlite.py
```diff
@@ -87,14 +87,14 @@ def __init__(
...
-                dbList = list(cursor.execute("PRAGMA database_list").fetchall())
+        with engine.connect() as connection, closing(connection.connection.cursor()) as cursor:
+            dbList = list(cursor.execute("PRAGMA database_list").fetchall())
         if len(dbList) == 0:
             raise RuntimeError("No database in connection.")
         if namespace is None:
             namespace = "main"
-        for _, dbname, filename in dbList:
+        # Look for the filename associated with this namespace.
+        for _, dbname, filename in dbList:  # B007
```
andy-slac, line 97: Does not `B007` need some tag or something (e.g. `noqa:`)?

python/lsst/daf/butler/script/transferDatasets.py
```diff
@@ -71,8 +71,8 @@ def transferDatasets(
     source_butler = Butler(source, writeable=False)
     dest_butler = Butler(dest, writeable=True)
 
-    dataset_type_expr = ... if not dataset_type else dataset_type
-    collections_expr: tuple[str, ...] | EllipsisType = ... if not collections else collections
+    dataset_type_expr = dataset_type if dataset_type else ...
+    collections_expr: tuple[str, ...] | EllipsisType = collections if collections else ...
```
andy-slac: When ternary operator was not a thing, everyone was happy with shorter `collections or ...` 🙂

andy-slac (approved):

python/lsst/daf/butler/core/ddl.py
timj: That's is probably better. ruff couldn't know that.

python/lsst/daf/butler/registry/databases/sqlite.py
timj: Yes, but I haven't turned "B" on in the ruff config so ruff complains about unneeded noqa. I could add "B" and then add ignore for any that we don't intend to fix. Same for SIM, I haven't enabled SIM by default.

python/lsst/daf/butler/script/transferDatasets.py
timj: Hmm. That is shorter and seems like it might be preferred.

PR #870, opened 2023-07-25
DM-40121 Optimizations in QuantumGraph loading

natelust:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/dimensions/_records.py
+++ b/python/lsst/daf/butler/core/dimensions/_records.py
@@ -23,6 +23,7 @@ from __future__ import annotations
 
 __all__ = ("DimensionRecord", "SerializedDimensionRecord")
 
+from collections.abc import Hashable
 from typing import TYPE_CHECKING, Any, ClassVar, Optional, Tuple
 
 import lsst.sphgeom
@@ -170,23 +171,22 @@ class SerializedDimensionRecord(_BaseModelCompat):
 
         This method should only be called when the inputs are trusted.
         """
-        _recItems = record.items()
+        # This method requires tuples as values of the mapping, but JSON
+        # readers will read things in as lists. Be kind and transparently
+        # transform to tuples
+        _recItems = {k: v if type(v) != list else tuple(v) for k, v in record.items()}  # type: ignore
+
         # Type ignore because the ternary statement seems to confuse mypy
         # based on conflicting inferred types of v.
         key = (
             definition,
-            frozenset((k, v if not isinstance(v, list) else tuple(v)) for k, v in _recItems),  # type: ignore
+            frozenset(_recItems.items()),
         )
         cache = PersistenceContextVars.serializedDimensionRecordMapping.get()
         if cache is not None and (result := cache.get(key)) is not None:
             return result
 
-        # This method requires tuples as values of the mapping, but JSON
-        # readers will read things in as lists. Be kind and transparently
-        # transform to tuples
-        serialized_record = {k: v if type(v) != list else tuple(v) for k, v in record.items()}  # type: ignore
-
-        node = cls.model_construct(definition=definition, record=serialized_record)  # type: ignore
+        node = cls.model_construct(definition=definition, record=_recItems)  # type: ignore
 
         if cache is not None:
             cache[key] = node
@@ -350,6 +350,7 @@ class DimensionRecord:
         simple: SerializedDimensionRecord,
         universe: DimensionUniverse | None = None,
         registry: Registry | None = None,
+        cacheKey: Hashable | None = None,
     ) -> DimensionRecord:
         """Construct a new object from the simplified form.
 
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/core/dimensions/_records.py
```diff
@@ -170,23 +171,22 @@ def direct(
 
         This method should only be called when the inputs are trusted.
         """
-        _recItems = record.items()
+        # This method requires tuples as values of the mapping, but JSON
+        # readers will read things in as lists. Be kind and transparently
+        # transform to tuples
+        _recItems = {k: v if type(v) != list else tuple(v) for k, v in record.items()}  # type: ignore
```
TallJimbo, line 177:
```suggestion
        _recItems = {k: tuple(v) for k, v in record.items()}  # type: ignore
```
?

I checked that
```
a = (1, 2)
assert tuple(a) is a
```
passes, so any conditional we write in Python is unlikely to beat the compiled `tuple` constructor itself on speed.

python/lsst/daf/butler/core/dimensions/_records.py
```diff
@@ -170,23 +171,22 @@ def direct(
...
+
         # Type ignore because the ternary statement seems to confuse mypy
         # based on conflicting inferred types of v.
         key = (
             definition,
```
TallJimbo, line 182: `definition.name` would probably make for a slightly more performant key.

python/lsst/daf/butler/core/dimensions/_records.py
```diff
@@ -170,23 +171,22 @@ def direct(
...
-            frozenset((k, v if not isinstance(v, list) else tuple(v)) for k, v in _recItems),  # type: ignore
+            frozenset(_recItems.items()),
```
TallJimbo, line 183: This is could be a pretty inefficient key for spatial dimension records in particular, in which there's a spatial region involved in any comparison.

You really just need to compare the required dimension values, not any of the metadata fields, and since those dimensions are fully specified `definition`,
```
key = (
    definition.name,
    tuple(_recItems[k] for k in definition.required.names),
)
```
is probably pretty close to the best key you can get for dimension records, _if_ the extra effort to extract it doesn't outweigh the advantage of faster comparisons.

TallJimbo (approved):

python/lsst/daf/butler/core/dimensions/_records.py
natelust: but not all of the things in record are lists, and we cant convert strings, numbers, et al into tuples

python/lsst/daf/butler/core/dimensions/_records.py
natelust: At this level definition is still a str from the serialization

python/lsst/daf/butler/core/dimensions/_records.py
TallJimbo: Ah, I misunderstood what these elements were; definitely ignore my previous comment.  I didn't actually think _any_ of the fields of a `DimensionRecord` could be `list` _or_ `tuple`, though; just `int`, `str`, `float`, `Timespan`, or `ConvexPolygon`.

natelust: Yeah https://github.com/lsst/daf_butler/blob/146dbb4537c4e3d171fc5a1c4bdb629136a93735/python/lsst/daf/butler/core/dimensions/_records.py#L131, tuple comes back as int, this is right from the raw python parsing

TallJimbo: Ah, ok, that must be the `Timespan`, then.  Thanks for explaining.

python/lsst/daf/butler/core/dimensions/_records.py
natelust: Same thing here, dimension is not yet anything, it is still an attribute on the serialized record. I could pull de-serializing the definition to before making the key, but it would mean deserializing every definition to create a key to see if there was already a cached record. Couple that with the iteration and I am not sure we would get a speed up, but I can test.

natelust: oh no I though this is still in the serialized object, I would prefer to not de-serialize anything if it could be avoided. If you thought it safe, I could always drop region

TallJimbo: FWIW, "deserializing the definition" should just be `universe[definition]`.  But I agree it's not clear whether it'd be an overall win.

natelust: but in this case it is in creating the serialized object itself. We are not sure yet what universe to use. There is a universe object that is passed to the `from_simple` method, but that comes later. I could see if this optimization helps there

natelust: ok, so I looked into it, and records are not trivial things that are results of dynamically created classes. For instance, the serialization of band is `(definition="band", records={"name":"i"})`. I don't want to get in the business of deconstructing all that again here, so I am going to leave it as is.

TallJimbo: Fair enough.  Sounds like it'd require a fair bit of restructuring if the universe isn't available here.

PR #874, opened 2023-08-04
DM-40303: Fix pydantic v2 warnings

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/datasets/ref.py
+++ b/python/lsst/daf/butler/core/datasets/ref.py
@@ -35,9 +35,10 @@ import uuid
 from collections.abc import Iterable
 from typing import TYPE_CHECKING, Any, ClassVar, Protocol, TypeAlias, runtime_checkable
 
-from lsst.daf.butler._compat import _BaseModelCompat
+import pydantic
+from lsst.daf.butler._compat import PYDANTIC_V2, _BaseModelCompat
 from lsst.utils.classes import immutable
-from pydantic import StrictStr, validator
+from pydantic import StrictStr
 
 from ..configSupport import LookupKey
 from ..dimensions import DataCoordinate, DimensionGraph, DimensionUniverse, SerializedDataCoordinate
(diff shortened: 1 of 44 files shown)


python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -179,24 +180,35 @@ class SerializedDatasetRef(_BaseModelCompat):
...
-        # Component should not be given if datasetType is given
-        if v and (d := "datasetType") in values and values[d] is not None:
-            raise ValueError(f"datasetType ({values[d]}) can not be set if component is given ({v}).")
-        return v
+    if PYDANTIC_V2:
+        # Can not use "after" validator since in some cases the validator
+        # seems to trigger with the datasetType field not yet set.
+        @pydantic.model_validator(mode="before")  # type: ignore[attr-defined]
+        @classmethod
+        def check_consistent_parameters(cls, data: Any) -> Any:
```
andy-slac: Should it take `data: dict[str, Any]` and return the same type?

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -1064,7 +1064,7 @@ def transfer_from(
             # ask the receiving datastore to copy it when it doesn't exist
             # so we have to filter again based on what the source datastore
             # understands.
-            known_to_source = source_child.knows_these([ref for ref in refs])
+            known_to_source = source_child.knows_these(list(refs))
```
andy-slac, line 1067: Not sure you need `list` here, `knows_these` says it takes `Iterable`, and `refs` is Iterable here too. I think this code already assumes that you can iterate many times over `refs`, which is probably not true for `Iterable` in general.

python/lsst/daf/butler/formatters/file.py
```diff
@@ -175,9 +175,9 @@ def _coerceBuiltinType(self, inMemoryDataset: Any, writeStorageClass: StorageCla
             and type(inMemoryDataset).__module__ == "builtins"
         ):
             # Try different ways of converting to the required type.
-            if hasattr(writeStorageClass.pytype, "parse_obj"):
+            if hasattr(writeStorageClass.pytype, "model_validate"):
```
andy-slac: Hmm, would it be safer to check `isinstance` instead with pydantic base class? `model_validate` is probably not super-popular method name but even that could potentially clash in some random class?

andy-slac (approved):

python/lsst/daf/butler/core/datasets/ref.py
timj: That is probably okay. I was following the pydantic porting guide.

python/lsst/daf/butler/datastores/chainedDatastore.py
timj: I think I was forcing it to a list originally so that we could go through it multiple times, which probably means the annotation is wrong. Should it be Sequence?

python/lsst/daf/butler/formatters/file.py
timj: I see I'm already using isinstance later so it probably does no harm to say this is explicitly the pydantic path. The best bit about this comment here is that it explains to me why ci_hsc_gen3 failed. I was trying to understand why ScarletModelData was not working properly and it turns out it's because that class tried to emulate pydantic by having a `parse_obj` and `json()` method and so `isinstance` here would break it. It does though tell me how to fix the scarlet problem, or at least I have to decide whether to keep parse_obj here or change `ScarletModelData` to use `model_validate` and `model_dump_json`...

python/lsst/daf/butler/datastores/chainedDatastore.py
andy-slac: We can keep `Iterable` but then make sure that code can handle single-pass iterables. This code in chainedDatastore cannot, it calls `set(refs)`  few lines above. fileDatastore also doing similar thing in `knows_these`. Or alternatively we should use Sequence everywhere.

PR #875, opened 2023-08-04
DM-40053: Add Datastore records to DatasetRef and use them in butler.get

andy-slac:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1182,62 +1182,42 @@ class Butler(LimitedButler):
         TypeError
             Raised if the butler is read-only or if no run has been provided.
         """
-        if not self.isWriteable():
-            raise TypeError("Butler is read-only.")
-
         if isinstance(datasetRefOrType, DatasetRef):
             # This is a direct put of predefined DatasetRef.
             log.debug("Butler put direct: %s", datasetRefOrType)
             if run is not None:
                 warnings.warn("Run collection is not used for DatasetRef", stacklevel=3)
-            ref = datasetRefOrType
-
             # If registry already has a dataset with the same dataset ID,
             # dataset type and DataId, then _importDatasets will do nothing and
             # just return an original ref. We have to raise in this case, there
             # is a datastore check below for that.
-            self._registry._importDatasets([ref], expand=True)
+            self._registry._importDatasets([datasetRefOrType], expand=True)
             # Before trying to write to the datastore check that it does not
             # know this dataset. This is prone to races, of course.
-            if self._datastore.knows(ref):
-                raise ConflictingDefinitionError(f"Datastore already contains dataset: {ref}")
+            if self._datastore.knows(datasetRefOrType):
+                raise ConflictingDefinitionError(f"Datastore already contains dataset: {datasetRefOrType}")
             # Try to write dataset to the datastore, if it fails due to a race
             # with another write, the content of stored data may be
             # unpredictable.
             try:
-                stored_refs = self._datastore.put_new(obj, ref)
+                self._datastore.put(obj, datasetRefOrType)
             except IntegrityError as e:
                 raise ConflictingDefinitionError(f"Datastore already contains dataset: {e}") from e
-            # TODO: we should re-order calls to datastore and registry and
-            # store records in _importDatasets()
-            self._registry.store_datastore_records(stored_refs)
+            return datasetRefOrType
 
-        else:
-            log.debug("Butler put: %s, dataId=%s, run=%s", datasetRefOrType, dataId, run)
-            if not self.isWriteable():
-                raise TypeError("Butler is read-only.")
-            datasetType, dataId = self._standardizeArgs(datasetRefOrType, dataId, **kwargs)
+        log.debug("Butler put: %s, dataId=%s, run=%s", datasetRefOrType, dataId, run)
+        if not self.isWriteable():
+            raise TypeError("Butler is read-only.")
+        datasetType, dataId = self._standardizeArgs(datasetRefOrType, dataId, **kwargs)
 
-            # Handle dimension records in dataId
-            dataId, kwargs = self._rewrite_data_id(dataId, datasetType, **kwargs)
+        # Handle dimension records in dataId
+        dataId, kwargs = self._rewrite_data_id(dataId, datasetType, **kwargs)
 
-            # Add Registry Dataset entry.
-            dataId = self.registry.expandDataId(dataId, graph=datasetType.dimensions, **kwargs)
+        # Add Registry Dataset entry.
+        dataId = self._registry.expandDataId(dataId, graph=datasetType.dimensions, **kwargs)
+        (ref,) = self._registry.insertDatasets(datasetType, run=run, dataIds=[dataId])
+        self._datastore.put(obj, ref)
 
-            # Add Registry Dataset entry.
-            dataId = self._registry.expandDataId(dataId, graph=datasetType.dimensions, **kwargs)
-            (ref,) = self._registry.insertDatasets(datasetType, run=run, dataIds=[dataId])
-            stored_refs = self._datastore.put_new(obj, ref)
-            # TODO: we should re-order calls to datastore and registry and
-            # store records in insertDatasets()
-            self._registry.store_datastore_records(stored_refs)
-
-        # We probably want to return a ref with datastore records, take any one
-        # from the stored refs (in case it's not empty).
-        # TODO: Disabled for now, this messes up logic in datastore.exists()
-        # and breaks unit tests.
-        # if stored_refs:
-        #     ref = next(iter(stored_refs.values()))
         return ref
 
     # TODO: remove on DM-40067.
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -237,6 +237,10 @@ class Butler(LimitedButler):
                     self._datastore = Datastore.fromConfig(
                         self._config, self._registry.getDatastoreBridgeManager(), butlerRoot=butlerRoot
                     )
+                # TODO: Once datastore drops dependency on registry we can
+                # construct datastore first and pass opaque tables to registry
+                # constructor.
+                self._registry.make_datastore_tables(self._datastore.get_opaque_table_definitions())
                 self.storageClasses = StorageClassFactory()
                 self.storageClasses.addFromConfig(self._config)
             except Exception:
@@ -1004,6 +1008,7 @@ class Butler(LimitedButler):
         collections: Any = None,
         predict: bool = False,
         run: str | None = None,
+        datastore_records: bool = False,
         **kwargs: Any,
     ) -> DatasetRef:
         """Shared logic for methods that start with a search for a dataset in
@@ -1029,6 +1034,8 @@ class Butler(LimitedButler):
         run : `str`, optional
             Run collection name to use for creating `DatasetRef` for predicted
             datasets. Only used if ``predict`` is `True`.
+        datastore_records : `bool`, optional
+            If `True` add datastore records to returned `DatasetRef`.
         **kwargs
             Additional keyword arguments used to augment or construct a
             `DataId`.  See `DataId` parameters.
@@ -1055,6 +1062,9 @@ class Butler(LimitedButler):
         if isinstance(datasetRefOrType, DatasetRef):
             if collections is not None:
                 warnings.warn("Collections should not be specified with DatasetRef", stacklevel=3)
+            # May need to retrieve datastore records if requested.
+            if datastore_records and datasetRefOrType.datastore_records is None:
+                datasetRefOrType = self._registry.get_datastore_records(datasetRefOrType)
             return datasetRefOrType
         timespan: Timespan | None = None
 
@@ -1081,7 +1091,13 @@ class Butler(LimitedButler):
             )
         # Always lookup the DatasetRef, even if one is given, to ensure it is
         # present in the current collection.
-        ref = self._registry.findDataset(datasetType, dataId, collections=collections, timespan=timespan)
+        ref = self._registry.findDataset(
+            datasetType,
+            dataId,
+            collections=collections,
+            timespan=timespan,
+            datastore_records=datastore_records,
+        )
         if ref is None:
             if predict:
                 if run is None:
@@ -1102,7 +1118,9 @@ class Butler(LimitedButler):
             # registry definition. The DatasetRef must therefore be recreated
             # using the user definition such that the expected type is
             # returned.
-            ref = DatasetRef(datasetType, ref.dataId, run=ref.run, id=ref.id)
+            ref = DatasetRef(
+                datasetType, ref.dataId, run=ref.run, id=ref.id, datastore_records=ref.datastore_records
+            )
 
         return ref
 
@@ -1419,7 +1437,9 @@ class Butler(LimitedButler):
         ``exposure`` is a temporal dimension.
         """
         log.debug("Butler get: %s, dataId=%s, parameters=%s", datasetRefOrType, dataId, parameters)
-        ref = self._findDatasetRef(datasetRefOrType, dataId, collections=collections, **kwargs)
+        ref = self._findDatasetRef(
+            datasetRefOrType, dataId, collections=collections, datastore_records=True, **kwargs
+        )
         return self._datastore.get(ref, parameters=parameters, storageClass=storageClass)
 
     def getURIs(
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1182,42 +1182,62 @@ class Butler(LimitedButler):
         TypeError
             Raised if the butler is read-only or if no run has been provided.
         """
+        if not self.isWriteable():
+            raise TypeError("Butler is read-only.")
+
         if isinstance(datasetRefOrType, DatasetRef):
             # This is a direct put of predefined DatasetRef.
             log.debug("Butler put direct: %s", datasetRefOrType)
             if run is not None:
                 warnings.warn("Run collection is not used for DatasetRef", stacklevel=3)
+            ref = datasetRefOrType
+
             # If registry already has a dataset with the same dataset ID,
             # dataset type and DataId, then _importDatasets will do nothing and
             # just return an original ref. We have to raise in this case, there
             # is a datastore check below for that.
-            self._registry._importDatasets([datasetRefOrType], expand=True)
+            self._registry._importDatasets([ref], expand=True)
             # Before trying to write to the datastore check that it does not
             # know this dataset. This is prone to races, of course.
-            if self._datastore.knows(datasetRefOrType):
-                raise ConflictingDefinitionError(f"Datastore already contains dataset: {datasetRefOrType}")
+            if self._datastore.knows(ref):
+                raise ConflictingDefinitionError(f"Datastore already contains dataset: {ref}")
             # Try to write dataset to the datastore, if it fails due to a race
             # with another write, the content of stored data may be
             # unpredictable.
             try:
-                self._datastore.put(obj, datasetRefOrType)
+                stored_refs = self._datastore.put_new(obj, ref)
             except IntegrityError as e:
                 raise ConflictingDefinitionError(f"Datastore already contains dataset: {e}") from e
-            return datasetRefOrType
+            # TODO: we should re-order calls to datastore and registry and
+            # store records in _importDatasets()
+            self._registry.store_datastore_records(stored_refs)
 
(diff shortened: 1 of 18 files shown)


andy-slac:
@timj, @TallJimbo, here is the current state of my branch, if you want to check it. Not so much code changes, but there were a lot of experiments along the way. Any comments are welcome!

timj:
I finally had a chance to look at this and given,

* We want datastores to have the option of not having to manage records at all.
* Each `DatasetRef` should carry its own datastore records.

then this looks like a reasonable approach. We do have to decide whether it's only ephemeral datastores that don't have to return records. Does SasquatchDatastore have to return empty records or only do that if `get()` is implemented?

andy-slac:
> Does SasquatchDatastore have to return empty records or only do that if `get()` is implemented?

I'm not familiar with SasquatchDatastore, but I can imagine that if it does not need datastore records stored in registry then it can return `None` or empty list/map for records.

python/lsst/daf/butler/core/datastore.py
```diff
@@ -85,6 +94,19 @@ class Event:
...
+class OpaqueTableDefinition:
+    """Definition of the opaque table which stores datastore records.
+
+    Table definition contains `.ddl.TableSpec` for a table and a class
+    of a record which must be a subclass of `StoredDatastoreItemInfo`.
+    """
+
+    __slots__ = {"table_spec", "record_class"}
+    table_spec: ddl.TableSpec
+    record_class: type[StoredDatastoreItemInfo]
```
TallJimbo, line 109: This is really a minor side issue, but I'd prefer not to assume anywhere that opaque tables are just for datastores, and I'd be great to cut down on the number of conversions (SQLAlchemy row <-> `StoredDatastoreItemInfo` <-> pydantic/json mapping).

So, could we put this `OpaqueTableDefinition` struct in `registry.interfaces.opaque`?

And could we make `record_class` just "some pydantic model type"?  Or maybe even just use bare `dict` here?   Dynamic class types aren't getting us any static-analysis advantages, and runtime checking is probably a waste of cycles (except at the JSON boundary).

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -413,11 +423,27 @@ def _get_stored_records_associated_with_refs(
...
+                    # Need to make sure they have correct type.
+                    for ref_record in ref_records:
+                        if not isinstance(ref_record, StoredFileInfo):
+                            raise TypeError(
+                                f"Datastore record has unexpected type {ref_record.__class__.__name__}"
+                            )
+                        records_by_ref[ref.id].append(ref_record)
+
+        # If there were any refs without datastore records, check opaque table.
+        records = self._table.fetch(dataset_id=[ref.id for ref in refs_with_no_records])
```
TallJimbo, line 521: This possibility is something I'd imagined disappearing from `Datastore` on DMTN-249: I was hoping we could make sure all possibly-relevant records were passed to a datastore whenever we'd get to code like this, such that if any were missing it'd represent repo corruption or logic bugs and hence the datastore should raise.  And that'd mean it wouldn't have a `self._table` at all.

Do you see this as transitional code, or something that'd stay for the long haul?

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -451,6 +447,53 @@ def put(self, inMemoryDataset: Any, ref: DatasetRef) -> None:
...
+                pass
+
+        if nsuccess == 0:
+            raise DatasetTypeNotSupportedError(f"None of the chained datastores supported ref {ref}")
+
+        if not isPermanent and npermanent > 0:
+            warnings.warn(f"Put of {ref} only succeeded in ephemeral databases", stacklevel=2)
+
+        if self._transaction is not None:
+            self._transaction.registerUndo("put", self.remove, ref)
```
TallJimbo, line 491: This is a side issue, but `DatastoreTransaction` is definitely something I'd hoped we could remove after the DMTN-249 transition is complete and we require file artifacts to be successfully written or transferred _before_ we insert the corresponding rows into registry.

I'm hoping this is still here just because this PR doesn't try to get us all the way there?

python/lsst/daf/butler/registries/sql.py
```diff
@@ -1341,6 +1350,48 @@ def queryDatasetAssociations(
                         timespan = None
                     yield DatasetAssociation(ref=ref, collection=collection_record.name, timespan=timespan)
 
+    def get_datastore_records(self, ref: DatasetRef) -> DatasetRef:
```
TallJimbo, line 1353: Eventually I think we'll really want this to be vectorized.

TallJimbo:
I think this is broadly consistent with what I think I'd have put in DMTN-249 if I hadn't been overzealous about optimizing the back-and-forth in `put` in client-server, but I need to go read that again (and maybe attempt to fix it) to be sure, at least if my question about datastores fetching missing records is resolved in favor of that being transitional.

python/lsst/daf/butler/core/datastore.py
andy-slac: I think it should be OK to move `OpaqueTableDefinition` to `registry.interfaces`. I was worried that this may cause `core` to be (logically) dependent on `registry`, but we can do it as mypy-only dependency, as we do for a couple of other classes already. We do need then to replace `StoredDatastoreItemInfo` with registry-side type (or possibly use `dict`), I need to look at the code and figure out a reasonable option.

python/lsst/daf/butler/datastores/fileDatastore.py
andy-slac: This is indeed transitional, the final goal is to remove `self._table` completely. I just keep it because not all functionality is replaced yet and there are many unit tests to be updated.

python/lsst/daf/butler/datastores/chainedDatastore.py
andy-slac: Yes, this PR is mostly to get a sense that new `DatasetRef` with all extra stuff in it looks reasonable, there are many things yet to update, including transaction handling by Datastore.

python/lsst/daf/butler/core/datastore.py
timj: I don't like the idea of completely abandoning typing and using dict for arbitrary datastore records. The `StoredDatastoreItemInfo` dataclass definition is seemingly very close to defining the database schema directly through introspection.

andy-slac: I would like to have typed datastore records, but it makes life a bit harder for registry. But it's probably unavoidable.

python/lsst/daf/butler/datastore/_datastore.py
```diff
@@ -1227,6 +1263,19 @@ def set_retrieve_dataset_type_method(self, method: Callable[[str], DatasetType |
         """
         pass
 
+    @abstractmethod
```
TallJimbo, line 1266: Should make this a `property` or add `get_` to the name.

python/lsst/daf/butler/datastore/_datastore.py
```diff
@@ -537,6 +553,26 @@ def put(self, inMemoryDataset: Any, datasetRef: DatasetRef) -> None:
         """
         raise NotImplementedError("Must be implemented by subclass")
 
+    @abstractmethod
+    def put_new(self, inMemoryDataset: Any, datasetRef: DatasetRef) -> Mapping[str, DatasetRef]:
```
TallJimbo:
```suggestion
    def put_new(self, in_memory_dataset: Any, dataset_ref: DatasetRef) -> Mapping[str, DatasetRef]:
```

python/lsst/daf/butler/_butler.py
```diff
@@ -1004,6 +1008,7 @@ def _findDatasetRef(
         collections: Any = None,
         predict: bool = False,
         run: str | None = None,
+        datastore_records: bool = False,
```
TallJimbo, line 1011: Needs a doc entry.

python/lsst/daf/butler/_dataset_ref.py
```diff
@@ -176,6 +182,10 @@ def makeDatasetId(
 # This is constant, so don't recreate a set for each instance
 _serializedDatasetRefFieldsSet = {"id", "datasetType", "dataId", "run", "component"}
 
+# Serialized representation of StoredDatastoreItemInfo collection, first item
+# is the record class name.
+_DatastoreRecords: TypeAlias = tuple[str, list[Mapping[str, Any]]]
```
TallJimbo, line 187: If this is just for serialization we may want to put that in the name.

What breaks if we don't serialize these records at all?  I am a bit worried that if we start including them in `SerializedDatasetRef` by default we might blow up the size of `QuantumGraph` without solving any problems (yet).

python/lsst/daf/butler/datastore/generic_base.py
```diff
@@ -32,136 +32,35 @@
...
-        else:
-            # There are only two columns and all that matters is the
-            # dataset ID.
-            self.bridge.ensure(registryRefs.values())
-        self.addStoredItemInfo(expandedRefs, expandedItemInfos, insert_mode=insert_mode)
-
     def _post_process_get(
         self,
-        inMemoryDataset: Any,
+        inMemoryDataset: object,
```
TallJimbo, line 59: Any particular reason for `object` vs. `Any` here?

python/lsst/daf/butler/datastore/record_data.py
```diff
@@ -63,15 +61,17 @@ class SerializedDatastoreRecordData(_BaseModelCompat):
     dataset_ids: list[uuid.UUID]
     """List of dataset IDs"""
 
-    records: Mapping[str, Mapping[str, list[_Record]]]
-    """List of records indexed by record class name and table name."""
+    records: Mapping[str, Mapping[str, Mapping[str, list[_Record]]]]
+    """List of records indexed by record class name, dataset ID (encoded as
+    str, because JSON), and opaque table name.
+    """
```
TallJimbo, line 67: It's too bad we can't use opaque table name to get the record class name instead of storing them both, but I see this is preexisting (and just used in `Quantum` and `QuantumBackedButler`, right?) so maybe this is not the time to try to fix that.

python/lsst/daf/butler/datastore/stored_file_info.py
```diff
@@ -87,13 +89,14 @@ def from_record(cls: type[StoredDatastoreItemInfo], record: Mapping[str, Any]) -
         """
         raise NotImplementedError()
 
-    def to_record(self) -> dict[str, Any]:
-        """Convert record contents to a dictionary."""
-        raise NotImplementedError()
+    def to_record(self, **kwargs: Any) -> dict[str, Any]:
```
TallJimbo, line 92: This may not be time to change it, but I'm not a fan of the name of this; we call these things "records" before we convert them to `dict`, so there's nothing here that implies conversion to `dict`.

python/lsst/daf/butler/datastore/stored_file_info.py
```diff
@@ -102,12 +105,56 @@ def update(self, **kwargs: Any) -> StoredDatastoreItemInfo:
...
+            raise TypeError(f"Class {class_name} is not a subclass of StoredDatastoreItemInfo")
+        return [klass.from_record(record) for record in records]
+
 
 @dataclass(frozen=True)
 class StoredFileInfo(StoredDatastoreItemInfo):
     """Datastore-private metadata associated with a Datastore file."""
 
-    __slots__ = {"formatter", "path", "storageClass", "component", "checksum", "file_size", "dataset_id"}
+    __slots__ = {"formatter", "path", "storageClass", "component", "checksum", "file_size"}
```
TallJimbo: We can use `dataclass(..., slots=True)` now to avoid spelling these out.

python/lsst/daf/butler/datastore/stored_file_info.py
```diff
@@ -239,14 +280,15 @@ def from_record(cls: type[StoredFileInfo], record: Mapping[str, Any]) -> StoredF
         storageClass = cls.storageClassFactory.getStorageClass(record["storage_class"])
         component = record["component"] if (record["component"] and record["component"] != NULLSTR) else None
 
+        # UUID may be converted to string, e.g. in round-trip through JSON,
+        # convert it back to UUID.
```
TallJimbo: This comment doesn't seem to be describing any code around it.

python/lsst/daf/butler/datastores/inMemoryDatastore.py
```diff
@@ -62,7 +60,7 @@ class StoredMemoryItemInfo(StoredDatastoreItemInfo):
     DatasetRef.
     """
 
-    __slots__ = {"timestamp", "storageClass", "parentID", "dataset_id"}
+    __slots__ = {"timestamp", "storageClass", "parentID"}
```
TallJimbo: Another opportunity for passing `slots=True` to the `dataclass` decorator.

python/lsst/daf/butler/registries/sql.py
```diff
@@ -219,6 +221,10 @@ def __init__(self, database: Database, defaults: RegistryDefaults, managers: Reg
         # initialized, and must be done before the property getter is used.
         self.defaults = defaults
 
+        # TODO: This is currently initialized by `make_datastore_tables`,
+        # eventually we'll need to do it during construction.
+        self._datastore_record_classes: Mapping[str, type[StoredDatastoreItemInfo]] = {}
```
TallJimbo, line 227: Are keys here datastore names or opaque table names?

I also wonder if it'd be better to just have a mapping of `DatastoreOpaqueTable`.

python/lsst/daf/butler/registry/_butler_registry.py
```diff
@@ -216,3 +219,53 @@ def getDatastoreBridgeManager(self) -> DatastoreRegistryBridgeManager:
...
+
+        Returns
+        -------
+        updated_ref : `DatasetRef`
+            Dataset reference with filled datastore records.
+
+        Notes
+        -----
+        If this method is called with the dataset ref that is not known to the
+        registry then the reference with an empty set of records is returned.
```
TallJimbo, line 241: I'm a bit surprised by this behavior; generally in methods that add records to refs, I think we want:

- `DatasetRef._datastore_records = None`: this ref might have records, but we haven't tried to "expand" the ref to get them yet;
- `DatasetRef._datastore_Records = {}`: this ref is known to the registry but is not in any datastore and hence has no datastore records;
- (raise an exception): this ref is not known to the registry at all.

But if it's expensive or otherwise inconvenient to maintain that invariant maybe it's not worth it.

TallJimbo (approved):
I've left a lot of little comments and questions but I have no big-picture concerns; it's all a step in the right direction.

FWIW, your `put_new` does not match the one I have been prototyping with transaction support over on DMTN-249 anymore, and I think it'll probably end up changing again when we do transactions and/or client-server `put`, but my prototyping doesn't include `ChainedDatastore` so I'm not confident it will work as-is either.

python/lsst/daf/butler/datastore/_datastore.py
andy-slac: I'll add `get_`, I'm not sure if property is the correct thing for this.

python/lsst/daf/butler/_dataset_ref.py
andy-slac: You are right, we don't have an immediate use for this, so it may be better to drop it completely. I guess DM-41115 will take care of datastore record serialization for future QuantumGraph?

python/lsst/daf/butler/datastore/generic_base.py
andy-slac: This is mostly personal preference, I think mypy can catch more potential issues with `object` as `Any` disables all checks. Some docstrings already specified `object` for parameters, this is just restoring documented behavior 🙂

andy-slac: I can revert to `Any` if consistency is important.

python/lsst/daf/butler/datastore/record_data.py
andy-slac: The mapping from opaque table name to record class would have to exist somewhere when de-serialization happens. I'm not sure what would be the right place for it.

python/lsst/daf/butler/datastore/stored_file_info.py
andy-slac: Yes, the naming is confusing, we call instances infos or records in different places. I can rename this method (and `from_record`, `to_records`, `from_records`) to `as_dict`, etc. @timj, do you have preference for keeping this naming or some other names?

python/lsst/daf/butler/datastore/stored_file_info.py
andy-slac: Leftover for previous commits, I dropped dataset_id from the class, will remove this comment.

python/lsst/daf/butler/registries/sql.py
andy-slac: Keys are opaque table names, I'll update comment. Registry itself does not need `TableSpec` part of `DatastoreOpaqueTable`, depending on how we initialize this structure in the future it may be better not to require `TableSpec` at that point.

python/lsst/daf/butler/registry/_butler_registry.py
andy-slac: The complication here is InMemoryDatastore which does not have any records. I want to avoid raising an exception in that case because Butler would need then to figure out what to do with it. This is not ideal of course as we now depend on Butler+Datastore doing correct thing when the records are empty. The alternative is to do an additional check for dataset_id in that case, which is one more query on a large table. I think I want to see additional use cases for this method to decide if this needs change.

python/lsst/daf/butler/_dataset_ref.py
TallJimbo: > I guess DM-41115 will take care of datastore record serialization for future QuantumGraph?

Yes, that's how I'm hoping things will go.

python/lsst/daf/butler/datastore/record_data.py
TallJimbo: I'm fine with leaving this.  DM-41115 may also change this completely anyway.

python/lsst/daf/butler/datastore/generic_base.py
TallJimbo: `object` is fine.  I was mostly curious for my own education - I knew mypy was a bit more permissive with `Any` but I didn't have any concrete examples of how it was different in my mind.

python/lsst/daf/butler/datastore/stored_file_info.py
TallJimbo: Renaming `StoredDatastoreItemInfo` -> `DatastoreRecord`  and `StoredFileInfo` -> `FileDatastoreRecord` would be ideal but maybe not a good idea (especially on this ticket).

I do think `to_dict` is better than `to_records` (same with `from_`), but I think we may well want these to be native pydantic models, which would mean we'd just call `model_dump` instead of having another method.

python/lsst/daf/butler/registry/_butler_registry.py
TallJimbo: If the answer is "because InMemoryDatastore" then let's just revisit it when that's dead and gone.

TallJimbo (approved):

python/lsst/daf/butler/datastore/stored_file_info.py
andy-slac: OK, I'll leave these as they are for now, and wait for big refactoring.

python/lsst/daf/butler/datastore/generic_base.py
andy-slac: It's probably mostly about accidental corner cases, but here is a trivial example to show differences:
```
from typing import Any

class Butler():
    def get_obj(self) -> object:
        return 42

    def get_any(self) -> Any:
        return 42

b = Butler()

x = [42]
x = b.get_obj()  # will be mypy error
x = b.get_any()  # mypy is OK
l = len(x)  # mypy is still OK, as x is Any here

y = b.get_obj()
y.something()  # mypy error

z = b.get_any()
z.something()  # mypy is OK
```

TallJimbo: Ah, interesting; `object` is _much_ less permissive than I thought - it seems to be an object you can't do anything with (except basic stuff like equality tests), while `Any` permits anything (including implicit downcasts).

python/lsst/daf/butler/_dataset_ref.py
andy-slac: Oops, I forgot to undo this part, will make another ticket.

PR #877, opened 2023-08-11
DM-40359: Implement Butler.get for matplotlib generated png files

fred3m:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/formatters/matplotlib.py
+++ b/python/lsst/daf/butler/formatters/matplotlib.py
@@ -27,6 +27,9 @@ __all__ = ("MatplotlibFormatter",)
 
 from typing import Any
 
+import matplotlib.pyplot as plt
+import numpy as np
+
 from .file import FileFormatter
 
 
@@ -38,8 +41,22 @@ class MatplotlibFormatter(FileFormatter):
 
     def _readFile(self, path: str, pytype: type[Any] | None = None) -> Any:
         # docstring inherited from FileFormatter._readFile
-        raise NotImplementedError(f"matplotlib figures cannot be read by the butler; path is {path}")
+        return plt.imread(path)
 
     def _writeFile(self, inMemoryDataset: Any) -> None:
         # docstring inherited from FileFormatter._writeFile
         inMemoryDataset.savefig(self.fileDescriptor.location.path)
+
+    @staticmethod
+    def fromArray(cls: np.ndarray) -> plt.Figure:
+        """Convert an array into a Figure."""
+        fig = plt.figure()
+        plt.imshow(cls)
+        return fig
+
+    @staticmethod
+    def dummyCovnerter(cls: np.ndarray) -> np.ndarray:
+        """This converter exists to trick the Butler into allowing
+        a numpy array on read with ``storageClass='NumpyArray'``.
+        """
+        return cls
--- a/tests/test_matplotlibFormatter.py
+++ b/tests/test_matplotlibFormatter.py
@@ -27,6 +27,8 @@ import tempfile
 import unittest
 from random import Random
 
+import numpy as np
+
 try:
     import matplotlib
 
@@ -78,8 +80,13 @@ class MatplotlibFormatterTestCase(unittest.TestCase):
             pyplot.gcf().savefig(file.name)
             self.assertTrue(filecmp.cmp(local.ospath, file.name, shallow=True))
         self.assertTrue(butler.exists(ref))
-        with self.assertRaises(ValueError):
-            butler.get(ref)
+
+        fig = butler.get(ref)
+        # Ensure that the result is a figure
+        self.assertTrue(isinstance(fig, pyplot.Figure))
+        image = butler.get(ref, storageClass="NumpyArray")
+        self.assertTrue(isinstance(image, np.ndarray))
+
         butler.pruneDatasets([ref], unstore=True, purge=True)
         self.assertFalse(butler.exists(ref))
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -18,6 +18,9 @@ ignore_missing_imports = True
 [mypy-numpy.*]
 ignore_missing_imports = True
 
+[mypy-matplotlib.*]
+ignore_missing_imports = True
+
 [mypy-pyarrow.*]
 ignore_missing_imports = True
 
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -342,10 +342,14 @@ storageClasses:
       dict: lsst.utils.packages.Packages
   NumpyArray:
     pytype: numpy.ndarray
+    converters:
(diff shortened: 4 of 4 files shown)


tests/test_matplotlibFormatter.py
```diff
@@ -27,6 +27,8 @@
 import unittest
 from random import Random
 
+import numpy as np
```
timj, line 30: This needs to be a protected import since numpy is not butler requirement (although effectively it is since astropy won't work without it).

tests/test_matplotlibFormatter.py
```diff
@@ -78,8 +80,13 @@ def testMatplotlibFormatter(self):
             pyplot.gcf().savefig(file.name)
             self.assertTrue(filecmp.cmp(local.ospath, file.name, shallow=True))
         self.assertTrue(butler.exists(ref))
-        with self.assertRaises(ValueError):
-            butler.get(ref)
+
+        fig = butler.get(ref)
+        # Ensure that the result is a figure
+        self.assertTrue(isinstance(fig, pyplot.Figure))
```
timj, line 86:
```suggestion
        self.assertIsinstance(fig, pyplot.Figure)
```

tests/test_matplotlibFormatter.py
```diff
@@ -78,8 +80,13 @@ def testMatplotlibFormatter(self):
...
+        image = butler.get(ref, storageClass="NumpyArray")
+        self.assertTrue(isinstance(image, np.ndarray))
```
timj, line 88:
```suggestion
        self.assertIsinstance(image, np.ndarray)
```

python/lsst/daf/butler/formatters/matplotlib.py
```diff
@@ -38,8 +41,22 @@ class MatplotlibFormatter(FileFormatter):
...
+
+    @staticmethod
+    def fromArray(cls: np.ndarray) -> plt.Figure:
+        """Convert an array into a Figure."""
+        fig = plt.figure()
+        plt.imshow(cls)
+        return fig
+
+    @staticmethod
+    def dummyCovnerter(cls: np.ndarray) -> np.ndarray:
```
timj, line 58: The name has a typo in it. Also see comment about naming I make above.

I am confused by the signature. Surely this should take a `plt.Figure` and return an `np.ndarray`? Is there no way to do that conversion? It is going to be problematic in some contexts (eg in-memory datastore) if this method is not implemented.

python/lsst/daf/butler/formatters/matplotlib.py
```diff
@@ -38,8 +41,22 @@ class MatplotlibFormatter(FileFormatter):
...
         # docstring inherited from FileFormatter._readFile
-        raise NotImplementedError(f"matplotlib figures cannot be read by the butler; path is {path}")
+        return plt.imread(path)
 
     def _writeFile(self, inMemoryDataset: Any) -> None:
         # docstring inherited from FileFormatter._writeFile
         inMemoryDataset.savefig(self.fileDescriptor.location.path)
+
+    @staticmethod
+    def fromArray(cls: np.ndarray) -> plt.Figure:
```
timj, line 51: This does not need to be a static method in the formatter class. It's okay for this to be a private function in the file that is not exported and only referenced from the yaml file. The first parameter should not be named `cls`, it should be `array` or something.

python/lsst/daf/butler/formatters/matplotlib.py
```diff
@@ -38,8 +41,22 @@ class MatplotlibFormatter(FileFormatter):
 
     def _readFile(self, path: str, pytype: type[Any] | None = None) -> Any:
         # docstring inherited from FileFormatter._readFile
-        raise NotImplementedError(f"matplotlib figures cannot be read by the butler; path is {path}")
+        return plt.imread(path)
```
timj, line 44: Given that this is returning something unexpected (and effectively backwards from expectations) please add a comment saying that it's not returning a figure.

timj (changes requested):
Thanks for doing this. Is it impossible for the Figure->numpy conversion to work?

I need to think about how easy it would be to add in-memory datastore tests like we do in test_parquet.py. As things stand I think that will break because of the dummy converter not being real.

python/lsst/daf/butler/formatters/matplotlib.py
fred3m: No, this is just what the name says, a dummy converter. This is what I was trying to get across in slack: In some cases we want to just load the png file directly, although a  numpy array is fine. Yes, it is possible to convert a Figure to a numpy array, but it's messy. You basically have to create a dummy file, save the figure to the dummy file, then read the dummy file as a numpy array. 👎 

I did the cursory stack overflow search, and that was the recommended solution. I also looked at the matplotlib codebase a bit and it does indeed look like it would be non-trivial to try and convert the figure to a numpy array directly. This is probably why the method wasn't implemented in the first place.

The other solution that I thought of is that we could define a `PNG` store class, which would be an `IoBytes` instance, and convert plots and arrays to and from the `PNG` class. In other words
```
@staticmethod
def fig_to_png(fig):
    buf = IoBytes()
    fig.savefig(buf)
    return buf

@staticmethod
def png_to_array(png):
    return plt.imread(png)

@staticmethod
def png_to_fig(png):
    arr = png_to_array(png)
    fig = plt.figure()
    plt.imshow(arr)
    return fig
```

fred3m:
The other thing that I'll add is that in general you don't want to load a saved figure as a figure, as what you're actually loading is a png file that doesn't have information about the axes, size of the figure, etc. So the dummy figure that is loaded is rarely (if ever) what the user actually desires.

timj:
Yes, I understand that what you get back isn't what you put in, and that is a bit of a problem in butler in general. The problem here though is that if an in-memory datastore is involved behind the scenes everything is going to break because you put a Figure and then you do a butler.get() and all it has is a Figure -- you get back what you put in but if you ask for the NumpyArray version by specifying a different storage class it won't work and will fall over because there is no functioning converter to even attempt it. In the general case you can't tell what the datastore is doing. If the converter has to use a BytesIO buffer to pretend to write a file to disk and then call imread on that buffer with format="png" then that would help a lot.

python/lsst/daf/butler/formatters/matplotlib.py
timj: Completely untested but can something like this work?

```python
def fig_to_np(fig: pyplot.figure.Figure) -> np.ndarray:
    buf = io.BytesIO()
    fig.savefig(buf, format="png")
    return plt.imread(buf, format="png")
```
?

timj:
There is code in the [wild](https://jira.lsstcorp.org/browse/DM-38358) that does:

```python
ref = butler.registry.findDataset(plot_name, detector=detector)
print ("Plot number", plot_name)
uri = butler.getURI(ref)
display(Image(data=uri.read()))
```

so we should consider adding a `IPythonImage` StorageClass definition and adding a simply converter that does this internally so people would only need to do `butler.get(ref, storageClass="IPythonImage")` instead.

TallJimbo:
I like the idea of using storage class conversions to deal with this.

PR #878, opened 2023-08-11
DM-40366: Use default pydantic model for testing

timj:
Previously we changed the pydantic models used in testing to be the compatibility models that understand v1 and v2. This is fine but means that any external users using pydantic models as butler datasets break if they are still using v1.

Change the test model to use whatever is the native pydantic and then the YAML formatter has been modified to support both APIs. The JSON formatter already supported both.

## Checklist

- [x] ran Jenkins


tgoldina (approved):
Thank you! The YamlFormatter failure with pydantic<2.0 is fixed.

PR #879, opened 2023-08-16
DM-40381: Deduplicate when merging DatastoreRecordData and document preconditions.

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-40381.bugfix.md
+++ b/doc/changes/DM-40381.bugfix.md
@@ -1,0 +1,1 @@
+Ensure Datastore record exports (as used in quantum-backed butler) are deduplicated when necessary.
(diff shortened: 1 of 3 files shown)


timj (approved):
Looks okay once I realized that the idea here is that there are fully populated subsets of `DatastoreRecordData` floating around and when you call `update` you are noting that, assuming these are all from the same datastore, that if one record for a dataset_id is already present there is no need to copy over the records from the incoming object. If there is a mismatch in definitions for that record we have bigger problems.

Ideally DatastoreRecordData would have some idea about the primary keys that StoredDatastoreItemsInfo could report to it.

TallJimbo:
> Looks okay once I realized that the idea here is that there are fully populated subsets of DatastoreRecordData floating around and when you call update you are noting that, assuming these are all from the same datastore, that if one record for a dataset_id is already present there is no need to copy over the records from the incoming object. If there is a mismatch in definitions for that record we have bigger problems.

Yes, that's exactly it.

> Ideally DatastoreRecordData would have some idea about the primary keys that StoredDatastoreItemsInfo could report to it.

For our current limited use cases for this export API (i.e. just QBB), I like keeping DatastoreRecordData being unaware of the real primary key, and just declaring that only full datasets (not components) can be exported this way.  It's a lot simpler than any alternative I can think of, and I can't think of a scenario where it'd be a serious limitation.

PR #880, opened 2023-08-16
DM-40395: Allow datastore record inserts to use ensure or replace

timj:
This doesn't solve the real problem of the files being overwritten without us knowing whether the datastore records are allowed to be modified but it does give the option.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -1037,7 +1037,6 @@ class FileDatastore(GenericBaseDatastore):
         record_validation_info: bool = True,
     ) -> None:
         # Docstring inherited from Datastore._finishIngest.
-        uses_uuid_v5 = True
         refsAndInfos = []
         progress = Progress("lsst.daf.butler.datastores.FileDatastore.ingest", level=logging.DEBUG)
         for dataset in progress.wrap(prepData.datasets, desc="Ingesting dataset files"):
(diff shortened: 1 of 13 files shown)


python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -365,10 +365,20 @@ def _delete_artifact(self, location: Location) -> None:
             raise
         log.debug("Successfully deleted file: %s", location.uri)
 
-    def addStoredItemInfo(self, refs: Iterable[DatasetRef], infos: Iterable[StoredFileInfo]) -> None:
+    def addStoredItemInfo(
+        self, refs: Iterable[DatasetRef], infos: Iterable[StoredFileInfo], insert_mode: str = "insert"
```
andy-slac: Maybe use `Literal["insert", ...]` instead of `str`?

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1036,7 +1047,16 @@ def _finishIngest(
                 record_validation_info=record_validation_info,
             )
             refsAndInfos.extend([(ref, info) for ref in dataset.refs])
-        self._register_datasets(refsAndInfos)
+            for ref in dataset.refs:
+                if ref.id.version != 5:
+                    uses_uuid_v5 = False
```
andy-slac: What happens if there is a mix of v4/v5 UUIDs? Should we split them into two groups? If we don't care then maybe add `break` after this assignment?

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -2729,7 +2749,10 @@ def transfer_from(
                 "" if len(direct_transfers) == 1 else "s",
             )
 
-        self._register_datasets(artifacts)
+        # We are overwriting previous datasets that may have already
+        # existed. We therefore should ensure that we force the
+        # datastore records to agree.
```
andy-slac: Could it happen that old and new datasets have different set of components?

tests/test_datastore.py
```diff
@@ -1101,6 +1102,36 @@ def testChecksum(self) -> None:
...
+            v4ref.datasetType, v4ref.dataId, v4ref.run, id_generation_mode=DatasetIdGenEnum.DATAID_TYPE_RUN
+        )
+
+        with lsst.utils.tests.getTempFilePath(".yaml", expectOutput=True) as path:
+            with open(path, "w") as fd:
+                yaml.dump(metrics._asdict(), stream=fd)
+
+            datastore.ingest(FileDataset(path=path, refs=v4ref), transfer="direct")
+
+            # This will fail because the ref is not using UUIDv4.
```
andy-slac:
```suggestion
            # This will fail because the ref is using UUIDv4.
```
?

andy-slac (approved):
Looks good, a couple of minor questions.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: Should I use an enum? This is an internal interface so maybe that is better.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: Hmm. Maybe I should split them into groups. I'm not expecting it to happen but you are right that it's the consistent thing to do with the current brokenness.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: For ingest we do not do disassembly so replace should always be fine. For arbitrary transfers in this method I suppose it is possible that someone ran the graph in disassemble mode, transferred it, then re-ran the graph in standard mode and tried to transfer it again. We don't really want that to happen from a provenance perspective and ideally transferring something that you've already got should result in it being skipped -- `registry._insertDatasets` doesn't help us here and race conditions in datastore if transfers are happening concurrently doesn't help us either. I think I should switch it back to "insert" for now and leave the real fix for the other ticket.

python/lsst/daf/butler/datastores/fileDatastore.py
andy-slac: 👍 Enum is probably more natural here.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: I now split them up.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: I've added an extra commentary but still use replace. We really need to sort this out properly.

PR #882, opened 2023-08-17
DM-40392: Add methods to update run collection name in a quantum

andy-slac:
QuantumGraph.updateRun() method needs a better implementation and these new methods in DatasetRef and Quantum classes are for supporting that implementation.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/core/datasets/ref.py
+++ b/python/lsst/daf/butler/core/datasets/ref.py
@@ -706,11 +706,48 @@ class DatasetRef:
             A new dataset reference that is the same as the current one but
             with a different storage class in the `DatasetType`.
         """
+        return self.replace(storage_class=storageClass)
+
+    def replace(
+        self,
+        *,
+        id: DatasetId | None = None,
+        run: str | None = None,
+        storage_class: str | StorageClass | None = None,
+    ) -> DatasetRef:
+        """Create a new `DatasetRef` from this one, but with some modified
+        attributes.
+
+        Parameters
+        ----------
+        id : `DatasetId` or `None`
+            If not `None` then update dataset ID.
+        run : `str` or `None`
+            If not `None` then update run collection name. If ``dataset_id`` is
+            `None` then this will also cause new dataset ID to be generated.
+        storage_class : `str` or `StorageClass` or `None`.
+            The new storage class. If not `None`, replaces existing storage
+            class.
+
+        Returns
+        -------
+        modified : `DatasetRef`
+            A new dataset reference with updated attributes.
+        """
+        if storage_class is None:
+            datasetType = self.datasetType
+        else:
+            datasetType = self.datasetType.overrideStorageClass(storage_class)
+        if run is None:
+            run = self.run
+            # Do not regenerate dataset ID if run is the same.
+            if id is None:
+                id = self.id
         return DatasetRef(
-            datasetType=self.datasetType.overrideStorageClass(storageClass),
+            datasetType=datasetType,
             dataId=self.dataId,
-            id=self.id,
-            run=self.run,
+            run=run,
+            id=id,
             conform=False,
         )
 
--- a/tests/test_datasets.py
+++ b/tests/test_datasets.py
@@ -643,6 +643,27 @@ class DatasetRefTestCase(unittest.TestCase):
             # of "object" which is compatible with everything.
             ref_new.overrideStorageClass(incompatible_sc)
 
+    def testReplace(self) -> None:
+        """Test for `DatasetRef.replace` method."""
+        ref = DatasetRef(self.datasetType, self.dataId, run="somerun")
+
+        ref2 = ref.replace(run="somerun2")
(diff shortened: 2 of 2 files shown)


timj:
The ruff problem comes from ruff 0.285 with astral-sh/ruff#6465

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -706,11 +706,48 @@ def overrideStorageClass(self, storageClass: str | StorageClass) -> DatasetRef:
             A new dataset reference that is the same as the current one but
             with a different storage class in the `DatasetType`.
         """
+        return self.update(storage_class=storageClass)
+
+    def update(
```
timj: Maybe `replace()` to match the name used in `dataclasses`? `update` doesn't immediately suggest that a new version is being returned. I called the equivalent `ResourcePath.replace` for that reason.

python/lsst/daf/butler/core/dimensions/_records.py
```diff
@@ -190,7 +190,7 @@ def direct(
         # This method requires tuples as values of the mapping, but JSON
         # readers will read things in as lists. Be kind and transparently
         # transform to tuples
-        _recItems = {k: v if type(v) != list else tuple(v) for k, v in record.items()}  # type: ignore
+        _recItems = {k: v if not isinstance(v, list) else tuple(v) for k, v in record.items()}  # type: ignore
```
timj: I think we deliberately used `type()` here for speed since we know it can only be a list coming from JSON and we need this to be fast. We don't want to worry about subclasses of list here.

andy-slac: Latest ruff was complaining about it, should I just suppress this?

andy-slac: And why don't we always use `tuple(v)`? `tuple(v)` will return `v` if it's already a tuple.

timj: Hmm. That might be a better answer. Let the tuple() constructor deal with it.

andy-slac: Ah, not all items are supposed to be tuples there, sorry for the noise.

timj: Right, we only want lists to migrate. I think adding a noqa will work but you'll also have to adjust pyproject.toml to stop ruff complaining about unused noqa -- the rubin-env ruff used in Jenkins will not trigger an error and will instead get upset by the unused noqa.

python/lsst/daf/butler/core/datasets/ref.py
```diff
@@ -706,11 +706,48 @@ def overrideStorageClass(self, storageClass: str | StorageClass) -> DatasetRef:
...
+            If not `None` then update run collection name. If ``dataset_id`` is
+            `None` then this will also cause new dataset ID to be generated.
+        storage_class : `str` or `StorageClass` or `None`.
+            The new storage class. If not `None`, replaces existing storage
+            class.
+
+        Returns
+        -------
+        modified : `DatasetRef`
+            A new dataset reference with updated attributes.
```
timj, line 735: If all 3 parameters are None do you want to return `self` instead?

python/lsst/daf/butler/core/quantum.py
```diff
@@ -569,6 +569,87 @@ def _reduceFactory(
...
+        dataset_id_map : `~collections.abc.MutableMapping`
+            The mapping with correspondence of the old dataset IDs to new
+            dataset IDs, filled previously by `update_output_run`.
+        """
+
+        def _update_input_refs(
+            refs: Iterable[DatasetRef], run: str, dataset_id_map: Mapping[DatasetId, DatasetId]
+        ) -> Iterator[DatasetRef]:
+            for ref in refs:
+                if dataset_id := dataset_id_map.get(ref.id):
```
timj: Add comment that this is for intermediates only and if it's not already in dataset_id_map the ref won't be modified?

timj (approved):
Please add trivial test for update_output_run and update_input_run.

If these two methods have to be called in exactly the right way and solely from `QuantumGraph.updateRun` maybe they should be private methods in the sense that middleware sometimes knows it can call private methods in different packages but we don't want them public? Else it begs the question why isn't there a `quantum.update_run` that combines these two methods?

python/lsst/daf/butler/core/datasets/ref.py
andy-slac: I could but I do not want an additional test for all `None`, I don't think it's going to used with no parameters.

andy-slac:
I agree that those two new methods in `Quantum` should be considered private. Maybe I need to move both of them to `Quantumgraph` as that is the only place that needs them, let me try that.

andy-slac:
I moved that private code to `QuantumGraph.updateRun`, does not look too terrible.

PR #887, opened 2023-09-07
DM-40582: Add support for serialization/deserialization of Arrow schemas to Parquet.

erykoff:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/formatters/parquet.py
+++ b/python/lsst/daf/butler/formatters/parquet.py
@@ -458,9 +458,9 @@ def astropy_to_arrow(astropy_table: atable.Table) -> pa.Table:
     for name, pa_type in type_list:
         field_metadata = {}
         if description := astropy_table[name].description:
-            field_metadata["doc"] = description
-        if units := astropy_table[name].unit:
-            field_metadata["units"] = str(units)
+            field_metadata["description"] = description
+        if unit := astropy_table[name].unit:
+            field_metadata["unit"] = str(unit)
         fields.append(
             pa.field(
                 name,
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -455,7 +453,23 @@ def astropy_to_arrow(astropy_table: atable.Table) -> pa.Table:
...
 
-    schema = pa.schema(type_list, metadata=md)
+    # Convert type list to fields with metadata.
+    fields = []
+    for name, pa_type in type_list:
+        field_metadata = {}
+        if description := astropy_table[name].description:
+            field_metadata["doc"] = description
+        if units := astropy_table[name].unit:
+            field_metadata["units"] = str(units)
```
TallJimbo: This is probably the point where we decide whether we want this to be "units" or "unit".  I may have put "units" in the drp_tasks PR that spawned this, but if astropy uses "unit" maybe we should, too.

TallJimbo (approved):
If I'm reading this correctly, we can write only `ArrowSchema` but read any of the convertible schema types?  That's fine with me, especially to minimize scope on this ticket, as long as what happens if someone tries to write one of the other schema types is a sufficiently graceful failure.

python/lsst/daf/butler/formatters/parquet.py
erykoff: This is an interesting question.  Right now, our afw Fields use `doc`/`units`, which is what I thought you were going for here.  Astropy tables use `description`/`unit`, which I'm using as exact convertibles.  And right here we're defining the precedent for what should go into an arrow schema.  So I think that either we should go with the afw convention or the astropy convention.

erykoff:
I can add support for serializing the other ones on this ticket; that makes sense.  Though I don't know exactly why you'd want to with anything but the arrow or perhaps astropy.

python/lsst/daf/butler/formatters/parquet.py
TallJimbo: I vote for astropy conventions, then.

TallJimbo:
> I can add support for serializing the other ones on this ticket; that makes sense. Though I don't know exactly why you'd want to with anything but the arrow or perhaps astropy.

Don't worry about it.  I see that those didn't get associated with `ParquetFormatter` in the formatters configuration, so they should already fail gracefully if you try to write them.

erykoff:
Okay, I'll leave this as just being able to serialize an Arrow schema (as described on the ticket) and in the future we can do others if it seems useful/necessary.

PR #888, opened 2023-09-14
DM-35599: Add transfer option that does not include -t

timj:
This allows pipetask to use the --transfer option without also overriding its own -t option.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/cli/opt/options.py
+++ b/python/lsst/daf/butler/cli/opt/options.py
@@ -48,6 +48,7 @@ __all__ = (
     "run_option",
     "transfer_dimensions_option",
     "transfer_option",
+    "transfer_option_no_short",
     "verbose_option",
     "where_option",
     "order_by_option",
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/cli/opt/options.py
```diff
@@ -255,6 +253,17 @@ def _config_split(*args: Any) -> dict[str, str]:
     ),
 )
 
+transfer_option_no_short = MWOptionDecorator(
```
timj, line 256: I couldn't really think of a good name...

TallJimbo: It's fine with me.

More broadly I think just one example of our approach to sharing command-line options not being quite right for the problem.  I see it most often in the `--help` strings, where we should probably should have a lot more context-specific information about how an option interacts with others, but the pattern we use discourages it (or, worse, the `--help` string includes stuff specific to its original context and we never notice that it's confusing for other contexts introduce later).

TallJimbo (approved):

python/lsst/daf/butler/cli/opt/options.py
timj: You can override the help string, indeed we do exactly that for the `--transfer` option in ctrl_mpexec. We can modify any kwarg, but the option strings themselves aren't kwargs and so can't be removed, only augmented. Another approach is to remove all the short options from the default options and require that people add the short form explicitly if they want it when they use it.

Standard options are important for the long names and for things like transfer where you really don't want to have to repeat all the allowed transfer options.

TallJimbo: Yeah, I completely agree we need some form of information sharing for the options.  And I can't claim to have understood the situation well enough to propose anything like a detailed alternative.  I'm more or less just being grumpy.

PR #891, opened 2023-09-29
DM-36303: Replace RFC-878/879 deprecations with removals

timj:
## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/direct_butler.py
```diff
@@ -1229,8 +1237,11 @@ def find_dataset(
         )
         if ref is not None and dimension_records:
             ref = ref.expanded(self._registry.expandDataId(ref.dataId, dimensions=ref.datasetType.dimensions))
+        if ref is not None and component_name:
+            ref = ref.makeComponentRef(component_name)
```
timj, line 1241: This change is sufficient to make all the `Butler.get()` tests with components work. I do understand that doing this means that `Butler.find_dataset` is handling components but `Butler.query_datasets()` does not. If we don't like this the above code will have to move into `_findDatasetRef`.

python/lsst/daf/butler/_registry_shim.py
```diff
@@ -281,7 +281,7 @@ def queryDatasetTypes(
         self,
         expression: Any = ...,
         *,
-        components: bool | None = None,
+        components: bool = False,
```
timj: When I was working on #937, it occurred to me that we really want to be warning if people set the components flag to False. What do you think about having this be something like:

```python
components : bool | _Sentinel = _DefaultSentinel
```
and then warning `if components is False` ?

timj (commented):
The rebase took a while because the first batch of changes turned out to already be applied on main but in slightly different ways. I think everything looks reasonable now and all tests pass and mypy is happy.

We do still issue a warning for components in the tests because we are calling `_iter_by_dataset_type` which calls the `.components` property on the result (and then calls `withComponents`. Presumably that code could be cleaned up as well.

python/lsst/daf/butler/_registry_shim.py
TallJimbo: :+1:, but we should instead warn `if components is not _DefaultSentinal` in case somebody is still passing `None` and isn't running MyPy.

python/lsst/daf/butler/direct_butler.py
TallJimbo: I'm fine with handling components here like this.

TallJimbo (approved):
`registry.queries.ParentDatasetQueryResults.withComponents` seems to have escaped deprecation originally, so we should definitely add that deprecation now.  I think that means we also have to defer removing the `_components` attribute until that's removed; after other changes on this ticket it should be possible for it to be constructed with anything other than `components=[None]` _except_ for via `withComponents`.

And then I think we want to reimplement `_iter_by_dataset_type` to just use `_components` and duplicate the implementation of `withComponents` internally.  That will keep it from emitting any warnings of its own, while the deprecation of `withComponents` will close the last remaining avenue for making it actually iterate over components without a warning.

There's also the issue that the "parent" in that class name and in `byParentDatasetTypes` is not so great anymore.  I'm inclined to leave it on the old registry result types, but we should fix the `Butler._query_results` versions since those aren't public yet.

python/lsst/daf/butler/_registry_shim.py
timj: Yes, I was trying to say that we should warn any time it's not `_DefaultSentinel` because that would imply someone actually tried to set it. We should likely raise if they set it to anything other than False.

TallJimbo:
I've pushed three new commits to deal with `ParentDatasetQueryResults`.

PR #894, opened 2023-10-16
DM-41116: Make Butler an abstract class, with DirectButler as default implementation

andy-slac:
## Checklist

- [X] ran Jenkins


python/lsst/daf/butler/_quantum_backed.py
```diff
@@ -659,9 +659,9 @@ class QuantumProvenanceData(_BaseModelCompat):
 
     @staticmethod
     def collect_and_transfer(
-        butler: Butler, quanta: Iterable[Quantum], provenance: Iterable[QuantumProvenanceData]
+        butler: DirectButler, quanta: Iterable[Quantum], provenance: Iterable[QuantumProvenanceData]
```
TallJimbo: Why does this need to be a `DirectButler`?

python/lsst/daf/butler/_registry_shim.py
```diff
@@ -64,7 +64,7 @@ class RegistryShim(Registry):
 
     Parameters
     ----------
-    butler : `Butler`
+    butler : `DirectButler`
```
TallJimbo, line 67: Is the plan to make this back into `Butler` once `_RegistryShim` can delegate to `Butler` methods instead of `DirectButler._registry` methods?

python/lsst/daf/butler/_butler.py
```diff
@@ -0,0 +1,1066 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+from __future__ import annotations
+
```
TallJimbo, line 29: Needs `__all__`.

python/lsst/daf/butler/_butler.py
```diff
@@ -0,0 +1,1066 @@
...
+from ._file_dataset import FileDataset
+from ._limited_butler import LimitedButler
+from ._storage_class import StorageClass
+from .datastore import DatasetRefURIs, Datastore
+from .dimensions import DataId, DimensionConfig
+from .registry import Registry, RegistryConfig, _RegistryFactory
+from .repo_relocation import BUTLER_ROOT_TAG
+from .transfers import RepoExportContext
+
+log = getLogger(__name__)
```
TallJimbo: I'd prefer to stick with our pattern of calling this `_LOG`, or find some other way make this look less like a local variable.

python/lsst/daf/butler/_butler.py
```diff
@@ -0,0 +1,1066 @@
...
+    config : `ButlerConfig`, `Config` or `str`, optional.
+        Configuration. Anything acceptable to the `ButlerConfig` constructor.
+        If a directory path is given the configuration will be read from a
+        ``butler.yaml`` file in that location. If `None` is given default
+        values will be used. If ``config`` contains "cls" key then its value is
+        used as a name of butler class and it must be a sub-class of this
+        class, otherwise `DirectButler` is instantiated.
+    **kwargs : `Any`
+        Optional keyword arguments passed to a constructor of actual butler
+        class.
```
TallJimbo, line 103: I think we should try harder to spell some of these out, maybe all of them, and add those explicit keyword arguments to `from_config` as well.  At least `collections`, `run`, `writeable`, and the default-data-ID arguments should be valid for all butlers and passed to `from_config`, while `searchPaths` should probably be explicit and passed to `_find_butler_class` only.

python/lsst/daf/butler/_butler.py
```diff
@@ -0,0 +1,1066 @@
...
+    -----
+    The preferred way to instantiate Butler is via the `from_config` method.
+    The call to ``Butler(...)`` is equivalent to ``Butler.from_config(...)``,
+    but ``mypy`` will complain about the former.
+    """
+
+    def __new__(cls, config: Config | ResourcePathExpression | None = None, **kwargs: Any) -> Butler:
+        if cls is Butler:
+            cls = cls._find_butler_class(config, **kwargs)
+        return super().__new__(cls)
```
TallJimbo, line 127: It took me a while to figure out how the `**kwargs` get passed to the concrete derived type here (i.e. `__new__` on the base class being followed up by `__init__` on the derived class).

I wish we could disable that behavior and make this more straightforward, e.g. by making this just delegate to `from_config` and have derived-type `__init__` only invoked by that.  But I think that'd require a metaclass and then "more straightforward" is not necessarily the case anymore.

Unless you've got better ideas, maybe just add a code comment about derived-type `__init__` being called automatically here?

TallJimbo (approved):
I've left a few line comments, and I've got one big-picture one: I think we want the `Butler` to eventually hold the `Datastore` and implement many of its methods itself, delegating to a much narrower suite of abstract methods for database access and (eventually) transactions.  We may not be able to do much of that now - a lot of that depends on moving the defaults and caching out of registry.  But since `Butler` doesn't have any state right now - and adding state (especially `Datastore`) might change how we'd want to do construction - I wanted to make sure we were on the same page.

python/lsst/daf/butler/_quantum_backed.py
andy-slac: Sorry, my confusion, I thought I'd need `_ButlerRegistry` here. I'll change it back to `Butler`.

python/lsst/daf/butler/_registry_shim.py
andy-slac: Indeed, once we have everything in `Butler` we can delegate to that.

python/lsst/daf/butler/_butler.py
andy-slac: Yes, I think implementing one in terms of another is more trouble that it's worth, this is why I added `_find_butler_class` method. I'll add a comment before `super().__new__()`.

python/lsst/daf/butler/_butler.py
andy-slac: OK, I added all those parameters to `Butler(...)`  and `Butler.from_config(...)`. I also remembered that I intended to move examples from `DirectButler` docstring to `Butler.from_config`, so that is also on a separate commit with these new parameters.

andy-slac:
> But since Butler doesn't have any state right now - and adding state (especially Datastore) might change how we'd want to do construction - I wanted to make sure we were on the same page.

LimitedButler already has `_datastore` so in a sense it already has state (but it does not initialize it, sub-classes do). And I agree we need to think about deep refactoring and cleaner separation of responsibilities.

PR #898, opened 2023-10-25
DM-41235: Make SqlRegistry a concrete class

andy-slac:
SqlRegistry un-inherits Registry and drops an intermediate _ButlerRegistry
interface. All clients of _ButlerRegistry now use SqlRegistry. Registry is
still an interface for client-facing `Butler.registry`, which is a shim for
SqlRegistry or other butler methods.

## Checklist

- [X] ran Jenkins


tests/test_cliCmdQueryDataIds.py
```diff
@@ -71,12 +72,13 @@ def loadData(self, *filenames: str) -> Butler:
...
         """
         butler = Butler.from_config(self.repo, writeable=True)
+        assert isinstance(butler, DirectButler), "Test expects DirectButler"
         for filename in filenames:
             with open(os.path.join(TESTDIR, "data", "registry", filename)) as stream:
                 # Go behind the back of the import code a bit to deal with
                 # the fact that this is just registry content with no actual
                 # files for the datastore.
-                backend = YamlRepoImportBackend(stream, butler.registry)
+                backend = YamlRepoImportBackend(stream, butler._registry)
```
timj, line 81: This is because `butler.registry` is no longer a registry but `butler._registry` is?

timj (approved):
Looks okay although I am now wondering what the purpose of `Registry` ABC is.

tests/test_cliCmdQueryDataIds.py
andy-slac: Mostly because I changed `YamlRepoImportBackend` to accept `SqlRegistry` instead of `Registry`. I think that `Registry` at this point exists only because we need backward-compatible interface for `Butler.registry`. When we move all functionality to `Butler` we will eventually remove `Registry` completely.

andy-slac:
> Looks okay although I am now wondering what the purpose of Registry ABC is.

This is for clients that still use `Butler.registry`. It does not have to be an ABC, I just kept it with minimal changes. I could merge `Registry` and `RegistryShim` if you prefer.

andy-slac:
> I could merge Registry and RegistryShim if you prefer.

OTOH if we need registry interface for remote butler then its implementation will likely be different from RegistryShim. We probably should keep Registry ABC for now just in case.

PR #899, opened 2023-10-26
DM-41365: Add findDataset/getDataset/getDatasetType to Butler API

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -773,6 +773,32 @@ class Butler(LimitedButler):
         raise NotImplementedError()
 
     @abstractmethod
+    def get_dataset_type(self, name: str) -> DatasetType:
+        """Get the `DatasetType`.
+
+        Parameters
+        ----------
+        name : `str`
+            Name of the type.
+
+        Returns
+        -------
+        type : `DatasetType`
+            The `DatasetType` associated with the given name.
+
+        Raises
+        ------
+        lsst.daf.butler.MissingDatasetTypeError
+            Raised if the requested dataset type has not been registered.
+
+        Notes
+        -----
+        This method handles component dataset types automatically, though most
+        other operations do not.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
     def retrieveArtifacts(
         self,
         refs: Iterable[DatasetRef],
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -800,7 +800,13 @@ class Butler(LimitedButler):
         raise NotImplementedError()
 
     @abstractmethod
-    def get_dataset(self, id: DatasetId, storage_class: str | StorageClass | None) -> DatasetRef | None:
+    def get_dataset(
+        self,
+        id: DatasetId,
+        storage_class: str | StorageClass | None,
+        dimension_records: bool = False,
+        datastore_records: bool = False,
+    ) -> DatasetRef | None:
         """Retrieve a Dataset entry.
 
         Parameters
@@ -810,6 +816,10 @@ class Butler(LimitedButler):
         storage_class : `str` or `StorageClass` or `None`
             A storage class to use when creating the returned entry. If given
             it must be compatible with the default storage class.
+        dimension_records: `bool`, optional
+            If `True` the ref will be expanded and contain dimension records.
+        datastore_records: `bool`, optional.
+            If `True` the ref will contain associated datastore records.
 
         Returns
         -------
@@ -828,6 +838,7 @@ class Butler(LimitedButler):
         collections: str | Sequence[str] | None = None,
         timespan: Timespan | None = None,
         storage_class: str | StorageClass | None = None,
+        dimension_records: bool = False,
         datastore_records: bool = False,
         **kwargs: Any,
     ) -> DatasetRef | None:
@@ -861,6 +872,10 @@ class Butler(LimitedButler):
         storage_class : `str` or `StorageClass` or `None`
             A storage class to use when creating the returned entry. If given
             it must be compatible with the default storage class.
+        dimension_records: `bool`, optional
+            If `True` the ref will be expanded and contain dimension records.
+        datastore_records: `bool`, optional.
+            If `True` the ref will contain associated datastore records.
         **kwargs
             Additional keyword arguments passed to
             `DataCoordinate.standardize` to convert ``dataId`` to a true
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -48,9 +48,10 @@ from ._deferredDatasetHandle import DeferredDatasetHandle
 from ._file_dataset import FileDataset
 from ._limited_butler import LimitedButler
 from ._storage_class import StorageClass
+from ._timespan import Timespan
 from .datastore import DatasetRefURIs, Datastore
 from .dimensions import DataId, DimensionConfig
-from .registry import Registry, RegistryConfig, _RegistryFactory
+from .registry import CollectionArgType, Registry, RegistryConfig, _RegistryFactory
 from .repo_relocation import BUTLER_ROOT_TAG
 from .transfers import RepoExportContext
 
@@ -795,6 +796,83 @@ class Butler(LimitedButler):
         -----
         This method handles component dataset types automatically, though most
         other operations do not.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def find_dataset(
+        self,
+        datasetType: DatasetType | str,
+        dataId: DataId | None = None,
+        *,
+        collections: CollectionArgType | None = None,
+        timespan: Timespan | None = None,
+        datastore_records: bool = False,
+        **kwargs: Any,
+    ) -> DatasetRef | None:
+        """Find a dataset given its `DatasetType` and data ID.
+
+        This can be used to obtain a `DatasetRef` that permits the dataset to
+        be read from a `Datastore`. If the dataset is a component and can not
+        be found using the provided dataset type, a dataset ref for the parent
+        will be returned instead but with the correct dataset type.
+
+        Parameters
+        ----------
+        datasetType : `DatasetType` or `str`
+            A `DatasetType` or the name of one.  If this is a `DatasetType`
+            instance, its storage class will be respected and propagated to
+            the output, even if it differs from the dataset type definition
+            in the registry, as long as the storage classes are convertible.
+        dataId : `dict` or `DataCoordinate`, optional
+            A `dict`-like object containing the `Dimension` links that identify
+            the dataset within a collection.
+        collections : collection expression, optional
+            An expression that fully or partially identifies the collections to
+            search for the dataset; see
+            :ref:`daf_butler_collection_expressions` for more information.
+            Defaults to ``self.defaults.collections``.
+        timespan : `Timespan`, optional
+            A timespan that the validity range of the dataset must overlap.
+            If not provided, any `~CollectionType.CALIBRATION` collections
+            matched by the ``collections`` argument will not be searched.
+        **kwargs
+            Additional keyword arguments passed to
+            `DataCoordinate.standardize` to convert ``dataId`` to a true
+            `DataCoordinate` or augment an existing one.
+
+        Returns
+        -------
+        ref : `DatasetRef`
+            A reference to the dataset, or `None` if no matching Dataset
+            was found.
+
+        Raises
+        ------
+        lsst.daf.butler.NoDefaultCollectionError
+            Raised if ``collections`` is `None` and
+            ``self.collections`` is `None`.
+        LookupError
+            Raised if one or more data ID keys are missing.
+        lsst.daf.butler.registry.MissingDatasetTypeError
+            Raised if the dataset type does not exist.
+        lsst.daf.butler.MissingCollectionError
+            Raised if any of ``collections`` does not exist in the registry.
+
+        Notes
+        -----
+        This method simply returns `None` and does not raise an exception even
+        when the set of collections searched is intrinsically incompatible with
+        the dataset type, e.g. if ``datasetType.isCalibration() is False``, but
+        only `~CollectionType.CALIBRATION` collections are being searched.
+        This may make it harder to debug some lookup failures, but the behavior
+        is intentional; we consider it more important that failed searches are
+        reported consistently, regardless of the reason, and that adding
+        additional collections that do not contain a match to the search path
+        never changes the behavior.
+
+        This method handles component dataset types automatically, though most
+        other registry operations do not.
         """
         raise NotImplementedError()
 
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -42,7 +42,7 @@ from ._butler_config import ButlerConfig
 from ._butler_repo_index import ButlerRepoIndex
 from ._config import Config, ConfigSubset
 from ._dataset_existence import DatasetExistence
-from ._dataset_ref import DatasetIdGenEnum, DatasetRef
+from ._dataset_ref import DatasetId, DatasetIdGenEnum, DatasetRef
 from ._dataset_type import DatasetType
 from ._deferredDatasetHandle import DeferredDatasetHandle
 from ._file_dataset import FileDataset
@@ -796,6 +796,23 @@ class Butler(LimitedButler):
         -----
         This method handles component dataset types automatically, though most
         other operations do not.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def get_dataset(self, id: DatasetId) -> DatasetRef | None:
+        """Retrieve a Dataset entry.
+
+        Parameters
+        ----------
+        id : `DatasetId`
+            The unique identifier for the dataset.
+
+        Returns
+        -------
+        ref : `DatasetRef` or `None`
+            A ref to the Dataset, or `None` if no matching Dataset
+            was found.
         """
         raise NotImplementedError()
 
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -51,7 +51,7 @@ from ._storage_class import StorageClass
 from ._timespan import Timespan
 from .datastore import DatasetRefURIs, Datastore
 from .dimensions import DataId, DimensionConfig
-from .registry import CollectionArgType, Registry, RegistryConfig, _RegistryFactory
+from .registry import Registry, RegistryConfig, _RegistryFactory
 from .repo_relocation import BUTLER_ROOT_TAG
 from .transfers import RepoExportContext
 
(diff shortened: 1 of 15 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -772,6 +773,126 @@ def getURI(
...
+        ref : `DatasetRef` or `None`
+            A ref to the Dataset, or `None` if no matching Dataset
+            was found.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def find_dataset(
+        self,
+        datasetType: DatasetType | str,
```
timj: @TallJimbo I assume we aren't going to start saying `dataset_type` and `data_id` in the public APIs?

tests/test_server.py
```diff
@@ -77,6 +82,10 @@ def create_factory_dependency():
         cls.client = TestClient(app)
         cls.butler = _make_remote_butler(cls.client)
 
+        # Populate the test server.
+        server_butler.import_(filename=os.path.join(TESTDIR, "data", "registry", "base.yaml"))
+        server_butler.import_(filename=os.path.join(TESTDIR, "data", "registry", "datasets-uuid.yaml"))
+
```
dhirving, line 90: Doesn't matter yet but you may want to move this above the call to _make_remote_butler, in case we decide to read some stuff from the server during RemoteButler initialization.

timj: I'm hoping that it doesn't make a difference here either way because the clients won't be downloading caches of dataset type and collection before this runs will they?

python/lsst/daf/butler/_butler.py
TallJimbo: I figured we _would_ start using snake_case for those, too, when they're new arguments or new methods.  I've already started doing that in various places.

python/lsst/daf/butler/_butler.py
```diff
@@ -798,6 +799,83 @@ def get_dataset_type(self, name: str) -> DatasetType:
...
+        dataId: DataId | None = None,
+        *,
+        collections: CollectionArgType | None = None,
+        timespan: Timespan | None = None,
+        datastore_records: bool = False,
+        **kwargs: Any,
+    ) -> DatasetRef | None:
+        """Find a dataset given its `DatasetType` and data ID.
```
TallJimbo, line 845: With this method moving to `Butler`, I think it'd make sense for it to become an entry point for the logic that handles `day_obs` and `seq_num` (etc) as data ID keys.  I think it has everything it needs.

python/lsst/daf/butler/_butler.py
```diff
@@ -798,6 +799,83 @@ def get_dataset_type(self, name: str) -> DatasetType:
...
 
+    @abstractmethod
+    def find_dataset(
+        self,
+        datasetType: DatasetType | str,
+        dataId: DataId | None = None,
+        *,
+        collections: CollectionArgType | None = None,
+        timespan: Timespan | None = None,
+        datastore_records: bool = False,
```
TallJimbo, line 842: If we provide this as an option, we should also provide a way to expand the data ID (i.e. add dimension records) at the same time, and we should think about whether we want just one option to do both or allow the two kinds of expansion to happen separately.

python/lsst/daf/butler/_butler.py
```diff
@@ -798,6 +799,83 @@ def get_dataset_type(self, name: str) -> DatasetType:
         """
         raise NotImplementedError()
 
+    @abstractmethod
+    def find_dataset(
+        self,
+        datasetType: DatasetType | str,
+        dataId: DataId | None = None,
+        *,
+        collections: CollectionArgType | None = None,
```
TallJimbo: This needs to be an ordered search path, so some of the types in this alias like `re.Pattern` and `EllipsisType` aren't actually valid here.

python/lsst/daf/butler/_butler.py
```diff
@@ -798,6 +799,83 @@ def get_dataset_type(self, name: str) -> DatasetType:
...
+        timespan: Timespan | None = None,
+        datastore_records: bool = False,
+        **kwargs: Any,
```
TallJimbo, line 843: Should we consider making this accept `storage_class`, too, so storage-class-overrides in `get` (etc) can be delegated here and then passed with the `ref` instead of separately?

python/lsst/daf/butler/_butler.py
```diff
@@ -798,6 +799,83 @@ def get_dataset_type(self, name: str) -> DatasetType:
...
+            was found.
+
+        Raises
+        ------
+        lsst.daf.butler.NoDefaultCollectionError
+            Raised if ``collections`` is `None` and
+            ``self.collections`` is `None`.
+        LookupError
+            Raised if one or more data ID keys are missing.
+        lsst.daf.butler.registry.MissingDatasetTypeError
```
TallJimbo: We need to move this exception out of `registry` now.

python/lsst/daf/butler/_butler.py
```diff
@@ -799,6 +799,23 @@ def get_dataset_type(self, name: str) -> DatasetType:
         """
         raise NotImplementedError()
 
+    @abstractmethod
+    def get_dataset(self, id: DatasetId) -> DatasetRef | None:
```
TallJimbo: I think we'll want a vectorized version of this in the future, but it's a pain to implement now.  I don't think this interface needs to change in order to live alongside a vectorized one, but I wanted to raise it in case you think otherwise.

TallJimbo (approved):
Just took a look at the APIs, and while I left a lot of comments they're all minor.  Did not look at the implementations (I gather that's not needed from me).

python/lsst/daf/butler/_butler.py
timj: I had in fact done this (for now I'm lifting it out) but I forgot to update the docs.

python/lsst/daf/butler/_butler.py
timj: There are two open questions:

* If we have vectorized version, `get_datasets` do we keep `get_dataset` around?
* What should `get_datasets` return? `list[DatasetRef]` or `DatasetRefContainer`?

I think we likely want a single request to return a single ref so it seems like we always want `Butler.get_dataset` even if we later add `Butler.get_datasets`.

python/lsst/daf/butler/_butler.py
timj: I had completely failed to notice that that flag was in there now. One option is for `find_dataset` to always expand and add everything. People shouldn't be calling it in bulk.

python/lsst/daf/butler/_butler.py
timj: Okay. This can in theory call `Butler._rewrite_data_id` now that it's moved up to Butler.

python/lsst/daf/butler/_butler.py
timj: So really it's `str | Iterable[str] | CollectionSearch`  and the original annotation isn't quite right.

python/lsst/daf/butler/_butler.py
timj: Can give that a try.

python/lsst/daf/butler/_butler.py
TallJimbo: Yes, and I think `CollectionSearch` is deprecated (and ready to be removed).

python/lsst/daf/butler/_butler.py
```diff
@@ -772,6 +773,126 @@ def getURI(
...
+
+        Raises
+        ------
+        lsst.daf.butler.MissingDatasetTypeError
+            Raised if the requested dataset type has not been registered.
+
+        Notes
+        -----
+        This method handles component dataset types automatically, though most
+        other registry operations do not.
```
andy-slac: Drop `registry` here or replace with `Butler`?

python/lsst/daf/butler/_butler.py
andy-slac: Hmm, do we ever expose bare `DatasetId` to butler clients (do we need this method at butler level)?

python/lsst/daf/butler/_butler.py
andy-slac: Always expanding could be inefficient, I think many clients do not care about datastore records at all, we should not be pessimizing things for them.

python/lsst/daf/butler/_butler.py
```diff
@@ -772,6 +773,126 @@ def getURI(
...
+            :ref:`daf_butler_collection_expressions` for more information.
+            Defaults to ``self.defaults.collections``.
+        timespan : `Timespan`, optional
+            A timespan that the validity range of the dataset must overlap.
+            If not provided, any `~CollectionType.CALIBRATION` collections
+            matched by the ``collections`` argument will not be searched.
+        **kwargs
+            Additional keyword arguments passed to
+            `DataCoordinate.standardize` to convert ``dataId`` to a true
+            `DataCoordinate` or augment an existing one.
```
andy-slac: `datastore_records` needs to be added to docstring, somehow I forgot to add it to the registry method when introduced it first.

python/lsst/daf/butler/_butler.py
```diff
@@ -772,6 +773,126 @@ def getURI(
...
+        the dataset type, e.g. if ``datasetType.isCalibration() is False``, but
+        only `~CollectionType.CALIBRATION` collections are being searched.
+        This may make it harder to debug some lookup failures, but the behavior
+        is intentional; we consider it more important that failed searches are
+        reported consistently, regardless of the reason, and that adding
+        additional collections that do not contain a match to the search path
+        never changes the behavior.
+
+        This method handles component dataset types automatically, though most
+        other registry operations do not.
```
andy-slac: "registry" here too.

python/lsst/daf/butler/_butler.py
```diff
@@ -42,15 +42,16 @@
...
 from ._dataset_type import DatasetType
 from ._deferredDatasetHandle import DeferredDatasetHandle
 from ._file_dataset import FileDataset
 from ._limited_butler import LimitedButler
 from ._storage_class import StorageClass
+from ._timespan import Timespan
 from .datastore import DatasetRefURIs, Datastore
 from .dimensions import DataId, DimensionConfig
-from .registry import Registry, RegistryConfig, _RegistryFactory
+from .registry import CollectionArgType, Registry, RegistryConfig, _RegistryFactory
```
andy-slac: We should probably move `CollectionArgType` out of `registry`. I think ideally `Butler` should not depend on anything from `registry` (we keep `Butler.makeRepo()` so we need three other classes here.)

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -179,6 +211,57 @@ def getURI(
         # Docstring inherited.
         raise NotImplementedError()
 
+    def get_dataset_type(self, name: str) -> DatasetType:
+        # In future implementation this should directly access the cache
+        # and only go to the server if the dataset type is not known.
+        path = f"dataset_type/{name}"
+        response = self._client.get(self._get_url(path))
+        response.raise_for_status()
```
andy-slac, line 223: Will this raise some HttpError instead of `MissingDatasetTypeError` when dataset type is not know?

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -179,6 +211,57 @@ def getURI(
...
+        return DatasetType.from_simple(SerializedDatasetType(**response.json()), universe=self.dimensions)
+
+    def get_dataset(self, id: DatasetId) -> DatasetRef | None:
+        path = f"dataset/{id}"
+        response = self._client.get(self._get_url(path))
+        if response.json() is None:
```
andy-slac, line 248: Does this need some error handling before you try to covert response to JSON (`raise_for_status()`?)

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -179,6 +211,57 @@ def getURI(
...
+    ) -> DatasetRef | None:
+        if collections is None:
+            if not self.collections:
+                raise NoDefaultCollectionError(
+                    "No collections provided to find_dataset, and no defaults from butler construction."
+                )
+            collections = self.collections
+        # Temporary hack. Assume strings for collections. In future
+        # want to construct CollectionWildcard and filter it through collection
+        # cache to generate list of collection names.
```
andy-slac, line 272: Working on caching ticket now, I would probably prefer to not have a complete list of collections on every client in the long term.

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -179,6 +211,57 @@ def getURI(
...
+        wildcards = CollectionWildcard.from_expression(collections)
```
andy-slac, line 273: Maybe do a quick check here and raise if `patterns` specifies something?

python/lsst/daf/butler/remote_butler/server/_server.py
```diff
@@ -56,8 +64,88 @@ def factory_dependency() -> Factory:
...
+    response_model_exclude_unset=True,
+    response_model_exclude_defaults=True,
+    response_model_exclude_none=True,
+)
+def get_dataset_type(
+    dataset_type_name: str, factory: Factory = Depends(factory_dependency)
+) -> SerializedDatasetType:
+    """Return the dataset type."""
+    butler = factory.create_butler()
+    datasetType = butler.get_dataset_type(dataset_type_name)
```
andy-slac, line 120: Is exception transported transparently back to client?

tests/test_server.py
```diff
@@ -90,6 +99,38 @@ def test_simple(self):
     def test_remote_butler(self):
         universe = self.butler.dimensions
         self.assertEqual(universe.namespace, "daf_butler")
+        self.assertFalse(self.butler.isWriteable())
+
+    def test_get_dataset_type(self):
+        bias_type = self.butler.get_dataset_type("bias")
+        self.assertEqual(bias_type.name, "bias")
```
andy-slac, line 108: Add a check for non-existent one?

andy-slac (approved):

python/lsst/daf/butler/remote_butler/_remote_butler.py
timj: Yes. Oops.

python/lsst/daf/butler/remote_butler/_remote_butler.py
timj: Interesting. I thought we said we would last week. I thought @TallJimbo wanted the client to know all possible collections when it was building the query object before sending it to the server. The caching story between client and server is a bit tricky.

python/lsst/daf/butler/remote_butler/_remote_butler.py
timj: Seems like I have to change the API to something a bit simpler for collection list anyhow.

python/lsst/daf/butler/_butler.py
timj: It certainly makes client/server testing a lot easier to begin with since it's the simplest possible API to implement. We do have VO services that are given a UUID (via datalink from, eg, ObsTAP) and need to convert that to a dataset ref so it's not impossible that the server will need to do that (depending on how services connect to butlers).

TallJimbo: I think we'll want to start exposing dataset IDs in more places, too; I've been thinking about how we might to get them into the `--where` expression language, for instance. They appear pretty prominently in QGs already.

python/lsst/daf/butler/remote_butler/_remote_butler.py
andy-slac: Yes, I know, but we may need to rethink that part, otherwise we may have scaling issue. I looked at Butler initialization time today, the simplest `butler query-data-ids --collections=HSC/raw/all /repo/main exposure` takes 10 seconds, large fraction of that comes from pre-loading of collections cache. We have 300k collections today, and it will only grow with time. There are ways to reduce that probably, but in my ideal world it would be better if we cache almost nothing on client side. 🙂

TallJimbo: I think having the caches in the client has worked well for us so far (i.e. with the server just being the DB), and I like the idea of being able to optimize server calls away entirely in the client sometimes.  And I do think caching on the server is a lot harder, since we can't just punt cache misses to the user there.

But I do hesitate a bit because I'm not sure if "cache all the collections the user can see" continues to be in the "small enough to be irrelevant" category after DR1, both in terms of downloading the cache and holding it in memory.  And I have no idea if my "do more on the client to keep the server load light" intuition is better than @dhirving 's "do nothing on the client to keep version management easier" intuition.

andy-slac: Downloading a million collection records just to figure out that pattern does not match any of them looks like a lot of overhead. It is of course about the right balance between what we do on server vs client, we need to find something that works and scales.

TallJimbo: Ok, I'm sold on not trying to aggressively fetch all collections the user can access.  It's the dataset type definitions, chained collection members, and collection summaries that are needed for query contruction, and we don't need all of any of those.

I think what may break if we drop the complete collection-name cache is our regex/glob collection query implementation - we'll need to start translating/limiting those patterns to something that can be run inside the DB.  Yes, we could do a full cache in the `RemoteButler` server, but this scaling problem will hit `DirectButler`, too.

I'll do a brain dump on Slack later today about the places we use the caches in the query system so somebody other than me can do some conceptual design work on what it should look like.  I think we need to have a better sense of where this is all going, even if we don't decide to switch fully from the current approach yet, but I want to stay off of at least this part of the critical path.

python/lsst/daf/butler/_butler.py
timj: I replaced it with an explicit `str | Sequence[str]` definition.

python/lsst/daf/butler/_butler.py
timj: I've done this and it works fine. It doesn't save us anything in `butler.get()` though because that does the day_obs/seq_num look up early so it can construct a proper DataId, and potentially expand it, before calling find_dataset.

python/lsst/daf/butler/_butler.py
timj: Presumably get_dataset should have the datastore_records and dimension_records booleans as well for consistency?

TallJimbo: :+1:

timj: I also need to add datastore records to the SerializedDatasetRef because that was never updated and so it's impossible to return datastore records from the server at the moment...

andy-slac: Jim asked me to not add them to `SerializedDatasetRef`, as it may increase graph size on disk (but I don't think we give records to refs that go into the graph, at least not yet).

timj: We are sending SerializedDatasetRef over the wire so we can only reconstruct things in the far end that were put in at the other end. We need to add them somehow (but obviously they should be optional). For now I've turned the feature off in client/server so you can't get the datastore records.

python/lsst/daf/butler/remote_butler/_remote_butler.py
timj: Turns out FastAPI exception handling is an adventure (you have to raise an HttpException and fill it with all the real exception information and then re-raise in the client). I think I'm going to punt on this for now.

PR #900, opened 2023-10-27
DM-41366: Add Docker image build for butler server

dhirving:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,99 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This software is dual licensed under the GNU General Public License and also
+# under a 3-clause BSD license. Recipients may choose which of these licenses
+# to use; please see the files gpl-3.0.txt and/or bsd_license.txt,
+# respectively.  If you choose the GPL option then the following text applies
+# (but note that there is still no warranty even if you opt for BSD instead):
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -0,0 +1,99 @@
...
+RUN pip install --quiet --no-cache-dir -r requirements.txt -r server_requirements.txt
+
+FROM dependencies-image AS install-image
+
+# Use the virtualenv
+ENV PATH="/opt/venv/bin:$PATH"
+
+COPY . /workdir
+WORKDIR /workdir
+RUN pip install --no-cache-dir .
```
timj: Maybe add `--no-deps` here so it doesn't try again to install all the dependencies that we've installed via `requirements.txt` above? Historically we have had some weird things going on with version number matches if requirements.txt is pointing to git repos and pip is using `pyproject.toml` dependencies from PyPI.

compose.yml
```diff
@@ -0,0 +1,13 @@
...
+# test data from the ci_hsc_gen3 package.  Assumes that ci_hsc_gen3
+# has been built via lsstsw in a directory adjacent to the directory
+# containing this file.
+services:
+  butler:
+    build: .
+    ports:
+      - "8080:8080"
+    environment:
+      BUTLER_SERVER_BUTLER_ROOT: "/butler_root"
```
timj: I am pondering whether BUTLER_ROOT is the right term now. This is really the URI to the butler configuration (with or without the `butler.yaml`).

server/requirements.txt
```diff
@@ -0,0 +1,2 @@
+uvicorn
+psycopg2
```
timj, line 2: In theory `pip install .[postgres]` should pick up this dependency (it's listed as an optional dep in the pyproject.toml), but not if I've just told you to use `--no-deps` for that part above...

timj (approved):
Looks good. It would be nice if the github action could run a simple test to check that the code can be loaded. eg by running something like:

```bash
$ butler create tmp
$ butler config-dump tmp
```

server/requirements.txt
dhirving: The dependencies actually want to be installed in a separate step from the pip install of the application itself -- if the requirements.txt files don't change, Docker can cache the intermediate dependency image and save a few minutes on the next build.

Also the requirements will need to be pinned in future so we won't want pyproject.toml involved.

compose.yml
dhirving: renamed BUTLER_SERVER_CONFIG_URI

PR #902, opened 2023-11-03
DM-41562: Rewrite much of DatasetRef.from_simple to fix cache problem.

TallJimbo:
The previous caching did not respect storage class overrides; all DatasetRefs with the same UUID and component would end up using the first storage class deserialized.

The fix here is to always call overrideStorageClass when using the cache, rather than add the storage class to the cache key.  That's because we expect the value of the cache to mostly be in avoiding reconstruction of the data ID, and overrideStorageClass doesn't touch that.

By the same token, the dataset type name has been removed from the cache key as well, with only refs for parent dataset types cached. This should shrink the cache a bit and improve performance in the usual (no-component) case.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-41562.bugfix.md
+++ b/doc/changes/DM-41562.bugfix.md
@@ -1,0 +1,3 @@
+Fix caching in DatasetRef deserialization that caused the serialized storage class to be ignored.
+
+This caused intermittent failures when running pipelines that use multiple storage classes for the same dataset type.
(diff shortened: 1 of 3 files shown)


natelust (approved):
There are a few places where I think there are various branches that are essentially the same, and I think could be smithed to be a less complicated branch structure, but I am not convinced that it is worth the time investment to do so.

I still feel there is subtle inconsistency here I can't quite put my finger on, but it could just be the mess that is transforming types giving me the impression, and it certainly seems better than what was here

TallJimbo:
My hope is that someday we'll be able to get rid of almost all of the `Serialized*` types anyway, in favor of normalized, directly-serializable collections of things.  We'll be adding those for client server, and then it'll be a matter of getting `Quantum` and `QuantumGraph` to stop using them.

PR #903, opened 2023-11-06
DM-34340: implement RFC-834 deprecations

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-34340.api.md
+++ b/doc/changes/DM-34340.api.md
@@ -1,0 +1,19 @@
+Deprecate most public APIs that use Dimension or DimensionElement objects.
+
+This implements [RFC-834](https://jira.lsstcorp.org/browse/RFC-834), deprecating the `DimensionGraph` class (in favor of the new, similar `DimensionGroup`) and a large number of `DataCoordinate` methods and attributes, including its `collections.abc.Mapping` interface.
+
+This includes:
+
+- use `DataCoordinate.dimensions` instead of `DataCoordinate.graph` (likewise for arguments to `DataCoordinate.standardize`);
+- use `dict(DataCoordinate.required)` as a drop-in replacement for `DataCoordinate.byName()`, but consider whether you want `DataCoordinate.required` (a `Mapping` view rather than a `dict`) or `DataCoordinate.mapping` (a `Mapping` with all *available* key-value pairs, not just the required ones);
+- also use `DataCoordinate.mapping` or `DataCoordinate.required` instead of treating `DataCoordinate` itself as a `Mapping`, *except* square-bracket indexing, which is still very much supported;
+- use `DataCoordinate.dimensions.required.names` or `DataCoordinate.required.keys()` as a drop-in replacement for `DataCoordinate.keys().names` or `DataCoordinate.names`, but consider whether you actually want `DataCoordinate.dimensions.names` or `DataCoordinate.mapping.keys` instead.
+
+`DimensionGroup` is almost identical to `DimensionGraph`, but it and its subset attributes are not directly iterable (since those iterate over `Dimension` and `DimensionElement` objects); use the `.names` attribute to iterate over names instead (just as names could be iterated over in `DimensionGraph`).
+
+`DimensionGraph` is still used in some `lsst.daf.butler` APIs (most prominently `DatasetType.dimensions`) that may be accessed without deprecation warnings being emitted, but iterating over that object or its subset attributes *will* yield deprecation warnings.
+And `DimensionGraph` is still accepted along with `DimensionGroup` without warning in most public APIs.
+When `DimensionGraph` is removed, methods and properties that return `DimensionGraph` will start returning `DimensionGroup` instead.
+
+Rare code (mostly in downstream middleware packages) that does need access to `Dimension` or `DimensionElement` objects should obtain them directly from the `DimensionUniverse`.
+For the pattern of checking whether a dimension is a skypix level, test whether its name is in `DimensionUniverse.skypix_dimensions` or `DimensionGroup.skypix` instead of obtaining a `Dimension` instance and calling `isinstance(dimension, SkyPixDimension)`.
--- a/python/lsst/daf/butler/_dataset_type.py
+++ b/python/lsst/daf/butler/_dataset_type.py
@@ -40,7 +40,7 @@ from pydantic import StrictBool, StrictStr
 
 from ._config_support import LookupKey
 from ._storage_class import StorageClass, StorageClassFactory
-from .dimensions import DimensionGraph, SerializedDimensionGraph
+from .dimensions import DimensionGraph, DimensionGroup, SerializedDimensionGraph
 from .json import from_json_pydantic, to_json_pydantic
 from .persistence_context import PersistenceContextVars
 
@@ -60,7 +60,7 @@ class SerializedDatasetType(_BaseModelCompat):
 
     name: StrictStr
     storageClass: StrictStr | None = None
-    dimensions: SerializedDimensionGraph | None = None
+    dimensions: SerializedDimensionGraph | list[StrictStr] | None = None
     parentStorageClass: StrictStr | None = None
     isCalibration: StrictBool = False
 
@@ -70,15 +70,16 @@ class SerializedDatasetType(_BaseModelCompat):
         *,
         name: str,
         storageClass: str | None = None,
-        dimensions: dict | None = None,
+        dimensions: list | dict | None = None,
         parentStorageClass: str | None = None,
         isCalibration: bool = False,
     ) -> SerializedDatasetType:
         """Construct a `SerializedDatasetType` directly without validators.
 
-        This differs from PyDantics construct method in that the arguments are
-        explicitly what the model requires, and it will recurse through
-        members, constructing them from their corresponding `direct` methods.
+        This differs from Pydantic's model_construct method in that the
+        arguments are explicitly what the model requires, and it will recurse
+        through members, constructing them from their corresponding `direct`
+        methods.
 
         This method should only be called when the inputs are trusted.
         """
@@ -87,9 +88,14 @@ class SerializedDatasetType(_BaseModelCompat):
         if cache is not None and (type_ := cache.get(key, None)) is not None:
             return type_
 
-        serialized_dimensions = (
-            SerializedDimensionGraph.direct(**dimensions) if dimensions is not None else None
-        )
+        serialized_dimensions: list[str] | None
+        match dimensions:
+            case list():
+                serialized_dimensions = dimensions
+            case dict():
+                serialized_dimensions = SerializedDimensionGraph.direct(**dimensions).names
+            case None:
+                serialized_dimensions = None
 
         node = cls.model_construct(
             name=name,
@@ -125,9 +131,11 @@ class DatasetType:
         and underscores.  Component dataset types should contain a single
         period separating the base dataset type name from the component name
         (and may be recursive).
-    dimensions : `DimensionGraph` or iterable of `Dimension` or `str`
+    dimensions : `DimensionGroup`, `DimensionGraph`, or \
+            `~collections.abc.Iterable` [ `Dimension` or `str` ]
         Dimensions used to label and relate instances of this `DatasetType`.
-        If not a `DimensionGraph`, ``universe`` must be provided as well.
+        If not a `DimensionGraph` or `DimensionGroup`, ``universe`` must be
+        provided as well.
     storageClass : `StorageClass` or `str`
         Instance of a `StorageClass` or name of `StorageClass` that defines
         how this `DatasetType` is persisted.
@@ -184,7 +192,7 @@ class DatasetType:
     def __init__(
         self,
         name: str,
-        dimensions: DimensionGraph | Iterable[Dimension | str],
+        dimensions: DimensionGroup | DimensionGraph | Iterable[Dimension | str],
         storageClass: StorageClass | str,
         parentStorageClass: StorageClass | str | None = None,
         *,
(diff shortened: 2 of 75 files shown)


doc/changes/DM-34340.api.md
```diff
@@ -0,0 +1,19 @@
...
+- use `dict(DataCoordinate.required)` as a drop-in replacement for `DataCoordinate.byName()`, but consider whether you want `DataCoordinate.required` (a `Mapping` view rather than a `dict`) or `DataCoordinate.mapping` (a `Mapping` with all *available* key-value pairs, not just the required ones);
+- also use `DataCoordinate.mapping` or `DataCoordinate.required` instead of treating `DataCoordinate` itself as a `Mapping`, *except* square-bracket indexing, which is still very much supported;
+- use `DataCoordinate.dimensions.required.names` or `DataCoordinate.required.keys()` as a drop-in replacement for `DataCoordinate.keys().names` or `DataCoordinate.names`, but consider whether you actually want `DataCoordinate.dimensions.names` or `DataCoordinate.mapping.keys` instead.
+
+`DimensionGroup` is almost identical to `DimensionGraph`, but it and its subset attributes are not directly iterable (since those iterate over `Dimension` and `DimensionElement` objects); use the `.names` attributes of to iterate over names instead (just as names could be iterated over in `DimensionGraph`).
```
timj:
```suggestion
`DimensionGroup` is almost identical to `DimensionGraph`, but it and its subset attributes are not directly iterable (since those iterate over `Dimension` and `DimensionElement` objects); use the `.names` attribute to iterate over names instead (just as names could be iterated over in `DimensionGraph`).
```

doc/changes/DM-34340.api.md
```diff
@@ -0,0 +1,19 @@
+Deprecate most public APIs that use Dimension or DimensionElement objects.
+
+This implements RFC-834, deprecating the `DimensionGraph` class (in favor of the new, similar `DimensionGroup`) and a large number of `DataCoordinate` methods and attributes, including its `collections.abc.Mapping` interface.
```
timj:
```suggestion
This implements [RFC-834](https://jira.lsstcorp.org/browse/RFC-834), deprecating the `DimensionGraph` class (in favor of the new, similar `DimensionGroup`) and a large number of `DataCoordinate` methods and attributes, including its `collections.abc.Mapping` interface.
```

python/lsst/daf/butler/_dataset_type.py
```diff
@@ -355,7 +362,7 @@ def dimensions(self) -> DimensionGraph:
         The dimensions label and relate instances of this
         `DatasetType` (`DimensionGraph`).
         """
-        return self._dimensions
+        return self._dimensions._as_graph()
```
timj, line 364: This property will change to DimensionGroup at some point? Also typo in doc string.

python/lsst/daf/butler/_dataset_type.py
```diff
@@ -661,7 +668,7 @@ def to_simple(self, minimal: bool = False) -> SerializedDatasetType:
                 "name": self.name,
                 "storageClass": self._storageClassName,
                 "isCalibration": self._isCalibration,
-                "dimensions": self.dimensions.to_simple(),
+                "dimensions": list(self.dimensions.names),
```
timj: Not `self._dimensions` ?

python/lsst/daf/butler/dimensions/_coordinate.py
```diff
@@ -119,45 +120,25 @@ def _intersectRegions(*args: Region) -> Region | None:
...
 
-    `DataCoordinate` itself is an ABC, but provides `staticmethod` factory
+    Notes
+    -----
+    `DataCoordinate` is an ABC, but it provides `staticmethod` factory
     functions for private concrete implementations that should be sufficient
     for most purposes.  `standardize` is the most flexible and safe of these;
-    the others (`makeEmpty`, `fromRequiredValues`, and `fromFullValues`) are
-    more specialized and perform little or no checking of inputs.
+    the others (`makeEmpty`, `from_required_values`, and `from_full_values`)
```
timj: Now the `makeEmpty` looks incongruous in that it's not `make_empty`...

python/lsst/daf/butler/dimensions/_coordinate.py
```diff
@@ -119,45 +120,25 @@ def _intersectRegions(*args: Region) -> Region | None:
...
-       dimensions represent a serious problem with the data that
-       `DataCoordinate` cannot generally recognize on its own, and a data ID
-       that knows implied dimension values should still be able to compare as
-       equal to one that does not.  This is of course not the way comparisons
-       between simple `dict` data IDs work, and hence using a `DataCoordinate`
-       instance for at least one operand in any data ID comparison is strongly
-       recommended.
+    Lookups for implied dimensions (those in ``self.dimensions.implied``) are
+    supported if and only if `has_full_values` is `True`.  This also sets the
+    keys of the `mapping` attribute.  This means that DataCoordinate equality
```
timj:
```suggestion
    keys of the `mapping` attribute.  This means that `DataCoordinate` equality
```

python/lsst/daf/butler/dimensions/_coordinate.py
```diff
@@ -310,15 +364,53 @@ def fromRequiredValues(graph: DimensionGraph, values: tuple[DataIdValue, ...]) -
...
+        values : `tuple` [ `int` or `str` ]
+            Tuple of primary key values corresponding to ``graph.required``,
+            in that order.
+
+        Returns
+        -------
+        dataId : `DataCoordinate`
+            A data ID object that identifies the given dimensions.
+            ``dataId.hasFull()`` will return `True` if and only if
+            ``graph.implied`` is empty, and ``dataId.hasRecords()`` will never
```
timj: I had a bit of trouble parsing this the first time. I think it would be clearer if there were two distinct sentences -- one about `hasFull` and one about `hasRecords`.

python/lsst/daf/butler/dimensions/_coordinate.py
```diff
@@ -340,52 +432,101 @@ def fromFullValues(graph: DimensionGraph, values: tuple[DataIdValue, ...]) -> Da
...
+            Tuple of primary key values corresponding to
+            ``itertools.chain(graph.required, graph.implied)``, in that order.
+            Note that this is _not_ the same order as ``graph.dimensions``,
+            though these contain the same elements.
+
+        Returns
+        -------
+        dataId : `DataCoordinate`
+            A data ID object that identifies the given dimensions.
+            ``dataId.hasFull()`` will return `True` if and only if
```
timj: Doesn't this always have `hasFull` return `True`?

python/lsst/daf/butler/dimensions/_coordinate.py
```diff
@@ -340,52 +432,101 @@ def fromFullValues(graph: DimensionGraph, values: tuple[DataIdValue, ...]) -> Da
...
+        # printing out the whole DimensionUniverse.
+        return str(self.mapping)
 
     def __lt__(self, other: Any) -> bool:
-        # Allow DataCoordinate to be sorted
-        if not isinstance(other, type(self)):
+        if not isinstance(other, DataCoordinate):
             return NotImplemented
-        # Form tuple of tuples for each DataCoordinate:
-        # Unlike repr() we only use required keys here to ensure that
```
timj, line 372: This comment is still helpful isn't it?

python/lsst/daf/butler/dimensions/_coordinate.py
```diff
@@ -533,15 +702,22 @@ def full(self) -> NamedKeyMapping[Dimension, DataIdValue]:
         """
         raise NotImplementedError()
 
-    @abstractmethod
+    # TODO: remove on DM-41326.
+    @deprecated(
+        "DataCoordinate.values_tuple() is deprecated in favor of .require_values, and will be dropped "
```
timj:
```suggestion
        "DataCoordinate.values_tuple() is deprecated in favor of .required_values, and will be dropped "
```

python/lsst/daf/butler/dimensions/_coordinate.py
```diff
@@ -742,7 +920,7 @@ def from_simple(
         simple : `dict` of [`str`, `Any`]
             The `dict` returned by `to_simple()`.
         universe : `DimensionUniverse`
-            The special graph of all known dimensions.
+            Object that manages of all known dimensions.
```
timj:
```suggestion
            Object that manages all known dimensions.
```

python/lsst/daf/butler/dimensions/_graph.py
```diff
@@ -118,83 +211,71 @@ def __new__(
...
-        # (many!) DataCoordinates will share the same DimensionGraph, and
-        # we want them to be lightweight.  The order here is what's convenient
-        # for DataCoordinate: all required dimensions before all implied
-        # dimensions.
-        self._dataCoordinateIndices: dict[str, int] = {
-            name: i for i, name in enumerate(itertools.chain(self.required.names, self.implied.names))
-        }
+    @property
+    @cached_getter
+    def dimensions(self) -> NamedValueAbstractSet[Dimension]:
```
timj, line 252: This is now a deprecated getter?

python/lsst/daf/butler/dimensions/_group.py
```diff
@@ -0,0 +1,435 @@
...
+    def __hash__(self) -> int:
+        return hash(self._seq)
+
+    def __eq__(self, other: object) -> bool:
+        if seq := getattr(other, "_seq", None):
+            return seq == self._seq
+        return super().__eq__(other)
+
+    @classmethod
+    def _from_iterable(cls, iterable: Iterable[str]) -> set[str]:
```
timj, line 79: What calls this private method?

python/lsst/daf/butler/dimensions/_group.py
```diff
@@ -0,0 +1,435 @@
...
+        return DimensionGroup(self.universe, names=names)
+
+    def __or__(self, other: DimensionGroup) -> DimensionGroup:
+        return self.union(other)
+
+    def __and__(self, other: DimensionGroup) -> DimensionGroup:
+        return self.intersection(other)
+
+    @property
+    def data_coordinate_keys(self) -> Set[str]:
```
timj, line 342: It seems odd that this is talking about returning an ordered entity that is a `Set` but this is really a `KeysView`.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -323,7 +323,7 @@ def register(self, datasetType: DatasetType) -> bool:
             )
         record = self._fetch_dataset_type_record(datasetType.name)
         if record is None:
-            dimensionsKey = self._dimensions.saveDimensionGraph(datasetType.dimensions)
+            dimensionsKey = self._dimensions.saveDimensionGroup(datasetType.dimensions.as_group())
```
timj: Did you consider switching to `save_dimension_group`?

python/lsst/daf/butler/script/queryDatasets.py
```diff
@@ -95,7 +95,10 @@ def getAstropyTable(self, datasetTypeName: str) -> AstropyTable:
             raise RuntimeError(f"No DatasetRefs were provided for dataset type {datasetTypeName}")
 
         refInfo = next(iter(self.datasetRefs))
-        dimensions = list(refInfo.datasetRef.dataId.full.keys())
+        dimensions = [
+            refInfo.datasetRef.dataId.universe.dimensions[k]
+            for k in refInfo.datasetRef.dataId.dimensions.data_coordinate_keys
```
timj, line 100: This seems a bit more convoluted than the original. I'm trying to understand why the `data_coordinate_keys` aren't enough given that they are converted to strings in the next line.

timj (approved):
Looks great. DimensionGroup is much better name than DimensionGraph and I see that many isinstance checks have been removed.

python/lsst/daf/butler/_dataset_type.py
TallJimbo: > This property will change to DimensionGroup at some point?

Yes, when `DimensionGraph` is removed this will become `DimensionGroup`, and I think it should be transparently to code that isn't doing deprecated things with it.

> Also typo in doc string.

Rewrote it completely, since `Dimension` instances are no longer in play and the parenthetical type somehow had gotten bumped down to the second paragraph.

python/lsst/daf/butler/dimensions/_coordinate.py
TallJimbo: Should I add `make_empty`, but leave `makeEmpty` un-deprecated (for now)?

python/lsst/daf/butler/dimensions/_coordinate.py
TallJimbo: Yes, but reading these docs again, the logic is actually wrong in a different sense; these need to check for the dimensions to be empty, and if so, call `makeEmpty`, since an empty data ID has all of the zero records it might have.  I've added the checks and updated all of the docs.

python/lsst/daf/butler/dimensions/_graph.py
TallJimbo: Good catch.  Will deprecate and fix any fallout.

python/lsst/daf/butler/dimensions/_group.py
TallJimbo: `~collections.abc.Set`, in mixin methods like `__and__`.  Will add a comment.

python/lsst/daf/butler/dimensions/_group.py
TallJimbo: `KeysView` implements `Set`, and `Set` does not specify order one way or the other, so annotating this as `Set` is a sort of encapsulation: the user should not care that `data_coordinate_keys` is implemented as the keys of a `dict`.

In fact, I actually think it's a huge shame that `KeysView`, `ValuesView`, and `ItemsView` are defined `collections.abc` at all, let alone included in the type annotation for `Mapping`, as they are concrete classes that add nothing in terms of interfaces.  `Mapping.keys()` should have been annotated to return `Set`, and the other two should have been annotated to return `Collection`.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
TallJimbo: Good idea; will do.

python/lsst/daf/butler/script/queryDatasets.py
TallJimbo: The list of actual `Dimension` instances is passed to `sortAstropyTable`, down below.

python/lsst/daf/butler/dimensions/_coordinate.py
timj: We could start aliasing things so that makeEmpty redirects to make_empty. Then we can use the consistent form internally at least.

PR #904, opened 2023-11-06
DM-41117: Improvements to caching and Butler initialization

andy-slac:
A bunch of improvements for collection caching to reduce the need to cache all collections in memory. In some cases when we need to filter collections based on regexes we still need to fetch everything into memory, but in other cases we only read collections that are explicitly specified.

Similarly, collection summary cache is eliminated completely to avoid upfront loading which takes significant time. Now we are only fetching summaries for collections that are needed.

Table reflection is eliminated for static tables, and is postponed for dynamic (tags/calib) tables until they are used.

## Checklist

- [X] ran Jenkins


python/lsst/daf/butler/registry/queries/_sql_query_backend.py
```diff
@@ -121,6 +121,7 @@ def filter_dataset_collections(
         result: dict[DatasetType, list[CollectionRecord]] = {
             dataset_type: [] for dataset_type in dataset_types
         }
+        summaries = self._managers.datasets.fetch_summaries(collections, result.keys())
```
TallJimbo, line 124: I'd prefer to move this fetch a little higher up the call stack, so if someone queries for many dataset types in the same list of collections in a single `queryDatasets` call, we can do one fetch for all dataset types in those collections.  This pattern in particular is something I'd like to make sure we optimize:
```
for result_for_dataset_type in butler.registry.queryDatasets(..., collections=x).byParentDatasetType():
    if result_for_dataset_type.any(exact=False, execute=False):
        print(f"Found {result_for_dataset_type.parentDatasetType.name}")
```
If that's too messy right now, we should keep it in mind for the new query interface.

python/lsst/daf/butler/registry/interfaces/_collections.py
```diff
@@ -50,7 +49,10 @@
     from ._dimensions import DimensionRecordStorageManager
 
 
-class CollectionRecord:
+_Key = TypeVar("_Key")
+
+
+class CollectionRecord(Generic[_Key]):
```
TallJimbo, line 56: This is probably a topic for another ticket, but I'm guessing we'll find the typing works even better (especially if we end up sending these things from server to client via pydantic) if we switch from an ABC pattern to a descriminated union one, e.g.:
```
@dataclasses.dataclass
class RunRecord(Generic[_K]):
    key: _K
    name: str
    type: Literal[CollectionType.RUN]

@dataclasses.dataclass
class ChainedCollectionRecord(Generic[_K]):
    key: _K
    name: str
    type: Literal[CollectionType.CHAINED]
    children: Sequence[str]

CollectionRecord: TypeAlias = Union[RunRecord, ChainedCollectionRecord]   # etc.
```

But :+1: to adding generics here.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
```diff
@@ -77,20 +77,36 @@ def __init__(
...
+            self._tags_table = self._tags_table_factory()
+        return self._tags_table
+
+    @property
+    def _calibs(self) -> sqlalchemy.schema.Table | None:
+        if self._calibs_table is None:
+            if self._calibs_table_factory is None:
+                return None
+            self._calibs_table = self._calibs_table_factory()
+        return self._calibs_table
```
TallJimbo, line 109: This is already an improvement as-is, but if we moved this pattern of only looking up the table on first use up a level, so that we only make a `DatasetRecordStorage` instance at all on first use, it might help us get the bad kind of caching out of that manager so we can use it on the server.  Happy to defer that work to another ticket; just figured it was worth mentioning here because it might make a lot of this change go away.

python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -323,12 +274,31 @@ def find(self, name: str) -> CollectionRecord[K]:
             raise MissingCollectionError(f"No collection with name '{name}' found.")
         return result
 
+    def _find_many(self, names: Iterable[str]) -> list[CollectionRecord[K]]:
+        """Return multiple records given their names."""
+        names = list(names)
+        # To protect against potential races in cache updates.
```
TallJimbo, line 283: Oof, if this is because FastAPI without async forces threading on us, and we expect the same `SqlRegistry` to be used by multiple threads, there's gonna be a lot more of this, enough so that it makes me think we need to do something else, like have explicit thread-local `SqlRegistry` instance (maybe sharing a connection pool?) or get all caching out of `SqlRegistry`.

python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -419,11 +415,13 @@ def getParentChains(self, key: Any) -> Iterator[ChainedCollectionRecord[K]]:
...
-        for key in parent_keys:
-            # TODO: Just in case cached records miss new parent collections.
-            # This is temporary, will replace with non-cached records soon.
-            with contextlib.suppress(KeyError):
-                yield cast(ChainedCollectionRecord[K], self._records[key])
+        # TODO: It would be more efficient to write a single query that both
+        # finds parents and all their children, but for now we do not care
+        # much about efficiency. Also the only client of this method does not
+        # need full records, only parent collection names, maybe we should
+        # change this method to return names instead.
```
TallJimbo: :+1: to making this return names only.

TallJimbo (approved):
Everything here looks great; definitely a step in the right direction.

python/lsst/daf/butler/registry/queries/_sql_query_backend.py
andy-slac: I think I could add summaries-based filtering logic to `SqlRegistry._standardize_query_dataset_args` if that feels like the right place. We probably still want filtering here in `SqlQueryBackend.filter_dataset_collections`, as it is called from `SqlRegistry.findDataset` (though it only passes a single dataset type)?

python/lsst/daf/butler/registry/interfaces/_collections.py
andy-slac: We'll probably need a `TaggedCollectionRecord` also?

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
andy-slac: I think there is a reason to delay table reflection even if we create DatasetRecordStorage instances on-demand. `resolve_wildcard` needs all dataset types when patters are used, so we'd probably need to make all storages at that point, but we still don't need to know their tables' schemas. Alternative is probably to just query dataset types on each `resolve_wildcard` without making storage instances?

python/lsst/daf/butler/registry/collections/_base.py
andy-slac: Well, I was just thinking about concurrency in general, without async context. It's true that caches make everything very difficult, would be better to get rid of caching entirely. We'll have to think about it very soon.

python/lsst/daf/butler/registry/collections/_base.py
andy-slac: I pushed [ceede1c](https://github.com/lsst/daf_butler/pull/904/commits/ceede1c2c838245844f5fe567a0f76b2df20a011) which changes return type to strings.

python/lsst/daf/butler/registry/interfaces/_collections.py
TallJimbo: Yes, and `CalibrationCollectionRecord`.  Just didn't try to write out the whole thing.

python/lsst/daf/butler/registry/queries/_sql_query_backend.py
TallJimbo: Another possibility might be be to pass the summaries down to this method and just fetch them in `SqlRegistry` methods.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
TallJimbo: We could also load the list of dataset type names to do the pattern-matching without necessarily loading the storages at the same time, in the same way we talked about decoupling loading the children-of-chains names from loading the records of those children.

python/lsst/daf/butler/registry/queries/_sql_query_backend.py
andy-slac: OK, I'll look at this later today, maybe I can add it on this branch still.

python/lsst/daf/butler/registry/interfaces/_collections.py
andy-slac: Looks reasonable, I think this deserves a dedicated ticket.

python/lsst/daf/butler/registry/queries/_sql_query_backend.py
andy-slac: I think this deserves a separate ticket too, I looked at what it needs and in some context we'd have to resolve both collection wildcards and dataset wildcards before fetching summaries. I want to think about it a bit more, maybe we need to change `fetch_summaries` to accept wildcards instead of just lists.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
andy-slac: OK, I added another commit [8d3458f](https://github.com/lsst/daf_butler/pull/904/commits/8d3458fade951351207fd051182261fe6e3f4439) which completely eliminates caching from datasets manager. It makes summaries manager a bit messier and more dependent on datasets manager (and dataset_type table). Could you check it and tell if you like this variant better?

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -359,15 +359,17 @@ def register(self, datasetType: DatasetType) -> tuple[DatasetRecordStorage, bool
...
                 raise ConflictingDefinitionError(
                     f"Given dataset type {datasetType} is inconsistent "
-                    f"with database definition {storage.datasetType}."
+                    f"with database definition {record.dataset_type}."
                 )
+            storage = self._make_storage(record)
             inserted = False
+        # TODO: We return storage instance from this method, but the only
+        # client that uses this method ignores it. Maybe we should drop it
+        # and avoid making storage instance above.
```
TallJimbo: :+1:

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -492,49 +491,77 @@ def getDatasetRef(self, id: DatasetId) -> DatasetRef | None:
...
-        """Return dataset type given its ID."""
-        return self._byId[dataset_type_id].datasetType
+    def _fetch_dataset_type_record(self, name: str) -> _DatasetTypeRecord | None:
+        """Retrieve all dataset types defined in database.
+
+        Yields
+        ------
+        dataset_types : `tuple` [`_DatasetTypeRecord`]
+            Information from a single database record.
+        """
```
TallJimbo, line 532: Docstring is not correct.

python/lsst/daf/butler/registry/datasets/byDimensions/summaries.py
```diff
@@ -240,21 +248,20 @@ def update(
...
-        dataset_type_ids : `~collections.abc.Iterable` [`int`]
-            IDs of dataset types to include into returned summaries. If `None`
-            then all dataset types will be included.
-        dataset_type_factory : `Callable`
-            Method that returns `DatasetType` instance given its dataset type
-            ID.
+        dataset_type_names : `~collections.abc.Iterable` [`str`]
+            Names of dataset types to include into returned summaries. If
+            `None` then all dataset types will be included.
+        dataset_type_ids : `~collections.abc.Mapping` [`int`, `DatasetType`]
```
TallJimbo: Docstring is inconsistent with type annotation.

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
TallJimbo: It's certainly more viable for use in the server, so I think we'll have to do something like this (or something more drastic in terms of refactoring managers) at some point.  I don't think the manager system is really serving us well at this point, so I'm not too bothered by the collection and dataset managers getting more dependent on each other - it just feels like yet another indication of something we already knew, or at least suspected.

I am a little worried that stripping the dataset type caching out of the managers without putting it somewhere else in `DirectButler` could be a noticeable regression for certain use cases, but it's probably okay for now since you should be able to claim a much more obvious speedup in `DirectButler` initialization times from this ticket.

TallJimbo (approved):

python/lsst/daf/butler/registry/datasets/byDimensions/_storage.py
andy-slac: I think I need to re-introduce some caching, at Buter level. I checked how many queries graph builder generates for ci_hsc_gen3 and there is a large increase compared to fully-cached case (timing is not very different, but this was with local postgres).

andy-slac:
@TallJimbo, sorry for another review request, I have added three more commits to try to improve things:
- [5833c91](https://github.com/lsst/daf_butler/pull/904/commits/5833c91f55f4be34daf795f115f44fdbc45a7085) - removes caching from dataset type manager, looking at the debug logs I realized that we now have too many SQL queries.
- [14721a2](https://github.com/lsst/daf_butler/pull/904/commits/14721a2eab1044892224da7477cd8a6a87d276e7) - adds context manager to Registry that enables caching and adds caches for collections and collection summaries
- [1e59418](https://github.com/lsst/daf_butler/pull/904/commits/1e59418b37380ef4b4ec85de290278c70fbd38db) - adds cache for dataset types.

I'm not very happy with how it is structured now, cache for dataset types looks particularly ugly, but this is the best that I could do. As the result, number of queries is approximately the same in both QG building and `pipetask run` in `ci_hsc_gen3`. (In QG generation the number of queries (but not time, in my setup) is actually dominated by `_skyFrameLookup`).

python/lsst/daf/butler/_registry_shim.py
```diff
@@ -102,6 +102,12 @@ def refresh(self) -> None:
         # Docstring inherited from a base class.
         self._registry.refresh()
 
+    @contextlib.contextmanager
+    def caching_context(self) -> Iterator[None]:
+        # Docstring inherited from a base class.
+        with self._registry.caching_context():
```
TallJimbo: I think it's reasonable to put the public interface on `RegistryShim` rather than `Butler` - it gives us more time to think about whether a future context for caching in the new query is the same or different - but I think we want to put the implementation in a `Butler` private method so it's easier to implement on `RemoteButler`, which won't have a real `Registry` on the client.  I think that's true whether it's a client-side cache in `RemoteButler` or a per-user cache on the server (both of which have some serious downsides), or both.

python/lsst/daf/butler/registry/_caching_context.py
```diff
@@ -0,0 +1,64 @@
...
+
+    Notes
+    -----
+    Caching is usually disabled for most of the record types, but it can be
+    explicitly and temporarily enabled in some context (e.g. quantum graph
+    building) using Registry method. This class is a collection of cache
+    instances which will be `None` when caching is disabled. Instance of this
+    class is passed to the relevant managers that can use it to query or
+    populate caches when caching is enabled.
+    """
```
TallJimbo, line 57: FYI I'm definitely thinking a lot about dimension record caching on DM-41114 right now (since all the complexity in a collection of `DataCoordinate` comes from associating the data IDs with dimension records), and I think it'll work quite nicely to just add dimension record caches as another attribute here.

python/lsst/daf/butler/registry/_collection_record_cache.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class CollectionRecordCache:
+    """Cache for collection records.
+
+    Notes
+    -----
+    This class stores collection records and can retrieve them using either
+    collection name or collection key. One complication is that key type can be
+    either collection name or a distinct integer value. To optimize storage
```
TallJimbo, line 46: Side comment about this complication: the more I think about it, the more I think we should migrate ourselves out of this problem, and move all repos to synthetic integer keys for collections.  It's >100 bytes per dataset tag savings (i.e. one tag for the `RUN` collection, more for each `TAGGED` or `CALIBRATION` collection) in storage, and I think that'll add up.  It will also allow us to rename collections easily, and that's a frequent feature request.

TallJimbo (approved):
Looks good - I think it's time to wrap this one up and start one or more new tickets for further caching work, as I think this is a clear step in the right direction with some immediate benefits and no obvious downsides.

I do still think there's a ton of work ahead here, because the caching context interface implies a per-user cache and that'd be a massive pain to put on the server, and yet some of what we cache now can only go on the server (SQLAlchemy table objects) or needs to be on the server as a check even if it's also on the client (collection summaries).  I think I'd advise some "step back and design it conceptually" work before further coding on this problem, though.

Typo: `additiona` in commit message for 14721a2eab1044892224da7477cd8a6a87d276e7.

python/lsst/daf/butler/_registry_shim.py
andy-slac: I can add `_caching_context()` method to `Butler`, though at this point it should not matter, as the only client of this context is QG builder, and I don't think that QG builder could start using RemoteButler any time soon.

andy-slac: I added [922359c](https://github.com/lsst/daf_butler/pull/904/commits/922359ce8e266e057c40a3cd47089f66ac0f6b74) as future-proofing for remote butler caching mechanism. RegistryShim still can only exist with actual SqlRegistry for now.

PR #905, opened 2023-11-15
DM-41705: Fix RemoteButler import

dhirving:
## Checklist

- [x] ran Jenkins


timj (approved):
This looks okay. I might quibble that server models are needed by both server and by remote butler and so putting them in either server or remote_butler seems wrong but I have no better place to suggest (previously they were treated as a shared model in `core`).

PR #907, opened 2023-11-17
DM-41724: make dimension record construction more strict

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/dimensions/_records.py
+++ b/python/lsst/daf/butler/dimensions/_records.py
@@ -109,7 +109,7 @@ def _createSimpleRecordSubclass(definition: DimensionElement) -> type[SpecificSe
             field_type = Optional[field_type]  # type: ignore
         members[field.name] = (field_type, ...)
     if definition.temporal:
-        members["timespan"] = (Tuple[int, int], ...)  # type: ignore
+        members["timespan"] = (Optional[Tuple[int, int]], ...)  # type: ignore
     if definition.spatial:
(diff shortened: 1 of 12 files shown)


python/lsst/daf/butler/dimensions/_records.py
```diff
@@ -299,6 +288,26 @@ def __init__(self, **kwargs: Any):
...
+            object.__setattr__(self, name, kwargs.pop(name, None))
+        if self.definition.temporal is not None and self.timespan is None and "datetime_begin" in kwargs:
+            object.__setattr__(
+                self,
+                "timespan",
+                Timespan(
+                    kwargs.pop("datetime_begin"),
+                    kwargs.pop("datetime_end"),
+                ),
+            )
```
dhirving, line 309: There's a small change in behavior here, not sure if it is intended to enforce that any Timespan here will always be bounded or not.

Previously if only `datetime_begin` was passed without also including `datetime_end`, it would create a Timespan without an upper bound.  Now it throws KeyError in this case instead.

dhirving (approved):
Looks OK to me, other than the one behavior change above where I wasn't sure if it was intentional or not.

python/lsst/daf/butler/dimensions/_records.py
TallJimbo: It was intentional, but the reader shouldn't have to wonder about that; I'll add a code comment saying that if you pass one you're required to pass both.

PR #908, opened 2023-11-17
DM-41750: Fix some problems with global state in tests

timj:
Sometimes tests would fail depending on what other tests have already run.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/queries/expressions/parser/parserYacc.py
+++ b/python/lsst/daf/butler/registry/queries/expressions/parser/parserYacc.py
@@ -78,7 +78,7 @@ _re_time_str = re.compile(
     (?P<value>
         (?P<number>-?(\d+(\.\d*)|(\.\d+)))   # floating point number
         |
-        (?P<iso>\d+-\d+-\d+([ T]\d+:\d+(:\d+([.]\d*)?)?)?)   # iso(t)
+        (?P<iso>\d+-\d+-\d+([ T]\d+:\d+(:\d+([.]\d*)?)?)?)   # iso(t) [no timezone]
         |
         (?P<fits>[+]\d+-\d+-\d+(T\d+:\d+:\d+([.]\d*)?)?)   # fits
         |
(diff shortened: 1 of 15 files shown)


python/lsst/daf/butler/registry/queries/expressions/parser/parserYacc.py
```diff
@@ -108,6 +108,11 @@ def _parseTimeString(time_str):
     ValueError
         Raised if input string has unexpected format
     """
+    # Check for time zone. Python datetime objects can be timezone-aware
+    # and if one has been stringified then there will be a +00:00 on the end.
+    # Special case UTC. Fail for other timezones.
+    time_str = time_str.replace("+00:00", "")
```
andy-slac, line 114: If you strip timezone then you don't need to modify regex to match it. OTOH this feels a bit fragile, I'm not really sure if timezone representation in Python is locale-specific. The updated regex allows both `+00` and `+00:00` (and non-zero variants). Maybe we can check what exactly regex matches and strip it, and probably verify that it is all zeros and raise if user gave it non-UTC time? Or allow only `+00:00` (optionally) in regex, so that parser fails for anything else.

andy-slac (approved):
Looks good, one minor comment.

python/lsst/daf/butler/registry/queries/expressions/parser/parserYacc.py
timj: I'm pretty sure the `+/-\d\d:\d\d` is not locale-specific for timezones in ISO standard.

This code will fail if a non-UTC timezone turns up because Astropy will not like it. I think the error message is a bit opaque though because it hides the complaint from astropy itself. Maybe it is better to change the regex to optional `+00:00` and strip it here since that will trigger the error message higher up.

andy-slac: I saw your recent commit - looks like you are still stripping `+00:00` before doing regex check, and there is no reason for regex to have timezone in it if you strip it beforehand?

timj: I thought that stripping happened after the regex check... I am going to check again.

timj: Ha. Okay. Now I see. The regex triggers after the stripping in that astropy routine. I will fix to only do the stripping.

PR #912, opened 2023-11-28
DM-41878: Implement RemoteButler.get() backed by a single FileDatastore

dhirving:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -814,6 +814,7 @@ class Butler(LimitedButler):
     def get_dataset(
         self,
         id: DatasetId,
+        *,
         storage_class: str | StorageClass | None = None,
         dimension_records: bool = False,
         datastore_records: bool = False,
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -814,7 +814,7 @@ class Butler(LimitedButler):
     def get_dataset(
         self,
         id: DatasetId,
-        storage_class: str | StorageClass | None,
+        storage_class: str | StorageClass | None = None,
         dimension_records: bool = False,
         datastore_records: bool = False,
     ) -> DatasetRef | None:
--- a/python/lsst/daf/butler/datastore/cache_manager.py
+++ b/python/lsst/daf/butler/datastore/cache_manager.py
@@ -1015,7 +1015,11 @@ class DatastoreDisabledCacheManager(AbstractDatastoreCacheManager):
         in lookup keys.
     """
 
-    def __init__(self, config: str | DatastoreCacheManagerConfig, universe: DimensionUniverse):
+    def __init__(
+        self,
+        config: str | DatastoreCacheManagerConfig | None = None,
+        universe: DimensionUniverse | None = None,
+    ):
         return
 
     def should_be_cached(self, entity: DatasetRef | DatasetType | StorageClass) -> bool:
--- a/python/lsst/daf/butler/datastore/stored_file_info.py
+++ b/python/lsst/daf/butler/datastore/stored_file_info.py
@@ -27,13 +27,14 @@
 
 from __future__ import annotations
 
-__all__ = ("StoredDatastoreItemInfo", "StoredFileInfo")
+__all__ = ("StoredDatastoreItemInfo", "StoredFileInfo", "SerializedStoredFileInfo")
 
 import inspect
 from collections.abc import Iterable, Mapping
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
 
+from lsst.daf.butler._compat import _BaseModelCompat
 from lsst.resources import ResourcePath
 from lsst.utils import doImportType
 from lsst.utils.introspection import get_full_type_name
@@ -260,6 +261,13 @@ class StoredFileInfo(StoredDatastoreItemInfo):
             **kwargs,
         )
 
+    def to_simple(self) -> SerializedStoredFileInfo:
+        record = self.to_record()
+        # We allow None on the model but the record contains a "null string"
+        # instead
+        record["component"] = self.component
+        return SerializedStoredFileInfo.model_validate(record)
+
     def file_location(self, factory: LocationFactory) -> Location:
         """Return the location of artifact.
 
@@ -307,6 +315,10 @@ class StoredFileInfo(StoredDatastoreItemInfo):
         )
         return info
 
+    @classmethod
+    def from_simple(cls: type[StoredFileInfo], model: SerializedStoredFileInfo) -> StoredFileInfo:
+        return cls.from_record(dict(model))
+
     def update(self, **kwargs: Any) -> StoredFileInfo:
         new_args = {}
         for k in self.__slots__:
@@ -320,3 +332,26 @@ class StoredFileInfo(StoredDatastoreItemInfo):
 
     def __reduce__(self) -> str | tuple[Any, ...]:
         return (self.from_record, (self.to_record(),))
+
+
+class SerializedStoredFileInfo(_BaseModelCompat):
+    """Serialized representation of `StoredFileInfo` properties"""
+
+    formatter: str
+    """Fully-qualified name of Formatter."""
+
+    path: str
+    """Path to dataset within Datastore."""
+
+    storage_class: str
+    """Name of the StorageClass associated with Dataset."""
+
+    component: str | None
+    """Component associated with this file. Can be None if the file does
+    not refer to a component of a composite."""
+
+    checksum: str | None
+    """Checksum of the serialized dataset."""
+
+    file_size: int
+    """Size of the serialized dataset in bytes."""
--- a/python/lsst/daf/butler/datastore/stored_file_info.py
+++ b/python/lsst/daf/butler/datastore/stored_file_info.py
@@ -215,7 +215,7 @@ class StoredFileInfo(StoredDatastoreItemInfo):
     """StorageClass associated with Dataset."""
 
     component: str | None
-    """Component associated with this file. Can be None if the file does
+    """Component associated with this file. Can be `None` if the file does
     not refer to a component of a composite."""
 
     checksum: str | None
@@ -347,7 +347,7 @@ class SerializedStoredFileInfo(_BaseModelCompat):
     """Name of the StorageClass associated with Dataset."""
 
     component: str | None
-    """Component associated with this file. Can be None if the file does
+    """Component associated with this file. Can be `None` if the file does
     not refer to a component of a composite."""
 
     checksum: str | None
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -37,10 +37,11 @@ import logging
 from collections import defaultdict
 from collections.abc import Callable, Iterable, Mapping, Sequence
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, ClassVar, cast
+from typing import TYPE_CHECKING, Any, ClassVar, cast, TypeAlias
 
 from lsst.daf.butler import (
     Config,
+    DataCoordinate,
     DatasetId,
     DatasetRef,
     DatasetType,
@@ -69,7 +70,7 @@ from lsst.daf.butler.datastore.cache_manager import (
 )
 from lsst.daf.butler.datastore.composites import CompositesMap
 from lsst.daf.butler.datastore.file_templates import FileTemplates, FileTemplateValidationError
-from lsst.daf.butler.datastore.generic_base import GenericBaseDatastore
+from lsst.daf.butler.datastore.generic_base import GenericBaseDatastore, post_process_get
 from lsst.daf.butler.datastore.record_data import DatastoreRecordData
 from lsst.daf.butler.datastore.stored_file_info import StoredDatastoreItemInfo, StoredFileInfo
 from lsst.daf.butler.registry.interfaces import (
@@ -136,6 +137,9 @@ class DatastoreFileGetInformation:
 
     readStorageClass: StorageClass
     """The `StorageClass` of the dataset being read."""
+
+
+DatasetLocationInformation: TypeAlias = tuple[Location, StoredFileInfo]
 
 
 class FileDatastore(GenericBaseDatastore[StoredFileInfo]):
@@ -577,7 +581,7 @@ class FileDatastore(GenericBaseDatastore[StoredFileInfo]):
 
     def _get_dataset_locations_info(
         self, ref: DatasetIdRef, ignore_datastore_records: bool = False
-    ) -> list[tuple[Location, StoredFileInfo]]:
+    ) -> list[DatasetLocationInformation]:
         r"""Find all the `Location`\ s  of the requested dataset in the
         `Datastore` and the associated stored file information.
 
@@ -702,7 +706,7 @@ class FileDatastore(GenericBaseDatastore[StoredFileInfo]):
             for location, formatter, storageClass, component in all_info
         ]
 
-    def _prepare_for_get(
+    def _prepare_for_direct_get(
         self, ref: DatasetRef, parameters: Mapping[str, Any] | None = None
     ) -> list[DatastoreFileGetInformation]:
         """Check parameters for ``get`` and obtain formatter and
@@ -738,8 +742,6 @@ class FileDatastore(GenericBaseDatastore[StoredFileInfo]):
             fileLocations = self._get_expected_dataset_locations_info(ref)
 
         if len(fileLocations) > 1:
-            disassembled = True
-
             # If trust is involved it is possible that there will be
             # components listed here that do not exist in the datastore.
             # Explicitly check for file artifact existence and filter out any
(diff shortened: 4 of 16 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -808,7 +808,7 @@ def get_dataset_type(self, name: str) -> DatasetType:
     def get_dataset(
         self,
         id: DatasetId,
-        storage_class: str | StorageClass | None,
+        storage_class: str | StorageClass | None = None,
```
timj, line 818: Maybe we mark this method as requiring keyword args after the `id`?

python/lsst/daf/butler/datastore/cache_manager.py
```diff
@@ -1015,7 +1015,11 @@ class DatastoreDisabledCacheManager(AbstractDatastoreCacheManager):
         in lookup keys.
     """
 
-    def __init__(self, config: str | DatastoreCacheManagerConfig, universe: DimensionUniverse):
+    def __init__(
+        self,
+        config: str | DatastoreCacheManagerConfig | None = None,
+        universe: DimensionUniverse | None = None,
```
timj, line 1021: I think we get away with this because we aren't yet trying to cache anything that uses the advanced LookupKey features?

python/lsst/daf/butler/datastore/stored_file_info.py
```diff
@@ -320,3 +332,26 @@ def update(self, **kwargs: Any) -> StoredFileInfo:
 
     def __reduce__(self) -> str | tuple[Any, ...]:
         return (self.from_record, (self.to_record(),))
+
+
+class SerializedStoredFileInfo(_BaseModelCompat):
```
timj, line 337: I still think it's a bit odd that we now have a dataclass defining the datastore record content, a pydantic model defining the datastore record content, and a method in FileDatastore for creating the database table that must match the other two but uses neither to define it. If `makeTableSpec` built up its definition from `StoredFileInfo` (if we are saying that it is required that `StoredFileInfo` and `makeTableSpec` agree but the serialized form is allowed to be different) that would at least limit the duplication.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1993,10 +1994,23 @@ def get(
...
         allGetInfo = self._prepare_for_direct_get(ref, parameters)
-        return get_dataset_as_python_object(
+        return get_dataset_as_python_object_from_get_info(
             allGetInfo, ref=ref, parameters=parameters, cache_manager=self.cacheManager
         )
 
+    def prepare_get_for_external_client(self, ref: DatasetRef) -> FileDatastoreGetPayload:
+        # Docstring inherited
+
+        def to_file_info_payload(info: DatasetLocationInformation) -> FileDatastoreGetPayloadFileInfo:
```
timj, line 2001: Why is the helper function needed here? Isn't it clearer to do instead:

```python
file_records = []
for location, file_info in self._get_dataset_locations_info(ref):
    file_records.append(FileDatastoreGetPayloadFileInfo(...)
```
?

python/lsst/daf/butler/datastores/fileDatastoreClient.py
```diff
@@ -0,0 +1,58 @@
...
+
+    datastore_file_info = generate_datastore_get_information(
+        fileLocations,
+        refStorageClass=ref.datasetType.storageClass,
+        refComponent=ref.datasetType.component(),
+        parameters=parameters,
+        dataId=ref.dataId,
+    )
+    return get_dataset_as_python_object_from_get_info(
+        datastore_file_info, ref=ref, parameters=parameters, cache_manager=DatastoreDisabledCacheManager()
```
timj, line 93: Why is the cache being disabled here? Isn't the cache the thing we want for remote data?

python/lsst/daf/butler/remote_butler/server/handlers/_external.py
```diff
@@ -152,3 +157,16 @@ def find_dataset(
...
+
+@external_router.post("/v1/get_file")
+def get_file(
+    request: GetFileRequestModel,
+    factory: Factory = Depends(factory_dependency),
+) -> GetFileResponseModel:
+    butler = factory.create_butler()
+    ref = butler.get_dataset(request.dataset_id, datastore_records=True)
+    if ref is None:
+        raise NotFoundException()
```
timj: You don't want to tell them which ref wasn't found because you expect the client to raise the exception itself based entirely on the 404 and not the exception raised here?

python/lsst/daf/butler/remote_butler/server/handlers/_external.py
```diff
@@ -152,3 +157,16 @@ def find_dataset(
         **data_id,
     )
     return ref.to_simple() if ref else None
+
+
+@external_router.post("/v1/get_file")
+def get_file(
+    request: GetFileRequestModel,
```
timj, line 164: This is a POST with a model because you might want to pass in more complex content in the future so using GET with the UUID is not something you want to do?

python/lsst/daf/butler/tests/utils.py
```diff
@@ -61,6 +61,9 @@ class TestCaseMixin:
         pass
 
 
+METRICS_EXAMPLE_SUMMARY = {"AM1": 5.2, "AM2": 30.6}
```
timj: Since this is now pseudo public and used elsewhere can you add a triple-quoted docstring below this explaining what it's for? And maybe add to `__all__`? Wouldn't it be more useful for the general case if this was the full MetricsExample that is being created so that other parts can be tested by people? Rather than solely the summary part of it? Then we have a story which says that we guarantee that this is the python object that has been stored as an example for test purposes?

tests/test_server.py
```diff
@@ -197,6 +203,29 @@ def override_read(http_resource_path):
...
         assert str(butler._config.remote_butler.url) == "https://test.example/api/butler/"
 
+    def test_get(self):
+        # Test get() of a DatasetRef
+        ref = self.butler.find_dataset(
+            "test_metric_comp",
+            {"instrument": "DummyCamComp", "visit": 423},
+            collections="ingest/run",
+        )
+        metric = self.butler.get(ref)
```
timj, line 206: I guess the line above that says:

```
# Not yet testing butler.get()
```
is no longer relevant?

python/lsst/daf/butler/datastores/fileDatastoreClient.py
```diff
@@ -15,34 +15,71 @@
...
     universe: DimensionUniverse,
     parameters: Mapping[str, Any] | None,
     storageClass: StorageClass | str | None,
 ) -> Any:
+    """Retrieve an artifact from storage and return it as a Python object
+
+    Parameters
+    ----------
+    payload : `FileDatastoreGetPayload`
+        Pre-processed information about each file associated with this artifact
```
timj: Will not comment on the period since we discussed that today.

python/lsst/daf/butler/datastores/fileDatastoreClient.py
```diff
@@ -15,34 +15,71 @@
...
+    universe: `DimensionUniverse`
+        The universe of dimensions associated with the `DatasetRef` contained
+        in `payload`.
+    parameters : `Mapping`[`str`, `Any`]
```
timj: Bizarrely github doesn't let me make suggestions on an individual commit but I think you want `typing.Any` here for it to render properly.

python/lsst/daf/butler/datastores/fileDatastoreClient.py
```diff
@@ -15,34 +15,71 @@
...
 ) -> Any:
+    """Retrieve an artifact from storage and return it as a Python object
+
+    Parameters
+    ----------
+    payload : `FileDatastoreGetPayload`
+        Pre-processed information about each file associated with this artifact
+    universe: `DimensionUniverse`
+        The universe of dimensions associated with the `DatasetRef` contained
+        in `payload`.
```
timj: Need double back ticks here since `payload` is not a class.

python/lsst/daf/butler/datastores/file_datastore/get.py
```diff
@@ -154,6 +175,8 @@ def _read_artifact_into_memory(
         The registry information associated with this artifact.
     isComponent : `bool`
         Flag to indicate if a component is being read from this artifact.
+    cache_manager: `AbstractDatastoreCacheManager`
+        The cache manager to use for caching retrieved files
```
timj, line 179: I'm concerned that there is no coverage for the caching code below.

python/lsst/daf/butler/datastore/stored_file_info.py
```diff
@@ -320,3 +332,26 @@ def update(self, **kwargs: Any) -> StoredFileInfo:
...
+    """Fully-qualified name of Formatter."""
+
+    path: str
+    """Path to dataset within Datastore."""
+
+    storage_class: str
+    """Name of the StorageClass associated with Dataset."""
+
+    component: str | None
+    """Component associated with this file. Can be None if the file does
```
timj:
```suggestion
    """Component associated with this file. Can be `None` if the file does
```

python/lsst/daf/butler/datastores/file_datastore/get.py
```diff
@@ -0,0 +1,538 @@
...
+
+    component: str | None
+    """The component to be retrieved (can be `None`)."""
+
+    readStorageClass: StorageClass
+    """The `StorageClass` of the dataset being read."""
+
+
+def generate_datastore_get_information(
+    fileLocations: list[DatasetLocationInformation],
```
timj, line 81: Maybe we `s/fileLocations/file_locations/` and `s/readStorageClass/read_storage_class/` in these new APIs to move into the future?

python/lsst/daf/butler/datastores/file_datastore/get.py
```diff
@@ -0,0 +1,538 @@
...
+        The registry information associated with this artifact.
+    parameters : `Mapping`[`str`, `Any`]
+        `StorageClass` and `Formatter` parameters
+    readStorageClass : `StorageClass` | `None`, optional
+        The StorageClass to use when ultimately returning the resulting object
+        from the get.  Defaults to the `StorageClass` specified by `ref`
+
+    Returns
+    -------
+    getInfo : `list[DatastoreFileGetInformation]`
```
timj:
```suggestion
    getInfo : `list` [`DatastoreFileGetInformation`]
```

python/lsst/daf/butler/datastores/file_datastore/get.py
```diff
@@ -0,0 +1,538 @@
...
+                component,
+                readStorageClass,
+            )
+        )
+
+    return fileGetInfo
+
+
+def _read_artifact_into_memory(
+    getInfo: DatastoreFileGetInformation,
```
timj, line 162: get_info? These are all internal methods so if we want to go all underscores we don't have to ask anyone.

python/lsst/daf/butler/datastores/file_datastore/get.py
```diff
@@ -0,0 +1,538 @@
...
+
+    Parameters
+    ----------
+    allGetInfo : `list`[`DatastoreFileGetInformation`]
+        Pre-processed information about each file associated with this artifact
+    ref : `DatasetRef`
+        The registry information associated with this artifact.
+    parameters : `Mapping`[`str`, `Any`]
+        `StorageClass` and `Formatter` parameters
+    cache_manager: `AbstractDatastoreCacheManager`
```
timj:
```suggestion
    cache_manager : `AbstractDatastoreCacheManager`
```
hopefully some kind of linter would spot this. The numpydoc one says it will.

python/lsst/daf/butler/datastores/file_datastore/get.py
```diff
@@ -0,0 +1,538 @@
...
+        The cache manager to use for caching retrieved files
+
+    Returns
+    -------
+    python_object: `Any`
```
timj:
```suggestion
    python_object : `typing.Any`
```

python/lsst/daf/butler/datastores/file_datastore/get.py
```diff
@@ -0,0 +1,538 @@
...
+        The retrieved artifact, converted to a Python object according to the
+        `StorageClass` specified in `ref`.
```
timj:
```suggestion
        `StorageClass` specified in ``ref``.
```

timj (approved):
Looks great.

There are some numpydoc issues and single backticks vs double backticks issues for rst fun.

python/lsst/daf/butler/datastore/cache_manager.py
dhirving: This is the constructor for the no-op `DatastoreDisabledCacheManager` which doesn't actually use either of these parameters because it does nothing... but I think it has to have them as optionals because of the way the cache manager is instantiated from configuration.

timj: Oh yes. I had missed that it was the disabled cache constructor.

python/lsst/daf/butler/datastores/fileDatastoreClient.py
dhirving: We do eventually want it to be enabled, but I'll need to make a place for it to live and still need to audit it for thread-safety.  I didn't see any locks in there so I assumed by default that it's probably not threadsafe.  

Also it probably should be configurable on the client side, I'm assuming that the services backend use case probably does not want caching on a local disk.  Probably should be on by default for the notebook use case.

I should have made a ticket for this though, I'll do that.

python/lsst/daf/butler/datastores/file_datastore/get.py
dhirving: I was trying not to unnecessarily alter the body of these functions to make it clear what I had actually changed when I was extracting them

python/lsst/daf/butler/datastores/fileDatastoreClient.py
timj: It's not using globals and is meant to be safe for use with multiple processes sharing a cache. I would not expect threads to be a problem -- it mostly scans the file system for its source of truth and when a file is being retrieved it puts it in a separate location to make sure another process doesn't delete the file whilst it's being read.

dhirving: Yeah so it seems that having a separate instance of DatastoreCacheManager per-thread is likely threadsafe, but sharing an instance across threads is not because of `self._cache_entries`.

timj: Yeah, it's a mutable object so that's always going to be a problem.

python/lsst/daf/butler/remote_butler/server/handlers/_external.py
dhirving: Yeah I hadn't decided yet whether the two lookup types for get() were different endpoints or not, will have to figure that out in DM-41941.

python/lsst/daf/butler/datastores/file_datastore/get.py
dhirving: The lack of coverage predates this change, see https://app.codecov.io/gh/lsst/daf_butler/blob/main/python%2Flsst%2Fdaf%2Fbutler%2Fdatastores%2FfileDatastore.py#L1459

And the client isn't using caching yet which is why coverage wasn't added here.

timj: As discussed yesterday, I think it's because the MetricsExample is not on the allowed caching list. It's likely that we changed a test at some point which changed the storage class. I can take a look at that.

python/lsst/daf/butler/datastore/stored_file_info.py
dhirving: StoredFileInfo isn't actually a dataclass, it's a subclass of the abstract base class StoredDatastoreItemInfo with a property that can't be instantiated without injecting configuration and several business logic methods.  And the representation of some of the fields varies between the various cases (e.g. component is `str | None` on the public API but in the database it's `str | Literal['__NULL_STRING__']`.)

I had considered making StoredFileInfo have-a SerializedStoredFileInfo and forwarding the properties to it, but it didn't seem to buy much because you still have to have the duplicate properties to do the forwarding.  I also think some of the representations might need to vary here (e.g. I kinda think SerializedStoredFileInfo shouldn't have a `path` at all, or if it does it should always be relative.  Whereas the one in the StoredFileInfo is sometimes absolute.)

I also considered making StoredFileInfo inherit-from SerializedStoreFileInfo or an additional shared pydantic model for the duplicated fields, but inheritance seems harder to understand than duplicating 5 string properties.  And if we add more properties, they will likely be non-nullable on StoredFileInfo but will have to be nullable on the serialized one to handle backwards compatibility.

I do think that the methods on StoredFileInfo could be moved to FileDatastore and StoredFileInfo could become an actual dataclass.  Or like if we were using the SqlAlchemy ORM layer StoredFileInfo could just be the ORM object.  But I know you guys had plans with exposing these more places which is why you added the abstract base class, so to my mind that kind of change is way outside the scope of this ticket.

python/lsst/daf/butler/datastores/fileDatastoreClient.py
dhirving: Opened DM-42066

python/lsst/daf/butler/datastores/fileDatastore.py
dhirving: :shrug: to me it's equally clear to have a small function explicitly declaring what its inputs and outputs are, but that might be the typescript programmer in me talking.

python/lsst/daf/butler/remote_butler/server/handlers/_external.py
dhirving: adding the dataset ID to the exception message so it ends up in the logs is a good idea, will do

python/lsst/daf/butler/datastores/fileDatastore.py
timj: A timing test in ipython hints that calling the helper function takes a third longer than doing the loop directly. It seems a lot of overhead for something that is simpler without it (and there are still type annotations you can add to the list above to make it clear.

PR #914, opened 2023-11-30
DM-41761: Add Butler.query property

andy-slac:
This new property defines an interface for making complex queries
on repository database. This initial commit includes only three
convenience methods that are a close match  to the query methods that
already exist in Registry. The implementation of that interface
for DirectButler just forwards calls to SqlRegistry.

RemoteButler does not implement those methods yet, we will need
to define representation for queries and results that client exchanges
server. This should become clearer as we add more methods to `Query`.

## Checklist

- [ ] ran Jenkins


python/lsst/daf/butler/_query.py
```diff
@@ -0,0 +1,267 @@
...
+from collections.abc import Iterable
+from typing import TYPE_CHECKING, Any, Mapping
+
+if TYPE_CHECKING:
+    from ._query_results import DataCoordinateQueryResults, DatasetQueryResults, DimensionRecordQueryResults
+    from .dimensions import DataId, DimensionElement, DimensionGroup
+    from .registry._registry import CollectionArgType
+
+
+class Query(ABC):
```
TallJimbo, line 42: As I mentioned at the meeting, I think we'll want to make this a concrete class with a polymorphic implementation object of some kind that's specialized on direct vs. remote butler; something like:
```
class Query:
    _relation_tree: StuffJimIsPrototypingNow  # description of the query to run, so far
    _result_rows: list[dict[ColumnTag, Any]] | None  # results from running the query, if we have them
    _backend: QueryBackend  # thing that knows how to execute _relation_tree and get back _result_rows
```
The concrete `Query` methods would then frequently return a new instance with a modified copy of `_relation_tree` and the same `_backend`, which and that wouldn't depend on whether this is a remote or direct butler.

Given that we don't actually have `StuffJimIsPrototypingNow` or a sense for what the `QueryBackend` interface would be, it might be premature to make that switch.  But I do think we'll want eventually to avoid a lot of duplication.

python/lsst/daf/butler/_query.py
```diff
@@ -0,0 +1,267 @@
...
+        """Query for data IDs matching user-provided criteria.
+
+        Parameters
+        ----------
+        dimensions : `DimensionGroup`, `Dimension`, or `str`, or \
+                `~collections.abc.Iterable` [ `Dimension` or `str` ]
+            The dimensions of the data IDs to yield, as either `Dimension`
+            instances or `str`.  Will be automatically expanded to a complete
+            `DimensionGroup`.  Support for `Dimension` instances is deprecated
+            and will not be supported after v27.
```
TallJimbo: Looks like you can drop mention of `Dimension` instances here, at least according to the type annotation (and I endorse not putting the backwards compatibility into these new APIs).

python/lsst/daf/butler/_query.py
```diff
@@ -0,0 +1,267 @@
...
+            dimensions or values, or when they contain inconsistent values.
+        lsst.daf.butler.registry.UserExpressionError
+            Raised when ``where`` expression is invalid.
+
+        Notes
+        -----
+        When multiple dataset types are queried in a single call, the
+        results of this operation are equivalent to querying for each dataset
+        type separately in turn, and no information about the relationships
+        between datasets of different types is included.  In contexts where
```
TallJimbo: Let's drop the rest of the note after `included.`, since it references `Registry` methods that are not going to remain the preferred way of doing things, and the new way has fully materialized yet.

python/lsst/daf/butler/_query.py
```diff
@@ -0,0 +1,267 @@
...
+        desired dataset types and collections passed as constraints to the
+        query), and then use multiple (generally much simpler) calls to
+        `queryDatasets` with the returned data IDs passed as constraints.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def dimension_records(
+        self,
+        element: DimensionElement | str,
```
TallJimbo: Let's just accept `str` here, and continue getting `DimensionElement` instances out of the public API.

python/lsst/daf/butler/_query_results.py
TallJimbo, line 1: I'd like to start with a cleaner slate on the result objects for the new system, including:
- no `DimensionGraph` backwards compatibility
- no camelCase
- no inheritance from `DataCoordinateIterable` (which I think we might want to replace rather than modify on DM-41114)
- Maybe a rethinking of `order_by` and `limit`?  Do we want more Pythonic `.sorted()` or `[begin:end]` syntax?  Or do we want to combine them into one method, so there's no ambiguity about `result.limit(...).sorted(...)` - in fact, interactions with `subset(unique=True)` are also tricky.

I also think some methods like `explain_no_results`, `any`, and `count` can probably go on `Query` itself at some point, since they don't depend on what kind of object we're coercing the result rows into, but that probably doesn't make sense yet.  I think they should probably stay here as well, but it's worth considering whether something like `results.query.any()` is preferable (where `.query` gets you back to the original query object).

We don't have to try to get these right now, but it might be better to make the new `*Results` objects thin shims over the old ones that just don't expose much functionality at all; just like with `RegistryShim`, we could invert that relationship once the new ones can actually do everything the old ones can.

On the other hand, if we're going to only advertise `Butler.query` as experimental initially, maybe this doesn't matter.  I just don't want us to find ourselves writing API-change RFCs in a couple of months for things we're only adding now.

TallJimbo (commented):
Only about halfway through and will have more comments tomorrow, but I figured it might be useful to publish some big-picture thoughts now.

python/lsst/daf/butler/_query_results.py
TallJimbo: One more reason for a clean slate: the dataset-query followups, `subset`, and `materialize` are all pretty advanced usage, and I could imagine them being better expressed using other calls on the `Query` object, and if we do that we may want to limit the number of ways to do things by not allowing them here.

Basically, I'm imagining that once you've selected the kind of result-object you want (data IDs, dataset refs, dimension records, etc.), you might not be able to call various modifiers that don't depend on the result type _and_ are in the "advanced" usage category.

python/lsst/daf/butler/tests/butler_query.py
```diff
@@ -0,0 +1,1228 @@
...
+            dataIds = tuple(tuple(dataId[k] for k in keys) for dataId in query)
+            self.assertEqual(dataIds, test.result)
+
+            # and materialize
+            query = do_query(keys).order_by(*order_by)
+            if test.limit is not None:
+                query = query.limit(*test.limit)
+            with self.assertRaises(RelationalAlgebraError):
+                with query.materialize():
+                    pass
```
TallJimbo: Making this
```
with query.materialize(): pass
```
seems like it'd silence the spurious code coverage issue here.

python/lsst/daf/butler/tests/butler_query.py
```diff
@@ -0,0 +1,1228 @@
...
+            else:
+                query = do_query(dimensions, test.dataId, test.where, bind=test.bind, **test.kwargs)
+                self.assertEqual(query.count(discard=True), test.count)
+
+            # and materialize
+            if test.exception:
+                with self.assertRaises(test.exception):
+                    query = do_query(dimensions, test.dataId, test.where, bind=test.bind, **test.kwargs)
+                    with query.materialize() as materialized:
+                        materialized.count(discard=True)
```
TallJimbo: Is code coverage telling us that it's actually entering the context manager that's failing here, not the intended test?  Might be worth reordering the context managers to make sure we're checking exactly the line we want.

python/lsst/daf/butler/tests/butler_query.py
```diff
@@ -0,0 +1,1228 @@
...
+        result = butler.query.dimension_records("detector", data_id={"instrument": "Cam1"})
+        self.assertEqual(result.count(), 4)
+        result = butler.query.dimension_records("detector", where="instrument='Cam1'")
+        self.assertEqual(result.count(), 4)
+        result = butler.query.dimension_records("detector", where="instrument=instr", bind={"instr": "Cam1"})
+        self.assertEqual(result.count(), 4)
+
+        with self.assertRaisesRegex(DataIdValueError, "dimension instrument"):
+            result = butler.query.dimension_records("detector", instrument="NotCam1")
+            result.count()
```
TallJimbo: Another case where we might want to move the `with self.assertRaisesRegex` a line lower to make sure the exception is happening where we expect.  Either that or the test is bad.

Lots more of these below.  It's possible the original tests got this wrong, and the behavior we're getting is what we want even if it's not what we're testing for.

TallJimbo (approved):
There are some potential (minor, probably preexisting) issues with which test lines are actually raising our expected exceptions, but otherwise I've got nothing much to say for the second of half of this review.

I've hit "Accepted" because I think everything I'd like to consider changing can be deferred to another ticket _if_ we're marking the `Butler.query` result types as experimental somehow.  But I would like to avoid committing long-term to supporting some aspects of the old result-type interface via `Butler.query`.

python/lsst/daf/butler/_query_results.py
andy-slac: OK, I'll keep these classes in Registry tree and will make new cleaner interfaces for Butler. 

I don't think there is an ambiguity between `limit`/`order_by`, in a sense they should behave similarly to `sqlachemy`. And `subset` should be similar to how query is built using SELECT from subquery (what comes before `.subset`). I think it should be obvious that `query.limit(...).subset(...)` does `LIMIT` before subsetting, and `query.subset(...).limit(...)` does subsetting first and then `LIMIT`? 

And I think I like `order_by` and `limit` as they (almost) match `sqlalchemy` chaining methods.

TallJimbo: I am worried about the following cases:

- User writes `query.limit(1).order_by("visit")` but meant to write `query.order_by("visit").limit(1)`; do we do what they actually wrote (return an arbitrary visit), what we think they should have written, or raise an exception?  I was thinking it'd be better if the interface just didn't let this happen at all.
- User writes `query.order_by("visit").subset(["visit"], unique=True).limit(1)`, not realizing that `subset(..., unique=True)` does not preserve order; again, we have to chose whether to rewrite this for them, execute what's probably a mistake, or raise.

andy-slac: I think that `query.limit(1).order_by("visit")` should do the same as `query.order_by("visit").limit(1)`. Maybe we can produce a warning if we prefer that `order_by` always comes before `limit`? 

The example with `subset` sort of implies that any time when `limit` is used without `order_by` it is a mistake, but I'm not sure that it is always true.

TallJimbo: I agree `limit` without `order_by` is not always a mistake; certainly `limit(1)` without `order_by` is very reasonable.  I do think `offset` without `order_by` might always be a mistake, and I'm not sure about `limit(n > 1)`.


> I think that query.limit(1).order_by("visit") should do the same as query.order_by("visit").limit(1). Maybe we can produce a warning if we prefer that order_by always comes before limit?

I think we can make that work, especially if it's also acceptable to warn and reorder if any other chaining method is called after either of them, so that their operation-tree representation is always at the tip (and I think that makes sense for these particular operations).  If these operations can get embedded deep within the relation-operation tree it gets messy.

python/lsst/daf/butler/tests/butler_query.py
andy-slac: Hmm, `black` is not happy about inlining.

timj: I am not surprised.

timj: You can always do

```python
pass  # pragma: no cover
```
I think.

python/lsst/daf/butler/tests/butler_query.py
andy-slac: It looks like exception is actually happening in `do_query`, so context is not even entered. I'll drop the whole `with` clause.

python/lsst/daf/butler/tests/butler_query.py
andy-slac: Call to `count()` is not really needed to trigger an exception. Maybe things changed since from when the test was originally written. I'll drop all `count` calls.

python/lsst/daf/butler/_query_results.py
```diff
@@ -0,0 +1,749 @@
...
+    single parent `DatasetType`.
+    """
+
+    @property
+    @abstractmethod
+    def components(self) -> Sequence[str | None]:
+        """The components of the parent dataset type included in these results
+        (`~collections.abc.Sequence` [ `str` or `None` ]).
+        """
+        raise NotImplementedError()
```
TallJimbo: I think we can drop this from the new API.  If its equivalent in the old API wasn't deprecated, I think it should have been, as RFC-879 included deprecating all support for components in query APIs.

python/lsst/daf/butler/_butler.py
```diff
@@ -1357,3 +1363,275 @@ def registry(self) -> Registry:
         will be replaced by equivalent `Butler` methods.
         """
         raise NotImplementedError()
+
+    @abstractmethod
+    def query(self) -> AbstractContextManager[Query]:
```
TallJimbo: Can we make this private for now, too?  While I'm not super worried about this method changing in backwards-compatible ways (it might get additional arguments, but they could be default), I do think we will want to tinker with the `Query` class APIs further as we actually implement more kinds of chaining, and it's easier to put the wall here than on every single method there.

TallJimbo (approved):
I would like to make `Butler.query` private (i.e. `Butler._query`) to avoid committing to the `Query` class APIs (and the APIs of its helper classes); I just think we probably want to keep all of these APIs private until we're moderately far along writing the `RemoteButler` versions, as I could see unexpected difficulties there shaking things up. 

I think the `Butler` convenience methods are probably closest to being ready for public consumption - the fact that those do not return objects capable of follow-up queries or lazy iteration makes me pretty confident `RemoteButler` won't throw a wrench in their works, but it's still conceivable we'd want to make small adjustments in the future to make them more similar to corresponding advanced-interface methods that _do_ need to be adjusted due to `RemoteButler` considerations.

But while I want to be very cautious about rolling this out because I think we _might_ want to change thigns, I don't actually see anything here that differs enough from how I'd imagined doing it for me to have recommendations to change more now, so I think we put up that last "privacy wall" and proceed.

python/lsst/daf/butler/_query_results.py
andy-slac: `Registry.queryDatasets`'s `component` parameter is documented as deprecated, I guess that makes that attribute deprecated too. I'll drop it in the new API and I can explicitly document it as deprecated in old API.

python/lsst/daf/butler/_butler.py
andy-slac: I guess I should also remove news fragment as we are hiding new API from users.

timj: Keep it in with the underscores and when I make the release notes I can decide what to do with it.

PR #915, opened 2023-11-30
DM-41158: (mostly) implement QueryDriver for DirectButler

TallJimbo:
I've inspected all of the missed-coverage lines and they fall into two categories:
 - they need `DatasetRef` or `DataCoordinate` result support to be tested (which I'm deferring to another ticket);
 - they're really trivial error cases that I don't really think are worth the effort.

## Checklist

- [x] ran Jenkins
- [ ] ~added a release note for user-visible changes to `doc/changes`~ (N/A)

### DIFF
--- a/python/lsst/daf/butler/column_spec.py
+++ b/python/lsst/daf/butler/column_spec.py
@@ -53,6 +53,7 @@ from lsst.sphgeom import Region
 
 from . import arrow_utils, ddl
 from ._timespan import Timespan
+from .name_shrinker import NameShrinker
 
 ColumnType: TypeAlias = Literal[
     "int", "string", "hash", "float", "datetime", "bool", "uuid", "timespan", "region"
@@ -79,11 +80,14 @@ class _BaseColumnSpec(pydantic.BaseModel, ABC):
         description="Whether the column may be ``NULL``.",
     )
 
-    def to_sql_spec(self, **kwargs: Any) -> ddl.FieldSpec:
+    def to_sql_spec(self, name_shrinker: NameShrinker | None = None, **kwargs: Any) -> ddl.FieldSpec:
         """Convert this specification to a SQL-specific one.
 
         Parameters
         ----------
+        name_shrinker : `NameShrinker`, optional
+            Object that should be used to shrink the field name to ensure it
+            fits within database-specific limits.
         **kwargs
             Forwarded to `ddl.FieldSpec`.
 
@@ -92,7 +96,10 @@ class _BaseColumnSpec(pydantic.BaseModel, ABC):
         sql_spec : `ddl.FieldSpec`
             A SQL-specific version of this specification.
         """
-        return ddl.FieldSpec(name=self.name, dtype=ddl.VALID_CONFIG_COLUMN_TYPES[self.type], **kwargs)
+        name = self.name
+        if name_shrinker is not None:
+            name = name_shrinker.shrink(name)
+        return ddl.FieldSpec(name=name, dtype=ddl.VALID_CONFIG_COLUMN_TYPES[self.type], **kwargs)
 
     @abstractmethod
     def to_arrow(self) -> arrow_utils.ToArrow:
@@ -162,9 +169,9 @@ class StringColumnSpec(_BaseColumnSpec):
     length: int
     """Maximum length of strings."""
 
-    def to_sql_spec(self, **kwargs: Any) -> ddl.FieldSpec:
+    def to_sql_spec(self, name_shrinker: NameShrinker | None = None, **kwargs: Any) -> ddl.FieldSpec:
         # Docstring inherited.
-        return super().to_sql_spec(length=self.length, **kwargs)
+        return super().to_sql_spec(length=self.length, name_shrinker=name_shrinker, **kwargs)
 
     def to_arrow(self) -> arrow_utils.ToArrow:
         # Docstring inherited.
@@ -182,9 +189,9 @@ class HashColumnSpec(_BaseColumnSpec):
     nbytes: int
     """Number of bytes for the hash."""
 
-    def to_sql_spec(self, **kwargs: Any) -> ddl.FieldSpec:
+    def to_sql_spec(self, name_shrinker: NameShrinker | None = None, **kwargs: Any) -> ddl.FieldSpec:
         # Docstring inherited.
-        return super().to_sql_spec(nbytes=self.nbytes, **kwargs)
+        return super().to_sql_spec(nbytes=self.nbytes, name_shrinker=name_shrinker, **kwargs)
 
     def to_arrow(self) -> arrow_utils.ToArrow:
         # Docstring inherited.
--- a/python/lsst/daf/butler/direct_butler.py
+++ b/python/lsst/daf/butler/direct_butler.py
@@ -2141,8 +2141,9 @@ class DirectButler(Butler):  # numpydoc ignore=PR02
             self._registry._db, self.dimensions, self._registry._managers, self._registry.defaults
         )
         query = Query(driver)
-        with driver:
-            yield query
+        with self._caching_context():
+            with driver:
+                yield query
 
     def _preload_cache(self) -> None:
         """Immediately load caches that are used for common operations."""
--- a/python/lsst/daf/butler/direct_butler.py
+++ b/python/lsst/daf/butler/direct_butler.py
@@ -66,6 +66,7 @@ from ._storage_class import StorageClass, StorageClassFactory
 from ._timespan import Timespan
 from .datastore import Datastore, NullDatastore
 from .dimensions import DataCoordinate, Dimension
+from .direct_query_driver import DirectQueryDriver
 from .progress import Progress
 from .queries import Query
 from .registry import (
@@ -2136,7 +2137,12 @@ class DirectButler(Butler):  # numpydoc ignore=PR02
     @contextlib.contextmanager
     def _query(self) -> Iterator[Query]:
         # Docstring inherited.
-        raise NotImplementedError("TODO DM-41159")
+        driver = DirectQueryDriver(
+            self._registry._db, self.dimensions, self._registry._managers, self._registry.defaults
+        )
+        query = Query(driver)
+        with driver:
+            yield query
 
     def _preload_cache(self) -> None:
         """Immediately load caches that are used for common operations."""
--- a/python/lsst/daf/butler/direct_query_driver/_driver.py
+++ b/python/lsst/daf/butler/direct_query_driver/_driver.py
@@ -145,10 +145,8 @@ class DirectQueryDriver(QueryDriver):
         # transaction starts safe against deadlocks, and it'd be more
         # complicated to implement anyway.
         #
-        # We start a transaction rather than just opening a connection both
-        # to establish consistency guarantees throughout the query context and
-        # to make temp table and cursors work with pg_bouncer transaction
-        # affinity.
+        # We start a transaction rather than just opening a connection to make
+        # temp table and cursors work with pg_bouncer transaction affinity.
         self._exit_stack.enter_context(self.db.transaction(for_temp_tables=True))
 
     def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
--- a/python/lsst/daf/butler/direct_query_driver/_driver.py
+++ b/python/lsst/daf/butler/direct_query_driver/_driver.py
@@ -217,10 +217,7 @@ class DirectQueryDriver(QueryDriver):
             raw_page_size=raw_page_size,
         )
         result_page = cursor.next()
-        if result_page.next_key is None:
-            # Cursor has been exhausted; close it here.
-            cursor.close()
-        else:
+        if result_page.next_key is not None:
             # Cursor has not been exhausted; add it to the driver for use by
             # fetch_next_page.
             self._cursors[result_page.next_key] = cursor
--- a/python/lsst/daf/butler/direct_query_driver/_driver.py
+++ b/python/lsst/daf/butler/direct_query_driver/_driver.py
@@ -31,6 +31,7 @@ import uuid
 
 __all__ = ("DirectQueryDriver",)
 
+import dataclasses
 import logging
 import sys
 from collections.abc import Iterable, Mapping, Set
@@ -124,9 +125,7 @@ class DirectQueryDriver(QueryDriver):
         self.managers = managers
         self._universe = universe
         self._defaults = defaults
-        self._materializations: dict[
-            qt.MaterializationKey, tuple[sqlalchemy.Table, frozenset[str], Postprocessing]
-        ] = {}
+        self._materializations: dict[qt.MaterializationKey, _MaterializationState] = {}
         self._upload_tables: dict[qt.DataCoordinateUploadKey, sqlalchemy.FromClause] = {}
         self._exit_stack: ExitStack | None = None
         self._raw_page_size = raw_page_size
@@ -281,7 +280,7 @@ class DirectQueryDriver(QueryDriver):
         table = self._exit_stack.enter_context(self.db.temporary_table(builder.make_table_spec()))
         self.db.insert(table, select=sql_select)
         key = uuid.uuid4()
-        self._materializations[key] = (table, datasets, builder.postprocessing)
+        self._materializations[key] = _MaterializationState(table, datasets, builder.postprocessing)
         return key
 
     def upload_data_coordinates(
@@ -864,15 +863,19 @@ class DirectQueryDriver(QueryDriver):
                     )
         # Add materializations, which can also bring in more postprocessing.
         for m_key, m_dimensions in tree.materializations.items():
-            _, _, m_postprocessing = self._materializations[m_key]
+            m_state = self._materializations[m_key]
             result.materializations[m_key] = m_dimensions
             # When a query is materialized, the new tree has an empty
             # (trivially true) predicate because the original was used to make
             # the materialized rows.  But the original postprocessing isn't
             # executed when the materialization happens, so we have to include
             # it here.
-            builder.postprocessing.spatial_join_filtering.extend(m_postprocessing.spatial_join_filtering)
-            builder.postprocessing.spatial_where_filtering.extend(m_postprocessing.spatial_where_filtering)
+            builder.postprocessing.spatial_join_filtering.extend(
+                m_state.postprocessing.spatial_join_filtering
+            )
+            builder.postprocessing.spatial_where_filtering.extend(
+                m_state.postprocessing.spatial_where_filtering
+            )
         # Add data coordinate uploads.
         result.data_coordinate_uploads.update(tree.data_coordinate_uploads)
         # Add dataset_searches and filter out collections that don't have the
@@ -1010,9 +1013,9 @@ class DirectQueryDriver(QueryDriver):
             materialization was created.
         """
         columns = qt.ColumnSet(dimensions)
-        table, datasets, postprocessing = self._materializations[key]
-        joiner.join(QueryJoiner(self.db, table).extract_columns(columns, postprocessing))
-        return datasets
+        m_state = self._materializations[key]
+        joiner.join(QueryJoiner(self.db, m_state.table).extract_columns(columns, m_state.postprocessing))
+        return m_state.datasets
 
     def _join_dataset_search(
         self,
@@ -1042,6 +1045,13 @@ class DirectQueryDriver(QueryDriver):
             resolved_search.name not in joiner.timespans
         ), "Dataset timespan has unexpectedly already been joined in."
         joiner.join(storage.make_query_joiner(resolved_search.collection_records, fields))
+
+
+@dataclasses.dataclass
+class _MaterializationState:
+    table: sqlalchemy.Table
+    datasets: frozenset[str]
+    postprocessing: Postprocessing
 
 
 class _Cursor:
--- a/python/lsst/daf/butler/direct_query_driver/_driver.py
+++ b/python/lsst/daf/butler/direct_query_driver/_driver.py
@@ -1,0 +1,1202 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This software is dual licensed under the GNU General Public License and also
+# under a 3-clause BSD license. Recipients may choose which of these licenses
+# to use; please see the files gpl-3.0.txt and/or bsd_license.txt,
+# respectively.  If you choose the GPL option then the following text applies
+# (but note that there is still no warranty even if you opt for BSD instead):
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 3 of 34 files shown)


python/lsst/daf/butler/column_spec.py
```diff
@@ -79,11 +80,14 @@ class _BaseColumnSpec(pydantic.BaseModel, ABC):
         description="Whether the column may be ``NULL``.",
     )
 
-    def to_sql_spec(self, **kwargs: Any) -> ddl.FieldSpec:
+    def to_sql_spec(self, name_shrinker: NameShrinker | None = None, **kwargs: Any) -> ddl.FieldSpec:
```
dhirving, line 83: Should this nameshrinker argument be non-optional?  I would think that if it is ever needed, it will always be needed... otherwise different parts of the code can think names are different.  Just creates the potential for certain columns to work with some types of queries/operations but not others.

python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -0,0 +1,1200 @@
...
+        # to reduce the time spent in transactions.  But it's worth noting that
+        # this is the default low-level behavior of the Python SQLite driver,
+        # and it makes it incredibly prone to deadlocks.  We might be okay
+        # here, because Query doesn't do true write operations - just temp
+        # table writes - but I'm not confident that's enough to make delayed
+        # transaction starts safe against deadlocks, and it'd be more
+        # complicated to implement anyway.
+        #
+        # We start a transaction rather than just opening a connection both
+        # to establish consistency guarantees throughout the query context and
```
dhirving: What consistency guarantees do we get from the transaction?  I thought Butler transactions were always READ COMMITTED, which is the same as what we'd get without the explicit transaction.

tests/test_query_direct_sqlite.py
```diff
@@ -0,0 +1,65 @@
...
+from lsst.daf.butler import Butler, ButlerConfig, StorageClassFactory
+from lsst.daf.butler.datastore import NullDatastore
+from lsst.daf.butler.direct_butler import DirectButler
+from lsst.daf.butler.registry.sql_registry import SqlRegistry
+from lsst.daf.butler.tests.butler_queries import ButlerQueryTests
+
+TESTDIR = os.path.abspath(os.path.dirname(__file__))
+
+
+class DirectButlerSQLiteTests(ButlerQueryTests, unittest.TestCase):
```
dhirving, line 45: Should there be another instance of these tests for Postgres as well?

python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -0,0 +1,1200 @@
...
+        raw_page_size: int = 10000,
+        constant_rows_limit: int = 1000,
+        postprocessing_filter_factor: int = 10,
+    ):
+        self.db = db
+        self.managers = managers
+        self._universe = universe
+        self._defaults = defaults
+        self._materializations: dict[
+            qt.MaterializationKey, tuple[sqlalchemy.Table, frozenset[str], Postprocessing]
```
dhirving: Making this tuple a NamedTuple or frozen dataclass would make it self-documenting and less fiddly to interact with.

python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -0,0 +1,1200 @@
...
+        # to establish consistency guarantees throughout the query context and
+        # to make temp table and cursors work with pg_bouncer transaction
+        # affinity.
+        self._exit_stack.enter_context(self.db.transaction(for_temp_tables=True))
+
+    def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
+        assert self._exit_stack is not None
+        while self._cursors:
+            _, cursor = self._cursors.popitem()
+            cursor.close(exc_type, exc_value, traceback)
```
dhirving, line 157: It might be a good idea to push this cursor closing onto the exit stack instead of directly calling close... that way the rest of cleanup will still occur even if closing a cursor raises an exception.

python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -0,0 +1,1200 @@
...
+            Exception type as obtained from `sys.exc_info`, or `None` if there
+            was no error.
+        exc_value : `BaseException` or `None`
+            Exception instance as obtained from `sys.exc_info`, or `None` if
+            there was no error.
+        traceback : `object`
+            Traceback as obtained from `sys.exc_info`, or `None` if there was
+            no error.
+        """
+        self._context.__exit__(exc_type, exc_value, traceback)
```
dhirving, line 1118: Is it allowed to call `__exit__` on a SQLAlchemy cursor multiple times?  If not it might be a good idea to guard against that here.

python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -0,0 +1,1200 @@
...
+            sql_select,
+            result_spec=result_spec,
+            name_shrinker=builder.joiner.name_shrinker,
+            postprocessing=builder.postprocessing,
+            raw_page_size=raw_page_size,
+        )
+        result_page = cursor.next()
+        if result_page.next_key is None:
+            # Cursor has been exhausted; close it here.
+            cursor.close()
```
dhirving: This cursor closing logic is already handled inside `cursor.next()`.

python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -0,0 +1,1200 @@
...
+                    record_set.add(record_cls(**{k: m[v] for k, v in column_map}))
+            else:
+                timespan_qualified_name = qt.ColumnSet.get_qualified_name(
+                    result_spec.element.name, "timespan"
+                )
+                for raw_row in raw_rows:
+                    m = raw_row._mapping
+                    d = {k: m[v] for k, v in column_map}
+                    d["timespan"] = self._timespan_repr_cls.extract(m, name=timespan_qualified_name)
+                    record_set.add(record_cls(**d))
```
dhirving: It might be better not to duplicate the logic for the non-timespan columns between these two branches... creates the potential for them to get out of sync as things change.
```suggestion
            if result_spec.element.temporal:
                timespan_qualified_name = qt.ColumnSet.get_qualified_name(
                    result_spec.element.name, "timespan"
                )
            else:
                timespan_qualified_name = None
                
            for raw_row in raw_rows:
                m = raw_row._mapping
                d = {k: m[v] for k, v in column_map}
                if timespan_qualified_name is not None:
                    d["timespan"] = self._timespan_repr_cls.extract(m, name=timespan_qualified_name)
                record_set.add(record_cls(**d))
```

python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -0,0 +1,1200 @@
...
+    def upload_data_coordinates(
+        self, dimensions: DimensionGroup, rows: Iterable[tuple[DataIdValue, ...]]
+    ) -> qt.DataCoordinateUploadKey:
+        # Docstring inherited.
+        if self._exit_stack is None:
+            raise RuntimeError(
+                "QueryDriver context must be entered before 'upload_data_coordinates' is called."
+            )
+        columns = qt.ColumnSet(dimensions).drop_implied_dimension_keys()
+        table_spec = ddl.TableSpec(
```
dhirving, line 292: For maintainability it might be a good idea to deduplicate this ColumnSet-to-TableSpec logic between here and `QueryBuilder.make_table_spec()`... and the one in QueryBuilder has nameshrinker logic that might also be needed here.

python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -0,0 +1,1200 @@
...
+                    QueryBuilder.EMPTY_COLUMNS_NAME, dtype=QueryBuilder.EMPTY_COLUMNS_TYPE, nullable=True
+                )
+            )
+            dict_rows = [{QueryBuilder.EMPTY_COLUMNS_NAME: None}]
+        else:
+            dict_rows = [dict(zip(dimensions.required, values)) for values in rows]
+        from_clause: sqlalchemy.FromClause
+        if len(dict_rows) > self._constant_rows_limit:
+            from_clause = self._exit_stack.enter_context(self.db.temporary_table(table_spec))
+            self.db.insert(from_clause, *dict_rows)
```
dhirving, line 308: The Postgres documentation [recommends running ANALYZE](https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-TEMPORARY) on temp tables if they are going to be used in complex queries... I wonder if that is worth trying here and in materialize() above, once we get to a point where we can benchmark performance.

python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -0,0 +1,1200 @@
...
+        # particularly easy mistake to make and it's almost never intentional.
+        # We also allow the registry data ID values to provide governor values.
+        where_columns = qt.ColumnSet(self.universe.empty.as_group())
+        result.predicate.gather_required_columns(where_columns)
+        for governor in where_columns.dimensions.governors:
+            if governor not in result.constraint_data_id:
+                if governor in self._defaults.dataId.dimensions:
+                    result.constraint_data_id[governor] = self._defaults.dataId[governor]
+                else:
+                    raise qt.InvalidQueryError(
```
dhirving, line 857: We're going to want any user-facing errors like this to be raised as a subclass of ButlerUserError in `lsst.daf.butler._exceptions`, so that they can be transferred to the client.

I can fix up InvalidQueryError to make it a ButlerUserError when I go in to do the server version of this, but if there are other exception types getting raised we'll want to keep track of that.  (The RuntimeErrors in this file are fine as-is... those are not the kind of thing that needs to go from server to client.)

python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -0,0 +1,1200 @@
...
+        columns = qt.ColumnSet(dimensions)
+        table, datasets, postprocessing = self._materializations[key]
+        joiner.join(QueryJoiner(self.db, table).extract_columns(columns, postprocessing))
+        return datasets
+
+    def _join_dataset_search(
+        self,
+        joiner: QueryJoiner,
+        resolved_search: ResolvedDatasetSearch,
+        fields: Set[str],
```
dhirving, line 1021: It would help clarify what is going on if instead of `str` this used the `DatasetFieldName` type alias from `query/tree/_base.py`.

(Also in `DatasetRecordStorage.make_query_joiner` and `ColumnSet.dataset_fields`.)

python/lsst/daf/butler/direct_butler.py
```diff
@@ -2135,7 +2136,12 @@ def dimensions(self) -> DimensionUniverse:
     @contextlib.contextmanager
     def _query(self) -> Iterator[Query]:
         # Docstring inherited.
-        raise NotImplementedError("TODO DM-41159")
+        driver = DirectQueryDriver(
+            self._registry._db, self.dimensions, self._registry._managers, self._registry.defaults
+        )
+        query = Query(driver)
```
dhirving, line 2143: I wonder if we also want to call `with self._caching_context` here to enable caching of collection info.

python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -0,0 +1,1200 @@
...
+                if collection_name not in done:
+                    done.add(collection_name)
+                    record = self.managers.collections.find(collection_name)
+
+                    if record.type is CollectionType.CHAINED:
+                        recurse(cast(ChainedCollectionRecord, record).children)
+                    else:
+                        result.append((record, self.managers.datasets.getCollectionSummary(record)))
+
+        recurse(collections)
```
dhirving, line 983: I think this same operation is already implemented with somewhat fewer queries at `self.managers.datasets.fetch_summaries`.  The keys of the returned dict aren't CollectionRecord but it could easily be changed so it is -- it's operating on CollectionRecord internally.

python/lsst/daf/butler/direct_query_driver/_query_builder.py
```diff
@@ -0,0 +1,573 @@
...
+    either a dimension element name or dataset type name.
+    """
+
+    timespans: dict[str, TimespanDatabaseRepresentation] = dataclasses.field(default_factory=dict)
+    """Mapping of timespan columns.
+
+    Keys are "logical tables" - dimension element names or dataset type names.
+    """
+
+    special: dict[str, sqlalchemy.ColumnElement[Any]] = dataclasses.field(default_factory=dict)
```
dhirving, line 362: I'm slightly concerned about the way assignments are made to this throughout the code... seems like there is opportunity for one chunk of code to accidentally clobber something added by another chunk.

I didn't spot any cases where that is actually occurring, though.

python/lsst/daf/butler/direct_query_driver/_sql_column_visitor.py
```diff
@@ -0,0 +1,263 @@
...
+        self, original: qt.PredicateLeaf, result: sqlalchemy.ColumnElement[bool], flags: PredicateVisitFlags
+    ) -> sqlalchemy.ColumnElement[bool]:
+        # Docstring inherited.
+        return sqlalchemy.not_(result)
+
+    def expect_scalar(self, expression: qt.OrderExpression) -> sqlalchemy.ColumnElement[Any]:
+        return cast(sqlalchemy.ColumnElement[Any], expression.visit(self))
+
+    def expect_timespan(self, expression: qt.ColumnExpression) -> TimespanDatabaseRepresentation:
+        return cast(TimespanDatabaseRepresentation, expression.visit(self))
```
dhirving: You might consider adding instanceof asserts to these `expect_` functions instead of casting... might save some debugging time if something of an unexpected type creeps in.

python/lsst/daf/butler/direct_query_driver/_query_builder.py
```diff
@@ -0,0 +1,573 @@
...
+    queries that have no real columns.
+    """
+
+    EMPTY_COLUMNS_TYPE: ClassVar[type] = sqlalchemy.Boolean
+    """Type of the column added to a SQL SELECT clause in order to construct
+    queries that have no real columns.
+    """
+
+    def __post_init__(self) -> None:
+        assert not (self.distinct and self.group_by), "At most one of distinct and group_by can be set."
```
dhirving: This check should probably be in `select()` instead of post_init... most of the assignments to distinct and group_by occur after initialization.

python/lsst/daf/butler/direct_query_driver/_query_builder.py
```diff
@@ -0,0 +1,573 @@
...
+    """Special columns that are available from the FROM clause and
+    automatically included in the SELECT clause when this joiner is nested
+    within a `QueryBuilder`.
+
+    These columns are not part of the dimension universe and are not associated
+    with a dataset.  They are never returned to users, even if they may be
+    included in raw SQL results.
+    """
+
+    name_shrinker: NameShrinker | None = None
```
dhirving, line 372: I wonder if it would be easier to add a `name_shrinker` property to `Database` instead of worrying about creating them and transferring records between them here.

dhirving (approved):
Looks good to me, just a bunch of nitpicks you can take or leave.  (Like last time GitHub is hiding some of the comments because there are too many.)

I really appreciated your extensive commenting in the hairier parts of query construction... I'm not going to claim that I fully understand all of the details but definitely closer than I was before.

Looks like it will be pretty easy to wrap the client/server version around this.

python/lsst/daf/butler/column_spec.py
TallJimbo: I prefer to do that at a higher level, because the set of names that need shrinking is limited to only those constructed from dataset type names, and that keeps name-shrinker arguments from sprouting up in many more places.

python/lsst/daf/butler/direct_butler.py
TallJimbo: Yes, good idea.  Long term I think this can probably be the only way a caching context gets entered, which will hopefully let us get the caching out of the managers entirely, but entering the current caching context here is what makes sense right now.

python/lsst/daf/butler/direct_query_driver/_driver.py
TallJimbo: Yeah, there was some lazy thinking going on here; I had been mostly thinking about consistency of things in caches (which you reminded me I hadn't actually done in another thread) and started conflating that with actual transaction consistency (which as you note here depends on isolation level).  I'll drop the first clause of this sentence and just keep the second.

python/lsst/daf/butler/direct_query_driver/_driver.py
TallJimbo: I tried something like that first, but got stuck being unable to pop a cursor out of the exit stack by name when it could be closed due to normal termination of iteration.  But you're right that there's a resource management gap here.  Adding a manual try/except here is nontrivial because we do want to re-raise at the end, so I'll punt this to [DM-43697](https://rubinobs.atlassian.net/browse/DM-43697).

dhirving:
I'd like to start working on the client/server version of this early next week, while it's still fresh in my mind.  There's absolutely nothing in these review comments above the level of a petty nitpick.  So since you've got a lot on your plate at the moment I think it would be fine to merge this down as-is.

python/lsst/daf/butler/direct_query_driver/_driver.py
TallJimbo: I'll look into this on [DM-43698](https://rubinobs.atlassian.net/browse/DM-43698) in order to get this branch merged faster.  I suspect  method on `ColumnSet` would work well.

python/lsst/daf/butler/direct_query_driver/_driver.py
TallJimbo: I was not aware of that recommendation, but I did think a bit about whether it would ever be useful to create indexes here, and I came to the conclusion that we probably would not be, because starting with a full table scan of the temp table and using indexes on regular tables is what we want the query optimizer to do in all cases I could imagine.  I have no idea what that says about whether ANALYZE would be helpful.

python/lsst/daf/butler/direct_query_driver/_driver.py
TallJimbo: This is one of a very small number of places where `InvalidQueryError` is raised by code I expect to move to the server; in almost all cases it's raised in client-side code (or both; it's also a model-validation error).

How about we invent a new exception for errors that come up when trying to execute a query, and replace server-side usage of `InvalidQueryError` with that?

python/lsst/daf/butler/direct_query_driver/_driver.py
TallJimbo: I'll look into this on [DM-43698](https://rubinobs.atlassian.net/browse/DM-43698).  We might want to replace both with that recursive CTE at the same time.

python/lsst/daf/butler/direct_query_driver/_driver.py
TallJimbo: Unfortunately those aren't quite the same - there are one or two for-internal-use dataset field names that don't appear in the more user-visibile `DatasetFieldName` type alias.  It would probably make sense to define an expanded type alias for that, though.  Will do on [DM-43698](https://rubinobs.atlassian.net/browse/DM-43698).

python/lsst/daf/butler/direct_query_driver/_driver.py
TallJimbo: It's not _quite_ as clearly stated as I'd like, but [the docs](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.CursorResult.close) say

> This closes out the underlying DBAPI cursor corresponding to the statement execution, if one is still present.

python/lsst/daf/butler/direct_query_driver/_driver.py
TallJimbo: Yeah, this was intentional - but possibly premature - optimization; I wanted to bring as much out of the loop over result rows as possible. But your version is all that different in that respect.

python/lsst/daf/butler/direct_query_driver/_query_builder.py
TallJimbo: Yeah, I really struggled with how much to encapsulate `QueryBuilder` and `QueryJoiner` state throughout this ticket, and after a couple of flip-flops I made myself stop in order to keep making progress overall.  So there's room for improvement, and if you spot a way to improve it sometime while you're working nearby, feel free to give it a try; you may spot a possibility I didn't.

python/lsst/daf/butler/direct_query_driver/_query_builder.py
TallJimbo: I'm not sure; I'd need to look for places where we might want the `NameShrinker` without having the `Database` it came from.  If we don't have any such places (or they're easy to address some other way), this sounds like a good idea.  But one for another ticket (I'll put it on [DM-43697](https://rubinobs.atlassian.net/browse/DM-43697) for now).

tests/test_query_direct_sqlite.py
TallJimbo: Yes, I meant to do that before I put it up for review and apparently I forgot.

But having added it now, I see some failures in cursor/temp-table lifetimes.  I wasn't able to figure it quickly, so I've marked the tests that use temp tables as expected failures and will try to fix them on [DM-43697](https://rubinobs.atlassian.net/browse/DM-43697).

python/lsst/daf/butler/direct_query_driver/_driver.py
dhirving: I was just thinking call `exit_stack.push(cursor)` right here instead of calling `cursor.close()`... `push()` calls `__exit__` without calling `__enter__`.

python/lsst/daf/butler/direct_query_driver/_driver.py
dhirving: It's fine for the same error to be raised either from client-side and server-side, and to me this makes sense as an `InvalidQueryError`.

The main distinction we need to make is between:

"this is a thing that happened because of the arguments the user specified, and the only way to fix it is for them to change their query, or add some more data to the server, or..." -> subclass of ButlerUserError, so it's re-raised on the client side.

vs

"this is a problem with the database server, an internal logic problem, an unexpected assertion failure, etc". -> not a subclass of ButlerUserError, so that the details aren't sent to the client and it triggers alerts for our team to investigate.

python/lsst/daf/butler/direct_query_driver/_query_builder.py
dhirving: The only obvious idea I had was a dict that only allows one write to each key by default, so you at least have to explicitly specify clobbering.  But yeah it's not an easy problem to solve.

python/lsst/daf/butler/direct_query_driver/_driver.py
dhirving: I think mainly it would let the planner know min/max values for the columns, which might affect its decisions on join order etc.  But whether the planner having that knowledge outweighs the cost of the analyze is something we'd have to benchmark probably.

PR #916, opened 2023-11-30
Add Spectrum objects for obs_fiberSpectrograph

aferte:
We are only doing the merge for daf_butler now and will do the obs package later.

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -234,6 +234,8 @@ storageClasses:
     pytype: lsst.ip.isr.CrosstalkCalib
   Linearizer:
     pytype: lsst.ip.isr.Linearizer
+  FiberSpectrum:
+    pytype: lsst.obs.fiberSpectrograph.FiberSpectrum
   PhotonTransferCurveDataset:
     pytype: lsst.ip.isr.PhotonTransferCurveDataset
   StrayLightData:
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -234,6 +234,8 @@ storageClasses:
     pytype: lsst.ip.isr.CrosstalkCalib
   Linearizer:
     pytype: lsst.ip.isr.Linearizer
+  Spectrum:
```
timj: I don't like the name of this. `Spectrum` sounds like a generic class whereas this is specifically a spectrum from the fiber spectrograph, which doesn't sound like it's a Spectrum that a different spectrograph would use. Maybe it is, but then it's probably in the wrong place. Can you please give me some guidance? Otherwise `FiberSpectrum` would be better.

aferte: I see - indeed I believe we only expect this from the fiber spectrograph. Would it be ok to replace it by fiberSpectroData?

aferte: Or rather fiberSpecData as FiberSpec is the name of the instrument in the obs package.

timj: Convention is for upper case first letter.

"Data from the fiber spectrograph" seems like it might mean the raw. That's why I was edging towards `FiberSpectrum` as in spectrum from fiber. `FiberSpecSpectrum` might work if you don't mind doubling up "Spec".

aferte: Sounds good - done in daf_butler and I will update the obs package accordingly later on this week.

timj (approved):

PR #917, opened 2023-12-04
DM-42004: Update the module name for FitsMaskedImage formatter

arunkannawadi:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/datastores/formatters.yaml
+++ b/python/lsst/daf/butler/configs/datastores/formatters.yaml
@@ -40,7 +40,7 @@ ImageU: lsst.obs.base.formatters.fitsExposure.FitsImageFormatter
 DecoratedImageU: lsst.obs.base.formatters.fitsExposure.FitsExposureFormatter
 Mask: lsst.obs.base.formatters.fitsExposure.FitsMaskFormatter
 MaskX: lsst.obs.base.formatters.fitsExposure.FitsMaskFormatter
-MaskedImageF: lsst.obs.base.formatters.fitsMaskedImage.FitsMaskedImageFormatter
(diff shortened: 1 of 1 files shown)


python/lsst/daf/butler/configs/datastores/formatters.yaml
```diff
@@ -40,7 +40,7 @@ ImageU: lsst.obs.base.formatters.fitsExposure.FitsImageFormatter
 DecoratedImageU: lsst.obs.base.formatters.fitsExposure.FitsExposureFormatter
 Mask: lsst.obs.base.formatters.fitsExposure.FitsMaskFormatter
 MaskX: lsst.obs.base.formatters.fitsExposure.FitsMaskFormatter
-MaskedImageF: lsst.obs.base.formatters.fitsMaskedImage.FitsMaskedImageFormatter
+MaskedImageF: lsst.obs.base.formatters.fitsExposure.FitsMaskedImageFormatter
```
TallJimbo, line 43: If you're moving this, the original needs to continue to work (basically forever) since this string is written into the butler database for datasets we've already written.  That makes me wonder if it's actually worthwhile to move it.

arunkannawadi: I'm not moving any code. I don't think this was ever correct, since the formatter had been defined in fitsExposure.py.

https://github.com/lsst/obs_base/blob/main/python/lsst/obs/base/formatters/fitsExposure.py#L26

timj: I'm pretty sure this is fixing a typo.

arunkannawadi: Yup. I should perhaps update the PR title/ticket title?

timj (approved):

PR #920, opened 2023-12-05
DM-41879: Implement URL signing for client/server

dhirving:
Requires lsst/resources#73

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -1998,10 +1998,23 @@ class FileDatastore(GenericBaseDatastore[StoredFileInfo]):
     def prepare_get_for_external_client(self, ref: DatasetRef) -> FileDatastoreGetPayload:
         # Docstring inherited
 
+        # 1 hour.  Chosen somewhat arbitrarily -- this is long enough that the
+        # client should have time to download a large file with retries if
+        # needed, but short enough that it will become obvious quickly that
+        # these URLs expire.
+        # From a strictly technical standpoint there is no reason this
+        # shouldn't be a day or more, but there seems to be a political issue
+        # where people think there is a risk of end users posting presigned
+        # URLs for people without access rights to download.
+        url_expiration_time_seconds = 1 * 60 * 60
+
         def to_file_info_payload(info: DatasetLocationInformation) -> FileDatastoreGetPayloadFileInfo:
             location, file_info = info
             return FileDatastoreGetPayloadFileInfo(
-                url=location.uri.geturl(), datastoreRecords=file_info.to_simple()
+                url=location.uri.generate_presigned_get_url(
+                    expiration_time_seconds=url_expiration_time_seconds
+                ),
+                datastoreRecords=file_info.to_simple(),
             )
 
         return FileDatastoreGetPayload(
--- a/python/lsst/daf/butler/datastores/file_datastore/get.py
+++ b/python/lsst/daf/butler/datastores/file_datastore/get.py
@@ -207,13 +207,14 @@ def _read_artifact_into_memory(
     # Do not do this if the size is negative since that indicates
     # we do not know.
     recorded_size = getInfo.info.file_size
-    resource_size = uri.size()
-    if recorded_size >= 0 and resource_size != recorded_size:
-        raise RuntimeError(
-            "Integrity failure in Datastore. "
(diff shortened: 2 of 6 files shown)


python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1998,10 +1998,19 @@ def get(
     def prepare_get_for_external_client(self, ref: DatasetRef) -> FileDatastoreGetPayload:
         # Docstring inherited
 
+        # 8 hours.  Chosen somewhat arbitrarily -- this is long enough that the
```
timj: Last time we talked with Russ in Princeton we decided that a few minutes was reasonable here since the signing should only happen close to the getting (and we wouldn't be automatically signing the results from query.datasets()) and the server shouldn't be returning URLs that live long enough for people to post on social media. If the getting of a single dataset is taking hours then we have bigger problems (and I was under the impression that once the download starts it wouldn't cut off the download if it passed expiration but that may be misunderstanding how chunked downloads would work)

python/lsst/daf/butler/datastores/file_datastore/get.py
```diff
@@ -225,7 +226,7 @@ def _read_artifact_into_memory(
 
     formatter = getInfo.formatter
     nbytes_max = 10_000_000  # Arbitrary number that we can tune
-    if resource_size <= nbytes_max and formatter.can_read_bytes():
+    if recorded_size >= 0 and recorded_size <= nbytes_max and formatter.can_read_bytes():
```
timj, line 229: Good catch. The size of -1 for "unknown" was added long after this check was written.

python/lsst/daf/butler/tests/utils.py
```diff
@@ -311,4 +315,6 @@ def addDataset(
         if run:
             self.butler.registry.registerCollection(run, type=CollectionType.RUN)
         metric = self._makeExampleMetrics()
-        self.butler.put(metric, self.datasetType if datasetType is None else datasetType, dataId, run=run)
+        return self.butler.put(
```
timj, line 326: Please add a Returns section to the docstring.

tests/test_server.py
```diff
@@ -51,6 +52,7 @@
     removeTestTempDir,
 )
 from lsst.resources.http import HttpResourcePath
+from lsst.resources.s3utils import clean_test_environment_for_s3, getS3Client
```
timj: S3 dependencies are not installed by default in the pyproject.toml definition so if server testing is now required to have boto3 available this is going to have to have import protection so the tests can be skipped.

tests/test_server.py
```diff
@@ -227,5 +249,14 @@ def test_get(self):
         self.assertEqual(metric, converted)
 
 
+def _create_corrupted_dataset(repo: MetricTestRepo) -> DatasetRef:
+    run = "corrupted-run"
+    ref = repo.addDataset({"instrument": "DummyCamComp", "visit": 423}, run=run)
+    uris = repo.butler.getURIs(ref, run=run)
```
timj: I'm hoping that the `run` parameter is not needed here because the `ref` knows its own run.

timj (approved):
Some nice clean ups. I agree that checking the file size after download is fine since that's the case that should always happen and the case of a bad size is meant to be very rare (so the cost of downloading the file and immediately raising an exception should not be burdensome).

python/lsst/daf/butler/datastores/fileDatastore.py
dhirving: It won't cut off the download, but if you need to retry at the HTTP layer it will need the signature to remain valid.  Its nice if you can allow HTTP retries without needing to throw an exception all the way up to Butler.get() to re-do the API request too.

Or e.g. for multiple range requests if it does some slow work in between reading the chunks, the signature will need to stay valid for the duration of the work.

Also we had discussed like a get_many(), so the download might not start immediately if there are other files in queue on a slow link.

I would argue that a few minutes isn't really long enough.  10-20 minutes is probably OK, but I think it's best if we can make this the longest possible length that isn't a political problem.  How would you feel about 1-2 hours? I think it's best to have timeouts one or two orders of magnitude higher than the worst expected typical case, so things don't blow up immediately when the system is under stress.

Also I really don't buy the social media thing given that we provide a function for downloading the files, which could then be stored wherever you want for as long as you want.

PR #923, opened 2023-12-08
DM-41941: Handle all options for RemoteButler.get() inputs

dhirving:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastores/fileDatastoreClient.py
+++ b/python/lsst/daf/butler/datastores/fileDatastoreClient.py
@@ -51,6 +51,7 @@ def get_dataset_as_python_object(
     universe: DimensionUniverse,
     parameters: Mapping[str, Any] | None,
     storageClass: StorageClass | str | None,
+    component: str | None,
 ) -> Any:
     """Retrieve an artifact from storage and return it as a Python object.
 
@@ -69,6 +70,8 @@ def get_dataset_as_python_object(
         Overrides the `StorageClass` to be used when converting the artifact to
         a Python object.  If `None`, uses the `StorageClass` specified by
         ``payload``.
+    component : `str` | `None`
+        Selects which component of the artifact to retrieve.
 
     Returns
     -------
@@ -81,6 +84,8 @@ def get_dataset_as_python_object(
     ]
 
     ref = DatasetRef.from_simple(payload.dataset_ref, universe=universe)
+    if component is not None:
+        ref = ref.makeComponentRef(component)
     if storageClass is not None:
         ref = ref.overrideStorageClass(storageClass)
 
--- a/python/lsst/daf/butler/datastores/fileDatastoreClient.py
+++ b/python/lsst/daf/butler/datastores/fileDatastoreClient.py
@@ -84,6 +84,10 @@ def get_dataset_as_python_object(
     ]
 
     ref = DatasetRef.from_simple(payload.dataset_ref, universe=universe)
+
+    # If we have both a component override and a storage class override, the
+    # component override has to be applied first.  DatasetRef cares because it
+    # is checking compatibility of the storage class with its DatasetType.
     if component is not None:
         ref = ref.makeComponentRef(component)
     if storageClass is not None:
--- a/python/lsst/daf/butler/dimensions/_coordinate.py
+++ b/python/lsst/daf/butler/dimensions/_coordinate.py
@@ -94,7 +94,7 @@ class SerializedDataCoordinate(_BaseModelCompat):
 
         Parameters
         ----------
-        dataId : `dict`
+        dataId : `SerializedDataId`
             The data ID.
         records : `dict` or `None`
             The dimension records.
--- a/python/lsst/daf/butler/dimensions/_coordinate.py
+++ b/python/lsst/daf/butler/dimensions/_coordinate.py
@@ -32,7 +32,14 @@
 
 from __future__ import annotations
 
-__all__ = ("DataCoordinate", "DataId", "DataIdKey", "DataIdValue", "SerializedDataCoordinate")
+__all__ = (
+    "DataCoordinate",
+    "DataId",
+    "DataIdKey",
+    "DataIdValue",
+    "SerializedDataCoordinate",
+    "SerializedDataId",
+)
 
 import numbers
 import warnings
@@ -69,17 +76,20 @@ DataIdValue = int | str | None
 DataCoordinate or other data ID.
 """
 
+SerializedDataId = dict[str, DataIdValue]
+"""Simplified model for serializing the ``mapping`` property of
+`DataCoordinate`.
+"""
+
 
 class SerializedDataCoordinate(_BaseModelCompat):
     """Simplified model for serializing a `DataCoordinate`."""
 
-    dataId: dict[str, DataIdValue]
+    dataId: SerializedDataId
     records: dict[str, SerializedDimensionRecord] | None = None
 
     @classmethod
-    def direct(
-        cls, *, dataId: dict[str, DataIdValue], records: dict[str, dict] | None
-    ) -> SerializedDataCoordinate:
+    def direct(cls, *, dataId: SerializedDataId, records: dict[str, dict] | None) -> SerializedDataCoordinate:
         """Construct a `SerializedDataCoordinate` directly without validators.
 
         Parameters
--- a/python/lsst/daf/butler/remote_butler/_remote_butler.py
+++ b/python/lsst/daf/butler/remote_butler/_remote_butler.py
@@ -239,6 +239,14 @@ class RemoteButler(Butler):
         response.raise_for_status()
         model = self._parse_model(response, GetFileResponseModel)
 
+        # If the caller provided a DatasetRef or DatasetType, they may have
+        # overridden the storage class on it.  We need to respect this, if they
+        # haven't asked to re-override it.
+        explicitDatasetType = _extract_dataset_type(datasetRefOrType)
+        if explicitDatasetType is not None:
+            if storageClass is None:
+                storageClass = explicitDatasetType.storageClass
+
         return get_dataset_as_python_object(
             model, parameters=parameters, storageClass=storageClass, universe=self.dimensions
         )
@@ -569,3 +577,15 @@ class RemoteButler(Butler):
             return DatasetTypeName(datasetTypeOrName.name)
         else:
             return DatasetTypeName(datasetTypeOrName)
+
+
+def _extract_dataset_type(datasetRefOrType: DatasetRef | DatasetType | str) -> DatasetType | None:
+    """Return the DatasetType associated with the argument, or None if the
+    argument is not an object that contains a DatasetType object.
+    """
+    if isinstance(datasetRefOrType, DatasetType):
+        return datasetRefOrType
+    elif isinstance(datasetRefOrType, DatasetRef):
+        return datasetRefOrType.datasetType
+    else:
+        return None
--- a/python/lsst/daf/butler/remote_butler/_remote_butler.py
+++ b/python/lsst/daf/butler/remote_butler/_remote_butler.py
@@ -230,7 +230,7 @@ class RemoteButler(Butler):
             request = GetFileByDataIdRequestModel(
                 dataset_type_name=self._normalize_dataset_type_name(datasetRefOrType),
                 collections=self._normalize_collections(collections),
-                data_id=self._simplify_dataId(dataId),
+                data_id=self._simplify_dataId(dataId, kwargs),
             )
             response = self._post("get_file_by_data_id", request)
             if response.status_code == 404:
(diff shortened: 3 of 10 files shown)


python/lsst/daf/butler/dimensions/_coordinate.py
```diff
@@ -69,17 +76,20 @@
 DataCoordinate or other data ID.
 """
 
+SerializedDataId = dict[str, DataIdValue]
```
timj, line 79: Oddly, `DataId` is defined below to use `Any` instead of `DataIdValue`.

python/lsst/daf/butler/dimensions/_coordinate.py
```diff
@@ -69,17 +76,20 @@
...
 
-    dataId: dict[str, DataIdValue]
+    dataId: SerializedDataId
     records: dict[str, SerializedDimensionRecord] | None = None
 
     @classmethod
-    def direct(
-        cls, *, dataId: dict[str, DataIdValue], records: dict[str, dict] | None
-    ) -> SerializedDataCoordinate:
+    def direct(cls, *, dataId: SerializedDataId, records: dict[str, dict] | None) -> SerializedDataCoordinate:
```
timj, line 92: Maybe change the type of the dataId parameter in the docstring below (although last I looked Sphinx can get upset with type aliases).

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -137,33 +144,29 @@ def dimensions(self) -> DimensionUniverse:
         self._dimensions = DimensionUniverse(config)
         return self._dimensions
 
-    def _simplify_dataId(
-        self, dataId: DataId | None, **kwargs: dict[str, int | str]
-    ) -> SerializedDataCoordinate | None:
+    def _simplify_dataId(self, dataId: DataId | None, kwargs: dict[str, int | str]) -> SerializedDataId:
```
timj:
```suggestion
    def _simplify_dataId(self, dataId: DataId | None, kwargs: dict[str, DataIdValue]) -> SerializedDataId:
```
?

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -137,33 +144,29 @@ def dimensions(self) -> DimensionUniverse:
...
-        **kwargs : `dict`
-            Additional values that should be included if this is not
-            a `DataCoordinate`.
+        kwargs : `dict`
+            Additional entries to augment or replace the values in ``dataId``.
 
         Returns
         -------
-        data_id : `SerializedDataCoordinate` or `None`
+        data_id : `SerializedDataId` or `None`
```
timj:
```suggestion
        data_id : `SerializedDataId`
```

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -216,19 +219,45 @@ def get(
...
-        dataset_id = datasetRefOrType.id
-        request = GetFileRequestModel(dataset_id=dataset_id)
-        response = self._post("get_file", request)
-        if response.status_code == 404:
-            raise LookupError(f"Dataset not found with ID ${dataset_id}")
+        if isinstance(datasetRefOrType, DatasetRef):
+            dataset_id = datasetRefOrType.id
+            response = self._get(f"get_file/{dataset_id}")
+            if response.status_code == 404:
+                raise LookupError(f"Dataset not found with ID {dataset_id}")
```
timj: I know this was in the old code but it might be better to print the `datasetRefOrType` here and not just the ID since that is more human readable (and will include the ID).

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -216,19 +219,45 @@ def get(
...
+        else:
+            request = GetFileByDataIdRequestModel(
+                dataset_type_name=self._normalize_dataset_type_name(datasetRefOrType),
+                collections=self._normalize_collections(collections),
+                data_id=self._simplify_dataId(dataId, kwargs),
+            )
+            response = self._post("get_file_by_data_id", request)
+            if response.status_code == 404:
+                raise LookupError(f"Dataset not found with DataId {dataId}")
```
timj: This error message needs to report the dataset type and run as well.

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -216,19 +219,45 @@ def get(
...
+        explicitDatasetType = _extract_dataset_type(datasetRefOrType)
+        if explicitDatasetType is not None:
+            if storageClass is None:
+                storageClass = explicitDatasetType.storageClass
+
+        # If the caller provided a DatasetRef, they may have overridden the
+        # component on it.  We need to explicitly handle this because we did
+        # not send the DatasetType to the server in this case.
+        componentOverride = None
+        if isinstance(datasetRefOrType, DatasetRef):
```
timj, line 255: I realize that we talked about this not being a hotspot, but doing the instance check twice when we already know the answer a few lines higher up seems wasteful. Can't we calculate componentOverride in the block above?

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -216,19 +219,45 @@ def get(
...
+            if response.status_code == 404:
+                raise LookupError(f"Dataset not found with DataId {dataId}")
+
         response.raise_for_status()
         model = self._parse_model(response, GetFileResponseModel)
 
+        # If the caller provided a DatasetRef or DatasetType, they may have
+        # overridden the storage class on it.  We need to respect this, if they
+        # haven't asked to re-override it.
+        explicitDatasetType = _extract_dataset_type(datasetRefOrType)
```
timj, line 246: We can avoid one of the isinstance checks if we retrieve the dataset type from the dataset ref we knew we got in the first branch.

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -280,23 +309,22 @@ def get_dataset(
...
-            params["storage_class"] = storage_class_name
         response = self._client.get(self._get_url(path), params=params)
         response.raise_for_status()
         if response.json() is None:
             return None
-        return DatasetRef.from_simple(SerializedDatasetRef(**response.json()), universe=self.dimensions)
+        ref = DatasetRef.from_simple(
+            self._parse_model(response, SerializedDatasetRef), universe=self.dimensions
+        )
+        if storage_class is not None:
```
timj, line 328: Thanks for fixing this.

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -545,9 +560,53 @@ def _get_url(self, path: str, version: str = "v1") -> str:
...
+        # want to construct CollectionWildcard and filter it through collection
+        # cache to generate list of collection names.
+        wildcards = CollectionWildcard.from_expression(collections)
+        return CollectionList(list(wildcards.strings))
+
+    def _normalize_dataset_type_name(self, datasetTypeOrName: DatasetType | str) -> DatasetTypeName:
+        """Convert DatasetType parameters in the format used by Butler methods
+        to a standardized string name for the REST API.
+        """
+        if isinstance(datasetTypeOrName, DatasetType):
```
timj, line 600: Would 

```python
return getattr(datasetTypeOrName, "name", datasetTypeOrName)
```
be more efficient?

python/lsst/daf/butler/remote_butler/server/handlers/_external.py
```diff
@@ -140,33 +138,57 @@ def find_dataset(
...
+    try:
+        ref = butler._findDatasetRef(
+            datasetRefOrType=request.dataset_type_name,
+            dataId=request.data_id,
+            collections=request.collections,
+            datastore_records=True,
+        )
+        return _get_file_by_ref(butler, ref)
+    except LookupError as e:
+        raise NotFoundException() from e
```
timj, line 188: No error message?

tests/test_server.py
```diff
@@ -216,30 +242,86 @@ def override_read(http_resource_path):
...
         # Test storage class override
         new_sc = self.storageClassFactory.getStorageClass("MetricsConversion")
-        converted = self.butler.get(ref, storageClass=new_sc)
-        self.assertNotEqual(type(metric), type(converted))
-        self.assertIs(type(converted), new_sc.pytype)
-        self.assertEqual(metric, converted)
+
+        def check_sc_override(converted):
+            self.assertNotEqual(type(metric), type(converted))
+            self.assertIs(type(converted), new_sc.pytype)
```
timj:
```suggestion
            self.assertIsInstance(converted, new_sc.pytype)
```
?

python/lsst/daf/butler/datastores/fileDatastoreClient.py
```diff
@@ -81,6 +84,8 @@ def get_dataset_as_python_object(
     ]
 
     ref = DatasetRef.from_simple(payload.dataset_ref, universe=universe)
+    if component is not None:
```
timj, line 91: Maybe note that you have to apply the component override before the storage class override.

timj (approved):
Looks great. Nice cleanups.

python/lsst/daf/butler/dimensions/_coordinate.py
dhirving: You mean in this bit?:
```
DataId = DataCoordinate | Mapping[str, Any]
```

Not really sure what that's about.  DataCoordinate proper does seem to use the DataIdValue pretty consistently everywhere.  That `DataId` union exists mainly as a parameter for Butler methods I think, so maybe when you guys added mypy there was too much random code that was passing in like stringifiable things that weren't actually strings or something.  There is a bit of code in `standardize` doing things like coercing numpy.int64 to int, but not as much as I would have thought.

I kind of feel like there are a values that aren't ints or strings though.  Isn't some stuff like floats or datetimes?

I sort of suspect that the existing QuantumGraph code or wherever this is getting used just happens to not use anything unusual, and we may end up with some stuff getting accidentally coerced to strings at some point.  In any case I just pulled out an alias for the existing definition from SerializedDataCoordinate, so a problem for another day.

python/lsst/daf/butler/remote_butler/server/handlers/_external.py
dhirving: It's not actually shipping the error message from those NotFoundExceptions to the client yet, and the `from e` will put the original error message in the server logs.  I suppose I could have put in "DataId not found" or something.

In a couple weeks when I go in and set up the error propagation framework, that try block will go away entirely.

python/lsst/daf/butler/dimensions/_coordinate.py
timj: Dimensions do not use floats because you can't do equality on them. Dimensions are strings or int (and as you found, integral types so really should be `numbers.Integral` or something instead of int).

python/lsst/daf/butler/remote_butler/_remote_butler.py
dhirving: That is how I had it originally actually, but I re-organized it for clarity.  This version moves the declaration of variables closer to the code that uses them and separate the concerns.  Above the block is concerned with deciding which REST endpoint to call, down here we are focused on post-processing the result.  

I originally had both the storage class and component override thing up above, but I discovered that above there are only two cases (DatasetRef or not), but for postprocessing there are three cases (DatasetRef, inflated DatasetType, string DatasetType).

As I have been reading the existing Butler methods to try to understand how they work, one of the things that has making it more difficult is jumping up and down multi-hundred-line functions to figure out where a variable came from and whether it has been mutated since its original definition.  Unfortunately python lacks the `const` declarations that most other languages have.  And because Butler methods all take extremely long argument lists with variable argument types, it's harder to add helper functions to break them into chunks because all the helper methods would need to take like 6 parameters.

Instance checks are not expensive.  It would be better if the Butler API was designed in a way that didn't make them necessary, but unfortunately it's not.  In the absolute worst case it's walking a very short list (like 3 items) and doing a couple of method calls to `__isinstance__`.  In the typical case (which this is) it's basically just a pointer comparison.  It's not worth worrying about these kinds of microoptimizations except in tight inner loops.  This is not a tight inner loop, since this function does network I/O to go out to a service that then does network I/O to ask a database to do disk I/O.

Computers are pretty fast now, but developer time is more expensive than ever.  This kind of stuff is not worth worrying about.  Clarity is a lot more important, as is minimizing the amount of time spent arguing over things that don't matter so that there is more time available for writing code.

python/lsst/daf/butler/remote_butler/_remote_butler.py
dhirving: Slightly, yeah, but less type safe and more ambiguous.  I personally have no idea off the top of my head whether `str` or any str-like objects that may be passed in by a user have a `name` attribute or not.

PR #924, opened 2023-12-14
DM-42187: Add RemoteButler.getURIs

dhirving:
## Checklist

- [ ] ran Jenkins


python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -287,9 +289,26 @@ def getURI(
         collections: Any = None,
         run: str | None = None,
         **kwargs: Any,
-    ) -> ResourcePath:
+    ) -> DatasetRefURIs:
         # Docstring inherited.
-        raise NotImplementedError()
+        if predict or run:
+            raise NotImplementedError("Predict mode is not supported by RemoteButler")
```
timj, line 345: We need to discuss whether we can make this work. Obviously a signed URL to a resource that doesn't exist is not all that helpful but people like to know what the filename might look like. We can tell them that in theory -- maybe for predict mode we return the unsigned `s3` URI? Can punt this to another ticket.

tests/test_server.py
```diff
@@ -104,9 +106,11 @@ def setUpClass(cls):
             configFile=os.path.join(TESTDIR, "config/basic/butler-s3store.yaml"),
             forceConfigRoot=False,
         )
-
         # Add a file with corrupted data for testing error conditions
         cls.dataset_with_corrupted_data = _create_corrupted_dataset(cls.repo)
+        # All of the datasets that come with MetricTestRepo are disassembled
```
timj, line 111: I had forgotten that the MetricTestRepo defaults to storage class `StructuredCompositeReadComp`.  Easy to change the default if we wanted to but also nice to know that getURIs should work.

tests/test_server.py
```diff
@@ -323,6 +328,50 @@ def check_sc_override(converted):
...
+            self.assertEqual(data, b"AM1: 5.2\nAM2: 30.6\n")
+
+        uris = self.butler.getURIs(dataset_type, dataId=data_id, collections=collections)
+        check_uris(uris)
+
+        # Calling getURI on a multi-file dataset raises an exception
+        with self.assertRaises(RuntimeError):
+            self.butler.getURI(dataset_type, dataId=data_id, collections=collections)
+
+        # getURIs does NOT respect component overrides on the DatasetRef,
```
timj, line 366: There is only one URI for any of the components if not disassembled. I imagine no-one has ever tried to request the URI of an individual component by passing in the ref but it seems like it should return the one URI of the component.

timj (approved):
Looks good.

PR #927, opened 2023-12-15
DM-42216: enable collection caching in several locations, including query-results iteration.

TallJimbo:
## Checklist

- [ ] ran [Jenkins](https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/233/pipeline)


timj (approved):
A quick look indicates that this is doing what you say it is doing (adding a context manager around many of the queries). I assume you see many fewer debug log messages with this enabled.

TallJimbo:
Yup, looking at the debug log messages from query-datasets was exactly how I tested this.

andy-slac (approved):
Looks great. Thanks for the fix, I should have thought about it on DM-41117, but I only looked at the queries from ci_hsc_gen3 at that time, unfortunately.

PR #929, opened 2023-12-21
DM-42302: drop support for Pydantic v1

TallJimbo:
## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/_compat.py
```diff
@@ -1,272 +0,0 @@
...
-
-    class _BaseModelCompat(BaseModel):
-        """Methods from pydantic v1 that we want to emulate in v2.
-
-        Some of these methods are provided by v2 but issue deprecation
-        warnings.  We need to decide whether we are also okay with deprecating
-        them or want to support them without the deprecation message.
-        """
-
-        def json(
```
timj, line 65: I wasn't really sure if we were going to break people by dropping the `json` method here. I think that there might be people who use `SerializedDatasetRef` outside of our middleware but maybe only `to_simple` and `from_simple` rather than calling `.json` and `.parse_obj`. Maybe @mfisherlevine ?

python/lsst/daf/butler/datastore/record_data.py
```diff
@@ -46,16 +46,10 @@
 if TYPE_CHECKING:
     from ..registry import Registry
 
-# Pydantic 2 requires we be explicit about the types that are used in
```
timj, line 49: Would this comment still be helpful?

timj (approved):

python/lsst/daf/butler/datastore/record_data.py
TallJimbo: Maybe; the interesting bit is _why_ `UUID` doesn't work with `Any`, and I'm guessing it's because there's nothing to distinguish it from `str`, and it doesn't see if a `str` is UUID-like unless you explicitly tell it that `UUID` is a possibility.  But that's pure speculation.

python/lsst/daf/butler/_compat.py
TallJimbo: I've been thinking of all the `Serialized*` as middleware-internal, but I guess that's not quite true since those `*_simple` methods are public.

I'd really rather not go through a deprecation period for an interface we just inherit from a third-party package, but I'm happy to stall merging this for a few days to see if Merlin chimes in.

timj: The compromise is to merge this tomorrow after the next weekly and then you have a week to find out from people if we need to put a shim back in specifically for `SerializedDatasetRef`.

python/lsst/daf/butler/datastore/record_data.py
timj: I think this was something that became necessary after some pydantic 2 releases had already been made.

TallJimbo: I'll restore the comment, or something like it.

PR #930, opened 2023-12-21
DM-42188: Make RemoteButler usable from services

dhirving:
## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/_butler.py
```diff
@@ -297,17 +270,44 @@ def from_config(
...
+                butler_config,
+                options=options,
+                without_datastore=without_datastore,
+            )
+        elif butler_type == ButlerType.REMOTE:
+            from .remote_butler import RemoteButlerFactory
+
+            factory = RemoteButlerFactory.create_factory_from_config(butler_config)
+            return factory.create_butler_with_credentials_from_environment(butler_options=options)
+
```
timj, line 313: Can you add a `raise` here for clarity if there is no match to REMOTE or DIRECT? I know that there are currently only two values in the enum but this ties it up nicely. Consider using `match` here since that is designed for enums.

python/lsst/daf/butler/_butler.py
```diff
@@ -1691,3 +1691,18 @@ def _query_dimension_records(
...
+        self,
+        *,
+        collections: Any = None,
+        run: str | None = None,
+        inferDefaults: bool = True,
+        **kwargs: Any,
+    ) -> Butler:
+        """Return a new Butler instance connected to the same repository
+        as this one, but overriding ``collections``, ``run``,
+        ``inferDefaults``, and default data ID.
```
timj, line 1709: When this method becomes public I think it will trigger the numpydoc warning. Might want to add the public docs here now rather than later.

python/lsst/daf/butler/_butler_config.py
```diff
@@ -191,3 +194,20 @@ def __init__(
...
+    def get_butler_type(self) -> ButlerType:
+        # Configuration optionally includes a class name specifying which
+        # implementation to use, DirectButler or RemoteButler.
+        butler_class_name = self.get("cls")
+        if butler_class_name is None:
+            # There are many existing DirectButler configurations that are
+            # missing the ``cls`` property.
+            return ButlerType.DIRECT
+        elif butler_class_name == "lsst.daf.butler.direct_butler.DirectButler":
+            return ButlerType.DIRECT
```
timj, line 207: Can you please add a `cls` entry to one of the direct butler config files in `tests/config/basic` so we can trigger this code path?

python/lsst/daf/butler/_butler_config.py
```diff
@@ -191,3 +194,20 @@ def __init__(
...
+        elif butler_class_name == "lsst.daf.butler.remote_butler.RemoteButler":
+            return ButlerType.REMOTE
+        else:
+            raise ValueError(
+                f"Butler configuration requests to load unknown Butler class {butler_class_name}"
```
timj, line 212:
```suggestion
                f"Butler configuration requests to load unknown Butler class {butler_class_name!r}"
```
so the class name is quoted.

python/lsst/daf/butler/_butler_factory.py
```diff
@@ -0,0 +1,165 @@
...
+
+class ButlerFactory:
+    """Factory for efficiently instantiating Butler instances from the
+    repository index file.  This is intended for use from long-lived services
+    that want to instantiate a separate Butler instance for each end user
+    request.
+
+    Parameters
+    ----------
+    repositories : `Mapping`[`str`, `str`], optional
```
timj:
```suggestion
    repositories : `~collections.abc.Mapping`[`str`, `str`], optional
```

python/lsst/daf/butler/_butler_factory.py
```diff
@@ -0,0 +1,165 @@
...
+
+            factory = self._create_butler_factory_function(label)
+            return self._factories.set_or_get(label, factory)
+
+    def _create_butler_factory_function(self, label: str) -> _FactoryFunction:
+        config_uri = self._get_config_uri(label)
+        config = ButlerConfig(config_uri)
+        butler_type = config.get_butler_type()
+
+        if butler_type == ButlerType.DIRECT:
```
timj: Use `match` and catch if doesn't match?

python/lsst/daf/butler/_butler_factory.py
```diff
@@ -0,0 +1,165 @@
...
+            return _create_direct_butler_factory(config)
+        elif butler_type == ButlerType.REMOTE:
+            return _create_remote_butler_factory(config)
+
+    def _get_config_uri(self, label: str) -> ResourcePathExpression:
+        if self._repositories is None:
+            return ButlerRepoIndex.get_repo_uri(label)
+        else:
+            return self._repositories[label]
```
timj: What happens if someone passes in a URI instead of a label? The `ButlerConfig` constructor tries to handle this. I'm not sure if that breaks here (or what the error message looks like if someone uses a label that isn't in the INDEX).

python/lsst/daf/butler/direct_butler.py
```diff
@@ -109,119 +109,133 @@ class ButlerValidationError(ValidationError):
...
         # butler.
         self._datastore.set_retrieve_dataset_type_method(self._retrieve_dataset_type)
 
-        if "run" in self._config or "collection" in self._config:
+        self._registry_shim = RegistryShim(self)
+
+        return self
+
+    @classmethod
+    def create_from_config(
```
timj, line 166: We now have a `create_from_config` and a `from_config` ? (which I see registry factory also has). Is `create_from_config` pseudo private? Is it conceptually what `DirectButler.from_config` calls with cleaned up API?

python/lsst/daf/butler/direct_butler.py
```diff
@@ -273,14 +287,11 @@ def _unpickle(
         butler : `Butler`
             A new `Butler` instance.
         """
-        # MyPy doesn't recognize that the kwargs below are totally valid; it
-        # seems to think '**defaultDataId* is a _positional_ argument!
-        return cls(
+        return cls.create_from_config(
```
timj, line 294: It occurs to me that pickling is not handling the without_datastore flag.

python/lsst/daf/butler/direct_butler.py
```diff
@@ -2369,3 +2382,5 @@ def _query_dimension_records(
     """An object that maps known storage class names to objects that fully
     describe them (`StorageClassFactory`).
     """
+
+    _registry_shim: RegistryShim
```
timj, line 2391: Docstring for this and `_config`? The others do have docstrings even though private.

python/lsst/daf/butler/remote_butler/_factory.py
```diff
@@ -0,0 +1,111 @@
...
+    """Factory for instantiating RemoteButler instances bound to a user's Rubin
+    Science Platform Gafaelfawr access token.  All Butler instances created by
+    this factory share a common HTTP connection pool.
+
+    Parameters
+    ----------
+    server_url : `str`
+        The URL of the Butler server that RemoteButler instances created by
+        this factory will connect to.
+
```
timj: Remove blank line. Not needed.

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -188,15 +153,19 @@ def isWriteable(self) -> bool:
     @property
     def dimensions(self) -> DimensionUniverse:
         # Docstring inherited.
-        if self._dimensions is not None:
-            return self._dimensions
+        with self._cache.access() as cache:
```
timj, line 156: Does this need the read lock? If the value `is not None` then no other thread is going to make it `None`.

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -629,18 +598,21 @@ def _get_url(self, path: str, version: str = "v1") -> str:
         path : `str`
             The full path to the endpoint.
         """
-        return f"{version}/{path}"
+        slash = "" if self._server_url.endswith("/") else "/"
```
timj, line 601: Can we force `self._server_url` to always have the `/` so we don't need to check every time on get?

tests/test_butler_factory.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        repo_dir = cls.enterClassContext(tempfile.TemporaryDirectory())
+        makeTestRepo(repo_dir)
+        cls.config_file_uri = f"file://{repo_dir}"
+
+    def test_factory_via_global_repository_index(self):
+        with tempfile.NamedTemporaryFile(suffix=".yaml") as index_file:
+            index_file.write(f"test_repo: {self.config_file_uri}\n".encode())
```
timj, line 52:
```suggestion
            print(f"test_repo: {self.config_file_uri}", file=index_file)
```
(and open temp file with `mode="w"`)? It is shorter and seems more natural.

tests/test_thread_safe_cache.py
```diff
@@ -0,0 +1,47 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import unittest
+
+from lsst.daf.butler._utilities.thread_safe_cache import ThreadSafeCache
```
timj: Maybe put the three thread safety tests in a single test file? They are very small.

timj (approved):
Looks really good.

python/lsst/daf/butler/_butler_factory.py
```diff
@@ -0,0 +1,165 @@
...
+from ._utilities.named_locks import NamedLocks
+from ._utilities.thread_safe_cache import ThreadSafeCache
+
+_FactoryFunction = Callable[[str | None], Butler]
+"""Function that takes an access token string or `None`, and returns a Butler
+instance."""
+
+
+class ButlerFactory:
+    """Factory for efficiently instantiating Butler instances from the
```
timj: Now that I'm reading #931, we seem to be saying that server is required to use a label from a DAF_BUTLER_REPOSITORY_INDEX.  This makes me ponder that the name of the class is a bit too generic since DirectButler is not required to use the index labels and we don't expect this class to be used in non-server context.

python/lsst/daf/butler/_butler_factory.py
dhirving: It just throws a KeyError for label not found.  It's important for security that this NOT work for URLs, because these labels are sometimes untrusted user input.

timj: KeyError from python is a famously opaque error message.

python/lsst/daf/butler/direct_butler.py
dhirving: I wanted to name this `from_config` but mypy wouldn't let me because it has a different signature than `Butler.from_config`.  There is no public `create_from_config` -- it's private because it's only on `DirectButler`.

timj: It is public in the sense that:

```python
>>> b = Butler("dp02")
>>> help(b)
```
will show the method, and presumably Jupyter Notebooks will report the method. Maybe add an extra note to the docstring pointing to `Butler.from_config` as the general interface.

python/lsst/daf/butler/remote_butler/_remote_butler.py
dhirving: I have not been able to locate any definitive information about Python's memory model for threading other than "if you use the synchronization primitives from the threading module then it works how you would expect".  

We probably don't need this on the read in the current version of CPython because of the GIL but that's not documented as permanent behavior anywhere that I am aware of.

python/lsst/daf/butler/remote_butler/_remote_butler.py
dhirving: Yeah I had considered that but I was concerned someone might assign the server URL from somewhere else, forgetting the slash, and it would break intermittently.  By putting this check here we enforce it at the point where the code depends on it having a slash.

timj: Isn't `_server_url` private?

python/lsst/daf/butler/_butler_factory.py
dhirving: Well, it's DAF_BUTLER_REPOSITORY_INDEX or "pass in whatever set of labels and config files you want."  I'd be open to renaming if you have any suggestions.

python/lsst/daf/butler/_butler.py
dhirving: Because mypy's exhaustiveness checking will ensure that all the cases are handled, there's some value in not having the raise.  Without the raise, when someone adds an enum value, mypy will tell them all the places that need to do something with it.

I don't feel that strongly about it one way or the other though.

python/lsst/daf/butler/_butler_factory.py
timj: Since it has to be able to return DirectButler and RemoteButler for now the class can't really be ServerButlerFactory (as is the name of the variable you assign to). `LabeledButlerFactory`? `NamedButlerFactory`?

python/lsst/daf/butler/_butler.py
timj: `match` naturally comes with a fall through clause that can raise. I feel better if the code looks like it's dealing with all the options without relying on mypy.

python/lsst/daf/butler/remote_butler/_remote_butler.py
dhirving: Yes but RemoteButler is already almost 700 lines long and we haven't implemented 95% of its methods yet.  I don't see any reason not to enforce correctness at the point of use instead of adding a comment like "make extra sure this variable always has a slash at the end of it!"

timj: I think I'm missing something. `_server_url` is set in one place in the constructor so my suggestion was to move this fixup line up to that point and normalize the string there. Then it's fixed once and never needs to be tested again.

I wasn't suggesting we document that all places that construct a RemoteButler must ensure the / is added.

dhirving: Currently it is, but with the rate that these constructors are proliferating we may end up with it getting set more than one place.  

The only place that must have a slash is _get_url, so what is the benefit to doing this logic in the constructor?  The constructor is plenty busy with other issues already.

timj: Ok. Leave it.  I don't really understand why we want to do a check and append every time the server is called rather than once when we construct the object. `_server_url` is entirely internal. We are in charge of how that variable is set.

python/lsst/daf/butler/_utilities/thread_safe_cache.py
```diff
@@ -0,0 +1,78 @@
...
+    """A simple thread-safe cache.  Ensures that once a value is stored for
+    a key, it does not change.
+    """
+
+    def __init__(self) -> None:
+        # This mutex may not actually be necessary for the current version of
+        # CPython, because both the 'get' and 'set' operation use underlying C
+        # functions that don't drop the GIL (at first glance anyway.) This is
+        # not documented anywhere that I can find, however, so better safe than
+        # sorry.
```
andy-slac, line 45: Extremely late comment - all built-in Python types are guaranteed to be thread-safe in CPython, there is a comment about it in https://docs.python.org/3/glossary.html#term-global-interpreter-lock. But an explicit mutex may be useful once we switch to GIL-free mode 😀

PR #931, opened 2023-12-27
DM-42189: Serve multiple Butler repositories from a single server

dhirving:
In most environments, users will access multiple Butler repositories (multiple data releases, prompt processing, etc.).  To avoid the need to instantiate a separate service for every repository, allow multiple repositories to be hosted from a single server.

This uses the same DAF_BUTLER_REPOSITORY_INDEX configuration system as DirectButler.

## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/remote_butler/server/_factory.py
```diff
@@ -25,22 +25,31 @@
...
+    """Class for instantiating per-request dependencies, following the pattern
+    in `SQR-072 <https://sqr-072.lsst.io/>`_.
 
     Parameters
     ----------
-    butler : `DirectButler`
-        Butler to use.
+    repository : `str`
+        The label of the Butler repository requested by the user.
+
```
timj: Numpydoc doesn't need the blank line here.

tests_integration/test_docker_container.py
```diff
@@ -24,21 +24,28 @@ def _run_server_docker():
 
         port = 8080
         butler_root = "/butler_root"
+
+        # Set up a repository index file to be read in by the server
+        index_filename = "repo_index.yaml"
+        repo_name = "testserver"
+        with open(os.path.join(temp_dir, index_filename), "wb") as fh:
+            fh.write(f"{repo_name}: {butler_root}\n".encode())
```
timj, line 32:
```suggestion
            yaml.dump({repo_name: butler_root}, stream=fh)
```
?

timj (approved):
Looks good. Is the `/repo` in the path really necessary?

dhirving:
> Looks good. Is the `/repo` in the path really necessary?

Yeah, it leaves a free namespace where we can put per-server endpoints, like "list_available_repositories" or "health_check".  Otherwise we'd have to worry about repository labels colliding with endpoint names.  (Not sure but I wouldn't be surprised if FastAPI doesn't even support having a `/{placeholder}` that potentially collides with another name.)

PR #935, opened 2024-01-03
DM-42317: Make cloned Butler instances threadsafe

dhirving:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastore/_datastore.py
+++ b/python/lsst/daf/butler/datastore/_datastore.py
@@ -287,8 +287,6 @@ class Datastore(metaclass=ABCMeta):
         referring to a configuration file.
     bridgeManager : `DatastoreRegistryBridgeManager`
         Object that manages the interface between `Registry` and datastores.
-    butlerRoot : `str`, optional
-        New datastore root to use to override the configuration value.
     """
 
     defaultConfigFile: ClassVar[str | None] = None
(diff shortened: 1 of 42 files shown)


python/lsst/daf/butler/datastore/_datastore.py
```diff
@@ -403,6 +401,39 @@ def __init__(
...
         self.constraints = Constraints(constraintsConfig, universe=bridgeManager.universe)
 
+    @classmethod
+    @abstractmethod
+    def _create_from_config(
+        cls,
+        config: DatastoreConfig,
+        bridgeManager: DatastoreRegistryBridgeManager,
+        butlerRoot: ResourcePathExpression | None,
+    ) -> Datastore:
```
TallJimbo, line 411: Since this and `Datastore.clone` are brand new interfaces, let's make the arguments `snake_case`.

python/lsst/daf/butler/registry/sql_registry.py
```diff
@@ -276,7 +276,7 @@ def isWriteable(self) -> bool:
 
     def copy(self, defaults: RegistryDefaults | None = None) -> SqlRegistry:
         """Create a new `SqlRegistry` backed by the same data repository
-        and connection as this one, but independent defaults.
+        as this one, but independent defaults and database connection.
```
TallJimbo: Since we're sharing the `sqlalchemy.Engine`, I'm not sure that's necessarily true; I think we're sharing a connection pool and hence potentially sharing connections (with the thread-safety details of that left to SQLAlchemy), but we think that at any particular time a particular connection can only be held by one `Database` instance.

I'm asking partly to make sure I understand how copying the `Database` object actually gives us thread-safety, not just to make this docstring more precise.  I think connections not being shared "at any particular time" is sufficient, but I wanted to see if that matches your mental model.

python/lsst/daf/butler/registry/_defaults.py
```diff
@@ -120,6 +120,11 @@ def finish(self, registry: SqlRegistry) -> None:
             Raised if a non-governor dimension was included in ``**kwargs``
             at construction.
         """
+        # Skip re-initialization if it's already been completed.
+        # Can't just say 'self._finished' because this class is immutable.
+        if hasattr(self, "_finished"):
```
TallJimbo, line 125: Would it work to set
```
_finished: bool = False
```
at class scope instead (i.e. so it starts out as a ClassVar, and then gets shadowed)?

TallJimbo (approved):

python/lsst/daf/butler/registry/sql_registry.py
dhirving: Yeah, what you described matches my understanding.  `sqlalchemy.Engine` guarantees that only one caller has a given "logical" connection checked out at a time, though you're correct that it doesn't necessarily map to a separate network connection.  

To be clear, the part that is now independent is that each Database instance will use a different `sqlalchemy.engine.Connection` object, whereas before they would sometimes be accidentally shared via the `Database._session_connection` instance variable.  This was problematic for a few reasons:
1. `Connection` is specifically documented to not be threadsafe
2. The transaction state would get mixed together between the threads because this is managed via the `Connection` object
3. Depending on the timing, one thread could close the `Connection` object before the other thread was done using it, or replace it with a different connection while another thread was in the middle of using it.

Perhaps a better phrasing would be "independent database session"... I'll tweak the comment.

python/lsst/daf/butler/registry/_defaults.py
dhirving: Tried it but it didn't work unfortunately.

PR #939, opened 2024-01-10
DM-42344: dimensions configuration rework and other cleanups from DM-42324

TallJimbo:
## Checklist

- [x] ran Jenkins
- [ ] ~added a release note for user-visible changes to `doc/changes`~ (considering this invisible)

### DIFF
--- a/python/lsst/daf/butler/registry/sql_registry.py
+++ b/python/lsst/daf/butler/registry/sql_registry.py
@@ -1643,10 +1643,11 @@ class SqlRegistry:
         else:
             # Ignore typing since caller said to trust them with conform=False.
             records = data  # type: ignore
+        if element.name in self.dimension_record_cache:
+            self.dimension_record_cache.reset()
         self._managers.dimensions.insert(
             element,
             *records,
-            cache=self.dimension_record_cache,
             replace=replace,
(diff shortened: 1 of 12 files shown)


python/lsst/daf/butler/registry/sql_registry.py
```diff
@@ -1643,10 +1643,11 @@ def insertDimensionData(
         else:
             # Ignore typing since caller said to trust them with conform=False.
             records = data  # type: ignore
+        if element.name in self.dimension_record_cache:
+            self.dimension_record_cache.reset()
```
dhirving, line 1647: I think I like having this cache invalidation logic (and ownership of this cache in general) up here in SqlRegistry.  I definitely didn't like it down in the manager, anyway -- too hidden and too much mixing of concerns between the layers.

The risk of having it up here is that we miss an invalidation if we add more uses of the manager.insert/sync later, but I think the clarity gains from having the caching visible and fully owned in this layer outweigh that.

dhirving (approved):
Looks good to me.  The DimensionUniverse configuration is a lot more readable now and it's good to have validation of that config.  The simplification of the caching logic is nice to see too.

PR #942, opened 2024-01-16
DM-42306: Performance optimizations

timj:
Depends on lsst/resources#76

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1309,6 +1309,7 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
         skip_missing: bool = True,
         register_dataset_types: bool = False,
         transfer_dimensions: bool = False,
+        dry_run: bool = False,
     ) -> Collection[DatasetRef]:
         """Transfer datasets to this Butler from a run in another Butler.
 
@@ -1335,6 +1336,10 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
         transfer_dimensions : `bool`, optional
             If `True`, dimension record data associated with the new datasets
             will be transferred.
+        dry_run : `bool`, optional
+            If `True` the transfer will be processed without any modifications
+            made to the target butler and as if the target butler did not
+            have any of the datasets.
 
         Returns
         -------
--- a/python/lsst/daf/butler/_location.py
+++ b/python/lsst/daf/butler/_location.py
@@ -47,7 +47,7 @@ class Location:
         path separator if a ``file`` scheme is being used for the URI,
         else a POSIX separator. Can be a full URI if the root URI is `None`.
         Can also be a schemeless URI if it refers to a relative path.
-    _trusted_path : `bool`, optional
+    trusted_path : `bool`, optional
         If `True`, the path is not checked to see if it is really inside
         the datastore.
     """
@@ -59,7 +59,7 @@ class Location:
         datastoreRootUri: None | ResourcePathExpression,
         path: ResourcePathExpression,
         *,
-        _trusted_path: bool = False,
+        trusted_path: bool = False,
     ):
         # Be careful not to force a relative local path to absolute path
         path_uri = ResourcePath(path, forceAbsolute=False, forceDirectory=False)
@@ -89,7 +89,7 @@ class Location:
 
         # Check that the resulting URI is inside the datastore
         # This can go wrong if we were given ../dir as path
-        if self._datastoreRootUri is not None and not _trusted_path:
+        if self._datastoreRootUri is not None and not trusted_path:
             pathInStore = self.uri.relative_to(self._datastoreRootUri)
             if pathInStore is None:
                 raise ValueError(f"Unexpectedly {path} jumps out of {self._datastoreRootUri}")
@@ -245,7 +245,7 @@ class LocationFactory:
         """Return the network location of root location of the `Datastore`."""
         return self._datastoreRootUri.netloc
 
-    def fromPath(self, path: ResourcePathExpression, *, _trusted_path: bool = False) -> Location:
+    def fromPath(self, path: ResourcePathExpression, *, trusted_path: bool = False) -> Location:
         """Create a `Location` from a POSIX path.
 
         Parameters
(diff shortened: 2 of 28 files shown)


python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -2449,9 +2485,38 @@ def transfer_from(
...
+                    new_records = _check_missing_files(source_datastore, missing_ids_chunk)
+                    for k, v in new_records.items():
+                        source_records[k].extend(v)
+
+            log.verbose("Completed scan for missing data files")
 
         # See if we already have these records
-        target_records = self._get_stored_records_associated_with_refs(refs, ignore_datastore_records=True)
+        if not dry_run:
+            target_records = self._get_stored_records_associated_with_refs(
```
timj: This is the line I'm not sure about. I imagine people expect dry-run to know which records are really needed to be transferred rather than having the system think that it has to do all the checks again because the target doesn't have the datastore entries even when it does.

python/lsst/daf/butler/_butler.py
```diff
@@ -1335,6 +1336,10 @@ def transfer_from(
         transfer_dimensions : `bool`, optional
             If `True`, dimension record data associated with the new datasets
             will be transferred.
+        dry_run : `bool`, optional
+            If `True` the transfer will be processed without any modifications
+            made to the target butler and as if the target butler did not
+            have any of the datasets.
```
TallJimbo, line 1342: It will check to see whether the artifacts in the graph exist, right?

I agree with your comment on Jira that it'd be better if it didn't assume that the datasets were not in the target butler, but this is useful enough as a starting point.  Maybe move that qualification to a `Notes` section, and amend it to say that this might change in the future?

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -2490,9 +2491,17 @@ def transfer_from(
...
+                refs, ignore_datastore_records=True
+            )
+        else:
+            # In dry-run mode we assume there is nothing in target repo.
+            # (otherwise everything would report that no work is needed.)
```
TallJimbo: I don't understand this.  In the typical production case, the file artifacts are mostly all present and the others are completely absent (since we run directly in the datastore root, so there are no file transfers), and the "work needed" is inserting stuff into the database.  What keeps the dry-run from doing the same, and then (doing the equivalent of) rolling back the DB transaction instead of committing it?

python/lsst/daf/butler/datastores/fileDatastore.py
TallJimbo: Is this here because we're worried that the datasets might be present in the registry database, but not have file artifacts in the datastore and/or datastore records in the database?

I think it ought to be fine to assume that if a ref in the graph is in the registry we can skip it, as if it was inserted into the registry without associated datastore content that was probably intentional.

python/lsst/daf/butler/datastore/file_templates.py
```diff
@@ -449,7 +449,7 @@ def format(self, ref: DatasetRef) -> str:
         # context should only happen in unit tests; in general we need all
         # dimensions to fill out templates).
         fields: dict[str, object] = {
-            k: ref.dataId.get(k) for k in ref.datasetType.dimensions.names if ref.dataId.get(k) is not None
+            k: v for k in ref.datasetType.dimensions.names if (v := ref.dataId.get(k)) is not None
```
TallJimbo:
```
fields = dict(dataId.mapping)
```
is all you need now.

[DM-21840](https://jira.lsstcorp.org/browse/DM-21840) is still open because `None` values in data IDs are not prohibited by DB constraints yet, but it's long been true that conditions like "exposure with NULL filter" is for all intents and purposes forbidden, because if you do it the exposure becomes unqueryable.

python/lsst/daf/butler/_location.py
```diff
@@ -49,7 +49,12 @@ class Location:
 
     __slots__ = ("_datastoreRootUri", "_path", "_uri")
 
-    def __init__(self, datastoreRootUri: None | ResourcePathExpression, path: ResourcePathExpression):
+    def __init__(
+        self,
+        datastoreRootUri: None | ResourcePathExpression,
+        path: ResourcePathExpression,
+        _trusted_path: bool = False,
```
TallJimbo: `Location` itself is already sufficiently internal that I don't think you need to make this argument extra-private.

TallJimbo (approved):
Optimizations look great.

I'm not sure I understand how the transfer-from dry-run option works (maybe because I'm not sure precisely how the checking for existing datasets in the target butler works even without dry-run).

I do think we should drop the `ThreadPoolExecutor` commit for now.

timj:
> I'm not sure I understand how the transfer-from dry-run option works (maybe because I'm not sure precisely how the checking for existing datasets in the target butler works even without dry-run).

It does all the work that it would have to do to do the transfer but it doesn't copy any files or write to the target butler. I just modified how it works such that it will report that N datasets have already been transferred rather than dry run assuming that it always has to transfer everything.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: I got rid of the change. The point was to try to make my profiles take as long as if the files weren't already transferred, but in reality it was a minor blip and it's better if the dry-run implementation does report that all the datasets have already been transferred and so you know that it won't do anything if you run it for real.

python/lsst/daf/butler/_butler.py
timj: It does everything except update target butler.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: Right, I've removed this. The loop now does act like a dry run in that if the dataset is known to the datastore already it doesn't even attempt to do the copy (the copy would never happen because that's masked out later with dry-run). I'm now wondering why this check doesn't happen at the start of the `Datastore.transfer_from()` on the basis that the target datastore thinking it has the files obviates the need to actually check the file system for them.

PR #947, opened 2024-01-23
DM-42605: list chained collection children in search order

kfindeisen:
This PR fixes a bug where the `TABLE` mode of `butler query-collections` listed children in alphabetical order, obscuring the exact definition of a chained collection.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_cliCmdQueryCollections.py
+++ b/tests/test_cliCmdQueryCollections.py
@@ -225,9 +225,9 @@ class ChainedCollectionsTest(ButlerTestHelper, unittest.TestCase):
                 array(
                     (
                         ("calibration1", "CALIBRATION", ""),
-                        ("chain1", "CHAINED", "chain2"),
+                        ("chain1", "CHAINED", "tag1"),
                         ("", "", "run1"),
-                        ("", "", "tag1"),
+                        ("", "", "chain2"),
                         ("chain2", "CHAINED", "calibration1"),
                         ("", "", "run1"),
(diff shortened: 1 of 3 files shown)


tests/test_cliCmdQueryCollections.py
```diff
@@ -225,9 +225,9 @@ def testChained(self):
                 array(
                     (
                         ("calibration1", "CALIBRATION", ""),
-                        ("chain1", "CHAINED", "chain2"),
+                        ("chain1", "CHAINED", "tag1"),
                         ("", "", "run1"),
-                        ("", "", "tag1"),
+                        ("", "", "chain2"),
```
kfindeisen, line 230: It's a bit odd that this test didn't catch the bug; how was the value of `expected` chosen?

TallJimbo: I assume it's a classic case of:
- Developer knows they're breaking the test for good reason.
- Developer looks at the test output is after their change.
- Developer thinks about whether that it reasonable and adopts it (but is subject to confirmation bias).

I suppose it's good that we were at least trying to test this before, even if we failed to do it correctly.

TallJimbo (approved):

tests/test_cliCmdQueryCollections.py
kfindeisen: I would say this is a classic case of the problem definition being simple enough that you can work out the correct output in your head... 😬

PR #950, opened 2024-01-26
DM-42660: Allow direct configuration of repository index

dhirving:
Instead of requiring the repository index to be loaded from a file, allow direct configuration as a string in a new environment variable DAF_BUTLER_REPOSITORIES.

The file was problematic for services for a few reasons:
1. We don't have anywhere to host it -- currently it is stored in S3, but services using Butler server will not have the credentials to access it.
2. There are cache invalidation issues.  If it's configured directly, Phalanx can automatically restart services using it when the configuration changes.
3. Manually editing live configuration files is error-prone.  If this configuration lives in Phalanx, we are able to validate the configuration before deploying it.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler_repo_index.py
+++ b/python/lsst/daf/butler/_butler_repo_index.py
@@ -30,9 +30,11 @@ from __future__ import annotations
 __all__ = ("ButlerRepoIndex",)
 
 import os
-from typing import ClassVar
+from typing import Any, ClassVar
 
+import yaml
 from lsst.resources import ResourcePath
+from pydantic import TypeAdapter, ValidationError
 
 from ._config import Config
 from ._utilities.thread_safe_cache import ThreadSafeCache
@@ -41,10 +43,15 @@ from ._utilities.thread_safe_cache import ThreadSafeCache
 class ButlerRepoIndex:
     """Index of all known butler repositories.
 
-    The index of butler repositories is found by looking for a
-    configuration file at the URI pointed at by the environment
-    variable ``$DAF_BUTLER_REPOSITORY_INDEX``. The configuration file
-    is a simple dictionary lookup of the form:
+    The index of butler repositories can be configured in two ways:
+
+    1. By setting the environment variable ``DAF_BUTLER_REPOSITORY_INDEX`` to
+    the URI of a configuration file.
+    2. By setting the environment variable ``DAF_BUTLER_REPOSITORIES`` to the
+    contents of the configuration file as a string.
+
+    In either case, the configuration is a simple dictionary lookup of the
+    form:
 
     .. code-block:: yaml
 
@@ -56,9 +63,15 @@ class ButlerRepoIndex:
     """
 
     index_env_var: ClassVar[str] = "DAF_BUTLER_REPOSITORY_INDEX"
-    """The name of the environment variable to read to locate the index."""
+    """The name of the environment variable containing the URI of the index
+    configuration file.
+    """
+    repositories_env_var: ClassVar[str] = "DAF_BUTLER_REPOSITORIES"
+    """The name of the environment variable containing the configuration
+    directly as a string.
+    """
 
-    _cache: ClassVar[ThreadSafeCache[ResourcePath, Config]] = ThreadSafeCache()
+    _cache: ClassVar[ThreadSafeCache[str, dict[str, str]]] = ThreadSafeCache()
     """Cache of indexes. In most scenarios only one index will be found
     and the environment will not change. In tests this may not be true."""
 
@@ -67,17 +80,17 @@ class ButlerRepoIndex:
     every read."""
 
     @classmethod
-    def _read_repository_index(cls, index_uri: ResourcePath) -> Config:
+    def _read_repository_index(cls, index_uri: str) -> dict[str, str]:
         """Read the repository index from the supplied URI.
 
         Parameters
         ----------
-        index_uri : `lsst.resources.ResourcePath`
+        index_uri : `str`
             URI of the repository index.
 
         Returns
         -------
-        repo_index : `Config`
+        repo_index : `dict` [ `str` , `str` ]
             The index found at this URI.
 
         Raises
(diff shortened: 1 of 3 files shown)


tests/test_butler.py
```diff
@@ -677,6 +677,27 @@ def testConstructor(self) -> None:
             Butler.from_config("not_there")
         self.assertEqual(Butler.get_known_repos(), set())
 
+    def testDafButlerRepositories(self):
+        with unittest.mock.patch.dict(os.environ, {"DAF_BUTLER_REPOSITORIES": "label: https://someuri.com"}):
```
timj: Can you put a second entry in the LUT to make it clear that this is a YAML string that will have embedded new lines?

python/lsst/daf/butler/_butler_repo_index.py
```diff
@@ -41,10 +43,14 @@
 class ButlerRepoIndex:
     """Index of all known butler repositories.
 
-    The index of butler repositories is found by looking for a
-    configuration file at the URI pointed at by the environment
-    variable ``$DAF_BUTLER_REPOSITORY_INDEX``. The configuration file
-    is a simple dictionary lookup of the form:
+    The index of butler repositories can be configured in two ways:
+    1. By setting the environment variable ``DAF_BUTLER_REPOSITORY_INDEX`` to
```
timj, line 48: I think this needs a blank line before it. It currently renders as:

> The index of butler repositories can be configured in two ways: 1. By setting the environment variable DAF_BUTLER_REPOSITORY_INDEX to the URI of a configuration file. 2. By setting the environment variable DAF_BUTLER_REPOSITORIES to the contents of the configuration file as a string.

python/lsst/daf/butler/_butler_repo_index.py
```diff
@@ -41,10 +43,14 @@
...
+    the URI of a configuration file.
+    2. By setting the environment variable ``DAF_BUTLER_REPOSITORIES`` to the
+    contents of the configuration file as a string.
```
timj, line 51: If you are wanting this to be YAML you should document that here. Are you sure you don't want to require it to be JSON so that you don't have to have embedded newlines and it can parse a little faster? `DAF_BUTLER_REPOSITORY_INDEX` can work with JSON files as well.

python/lsst/daf/butler/_butler_repo_index.py
```diff
@@ -136,16 +129,29 @@ def _read_repository_index_from_environment(cls) -> Config:
...
+            direct_configuration = os.getenv(cls.repositories_env_var)
+
+            if index_uri and direct_configuration:
+                raise RuntimeError(
+                    f"Only one of the environment variables {cls.repositories_env_var} and"
+                    f" {cls.index_env_var} should be set."
+                )
+
+            if direct_configuration:
+                return cls._validate_configuration(yaml.safe_load(direct_configuration))
```
timj, line 143: Have I missed where this is cached or is it parsing the YAML every single time it's called? `_read_repository_index` has cached the result of reading the external file. YAML parsing is notoriously slow.

python/lsst/daf/butler/_butler_repo_index.py
```diff
@@ -107,26 +119,7 @@ def _read_repository_index(cls, index_uri: ResourcePath) -> Config:
...
-            Raised if the location of the index could not be determined.
-        """
-        index_uri = os.environ.get(cls.index_env_var)
-        if index_uri is None:
-            raise KeyError(f"No repository index defined in environment variable {cls.index_env_var}")
-        return ResourcePath(index_uri)
-
-    @classmethod
-    def _read_repository_index_from_environment(cls) -> Config:
+    def _read_repository_index_from_environment(cls) -> dict[str, str]:
```
timj, line 123: The docstring is wrong if you are changing to dict.

python/lsst/daf/butler/_butler_repo_index.py
```diff
@@ -67,7 +79,7 @@ class ButlerRepoIndex:
     every read."""
 
     @classmethod
-    def _read_repository_index(cls, index_uri: ResourcePath) -> Config:
+    def _read_repository_index(cls, index_uri: str) -> Config:
```
timj: You can change this to dict if you want to call `config.toDict()` here. `Config` was used since it understands ResourcePath and YAML vs JSON formats.

python/lsst/daf/butler/_butler_repo_index.py
```diff
@@ -222,10 +228,12 @@ def get_repo_uri(cls, label: str, return_label: bool = False) -> ResourcePath:
...
-            try:
-                index_uri = str(cls._get_index_uri())
-            except KeyError:
-                index_uri = "<environment variable not defined>"
-            raise KeyError(f"Label '{label}' not known to repository index at {index_uri}")
+            raise KeyError(f"Label '{label}' not known to repository index")
         return ResourcePath(repo_uri)
+
+    @classmethod
+    def _validate_configuration(cls, obj: Any) -> dict[str, str]:
```
timj, line 235: Can we validate once on first read and then know that we have a validated dict in the cache?

timj (changes requested):
I have a couple of questions:

* Parsing every time vs caching the validated dict.
* JSON vs YAML for the string version.

python/lsst/daf/butler/_butler_repo_index.py
dhirving: In the real deployment I am passing it as JSON, but I thought it was less confusing to have the same file format for both instead of trying to document "well if you do it directly it has to be JSON, but if it's the other one it can be YAML or JSON".

python/lsst/daf/butler/_butler_repo_index.py
dhirving: I didn't think caching was worth the added complexity to avoid parsing a few hundred bytes of in-memory data.  For instantiations via Butler() we are going to go off and read a file from the network, then do dozens of database round trips immediately after this.  LabeledButlerFactory only reads each label once.

timj (approved):
Changing to approved. I have been convinced that calling `yaml.safe_load()` is only going to happen a couple of times in the server when it launches.

Clients will be parsing the string with yaml.safe_load every time they call `Butler` and the theory is that this will be nothing compared to all the other activity needed to create a `Butler`.

PR #951, opened 2024-01-26
DM-42676: Fix quantum backed butler trust mode with chained datastore

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_quantum_backed.py
+++ b/python/lsst/daf/butler/_quantum_backed.py
@@ -357,9 +357,7 @@ class QuantumBackedButler(LimitedButler):
         # TODO: We need to inform `Datastore` here that it needs to support
         # predictive reads; This only really works for file datastore but
         # we need to try everything in case there is a chained datastore.
-        for this_datastore in getattr(datastore, "datastores", [datastore]):
-            if hasattr(this_datastore, "trustGetRequest"):
-                this_datastore.trustGetRequest = True
+        datastore._set_trust_mode(True)
 
         if datastore_records is not None:
             datastore.import_records(datastore_records)
--- a/python/lsst/daf/butler/_quantum_backed.py
+++ b/python/lsst/daf/butler/_quantum_backed.py
@@ -352,12 +352,15 @@ class QuantumBackedButler(LimitedButler):
                 datasets=_DatasetRecordStorageManagerDatastoreConstructionMimic,  # type: ignore
                 universe=dimensions,
             )
-        # TODO: We need to inform `Datastore` here that it needs to support
-        # predictive reads; right now that's a configuration option, but after
-        # execution butler is retired it could just be a kwarg we pass here.
-        # For now just force this option as we cannot work without it.
-        butler_config["datastore", "trust_get_request"] = True
         datastore = Datastore.fromConfig(butler_config, bridge_manager, butler_root)
+
+        # TODO: We need to inform `Datastore` here that it needs to support
+        # predictive reads; This only really works for file datastore but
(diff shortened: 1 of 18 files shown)


python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -1116,6 +1116,11 @@ def transfer_from(
             # child datastores.
             source_datastores = (source_datastore,)
 
+        refs = list(refs)
```
timj: This seems wrong to me but we've declared `refs` to be `Iterable` and that doesn't support `len`.

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -1116,6 +1116,11 @@ def transfer_from(
...
+        if len(refs) == 0:
```
timj:
```python
if not refs:
```
does pass mypy for an Iterator but not sure if that's safe.

andy-slac: If `refs` is a list then `not refs` and `len(refs) == 0` work identically.

python/lsst/daf/butler/_quantum_backed.py
```diff
@@ -352,12 +352,15 @@ def _initialize(
...
-        # For now just force this option as we cannot work without it.
-        butler_config["datastore", "trust_get_request"] = True
         datastore = Datastore.fromConfig(butler_config, bridge_manager, butler_root)
+
+        # TODO: We need to inform `Datastore` here that it needs to support
+        # predictive reads; This only really works for file datastore but
+        # we need to try everything in case there is a chained datastore.
+        for this_datastore in getattr(datastore, "datastores", [datastore]):
+            if hasattr(this_datastore, "trustGetRequest"):
+                this_datastore.trustGetRequest = True
```
andy-slac: So, if by some chance non-chained datastore adds `datastores` attribute, we won't try to set its `trustGetRequest`? And it will also crash if `datastore.datastores` is not an Iterable?

andy-slac (approved):
Looks good, one question about possible issues with `.datastores`.

python/lsst/daf/butler/datastores/chainedDatastore.py
timj: Yes. That's my point, but in the more general case for Iterator am I setting myself up for pain if I do that. I am wondering if we should make the type annotation be `Collection` instead and therefore guarantee that `len` works, or do something like:

```python
if not isinstance(refs, Sized):
    refs = list(refs)
```

python/lsst/daf/butler/_quantum_backed.py
timj: If `datastores` is not an iterable we are in trouble indeed. I was pondering adding `datastores` to `Datastore` and making it return `[self]` by default. Then it would be clear that it has a special purpose that only something like `ChainedDatastore` can override.

andy-slac: Maybe just add "private" method `_set_trust_mode()` to the base class?

python/lsst/daf/butler/datastores/chainedDatastore.py
andy-slac: If we need `len` then `Collection` annotation is probably better.

python/lsst/daf/butler/_quantum_backed.py
timj: I will do that. I wasn't sure I wanted to make it something that top-level given we can't think of anything outside FileDatastore caring.

PR #953, opened 2024-01-30
DM-42636: add group and day_obs first-class dimensions

TallJimbo:
Depends on lsst/astro_metadata_translator#71.

## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/transfers/_yaml.py
```diff
@@ -310,6 +332,44 @@ def __init__(self, stream: IO, registry: SqlRegistry):
...
+                                # The instrument is already present so use
+                                # the day_obs_offset from that record.
+                                record["day_obs_offset"] = instrument_offsets[name]
+                                continue
+
+                            # Try to ask the registered instrument class.
+                            try:
+                                instrument_class = doImportType(record["class_name"])
+                            except ImportError:
+                                record["day_obs_offset"] = 0
```
TallJimbo: I think in these failure cases we should use `None` instead of `0`.  If we ever have to go back and repair an import that failed this way, it'll make it easier to spot the ones that need fixing.

python/lsst/daf/butler/dimensions/_records.py
```diff
@@ -112,9 +112,12 @@ def _createSimpleRecordSubclass(definition: DimensionElement) -> type[SpecificSe
     if definition.spatial:
         members["region"] = (str, ...)
 
-    # mypy does not seem to like create_model
+    # For the new derived class name need to convert to camel case.
+    # so "day_obs" -> "DayObs".
+    derived_name = "".join([part.capitalize() for part in definition.name.split("_")])
```
TallJimbo, line 117: Was this always a problem for `physical_filter` in the past, too?

doc/lsst.daf.butler/dimensions.rst
```diff
@@ -81,3 +81,34 @@ The same is not true of regions (especially regions on the sphere), at least not
...
+   Updated storage classes for many dimensions.
+   Defined governor dimensions (``instrument`` and ``skymap``).
+   Added ``day_obs`` and ``seq_num`` to ``visit`` and ``exposure``.
+   Renamed ``exposure.name`` to ``exposure.obs_id``.
+2. Modified how visits are defined
+   (Added the default ``visit_system`` to ``instrument``,
+   added ``visit_system_membership`` dimension, removed ``visit_system`` from ``visit`` dimension).
+   Added ``has_simulated``, ``azimuth``, ``seq_start``, and ``seq_end`` to ``exposure``.
+   Added ``seq_num``, ``azimuth`` to ``visit``.
+3. Updated the length of the ``observation_reason`` from 32 to 68 in ``exposure`` and ``visit`` fields.
```
TallJimbo:
```suggestion
3. Updated the length of the ``observation_reason`` field from 32 to 68 in ``exposure`` and ``visit``.
```

TallJimbo (commented):
Looks great; I especially appreciate the effort to reconstruct the dimension universe changelog.

(I can't hit approve because I created the PR.)

python/lsst/daf/butler/dimensions/_records.py
timj: Yes. Most people don't notice since the class is only constructed transiently to validate the JSON content as it's converted to a real DimensionRecord so you only get to see it if your record content fails validation (and physical filter mostly has not changed over the years)

python/lsst/daf/butler/transfers/_yaml.py
timj: Ok. Will see what happens. For reasons I can't explain validation with `None` for the timespan does work so I need to understand why it wasn't working in ci_middleware.

TallJimbo (commented):
New changes look good!  (Approved again in spirit.)

PR #954, opened 2024-01-31
DM-42733: minor improvement spin-offs from query system rewrite

TallJimbo:
## Checklist

- [x] ran Jenkins
- [ ] ~added a release note for user-visible changes to `doc/changes`~ (nothing here should be user-visible)

### DIFF
--- a/python/lsst/daf/butler/_query_results.py
+++ b/python/lsst/daf/butler/_query_results.py
@@ -389,17 +389,17 @@ class DataCoordinateQueryResults(Iterable[DataCoordinate]):
         raise NotImplementedError()
 
     @abstractmethod
-    def limit(self, limit: int, offset: int | None = 0) -> DataCoordinateQueryResults:
+    def limit(self, limit: int | None = None, offset: int = 0) -> DataCoordinateQueryResults:
         """Make the iterator return limited number of records.
 
         Parameters
         ----------
-        limit : `int`
-            Upper limit on the number of returned records.
-        offset : `int` or `None`, optional
+        limit : `int` or `None`, optional
+            Upper limit on the number of returned records.  `None` (default) is
+            no limit.
+        offset : `int`, optional
             The number of records to skip before returning at most ``limit``
-            records.  `None` is interpreted the same as zero for backwards
-            compatibility.
+            records.
 
         Returns
         -------
@@ -683,17 +683,16 @@ class DimensionRecordQueryResults(Iterable[DimensionRecord]):
         raise NotImplementedError()
 
     @abstractmethod
-    def limit(self, limit: int, offset: int | None = 0) -> DimensionRecordQueryResults:
+    def limit(self, limit: int | None = None, offset: int = 0) -> DimensionRecordQueryResults:
         """Make the iterator return limited number of records.
 
         Parameters
         ----------
-        limit : `int`
+        limit : `int` or `None`, optional
             Upper limit on the number of returned records.
-        offset : `int` or `None`
+        offset : `int`, optional
             The number of records to skip before returning at most ``limit``
-            records.  `None` is interpreted the same as zero for backwards
-            compatibility.
+            records.
 
         Returns
         -------
--- a/python/lsst/daf/butler/nonempty_mapping.py
+++ b/python/lsst/daf/butler/nonempty_mapping.py
@@ -1,0 +1,99 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This software is dual licensed under the GNU General Public License and also
+# under a 3-clause BSD license. Recipients may choose which of these licenses
+# to use; please see the files gpl-3.0.txt and/or bsd_license.txt,
+# respectively.  If you choose the GPL option then the following text applies
+# (but note that there is still no warranty even if you opt for BSD instead):
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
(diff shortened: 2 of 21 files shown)


TallJimbo:
Docker build is failing because I apparently used a feature (`string_agg`) that wasn't in SQLite until [3.44.2](https://www.sqlite.org/releaselog/3_44_2.html).

Can we bump the SQLite version in the docker build easily, or should I add a new `Database` method to use `string_agg` in PostgreSQL and `group_agg` in SQLite?

dhirving:
> Can we bump the SQLite version in the docker build easily

It doesn't look like sqlite is one of the packages available in debian-backports, so "easily" no because the version you want isn't available in the package manager.  If there's some other way to replace the system sqlite library with a newer version we can use it.

We don't "have" to run Debian if you need bleeding edge versions of packages, that's just what SQRE is doing for all of their services.  I decided to follow them so we can free-ride on them knowing what is going on with security and compatibility with other tooling.

It should be relatively straightforward to switch it over to be based on minimamba with rubin-env, just the image will get bigger (which doesn't really matter other than there's more surface area for security issues).  Also I don't know enough about conda so I'm not sure if it's possible to get a reproducible set of packages from installing rubin-env because they don't pin versions in their recipe.

python/lsst/daf/butler/_query_results.py
```diff
@@ -389,17 +389,16 @@ def order_by(self, *args: str) -> DataCoordinateQueryResults:
...
 
     @abstractmethod
-    def limit(self, limit: int, offset: int | None = 0) -> DataCoordinateQueryResults:
+    def limit(self, limit: int | None = None, offset: int = 0) -> DataCoordinateQueryResults:
         """Make the iterator return limited number of records.
 
         Parameters
         ----------
-        limit : `int`
+        limit : `int` or `None`, optional
```
timj, line 397: Maybe state that `None` means all records are returned?

python/lsst/daf/butler/nonempty_mapping.py
```diff
@@ -0,0 +1,98 @@
...
+from collections.abc import Callable, Iterator, Mapping
+from typing import Any, TypeVar, overload
+
+_K = TypeVar("_K")
+_T = TypeVar("_T")
+_V = TypeVar("_V", covariant=True)
+
+
+class NonemptyMapping(Mapping[_K, _V]):
+    """A `Mapping` that implicitly adds values (like `defaultdict`) but treats
```
timj:
```suggestion
    """A `Mapping` that implicitly adds values (like `~collections.defaultdict`) but treats
```
(which will likely make the line too long so this change can't be accepted as-is).

python/lsst/daf/butler/nonempty_mapping.py
```diff
@@ -0,0 +1,98 @@
...
+
+    Parameters
+    ----------
+    default_factory : `~collections.abc.Callable`
+        A callable that takes no arguments and returns a new instance of the
+        value type.
+
+    Notes
+    -----
+    Unlike `defaultdict`, this class implements only `collections.abc.Mapping`,
```
timj: `~collections.defaultdict`

python/lsst/daf/butler/nonempty_mapping.py
```diff
@@ -0,0 +1,98 @@
...
+    not `~collections.abc.MutableMapping`, and hence it can be modified only by
+    invoking ``__getitem__`` with a key that does not exist.  Is is expected
```
timj:
```suggestion
    invoking ``__getitem__`` with a key that does not exist.  It is expected
```
?

TallJimbo:
I generally like the idea of getting this container based on rubin-env ones, because it seems like that shrinks the space of environments we're trying to support.  But it does sound like more than I wanted to do on this branch.  @timj, I think it's your call.

timj:
I see that Debian trixie (the next version) has 3.45 so there is hope for the future. How much work is it to support the old sqlite? Does @rra have any thoughts on trying to get a newer sqlite into the contanier?

rra:
I agree with basically everything @dhirving said: that version of SQLite is extremely new, does not appear in any released version of Debian, and hasn't been backported. If you want to use a prepackaged version, you would need to build a Docker image from Debian testing (an unreleased version of Debian). I'm not sure you want to do that.

You could instead switch to the stack image, which will get you a much older base OS image with dubious security support but a conda environment, if conda packages the new version of SQLite. SQuaRE avoids the stack images whenever possible because they are huge and are not supported by the broader open source community, so involve some reinventing of wheels to get things onto them, but the deeper you get into stack package land, the more it would make sense to go down that path.

Another option would be to install SQLite via some other mechanism, but you would still need to find a source of packages that is newer than the Debian stable release from last summer.

TallJimbo:
Sounds like I should just support the old SQLite; I don't think it should be hard, and I only kept this conversation going in case we thought it was better otherwise to use the stack images.

dhirving:
> Sounds like I should just support the old SQLite

Another option that isn't too bad is just building sqlite from source in the Docker container.  Let me know if you want me to set it up.

```
RUN wget https://www.sqlite.org/2024/sqlite-autoconf-3450100.tar.gz
RUN tar -xzf sqlite-autoconf-3450100.tar.gz
WORKDIR sqlite-autoconf-3450100
RUN ./configure
RUN make
RUN make install
RUN ldconfig /usr/local/lib
CMD python -c "import sqlite3;print(sqlite3.sqlite_version)"
# 3.45.1
```

tests/test_nonempt_mapping.py
```diff
@@ -0,0 +1,52 @@
+# This file is part of daf_butler.
```
timj: Maybe rename this file to nonempty rather than nonempt?

timj (approved):

TallJimbo:
Supporting older SQLite was even easier than I expected, because SQLAlchemy already had an abstraction that resolved to the right thing for each dialect.

PR #961, opened 2024-02-16
DM-42934: Run existing unit tests against RemoteButler

dhirving:
Added a new Butler subclass HybridButler that delegates methods to RemoteButler where implemented, and uses DirectButler for the rest.  Extracted the test server setup from test_server.py into a reusable function.  Modified test_butler.py to run tests against Butler server via HybridButler.

This revealed a few minor incompatibilities between RemoteButler and DirectButler, so updated RemoteButler to conform with the DirectButler behavior.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/tests/hybrid_butler.py
+++ b/python/lsst/daf/butler/tests/hybrid_butler.py
@@ -1,0 +1,433 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This software is dual licensed under the GNU General Public License and also
+# under a 3-clause BSD license. Recipients may choose which of these licenses
+# to use; please see the files gpl-3.0.txt and/or bsd_license.txt,
+# respectively.  If you choose the GPL option then the following text applies
+# (but note that there is still no warranty even if you opt for BSD instead):
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import annotations
+
+from collections.abc import Collection, Iterable, Mapping, Sequence
+from contextlib import AbstractContextManager
+from typing import Any, TextIO, cast
+
+from lsst.resources import ResourcePath, ResourcePathExpression
+
+from .._butler import Butler
+from .._dataset_existence import DatasetExistence
+from .._dataset_ref import DatasetId, DatasetRef
+from .._dataset_type import DatasetType
+from .._deferredDatasetHandle import DeferredDatasetHandle
+from .._file_dataset import FileDataset
+from .._limited_butler import LimitedButler
+from .._query import Query
+from .._storage_class import StorageClass
+from .._timespan import Timespan
+from ..datastore import DatasetRefURIs
+from ..dimensions import DataCoordinate, DataId, DimensionGroup, DimensionRecord, DimensionUniverse
+from ..registry import CollectionArgType, Registry
+from ..remote_butler import RemoteButler
+from ..transfers import RepoExportContext
+
+
+class HybridButler(Butler):
+    """A `Butler` that delegates methods to internal RemoteButler and
+    DirectButler instances.  Intended to allow testing of RemoteButler before
+    its implementation is complete, by delegating unsupported methods to
+    DirectButler.
+    """
+
+    _remote_butler: RemoteButler
+    _direct_butler: Butler
+
+    def __new__(cls, remote_butler: RemoteButler, direct_butler: Butler) -> HybridButler:
+        self = cast(HybridButler, super().__new__(cls))
+        self._remote_butler = remote_butler
+        self._direct_butler = direct_butler
+        self._datastore = direct_butler._datastore
+        return self
+
+    def isWriteable(self) -> bool:
+        return self._remote_butler.isWriteable()
+
+    def _caching_context(self) -> AbstractContextManager[None]:
+        return self._direct_butler._caching_context()
+
+    def transaction(self) -> AbstractContextManager[None]:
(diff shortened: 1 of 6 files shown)


python/lsst/daf/butler/tests/hybrid_butler.py
```diff
@@ -0,0 +1,431 @@
...
+    """A `Butler` that delegates methods to internal RemoteButler and
+    DirectButler instances.  Intended to allow testing of RemoteButler before
+    its implementation is complete, by delegating unsupported methods to
+    DirectButler.
+    """
+
+    _remote_butler: RemoteButler
+    _direct_butler: Butler
+
+    def __new__(cls, remote_butler: RemoteButler, direct_butler: Butler) -> "HybridButler":
```
timj:
```suggestion
    def __new__(cls, remote_butler: RemoteButler, direct_butler: Butler) -> Self:
```
Does `typing.Self` work here? (maybe also need `from __future__ import annotations` to remove the quotes?)

python/lsst/daf/butler/tests/hybrid_butler.py
```diff
@@ -0,0 +1,431 @@
...
+        datasetRefOrType: DatasetRef | DatasetType | str,
+        /,
+        dataId: DataId | None = None,
+        *,
+        parameters: dict | None = None,
+        collections: Any = None,
+        storageClass: str | StorageClass | None = None,
+        **kwargs: Any,
+    ) -> DeferredDatasetHandle:
+        return self._direct_butler.getDeferred(
```
timj, line 102: `getDeferred` in theory is simple to implement in remote butler since it is a deferred get and get works already. The minor complication is that the direct butler implementation does a check to make sure the get is likely to work (calling datastore at the moment but it should really call `self.exists()` and `self.stored()`) before it returns the deferred handle (this is so that you don't spend hours doing some processing only to find that the get fails). This might be something to add fairly soon (not on this ticket) along with `exists` and `stored`.

python/lsst/daf/butler/tests/hybrid_butler.py
```diff
@@ -0,0 +1,431 @@
...
+        )
+
+    def _clone(
+        self,
+        *,
+        collections: Any = None,
+        run: str | None = None,
+        inferDefaults: bool = True,
+        **kwargs: Any,
+    ) -> Butler:
```
timj: Is `Self` better here?

tests/test_butler.py
```diff
@@ -205,17 +211,32 @@ def assertGetComponents(
...
+            removeTestTempDir(self.root)
 
     def create_butler(
         self, run: str, storageClass: StorageClass | str, datasetTypeName: str
-    ) -> tuple[DirectButler, DatasetType]:
+    ) -> tuple[Butler, DatasetType]:
+        """Create a Butler for the test repository and insert some test data
+        into it.
+        """
         butler = Butler.from_config(self.tmpConfigFile, run=run)
```
timj: Can this call `self.create_empty_butler()` here? Or is there some restriction on this having to be a direct butler? If it can't call it, maybe add a comment to that effect?

tests/test_butler.py
```diff
@@ -205,17 +211,32 @@ def assertGetComponents(
...
         assert isinstance(butler, DirectButler), "Expect DirectButler in configuration"
 
         collections = set(butler.registry.queryCollections())
         self.assertEqual(collections, {run})
 
+        return self._setup_butler_data(butler, storageClass, datasetTypeName)
+
+    def create_empty_butler(self, run: str | None = None, writeable: bool | None = None):
```
timj:
```suggestion
    def create_empty_butler(self, run: str | None = None, writeable: bool | None = None) -> Butler:
```

tests/test_butler.py
```diff
@@ -723,12 +750,13 @@ def testCompositePutGetConcrete(self) -> None:
         self.assertIn("423", str(uri), f"Checking visit is in URI {uri}")
 
         # Predicted dataset
-        dataId = {"instrument": "DummyCamComp", "visit": 424}
-        uri, components = butler.getURIs(datasets[0].datasetType, dataId=dataId, predict=True)
-        self.assertFalse(components)
-        self.assertIsInstance(uri, ResourcePath)
-        self.assertIn("424", str(uri), f"Checking visit is in URI {uri}")
-        self.assertEqual(uri.fragment, "predicted", f"Checking for fragment in {uri}")
+        if self.predictionSupported:
```
timj, line 748: Is this a temporary restriction on remote butler or a permanent one?

tests/test_butler.py
```diff
@@ -2478,6 +2507,63 @@ def test_fallback(self) -> None:
...
+
+    # Pickling not yet implemented for RemoteButler/HybridButler.
+    @unittest.expectedFailure
+    def testPickle(self) -> None:
+        return super().testPickle()
+
+    def testStringification(self) -> None:
+        # RemoteButler does not have datastore/registry strings like
+        # DirectButler does.
+        pass
```
timj: Can we have a bespoke test here that checks that RemoteButler stringifies into something reasonable? Maybe with the URL of the server?

timj (approved):
Looks great. I really like the hybrid butler approach -- makes it very clear what works with remote server.

python/lsst/daf/butler/tests/hybrid_butler.py
dhirving: Self doesn't for some reason... it causes mypy to throw "expected Self but got HybridButler".

`from __future__ import annotations` works though, thanks -- I could not for the life of me figure out why this exact same pattern worked in DirectButler and RemoteButler but not in this file.

tests/test_butler.py
dhirving: Yeah that makes sense and will let me get rid of the other method override below.  I didn't discover the need for `create_empty_butler` until later in the process so I didn't think about it.

tests/test_butler.py
dhirving: It's not currently implemented.  I don't really understand the use cases for that feature so I'm not sure whether it makes sense to implement in the future or not.

timj: Yes, it's a tricky one. People wanted to know what the path would be if they were to store the dataset. In client/server this becomes a bit odder since you have to decide what it comes back as since returning a signed URL to something that doesn't exist seems odd. 

We have discussed in the past whether getURI should have the option of returning the server-side URI (s3://bucket/path_in_store/file.fits) rather than the client-side signed URL -- presumably this is assumed to be a bad idea because the bucket name shouldn't leak to an untrusted user? Maybe for predicted URIs we could return something like `butler://datastore_name/path_in_store/file.fits#predicted` so that people could see what the final filename template would resolve to. I'm not sure how many users care about this at this point.

TallJimbo: Maybe this could become a `DirectButler`-only interface?  It certainly won't be the last one; we might even want to go ahead and move the entire write interface from `Butler` to `DirectButler` at some point.

PR #964, opened 2024-02-20
DM-42947: Implement RemoteButler.retrieveArtifacts

dhirving:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastores/file_datastore/retrieve_artifacts.py
+++ b/python/lsst/daf/butler/datastores/file_datastore/retrieve_artifacts.py
@@ -1,0 +1,72 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This software is dual licensed under the GNU General Public License and also
+# under a 3-clause BSD license. Recipients may choose which of these licenses
+# to use; please see the files gpl-3.0.txt and/or bsd_license.txt,
+# respectively.  If you choose the GPL option then the following text applies
+# (but note that there is still no warranty even if you opt for BSD instead):
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import annotations
+
+__all__ = ("determine_destination_for_retrieved_artifact",)
+
+from lsst.resources import ResourcePath, ResourcePathExpression
+
+
+def determine_destination_for_retrieved_artifact(
+    destination_directory: ResourcePath, source_path: ResourcePath, preserve_path: bool
+) -> ResourcePath:
+    """Determine destination path for an artifact retrieved from a datastore.
+
+    Parameters
+    ----------
+    destination_directory : `ResourcePath`
+        Path to the output directory where file will be stored.
+    source_path : `ResourcePath`
+        Path to the source file to be transferred.  This may be relative to the
+        datastore root, or an absolute path.
+    preserve_path : `bool`, optional
+        If `True` the full path of the artifact within the datastore
+        is preserved. If `False` the final file component of the path
+        is used.
+
+    Returns
+    -------
+    destination_uri : `~lsst.resources.ResourcePath`
+        Absolute path to the output destination.
+    """
+    destination_directory = destination_directory.abspath()
+
+    target_path: ResourcePathExpression
+    if preserve_path:
+        target_path = source_path
+        if target_path.isabs():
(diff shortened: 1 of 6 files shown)


python/lsst/daf/butler/datastores/file_datastore/retrieve_artifacts.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def determine_destination_for_retrieved_artifact(
+    destination_directory: ResourcePath, source_path: ResourcePath, preserve_path: bool
+) -> ResourcePath:
+    """Determine destination path for an artifact retrieved from a datastore.
+
+    Parameters
+    ----------
+    destination_directory : `ResourcePath`
```
timj, line 42:
```suggestion
    destination_directory : `~lsst.resources.ResourcePath`
```
I don't think Sphinx resolves them otherwise because `ResourcePath` is not part of `daf_butler`. It only really matters if this docstring turns up in the sphinx and in the longer term we keep dreaming that sphinx will pick up the type annotations and we can remove the type from here...

python/lsst/daf/butler/datastores/file_datastore/retrieve_artifacts.py
```diff
@@ -0,0 +1,72 @@
...
+    Returns
+    -------
+    destination_uri : `ResourcePath`
+        Absolute path to the output destination.
+    """
+    destination_directory = destination_directory.abspath()
+
+    target_path: ResourcePathExpression
+    if preserve_path:
+        target_path = source_path
```
timj, line 61: I know this was copied from above but looking at it again it might be clearer from a typing perspective as:

```python
target_path: str
if preserve_path:
    if source_path.isabs():
        target_path = source_path.relativeToPathRoot
    else:
        target_path = source_path.path
else:
    target_path = source_path.basename()
```
then target_path is a simple string and doesn't have to be a ResourcePath in one branch. Thoughts?

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -428,8 +433,26 @@ def retrieveArtifacts(
...
+        output_uris: list[ResourcePath] = []
+        for ref in refs:
+            file_info = _to_file_payload(self._get_file_info_for_ref(ref)).file_info
+            for file in file_info:
+                source_uri = ResourcePath(str(file.url))
+                relative_path = ResourcePath(file.datastoreRecords.path, forceAbsolute=False)
+                target_uri = determine_destination_for_retrieved_artifact(
+                    destination, relative_path, preserve_path
+                )
+                target_uri.transfer_from(source_uri, transfer="copy", overwrite=overwrite)
```
timj, line 454: This loop is organized slightly different to `FileDatastore` in that it does the transfer in the main loop and can't report how many files it's about to transfer.  The debug message was there before the transfer for cases where people have mistakenly asked for 10,000 datasets and are wondering why it's taking so long. We have had discussions in the past about refusing to do the transfers if there are too many refs and also implementing parallelism in the transfers (which is why it was written initially to do the transfers after all the database queries so it could be modified to use futures.

I realize looking at the original now that it was never upgraded to use the bulk query interface `_get_stored_records_associated_with_refs` but instead uses the much slower per-ref API that we had originally. This is also another slowdown -- we have had people wondering what is going on even with a few hundred datasets so a reorganization of this implementation to have a bulk endpoint ("give me the payloads of N refs") with futures to download with 10 threads will likely be important at some point.

timj (approved):
Looks okay. I think there will need to be a refactor of it at some point in the future to be more efficient (we have had people complaining that it's slow in direct butler when the number of refs is reasonably large)

python/lsst/daf/butler/remote_butler/_remote_butler.py
dhirving: This is intentional here to some extent... because of URL expiration, we don't want to download a huge number of URLs upfront and then download all of them at the end.  If the files are large the URLs will expire before we start downloading the later ones.

I think a bulk/parallel version of this would probably request like 10 URLs at a time from the server, instead of all of them.

python/lsst/daf/butler/datastores/file_datastore/retrieve_artifacts.py
dhirving: FYI I asked Jonathan about this at JTM.  The latest version of the documentation tooling does do that, just science pipelines has not upgraded to use it yet.

python/lsst/daf/butler/datastores/file_datastore/retrieve_artifacts.py
dhirving: seems reasonable

python/lsst/daf/butler/remote_butler/_remote_butler.py
timj: Can you add a comment here to say that the reason is because of expiration?

Maybe also add a verbose log message after the loop to report how many files were transferred.

python/lsst/daf/butler/datastores/file_datastore/retrieve_artifacts.py
dhirving: Actually should it be `unquoted_path` instead?  I'm not sure on the vagaries of ResourcePath quoting and unquoting.  I might just leave it as it was since otherwise it's going to get immediately converted back into a ResourcePath in join() anyway.

timj: Yes, `join` uses `unquoted_path` (so it does convert the resource path to a string inside there).

https://github.com/lsst/resources/blob/main/python/lsst/resources/_resourcePath.py#L742

Leaving it is fine, it is just a bit more convoluted with the typing as it currently stands.

PR #966, opened 2024-02-26
DM-42740: rewrite new query system interfaces and division of responsibilities

TallJimbo:
## Checklist

- [x] ran Jenkins
- [ ] ~added a release note for user-visible changes to `doc/changes`~ (not yet user-visibile)


python/lsst/daf/butler/queries/tree/_query_tree.py
```diff
@@ -0,0 +1,314 @@
...
+
+        Returns
+        -------
+        result : `QueryTree`
+            A new tree that joins in the dataset search.
+
+        Notes
+        -----
+        If this dataset type was already joined in, the new `DatasetSearch`
+        replaces the old one.
```
dhirving, line 244: I wonder if it should raise an exception instead?  Presumably the existing DatasetSearch was added for reasons, so silently discarding it could lead to potentially confusing bugs.  I see that you guard against this at the higher level but as the code churns it might be good to check it here too.

python/lsst/daf/butler/queries/tree/_query_tree.py
```diff
@@ -0,0 +1,314 @@
...
+    this type is used as a value.
+    """
+
+    collections: tuple[str, ...]
+    """The collections to search.
+
+    Order matters if this dataset type is later referenced by a `FindFirst`
+    operation.  Collection wildcards are always resolved before being included
+    in a dataset search.
+    """
```
dhirving, line 92: How important is this constraint?  I think it may save some round-trips and complexity in the client if the server is able to resolve any collection wildcards during the normal course of the query execution, instead of the client needing to pre-resolve things itself.

(One other reason to defer resolution is that users might expect that the collections used during query execution are those that are present at the time the query is executed, not the ones at the time the query is defined.  This might come up in use cases like prompt processing where you do repeated queries in a long-lived process.)

python/lsst/daf/butler/queries/tree/_query_tree.py
```diff
@@ -0,0 +1,314 @@
...
+    carry information about the columns that are actually included in result
+    rows, or what kind of butler primitive (e.g. `DataCoordinate` or
+    `DatasetRef`) those rows might be transformed into.
+    """
+
+    dimensions: DimensionGroup
+    """The dimensions whose keys are joined into the query.
+    """
+
+    datasets: Mapping[str, DatasetSearch] = pydantic.Field(default_factory=dict)
```
dhirving, line 133: Since this is declaring the wire format for the data, it might be a good idea to declare all these `Mapping`s as `dict` instead.  Why leave the reader wondering if this could potentially be an instance of some other class in some circumstances?  Any non-dict type is going to become a dict on the other side of the JSON boundary, so we may as well declare that here to ensure that things are consistent on both sides.

python/lsst/daf/butler/queries/tree/_query_tree.py
```diff
@@ -0,0 +1,314 @@
...
+    tree is validated (since it requires a database query) and hence must be
+    checked later.
+    """
+
+    storage_class_name: str | None
+    """Name of the storage class to use when returning `DatasetRef` results.
+
+    May be `None` if the dataset is only used as a constraint or to return
+    columns that do not include a full dataset type.
+    """
```
dhirving, line 108: Just FYI the strategy I've been using so far on StorageClass is that the client never sends the server anything about a StorageClass -- the server sends back a ref with the default storage class for the data type and the client overrides it if required.  This theoretically lets us support clients loading data using custom/experimental storage classes not known to the server.

I don't think it's harmful to have this get sent to the server and ignored, though at first glance I would have expected this bit of data to be attached to the SingleTypeDatasetQueryResults object and not in this part of the QueryTree... since the dataset types provided in the call to `datasets()` define the output from the query, not any intermediate joins.

python/lsst/daf/butler/queries/_dataset_query_results.py
```diff
@@ -0,0 +1,236 @@
...
+    def has_dimension_records(self) -> bool:
+        """Whether all data IDs in this iterable contain dimension records."""
+        raise NotImplementedError()
+
+    @abstractmethod
+    def with_dimension_records(self) -> DatasetQueryResults:
+        """Return a results object for which `has_dimension_records` is
+        `True`.
+        """
+        raise NotImplementedError()
```
dhirving, line 79: I'm not sure this `DatasetQueryResults` base class is necessary.  If `Query.datasets` always returned a `ChainedDatasetQueryResults` (potentially of length 1, and presumably renamed to `DatasetQueryResults`) then:
* The `DatasetQueryResults` base class could be removed, leaving the reader to with less of a multiple inheritance tangle to puzzle out (there's at least one diamond in the current setup.)
* There would be fewer code paths to go down -- users and maintainers wouldn't have to worry about potentially varying behavior between the 'direct' case and the 'chained' case.  All test coverage would be covering the same code paths.
* `by_dataset_type` could be removed from SingleTypeDatasetQueryResults (and possibly other methods with a bit of restructuring that becomes possible when the interfaces of SingleTypeDatasetQueryResults and ChainedDatasetQueryResults don't have to be identical.)
* SingleTypeDatasetQueryResults could potentially become private to this file (if all available methods were forwarded from ChainedDatasetQueryResults, which is probably desirable for consistency anyway.)

python/lsst/daf/butler/queries/_dataset_query_results.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def __init__(self, by_dataset_type: tuple[SingleTypeDatasetQueryResults, ...]):
+        self._by_dataset_type = by_dataset_type
+
+    def __iter__(self) -> Iterator[DatasetRef]:
+        return itertools.chain.from_iterable(self._by_dataset_type)
+
+    def by_dataset_type(self) -> Iterator[SingleTypeDatasetQueryResults]:
+        # Docstring inherited.
+        return iter(self._by_dataset_type)
```
dhirving, line 186: This might be easier to use if it was more like `get_by_dataset_type(self, datasetType) -> Iterator[DatasetRef]` instead of returning the types in a random order (if wildcards/ellipsis are involved.)

This also kind of raises the issue of whether giving public access to the inner results objects is the right thing here, since the fact that the queries are separable might end up being an implementation detail  -- we probably want to be able to run all the queries as a batch.  Maybe this could be typed as `Iterator[Iterable[DatasetRef]]` to discourage people from using the "extra" methods on the inner objects.

(If the long-term plan is to run queries individually and expose the individual query objects, I would argue that ChainedDatasetQueryResults and the collection-related overload of `Query.datasets` shouldn't even exist -- the user can write an extra 1-2 lines to loop over the output from `queryDatasetTypes` themselves.)

python/lsst/daf/butler/queries/_query.py
```diff
@@ -0,0 +1,652 @@
...
+    ) -> SingleTypeDatasetQueryResults: ...  # pragma: no cover
+
+    @overload
+    def datasets(
+        self,
+        dataset_type: Iterable[str | DatasetType] | EllipsisType,
+        collections: str | Iterable[str] | None = None,
+        *,
+        find_first: bool = True,
+    ) -> DatasetQueryResults: ...  # pragma: no cover
```
dhirving, line 224: If you don't take me up on the suggestion above to hide the concept of SingleTypeDatasetQueryResults, I wonder if these overloads should actually be two separate methods (`datasets()` and `datasets_many()` or something).  

Just kind of confusing getting potentially different objects with different sets of methods available from the same function call.  (Like 'OK I was testing my query against one datatype, now I want two so let's make it a list... um `AttributeError order_by not found` what)

python/lsst/daf/butler/queries/_query.py
```diff
@@ -0,0 +1,652 @@
...
+        """
+        queries: dict[str, Query] = {}
+        if dataset_type is ...:
+            if collections is None:
+                collections = self._driver.get_default_collections()
+            else:
+                collections = tuple(ensure_iterable(collections))
+            for _, summary in self._driver.resolve_collection_path(collections):
+                for dataset_type_name in summary.dataset_types.names:
+                    queries[dataset_type_name] = self.join_dataset_search(dataset_type_name, collections)
```
dhirving, line 290: Would it be possible to just record the fact that we want all dataset types and send it to the server, instead of forcing the client to pre-resolve the collections here?

python/lsst/daf/butler/queries/_query.py
```diff
@@ -0,0 +1,652 @@
...
+from ._dimension_record_query_results import DimensionRecordQueryResults
+from .convert_args import convert_where_args
+from .driver import QueryDriver
+from .expression_factory import ExpressionFactory
+from .result_specs import DataCoordinateResultSpec, DatasetRefResultSpec, DimensionRecordResultSpec
+from .tree import DatasetSearch, InvalidQueryError, Predicate, QueryTree, make_identity_query_tree
+
+
+@final
+class Query(HomogeneousQueryBase):
```
dhirving, line 58: I wonder if instead of having Query inherit from HomogeneousQueryBase to provide `any()`, `explain_no_results()`, etc, if it could have a method `result_summary` or something that returns a QueryResult object with these methods.

Then the pattern is always consistently "build a query, call a method to choose a result type, call a method on the result object to get the actual data".  Would let you remove some (all?) of the intermediate base classes below QueryResultsBase.  Might also prevent some user confusion when they intend to call a method on the result object but accidentally call it on the query instead -- if the interfaces of Query and QueryResult are disjoint there is no possible confusion.  Let's you draw a clearer boundary in the documentation between the two concepts if you don't have to say "and also by the way Query is implicitly partially a QueryResult".

python/lsst/daf/butler/queries/_base.py
```diff
@@ -0,0 +1,287 @@
...
+        Notes
+        -----
+        If this method is called multiple times, the new sort terms replace
+        the old ones.
+        """
+        return self._copy(
+            self._tree, order_by=convert_order_by_args(self.dimensions, self._get_datasets(), *args)
+        )
+
+    def limit(self, limit: int | None = None, offset: int = 0) -> Self:
```
dhirving, line 236: Are we sure we want to support `offset` for limit?  It seems like it will encourage users to try to do things like manually asking for one page at a time in a loop for "efficiency" when that it is likely to actually be slower and increase database load by repeating queries.

Are there any important use cases that aren't covered by "incrementally read rows in order up to the specified limit, with our code implicitly retrieving partial result sets in a way that makes sense for the driver"?

python/lsst/daf/butler/queries/_query.py
```diff
@@ -0,0 +1,652 @@
...
+        """The names of all dataset types joined into the query.
+
+        The existence of datasets of these types constrains the data IDs of any
+        type of result.  Fields for these dataset types are also usable in
+        'where' expressions.
+
+        Note that this includes only dataset type names, not `DatasetType`
+        instances; the `DatasetQueryResults` adapter returned by the `datasets`
+        method does include `DatasetType` instances, since it is in a better
+        position to track and respect any storage class override specified.
```
dhirving: This paragraph is useful information for a maintainer, but I would encourage you to put this kind of thing in a comment instead of in the user-facing documentation.  This information speaks to "why was this code implemented the way it is" instead of "how can I use this to accomplish my task".

As someone who has recently read the Butler documentation with no prior knowledge of the system, I can tell you that this kind of verbose technical detail in the user-facing documentation contributes to a feeling of overwhelm that isn't what you want when you're trying to get someone to use your thing :)

doc/lsst.daf.butler/index.rst
```diff
@@ -101,48 +101,59 @@ Python API reference
 
 .. automodapi:: lsst.daf.butler
    :no-main-docstr:
+   :no-inherited-members:
```
dhirving, line 104: I'm not sure this is a good idea?  Given the deep inheritance hierarchies in this library, many of the important public-facing methods on classes are inherited from base classes.  As a user it's really useful to be able to see all the available methods for a class on one page, and most of our inheritance hierarchies aren't really meant to be a user-facing thing that we want people clicking through.

python/lsst/daf/butler/queries/_query.py
```diff
@@ -0,0 +1,652 @@
...
+            used.
+        dimensions : `DimensionGroup`, optional
+            The dimensions to assume for the dataset type if it is not
+            registered, or to check against if it is registered.  When the
+            dataset is not registered and this is not provided,
+            `MissingDatasetTypeError` is raised, since we cannot construct a
+            query without knowing the dataset's dimensions.  Providing this
+            argument causes the returned query to instead return no rows (as it
+            does when the dataset type is registered but no matching datasets
+            are found).
```
dhirving, line 429: What are the use cases for which you anticipate users providing the `dimensions` parameter?

Might be easier to just throw the MissingDatasetTypeError if the dataset isn't registered, instead of spending a long paragraph documenting the fact that if you provide a difficult-to-provide extra parameter you can get zero rows back instead.

python/lsst/daf/butler/queries/_base.py
```diff
@@ -0,0 +1,287 @@
...
+            results.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def where(
+        self,
+        *args: str | Predicate | DataId,
+        bind: Mapping[str, Any] | None = None,
+        **kwargs: Any,
```
dhirving, line 99: Aren't only `str` and `int` allowed as the values of `kwargs` here?  More strictly typing `kwargs` might help prevent some of the carnival of punctuation/spelling-related bugs invited by the combination of `*args` and `**kwargs`... it would at least flag `.where(biind={...})`.

python/lsst/daf/butler/queries/_query.py
```diff
@@ -0,0 +1,652 @@
...
+        *,
+        find_first: bool = True,
+    ) -> DatasetQueryResults: ...  # pragma: no cover
+
+    def datasets(
+        self,
+        dataset_type: str | DatasetType | Iterable[str | DatasetType] | EllipsisType,
+        collections: str | Iterable[str] | None = None,
+        *,
+        find_first: bool = True,
```
dhirving, line 231: Should find_first default to `False` so that you get all the results by default?

python/lsst/daf/butler/queries/driver.py
```diff
@@ -0,0 +1,458 @@
...
+    ) -> DimensionRecordResultPage: ...
+
+    @overload
+    def execute(self, result_spec: DatasetRefResultSpec, tree: QueryTree) -> DatasetRefResultPage: ...
+
+    @overload
+    def execute(self, result_spec: GeneralResultSpec, tree: QueryTree) -> GeneralResultPage: ...
+
+    @abstractmethod
+    def execute(self, result_spec: ResultSpec, tree: QueryTree) -> ResultPage:
```
dhirving, line 185: What's the advantage of jamming all this functionality into a single overloaded function name instead of having these be separate functions `execute_data_coordinate`, `execute_dimension_record` etc?

(Same for fetch_next_page below)

python/lsst/daf/butler/queries/driver.py
```diff
@@ -0,0 +1,458 @@
...
+    ) -> DimensionRecordResultPage: ...
+
+    @overload
+    def fetch_next_page(self, result_spec: DatasetRefResultSpec, key: PageKey) -> DatasetRefResultPage: ...
+
+    @overload
+    def fetch_next_page(self, result_spec: GeneralResultSpec, key: PageKey) -> GeneralResultPage: ...
+
+    @abstractmethod
+    def fetch_next_page(self, result_spec: ResultSpec, key: PageKey) -> ResultPage:
```
dhirving, line 227: It's hard to say without seeing an implementation here, but I wonder if it would be simpler for `execute` to return an `Iterator[DataCoordinateResultPage]` instead of having the driver need to keep track of the PageKeys that go with a ResultSpec for every query that has been run against it?

Or maybe execute should return a small class that is a context manager that does cleanup of the resources from the query, and is an `Iterable[Page]`.  Just to encapsulate the state of each individual execution separately instead of tracking it all in the single driver object.  In addition to maybe being simpler, the extra level of context manager would allow earlier cleanup of cursors or result parquet files if the user doesn't finish iteration.  I guess in order for that to work the calls to the QueryResults iterators would need to be wrapped in context managers too, which may be too much context manager in the user-facing interface.

One thing that is a little weird with the current setup is that the driver (and hence the cleanup) is bound to the query object.  Use cases where you construct a Query once and then execute multiple QueryResults objects against it will leak resources until you release the Query.

python/lsst/daf/butler/queries/driver.py
```diff
@@ -0,0 +1,458 @@
...
+)
+from .tree import DataCoordinateUploadKey, MaterializationKey, QueryTree
+
+PageKey: TypeAlias = uuid.UUID
+
+
+# The Page types below could become Pydantic models instead of dataclasses if
+# that makes them more directly usable by RemoteButler (at least once we have
+# Pydantic-friendly containers for all of them).  We may want to add a
+# discriminator annotation to the ResultPage union if we do that.
```
dhirving, line 75: I don't think these will want to be pydantic models.  Over the wire this data will probably be transmitted as like parquet rather than JSON, and the driver may want to return multiple pages for a given parquet file (since a 10MB parquet file is potentially hundreds of megabytes of in-memory data).

python/lsst/daf/butler/queries/visitors.py
```diff
@@ -0,0 +1,540 @@
...
+
+        Returns
+        -------
+        result : `object`
+            Implementation-defined.
+        """
+        raise NotImplementedError()
+
+
+class PredicateVisitor(Generic[_A, _O, _L]):
```
dhirving, line 188: What advantage do we get from having this extra generic/abstract base class instead of just having the more concrete one below?

python/lsst/daf/butler/queries/overlaps.py
```diff
@@ -0,0 +1,466 @@
...
+
+from .._topology import TopologicalFamily
+from ..dimensions import DimensionElement, DimensionGroup
+from . import tree
+from .visitors import PredicateVisitFlags, SimplePredicateVisitor
+
+_T = TypeVar("_T", bound=Hashable)
+
+
+class _NaiveDisjointSet(Generic[_T]):
```
dhirving, line 46: It might be a good idea to add a few tests targeting just this data structure by itself... even though it's pretty simple these kinds of things are prone to subtle bugs that can propagate to surprising bugs in consumers.

python/lsst/daf/butler/queries/tree/_predicate.py
```diff
@@ -0,0 +1,629 @@
...
+        ----------
+        value : `bool`
+            Value the predicate should evaluate to.
+
+        Returns
+        -------
+        predicate : `Predicate`
+            Predicate that evaluates to the given boolean value.
+        """
+        return cls.model_construct(operands=() if value else ((),))
```
dhirving, line 144: This doesn't really make sense to me.  If "empty operand" means `True`, then how does empty operand (`True`) anded with nothing evaluate to `False`?

python/lsst/daf/butler/queries/tree/_column_literal.py
```diff
@@ -0,0 +1,372 @@
...
+
+
+@final
+class IntColumnLiteral(ColumnLiteralBase):
+    """A literal `int` value in a column expression."""
+
+    expression_type: Literal["int"] = "int"
+
+    value: int
+    """The wrapped value after base64 encoding."""
```
dhirving: Several of the classes in this file have this incorrect comment about base64 encoding.

python/lsst/daf/butler/queries/tree/_column_literal.py
```diff
@@ -0,0 +1,372 @@
...
+        return repr(self.value)
+
+
+@final
+class FloatColumnLiteral(ColumnLiteralBase):
+    """A literal `float` value in a column expression."""
+
+    expression_type: Literal["float"] = "float"
+
+    value: float
```
dhirving, line 116: JSON encoders have a long history of doing whatever they feel like with floating point precision.  I'm not sure what Pydantic/Python's versions do... have you confirmed that they guarantee they will round-trip values with sufficient precision?

python/lsst/daf/butler/queries/tree/_base.py
```diff
@@ -0,0 +1,188 @@
...
+    @property
+    @abstractmethod
+    def precedence(self) -> int:
+        """Operator precedence for this operation.
+
+        Lower values bind more tightly, so parentheses are needed when printing
+        an expression where an operand has a higher value than the expression
+        itself.
+        """
+        raise NotImplementedError()
```
dhirving, line 103: I'm a little surprised to see precedence here... I would have expected that to be resolved at the higher level where we parse the input expressions, so that we just have a tree of expressions that could be evaluated directly here without worrying about ordering.

Is this only used for debug stringifications?  If it's me I would just always put the parentheses... less complexity, and it makes the tree structure explicit (which is often what you're trying to debug if you're looking at these?)

python/lsst/daf/butler/queries/tree/_column_set.py
```diff
@@ -0,0 +1,357 @@
...
+            )
+        match field:
+            case "dataset_id":
+                return column_spec.UUIDColumnSpec.model_construct(name=qualified_name, nullable=False)
+            case "ingest_date":
+                return column_spec.DateTimeColumnSpec.model_construct(name=qualified_name)
+            case "run":
+                # TODO: string length matches the one defined in the
+                # CollectionManager implementations; we need to find a way to
+                # avoid hard-coding the value in multiple places.
```
dhirving, line 341: Why not just declare a constant and reference it in both places?

python/lsst/daf/butler/queries/tree/_column_set.py
```diff
@@ -0,0 +1,357 @@
...
+            columns.
+
+        Returns
+        -------
+        name : `str`
+            Fully-qualified name.
+        """
+        return logical_table if field is None else f"{logical_table}:{field}"
+
+    def get_column_spec(self, logical_table: str, field: str | None) -> column_spec.ColumnSpec:
```
dhirving, line 305: What's the story on how you plan to use these ColumnSpec models?  Do they go over the wire?

dhirving (approved):
I think this should work pretty well.  It pushes a bit more complexity to the client than I would prefer -- mostly in `join_dataset_search` where it forces the client to do work deciding what dimensions will need to be involved in the query instead of just recording what dataset types the client asked for and sending it to the server.  However, since the dimensions are fairly high-level logical things inherent to the way Butler works I suspect this won't cause many backwards-compatibility issues.  The serialized data sent in QueryTree and the ResultSpec classes is quite straightforward and simple, so that's a win.

I don't see anything here that would be harmful to merge in its current form.  To the extent that you decide to make changes in response to these comments it can probably happen in a later (and hopefully much smaller!) PR.  (Also note there are 26 comments here -- sometimes GitHub likes to hide some when it goes over 10.)

doc/lsst.daf.butler/index.rst
TallJimbo: I can try to remove it and see if it's just warnings we get, but I know it's at least possible to get hard errors in the doc-build from the Sphinx bug this works around, and if that happens I'll have to put it back in at least a few places.  From the [slack conversation](https://lsstc.slack.com/archives/C2B6DQBAL/p1706715101824359) that led to this commit it sounds like if we can finally upgrade our doc tooling to a newer Sphinx it may go away.

python/lsst/daf/butler/queries/_base.py
TallJimbo: I'll give it a try.  I've been frustrated with MyPy not being as smart as it seems it ought to be about `**kwargs` in the past and have gotten in the habit of just always using `Any`.

python/lsst/daf/butler/queries/_base.py
TallJimbo: I don't know of any direct use cases, and I'd be happy to remove it.  @timj , @andy-slac , do you have a reason we need `offset` support, or was it added originally just because it felt reasonable when we added `limit`?

python/lsst/daf/butler/queries/_dataset_query_results.py
TallJimbo: I consider the single-type results to be the important thing and the chain to be secondary, because I expect most queries to be for a specific dataset type.  But I very much like the idea in one of your other comments of just dropping the support for querying more than one dataset type at a time in favor of the user writing their own loop over dataset types (let's discuss that further on that other thread).

python/lsst/daf/butler/queries/_dataset_query_results.py
TallJimbo: :+1: to dropping `ChainedDatasetQueryResults`.  That will simplify the mess of base classes a lot, and I do think the long term plan is to run those queries separately.

We would then need a separate method or methods for some important summary queries, like "which dataset types exist at all in these collections".  I had originally planned to add those in the future as convenience methods on `Query` that would delegate to `ChainedDatasetQueryResults.by_dataset_type()` and then call `any` on each (and hence not require any new `QueryDriver` methods).  But having a driver method dedicated to those summaries is not really a burden, and it could save us from having to bring the `CollectionSummary` objects down to the client at all (I'll have to work through that possibility to be sure) since that was largely about optimizing those "dataset types in collections" summary queries.  (The other reason to bring them down to the client is that caching them there may be easier than caching them on the server, but that's also very much a "maybe".)

I'll start with deleting `ChainedDatasetQueryResults` and whatever base classes that renders useless on this branch before I merge.  If that is goes well I'll stop there and put summarization methods on another ticket, but I think I might want to prototype those out before I dive back into implementing the driver to make sure I don't want to delete or change anything in the lower levels that I should have kept.

python/lsst/daf/butler/queries/_query.py
TallJimbo: Good idea.  I'll give it a try.

I think there are good reasons for both `Query` and `*QueryResults` to have at least `where` in common:
- on `Query` so you can add a WHERE clause before you materialize or branch that query into multiple kinds of results;
- on dataset query-result objects in particular so you can add a WHERE clause referencing a dataset field while letting `Query.datasets` join it in (i.e. instead of having to call `Query.join_dataset_search` , then `Query.where`, then `datasets`).

So I don't think we can get them fully disjoint, and that probably means one base class will remain.  But if we combine this with getting rid of `DatasetQueryResults` and `ChainedDatasetQueryResults` I think we can get it down to one base class and that would indeed be a lot better.

python/lsst/daf/butler/queries/_query.py
TallJimbo: Will do!

python/lsst/daf/butler/queries/_query.py
TallJimbo: (Getting rid of chained dataset query results takes care of this, too.)

python/lsst/daf/butler/queries/_query.py
TallJimbo: I don't think so.  This _is_ a break with the past, but it's an attempt to correct a mistake: the registry query interface does default `find_first=False`, but `find_first=True` is the behavior of `Butler.get`, `Butler.find_dataset`, and QuantumGraph generation, and it's what makes the results consistent with what we conceptually consider the members of a `CHAINED` collection (which does not include the "shadowed" datasets).

python/lsst/daf/butler/queries/_query.py
TallJimbo: This would be in a different interface per other threads, but I think the answer is yes - assuming we can manage cached collection summary lifetimes on the server as effectively as the client can via its context manager (though if the server can't trust collection summaries passed back from the client that's a moot point).

python/lsst/daf/butler/queries/_query.py
TallJimbo: It's pretty much entirely for `QuantumGraph` generation, and I that means we should drop it, though there's a sad history of us going back and forth between raising for unregistered datasets and returning empty results. and breaking downstream code every time we change it.

The registry query system does not raise and does not need or allow the caller to pass in the dimensions; instead it makes a special query that doesn't actually know its own dimensions but knows it cannot return results, which required special-casing it all over the implementation.  I really want to avoid that complexity this time, so this was my attempt to provide some of that past behavior (that history is also why `join_dataset_search` sucks down dataset type dimension information to the client).

But I think I agree we should just drop it, and just raise.  QuantumGraph generation can certainly work around that itself, and I think other callers can, too.

python/lsst/daf/butler/queries/driver.py
TallJimbo: Agreed.

As an aside, I think we'll might want to use JSON for the first page (which may be smaller than the rest), because Parquet is actually not efficient at all for very small amounts of data (most queries) and it's much more rigid than I'd hoped about normalizing data from multiple tables.  But that still doesn't mean these are the things we'd want to make into models.

python/lsst/daf/butler/queries/driver.py
TallJimbo: My guess was that the implementations would be more common than not.  I don't know if that'll be the case yet, since even on my other branch I've only done dimension record results.  I have no objection to splitting that up if it turns out not to be the case.

It's also very annoying that MyPy requires derived classes to spell out all the overloads in order to override the implementation.

python/lsst/daf/butler/queries/driver.py
TallJimbo: I don't see the driver as being bound to any particular `Query` object - I see it as being bound to one `with Butler._query() as query` context manager block (i.e. you could do `results = query.data_ids()` and then `del query` and its still the closing of the context manager that drops everything.

I do think a separate context manager for each result object is a bit much for the user interface, and I hadn't been worried about cleaning things up late because I was thinking that `with butler._query() ...` blocks would be quite short, and rarely involve more than one set of results (with the main exception being QG generation or anything else that makes a temp table).

I don't know how much to worry about users writing much longer-lived blocks for no good reason, which could indeed be a problem for us.  I think it helps a lot that in a Jupyter notebook you can't have a `with` block that spans more than one cell, so users will have to keep them short to see results interactively.  But of course that's not the only way butler queries will be used.

The driver being able to clean up unread pages as a context manager is why I went with this interface instead of an iterator one (clearly you could do that with an iterator interface, too; I figured I was saving each driver from having to write its own iterator that still ends up pointing back at internal driver state).  I'm inclined to leave this as it is for now, and see how it holds up when we implement real drivers.

python/lsst/daf/butler/queries/overlaps.py
TallJimbo: Will do.

python/lsst/daf/butler/queries/tree/_base.py
TallJimbo: It is just about stringification and it might now be overkill.  I added it back when `Predicate` was a flexible tree of binary AND and OR operators, and `A AND (B AND (C AND (D AND E)))` happened a lot and did get in the way of readability.  But now that `Predicate` always maintains conjunctive normal form (ANDs-of-ORs-[of NOT]) we don't need it to solve that particular problem.  Let me see how it goes if I rip it out.

python/lsst/daf/butler/queries/tree/_column_literal.py
TallJimbo: Will fix.

python/lsst/daf/butler/queries/tree/_column_literal.py
TallJimbo: I have not.  Thankfully `float` fields are pretty rare in our schema and I can't think of anything that would require high precision, but it would be nice to know if we're losing anything here.

python/lsst/daf/butler/queries/tree/_column_set.py
TallJimbo: They're used to make temporary SQL tables for `materialize` for direct butler on my in-progress branch, and I anticipate using them for some Arrow schemas, too.

I don't think they need to go over the wire.  They're models because they're used to validate the dimensions-configuration YAML file.

python/lsst/daf/butler/queries/tree/_column_set.py
TallJimbo: I think I just forgot to come back to this and do it later on the branch; will do. I think I was just hung up on where to put the constant.

python/lsst/daf/butler/queries/tree/_predicate.py
TallJimbo: Yeah, it could do with some commenting.  It follows from
`all([]) == True` and `any([]) == False` and the fact that the outer tuple is combined with `AND` while the inner tuples are combined with `OR`.   And I think that's what `all` and `any` do because that's what gets you the right behavior if you use it as the initial value when you fold with that operator (i.e. `(True AND A AND B) == (A AND B)`, `(False OR A OR B) == (A OR B)`).

python/lsst/daf/butler/queries/tree/_query_tree.py
TallJimbo: I see this as the user being forced to resolve collection wildcards in advance rather than it being something client methods do automatically, just as we're (now) thinking they should do with dataset types.  In addition to keeping our code simpler, I think it's a case of discouraging bad behavior ("find me all dataset of this type in any collection") being more important than making certain legitimate collection-wildcard queries more convenient.

Because queries only get defined in context-manager blocks, I don't think there should be any cases of `Query` or `QueryResult` objects living a long time, and in `DirectButler` that context can hold a transaction (maybe created on first-use rather than `__enter__`) and guarantee consistency.  We probably can't fully guarantee that in `RemoteButler`, but given that most collections relevant for `RemoteButler` are frozen I think it's less important there.

python/lsst/daf/butler/queries/tree/_query_tree.py
TallJimbo: I struggled with this a lot myself.  I added it because:
 - if `Query.join_dataset_search` can take a `DatasetType` and the user wants dataset results, it'll be confusing if they passed in a storage class override initially and then we forgot it just because we didn't have a place to put it;
 - I wanted `Query.join_dataset_search` to be able to take `DatasetType` as well as `str`, both so `Query.datasets` could delegate to it and so I didn't have to document why it was unusual in that respect.

I think the real problem is that `DatasetType` has a storage class, and that's a problem in lots of other places, but of course it's tremendously useful in probably even more other places and having a `DatasetTypeWithoutStorageClass` public class is its own problem.  And that's all thoroughly water under the bridge.

In any case, yes, I think the server can ignore this (or just blindly copy it into Parquet files, so the client doesn't have to edit them); the important thing is that the client doesn't forget it, and I think it's slightly better (even if that means sending it to the server) than out on its own in `Query` (and it would be `Query`, not `SingleTypeDatasetResults`, to serve its current purpose).

python/lsst/daf/butler/queries/tree/_query_tree.py
TallJimbo: I'm using `Mapping` just because there's no (built-in, Pydantic-friendly) `frozendict` and I want mypy to prevent in-place modifications.  I'm open to other ideas.

python/lsst/daf/butler/queries/tree/_query_tree.py
TallJimbo: I'll make it assert.  It used to raise an exception with a friendly error message then I dropped that to make it as simple as possible when trying to remove duplication between `QueryTree` and `Query`; I probably just went a bit too far in this case.

python/lsst/daf/butler/queries/visitors.py
TallJimbo: One of the more important visitors I anticipate is one whose return values are `sqlalchemy.ColumnElement[bool]`, and the more concrete one can't do that.  There are three generic arguments instead of one because I recalled needing that on a similar visitor class in the registry query system, but I don't recall why.  I'd like to leave it until we have a concrete driver or two.

TallJimbo (commented):
Here's a round of replies about changes I now plan to make but haven't made yet.

I'm not really accustomed to leaving even small review-response changes to another ticket, but this is such a behemoth PR that I can imagine that working better for the reviewer - so I'll leave that up to you. I'll start working on those on strictly separate commits, and I can put them on some other branch or push them here later.

I do think I want to make a number of those changes before getting back to implementing the driver, because I think they may actually make that easier, or at least differently hard.

dhirving:
> I'm not really accustomed to leaving even small review-response changes to another ticket, but this is such a behemoth PR that I can imagine that working better for the reviewer - so I'll leave that up to you.

:shrug: Probably makes sense to clean up some of the little stuff like the wrong comments before merging, but to me it makes sense to get this version merged down before removing the ChainedDatasetQueryResults or any changes like that.

For a new feature like this where the impact on users is basically nil until we make it public, there's basically no downside to merging early and merging often :) (Other than that our development workflow is kind of annoying with the need to generate a new JIRA ticket to make a PR.)

Benefits to you of having this code merged:
* You have a record of how the design has evolved as you tweak it, and you can more easily backtrack a change or resurrect a useful utility function because there is a linear history on a single branch to look through to find it.
* You don't need to rebase it anymore, and any new changes people make will take into account the code you've written here.
* You reduce the risk of subtle breakage from rebasing, because any rebasing will happen in smaller chunks.
* Having the rough framework merged makes it easier to delegate parts to other people

Benefits to the rest of the team of having this code merged:
* We can make changes to it (e.g. as I move the exception hierarchy around this week I can see which ones you've used here and make decisions accordingly.)
* It's easier to learn and think about it, because it's visible in the same branches we're working on.
* We don't have to wonder which of your side branches we should be looking at, because we can just see the progress on main.

python/lsst/daf/butler/queries/tree/_query_tree.py
dhirving: That's reasonable -- I forgot that `Mapping` and `MutableMapping` are two separate things and yeah there's not a great way to control the const-ness of a dict otherwise.

python/lsst/daf/butler/queries/tree/_query_tree.py
dhirving: > if `Query.join_dataset_search` can take a `DatasetType` and the user wants dataset results, it'll be confusing if they passed in a storage class override initially and then we forgot it just because we didn't have a place to put it

Don't they already have to pass the dataset type again when they call `datasets()` to declare that they want refs back for this dataset type?  To me it makes sense that whatever you pass into `datasets` is the authoritative thing, since that's where you're declaring the output format.

>  not `SingleTypeDatasetResults`, to serve its current purpose).

It's possible I'm misunderstanding what is going on with this StorageClass (or maybe all the effects that a StorageClass can have?)  To me it looks like you could just pass in the StorageClass when you create the SingleTypeDatasetQueryResults, and call `ref.overrideStorageClass` if needed just before yielding from `__iter__`.  If this doesn't work I'd like to know because I'm using a similar pattern a few places in RemoteButler.

python/lsst/daf/butler/queries/_dataset_query_results.py
dhirving: Cool, that seems like it will buy us a lot of simplicity.  Seems like it should be easy enough to add back a multi/batch version later if it turns out we want it.

python/lsst/daf/butler/queries/_query.py
dhirving: Yeah I buy the argument for having `where` on both, and one base class with one method is plenty straightforward.

python/lsst/daf/butler/queries/_base.py
dhirving: I think they put some work into it at some point... or at least they made it more complicated, there are a few different things you can do to type the keys as well as the values now etc.  I did check that a simple `**kwargs: str | int` typing worked as I expected before making the original comment.

python/lsst/daf/butler/queries/tree/_query_tree.py
TallJimbo: In almost all cases, that's right - it makes more sense to pass the storage class in later.  It's really quite an edge case I'm worried about here:
- user calls `Query.join_dataset_search` with a `DatasetType` instance that has a storage class that is not the same as the default one for that dataset type in the repository;
- user calls `Query.datasets` with a `str` dataset type name that has no storage class information.  Do we respect their previous storage class override?

Maybe it would be better to raise an exception or emit a warning if the `DatasetType` passed to `join_dataset_search` has a storage class that isn't the same as the repository one (since we need to look up the dataset type definition in the repository at that point anyway, to get the dimensions), saying that the storage class passed in there will be ignored.

TallJimbo:
Ok, I've fixed a few of the doc/comment issues and I'm punting the rest to [DM-43146](https://rubinobs.atlassian.net/browse/DM-43146).

[DM-43146]: https://rubinobs.atlassian.net/browse/DM-43146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

python/lsst/daf/butler/queries/_base.py
TallJimbo: Done on DM-43146.

python/lsst/daf/butler/queries/_query.py
TallJimbo: Done on DM-43146.

python/lsst/daf/butler/queries/tree/_base.py
TallJimbo: Done (on DM-43146): we now just add parentheses for any kind of nested expression (but do try to avoid them on literals, direct references to columns, and predicates).

python/lsst/daf/butler/queries/tree/_column_set.py
TallJimbo: Done on DM-43146, and a good thing, too - I had actually gotten the number wrong here before.

python/lsst/daf/butler/queries/tree/_query_tree.py
TallJimbo: Done on DM-43146.

python/lsst/daf/butler/queries/tree/_query_tree.py
TallJimbo: Done on DM-43146: `DatasetSearch` no longer has `storage_class_name`.  Instead `join_dataset_search` raises if you give it a storage class override.

python/lsst/daf/butler/queries/_dataset_query_results.py
TallJimbo: Done on DM-43146.

python/lsst/daf/butler/queries/_dataset_query_results.py
TallJimbo: Done on DM-43146.

python/lsst/daf/butler/queries/overlaps.py
TallJimbo: Done on DM-43146.

python/lsst/daf/butler/queries/_query.py
TallJimbo: Dropping `DatasetQueryResults` and `ChainedDatasetQueryResults` changes the picture here a bit, and I'm now inclined to punt this proposal to another later ticket:

- We're now down to one base class already.
- When we add a dataset-types-in-collections summary interface ([DM-43157](https://rubinobs.atlassian.net/browse/DM-43157)), we'll have something else we might want to call a "result summary"; it might even just be an expanded version of this proposal that knows how to group by dataset type and/or collection.
- I think we have a _decent_ story right now: you don't need to specify the result type to know whether there will be any results, so `Query` can tell you; for anything beyond that you have to use `QueryResult`.  It's not a great story, because you have to ponder a bit about duplication to understand why `Query` can't at least do `count`.  But I think it isn't terrible.

dhirving: That seems reasonable.

python/lsst/daf/butler/queries/_base.py
andy-slac: I think the intent was indeed to support paging. Efficiency of course depends on implementation. For complex queries it may be very inefficient, but for materialized tables it may work fine. OK for me to drop it if we don't use it.

PR #967, opened 2024-02-26
DM-42037: Make RemoteButler.find_dataset consistent with DirectButler

dhirving:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/remote_butler/_collection_args.py
+++ b/python/lsst/daf/butler/remote_butler/_collection_args.py
@@ -1,0 +1,57 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This software is dual licensed under the GNU General Public License and also
+# under a 3-clause BSD license. Recipients may choose which of these licenses
+# to use; please see the files gpl-3.0.txt and/or bsd_license.txt,
+# respectively.  If you choose the GPL option then the following text applies
+# (but note that there is still no warranty even if you opt for BSD instead):
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import annotations
+
+import re
+from typing import cast
+
+from ..registry import CollectionArgType
+from ..registry.wildcards import CollectionWildcard
+from .server_models import CollectionList
+
+
+def convert_collection_arg_to_glob_string_list(arg: CollectionArgType) -> CollectionList:
+    if arg is ...:
+        return CollectionList(["*"])
+    elif isinstance(arg, str):
+        return CollectionList([arg])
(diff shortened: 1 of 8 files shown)


python/lsst/daf/butler/remote_butler/server_models.py
```diff
@@ -31,31 +31,41 @@
...
     """Request model for find_dataset."""
 
     data_id: SerializedDataId
     collections: CollectionList
+    timespan: SerializedTimespan | None
     dimension_records: bool = False
     datastore_records: bool = False
 
 
+class FindDatasetResponseModel(pydantic.BaseModel):
```
dhirving, line 63: This change breaks backwards compatibility but we have breakage from other recent PRs that I was going to make new releases for already.

In theory I could have just changed the response type to be `SerializedDatasetRef | None`, but it's really useful to have a wrapper object for responses.  If you find out you need to add more data later, you can add it as an additional property without it being a backwards-incompatible change.

dhirving:
n.b. there are many exception-handling checks in the registry unit tests that look like they are working, but would incorrectly return a `ButlerServerErrror` in real usage.  Because of the way `httpx.TestClient` handles error propagation, all the test is showing is that we're getting the server to do something that triggers the expected errors server-side.

I have a theory on how to deal with this that we can discuss in the standup tomorrow.

python/lsst/daf/butler/remote_butler/_collection_args.py
```diff
@@ -0,0 +1,76 @@
...
+        return CollectionList([arg])
+    elif isinstance(arg, re.Pattern):
+        raise TypeError("RemoteButler does not support regular expressions as search patterns")
+    elif isinstance(arg, CollectionWildcard):
+        # In theory this could work, but CollectionWildcard eagerly converts
+        # any glob strings to regexes.  So it would need some rework to
+        # preserve the original string inputs.  CollectionWildcard has likely
+        # never been used in the wild by an end-user so this probably isn't
+        # worth figuring out.
+        raise TypeError("RemoteButler does not accept CollectionWildcard instances as search patterns")
```
timj, line 70: As far as I can tell, it's only used in ctrl_mpexec outside of daf_butler.

python/lsst/daf/butler/remote_butler/_registry.py
```diff
@@ -146,10 +154,30 @@ def findDataset(
...
-        # definition of the collections parameter is incompatible and timespans
-        # aren't supported yet.
-        raise NotImplementedError()
+        # Components are supported by Butler.find_dataset, but are required to
+        # raise an exception in registry.findDataset.  Not sure why, but
+        # someone went through the trouble of writing a fairly lengthy unit
+        # test (RegistryTests.testComponentLookups) only for the purpose of
+        # verifying this behavior with a note that "Components can no longer be
+        # found by registry", implying that it used to work but was
+        # intentionally removed for reasons.
```
timj, line 163: Yes. We recently purged registry interfaces so they no longer need to know about components. Butler is in a slightly different position and has a broader overview of components.

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -776,6 +765,37 @@ def _extract_dataset_type(datasetRefOrType: DatasetRef | DatasetType | str) -> D
...
+    explicit_storage_class: StorageClass | str | None,
+) -> DatasetRef:
+    """Return a DatasetRef with its storage class overridden to match the
+    StorageClass supplied by the user as input to one of the search functions.
+
+    Parameters
+    ----------
+    ref : `DatasetRef`
+        The ref to which we will apply the StorageClass override.
+    original_dataset_ref_or_type : `DatasetRef` | `DatasetType` | `str
```
timj, line 780:
```suggestion
    original_dataset_ref_or_type : `DatasetRef` | `DatasetType` | `str`
```

timj (approved):

timj:
PS Ticket name is wrong. Should be DM-43027. I think that means a new PR with the right branch name so that it links to the Jira pipelines change log properly.

dhirving:
Ticket number is wrong in branch name, re-opening this as a new PR: https://github.com/lsst/daf_butler/pull/968

PR #970, opened 2024-03-03
Tell how to avoid bind collisions.

ktlim:
## Checklist

- [ ] ran Jenkins


doc/lsst.daf.butler/queries.rst
```diff
@@ -168,6 +168,7 @@ filter query results based on that property of datasets.
 
 Registry methods accepting user expressions also accept a ``bind`` parameter, which is a mapping from identifier name to its corresponding value.
 Identifiers appearing in user expressions will be replaced with the corresponding value from this mapping.
+Choose bind identifiers that are distinct and not the same as a dimension or dimension attribute name.
```
timj, line 170: The second half of this comment is only true if the dimension attribute is an integer or string and it gets promoted to be a dimension at some point in the future... Floating point attributes can't become dimensions. I realize it's easy to tell people never to do it though.

PR #972, opened 2024-03-04
DM-43105: Add exposure.can_see_sky metadata field

timj:
Defaults to None when importing old YAML files. The importer can't really work out what the flag should be on import in that situation.

## Checklist

- [ ] ran Jenkins


timj:
This PR is using the wrong ticket number. Closing.

PR #973, opened 2024-03-06
DM-43121: Set up server to client error propagation

dhirving:
Set up a framework for transferring exceptions from the server to the client and raising them in RemoteButler.  Exceptions intended as user-facing error messages should now be raised as exceptions that subclass a new class `ButlerUserError`.  This allows us to distinguish between "intentional"/"expected" error messages which should be sent to the client and should not raise alarms, vs server-internal breakage that our team will have to investigate and fix.

Any `ButlerUserError` raised on the server will be caught and transferred to the client as JSON with a 422 status code, then raised as the original exception type on the client.

The top-level `_exceptions.py` will eventually become the single file where Butler-specific exceptions for the public API are defined.  Added a new exception type `DatasetNotFoundError` to replace some `LookupError`s raised when datasets were not found.  Hoisted `MissingDatasetTypeError` from `registry._exceptions` to the top-level `_exceptions`, and put its base classes in a separate file `_exceptions_legacy.py` to indicate they should not be used from new code.

Updated enough exceptions to pass the tests in `test_butler.py` using this framework. The registry tests in `test_remote_butler.py` will require more work in a later PR.

## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/_exceptions.py
```diff
@@ -61,3 +114,30 @@ def __init__(self, reasons: list[str]):
...
+    MissingDatasetTypeError,
+    UnknownButlerUserError,
+)
+_USER_ERROR_MAPPING = {e.error_type: e for e in _USER_ERROR_TYPES}
+assert len(_USER_ERROR_MAPPING) == len(
+    _USER_ERROR_TYPES
+), "Subclasses of ButlerUserError must have unique 'error_type' property"
+
+
+def create_butler_user_error(error_type: str, message: str) -> ButlerUserError:
```
timj, line 139: Can you please add a docstring for this since it is used elsewhere and would help people reading the code and getting previews from VSCode.

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -272,7 +264,10 @@ def get(
         **kwargs: Any,
     ) -> Any:
         # Docstring inherited.
-        model = self._get_file_info(datasetRefOrType, dataId, collections, kwargs)
+        try:
+            model = self._get_file_info(datasetRefOrType, dataId, collections, kwargs)
+        except DatasetNotFoundError as e:
+            raise FileNotFoundError(str(e)) from e
```
timj: Maybe we need to talk about this oddity for the future.

timj (approved):
Looks good.

python/lsst/daf/butler/remote_butler/_remote_butler.py
dhirving: This is a good call out... I'm pretty sure this is just incorrect even though it passes tests.

I thought this was needed because of the datastore vs registry thing.  Datastore pretty consistently throws FileNotFoundError when stuff can't be found, and we're mostly acting in place of datastore here.

But I took a closer look at DirectButler and it does actually throw DatasetNotFoundError (formerly LookupError) for most common cases of not being able to lookup a ref in get().  Apparently that just isn't covered in `test_butler.py`.  

The FileNotFoundErrors being expected in the tests are actually for a couple of weird edge cases:
* changing the UUID of a ref to something invalid (line 500 in test_butler.py)
* holding on to a ref after calling `pruneDatasets(purge=True)`, then trying to call get() on it (line 417 in test_butler.py)

Arguably in both these cases it would make just as much sense to throw the DatasetNotFoundError, but because these checks are inside Datastore they happen to throw FileNotFoundError because that's what Datastore does.

PR #974, opened 2024-03-06
DM-43187: Fix masked column reading into astropy tables.

erykoff:
## Checklist

- [x] ran Jenkins

Reading masked parquet columns into astropy Tables will now use appropriate
fill values.  In addition, floating point columns will be filled with NaN
instead of using a masked column.  This fixes discrepancies when accessing
masked columns with `.filled()` or not.

### DIFF
--- a/python/lsst/daf/butler/formatters/parquet.py
+++ b/python/lsst/daf/butler/formatters/parquet.py
@@ -287,15 +287,22 @@ def arrow_to_numpy_dict(arrow_table: pa.Table) -> dict[str, np.ndarray]:
             # For a masked column, we need to ask arrow to fill the null
             # values with an appropriately typed value before conversion.
             # Then we apply the mask to get a masked array of the correct type.
-
-            if t in (pa.string(), pa.binary()):
-                dummy = ""
-            else:
-                dummy = t.to_pandas_dtype()(0)
+            match t:
+                case t if t in (pa.float64(), pa.float32(), pa.float16()):
+                    null_value = np.nan
+                case t if t in (pa.int64(), pa.int32(), pa.int16(), pa.int8()):
+                    null_value = -1
+                case t if t in (pa.bool_(),):
+                    null_value = True
+                case t if t in (pa.string(), pa.binary()):
+                    null_value = ""
+                case _:
+                    null_value = 0
 
             col = np.ma.masked_array(
-                data=arrow_table[name].fill_null(dummy).to_numpy(),
+                data=arrow_table[name].fill_null(null_value).to_numpy(),
                 mask=arrow_table[name].is_null().to_numpy(),
+                fill_value=null_value,
             )
 
         if t in (pa.string(), pa.binary()):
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -287,16 +287,31 @@ def arrow_to_numpy_dict(arrow_table: pa.Table) -> dict[str, np.ndarray]:
...
+                    # the masked array.
+                    null_value = np.nan
+                    use_masked = False
+                case t if t in (pa.int64(), pa.int32(), pa.int16(), pa.int8()):
+                    null_value = -1
+                case t if t in (pa.bool_(),):
+                    null_value = True
+                case t if t in (pa.string(), pa.binary()):
+                    null_value = ""
+                case _:
```
taranu, line 304: Out of curiosity, which types are left over? Just hypothetical future additions?

tests/test_parquet.py
```diff
@@ -1019,7 +1033,22 @@ def _checkAstropyTableEquality(self, table1, table2, skip_units=False, has_bigen
...
+                    c1 = np.array(table1[name])
+                if isinstance(table2[name], atable.column.MaskedColumn):
+                    c2 = table2[name].filled()
+                    has_masked = True
+                else:
+                    c2 = np.array(table2[name])
+                np.testing.assert_array_equal(c1, c2)
+                # If we have a masked column then we test the underlying data.
+                if has_masked:
+                    np.testing.assert_array_equal(np.array(c1), np.array(c2))
```
taranu, line 1051: You may end up casting to array twice if only one of c1/c2 is masked, but I suppose that's harmless.

tests/test_parquet.py
```diff
@@ -1579,7 +1614,22 @@ def _checkAstropyTableEquality(self, table1, table2):
...
+                c1 = np.array(table1[name])
+            if isinstance(table2[name], atable.column.MaskedColumn):
+                c2 = table2[name].filled()
+                has_masked = True
+            else:
+                c2 = np.array(table2[name])
+            np.testing.assert_array_equal(c1, c2)
+            # If we have a masked column then we test the underlying data.
+            if has_masked:
+                np.testing.assert_array_equal(np.array(c1), np.array(c2))
```
taranu, line 1632: Same comment as above on casting twice.

taranu (commented):
Looks fine... if it works, it works.

python/lsst/daf/butler/formatters/parquet.py
erykoff: Unsigned ints in particular (and I added a test for those).  I could add a comment.

tests/test_parquet.py
erykoff: These are actually astropy columns, so there is a cast in both cases!  (But it's a fast no-copy cast which would happen anyway via the `__array__` interface).

taranu (approved):

PR #975, opened 2024-03-07
DM-43208: Expose Loader

arunkannawadi:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-43208.md
+++ b/doc/changes/DM-43208.md
@@ -1,0 +1,1 @@
+Exposure `Loader` class from `_config.py` so it can be imported from `lsst.daf.butler`.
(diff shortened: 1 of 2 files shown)


jchiang87 (approved):
Looks good!  Thanks so much!

timj:
`Loader` is an internal helper class. Why do we need it to be public? If people really need it, it needs to have a better name.

arunkannawadi:
If @jchiang87 needs it by the weekly that gets tagged tonight, I'll merge this and we can use a better name later. If not, I can wait on this.

timj:
Why though? Is there some discussion on Slack? Is he reading his own YAML files and wants the include support? Can he read them using the Config class directly and then converting Config to a dict? This really is an internal routine that was not designed to be public.

PR #977, opened 2024-03-15
DM-42637: Keep copies of old dimension universes

dhirving:
Store the dimensions.yaml configuration for universes 5 and 6.

This will be used by daf_butler_migrate to verify that the dimension universe stored in the DB is as expected before performing the migration.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/old_dimensions/README.md
+++ b/python/lsst/daf/butler/configs/old_dimensions/README.md
@@ -1,0 +1,6 @@
+These files are used by
+[daf_butler_migrate](https://github.com/lsst-dm/daf_butler_migrate) to verify
+and update the dimension universe stored in the database when upgrading the
+database schema.
+
+The naming convention is `<namespace>_universe<version>.yaml`.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,3 +2,4 @@
 
 - [ ] ran Jenkins
 - [ ] added a release note for user-visible changes to `doc/changes`
+- [ ] (if changing dimensions.yaml) make a copy of dimensions.yaml in `configs/old_dimensions`
--- a/python/lsst/daf/butler/configs/old_dimensions/daf_butler_universe5.yaml
+++ b/python/lsst/daf/butler/configs/old_dimensions/daf_butler_universe5.yaml
@@ -1,0 +1,452 @@
+version: 5
+namespace: daf_butler
+skypix:
+  # 'common' is the skypix system and level used to relate all other spatial
+  # dimensions.  Its value is a string formed by concatenating one of the
+  # other keys under the 'skypix' headerin (i.e. the name of a skypix system)
+  # with an integer level (with no zero-padding).
+  common: htm7
+  htm:
+    class: lsst.sphgeom.HtmPixelization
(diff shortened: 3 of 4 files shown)


python/lsst/daf/butler/dimension_universe_history.py
```diff
@@ -0,0 +1,51 @@
...
+    ----------
+    universe_version : `int`
+        Version number of the universe to be loaded.
+
+    Returns
+    -------
+    universe : `str`
+        Dimension universe configuration encoded as a JSON string.
+    """
+    # This is used by daf_butler_migrate during dimension universe migrations.
```
andy-slac: In general for `daf_butler_migrate` I try to avoid depending on `daf_butler` interfaces. `daf_butler_migrate` is very special in that it may need to work with different daf_butler versions, not just most recent (would be better to avoid this dependency completely, but some things are too expensive to reimplement).

timj:
As a general comment, I think I'd like the file name to include the namespace and the version number for completeness. I realize that daf_butler won't really have universes in it other than for daf_butler namespace but it would make it more explicit for people looking in from the outside if both were included.

python/lsst/daf/butler/dimension_universe_history.py
dhirving: So would you prefer that I put the historical dimension universe YAMLs in daf_butler_migrate instead then?  (I actually slightly prefer that but Tim wanted them in daf_butler.)

timj: My argument for putting them in daf_butler is that I have been asked by people in the past about old universes and it seems easier to point them at a file in daf_butler than try to explain to them what daf_butler_migrate is. If the consensus is that people want them in daf_butler_migrate I will go along with that (and we have to remember to add them there when we create new ones.

andy-slac: I think I like them better in daf_butler, but the code that reads them can go to daf_butler_migrate. I'm not sure that `load_historical_dimension_universe_json` is even generally useful in daf_butler.

dhirving: I thought it would be a good idea to have a stable function that daf_butler_migrate could call to insulate it from changes in daf_butler's package structure, and to make it obvious to readers in daf_butler that there is code elsewhere using these files, but I don't feel that strongly about it -- I'll put the code in daf_butler_migrate and the YAMLs in daf_butler.

python/lsst/daf/butler/configs/old_dimensions/README.md
```diff
@@ -0,0 +1,4 @@
+These files are used by
+[daf_butler_migrate](https://github.com/lsst-dm/daf_butler_migrate) to verify
+and update the dimension universe stored in the database when upgrading the
+database schema.
```
timj, line 4: Maybe add a note here that the naming convention is `<namespace>_universe<version>.yaml` ?

timj (approved):
This looks okay. I'll see about copying in earlier ones at some point.

timj:
Also consider adding a new PR check list item to remind people to copy universes in here whenever there is a new universe.

PR #979, opened 2024-03-19
DM-43101: Add exposure.can_see_sky metadata field

timj:
No attempt is made to generate a value when importing from an old universe. (replaces #972)

## Checklist

- [x] ran Jenkins
- [x] (if changing dimensions.yaml) make a copy of dimensions.yaml in `configs/old_dimensions`

### DIFF
--- a/python/lsst/daf/butler/json.py
+++ b/python/lsst/daf/butler/json.py
@@ -30,7 +30,9 @@ from __future__ import annotations
 __all__ = ("to_json_generic", "from_json_generic", "to_json_pydantic", "from_json_pydantic")
 
 import json
-from typing import TYPE_CHECKING, Any, Protocol, Type
+from typing import TYPE_CHECKING, Any, ClassVar, Protocol
+
+from pydantic import BaseModel
 
 if TYPE_CHECKING:
     from .dimensions import DimensionUniverse
@@ -42,7 +44,7 @@ class SupportsSimple(Protocol):
     serialization using "simple" methods names.
     """
 
-    _serializedType: Type
+    _serializedType: ClassVar[type[BaseModel]]
 
     def to_simple(self, minimal: bool) -> Any: ...
 
--- a/python/lsst/daf/butler/logging.py
+++ b/python/lsst/daf/butler/logging.py
@@ -25,6 +25,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+from __future__ import annotations
+
 __all__ = ("ButlerMDC", "ButlerLogRecords", "ButlerLogRecordHandler", "ButlerLogRecord", "JsonLogFormatter")
 
 import datetime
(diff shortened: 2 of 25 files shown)


python/lsst/daf/butler/timespan_database_representation.py
```diff
@@ -66,7 +66,7 @@ class _SpecialTimespanBound(enum.Enum):
     """
 
 
-TimespanBound = Union[astropy.time.Time, _SpecialTimespanBound, None]
+TimespanBound: TypeAlias = astropy.time.Time | _SpecialTimespanBound | None
```
timj: I'm not sure why this code is copied from `_timespan.py`.

python/lsst/daf/butler/logging.py
```diff
@@ -292,18 +292,18 @@ class ButlerLogRecords(_ButlerLogRecords):
...
+    def from_records(cls, records: Iterable[ButlerLogRecord]) -> Self:
         """Create collection from iterable.
 
         Parameters
         ----------
         records : iterable of `ButlerLogRecord`
             The records to seed this class with.
         """
-        return cls.model_construct(root=list(records))
+        return cast(Self, cls.model_construct(root=list(records)))
```
timj: `cls.model_construct()` returns `ButlerLogRecords` and mypy can't work out that's not `Self`.

python/lsst/daf/butler/timespan_database_representation.py
TallJimbo: It was just a mistake made when refactoring one module into two modules.  We should remove it and import instead it if it's actually used here.

python/lsst/daf/butler/json.py
```diff
@@ -42,7 +44,7 @@ class SupportsSimple(Protocol):
     serialization using "simple" methods names.
     """
 
-    _serializedType: Type
+    _serializedType: type[BaseModel]
```
TallJimbo: Should actually by `ClassVar[type[BaseModel]]`, I suspect.

Aside: this is correct for the code as it is now, but I think we really want to support things that are recognized by Pydantic other than `BaseModel` (like `typing.TypedDict`) as "simple" objects in the future.  I think that would involve making this into:

```
class SupportsSimple(Protocol, Generic[_T]):

    _adapter: ClassVar[pydantic.TypeAdapter[_T]]
```

TallJimbo (approved):
I'd prefer to revert the `Union -> |` commit and tell Ruff to either ignore that rule (ideal) or ignore at least the majority of those lines, as I think `Union` is better for anything multi-line (or potentially multi-line in the future).

I'm also not sold on the `Self` commit, as that is a subtle definition change rather than just notation, when inheritance is in play; I prefer not to use `Self` unless I want to say that derived classes must return the most-derived type.  Looks like we could be using `from __future__ import annotations` in `ButlerLogRecords` instead, though - probably dates from when pydantic couldn't handle that.

python/lsst/daf/butler/json.py
timj: Maybe, but `TypeAdapter` doesn't support `model_dump_json` which is what the code relies on (it's called `dump_json` with `TypeAdapter`).

I'll do your `ClassVar` fix.

PR #982, opened 2024-03-25
DM-43499: Add timespan to Butler.get

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/remote_butler/server_models.py
+++ b/python/lsst/daf/butler/remote_butler/server_models.py
@@ -73,6 +73,7 @@ class GetFileByDataIdRequestModel(pydantic.BaseModel):
     dataset_type_name: DatasetTypeName
     data_id: SerializedDataId
     collections: CollectionList
+    timespan: SerializedTimespan | None = None
 
 
 class GetFileResponseModel(pydantic.BaseModel):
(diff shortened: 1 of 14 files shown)


python/lsst/daf/butler/remote_butler/server_models.py
```diff
@@ -73,6 +73,7 @@ class GetFileByDataIdRequestModel(pydantic.BaseModel):
     dataset_type_name: DatasetTypeName
     data_id: SerializedDataId
     collections: CollectionList
+    timespan: SerializedTimespan | None = None
```
timj, line 76: @dhirving what's the policy regarding changes to this model? Should I default to None (so that serialization could drop it if it has not been set)?

dhirving: Normally yes, you should always provide a default when adding a new property to an existing model.  That will let the server remain backwards-compatible with older clients that don't send the new field.

However, at this exact moment I've been breaking backwards compatibility repeatedly for the past few weeks.  Russ has been advised that we need to do a synchronized release of `butler-server`, `datalinker` and `vo-cutouts` at the next release of one of the three.

So do whatever you want :)  What you have there already seems fine to me.

TallJimbo (approved):

dhirving:
You didn't add the new parameter to the public `Butler` interface... mypy didn't complain because of the presence of kwargs.

timj:
I put a fix on #979.

PR #983, opened 2024-03-27
DM-43526: Implement registry collection methods for RemoteButler

dhirving:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/remote_butler/_registry.py
+++ b/python/lsst/daf/butler/remote_butler/_registry.py
@@ -51,6 +51,7 @@ from ..registry import (
     CollectionArgType,
     CollectionSummary,
     CollectionType,
+    CollectionTypeError,
     DatasetTypeError,
     Registry,
     RegistryDefaults,
@@ -106,7 +107,7 @@ class RemoteButlerRegistry(Registry):
         raise NotImplementedError()
 
     def getCollectionType(self, name: str) -> CollectionType:
-        raise NotImplementedError()
+        return self._butler._get_collection_info(name).type
 
     def registerRun(self, name: str, doc: str | None = None) -> bool:
         raise NotImplementedError()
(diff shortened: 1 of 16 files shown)


python/lsst/daf/butler/remote_butler/_registry.py
```diff
@@ -115,22 +119,28 @@ def removeCollection(self, name: str) -> None:
...
+            raise CollectionTypeError(f"Collection '{parent}' has type {info.type.name}, not CHAINED.")
+        return info.children
 
     def setCollectionChain(self, parent: str, children: Any, *, flatten: bool = False) -> None:
         raise NotImplementedError()
 
     def getCollectionParentChains(self, collection: str) -> set[str]:
-        raise NotImplementedError()
+        info = self._butler._get_collection_info(collection, include_parents=True)
+        assert info.parents is not None
```
timj: `assert` does trigger a philosophical debate here as to what happens if it ever fails. The error message that you get is never helpful. Consider `RuntimeError` with an error message instead? (or at least add a message to the assert).

python/lsst/daf/butler/remote_butler/_registry.py
```diff
@@ -271,7 +281,15 @@ def queryCollections(
         flattenChains: bool = False,
         includeChains: bool | None = None,
     ) -> Sequence[str]:
-        raise NotImplementedError()
+        if includeChains is None:
```
timj, line 284: @TallJimbo what's the plan for the `datasetType` parameter? It's not in this implementation and is ignored in sql registry, but should the client code forward it to the server on the assumption that at some point the direct butler might use it? Or are we okay acting like it doesn't exist (the documentation doesn't seem to say that it's ignored and sql_registry mentions DM-24939 which is Done but hasn't fixed the problem).

timj (approved):

python/lsst/daf/butler/remote_butler/_registry.py
dhirving: If it gets implemented in DirectButler someday it can get added here at the same time.  It sounds like we're leaning towards re-designing this interface entirely, so we can drop it from the new interface or make it actually work there.

I don't like having unused variables/features hanging around because their existence implies to future readers that they are used for something, which can end up wasting a lot of time.

TallJimbo: I think we drop it here, and RFC dropping it from DirectButler (and the CLI) when we RFC dropping regex support there.

PR #984, opened 2024-03-28
DM-43574: Run test_simpleButler against RemoteButler

dhirving:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_simpleButler.py
+++ b/tests/test_simpleButler.py
@@ -54,45 +54,30 @@ from lsst.daf.butler import (
 from lsst.daf.butler.datastore.file_templates import FileTemplate
 from lsst.daf.butler.registry import RegistryConfig, RegistryDefaults, _RegistryFactory
 from lsst.daf.butler.tests import DatastoreMock
-from lsst.daf.butler.tests.utils import makeTestTempDir, removeTestTempDir
+from lsst.daf.butler.tests.utils import TestCaseMixin, makeTestTempDir, removeTestTempDir
+
+try:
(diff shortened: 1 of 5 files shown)


tests/test_simpleButler.py
```diff
@@ -767,5 +757,68 @@ def test_skypix_templates(self):
         self.assertEqual(path, "test/warp/HSCA02713600_12345_12345_12345_12345")
 
 
+class DirectSimpleButlerTestCase(SimpleButlerTests, unittest.TestCase):
+    """Run tests against DirectButler implementation."""
+
+    datasetsManager = (
```
timj, line 763: I think this is the only one we use now and is the default so in some future clean up we can drop this manager override completely.

tests/test_simpleButler.py
```diff
@@ -767,5 +757,68 @@ def test_skypix_templates(self):
...
+@unittest.skipIf(create_test_server is None, "Server dependencies not installed.")
+class RemoteSimpleButlerTestCase(SimpleButlerTests, unittest.TestCase):
+    """Run tests against Butler client/server."""
+
+    supportsCollectionRegex = False
+
+    def makeButler(self, writeable: bool = False) -> Butler:
+        server_instance = self.enterContext(create_test_server(TESTDIR))
+        butler = server_instance.hybrid_butler
+        DatastoreMock.apply(butler)
```
timj, line 800: Maybe add a comment here that the mock can't be used directly because remote butler isn't using a real datastore interface?

timj (approved):
Looks great. Very useful change.

PR #989, opened 2024-04-03
DM-43671: Implement atomic collection prepend

dhirving:
Added an atomic chained collection prepend method to Butler.  It works by taking a row lock on the parent collection's row in the collections tables, which acts as a mutex for the modification of the collection_chain table.  The existing setCollectionChain method now also uses this lock, so that it can interact safely with the new operation.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/collections/_base.py
+++ b/python/lsst/daf/butler/registry/collections/_base.py
@@ -32,9 +32,8 @@ __all__ = ()
 
 import itertools
 from abc import abstractmethod
-from collections import namedtuple
 from collections.abc import Iterable, Iterator, Set
-from typing import TYPE_CHECKING, Any, TypeVar, cast
+from typing import TYPE_CHECKING, Any, Generic, NamedTuple, TypeVar, cast
 
 import sqlalchemy
 
@@ -77,7 +76,13 @@ def _makeCollectionForeignKey(
     return ddl.ForeignKeySpec("collection", source=(sourceColumnName,), target=(collectionIdName,), **kwargs)
 
 
-CollectionTablesTuple = namedtuple("CollectionTablesTuple", ["collection", "run", "collection_chain"])
+_T = TypeVar("_T")
+
+
+class CollectionTablesTuple(NamedTuple, Generic[_T]):
+    collection: _T
+    run: _T
+    collection_chain: _T
 
 
 def makeRunTableSpec(
@@ -188,7 +193,7 @@ class DefaultCollectionManager(CollectionManager[K]):
     def __init__(
         self,
         db: Database,
-        tables: CollectionTablesTuple,
+        tables: CollectionTablesTuple[sqlalchemy.Table],
         collectionIdName: str,
         *,
         caching_context: CachingContext,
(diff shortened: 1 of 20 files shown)


dhirving:
One issue this doesn't yet address is duplicates in the collection chain.  Not sure what would make the most sense:
1. Remove the existing collections before re-inserting, so they effectively move up to their new position at the front of the chain.
2. Throw if there are duplicates.  (I don't think there is a unique constraint enforcing this but there could be easily.)
3. Just allow them?  

I'd say probably #1 unless you have other opinions?  It's idempotent and makes sense in the context of find-first.

andy-slac:
> One issue this doesn't yet address is duplicates in the collection chain.

I think that common use case in AP is when multiple clients do prepend concurrently. So 1 is probably the only useful option.

python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -426,22 +426,29 @@ def update_chain(
...
         self._addCachedRecord(record)
         return record
+
+    def _insert_collection_chain_rows(
+        self,
+        parent_key: K,
+        starting_position: int,
+        child_keys: Iterable[K],
+    ) -> None:
+        position = itertools.count(starting_position)
```
TallJimbo: Using `enumerate` inside the list comprehension plus an offset seems a little less error-prone.  I don't know why that wasn't the case before.

python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -452,3 +453,94 @@ def _insert_collection_chain_rows(
             for child in child_keys
         ]
         self._db.insert(self._tables.collection_chain, *rows)
+
+    def prepend_collection_chain(
+        self, parent_collection_name: str, child_collection_names: list[str]
+    ) -> None:
```
TallJimbo, line 485: I think we want to insert some kind of cache-invalidation or cache check on `self._caching_context` here.  I think
```
if self._caching_context.is_enabled:
    raise <something>
```
would be fine, as I'm not sure we actually need to support modifying a collection while a caching context is active.  Same for `update_chain`.

TallJimbo (approved):
Looks good!  Two minor comments.

python/lsst/daf/butler/registry/collections/_base.py
dhirving: It does handle updating the cache at the end already.  But I like forbidding it entirely, that will let me remove a decent amount of code.

dhirving:
I'm updating this to handle duplicates according to strategy 1 and will make the other changes you suggested.

python/lsst/daf/butler/registry/collections/_base.py
dhirving: It turns out that `setCollectionChain` is called inside a caching context from `Butler.import_`.  Whether that caching context is useful or not I can't really say.

So I'll do this in prepend_collection_chain but leave update_chain alone for now.

dhirving: I do think that making caching context a purely read-only thing would let us remove a decent amount of cache-maintenance code throughout, so might be worth exploring in a later ticket.

PR #990, opened 2024-04-04
DM-43315: Implement additional collection chain operations

dhirving:
Added additional collection chain methods to the top-level `Butler` interface: `extend_collection_chain`, `remove_from_collection_chain`, and `redefine_collection_chain`.  These methods are all "atomic" functions that can safely be used concurrently from multiple processes.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1772,6 +1772,41 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
         raise NotImplementedError()
 
     @abstractmethod
+    def extend_collection_chain(
+        self, parent_collection_name: str, child_collection_names: str | Iterable[str]
+    ) -> None:
+        """Add children to the end of a CHAINED collection.
+
+        If any of the children already existed in the chain, they will be moved
+        to the new position at the end of the chain.
+
+        Parameters
+        ----------
+        parent_collection_name : `str`
+            The name of a CHAINED collection to which we will add new children.
+        child_collection_names : `Iterable` [ `str ` ] | `str`
+            A child collection name or list of child collection names to be
+            added to the parent.
+
+        Raises
+        ------
+        MissingCollectionError
+            If any of the specified collections do not exist.
+        CollectionTypeError
+            If the parent collection is not a CHAINED collection.
+        CollectionCycleError
+            If this operation would create a collection cycle.
+
+        Notes
+        -----
+        If this function is called within a call to ``Butler.transaction``, it
+        will hold a lock that prevents other processes from modifying the
+        parent collection until the end of the transaction.  Keep these
+        transactions short.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
     def remove_from_collection_chain(
         self, parent_collection_name: str, child_collection_names: str | Iterable[str]
     ) -> None:
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1770,3 +1770,34 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
         transactions short.
         """
         raise NotImplementedError()
+
+    @abstractmethod
+    def remove_from_collection_chain(
+        self, parent_collection_name: str, child_collection_names: str | Iterable[str]
+    ) -> None:
+        """Remove children from a CHAINED collection.
+
+        Parameters
+        ----------
+        parent_collection_name : `str`
+            The name of a CHAINED collection from which we will remove
+            children.
+        child_collection_names : `Iterable` [ `str ` ] | `str`
+            A child collection name or list of child collection names to be
+            removed from the parent.
+
+        Raises
+        ------
+        MissingCollectionError
+            If any of the specified collections do not exist.
+        CollectionTypeError
+            If the parent collection is not a CHAINED collection.
+
+        Notes
+        -----
+        If this function is called within a call to ``Butler.transaction``, it
+        will hold a lock that prevents other processes from modifying the
+        parent collection until the end of the transaction.  Keep these
+        transactions short.
+        """
+        raise NotImplementedError()
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1737,6 +1737,39 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
         raise NotImplementedError()
 
     @abstractmethod
+    def redefine_collection_chain(
+        self, parent_collection_name: str, child_collection_names: str | Iterable[str]
+    ) -> None:
+        """Replace the contents of a CHAINED collection with new children.
+
+        Parameters
+        ----------
+        parent_collection_name : `str`
+            The name of a CHAINED collection to which we will assign new
+            children.
+        child_collection_names : `Iterable` [ `str ` ] | `str`
+            A child collection name or list of child collection names to be
+            added to the parent.
+
+        Raises
+        ------
+        MissingCollectionError
+            If any of the specified collections do not exist.
+        CollectionTypeError
+            If the parent collection is not a CHAINED collection.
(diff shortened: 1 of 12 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -1770,3 +1803,69 @@ def prepend_collection_chain(
...
         """
         raise NotImplementedError()
+
+    @abstractmethod
+    def extend_collection_chain(
+        self, parent_collection_name: str, child_collection_names: str | Iterable[str]
+    ) -> None:
+        """Add children to the end of a CHAINED collection.
+
+        If any of the children already existed in the chain, they will be moved
```
dhirving, line 1813: Moving existing children is less obviously useful/correct than it is for `prepend`.  Not sure what would be better though, and this is at least consistent.

python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -469,6 +529,21 @@ def _insert_collection_chain_rows(
...
+        # space to become fragmented.
+        #
+        # If this ever actually happens, we should consider doing a migration
+        # to increase the position column to a 32-bit number.
+        # To fix it in the short term, you can re-write the collection chain to
+        # defragment it by doing something like:
+        # registry.setCollectionChain(
+        #     parent,
+        #     registry.getCollectionChain(parent)
+        # )
```
dhirving, line 558: Not sure whether me not doing anything to address this is "efficient" or merely "lazy".  The most realistic scenario where this could happen is someone trying to use a collection chain as a queue... after around 32,000 prepends this will happen if you only remove from the end of the chain.

I should probably just write a migration script to make that column 32-bit.

python/lsst/daf/butler/_butler.py
```diff
@@ -1736,6 +1736,39 @@ def _clone(
...
+        self, parent_collection_name: str, child_collection_names: str | Iterable[str]
+    ) -> None:
+        """Replace the contents of a CHAINED collection with new children.
+
+        Parameters
+        ----------
+        parent_collection_name : `str`
+            The name of a CHAINED collection to which we will assign new
+            children.
+        child_collection_names : `Iterable` [ `str ` ] | `str`
```
```suggestion
        child_collection_names : `~collections.abc.Iterable` [ `str ` ] | `str`
```
otherwise Sphinx doesn't recognize it, unfortunately.

python/lsst/daf/butler/_butler.py
```diff
@@ -1770,3 +1803,69 @@ def prepend_collection_chain(
         transactions short.
         """
         raise NotImplementedError()
+
+    @abstractmethod
+    def extend_collection_chain(
```
TallJimbo, line 1808: I'm not sure which I prefer, but I'll throw it out there as food for thought: what about merging the `prepend_`, `extend_`, `redefine_`, and `remove_from_` methods into a single `modify_` method with a  `mode: Literal["prepend", "extend", "remove", "set"]` argument?  That's closer to the CLI and it slightly reduces the number of `Butler` methods.

python/lsst/daf/butler/_butler.py
```diff
@@ -1770,3 +1803,69 @@ def prepend_collection_chain(
...
+        """Add children to the end of a CHAINED collection.
+
+        If any of the children already existed in the chain, they will be moved
+        to the new position at the end of the chain.
+
+        Parameters
+        ----------
+        parent_collection_name : `str`
+            The name of a CHAINED collection to which we will add new children.
+        child_collection_names : `Iterable` [ `str ` ] | `str`
```
```suggestion
        child_collection_names : `~collections.abc.Iterable` [ `str ` ] | `str`
```

python/lsst/daf/butler/_butler.py
```diff
@@ -1770,3 +1803,69 @@ def prepend_collection_chain(
...
+        self, parent_collection_name: str, child_collection_names: str | Iterable[str]
+    ) -> None:
+        """Remove children from a CHAINED collection.
+
+        Parameters
+        ----------
+        parent_collection_name : `str`
+            The name of a CHAINED collection from which we will remove
+            children.
+        child_collection_names : `Iterable` [ `str ` ] | `str`
```
```suggestion
        child_collection_names : `~collections.abc.Iterable` [ `str ` ] | `str`
```

python/lsst/daf/butler/registry/collections/_base.py
TallJimbo: Collection chains are very frequently LIFO stacks, but I can't think of any context in which one would be used as a FIFO queue.  So I think that migration script can be very low priority.

python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -408,31 +409,90 @@ def _fetch_by_key(self, collection_ids: Iterable[K] | None) -> list[CollectionRe
...
+            # Avoid having cache-maintenance code around that is unlikely to
+            # ever be used.
+            raise RuntimeError("Chained collection modification not permitted with active caching context.")
+
+        if not skip_cycle_check:
+            self._sanity_check_collection_cycles(parent_collection_name, child_collection_names)
+
+        child_records = self.resolve_wildcard(
+            CollectionWildcard.from_names(child_collection_names), flatten_chains=False
+        )
```
TallJimbo, line 498: It seems a little wasteful to query for child rows both here and in the sanity check, though I don't know if that's likely to matter.

But more importantly, it seems like we need to move these queries after the row-level locks.  As is, couldn't another process modify the chain in between the queries for its contents and the point at which we acquire the lock?  Or is it that the atomic operations are precisely those that don't use the results of those queries?  If that's the case, I think some comments would help.

TallJimbo (approved):

python/lsst/daf/butler/_butler.py
dhirving: I also had that thought... but I'm concerned we're going to eventually add options that only apply to some subset of the modes.  Then we'll end up in the situation we're at with a lot of the other Butler methods where it takes multiple pages of documentation/checks to explain which sets of parameters are legal together.

One other idea I had was maybe naming them `collection_chain_*` instead of `*_collection_chain`, so they group together a little better in the docs.  But the rest of the butler methods are verb-first.

python/lsst/daf/butler/registry/collections/_base.py
dhirving: > It seems a little wasteful to query for child rows both here and in the sanity check, though I don't know if that's likely to matter.

It's definitely wasteful but I think not to an extent that is worth worrying about... in any case it's no slower than the previous version because all of this was re-used from the existing implementation.

The sanity check one is restricted to only `type==CHAINED` and is recursive so it's sort of a different query.  I think the way these two queries are done will change up some more when we add the recursive query thing to resolve_wildcard... all we need from the second query is the IDs corresponding to the names and it does more work than necessary right now.

> As is, couldn't another process modify the chain in between the queries for its contents and the point at which we acquire the lock? 

These aren't queries for the chain contents -- they're queries about the user-provided list of new child collections, and do not involve the parent collection.

The lock is only on the parent collection.  So moving these inside the lock wouldn't buy us anything and would make the critical section longer.  

For the child name -> ID query, the only concurrent change that affects anything is deleting one of the children (or delete and replace with a new collection of the same name).  In those cases the foreign key constraint will prevent us from inserting a child that existed at the time we looked up its ID but no longer exists.

The sanity check thing is hairier, and as I mentioned in the comment does not guarantee that cycles cannot exist.  To guarantee no cycles we'd have to recursively lock all child chained collections as well.  The locking order is  tricky to get right and locking multiple collections makes the availability issues with this approach worse.  It's isomorphic to the locking for the denormalized, flattened chain table we discussed the other week.  (That table would actually be a good way to enforce this... we could have a check constraint that `parent != child`.) But in any case just locking the parent doesn't help any -- this query wants to know if the parent is a recursive child of any of the children.  Locking the parent only prevents its children from changing and we don't care about the parent's children.

> I think some comments would help.

Yeah I'll add a bit more info here.

PR #992, opened 2024-04-08
DM-43746: Clean up RemoteButler

dhirving:
RemoteButler is getting too large, cut down clutter in the main file by factoring out a `RemoteButlerHttpConnection` class and moving some standalone functions to their own file.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/remote_butler/_http_connection.py
+++ b/python/lsst/daf/butler/remote_butler/_http_connection.py
@@ -46,6 +46,19 @@ _AnyPydanticModel = TypeVar("_AnyPydanticModel", bound=BaseModel)
 
 
 class RemoteButlerHttpConnection:
+    """HTTP connection to a Butler server.
+
+    Parameters
+    ----------
+    http_client : `httpx.Client`
+        HTTP connection pool we will use to connect to the server.
+    server_url : `str`
+        URL of the Butler server we will connect to.
+    access_token : `str`
+        Rubin Science Platform Gafaelfawr access token that will be used to
+        authenticate with the server.
+    """
+
     def __init__(self, http_client: httpx.Client, server_url: str, access_token: str) -> None:
         self._client = http_client
         self.server_url = server_url
@@ -95,8 +108,7 @@ class RemoteButlerHttpConnection:
         path : `str`
             A relative path to an endpoint.
         params : `Mapping` [ `str` , `str` | `bool` ]
-            Pydantic model containing the request body to be sent to the
-            server.
+            Query parameters included in the request URL.
 
         Returns
         -------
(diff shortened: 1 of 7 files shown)


python/lsst/daf/butler/remote_butler/_http_connection.py
```diff
@@ -0,0 +1,226 @@
...
+        self._headers = auth_headers | headers
+
+    def post(self, path: str, model: BaseModel) -> httpx.Response:
+        """Send a POST request to the Butler server.
+
+        Parameters
+        ----------
+        path : `str`
+            A relative path to an endpoint.
+        model : ``BaseModel``
```
timj:
```suggestion
        model : `pydantic.BaseModel`
```
?

python/lsst/daf/butler/remote_butler/_http_connection.py
```diff
@@ -0,0 +1,226 @@
...
+        )
+
+    def get(self, path: str, params: Mapping[str, str | bool] | None = None) -> httpx.Response:
+        """Send a GET request to the Butler server.
+
+        Parameters
+        ----------
+        path : `str`
+            A relative path to an endpoint.
+        params : `Mapping` [ `str` , `str` | `bool` ]
```
timj:
```suggestion
        params : `~collections.abc.Mapping` [ `str` , `str` | `bool` ]
```
Sphinx requires this to be fully specified.

python/lsst/daf/butler/remote_butler/_http_connection.py
dhirving: I thought there was some problem with the Sphinx and the pydantic documentation

python/lsst/daf/butler/remote_butler/_http_connection.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def _send_request(
+        self,
+        method: str,
+        path: str,
+        *,
+        content: bytes | None = None,
+        params: Mapping[str, str | bool] | None = None,
+        headers: Mapping[str, str] | None = None,
+    ) -> httpx.Response:
```
timj, line 153: In some sense, this is the most complicated method in the file and the one that might want a couple of sentences to explain it.

python/lsst/daf/butler/remote_butler/_http_connection.py
timj: No, the problem is when you have a model using numpydoc that is inheriting from BaseModel. Sphinx can't render the methods inherited from the base class not using numpydoc. There is no problem with linking in a docstring to a pydantic base model.

timj (approved):
Nice cleanup. Thanks.

PR #993, opened 2024-04-08
DM-43750: Fix crash when using NameKeyCollectionManager

dhirving:
Fix an issue where NameKeyCollectionManager would throw an exception when assigning a collection chain for a collection whose record was already in the collection cache.

This was happening because of an optimization that caused the "by key" cache to not be populated when using NameKeyCollectionManager, but some code paths assumed that both the "by name" and "by key" caches are always available.  Fixed the bug by removing the optimization, ensuring the behavior is the same for all implementations.

## Checklist

- [x] ran Jenkins


TallJimbo (approved):
Looks great!

PR #994, opened 2024-04-09
DM-43759: Move collection chains interface to sub-object

dhirving:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_butler/_direct_butler_collections.py
+++ b/python/lsst/daf/butler/direct_butler/_direct_butler_collections.py
@@ -42,7 +42,7 @@ class DirectButlerCollections(ButlerCollections):
 
     Parameters
     ----------
-    registry : `SqlRegistry`
+    registry : `~lsst.daf.butler.registry.sql_registry.SqlRegistry`
         Registry object used to work with the collections database.
     """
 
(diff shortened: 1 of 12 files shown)


python/lsst/daf/butler/direct_butler/_direct_butler_collections.py
```diff
@@ -0,0 +1,74 @@
...
+from .._butler_collections import ButlerCollections
+from ..registry.sql_registry import SqlRegistry
+
+
+class DirectButlerCollections(ButlerCollections):
+    """Implementation of ButlerCollections for DirectButler.
+
+    Parameters
+    ----------
+    registry : `SqlRegistry`
```
timj: Sphinx won't pick this up because `lsst.daf.butler.SqlRegistry` does not work.

```suggestion
    registry : `~lsst.daf.butler.registry.sql_registry.SqlRegistry`
```

timj (approved):

python/lsst/daf/butler/direct_butler/_direct_butler_collections.py
dhirving: I'm not importing it from lsst.daf.butler.SqlRegistry, I'm importing it from the relative path.  Shouldn't it just be able to find it?  What are the rules for when you need this weird `~` package syntax and when it just works normally?

PR #996, opened 2024-04-10
DM-43741: Add minimal RemoteButler QueryDriver

dhirving:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/dimensions/_group.py
+++ b/python/lsst/daf/butler/dimensions/_group.py
@@ -27,12 +27,12 @@
 
 from __future__ import annotations
 
-__all__ = ["DimensionGroup"]
+__all__ = ("DimensionGroup", "SerializedDimensionGroup")
 
 import itertools
 from collections.abc import Iterable, Iterator, Mapping, Set
 from types import MappingProxyType
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, TypeAlias
 
 import pydantic
 from lsst.utils.classes import cached_getter, immutable
@@ -526,12 +526,24 @@ class DimensionGroup:  # numpydoc ignore=PR02
         This satisfies the `pydantic.WithInfoPlainValidatorFunction` signature.
         """
         universe = pydantic_utils.get_universe_from_context(info.context)
+        return cls.from_simple(data, universe)
+
+    @classmethod
+    def from_simple(cls, data: SerializedDimensionGroup, universe: DimensionUniverse) -> DimensionGroup:
+        """Create an instance of this class from serialized data.
+
+        Parameters
+        ----------
+        data : `SerializedDimensionGroup`
+            Serialized data from a previous call to ``to_simple``.
+        universe : `DimensionUniverse`
+            Dimension universe in which this dimension group will be defined.
+        """
         return universe.conform(data)
 
-    def _serialize(self) -> list[str]:
-        """Pydantic serializer for `DimensionGroup`.
-
-        This satisfies the `pydantic.PlainSerializerFunction` signature.
+    def to_simple(self) -> SerializedDimensionGroup:
+        """Convert this class to a simple data format suitable for
+        serialization.
         """
         return list(self.names)
 
@@ -555,6 +567,9 @@ class DimensionGroup:  # numpydoc ignore=PR02
             ),
             # When serializing convert it to a `list[str]`.
             serialization=core_schema.plain_serializer_function_ser_schema(
-                cls._serialize, return_schema=list_of_str_schema
+                cls.to_simple, return_schema=list_of_str_schema
             ),
         )
+
+
+SerializedDimensionGroup: TypeAlias = list[str]
--- a/python/lsst/daf/butler/direct_query_driver/_driver.py
+++ b/python/lsst/daf/butler/direct_query_driver/_driver.py
@@ -42,6 +42,7 @@ import sqlalchemy
 
 from .. import ddl
 from .._dataset_type import DatasetType
+from .._exceptions import InvalidQueryError
 from ..dimensions import DataIdValue, DimensionGroup, DimensionRecordSet, DimensionUniverse, SkyPixDimension
 from ..name_shrinker import NameShrinker
 from ..queries import tree as qt
@@ -854,7 +855,7 @@ class DirectQueryDriver(QueryDriver):
                 if governor in self._defaults.dataId.dimensions:
                     result.constraint_data_id[governor] = self._defaults.dataId[governor]
                 else:
-                    raise qt.InvalidQueryError(
+                    raise InvalidQueryError(
                         f"Query 'where' expression references a dimension dependent on {governor} without "
                         "constraining it directly."
                     )
@@ -937,7 +938,7 @@ class DirectQueryDriver(QueryDriver):
             # This is really for server-side defensiveness; it's hard to
             # imagine the query getting different dimensions for a dataset
             # type in two calls to the same query driver.
-            raise qt.InvalidQueryError(
+            raise InvalidQueryError(
                 f"Incorrect dimensions {result.dimensions} for dataset {dataset_type_name} "
                 f"in query (vs. {self.get_dataset_type(dataset_type_name).dimensions.as_group()})."
             )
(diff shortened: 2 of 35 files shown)


python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -329,7 +346,7 @@ def count(
             builder.postprocessing = Postprocessing()
         if builder.postprocessing:
             if not discard:
-                raise RuntimeError("Cannot count query rows exactly without discarding them.")
+                raise InvalidQueryError("Cannot count query rows exactly without discarding them.")
```
dhirving, line 349: This one and the similar one below are the only exceptions that changed type... I think these are roughly in the same spirit as the other InvalidQueryErrors in that the user can fix it by changing one of the parameters in their function call.

python/lsst/daf/butler/remote_butler/server_models.py
```diff
@@ -141,3 +154,93 @@ class QueryCollectionsResponseModel(pydantic.BaseModel):
...
+    MaterializedQuery | DataCoordinateUpload, pydantic.Discriminator("type")
+]
+"""Information about additional data tables that may be used by a query."""
+
+
+class QueryInputs(pydantic.BaseModel):
+    """Serialized Butler query with additional context needed to execute it."""
+
+    tree: SerializedQueryTree
+    default_data_id: SerializedDataCoordinate
```
dhirving, line 190: Really the only rough edge in the separation of concerns is that the default data ID had to passed to the server here so the DirectQueryDriver can use it.

Not really a big deal, and many of the other Butler methods have a similar issue. When I implement default data ID for RemoteButler basically every method that uses `findDataset` internally is going to need the default data ID sent like this.

dhirving (commented):
Basically zero of this code is going to survive into the final version, but it passes all tests, proves that the serialization is viable, and puts the framework in place to start making things more real.

For the most part I just serialize the stuff sent to RemoteQueryDriver, forward it to the server, and pass it on to DirectQueryDriver there.  Materialization and data coordinate upload are basically stubs -- they just capture what the user asked to materialize/upload, and they get re-executed for each query.

This turned into a bit of a blob, sorry.  A lot of the noise is coming from:
* InvalidQueryError became a top-level Butler exception and is now a ButlerUserError that can be sent from the server.
* Default data ID in the RemoteButler registry defaults was not implemented yet, and I had to stub it out to get tests passing.  "Finishing" a RegistryDefaults requires a dimension universe.  Some of the RemoteButler tests weren't able to handle network traffic during RemoteButler construction, so that needing some tweaking throughout.  Unfortunately we can't defer finishing the RegistryDefaults because without it we can't raise errors for typos in the kwargs passed to the Butler constructor.

python/lsst/daf/butler/pydantic_utils.py
```diff
@@ -163,7 +163,9 @@ def validated(self, **kwargs: Any) -> _T:
             call to `validated`, *which will ignore ``**kwargs``*.
         """
         if not self._is_validated:
-            self._data = self._get_wrapped_type_adapter().validate_python(self._data, context=kwargs)
+            self._data = self._get_wrapped_type_adapter().validate_python(
+                self._data, strict=False, context=kwargs
```
dhirving, line 171: Without strict=False here, Pydantic wasn't doing any of its usual conversions like list -> tuple etc, which meant that a lot of the things in QueryTree weren't able to roundtrip through JSON.

python/lsst/daf/butler/remote_butler/server_models.py
TallJimbo: The default collections are not a similar problem because it's straightforward to "apply" them on the client by modifying the `QueryTree` before we send it to the server, right?

I think we could probably do that for the default data ID, too, by refactoring some or all of the `_DataIdExtractionVisitor` out of the `direct_query_driver` package and running it in the client to fill in the values that would be provided by the defaults (e.g. by `AND'ing` in more predicates to `QueryTree.predicate`.

While that means a bit more logic is running on the client side, I don't think it's too complex, and it has the advantage of providing feedback on bad queries faster and without a server call.

But we'll eventually have per-user query state that will have to be persistent server-side (materializations, data ID uploads, and result pages), and possibly some caches that really want to be per-user somewhere, and whatever we do to solve those might emit a solution for default data IDs as a byproduct.

As for `find_dataset`, I'll be interested to see whether we actually still need a server endpoint for that after we add dataset query support to the full query system.  We might be able to just turn it into a client-side convenience method.

python/lsst/daf/butler/direct_query_driver/_driver.py
TallJimbo: FWIW, in both of these cases I thought of them differently because when they occur is tied into implementation details (i.e. postprocessing) that I don't really want users to think about.  But that's kind of intrinsic to the `execute=False` and `discard=False` arguments, so I agree this change makes sense.

python/lsst/daf/butler/dimensions/_group.py
```diff
@@ -526,12 +526,24 @@ def _validate(cls, data: Any, info: pydantic.ValidationInfo) -> DimensionGroup:
...
+        data : `SerializedDimensionGroup`
+            Serialized data from a previous call to ``to_simple``.
+        universe : `DimensionUniverse`
+            Dimension universe in which this dimension group will be defined.
+        """
+        return universe.conform(data)
+
+    def to_simple(self) -> SerializedDimensionGroup:
+        """Convert this class to a simple data format suitable for
+        serialization.
```
TallJimbo, line 546: I'm not 100% on board with adding `to/from_simple` for types that are naturally Pydantic-friendly; I'd hoped to move away from them instead, as we made more things naturally Pydantic-friendly.  But I'm guessing this is to avoid using `DeferredValidation` on some other model, and that's a carve-out I can live with.

TallJimbo (approved):

python/lsst/daf/butler/remote_butler/server_models.py
timj: > But we'll eventually have per-user query state that will have to be persistent server-side (materializations, data ID uploads, and result pages)

I must have missed something because I thought we were wanting the server to be mostly stateless. Query workers would be running and have an ID that would be used by the client to find what the worker state was but the server wouldn't be keeping track of it itself internally -- there would not be a worker queue per server instance. I thought that temp tables in specific database servers known to specific butler servers was difficult unless you use a websocket or something to keep the connection from the client pinned to a specific server?

TallJimbo: I think this is in the space between "mostly" and "completely" stateless.

I had imagined temp tables being emulated via a reverse of the parquet-pagination approach, where we'd dump them to a file on the server and re-upload to the database as needed, but I'm not sure that actually makes the most sense - or even any sense at all, if it's a bigger hop from server to DB than it is from client to server.  Regular database tables that we clean up manually may be another option, and `raise NotImplementedError` might be, too.

We've also talked about spawning a new process for each query, and while that wouldn't solve this problem directly, I could imagine us spawning a new process at the start of each user query context instead if we find the overheads to be a problem, and regardless it's another kind of user-specific resource the server has to manage the lifetime of.

All of these things are difficult, and I'm all for putting them off until we have basic functionality working.

dhirving: The easiest way they would work is basically as Jim described -- they are stored as files in S3, there is an entry in redis so we can expire them later, and that redis entry is tied to the UUID returned to the client for use in its QueryTree.   

It's easiest if the query worker pool re-uploads the data to the DB for each query execution.  In theory we might have the query workers maintain an LRU cache of already-uploaded tables and preferentially route queries to a  database instance that has the needed tables already.

dhirving: i.e. each individual server process is stateless but there is global state for the pool stored in Redis.

python/lsst/daf/butler/dimensions/_group.py
dhirving: Yeah that's why I added this -- I want to keep the top-level request/response models pydantic-context-free so that they can be validated during FastAPI's normal validation process.

In theory I could have avoided the to_simple/from_simple by adding a `SerializedDimensionGroup` as a `DeferredValidation[DimensionGroup]` instead.  But then FastAPI wouldn't be able to structurally validate it during its normal validation process, and to me having two one-line functions that explicitly declare inputs/outputs is easier to understand anyway.

PR #997, opened 2024-04-11
DM-43769: Improve Pydantic support for sphgeom.Region and Timespan

TallJimbo:
## Checklist

- [x] ran Jenkins ([in progress](https://rubin-ci.slac.stanford.edu/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/1123/pipeline))

### DIFF
--- a/python/lsst/daf/butler/pydantic_utils.py
+++ b/python/lsst/daf/butler/pydantic_utils.py
@@ -249,50 +249,31 @@ class DeferredValidation(Generic[_T]):
         return handler.resolve_ref_schema(json_schema)
 
 
-class _SerializedRegionAnnotation:
-    @classmethod
-    def __get_pydantic_core_schema__(
-        cls, _source_type: Any, _handler: pydantic.GetCoreSchemaHandler
-    ) -> core_schema.CoreSchema:
-        # This is Pydantic's way of declaring a "schema" that expects a `str`
-        # and then calls _deserialize.
-        from_str_schema = core_schema.chain_schema(
-            [
-                core_schema.str_schema(),
-                core_schema.no_info_plain_validator_function(cls._deserialize),
-            ]
-        )
-        return core_schema.json_or_python_schema(
-            # When reading JSON, do just that.
-            json_schema=from_str_schema,
-            # When validating Python, do that only if we don't already have a
-            # Region instance.
-            python_schema=core_schema.union_schema([core_schema.is_instance_schema(Region), from_str_schema]),
-            # When serializing to JSON, just call the _serialize method.
-            serialization=core_schema.plain_serializer_function_ser_schema(cls._serialize),
-        )
+def _deserialize_region(value: object, handler: pydantic.ValidatorFunctionWrapHandler) -> Region:
+    if isinstance(value, Region):
+        return value
 
-    @classmethod
-    def __get_pydantic_json_schema__(
-        cls, _core_schema: core_schema.CoreSchema, handler: pydantic.json_schema.GetJsonSchemaHandler
-    ) -> pydantic.json_schema.JsonSchemaValue:
-        # JSON schema is the usual one for `str` fields...
-        result = handler.resolve_ref_schema(handler(core_schema.str_schema()))
-        # ...with a custom description and a note that it's base64-encoded.
-        result["description"] = "A region on the sphere from the lsst.sphgeom package."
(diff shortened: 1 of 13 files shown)


python/lsst/daf/butler/pydantic_utils.py
```diff
@@ -243,3 +247,108 @@ def __get_pydantic_json_schema__(
...
+    @staticmethod
+    def _deserialize(value: str) -> Region:
+        return Region.decode(bytes.fromhex(value))
+
+    @staticmethod
+    def _serialize(region: Region) -> str:
+        return region.encode().hex()
+
+
+SerializableRegion: TypeAlias = Annotated[Region, _SerializedRegionAnnotation]
```
dhirving:
```suggestion
def _deserialize_region(value: object, handler: pydantic.ValidatorFunctionWrapHandler) -> Region:
    if isinstance(value, Region):
        return value

    string = handler(value)
    return Region.decode(bytes.fromhex(string))


def _serialize_region(region: Region) -> str:
    return region.encode().hex()


SerializableRegion: TypeAlias = Annotated[
    Region,
    pydantic.GetPydanticSchema(lambda _, h: h(str)),
    pydantic.WrapValidator(_deserialize_region),
    pydantic.PlainSerializer(_serialize_region),
    pydantic.WithJsonSchema(
        {
            "type": "string",
            "description": "A region on the sphere from the lsst.sphgeom package.",
            "media": {"binaryEncoding": "base16", "type": "application/lsst.sphgeom"},
        }
    ),
]
```

python/lsst/daf/butler/pydantic_utils.py
```diff
@@ -243,3 +247,108 @@ def __get_pydantic_json_schema__(
...
+    @staticmethod
+    def _deserialize(value: int) -> Time:
+        return TimeConverter().nsec_to_astropy(value)
+
+    @staticmethod
+    def _serialize(time: Time) -> int:
+        return TimeConverter().astropy_to_nsec(time)
+
+
+SerializableTime: TypeAlias = Annotated[Time, _SerializedTimeAnnotation]
```
dhirving:
```suggestion
def _deserialize_time(value: object, handler: pydantic.ValidatorFunctionWrapHandler) -> Region:
    if isinstance(value, Time):
        return value

    integer = handler(value)
    return TimeConverter().nsec_to_astropy(integer)


def _serialize_time(time: Time) -> int:
    return TimeConverter().astropy_to_nsec(time)


SerializableTime: TypeAlias = Annotated[
    Time,
    pydantic.GetPydanticSchema(lambda _, h: h(int)),
    pydantic.WrapValidator(_deserialize_time),
    pydantic.PlainSerializer(_serialize_time),
    pydantic.WithJsonSchema(
        {
            "type": "integer",
            "description": "A TAI time represented as integer nanoseconds since 1970-01-01 00:00:00.",
        }
    ),
]
```

tests/test_pydantic_utils.py
```diff
@@ -97,5 +99,42 @@ def test_dump_and_validate(self) -> None:
...
+        python_roundtripped = adapter.validate_python(adapter.dump_python(region))
+        self.assertIsInstance(json_roundtripped, ConvexPolygon)
+        self.assertEqual(python_roundtripped.getVertices(), region.getVertices())
+        with self.assertRaises(ValueError):
+            adapter.validate_python(12)
+        with self.assertRaises(ValueError):
+            adapter.validate_json({})
+        with self.assertRaises(ValueError):
+            adapter.validate_json((b"this is not a region").hex())
+
```
dhirving, line 124: Might be worth adding a test for "a string, but not a hex string."  (Pretty sure it will pass already.)

dhirving (approved):
Looks alright to me.  I suggested a possible implementation that doesn't require resorting to pydantic core.  Whether this is easier to understand is a matter of taste, so take it or leave it. 🤷

dhirving:
Also in the future feel free to delegate any little tickets like this to me, I'm sure you have much bigger fish to fry :D

python/lsst/daf/butler/pydantic_utils.py
TallJimbo: Thanks - I'll go with your versions, because while the meager `pydantic_core` docs clicked with my brain better than the meager `GetPydanticSchema` / ``WrapValidator` docs, I think your version is a bit more likely to be stable and see its docs improve in the future.

PR #999, opened 2024-04-22
DM-43020: Implement region and time extraction for preload

kfindeisen:
This PR adds support for group datasets and the `Timespan` and `RegionTimeInfo` storage classes.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -415,3 +415,5 @@ storageClasses:
     pytype: lsst.cell_coadds.MultipleCellCoadd
   NNModelPackagePayload:
     pytype: lsst.meas.transiNet.modelPackages.formatters.NNModelPackagePayload
+  Timespan:
+    pytype: lsst.daf.butler.Timespan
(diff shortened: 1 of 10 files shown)


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -417,3 +417,6 @@ storageClasses:
     pytype: lsst.meas.transiNet.modelPackages.formatters.NNModelPackagePayload
   Timespan:
     pytype: lsst.daf.butler.Timespan
+  RegionTimeInfo:
+    pytype: lsst.pipe.base.utils.RegionTimeInfo
+    # TODO: add region and timespan components once Region is Butler-persistable
```
TallJimbo: I think you can add them now once they have storage classes (you've done `Timespan`, and are welcome to do `Region` as well).  They don't need to have a formatter, since accessing components on a `RegionTimeInfo` (or anything else stored as JSON) is going to use the `StorageClassDelegate` to read the whole thing and then do attribute access, and I think all the defaults should just work for that.  Given that, I'm not sure adding the components is actually useful, but it's not going to get any more doable in the future.

TallJimbo (approved):

PR #1001, opened 2024-04-24
DM-44050: Configure Postgres connection pool parameters

dhirving:
Adjusted SQLAlchemy's connection parameters for postgres:
* Enabled `pool_pre_ping` to prevent exceptions from being thrown due to stale DB connections.
* Increased the pool size from 1 to 10, to prevent connections from being re-created excessively in multi-threaded services.

The pool_size increase is less risky than it at first appears... only a handful of services are constructing Butlers in a way that causes more than 1 connection to ever be created in the pool:
* At SLAC, it's just the Butler server which has only 1 instance and is rarely used currently.
* At the summit, it is only exposurelog.  This service is pretty low concurrency.  They are going to do a Butler update soon to get pool_pre_ping, so I will let them know to watch out for other side effects in their testing.
* If changing this breaks IDF we want to know that now, because it means we likely have serious scalability problems.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-44050.bugfix.md
+++ b/doc/changes/DM-44050.bugfix.md
@@ -1,0 +1,1 @@
+Postgres database connections are now checked for liveness before they are used, significantly reducing the chance of exceptions being thrown due to stale connections.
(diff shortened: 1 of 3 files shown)


andy-slac (approved):
Looks great! I'm a bit concerned with a hard-coded `pool_size` value, it could happen that for some services we may want to play with this parameter. It would be nice to have an option for updating that on client side, but our current model does not support it.

PR #1003, opened 2024-04-25
DM-43845: Implement default data ID for RemoteButler

dhirving:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/remote_butler/server_models.py
+++ b/python/lsst/daf/butler/remote_butler/server_models.py
@@ -140,6 +140,18 @@ class GetCollectionSummaryResponseModel(pydantic.BaseModel):
     summary: SerializedCollectionSummary
 
 
+class ExpandDataIdRequestModel(pydantic.BaseModel):
+    """Request model for expand_data_id."""
+
+    data_id: SerializedDataId
+
+
+class ExpandDataIdResponseModel(pydantic.BaseModel):
+    """Response model for expand_data_id."""
+
+    data_coordinate: SerializedDataCoordinate
+
+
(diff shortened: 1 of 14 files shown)


python/lsst/daf/butler/remote_butler/server_models.py
```diff
@@ -140,6 +148,18 @@ class GetCollectionSummaryResponseModel(pydantic.BaseModel):
...
+class ExpandDataIdRequestModel(pydantic.BaseModel):
+    """Request model for expand_data_id."""
+
+    data_id: SerializedDataId
+
+
+class ExpandDataIdResponseModel(pydantic.BaseModel):
+    """Response model for expand_data_id."""
+
+    data_coordinate: SerializedDataCoordinate
```
timj, line 160: It's odd, but we never really use `data_coordinate` in this context. It would normally be called a `data_id` (and the docs call it `expanded`). I'm not asking you to change it, I'm commenting that we use data_id and "data coordinate" interchangeably but never use coordinate in variable names...

timj (approved):
👍

python/lsst/daf/butler/remote_butler/server_models.py
dhirving: There's a clearer distinction for the serialized ones than the unserialized ones... 'SerializedDataCoordinate' means "with dimension records" and 'SerializedDataId' means "no dimension records".  It's a lot fuzzier in the business objects because a DataCoordinate could be either of those, and a DataId can be a DataCoordinate or a dict or something else.

PR #1005, opened 2024-04-26
DM-44095: Implement queryDimensionRecords for RemoteButler

dhirving:
Implemented a large subset of `registry.queryDimensionsRecords` for `RemoteButler` by forwarding to the new query system.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/queries/_expression_strings.py
+++ b/python/lsst/daf/butler/queries/_expression_strings.py
@@ -1,0 +1,296 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This software is dual licensed under the GNU General Public License and also
+# under a 3-clause BSD license. Recipients may choose which of these licenses
+# to use; please see the files gpl-3.0.txt and/or bsd_license.txt,
+# respectively.  If you choose the GPL option then the following text applies
+# (but note that there is still no warranty even if you opt for BSD instead):
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import annotations
+
+from collections.abc import Mapping, Set
+from typing import Literal, NamedTuple, TypeAlias
+
+import astropy.time
+
+from .._exceptions import InvalidQueryError
+from .._timespan import Timespan
+from ..column_spec import ColumnType
+from ..dimensions import DimensionUniverse
+from ..registry.queries.expressions.categorize import (
+    ExpressionConstant,
+    categorizeConstant,
+    categorizeElementId,
+)
+from ..registry.queries.expressions.parser import Node, RangeLiteral, TreeVisitor, parse_expression
+from .tree import (
+    BinaryExpression,
+    ColumnExpression,
+    ComparisonOperator,
+    DimensionFieldReference,
+    DimensionKeyReference,
+    LiteralValue,
+    Predicate,
+    UnaryExpression,
+    make_column_literal,
+)
+
+BindValue = LiteralValue | list[LiteralValue] | tuple[LiteralValue] | Set[LiteralValue]
+
+
+def convert_expression_string_to_predicate(
+    expression: str, *, bind: Mapping[str, BindValue] | None, universe: DimensionUniverse
+) -> Predicate:
+    """Convert a Butler query expression string to a `Predicate` for use in a
+    QueryTree.
+
+    Parameters
+    ----------
+    expression : `str`
+        Butler expression query string, as used by the old query system to
+        specify filtering.
+    bind : `~collections.abc.Mapping` [ `str` , `BindValue` ]
+        User-provided key-value pairs that can be referenced by name in the
+        query expression.
+    universe : `DimensionUniverse`
+        Dimension metadata for the Butler database being queried.
+
+    Returns
+    -------
+    predicate : `Predicate`
+        Predicate corresponding to that filter, for use in `QueryTree`.
+    """
+    if bind is None:
+        bind = {}
+
+    try:
+        tree = parse_expression(expression)
+    except Exception as exc:
+        raise InvalidQueryError(f"Failed to parse expression '{expression}'") from exc
+
+    converter = _ConversionVisitor(bind, universe)
+    predicate = tree.visit(converter)
+    assert isinstance(
+        predicate, Predicate
+    ), "The grammar should guarantee that we get a predicate back at the top level."
+
+    return predicate
+
+
+class _ColExpr(NamedTuple):
+    """Represents a portion of the original expression that has been converted
+    to a ColumnExpression object.
+    """
+
+    # This wrapper object mostly exists to help with typing and match() --
+    # ColumnExpression is a big discriminated union, and mypy was having a lot
+    # of trouble dealing with it in the context of _VisitorResult's extra
+    # layers of union.
+
+    value: ColumnExpression
+
+    @property
+    def column_type(self) -> ColumnType:
+        return self.value.column_type
+
+
+class _Null:
+    """Class representing a literal 'null' value in the expression."""
+
+    column_type: Literal["null"] = "null"
+
+
+class _RangeLiteral(NamedTuple):
+    """Class representing a range expression."""
+
+    value: RangeLiteral
+    column_type: Literal["range"] = "range"
+
+
+class _Sequence(NamedTuple):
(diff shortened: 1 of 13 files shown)


python/lsst/daf/butler/queries/_expression_strings.py
```diff
@@ -0,0 +1,296 @@
...
+        Predicate corresponding to that filter, for use in `QueryTree`.
+    """
+    if bind is None:
+        bind = {}
+
+    try:
+        tree = parse_expression(expression)
+    except Exception as exc:
+        raise InvalidQueryError(f"Failed to parse expression '{expression}'") from exc
+
```
dhirving, line 81: The original implementation does some additional processing on the tree here.  I'm not sure if that is stuff already handled by the new query system's validation logic or if we need to handle it here too.

```
    expr = NormalFormExpression.fromTree(tree, NormalForm.DISJUNCTIVE)
    # Check the expression for consistency and completeness.
    visitor = CheckVisitor(data_id, dimensions, bind, defaults, allow_orphans=allow_orphans)
```

python/lsst/daf/butler/queries/_expression_strings.py
```diff
@@ -0,0 +1,296 @@
...
+class _Sequence(NamedTuple):
+    value: list[ColumnExpression]
+    column_type: Literal["sequence"] = "sequence"
+
+
+_VisitorResult: TypeAlias = Predicate | _ColExpr | _Null | _RangeLiteral | _Sequence
+
+
+class _ConversionVisitor(TreeVisitor[_VisitorResult]):
```
dhirving, line 129: The vast majority of the coverage for the query expression language is on `queryDataIds`, so I'm going to implement that next instead of trying to replicate the tests using `queryDimensionRecords`.

The main unimplemented features here are `IN` and miscellaneous `Time`/`Timespan` conversions in `visitBinaryOp`.

There is missing test coverage in the original implementation as well, so I will backfill the tests and complete this implementation in a follow-up to the implementation of queryDataIds.

python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -3212,18 +3236,19 @@ def do_query(element, datasets=None, collections=None):
...
             Test("detector", "raft,-name_in_raft", (2, 1, 4, 3)),
             Test("detector", "-detector.purpose", (4,), limit=(1,)),
-            Test("detector", "-purpose,detector.raft,name_in_raft", (2, 3), limit=(2, 2)),
             Test("visit", "visit", (1, 2)),
             Test("visit", "-visit.id", (2, 1)),
             Test("visit", "zenith_angle", (1, 2)),
             Test("visit", "-visit.name", (2, 1)),
-            Test("visit", "day_obs,-timespan.begin", (2, 1)),
-        )
+            Test("visit", "day_obs,-visit.timespan.begin", (2, 1)),
```
dhirving, line 3248: When day_obs became its own dimension, this query became ambiguous because `timespan` is present in both day_obs and visit (with different meaning/values).  The new query system correctly threw an exception complaining about the ambiguity, but the old one silently picks the visit one.

python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -3386,7 +3423,7 @@ def testExposureQueries(self):
             [
                 record.id
                 for record in registry.queryDimensionRecords("visit", instrument="HSC")
-                .order_by("id")
+                .order_by("visit")
```
dhirving, line 3434: Similar to above, `id` became ambiguous when `day_obs` became a dimension.

python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -106,6 +107,22 @@ class RegistryTests(ABC):
     """True if the registry class being tested supports regex searches for
     collections."""
 
+    supportsDetailedQueryExplain: bool = True
```
dhirving, line 110: This is mostly stuff that requires `LIMIT 1` execution of subqueries to diagnose.  This is marked as a TODO in DirectQueryDriver, so we can flip this feature flag to get these tests when someone goes to backfill that feature.

python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -106,6 +107,22 @@ class RegistryTests(ABC):
...
+    explanations for queries that return no rows by running additional queries
+    to diagnose the problem.
+    """
+
+    supportsQueryOffset: bool = True
+    """True if the registry class being tested supports the 'offset' parameter
+    to query methods.
+    """
+
+    supportsQueryGovernorValidation: bool = True
```
dhirving, line 121: I think this just wasn't implemented for the new query system?  Not sure how important/desirable this validation is -- I think it's more consistent to just get no results here instead of sometimes throwing an exception depending which field you tried to filter on.

python/lsst/daf/butler/queries/_expression_strings.py
TallJimbo: This stuff should indeed already be handled.

python/lsst/daf/butler/registry/tests/_registry.py
TallJimbo: We've gone back and forth on whether to raise or get no results here, due to conflicting requests from users (which we didn't always recognize as conflicting).  For the new query system I came down on the side of just returning no results, both because I also think that's more consistent and because in the new system (especially thinking about `RemoteButler`) it looked harder to have the exception raised at the point at which an error message would be most useful.

I've added this to the list of changes to be RFC'd on [DM-43587](https://rubinobs.atlassian.net/browse/DM-43587).

python/lsst/daf/butler/queries/_expression_strings.py
TallJimbo: Did `IN` and the `Time`/`Timespan` operators look like they were impossible to represent in the new `Predicate` tree, or did you just not want to do them yet?

Another way to approach the testing is that the coverage for `Predicate` trees and their mapping to SQL should now be pretty decent via the new query system's tests, so just testing the conversion from string to `Predicate` tree is probably sufficient for poking at all of the edge cases.

python/lsst/daf/butler/queries/_expression_strings.py
```diff
@@ -0,0 +1,296 @@
...
+                # I don't know how to provide.
+                raise NotImplementedError("INGEST_DATE not supported yet")
+            case ExpressionConstant.NULL:
+                return _Null()
+            case None:
+                pass
+            case _:
+                raise AssertionError("Unhandled expression constant")
+
+        element, column = categorizeElementId(self.universe, name)
```
TallJimbo: I'd prefer to delegate to `lsst.daf.butler.queries.convert_args.interpret_identifier` here.  That will require initializing the visitor with more information about the state of the query the expression is being added to, but that'll solve the problem of how to handle `INGEST_DATE`, provide hopefully-better error messages, and make `where` and `order_by` identifier intepretation more consistent.

(I actually think it will only recognize lower-case `ingest_date`, so we may need some case-wrangling in `interpret_identifier).

python/lsst/daf/butler/queries/_expression_strings.py
```diff
@@ -0,0 +1,296 @@
...
+        except ValueError:
+            # int() raises for float-like strings
+            numeric = float(value)
+        return _make_literal(numeric)
+
+    def visitParens(self, expression: _VisitorResult, node: Node) -> _VisitorResult:
+        return expression
+
+    def visitPointNode(self, ra: _VisitorResult, dec: _VisitorResult, node: Node) -> _VisitorResult:
+        raise NotImplementedError("POINT() function is not supported yet")
```
TallJimbo, line 211: I'm pretty confident it's never actually been supported, at least not beyond the parser.

python/lsst/daf/butler/remote_butler/registry/_query_dimension_records.py
```diff
@@ -0,0 +1,137 @@
...
+    @property
+    def element(self) -> DimensionElement:
+        return self._args.element
+
+    def __iter__(self) -> Iterator[DimensionRecord]:
+        with self._build_query() as result:
+            # We have to eagerly fetch the results to prevent
+            # leaking the resources associated with QueryDriver.
+            records = list(result)
+        yield from records
```
TallJimbo: I suspect (but do not know) that `return iter(records)` is a little more efficient in cases like this.

python/lsst/daf/butler/remote_butler/registry/_query_dimension_records.py
```diff
@@ -0,0 +1,137 @@
...
+        with self._query_factory() as query:
+            a = self._args
+            for dataset_type in a.dataset_types:
+                query = query.join_dataset_search(dataset_type, a.collections)
+            if a.where:
+                query = query.where(a.where, bind=a.bind)
+            if a.dataId:
+                query = query.where(a.dataId)
+            if a.kwargs:
+                query = query.where(a.kwargs)
```
TallJimbo: We should combine these last two cases into one:
```
if a.dataId:
     kwargs = a.kwargs if a.kwargs else {}  # or use default_factory=dict in dataclass
     query = query.where(a.dataId, **a.kwargs)
```
This lets the values in `kwargs` replace those in `dataId`, which is what users have come to expect from data IDs in kwargs, instead of putting the conflicting constraint into the query and dooming it yield no results.

TallJimbo (approved):

python/lsst/daf/butler/queries/_expression_strings.py
dhirving: At first glance they appear to be relatively straightforward to implement -- there are just a lot of cases and no test coverage, so I wanted to defer to a later PR with less going on.

PR #1009, opened 2024-05-01
DM-44147: Allow alternate dimensions in file template

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastore/file_templates.py
+++ b/python/lsst/daf/butler/datastore/file_templates.py
@@ -334,6 +334,14 @@ class FileTemplate:
     ``detector.name_in_raft`` would use the name of the detector within the
     raft.
 
+    In some cases the template may want to support multiple options for a
+    single part of the template. For example, you may not want to include
+    ``group`` if ``exposure`` is in the data ID. To handle this situation a
+    ``|`` character can be used to specify multiple data Id keys in the
+    same format specifier. For example ``{exposure.obs_id|group}`` would
+    choose ``exposure.obs_id`` if ``exposure`` is in the data ID but otherwise
+    would use ``group``.
+
     The mini-language is extended to understand a "?" in the format
     specification. This indicates that a field is optional. If that
     Dimension is missing the field, along with the text before the field,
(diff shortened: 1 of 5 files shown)


python/lsst/daf/butler/datastore/file_templates.py
```diff
@@ -522,6 +531,24 @@ def format(self, ref: DatasetRef) -> str:
         output = ""
 
         for literal, field_name, format_spec, _ in parts:
+            if field_name and "|" in field_name:
+                alternates = field_name.split("|")
+                for alt in alternates:
+                    if "." in alt:
+                        primary, _ = alt.split(".")
```
timj, line 548: In theory we could extend the test here so that if someone asked for `exposure.x|exposure` (say, because `x` was new in the universe but if it wasn't there fall back to `exposure`) fairly simply. I'm not sure if it's worth doing though.

python/lsst/daf/butler/datastore/file_templates.py
```diff
@@ -406,25 +414,26 @@ def grouped_fields(self) -> tuple[dict[str, set[str]], dict[str, set[str]]]:
             "parent": set(),
         }
```
TallJimbo, line 425: Preexisting, but it'd be great to annotate these return objects with `TypedDict` (or just use dataclasses if that wouldn't break anything).

python/lsst/daf/butler/datastore/file_templates.py
```diff
@@ -406,25 +414,26 @@ def grouped_fields(self) -> tuple[dict[str, set[str]], dict[str, set[str]]]:
...
+                for field_name in field_names.split("|"):  # Treat alternates as equals.
+                    subfield = None
+                    key = "standard"
+                    if field_name in self.specialFields:
+                        key = "special"
+                    elif "." in field_name:
+                        # This needs to be added twice.
+                        subfield = field_name
+                        key = "parent"
+                        field_name, _ = field_name.split(".")
```
TallJimbo, line 441: I think the logic would be clearer if you moved the `if "?"` block that sets `target` above this one, so you don't need the separate `if subfield` later.

python/lsst/daf/butler/datastore/file_templates.py
TallJimbo: If we don't have a use case let's not do it.  Even though I buy that the approach you've taken on this ticket is the minimal change to solve the problem at hand, I'd still like to search in the future for ways to simplify all of this by offering less functionality, as opposed to extending that functionality further.

TallJimbo (approved):

python/lsst/daf/butler/datastore/file_templates.py
timj: I used `TypedDict` -- it did require a slight reorganization of the code since you can't say `mydict[var]` -- it has to be a literal. I didn't use a dataclass because that would be a much bigger change and this method is a bit of a hotspot.

PR #1014, opened 2024-05-13
DM-44294: Initial queryDataIds implementation for RemoteButler

dhirving:
Add minimal support for DataCoordinate result type for the new query system, and shim RemoteButler.registry.queryIds to use it.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_query_driver/_driver.py
+++ b/python/lsst/daf/butler/direct_query_driver/_driver.py
@@ -226,6 +226,7 @@ class DirectQueryDriver(QueryDriver):
             name_shrinker=builder.joiner.name_shrinker,
             postprocessing=builder.postprocessing,
             raw_page_size=raw_page_size,
+            column_order=builder.columns.get_column_order(),
         )
         result_page = cursor.next()
         if result_page.next_key is not None:
@@ -1094,6 +1095,9 @@ class _Cursor:
     raw_page_size : `int`
         Maximum number of SQL result rows to return in each page, before
         postprocessing.
+    column_order : `ColumnOrder`
+        Definition of the columns available in the rows returned by the SQL
+        query.
     """
 
     def __init__(
@@ -1104,6 +1108,7 @@ class _Cursor:
         name_shrinker: NameShrinker | None,
         postprocessing: Postprocessing,
         raw_page_size: int,
+        column_order: qt.ColumnOrder,
     ):
         self._result_spec = result_spec
         self._name_shrinker = name_shrinker
@@ -1111,6 +1116,7 @@ class _Cursor:
         self._postprocessing = postprocessing
         self._timespan_repr_cls = db.getTimespanRepresentation()
         self._context = db.query(sql, execution_options=dict(yield_per=raw_page_size))
+        self._column_order = column_order
         cursor = self._context.__enter__()
         try:
             self._iterator = cursor.partitions()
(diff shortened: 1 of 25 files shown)


python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -1222,3 +1224,33 @@ def _convert_dimension_record_results(
...
+    next_key : `PageKey` or `None`
+        Key for the next page to add into the returned page object.
+
+    Returns
+    -------
+    result_page : `DataCoordinateResultPage`
+        Page object that holds a `DataCoordinate` container.
+    """
+    rows = [
+        DataCoordinate.standardize(cast(Mapping, row._mapping), dimensions=spec.dimensions)
```
TallJimbo: `standardize` is full of checks and argument-wrangling we'd prefer to avoid here.  I tried to set things up such that
```
DataCoordinate.from_full_values(spec.dimensions, tuple(row[:len(spec.dimensions)])
```
will work, ultimately because that's the [iteration order for `ColumnSet`](https://github.com/lsst/daf_butler/blob/faa84923553689904c74f88840b8316e8f1de562/python/lsst/daf/butler/queries/tree/_column_set.py#L249-L251).  I'm not thrilled with that just being an unenforced behavioral requirement imposed on one piece of code by another piece of code several layers away, though, so I'm open to opinions on how to clean that up.

python/lsst/daf/butler/remote_butler/server_models.py
```diff
@@ -236,12 +236,35 @@ class QueryExecuteRequestModel(pydantic.BaseModel):
...
 
-class QueryExecuteResponseModel(pydantic.BaseModel):
-    """Response model for /query/execute/."""
+class DataCoordinateResultModel(pydantic.BaseModel):
+    """Result model for /query/execute/ when user requested DataCoordinate
+    results.
+    """
+
+    type: Literal["data_coordinate"] = "data_coordinate"
+    rows: list[SerializedDataCoordinate]
```
TallJimbo, line 245: This is probably not the right time (I'd wait until we do `.expanded()`), but I think we're going to want this model to factor out the `DimensionGroup` that's common to all rows and store a `tuple` of values for each row, instead of using `SerializedDataCoordinate`.

I think we'll want something similar for dimension record results, too, but that's harder, because while `DataCoordinate` values are just `str` or `int`, `DimensionRecord` can have `Region` and `Timespan` and hence needs some kind of serialization logic involved.

TallJimbo (approved):
Looks great.  Only comments are probably-premature optimization things.

python/lsst/daf/butler/remote_butler/server_models.py
dhirving: Yeah these response models are basically just a "simplest thing that could possibly work" placeholder.  All the result data is eventually going to go through the pagination mechanism and I assume probably parquet or some other more efficient format.

python/lsst/daf/butler/direct_query_driver/_driver.py
dhirving: Yeah I don't think conceptually it's too terrible, since the `ResultSpec` we have available here owns the list of dimensions (`spec.dimensions`) and also owns the process of deciding what columns will be in the result (`spec.get_result_columns`).

Maybe I can add a method to ColumnSet (or ResultSpec-delegating-to-ColumnSet) that is equivalent to the `row[:len(spec.dimensions)]` bit of your code, and put it right next to `__iter__` so it's more local.  Or something similar to that -- basically explicitly keep track of which columns correspond to what somewhere.

dhirving: How do you feel about this?: https://github.com/lsst/daf_butler/pull/1014/commits/d8a0a1bb9a772fa55a533722811bf04452ffbc78

(edited: the original commit didn't work because I didn't know the difference between DimensionGroup.names and DimensionGroup.data_coordinate_keys).

PR #1015, opened 2024-05-13
DM-44280: Allow transfers with compatible dataset type definitions

timj:
Previously, the dataset types had to match exactly. Now the transfer is allowed even if the storage classes differ so long as the storage classes are compatible.

## Checklist

- [x] ran Jenkins


TallJimbo (approved):
Looks great.

Out of scope for this ticket: reading code using `DatasetType.is_convertible` is always hard because it isn't obvious from the method name which direction it's testing.  We might want to rename it to `can_convert_from` or `can_convert_to` and consider adding a `can_roundtrip` or something for bidirectional-comparison testing.

timj:
Regarding naming, yes, I also have to ponder what it means at times. The implementation uses `StorageClass.can_convert()` which makes more sense but I called it `DatasetType.is_compatible_with()` because it was taking a more holistic view of compatibility by comparing dimensions and storage classes.

PR #1016, opened 2024-05-17
DM-44414: Initial implementation of queryDatasets for RemoteButler

dhirving:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_query_driver/_result_page_converter.py
+++ b/python/lsst/daf/butler/direct_query_driver/_result_page_converter.py
@@ -33,10 +33,18 @@ from typing import TYPE_CHECKING
 
 import sqlalchemy
 
+from .._dataset_ref import DatasetRef
+from .._dataset_type import DatasetType
 from ..dimensions import DataCoordinate, DimensionGroup, DimensionRecordSet, SkyPixDimension
 from ..queries import tree as qt
-from ..queries.driver import DataCoordinateResultPage, DimensionRecordResultPage, PageKey, ResultPage
-from ..queries.result_specs import DataCoordinateResultSpec, DimensionRecordResultSpec
+from ..queries.driver import (
+    DataCoordinateResultPage,
+    DatasetRefResultPage,
+    DimensionRecordResultPage,
+    PageKey,
+    ResultPage,
+)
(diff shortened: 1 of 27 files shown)


python/lsst/daf/butler/direct_query_driver/_result_page_converter.py
```diff
@@ -129,12 +137,45 @@ def convert(
...
+        self._data_coordinate_converter = _DataCoordinateRowConverter(spec.dimensions, column_order)
+        self._column_order = column_order
+
+    def convert(
+        self,
+        raw_rows: Iterable[sqlalchemy.Row],
+        next_key: PageKey | None,
+    ) -> DatasetRefResultPage:
+        run_column = qt.ColumnSet.get_qualified_name(self._spec.dataset_type_name, "run")
+        dataset_id_column = qt.ColumnSet.get_qualified_name(self._spec.dataset_type_name, "dataset_id")
```
TallJimbo, line 164: I think these two names will need to get some name-shrinking applied.  I think everything else is safe.  I thought we already had a test that would flag this, but if not, I think we need to add one with a very long dataset type name.

python/lsst/daf/butler/remote_butler/registry/_query_datasets.py
```diff
@@ -0,0 +1,120 @@
...
+    """Implementation of the legacy ``DimensionRecordQueryResults`` interface
+    using the new query system.
+
+    Parameters
+    ----------
+    query_factory : `QueryFactory`
+        Function that can be called to access the new query system.
+    args : `CommonQueryArguments`
+        User-facing arguments forwarded from
+        ``registry.queryDimensionRecords``.
```
TallJimbo: Should be `queryDatasets`.

TallJimbo (approved):
Looks great.

I hope we can migrate our way out of the new `ingest_date` specialization in the near future.

python/lsst/daf/butler/direct_query_driver/_result_page_converter.py
dhirving: Added a note to DM-44505 to deal with this... I'll check in with Andy on nameshrinker stuff and decide the best way to pass everything down to here.

queryDatasets test coverage is actually pretty sparse... there's a lot more cover of queryDataIds and I think it's assuming that most of the implementation is common between the two.

PR #1017, opened 2024-05-17
DM-44399: Add more helpful exception notes to parquet serializer when it fails.

erykoff:
## Checklist

- [ ] ran Jenkins

Further checking of the exception notes will happen after "formatter v2".


timj (approved):
Looks good. Ned to run black.

PR #1018, opened 2024-05-17
DM-26658: Re-implement the Formatter class

timj:
* Use entirely new class: FormatterV2
* Provide FormatterV1inV2 shim class.
* Since FormatterV2 needs access to the cache, moved all the cache code to FormatterV2 itself out of Datastore.
* Three read methods can be provided, two write methods.
* Can now read content from within zip archives.

For read:

* `read_from_uri` gives you the full URI to do with whatever you want but *might* give you a locally file if it was in the cache.
* `read_from_stream` gives you a file handle for the resource.
* `read_from_local_file` gives you a guaranteed local file.

Flags declare which of those 3 are available for a given formatter to allow the `FormatterV2.read` method to pick a suitable option based on what it knows (or if a zip file is in play, in which case `read_from_uri` can't be called).

For write:

* `to_bytes` is called first (can be Not Implemented)
* Falls back to `write_local_file`

The default implementation of `write_local_file` calls `to_bytes`.

All daf_butler formatters ported to V2. There are some V1 test formatters since otherwise nothing will test `FormatterV1inV2`.

This mostly seems to work but there are some things I still need to do:

- [x] There is some code duplication within FormatterV2 that I should tidy up.
- [x] I agreed to add any notes from sub exceptions when I re-raise so that notes are more visible in the final stack trace.
- [x] Work out how to prevent storage class conversion before handing off to a formatter.

## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +60,927 @@
...
+class FileIntegrityError(RuntimeError):
+    """The file metadata is inconsistent with the metadata supplied by
+    the datastore.
+    """
+
+
+class FormatterV2(metaclass=ABCMeta):
+    """Interface for reading and writing datasets using URIs.
+
+    The formatters are associated with a particular `StorageClass`.
```
TallJimbo, line 87: One or more storage classes, right?  `ParquetFormatter` speaks can be used with multiple storage classes.

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +60,927 @@
...
 
 
+class FileIntegrityError(RuntimeError):
+    """The file metadata is inconsistent with the metadata supplied by
+    the datastore.
+    """
+
+
+class FormatterV2(metaclass=ABCMeta):
+    """Interface for reading and writing datasets using URIs.
```
TallJimbo, line 85: One thing I was a little uncomfortable with in the original `Formatter` interface is that the same class is responsible for both reading and writing, but the state the class wants to hold for reading is not generally the same as the state it wants to hold for writing.  It's nice to have one class name to configure rather than two, but other than that I think reading and writing are best kept separate.

So, while I realize this would be another pretty big change (apologies for not raising the possibility earlier), what do you think about making the `write` interface go entirely through `classmethod`s, so that constructing an instance is something you do only for reads?  The reason I'd let the instance state be for reads is to give the formatter the URL and `DatasetRef` at construction to let it do some work up front, and then support multiple calls to single-component or dict-of-parameters reads that it may be able to optimize internally (e.g. using [fsspec's buffering/caching file-like object wrappers](https://filesystem-spec.readthedocs.io/en/latest/api.html#readbuffering)).  Getting multiple components or sets of parameters would require new interfaces higher up, too, of course, but I think that's easier than the formatter side.

If we don't go this route, I think we want to be very clear about these being two-state instances that are constructed in different ways for read and write and may have different attributes set as a result.  And we really want to limit to just two states.

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +60,927 @@
...
+            The Python object read from the resource.
+
+        Notes
+        -----
+        This method will only be called if `read_cached_file` was not called
+        or if `read_from_bytes` was called successfully.
+        """
+        raise NotImplementedError("This formatter does not know how to read a local file.")
+
+    def read_from_bytes(self, serialized_bytes: bytes, component: str | None = None) -> Any:
```
TallJimbo: I think a "read from file-like object" is more general than "read from bytes" and a complete replacement for it.  I wonder if it'd also be a replacement either of the other methods or something they could delegate to by default, but I haven't wrapped my head around the caching enough to have a sense for whether that's the case.

timj:
For the record, we have been asked to support reading a file from within a zip file (where a single zip file is registered as multiple datasets). This might affect some of our thinking here as to the best interface.

python/lsst/daf/butler/_formatter.py
timj: A storage class is associated with one formatter in configuration.  I need to think about how Formatter can say what it supports. The current situation is:

* On read the formatter can see what the requested storage class is and generate that directly.
* On write the datastore does a type conversion before sending the in-memory dataset to the formatter

We would need something like a `FormatterV2.can_accept(in_memory_dataset)` API that would allow a formatter to say it's all under control, bypass the type conversion. Maybe it has to take the storageClass instead. Need to think. It's not solely a case of "can you accept X" it's also an assertion that the formatter can read it back as the registry storage class python type.

timj:
The `.name` property used for stream handles requires lsst/resources#89.

python/lsst/daf/butler/_formatter.py
timj: Writing currently needs to be an instance method because the write parameters are read from configuration and attached to the formatter instance by the formatter factory.

I've only just read the second half of your second paragraph where you are suggesting that `formatter.read()` might want to be called multiple times with different components. Datastore isn't set up to be able to do that at the moment because butler only asks for one component for one dataset or else all components from a disassembled dataset.

TallJimbo: I do think reading N components might be useful, but I don't have a concrete use case in mind and I could imagine it being better implemented by being able to pass a list of components rather than by reworking what is considered formatter instance state.  It's not something I'd go add at a high level right now, just something to keep in mind as a possibility (e.g. is the formatter interface such that we could add a method for that with a default implementation and override it in just some concrete formatters, rather than having to change them all again).

timj: Now that the user subclasses something like `read_from_uri(uri, component)` and doesn't subclass `read` itself then we could pass multiple components to `read` and have that by default be N calls to `read_from_uri`.

python/lsst/daf/butler/_formatter.py
timj: I have implemented `can_accept` so that @erykoff can prevent conversions to DataFrame on write.

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1058 @@
...
+                f"Formatter {self.name()} could not read the file using any method."
+            )
+
+        # First see if the formatter can support direct remote read from
+        # a URI. We allow FormatterNotImplementedError to be raised by
+        # the formatter in case the formatter wishes to defer to the
+        # local file implementation which will trigger a cache write.
+        if self.can_read_from_uri:
+            with contextlib.suppress(FormatterNotImplementedError):
+                return self.read_directly_from_possibly_cached_uri(
```
timj: I wonder if we should trap explicitly the no component / no parameters case and not call the read_from_uri here but defer it to the later situation where we have downloaded and possibly cached the file and then can pass the local URI to read_from_uri. That might be better than having the formatter author decide for themselves whether they might want to use the cached file instead.

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1058 @@
...
+        local cache. Consider raising `FormatterNotImplementedError` in
+        this situation, for example if there are no parameters or component
+        specified, and allowing the system to fall back to calling
+        `read_from_local_file` (which will populate the cache if configured
+        to do so).
+        """
+        raise FormatterNotImplementedError("This formatter does not know how to read a file.")
+
+    def read_from_stream(
+        self, stream: BinaryIO | ResourceHandleProtocol, component: str | None = None
```
timj: Is it worth passing in the expected size to this method (and read_from_uri) in case they want to decide that it would be more efficient to read from a local file and raise in here? (which would need to be caught above).

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+    ----------
+    file_descriptor : `FileDescriptor`, optional
+        Identifies the file to read or write, and the associated storage
+        classes and parameter information.  Its value can be `None` if the
+        caller will never call `Formatter.read` or `Formatter.write`.
+    ref : `DatasetRef`
+        The dataset associated with this formatter. Should not be a component
+        dataset ref.
+    write_parameters : `dict`, optional
+        Any parameters to be hard-coded into this instance to control how
```
TallJimbo: "hard-coded" seems like a strange choice of words for something that seems explicitly configurable.

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+        the dataset is serialized.
+    write_recipes : `dict`, optional
+        Detailed write Recipes indexed by recipe name.
```
TallJimbo: Is Recipes capitalization intentional?  Is there a type this was supposed to link to?

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+    ) -> AbstractDatastoreCacheManager:
+        """Return the cache if given else return a null cache."""
+        if cache_manager is None:
+            # Circular import avoidance.
+            from .datastore.cache_manager import DatastoreDisabledCacheManager
+
+            cache_manager = DatastoreDisabledCacheManager(None, None)
+        return cache_manager
+
+    def read(
```
TallJimbo, line 353: Might want to use `@typing.final` to enforce what the docs say.

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+
+        Notes
+        -----
+        This method will only be called if the class property
+        ``can_read_from_local_file`` is `True` and other options were not
+        used.
+        """
+        raise FormatterNotImplementedError("This formatter does not know how to read a local file.")
+
+    def write(
```
TallJimbo, line 867: Might want to use `@typing.final` to enforce what the docs say.

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+    Notes
+    -----
+    A `FormatterV2` author should not override the default `read` or `write`
+    method. Instead for read the should implement one or all of
+    `read_from_stream`, `read_from_uri`, or `read_from_local_file`. The method
+    `read_from_uri` will always be attempted first and could be more efficient
+    (since it allows the possibility for a subset of the data file to be
+    accessed when parameters or components are specified) but it will not
+    update the local cache. If the entire contents of the remote file are
+    being accessed consider raising `FormatterNotImplementedError` and falling
```
TallJimbo: Did you consider having it _return_ `NotImplemented` instead, like e.g. `__add__` and `__radd__` can?

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+    back to `read_from_local_file` since this will store the file in the
+    cache to allow a more efficient subsequent read.
+
+    Similarly for writes, the recommendation is not to override the `write`
+    method but instead to implement `write_local_file` and, optionally,
+    `to_bytes`. For local URIs the system will always call `write_local_file`
+    to ensure atomic writes are implemented. For remote URIs with local caching
+    disabled, `to_bytes` will be called first.
```
TallJimbo: Implication here is that for remote URIs with local caching enabled, `write_local_file` will also be called first.  Is that right?

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+    """
+
+    unsupported_parameters: ClassVar[Set[str] | None] = frozenset()
+    """Set of read parameters not understood by this `Formatter`. An empty set
+    means all parameters are supported.  `None` indicates that no parameters
+    are supported. These param (`frozenset`).
```
TallJimbo: Unfinished last sentence.

TallJimbo: Might also be clearer to explicitly state that these are read parameters _declared by a storage class that is otherwise supported by this formatter_.  Even though I should know better it took me a moment to figure out what set of parameters this represented a difference from.

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+    """
+
+    supported_write_parameters: ClassVar[Set[str] | None] = None
+    """Parameters understood by this formatter that can be used to control
+    how a dataset is serialized. `None` indicates that no parameters are
+    supported."""
+
+    default_extension: str | None = None
```
TallJimbo: Not a `ClassVar`, then?

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+    """
+
+    can_read_from_uri: bool = False
+    """Declare whether `read_from_uri` is available to this formatter."""
+
+    can_read_from_stream: bool = False
+    """Declare whether `read_from_stream` is available to this formatter."""
+
+    can_read_from_local_file: bool = False
+    """Declare whether `read_from_file` is available to this formatter."""
```
TallJimbo, line 163: Are these `can_read_` attributes intentionally not `ClassVar`?

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+        # Compatibility parameters. Unused in v2.
+        **kwargs: Any,
+    ):
+        if not isinstance(file_descriptor, FileDescriptor):
+            raise TypeError("File descriptor must be a FileDescriptor")
+
+        self._file_descriptor = file_descriptor
+
+        if ref.isComponent():
+            # Raise or convert to composite ref?
```
TallJimbo: I think I lean towards raising, since that's what the docs imply to me and it seems like a bug in higher-level code if a component gets here and it might be confusing to that code if the component is silently ignored.

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+        return f"{self.name()}@{self.file_descriptor.location.uri}"
+
+    def __repr__(self) -> str:
+        return f"{self.name()}({self.file_descriptor!r})"
+
+    @property
+    def file_descriptor(self) -> FileDescriptor:
+        """File descriptor associated with this formatter (`FileDescriptor`).
+
+        Read-only property.
```
TallJimbo: I don't think you need to call this out here, especially since all the other properties are read-only as well.

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1058 @@
...
+    def _get_cache_ref(self) -> DatasetRef:
+        """Get the `DatasetRef` to use for cache look ups.
+
+        Returns
+        -------
+        ref : `lsst.daf.butler.DatasetRef`
+            The dataset ref to use when looking in the cache.
+            For single-file dataset this will be the dataset ref directly.
+            If this is disassembled we need the component and the component
+             will be in the `FileDescriptor`.
```
TallJimbo:
```suggestion
            will be in the `FileDescriptor`.
```

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+        cache_manager: AbstractDatastoreCacheManager | None = None,
+    ) -> Any:
+        """Read a Dataset.
+
+        Parameters
+        ----------
+        component : `str`, optional
+            Component to read from the file. Only used if the `StorageClass`
+            for reading differed from the `StorageClass` used to write the
+            file.
```
TallJimbo, line 366: Where is the "StorageClass for reading" specified now?  The `DatasetRef` attached to the formatter is now the parent ref, I think.

If the answer is that it's not specified, I think that's great - it'd be a great improvement to just obey `component` and let the "read vs write storage class" stuff focus on storage class conversions.

That does beg the question of what it means to get a component _while_ doing a storage class conversion, but I hope we can get away with banning it, because IIRC QG generation can't handle it, either.

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1058 @@
...
+        expected_size : `int`, optional
+            If known the expected size of the resource to read. This can be
```
TallJimbo:
```suggestion
            If known, the expected size of the resource to read. This can be
```

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+
+        For a single file:
+
+        * No component requested, read the whole file.
+        * Component requested, optionally read the component efficiently,
+          else read the whole file and extract the component.
+        * Derived component requested, read whole file or read relevant
+          component and derive.
+
+        Disassembled Composite:
```
TallJimbo, line 421: Time to ask this again, I think: should we just drop support for disassembled composites?  The too-many-small-files problem that's led to using Zip archives is another argument against it (unless you see us disassembling and then zipping up the per-component files).

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+            # Open the Zip file using ResourcePath.
+            with uri.open("rb") as fd:
+                with zipfile.ZipFile(fd) as zf:  # type: ignore
+                    if self.can_read_from_stream:
+                        zip_fd = zf.open(path_in_zip)
+                        try:
+                            in_memory_dataset = self.read_from_stream(zip_fd, component)
+                        except Exception:
+                            zip_fd.close()
+                            raise
```
TallJimbo: Could this be written as:
```
with zf.open(path_in_zip) as zip_fd:
    ...
```
If not, maybe do:
```
with contextlib.closing(zf.open(path_in_zip)) as zip_fd:
    ...
```

python/lsst/daf/butler/_formatter.py
TallJimbo: If we do run into a case where we find it's important to download and cache the full file in order to do a component read (and it's not just because the formatter doesn't implement partial reads for that component), I think the formatter won't really able to know that any better than the datastore.  And I don't think those cases are likely to begin with, so simplicity arguments probably should carry the day.

In the scenarios where I imagine both a component of a dataset and the full dataset being inputs to a single `Quantum`, we actually don't want to bring down and cache the full dataset if the component is requested first (at least not automatically), because there's good chance the component is only being requested because looking at that might be enough for the task to raise `NoWorkFound`, so the full dataset might never be read.

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+                msg = ""
+
+            with uri.as_local() as local_uri:
+                self._check_resource_size(self.file_descriptor.location.uri, expected_size, local_uri.size())
+                can_be_cached = False
+                if uri != local_uri:
+                    # URI was remote and file was downloaded
+                    cache_msg = ""
+
+                    if cache_manager.should_be_cached(cache_ref):
```
TallJimbo, line 719: If we wanted to support caching from one POSIX location to another faster POSIX location (e.g. NFS mount to local scratch), it would be nice if we could do that by just changing what this method returns - I think Princeton might have some interest in that, since we have some very slow POSIX storage here (though a `ChainedDatastore` or even two repos might be better for us).

Right now, I think the "slow POSIX" location would still be interpreted as "local", which would cause two problems:
 - We'd do the read from that slow location first, and then do another slow copy later to move it into the cache.
 - `uri` would be the same as `local_uri`, so we'd never even get to calling `cache_manager.should_be_cached`.

Having written all that, I suppose the other possibility is to look into changing what `ResourcePath.as_local` does for those files; maybe via a `NetworkPosixResourcePath` implementation or something.

And to be clear, I don't think we need that now, and certainly not on this ticket.  But it'd be nice to have a sense for how we might add it.

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+
+                    msg = f"(via download to local file{cache_msg})"
+
+                log.debug(
+                    "Reading%s from location %s %s with formatter %s",
+                    f" component {log_component}" if log_component else "",
+                    uri,
+                    msg,
+                    self.name(),
+                )
```
TallJimbo: Is this logging redundant with the the timer logging below?

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+            Raised if there is no support for direct reads from a, possibly,
+            remote URI.
+
+        Notes
+        -----
+        This method is only called if the class property
+        ``can_read_from_uri`` is set to `True`.
+
+        It is possible that a cached local file will be given to this method
+        even if it was originally a remote URI. This can happen if the write
```
TallJimbo: "if the read", you mean?  And what precisely is "it" in this sentence?  If the formatter can't see what the original URI was at all anyway, it may be better not to mention this.

python/lsst/daf/butler/_formatter.py
TallJimbo: I think so, not because I think it's likely to be used, but because adding a new argument in the future would be a breaking change and it seems pretty harmless now.

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1058 @@
...
+
+    def read_from_stream(
+        self, stream: BinaryIO | ResourceHandleProtocol, component: str | None = None
+    ) -> Any:
+        """Read from an open file descriptor.
+
+        Parameters
+        ----------
+        stream : `lsst.resources.ResourceHandleProtocol` or \
+                `typing.BinariyIO`
```
TallJimbo:
```suggestion
                `typing.BinaryIO`
```

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+        ----------
+        uri : `lsst.resources.ResourcePath`
+            URI to use to read the dataset. This URI can be local or remote
+            and can refer to the actual resource or to a locally cached file.
+        component : `str` or `None`, optional
+            The component to be read from the dataset.
+
+        Returns
+        -------
+        in_memory_dataset : `typing.Any`
```
TallJimbo: I like `object` better than `typing.Any` for at least the docs (as some of the methods already do).  I think it'd also be better for the type annotations unless it leads to a lot of `type: ignore` or casting.

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+            The Python object read from the stream.
+
+        Notes
+        -----
+        Only called if the class property ``can_read_from_stream`` is `True`.
+        """
+        raise FormatterNotImplementedError("This formatter does not know how to read from a stream.")
+
+    def read_from_local_file(self, local_uri: ResourcePath, component: str | None = None) -> Any:
```
TallJimbo: Since this has to be a local file, it might be better to just pass `local_uri.ospath` (i.e. a `str`) as the first argument.

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+                serialized_dataset = self.to_bytes(in_memory_dataset)
+            except FormatterNotImplementedError:
+                # Fallback to the file writing option.
+                pass
+            except Exception as e:
+                e.add_note(
+                    f"Failed to serialize dataset {self.dataset_ref} of "
+                    f"type {get_full_type_name(in_memory_dataset)} to bytes."
+                )
+                raise
```
TallJimbo, line 964: It might be better to re-raise with `from e` in order to standardize the exception type here so high level code doesn't have to worry about what I/O errors might manifest as.  I don't have a super strong opinion about that, but it's worth thinking about.

python/lsst/daf/butler/_formatter.py
```diff
@@ -75,13 +1109,22 @@ class Formatter(metaclass=ABCMeta):
         the dataset is serialized.
     writeRecipes : `dict`, optional
         Detailed write Recipes indexed by recipe name.
+    **kwargs
+        Additional parameters that can allow parameters
+        from `FormatterV2` to be provided.
```
TallJimbo, line 1219: Does this mean that any concrete V1 formatters now need to accept `**kwargs` as well, because V2 args are going to be passed in?

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+        The dataset associated with this formatter. Should not be a component
+        dataset ref.
+    write_parameters : `dict`, optional
+        Any parameters to be hard-coded into this instance to control how
+        the dataset is serialized.
+    write_recipes : `dict`, optional
+        Detailed write Recipes indexed by recipe name.
+    **kwargs
+        Additional arguments that will be ignored but allow for
+        `Formatter` V1 parameters to be given.
```
TallJimbo, line 103: I'm a bit confused as to why we need this.  Under what conditions would we construct a V2 formatter while thinking it's really a V1 formatter?  Seems like we'd just make a `FormatterV1inV2` wrapper instance if we see a V1 formatter configured.

python/lsst/daf/butler/_formatter.py
```diff
@@ -797,5 +1884,109 @@ def registerFormatter(
...
+
+    def segregate_parameters(self, parameters: dict[str, Any] | None = None) -> tuple[dict, dict]:
+        return self._formatter.segregate_parameters(parameters)
+
+    def validate_write_recipes(  # type: ignore
+        self,
+        recipes: Mapping[str, Any] | None,
+    ) -> Mapping[str, Any] | None:
+        # This should be a class method but a class method can not work
+        # for a dynamic shim. Luckily the shim is only used as an instance.
```
TallJimbo, line 2080: Is it possible to make this an instance method across the board, then?

python/lsst/daf/butler/_formatter.py
```diff
@@ -797,5 +1884,109 @@ def registerFormatter(
...
+        # Need to temporarily override the location since the V1 formatter
+        # will not know anything about this local file.
+
+        # V2 does not have a fromBytes equivalent.
+        if self._formatter.can_read_bytes():
+            serialized_dataset = local_uri.read()
+            return self._formatter.fromBytes(serialized_dataset, component=component)
+
+        location = Location(*local_uri.split())
+        with self._formatter._updateLocation(location):
```
TallJimbo, line 2094: This gets me wondering whether the V2inV1 wrapper would be simpler if it deferred constructing the V1 formatter until its methods are called, and maybe didn't even hold it as an instance attribute.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -786,7 +789,9 @@ def _prepare_for_direct_get(
             parameters=parameters,
         )
 
-    def _prepare_for_put(self, inMemoryDataset: Any, ref: DatasetRef) -> tuple[Location, Formatter]:
+    def _prepare_for_put(
+        self, inMemoryDataset: Any, ref: DatasetRef
+    ) -> tuple[Location, Formatter | FormatterV2]:
```
TallJimbo: Could you do the wrapping of `Formatter` into `FormatterV1inV2` earlier (i.e. in this mehtod), to simplify downstream code?

python/lsst/daf/butler/_formatter.py
```diff
@@ -51,12 +63,1034 @@
...
+
+    Parameters
+    ----------
+    file_descriptor : `FileDescriptor`, optional
+        Identifies the file to read or write, and the associated storage
+        classes and parameter information.  Its value can be `None` if the
+        caller will never call `Formatter.read` or `Formatter.write`.
+    ref : `DatasetRef`
+        The dataset associated with this formatter. Should not be a component
+        dataset ref.
```
TallJimbo, line 96: Which storage class should this have?  A storage class appears at least here and (twice) in the `file_descriptor`, and there are many different potential storage classes and scenarios in play (read/write, component/full-dataset/[dis]assembly, in-formatter coercion, datastore-level coercion).

More than anything I'd like to drop support for any of those scenarios we don't need, but given that we've accumulated a lot historically and probably can't drop anything other than edge-case combinations without a deprecation period, we probably need to document all possible scenarios (of course, with the rules for those as simple as we can make them).

python/lsst/daf/butler/formatters/pickle.py
```diff
@@ -34,10 +34,35 @@
...
+    can_read_from_uri = True
+
+    def read_from_uri(self, uri: ResourcePath, component: str | None = None) -> Any:
+        # Read the pickle file directly from the resource into memory.
+        try:
+            data = pickle.loads(uri.read())
+        except pickle.PicklingError:
+            data = None
+
+        return data
```
TallJimbo, line 58: For a simple formatter like this one, where all we need is bytes, I was expecting it'd be better practice to implement `read_from_stream` rather than `read_from_uri`, because I think of the former as more general.  I can see why others might think the opposite, though, since we could in theory add a `ResourcePath` implementation for anything, even though there are stream types for which there is at present no `ResourcePath` implementation (e.g. `io.BytesIO`).  It's not a big deal because I think either is fine, but it might be good to think about what we recommend when there doesn't seem to be an obvious reason to prefer one or the other.

python/lsst/daf/butler/formatters/typeless.py
```diff
@@ -0,0 +1,223 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Support for reading and writing files to a POSIX file system."""
```
TallJimbo, line 28: Module docstring doesn't seem accurate.

python/lsst/daf/butler/formatters/typeless.py
```diff
@@ -0,0 +1,223 @@
...
+        component: str | None = None,
+        expected_size: int = -1,
+        cache_manager: AbstractDatastoreCacheManager | None = None,
+    ) -> Any:
+        # Do the standard read of the base class.
+        log = logging.getLogger("lsst.tests")
+        log.warning("About to call super read")
+        data = super().read(component, expected_size, cache_manager)
+
+        log.warning("Read data as type *****************************************: %s", type(data))
```
TallJimbo: Debug relic.

python/lsst/daf/butler/formatters/typeless.py
```diff
@@ -0,0 +1,223 @@
...
+
+        # Assemble the requested dataset and potentially return only its
+        # component coercing it to its appropriate pytype.
+        data = self._assemble_dataset(data, component)
+
+        log.warning("Got after assembly: %s", type(data))
```
TallJimbo: Debug relic.

python/lsst/daf/butler/formatters/typeless.py
```diff
@@ -0,0 +1,223 @@
...
+
+        # Special case components by allowing a formatter to return None
+        # to indicate that the component was understood but is missing.
+        if data is None and component is None:
+            raise ValueError(f"Unable to read data with URI {self.file_descriptor.location.uri}")
+
+        log.warning("&************************** Got something of type %s", get_full_type_name(data))
```
TallJimbo: Debug relic.

python/lsst/daf/butler/formatters/typeless.py
```diff
@@ -0,0 +1,223 @@
...
+class TypelessFormatter(FormatterV2):
+    """Formatter V2 base class that attempts to coerce generic objects
+    read in subclasses into the correct Python type.
+
+    Notes
+    -----
+    This class provides a ``read()`` method that will run `FormatterV2.read`
+    and coerce the return type using a variety of techniques. Use the
+    standard `FormatterV2` methods for reading bytes/files and writing
+    bytes/files.
```
TallJimbo, line 53: This gets me thinking that we should have different JSON/YAML formatters for different categories of target types (one for things that are natively pydantic models, another for those that have `to_simple` and `from_simple`, etc.).

May be too late for `PropertySet` vs. `TaskMetadata`, of course, but if we'd had different formatters earlier that transition would have been easier to handle, and I'd like to do better in future similar transitions.

doc/lsst.daf.butler/formatters.rst
```diff
@@ -228,69 +228,69 @@ Formatters
...
-  The interface may evolve as other types of datastore become available and make use of the formatter system.
-  The interface is being reassessed on :jira:`DM-26658`.
+A datastore knows which formatter was used to write or ingest a dataset.
+There are three methods a formatter author can implement in order to read a Python type from a file:
 
-When ingesting files from external sources formatters are associated with each incoming file but these formatters are only required to support a `Formatter.read()` method.
-They must though declare all the file extensions that they can support.
-This allows the datastore to ensure that the image being ingested has not obviously been associated with a formatter that does not recognize it.
+``read_from_local_file``
+    The ``read_from_local_file`` method guarantees to pass in a local file resource.
```
TallJimbo:
```suggestion
    The ``read_from_local_file`` method is guaranteed to be to passed a local file resource.
```

TallJimbo (approved):
Looks great.  I've left a lot of comments because there's a lot here, but overall I'm very happy with all of this.

I think my biggest concern is that I'd like to make it really clear what all the storage classes are in the various places they appear in a formatter, for the various scenarios in which a formatter is used, and I could imagine trying to document/test that more rigorously leading to some changes to make it more consistent.  There's a line comment thread on this we can use for further discussion.

python/lsst/daf/butler/_formatter.py
timj: I was thinking of the non-component read case since that one will need the full file anyhow (if there are no parameters). If a formatter author believes that it's always better to read from the local file then they should only implement `read_from_local_file`. What we don't want is for an author to implement `read_from_uri` by transferring the whole file locally and then opening it -- that would break caching. I think I need to add that to the documentation.

As currently implemented, if someone asks for a component and `read_from_uri` can handle a component read without downloading the whole file, then the whole file won't be cached.

python/lsst/daf/butler/formatters/typeless.py
timj: We could have a PydanticFormatter implementation explicitly. This wouldn't have solved the PropertySet vs TaskMetadata problem though because those are entirely different types and the fix there would have been to have a specialist class from the beginning for task metadata that might have internally used a PropertySet rather than using a generic container class at the start.

python/lsst/daf/butler/_formatter.py
timj: I did not think about that.  That's presumably more efficient than a try/except and we'd have `read` raise `FormatterNotImplementedError` if nothing worked.

python/lsst/daf/butler/_formatter.py
timj: That was not intentional.

python/lsst/daf/butler/_formatter.py
timj: Great idea. Except `TypelessFormatter` only works because it overrides `read`...

python/lsst/daf/butler/_formatter.py
timj: Yes. The code calls write_local_file if the file is local or the ref should be cached. The default implementation of write_local_file is to call to_bytes.

python/lsst/daf/butler/_formatter.py
timj: There is an underlying assumption that if there are multiple storage classes that are configured to use a specific formatter that they have compatible parameter definitions. I probably should also add that a storage class delegate would be required to handle parameters that the formatter can not support.

python/lsst/daf/butler/_formatter.py
timj: It should be. I forgot (mostly because it wasn't defined that way originally for Formatter)

python/lsst/daf/butler/_formatter.py
timj: No. They are class properties.

python/lsst/daf/butler/_formatter.py
timj: The read vs write storage class definitions haven't moved. They are still in `FileDescriptor`.  The write storage class is the storage class that was used to write the dataset -- it could be the component storage class or the composite depending on whether it was disassembled. You can tell if it was disassembled because `FileDescriptor.component` will be defined.

The read storage class used to only differ from the write one for disassembly but now with storage class conversion they can be different intentionally. I need to make sure that Eli is getting the different value so that if someone stored pandas data frame they can get back a different thing without first reading it as pandas.

python/lsst/daf/butler/_formatter.py
timj: The first option doesn't work. I read the code and it doesn't do anything with context lib. I will do the second option you suggested.

python/lsst/daf/butler/_formatter.py
timj: I have changed the code so that read_from_uri can be called later after downloading to the cache if `read_from_uri` raised not implemented error. What do you think of always downloading the file to local storage (as we do now) if there is no component and not parameters specified?

python/lsst/daf/butler/_formatter.py
timj: Isn't it notoriously hard to find out what type of local file system you have? I recall in the past having frustrations trying to work out in a generic way if a directory is on NFS or not.

This has been requested previously in [DM-36150](https://rubinobs.atlassian.net/issues/DM-36150).

python/lsst/daf/butler/_formatter.py
timj: No. I mean the write. The file can be written to the cache during `butler.put`. The "it" is the URI to the thing being read. It might not be remote even if originally it was a remote URI.

python/lsst/daf/butler/_formatter.py
timj: Added to `read_from_uri` as well. Not added to `read_from_local_file` on the basis that there isn't really much choice even if it's a large file and also if they really need to know the actual size they can ask easily enough and performance shouldn't be an issue. Maybe I should add it for symmetry though.

python/lsst/daf/butler/_formatter.py
timj: I tried that and then realized that for convenience in some formatters I do `local_uri.read()` and replacing them all with `with open` wrappers will be a bit annoying. I can change them if you prefer it.

python/lsst/daf/butler/_formatter.py
timj: Datastore is doing that catch and re-raise.

python/lsst/daf/butler/_formatter.py
timj: Yes. But no V1 formatters should have been needing to override the `__init__` so shouldn't notice.

python/lsst/daf/butler/_formatter.py
timj: This is because FormatterFactory is creating these instances and it's given `**kwargs` that work for V1 and V2 without it knowing which params are relevant (and that's because the MappingFactory creates the instance and can't know the distinction).

python/lsst/daf/butler/_formatter.py
timj: That `updateLocation` call has always been a hack that we needed for V1 formatter. The way the formatter instance is constructed (from the formatter factory) makes it hard to defer construction.

python/lsst/daf/butler/_formatter.py
timj: As written it will have the original storage class that the user passed in if the formatter says it can accept that, else it will have the registry storage class. Before I implemented `can_accept` it was always going to be the registry storage class. Having this available could make the file descriptor `readStorageClass` obsolete. Will need to think.

python/lsst/daf/butler/formatters/pickle.py
timj: I haven't tried to see if that works specifically for pickles. It definitely doesn't work for many other cases because `readline` is not implemented by `ResourcePath`.

python/lsst/daf/butler/_formatter.py
timj: I also added it to `read_from_local_file`...

python/lsst/daf/butler/_formatter.py
timj: Changing to `object` for the type annotation doesn't seem to make any difference to mypy so I can do that.

timj: Although we use `Any` for `in_memory_dataset` all over the code so are you saying we have to change all those to `object` as well?

python/lsst/daf/butler/_formatter.py
timj: I tried this and it does work although:

* End up with a lot of extra `if result is not NotImplemented` checks
* It's not an `object` so changing the return type from `Any` to `object` means you have to change all the `read_from_*` methods to be explicitly returning `types.NotImplementedType | object` (or we can make a type alias).

What do you think? returning but then complicating the return type? Or returning and `Any`? or keeping it with `raise`?

python/lsst/daf/butler/_formatter.py
timj: Turns out that for composite disassembly we pass in the component so it can't raise.

python/lsst/daf/butler/_formatter.py
timj: Modified to do the download and cache if there are no components or parameters specified.

python/lsst/daf/butler/_formatter.py
timj: Now return `NotImplemented` but switched back to `Any` in the type annotation.

python/lsst/daf/butler/_formatter.py
timj: That's too big a change for this ticket and something that should have a broader discussion first.

Disassembling during processing and then zipping when ingesting is an interesting idea and probably would work. Would be a bit slower for the "read the whole thing" case but would be faster for reading a small component (at least until the Exposure replacement arrives).

python/lsst/daf/butler/_formatter.py
timj: Did not do this because it was made more complicated once we returned `NotImplemented`.

python/lsst/daf/butler/formatters/pickle.py
timj: I checked and `pickle.load()` requires a `readline` implementation. I'm starting to think we are going to have to implement that in ResourcePath.

timj: `yaml.load` does seem to work with ResourcePath though it 's much worse performance because when you hand a stream to `yaml.load` it does reads in chunks of 4096 bytes. There is a 8MB yaml file in summit_utils. If I read that up front and then parse it it completes in 3 seconds but if I pass the handle to the loader it takes 24 seconds so reading up front is definitely better.

python/lsst/daf/butler/formatters/json.py
```diff
@@ -34,90 +34,45 @@
...
-
-        Returns
-        -------
-        inMemoryDataset : `object`
-            The requested data as a Python object or None if the string could
-            not be read.
-        """
         try:
-            data = json.loads(serializedDataset)
+            data = json.loads(json_bytes)
```
timj: The discussion on https://docs.pydantic.dev/latest/concepts/json/#json-parsing implies that 

1. Trying `pytype.model_validate_json()` here might be significantly faster.
2. We might want to consider using `jiter` for any other JSON if it's meant to be so much faster.
3. @TallJimbo might be right that a special PydanticJsonFormatter is better (although we still have to let JsonFormatter support Pydantic because of backwards compatibility).

@ktlim I assume that `jiter` is part of rubinenv 9 because of the pydantic dependency but maybe it should be explicit rather than explicit if I want to go with point 2 above.

timj: Oh. The new json parser is part of pydantic itself: https://docs.pydantic.dev/latest/api/pydantic_core/#pydantic_core.from_json

timj: Comparison reading 100MB JSON file:
```
In [8]: %timeit json.loads(data)
539 ms ± 6.83 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [9]: %timeit from_json(data)
266 ms ± 5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
So factor of two improvement but I don't know how many JSON files we are going to be writing that are around that size or larger. (YAML reading the same data is a disaster @ nearly 2 minutes).

python/lsst/daf/butler/_formatter.py
timj: I made this change so it uses a `str` for the local read and `ResourcePath` for the `read_from_uri`.

python/lsst/daf/butler/_formatter.py
timj: I added some extra documentation on the storage class overrides and how they appear to the formatter. I've verified that if you ask for an override it will always turn up in the readStorageClass.

python/lsst/daf/butler/_formatter.py
TallJimbo: Yeah, I think that's a good reason to not try to fix this via `as_local`; sounds like it'd require inventing a new `file`-like URL schema, which doesn't sound great.  Anyway, I'm happy to leave this for the other ticket.

python/lsst/daf/butler/_formatter.py
TallJimbo: I think that reinforces my thinking that we should move away from having a JSON formatter and a single YAML formatter that try to do everything, and specialize for Pydantic/dataclasses,etc.  But not on this ticket.

TallJimbo (approved):

PR #1020, opened 2024-05-24
DM-43697: Fix server-side cursor lifetime issues

andy-slac:
A couple of other small improvements:
- better protection against exception in DirectQueryDriver context exit
- NameShrinker is now a Database property used by all other clients that need to shrink something.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_query_driver/_result_page_converter.py
+++ b/python/lsst/daf/butler/direct_query_driver/_result_page_converter.py
@@ -47,6 +47,7 @@ from ..queries.driver import (
 from ..queries.result_specs import DataCoordinateResultSpec, DatasetRefResultSpec, DimensionRecordResultSpec
 
 if TYPE_CHECKING:
+    from ..name_shrinker import NameShrinker
     from ..registry.interfaces import Database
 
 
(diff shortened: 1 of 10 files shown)


python/lsst/daf/butler/direct_query_driver/_result_page_converter.py
```diff
@@ -148,20 +149,29 @@ class DatasetRefResultPageConverter(ResultPageConverter):  # numpydoc ignore=PR0
...
 
     def convert(
         self,
         raw_rows: Iterable[sqlalchemy.Row],
         next_key: PageKey | None,
     ) -> DatasetRefResultPage:
-        run_column = qt.ColumnSet.get_qualified_name(self._spec.dataset_type_name, "run")
-        dataset_id_column = qt.ColumnSet.get_qualified_name(self._spec.dataset_type_name, "dataset_id")
+        run_column = self._name_shrinker.shrink(
+            qt.ColumnSet.get_qualified_name(self._spec.dataset_type_name, "run")
```
andy-slac, line 170: `get_qualified_name` is also used by `ColumnSet.get_column_spec()` but we don't shrink anything there. Is this a potential problem? Should `get_qualified_name` instead return a shrunken name?

TallJimbo: I don't think so, because `get_column_spec` is intended to also be used to create a schema representation that might get converted to Arrow instead of SQL, and we don't want the shrinking to happen to the Arrow schema (I think; I haven't actually looked to see if there are length limits there).

TallJimbo (approved):
Looks great!  Wish I'd thought of some of these solutions back when I wrote the original code you're fixing here.

PR #1021, opened 2024-05-28
DM-44501: Implement expanded() for RemoteButler query results

dhirving:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_query_driver/_result_page_converter.py
+++ b/python/lsst/daf/butler/direct_query_driver/_result_page_converter.py
@@ -44,6 +44,7 @@ from ..dimensions import (
     DimensionRecordSet,
     SkyPixDimension,
 )
+from ..dimensions.record_cache import DimensionRecordCache
 from ..queries import tree as qt
 from ..queries.driver import (
     DataCoordinateResultPage,
@@ -88,6 +89,7 @@ class ResultPageConverterContext:
 
     db: Database
     column_order: qt.ColumnOrder
+    dimension_record_cache: DimensionRecordCache
 
 
 class DimensionRecordResultPageConverter(ResultPageConverter):  # numpydoc ignore=PR01
@@ -95,7 +97,7 @@ class DimensionRecordResultPageConverter(ResultPageConverter):  # numpydoc ignor
 
     def __init__(self, spec: DimensionRecordResultSpec, ctx: ResultPageConverterContext) -> None:
         self._result_spec = spec
-        self._converter = _create_dimension_record_row_converter(spec.element, ctx)
+        self._converter = _create_dimension_record_row_converter(spec.element, ctx, use_cache=False)
 
     def convert(
         self, raw_rows: Iterable[sqlalchemy.Row], next_key: PageKey | None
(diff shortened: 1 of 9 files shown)


python/lsst/daf/butler/direct_query_driver/_result_page_converter.py
```diff
@@ -74,63 +83,137 @@ def convert(self, raw_rows: Iterable[sqlalchemy.Row], next_key: PageKey | None)
...
+) -> _DimensionRecordRowConverter:
+    if use_cache and element.is_cached:
+        return _CachedDimensionRecordRowConverter(element, ctx.dimension_record_cache)
+    elif isinstance(element, SkyPixDimension):
+        return _SkypixDimensionRecordRowConverter(element)
+    else:
+        return _NormalDimensionRecordRowConverter(element, ctx.db)
+
+
+class _DimensionRecordRowConverter:
```
TallJimbo, line 122:
```suggestion
class _DimensionRecordRowConverter(ABC):
```
I think this is needed in order for `@abstractmethod` to do anything.

python/lsst/daf/butler/direct_query_driver/_result_page_converter.py
```diff
@@ -178,6 +182,24 @@ def convert(self, row: sqlalchemy.Row) -> DimensionRecord:
...
+    """
+
+    def __init__(self, element: DimensionElement, cache: DimensionRecordCache) -> None:
+        self._element = element
+        self._cache = cache
+        self._key_columns = [qt.ColumnSet.get_qualified_name(name, None) for name in element.required.names]
+
+    def convert(self, row: sqlalchemy.Row) -> DimensionRecord:
+        mapping = row._mapping
+        values = tuple(mapping[key] for key in self._key_columns)
```
TallJimbo, line 199: I could see this line being enough of a hotspot in returning query results that it may be worth timing some alternatives; in particular, I've always wondered whether wrapping the generator expression in a list comprehension before wrapping it in `tuple`  is faster or slower.

python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -3929,3 +3949,76 @@ def test_dataset_followup_spatial_joins(self) -> None:
...
+        datasets = list(
+            registry.queryDatasets(
+                "flat", collections="imported_g", where="instrument = 'Cam1' and detector <= 3"
+            ).expanded()
+        )
+        datasets.sort(key=lambda ref: ref.dataId["detector"])
+        self.assertEqual(len(datasets), 2)
+        self.assertEqual(datasets[0].id, uuid.UUID("60c8a65c-7290-4c38-b1de-e3b1cdcf872d"))
+        self.assertEqual(datasets[1].id, uuid.UUID("84239e7f-c41f-46d5-97b9-a27976b98ceb"))
+        # All of the dimensions for flat are "cached "dimensions.
```
TallJimbo:
```suggestion
        # All of the dimensions for flat are "cached" dimensions.
```

TallJimbo (approved):
Looks great; only line comments are tiny nits.

One big-picture but very slight concern: with the new test coverage you're using the `Registry` interface, we'll have some test-writing work to do when we actually get around to retiring that.  And while there are good reasons for some of those tests to use the `Registry` interface (to test those shims, for example), for some things I think testing the `Butler._query` interface would have been fine and maybe more future-proof.  But that's more something to think about for the future than something to change on this PR, since the new tests are already in place here.

dhirving:
> test-writing work to do when we actually get around to retiring that.

Yeah I've had that in the back of my mind for the last few tickets, but I wanted to be able to run the tests against the old system to make sure that the tests I was writing were actually the desired behavior (as well as covering the shims.)

There are a lot of useful pre-existing tests we would have wanted to port already anyway.  The good news is I think porting the tests should be pretty mechanical, just adding a lot of periods and parentheses.

python/lsst/daf/butler/direct_query_driver/_result_page_converter.py
dhirving: Actually mypy will statically check it without the metaclass

https://mypy.readthedocs.io/en/stable/class_basics.html#abstract-base-classes-and-multiple-inheritance

PR #1023, opened 2024-06-04
DM-44502: Stream query results from Butler server

dhirving:
Modified Butler server query responses so that they write the response incrementally, one page at a time, as newline-separated chunks of JSON.  This allows clients to start reading results earlier and prevents the server from exhausting all its memory buffering rows from large queries.

To simplify the implementation of this, reworked the structure of row fetching in the QueryDriver interface.  It previously had a `fetch_next_page` method that required passing a UUID for each page of data to be read.  For both QueryDriver implementations there has never been a natural source of those UUID page keys and it was just getting in the way.  So  instead we now return an iterator from QueryDriver.execute().

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_query_driver/_driver.py
+++ b/python/lsst/daf/butler/direct_query_driver/_driver.py
@@ -34,7 +34,7 @@ __all__ = ("DirectQueryDriver",)
 import dataclasses
 import logging
 import sys
-from collections.abc import Iterable, Mapping, Set
+from collections.abc import Iterable, Iterator, Mapping, Set
 from contextlib import ExitStack
 from typing import TYPE_CHECKING, Any, cast, overload
 
@@ -51,7 +51,6 @@ from ..queries.driver import (
     DatasetRefResultPage,
     DimensionRecordResultPage,
     GeneralResultPage,
-    PageKey,
     QueryDriver,
     ResultPage,
 )
@@ -147,7 +146,7 @@ class DirectQueryDriver(QueryDriver):
         self._raw_page_size = raw_page_size
         self._postprocessing_filter_factor = postprocessing_filter_factor
         self._constant_rows_limit = constant_rows_limit
-        self._cursors: dict[PageKey, _Cursor] = {}
+        self._cursors: set[_Cursor] = set()
 
     def __enter__(self) -> None:
         self._exit_stack = ExitStack()
@@ -170,10 +169,10 @@ class DirectQueryDriver(QueryDriver):
         self._upload_tables.clear()
         # Transfer open cursors' close methods to exit stack, this will help
         # with the cleanup in case a cursor raises an exceptions on close.
-        for cursor in self._cursors.values():
+        for cursor in self._cursors:
             self._exit_stack.push(cursor.close)
         self._exit_stack.__exit__(exc_type, exc_value, traceback)
-        self._cursors = {}
+        self._cursors = set()
         self._exit_stack = None
 
     @property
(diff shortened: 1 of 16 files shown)


python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -233,12 +234,25 @@ def execute(self, result_spec: ResultSpec, tree: qt.QueryTree) -> ResultPage:
...
+        return self._read_results(cursor)
+
+    def _read_results(self, cursor: _Cursor) -> Iterator[ResultPage]:
+        """Read out all of the result pages from the database."""
+        try:
+            while (result_page := cursor.next()) is not None:
+                yield result_page
+        finally:
+            cursor.close()
+            self._cursors.discard(cursor)
```
andy-slac: Could `cursor.close()` raise an exception? Maybe re-order these two lines to discard first?

python/lsst/daf/butler/remote_butler/server/handlers/_external_query.py
```diff
@@ -29,40 +29,85 @@
...
+            # Transfer control of the context manager to
+            # _stream_query_pages.
+            exit_stack.pop_all(),
+            spec,
+            response_pages,
+        )
+        return StreamingResponse(output_generator, media_type="application/jsonlines")
+
+    # Mypy thinks that ExitStack might swallow an exception.
+    assert False, "This line is unreachable."
```
andy-slac, line 88: I presume there is no other more explicit way to suppress this with `# type: ignore[something]`?

python/lsst/daf/butler/remote_butler/server/handlers/_external_query.py
```diff
@@ -29,40 +29,85 @@
...
+    #
+    # If it was sync, it would get wrapped in an async function internal to
+    # FastAPI that does not guarantee that the generator is fully iterated or
+    # closed.
+    # (There is an example in the FastAPI docs showing StreamingResponse with a
+    # sync generator with a context manager, but after reading the FastAPI
+    # source code I believe that for sync generators it will leak the context
+    # manager if the client disconnects, and that it would be
+    # difficult/impossible for them to fix this in the general case within
+    # FastAPI.)
```
andy-slac, line 107: Hmm, it bothers me a bit that you need to read sources to understand their behavior. 🙁

python/lsst/daf/butler/remote_butler/server/handlers/_external_query.py
```diff
@@ -29,40 +29,85 @@
...
+        # reading results earlier and prevents the server from exhausting
+        # all its memory buffering rows from large queries.
+        output_generator = _stream_query_pages(
+            # Transfer control of the context manager to
+            # _stream_query_pages.
+            exit_stack.pop_all(),
+            spec,
+            response_pages,
+        )
+        return StreamingResponse(output_generator, media_type="application/jsonlines")
```
andy-slac, line 85: The context management logic in this method looks rather generic. If we expect more server methods that need similar approach it maybe worth moving this logic to a separate utility class/method?

andy-slac (approved):
Looks great, a couple of minor questions. I'm not super familiar with FastAPI/httpx, but comments do explain why it has to be done that way. 🙂

python/lsst/daf/butler/remote_butler/server/handlers/_external_query.py
dhirving: The FastAPI documentation is for the most part pretty good, but they definitely assume async handlers by default so some of the subtle details for interacting with non-async code are a bit more ambiguous.  `StreamingResponse` is a less-frequently-used part of the API too.

BTW the most useful parts of their documentation are the `Tutorial - User Guide` and `Advanced User Guide`, both found on this page: https://fastapi.tiangolo.com/tutorial/

python/lsst/daf/butler/remote_butler/server/handlers/_external_query.py
dhirving: I'm hoping we don't have anything else that will need StreamingResponse... the queries are really the only thing that slowly return an extremely large result set, and REST APIs really work better when they can promptly return a modest amount of data.

I suspect this approach will not survive into the final version even for the queries.  There are some issues like:
* Timeouts we do not control at the Kubernetes ingress... e.g. qserv has to kill sync queries after 30 minutes because the connection gets dropped otherwise
* Letting the client force us to hold a DB cursor open for an extended period of time could cause resource problems in the DB
* etc

python/lsst/daf/butler/remote_butler/server/handlers/_external_query.py
dhirving: Not that I'm aware of, the issue is pretty basic to how mypy analyzes control flow.

Most context managers are declared with a return type of `Literal[False]` so that mypy knows they won't swallow an exception, but because ExitStack could contain any context manager it returns a `bool` and mypy has to assume that we might bypass the return statement inside the context manager.

PR #1024, opened 2024-06-04
DM-44311: Fix collection summary updates on associate

andy-slac:
Datasets manager always updated collection summaries even if datasets
list was empty which resulted in many false positives in summaries.
One-line fix to skip empty datasets lists. Other methods get similar
protection for the case of empty input set of datasets.

Add registry method to cleanup collection summaries.
For now only the dataset type summaries are refreshed, governor
summaries cleanup may be added later.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/tests/_registry.py
+++ b/python/lsst/daf/butler/registry/tests/_registry.py
@@ -43,7 +43,7 @@ from collections.abc import Callable, Iterable, Iterator
 from concurrent.futures import ThreadPoolExecutor
 from datetime import timedelta
 from threading import Barrier
-from typing import TypeVar
+from typing import TypeVar, overload
 
 import astropy.time
 import sqlalchemy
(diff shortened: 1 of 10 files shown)


python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -4022,3 +4026,60 @@ def test_expanded_data_id_queries(self) -> None:
...
+        # Summaries should not have changed.
+        summary = registry.getCollectionSummary(tagged_coll)
+        self.assertEqual(summary.dataset_types.names, {"bias", "flat"})
+        self.assertEqual(summary.governors, {"instrument": {"Cam1"}})
+
+        # Cleanup summaries.
+        registry.refresh_collection_summaries()
+        summary = registry.getCollectionSummary(tagged_coll)
+        self.assertFalse(summary.dataset_types.names)
+        # We do not clean governor summaries yet, but beacuse how the query is
```
TallJimbo:
```suggestion
        # We do not clean governor summaries yet, but because how the query is
```

TallJimbo (approved):

PR #1026, opened 2024-06-15
DM-44842: Fix query crashes on postgres

dhirving:
Fixed two issues where invalid SQL statements were being generated for Postgres by the new query system.  (See individual commit messages for details.)  

We now also run the Butler server tests using Postgres for the registry, since this is the configuration being used in production.  In the process, deduplicated the logic for setting up a postgres instance for testing -- it had been copy/pasted five places.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_query_driver/_driver.py
+++ b/python/lsst/daf/butler/direct_query_driver/_driver.py
@@ -539,16 +539,6 @@ class DirectQueryDriver(QueryDriver):
         for dataset_type, fields_for_dataset in projection_plan.columns.dataset_fields.items():
             if len(joins_plan.datasets[dataset_type].collection_records) > 1:
                 fields_for_dataset.add("collection_key")
-        if projection_plan:
-            # If there's a projection and we're doing postprocessing, we might
-            # be collapsing the dimensions of the postprocessing regions.  When
-            # that happens, we want to apply an aggregate function to them that
-            # computes the union of the regions that are grouped together.
-            for element in builder.postprocessing.iter_missing(projection_plan.columns):
-                if element.name in projection_plan.columns.dimensions.elements:
-                    projection_plan.region_non_aggregates.append(element)
-                else:
-                    projection_plan.region_aggregates.append(element)
 
         # The joins-stage query also needs to include all columns needed by the
         # downstream projection query.  Note that this:
@@ -663,6 +653,13 @@ class DirectQueryDriver(QueryDriver):
         unique_keys: list[sqlalchemy.ColumnElement[Any]] = [
             builder.joiner.dimension_keys[k][0] for k in plan.columns.dimensions.data_coordinate_keys
         ]
+
+        # Many of our fields derive their uniqueness from the unique_key
+        # fields: if rows are uniqe over the 'unique_key' fields, then they're
+        # automatically unique over these 'derived_fields'.  We just remember
+        # these as pairs of (logical_table, field) for now.
+        derived_fields: list[tuple[str, str]] = []
+
         # There are two reasons we might need an aggregate function:
         # - to make sure temporal constraints and joins have resulted in at
         #   most one validity range match for each data ID and collection,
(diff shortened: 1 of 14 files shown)


python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -686,6 +688,11 @@ def apply_query_projection(self, plan: QueryProjectionPlan, builder: QueryBuilde
         # automatically unique over these 'derived_fields'.  We just remember
         # these as pairs of (logical_table, field) for now.
         derived_fields: list[tuple[str, str]] = []
+        # The region associated with dimension keys returned by the query are
+        # derived fields, since there is only one region associated with each
+        # dimension key value.
+        for element in plan.region_non_aggregates:
+            derived_fields.append((element.name, "region"))
```
TallJimbo: I think I originally expected these to be added by the loop below, since regions are a kind of dimension field.  The bug was that that was only happening for regions included in the output rows, not regions needed only by `Postprocessing`, since we don't put the latter in `plan.columns`.  I think that means we should also guard against a region appearing in both places.

To do that might make sense to drop the `region_aggregate` and `region_non_aggregate` attributes from `QueryProjectionPlan`, and instead handle them both together here in one loop:
```
for element in builder.postprocessing.iter_missing(plan.columns):
    if element.name not in plan.columns.dimensions.elements:
        builder.joiner.fields[element.name]["region"] = ddl.Base64Region.union_aggregate(
            builder.joiner.fields[element.name]["region"]
        )
        have_aggregates = True
    else:
        derived_fields.append((element.name, "region"))
```
It doesn't look like we need to split out the aggregate regions from the non-aggregate regions in the plan data structure anywhere else.

python/lsst/daf/butler/tests/butler_queries.py
```diff
@@ -467,6 +467,21 @@ def test_common_skypix_overlaps(self) -> None:
...
+        """
+        butler = self.make_butler("hsc-rc2-subset.yaml")
+        with butler._query() as query:
+            self.assertEqual(
+                [(9813, 72)],
+                [
+                    (data_id["tract"], data_id["patch"])
+                    for data_id in query.data_ids(["patch"]).where({"instrument": "HSC", "visit": 318})
+                ],
+            )
```
TallJimbo, line 486: A version of this test that returned patch dimension records instead of data coordinates would cover the case of the region appear in both the result rows and the `Postprocessing` that I was worried about in my other comment, but I think the original query system tests I wrote should have covered that already. I'm guessing that means the duplication I was worried about either gets fixed already in another layer or it manifests as something harmless, like a column nothing uses in the SELECT clause.

tests/test_postgresql.py
```diff
@@ -27,75 +27,45 @@
...
-
-    @classmethod
-    def tearDownClass(cls):
-        # Clean up any lingering SQLAlchemy engines/connections
-        # so they're closed before we shut down the server.
-        gc.collect()
-        cls.server.stop()
-        removeTestTempDir(cls.root)
+        super().setUpClass()
+        cls.postgres = cls.enterClassContext(setup_postgres_test_db())
```
TallJimbo, line 59: I was getting all ready to be concerned about using a context manager for this utility code since I knew it was going to be needed in a `setUpClass`, but I didn't know about `enterClassContext`.  Nice!

TallJimbo (approved):

python/lsst/daf/butler/tests/butler_queries.py
dhirving: Added the test and it appears fine.

Unless I'm missing something there shouldn't be any duplication, though -- isn't the point of `iter_missing` that it doesn't include columns that already appear in the result rows?

TallJimbo: Right, with the `iter_missing` version there shouldn't be any duplication.  I think with the previous version of this PR there would have been duplicate entries in `derived_fields`, but I don't know if that would have caused any real problems.

PR #1027, opened 2024-06-24
DM-44503: Clean up Butler server REST API

dhirving:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/dimensions/_config.py
+++ b/python/lsst/daf/butler/dimensions/_config.py
@@ -161,7 +161,7 @@ class DimensionConfig(ConfigSubset):
 
         Returns
         -------
-        model : `SerializedUniverseConfig`
+        model : `SerializedDimensionConfig`
             Serializable Pydantic version of this configuration.
         """
         return SerializedDimensionConfig.model_validate(self.toDict())
--- a/python/lsst/daf/butler/dimensions/_config.py
+++ b/python/lsst/daf/butler/dimensions/_config.py
@@ -27,7 +27,7 @@
 
 from __future__ import annotations
 
-__all__ = ("DimensionConfig",)
+__all__ = ("DimensionConfig", "SerializedDimensionConfig")
 
 import textwrap
 from collections.abc import Mapping, Sequence, Set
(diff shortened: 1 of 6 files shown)


python/lsst/daf/butler/dimensions/_config.py
```diff
@@ -156,6 +156,32 @@ def _updateWithOtherConfigFile(self, file: Config | str | ResourcePath | Mapping
...
+
+        Returns
+        -------
+        model : `SerializedUniverseConfig`
+            Serializable Pydantic version of this configuration.
+        """
+        return SerializedDimensionConfig.model_validate(self.toDict())
+
+    @staticmethod
+    def from_simple(simple: SerializedDimensionConfig) -> DimensionConfig:
```
andy-slac, line 170: I think in other places we prefer to use `@classmethod` for this so you can do `cls(...)` to support potential subclassing. Not sure it matters here though.

python/lsst/daf/butler/dimensions/_config.py
```diff
@@ -156,6 +156,32 @@ def _updateWithOtherConfigFile(self, file: Config | str | ResourcePath | Mapping
         externalConfig = type(self)(file, validate=False)
         self.update(externalConfig)
 
+    def to_simple(self) -> SerializedDimensionConfig:
+        """Convert this configuration to a serializable Pydantic model.
+
+        Returns
+        -------
+        model : `SerializedUniverseConfig`
```
andy-slac:
```suggestion
        model : `SerializedDimensionConfig`
```

andy-slac (approved):
Looks great, a couple of minor comments.

python/lsst/daf/butler/remote_butler/server/handlers/_external.py
```diff
@@ -83,76 +85,60 @@ async def get_client_config() -> dict[str, Any]:
...
+) -> GetUniverseResponseModel:
     # Allow remote client to get dimensions definition.
     butler = factory.create_butler()
-    return butler.dimensions.dimensionConfig.toDict()
+    return GetUniverseResponseModel(universe=butler.dimensions.dimensionConfig.to_simple())
 
 
 @external_router.get(
-    "/v1/dataset_type/{dataset_type_name}",
+    "/v1/dataset_type",
```
timj: If we know that we are enforcing that a name has no slashes it would be really nice from an API perspective if we did allow this simple form for retrieving the dataset type information. It seems like a very natural interface for single dataset type with no wildcards returning the details.

python/lsst/daf/butler/dimensions/_config.py
dhirving: Generally speaking I'm not a big fan of that pattern, since it assumes subclasses have the same constructor signature as the base class.

And yeah specifically here no subclasses exist or should exist (and if they did, they would have a different serialized form with their own data so the function signatures wouldn't match up.)

PR #1029, opened 2024-06-25
DM-44875: Handle ambiguous calibration lookups on older postgres

dhirving:
For Postgres older than version 16, we now throw an error for find-first calibration searches instead of silently returning a potentially-incorrect result.   This is similar to the behavior of the old query system.

The query requires the ANY_VALUE aggregate function to generate the validity match count column used by postprocessing to check for ambiguous results.  This function is not available on older Postgres.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_query_driver/_driver.py
+++ b/python/lsst/daf/butler/direct_query_driver/_driver.py
@@ -459,7 +459,7 @@ class DirectQueryDriver(QueryDriver):
         # construction do.
         plan, builder = self.analyze_query(tree, final_columns, order_by, find_first_dataset)
         self.apply_query_joins(plan.joins, builder.joiner)
-        self.apply_query_projection(plan.projection, builder)
+        self.apply_query_projection(plan.projection, builder, order_by)
         builder = self.apply_query_find_first(plan.find_first, builder)
         builder.columns = plan.final_columns
         return plan, builder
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -715,11 +717,11 @@ def apply_query_projection(
                         unique_keys.append(builder.joiner.fields[dataset_type]["collection_key"])
                     else:
                         derived_fields.append((dataset_type, "collection_key"))
-                elif dataset_field == "timespan" and plan.datasets[dataset_type].is_calibration_search:
+                elif dataset_field == "timespan" and is_calibration_search:
```
dhirving, line 720: This `dataset_field == "timespan"` case was already not covered before I got here.  I wanted to add a test but can't figure out what circumstances would trigger it, since the validity range comparison is handled in-DB, and DatasetRefs don't include the calibration collection timespan field.

TallJimbo: Right, this can't be tested until we add support for "general" results.

python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -715,11 +717,11 @@ def apply_query_projection(
...
                     # If we're doing a non-find-first query against a
                     # CALIBRATION collection, the timespan is also a unique
                     # key...
-                    if dataset_type == plan.find_first_dataset:
+                    if is_find_first_search:
```
dhirving, line 724: also I realized last night that we need to add dataset_id to unique keys for non-find-first search, I'll do that today before you review this for real hopefully

PR #1031, opened 2024-06-27
DM-45045: Preserve the order of the keys in YAML

tgoldina:
Currently, YamlFormatter always sorts keys in alphabetical order. The reason is that we do not pass any keyword parameters to yaml's `dump` or `safe_dump`, and by default `sort_keys` is True. Added `sort_keys=False` to preserve the order of the keys in the Pydantic model.

## Checklist

- [ ] ran Jenkins


timj:
@tgoldina let me know if you want me to start a Jenkins run.

timj (approved):

PR #1032, opened 2024-06-30
DM-44362: Fix unit test for spatial overlaps

andy-slac:
Using non-common skypix is going to be deprecated in favor of a
direct use of regions. New query system raises NotImplementedError
for non-common skypix, this patch replaces that exception with
a new exception class to support exception propagation from remote
butler to client.

`test_skypix_constraint_queries` unit test is updated to handle
exceptions from the new query system. When direct butler switches
to the new query system, that test can be removed or changed to
always expect the exceptions.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_exceptions.py
+++ b/python/lsst/daf/butler/_exceptions.py
@@ -155,6 +155,14 @@ class MissingCollectionError(CollectionError, ButlerUserError):
     error_type = "missing_collection"
 
 
+class UnimplementedQueryError(NotImplementedError, ButlerUserError):
+    """Exception raised when the query system does not support the query
+    specified by the user.
+    """
+
+    error_type = "unimplemented_query"
+
+
 class MissingDatasetTypeError(DatasetTypeError, KeyError, ButlerUserError):
     """Exception raised when a dataset type does not exist."""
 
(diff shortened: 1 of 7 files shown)


andy-slac:
@TallJimbo, I tried few different ideas and ended up with a special visitor that rewrites predicates which use non-common skypix. A complicated thing is that it also needs to override dimension group in the tree to replace skypix there too. It fixes that particular query that was broken in the unit test with `htm11` constraint, but that unit test is still failing on a subsequent query. Before I try to fix the next thing, could you check that this new visitor makes sense or maybe I need something different?

TallJimbo:
We talked about this at the Tuesday standup (while you were on vacation), and given the new complexity that seems to be required, I think we should really consider deprecating this functionality instead, since I think the primary reason it's important in the old query system is as a workaround for general geometry overlap queries, and we can do that better in the new query system if we do it more directly.

andy-slac:
I'm ok with deprecation, but does that mean that we still need to support until next release? Or is it fine if it's broken for remote butler only?

TallJimbo:
I think it's fine for it to be broken for RemoteButler only for now; the deprecation period will constrain the timeline for having the DirectButler's old query methods delegate to the new system, but there are other differences we plan to deprecate in that category already.  Those are being tracked on DM-43587, and I've added this issue there as well.

TallJimbo (approved):

PR #1034, opened 2024-07-08
DM-45140: Skip datastore root check for non-POSIX filesystems

dhirving:
We now skip existence checks for the datastore root for non-POSIX root paths.  This fixes an issue where FileDatastore was throwing a permissions-related exception on startup when trying to use a read-only S3 bucket via either `s3://` or `https://`.

This was occurring because on S3 there is no concept of a directory, so even though files exist in the datastore, the root directory does not.

## Checklist

- [x] ran Jenkins


timj (approved):

gpdf:
Thank you!

PR #1036, opened 2024-07-10
DM-41326: remove DimensionGraph and related interfaces

TallJimbo:
## Checklist

- [x] ran Jenkins


doc/changes/DM-41326.removal.md
```diff
@@ -0,0 +1,3 @@
+Remove `DimensionGraph` and the `Mapping` interface to `DataCoordinate`, along with most other public interfaces that utilize `DimensionElement` instances instead of just their string names.
+
+See RFC-834 for full details and rationale.
```
timj:
```suggestion
See [RFC-834](https://rubinobs.atlassian.net/browse/RFC-834) for full details and rationale.
```

python/lsst/daf/butler/_dataset_type.py
```diff
@@ -748,8 +742,6 @@ def from_simple(
         match simple.dimensions:
```
timj: Seems like `match` is now overkill if this is now doing:
```python
if simple.dimensions is None:
    raise(...)
dimensions = universe.conform(simple.dimensions)
```

python/lsst/daf/butler/_dataset_type.py
```diff
@@ -109,8 +109,6 @@ def direct(
         match dimensions:
```
timj:
```python
serialized_dimensions = dimensions if dimensions else None
```
? Or are you worried that the type really might not be correct?

I think the type annotation for this method needs to remove the `dict` now.

python/lsst/daf/butler/_dataset_type.py
```diff
@@ -59,7 +59,7 @@ class SerializedDatasetType(BaseModel):
 
     name: StrictStr
     storageClass: StrictStr | None = None
-    dimensions: SerializedDimensionGraph | list[StrictStr] | None = None
+    dimensions: list[StrictStr] | None = None
```
timj, line 62: I see that this leads to any SerializedDatasetType JSON we have lying around being incompatible if it had the graph in it. People shouldn't have that anywhere (not even in older graphs?) but it probably means that we need to ensure the butler server deployments are updated so they don't send back the wrong form of SerializedDatasetType.

timj (approved):
Looks great. Impressive amount of cleanup work.

timj:
I see there is a failure:
```
FAILED tests/test_packages.py::PackagesFormatterTestCase::testPackages - TypeError: expected string or bytes-like object, got 'NoneType'
```
which is great because Fabio reported that failure this morning when building the new weekly. This means there is a chance it can be debugged more easily.

python/lsst/daf/butler/_dataset_type.py
TallJimbo: That's right, but we've been writing this with `list` for a while now, so butler server should be fine, and I figured any graphs old enough to suffer from this would probably be impossible to read already due to other incompatibilities (though I don't think I can prove that in general).

PR #1038, opened 2024-07-13
DM-45237: Add ability to disable file validation info on ingest-files and import

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-45237.misc.rst
+++ b/doc/changes/DM-45237.misc.rst
@@ -1,0 +1,5 @@
+File ingest no longer checks that every file exists.
+This can take a very long time if thousands of files are being ingested from an object store.
+Now at most 200 files will be checked.
+Whether all files are subsequently checked depends on the transfer mode and whether ``--no-track-file-attrs`` is enabled.
+For ``direct`` or in-place ingest coupled with ``--no-track-file-attrs`` the file existence might never be verified.
--- a/python/lsst/daf/butler/cli/utils.py
+++ b/python/lsst/daf/butler/cli/utils.py
@@ -352,10 +352,10 @@ def split_kv(
     separator: str = "=",
     unseparated_okay: bool = False,
     return_type: type[dict] | type[tuple] = dict,
-    default_key: str = "",
+    default_key: str | None = "",
     reverse_kv: bool = False,
     add_to_default: bool = False,
-) -> dict[str, str] | tuple[tuple[str, str], ...]:
+) -> dict[str | None, str] | tuple[tuple[str | None, str], ...]:
     """Process a tuple of values that are key-value pairs separated by a given
     separator. Multiple pairs may be comma separated. Return a dictionary of
     all the passed-in values.
@@ -401,8 +401,9 @@ def split_kv(
         tuple will be 2-item tuple, the first item will be the value to the
         left of the separator and the second item will be the value to the
         right. By default `dict`.
-    default_key : `Any`
+    default_key : `str` or `None`
         The key to use if a value is passed that is not a key-value pair.
+        `None` can imply no separator depending on how the results are handled.
         (Passing values that are not key-value pairs requires
         ``unseparated_okay`` to be `True`).
     reverse_kv : bool
@@ -453,26 +454,26 @@ def split_kv(
 
     class RetDict:
         def __init__(self) -> None:
-            self.ret: dict[str, str] = {}
+            self.ret: dict[str | None, str] = {}
 
-        def add(self, key: str, val: str) -> None:
+        def add(self, key: str | None, val: str) -> None:
             if reverse_kv:
                 key, val = val, key
             self.ret[key] = val
 
-        def get(self) -> dict[str, str]:
+        def get(self) -> dict[str | None, str]:
             return self.ret
 
     class RetTuple:
         def __init__(self) -> None:
-            self.ret: list[tuple[str, str]] = []
+            self.ret: list[tuple[str | None, str]] = []
 
-        def add(self, key: str, val: str) -> None:
+        def add(self, key: str | None, val: str) -> None:
             if reverse_kv:
                 key, val = val, key
             self.ret.append((key, val))
 
-        def get(self) -> tuple[tuple[str, str], ...]:
+        def get(self) -> tuple[tuple[str | None, str], ...]:
             return tuple(self.ret)
 
     if separator in (",", " "):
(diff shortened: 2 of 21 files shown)


python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -963,12 +961,6 @@ def _standardizeIngestPath(
         if not srcUri.isabs():
             srcUri = self.root.join(path)
 
-        if not srcUri.exists():
-            raise FileNotFoundError(
-                f"Resource at {srcUri} does not exist; note that paths to ingest "
-                f"are assumed to be relative to {self.root} unless they are absolute."
```
timj, line 969: I can't see anyway around removing this. Maybe I should do the check if the ingest is direct or None and then only if local file? There is no reason to do it for a copy since the copy will fail later on anyhow.

ktlim: This doesn't seem like a primary function of this method, given its name and description.  I'm fine with a check (if any) happening elsewhere.

ktlim (approved):

python/lsst/daf/butler/datastores/fileDatastore.py
timj: It sounds like any check is going to be prohibitively slow. Options seem to be:

* Never check and hope the file really is there when you ingest with direct / no-track-file-attrs
* Always check if local else never check
* Only check if there are fewer than 200 files being ingested
* Always check the first 10 (for example) regardless and if those do exist trust that the person has a reasonable set of data that is likely to really exist.

I think I'm talking myself into the last option (maybe above this method). That would catch the egregious case where someone is trying to ingest without really having access to the remote bucket or they copied the export file to their laptop and are assuming the files are still accessible. Even only checking the first might be enough.

cc/ @TallJimbo in case you have an opinion.

TallJimbo: This is all for `direct` and `None` transfer modes, and not the others, right?  Not some case where a genuine artifact transfer silently fails?  If that's the case then some sort of spot-checking like the last bullet sounds reasonable.  Maybe the first one and a random 1% or something thereafter?

timj: Right. If "copy" mode is being used then the ingest will break as soon as it tries to copy the file (maybe "symlink" will also fail?). If we are tracking file attributes the ingest will fail when it can't find the size.

So we only really need to do any spot checks if we are using direct/None with no-track-file-attrs.

timj: Ok. I've changed it to only check 200 of the files being ingested. That doesn't take very long and you can at least get a sense that things are probably okay.

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1122,7 +1127,24 @@ def _extractIngestInfo(
...
+        # anyhow when they are relocated. For direct or None transfer modes
+        # it is possible to not know if the file is accessible at all.
+        # Therefore limit number of files that will be checked (but always
+        # include the first one).
+        max_checks = 200
+        n_datasets = len(datasets)
+        if n_datasets <= max_checks:
+            check_every_n = 1
+        elif transfer in ("direct", None):
+            check_every_n = int(n_datasets / max_checks + 1)  # +1 so that if n < 100 the answer is 1.
```
timj: @ktlim I've asked for a re-review since this change is quite significant. The code coverage doesn't cover the important bit because none of the tests try to ingest large numbers of files. I have verified that @gpdf 's example does only check 200 files out of 155k and that doesn't notable affect the import time.

python/lsst/daf/butler/cli/utils.py
```diff
@@ -352,10 +352,10 @@ def split_kv(
     separator: str = "=",
     unseparated_okay: bool = False,
     return_type: type[dict] | type[tuple] = dict,
-    default_key: str = "",
+    default_key: str | None = "",
```
timj, line 355: This `None` is here specifically to support `--log-levels` because the code in `cliLog` is using `None` to indicate the default logger and "" or "." to indicate the root logger. They are not the same because we want `--log-level INFO` to affect the `lsst` logger and not the root logger.

python/lsst/daf/butler/cli/utils.py
```diff
@@ -789,7 +790,7 @@ class MWOptionDecorator:
     """
 
     def __init__(self, *param_decls: Any, **kwargs: Any) -> None:
-        self.partialOpt = partial(click.option, *param_decls, cls=partial(MWOption), **kwargs)
+        self.partialOpt = partial(click.option, *param_decls, cls=partial(MWOption), **kwargs)  # type: ignore
```
timj, line 793: mypy doesn't think that `partial(MWOption)` is the same as an option and so now complains.

doc/changes/DM-45237.misc.rst
```diff
@@ -0,0 +1,5 @@
+File ingest no longer checks that ever file exists.
```
andy-slac:
```suggestion
File ingest no longer checks that every file exists.
```

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -1122,7 +1127,24 @@ def _extractIngestInfo(
...
         filtered = []
-        for dataset in datasets:
+
+        # Ingest could be given tens of thousands of files. It is not efficient
+        # to check for the existence of every single file (especially if they
+        # are remote URIs) but in some transfer modes the files will be checked
+        # anyhow when they are relocated. For direct or None transfer modes
+        # it is possible to not know if the file is accessible at all.
+        # Therefore limit number of files that will be checked (but always
+        # include the first one).
```
andy-slac, line 1137: Could there be a use case when clients want to check every file even if there are thousands of them? 200 looks like an arbitrary number, maybe we want a new option for that (or per-datastore config parameter)?

andy-slac (approved):
Looks great.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: Yes. It's an issue. It's why part of the conversation was considering always doing the checks for POSIX and only skipping for remote. Gregory's original 150k import example was so long (many hours) that it seemed like it wouldn't ever be something people wanted to do. I picked 200 because that's about the size of the focal plane... and it doesn't take very long.

PR #1040, opened 2024-07-18
DM-44843: Improve performance of collection resolution in the new query system

andy-slac:
This replaces Python-side recursive collection search, which caused
database query per collection with a call to collection manager
`resolve_wildcard` which is more efficient.

The code now uses recursive CTE in some cases to query collections
and their children. This results in reduced number of queries when
querying complex collection chains.

Small refactoring of collections manager, trying to simplify `resolve_wildcard`
method and possibly improve its performance.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/collections/_base.py
+++ b/python/lsst/daf/butler/registry/collections/_base.py
@@ -348,26 +348,24 @@ class DefaultCollectionManager(CollectionManager[K]):
         wildcard: CollectionWildcard,
         *,
         collection_types: Set[CollectionType] = CollectionType.all(),
-        done: set[str] | None = None,
         flatten_chains: bool = True,
         include_chains: bool | None = None,
     ) -> list[CollectionRecord[K]]:
         # Docstring inherited
-        if done is None:
-            done = set()
+        done: set[K] = set()
         include_chains = include_chains if include_chains is not None else not flatten_chains
 
         def resolve_nested(
-            record: CollectionRecord, done: set[str], collection_cache: CollectionRecordCache | None
+            record: CollectionRecord, done: set[K], collection_cache: CollectionRecordCache | None
         ) -> Iterator[CollectionRecord[K]]:
-            if record.name in done:
+            if record.key in done:
                 return
             if record.type in collection_types:
-                done.add(record.name)
+                done.add(record.key)
                 if record.type is not CollectionType.CHAINED or include_chains:
                     yield record
             if flatten_chains and record.type is CollectionType.CHAINED:
-                done.add(record.name)
+                done.add(record.key)
                 children = cast(ChainedCollectionRecord[K], record).children
                 for child in self._find_many(children, flatten_chains, collection_cache):
                     # flake8 can't tell that we only delete this closure when
--- a/python/lsst/daf/butler/registry/collections/_base.py
+++ b/python/lsst/daf/butler/registry/collections/_base.py
@@ -31,7 +31,7 @@ from ... import ddl
 __all__ = ()
 
 from abc import abstractmethod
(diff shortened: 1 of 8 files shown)


andy-slac:
mypy is failing due to the new version, I'll look at it later before merging.

python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -281,28 +284,112 @@ def find(self, name: str) -> CollectionRecord[K]:
...
+                raised then no records are generated.
+            """
+            record = cache.get_by_name(name)
+            if record is not None:
+                if flatten_chains and record.type is CollectionType.CHAINED:
+                    # Check all children recursively.
+                    for child_name in cast(ChainedCollectionRecord, record).children:
+                        # Make a list so that LookupError can happen before we
+                        # return anything in iterator.
+                        yield from list(check_cache(child_name, cache, flatten_chains))
```
TallJimbo: I suspect it might be more efficient to just have this return a `list` instead of being a generator, even if that means returning single-element lists sometimes and appending them.  That probably doesn't matter much, but if a `list` version is just as easy to read and it also sidesteps issues like the one the comment points out, it may be better overall.

python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -281,28 +284,112 @@ def find(self, name: str) -> CollectionRecord[K]:
...
-        if self._caching_context.collection_records is not None:
+        records: dict[str, CollectionRecord[K]] = {}
+        fetch_names = []
+        if collection_cache is not None:
             for name in names:
-                records[name] = self._caching_context.collection_records.get_by_name(name)
-            fetch_names = [name for name, record in records.items() if record is None]
+                try:
+                    # Make a list first so that we can avoid updating records
+                    # dict if any record is missing.
```
TallJimbo: Where is the `list` this comment refers to?

python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -322,44 +409,49 @@ def resolve_wildcard(
...
-            for record in self._fetch_all():
-                if any(p.fullmatch(record.name) for p in wildcard.patterns):
-                    result.extend(resolve_nested(record, done))
-        del resolve_nested
+            if all_records is None:
+                all_records = self._fetch_all(cache)
+            for record in filter_types(all_records):
+                if record.key not in done:
+                    if any(p.fullmatch(record.name) for p in wildcard.patterns):
+                        result.append(record)
```
TallJimbo: Seems like we should move the `if wildcard.patterns` block above the `if wildcard.strings` block, so when both are in play we just fell all collections and then use the cache, instead of first doing a query for certain collections and then fetching all of them.

TallJimbo (approved):

timj:
@andy-slac  I have mypy fixes pushed to #1038 (although whether you think they are correct is a different issue)

python/lsst/daf/butler/registry/collections/_base.py
andy-slac: Indeed, it should simplify things, I'll switch to `list`.

python/lsst/daf/butler/registry/collections/_base.py
andy-slac: It's a leftover from a previous experiment, I'll drop it.

python/lsst/daf/butler/registry/collections/_base.py
andy-slac: I kept this order just to be consistent with previous code and it also made sense to me that `strings` collections are returned first, but in the end the ordering of the result should not matter here, I'll re-order them.

PR #1041, opened 2024-07-23
DM-45386: Fix serialization of datetime64 to parquet via numpy and astropy.

erykoff:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_parquet.py
+++ b/tests/test_parquet.py
@@ -30,6 +30,7 @@
 Tests in this module are disabled unless pandas and pyarrow are importable.
 """
 
+import datetime
 import os
 import unittest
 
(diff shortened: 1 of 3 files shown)


tests/test_parquet.py
```diff
@@ -138,6 +139,7 @@ def _makeSimpleNumpyTable(include_multidim=False, include_bigendian=False):
         ("f", "i8"),
         ("strcol", "U10"),
         ("bytecol", "a10"),
+        ("dt", "datetime64[ns]"),
```
BrunoSanchez: Can you add to the test the `us` as well? Is it relevant?

BrunoSanchez (approved):
This looks good to me. Just one suggestion:
This bug I found it first using the `datetime[us]` so adding it to the test cases might be good.

tests/test_parquet.py
erykoff: I will add that!

PR #1042, opened 2024-07-30
DM-45489: Add more test cover for RemoteButler queries

dhirving:
Fixed an issue where the "Hybrid" query results object for queryDataIds was testing the DirectButler implementation in some cases where it should have been testing the RemoteButler implementation.  This revealed some small missing exception handling and minor discrepancies between the two implementations, which are now fixed.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/tests/_registry.py
+++ b/python/lsst/daf/butler/registry/tests/_registry.py
@@ -3138,12 +3138,20 @@ class RegistryTests(ABC):
         registry = self.makeRegistry()
         self.loadData(registry, "base.yaml")
         bind = {"time": astropy.time.Time("2020-01-01T01:00:00", format="isot", scale="tai")}
-        with self.assertRaisesRegex(LookupError, r"No dimension element with name 'foo' in 'foo\.bar'\."):
-            registry.queryDataIds(["detector"], where="foo.bar = 12")
+        # The diagnostics raised are slightly different between the old query
+        # system (ValueError, first error string) and the new query system
+        # (InvalidQueryError, second error string).
         with self.assertRaisesRegex(
-            LookupError, "Dimension element name cannot be inferred in this context."
+            (LookupError, InvalidQueryError),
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -2663,11 +2671,12 @@ def testSkipCalibs(self):
...
-        with self.assertRaises(NotImplementedError):
-            registry.queryDataIds(
-                ["instrument", "detector", "exposure"], datasets="bias", collections=coll_list
-            ).count()
+        if not self.supportsCalibrationCollectionInFindFirst:
+            # explicit list will raise if there are temporal dimensions
+            with self.assertRaises(NotImplementedError):
+                registry.queryDataIds(
+                    ["instrument", "detector", "exposure"], datasets="bias", collections=coll_list
+                ).count()
```
dhirving, line 2679: In the new query system this is returning zero rows, which I don't think is right.. it should probably be returning the data IDs for all the bias datasets instead?

TallJimbo: I think the the behavior is correct, and the comment above this test is wrong: I don't see anything that inserts `exposure` dimension records, so the query is going to return no rows because it's got a join to that empty table.  Note that the dimensions of `bias` are just `{instrument, detector}`; the match to `exposure` is going to be a temporal join between the CALIBRATION collection's validity range and `exposure.timespan`.

So I suspect there's supposed to be an `exposure` record inserted with a timespan that overlaps the validity range of exactly one `bias` for or or two `detector`s in this test, as that'd make the test much more interesting for the new system (and it wouldn't affect the behavior of the old query system, which is probably why it wasn't done).  And if we do that the new query system should return rows for that exposure and whatever detectors have a bias with the matching timespan.  But it's quite possible that behavior is already covered in other tests of the new query system and hence there's no need to re-check it here.

Note that we don't really care about the case where the exposure's timespan overlaps the validity ranges of multiple biases; this query might still be sound, but it'd make any find-first search for the `bias` fail, and hence it represents a practically useless validity range.

TallJimbo (approved):

PR #1043, opened 2024-07-30
DM-45119: Update logic for checking overlap dimensions in relations

andy-slac:
New logic only checks for required elements instead of full set of
dimensions. This fixes special case of `healpix11` dataset types
in dc2 repo. There is no unit test for this special case, but I verified
that query-datasets and query-data-ids now work correctly.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/dimensions/static.py
+++ b/python/lsst/daf/butler/registry/dimensions/static.py
@@ -385,9 +385,9 @@ class StaticDimensionRecordStorageManager(DimensionRecordStorageManager):
     ) -> tuple[Relation, bool]:
         # Docstring inherited.
         overlap_relationship = frozenset(
-            self.universe[element1].dimensions.names | self.universe[element2].dimensions.names
+            self.universe[element1].required.names | self.universe[element2].required.names
         )
-        if overlap_relationship in existing_relationships:
+        if any(overlap_relationship.issubset(rel) for rel in existing_relationships):
             return context.preferred_engine.make_join_identity_relation(), False
         overlaps: Relation | None = None
         needs_refinement: bool = False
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/registry/dimensions/static.py
```diff
@@ -385,9 +385,9 @@ def make_spatial_join_relation(
     ) -> tuple[Relation, bool]:
         # Docstring inherited.
         overlap_relationship = frozenset(
-            self.universe[element1].dimensions.names | self.universe[element2].dimensions.names
+            self.universe[element1].required.names | self.universe[element2].required.names
         )
-        if overlap_relationship in existing_relationships:
+        if any(overlap_relationship.issubset(rel) for rel in existing_relationships):
```
TallJimbo: I think we can make this a little safer by keeping the old check and the new one, and combinations between them:
```
group1 = self.universe[element1].minimal_group
group2 = self.universe[element2].minimal_group
overlap_relationships = {
    frozenset(a | b) for a, b in itertools.product(
        [group1.names, group1.required],
        [group2.names, group2.required],
    )
}
if not overlap_relationships.isdisjoint(existing_relationships):
    return context.preferred_engine.make_join_identity_relation(), False
```

I'm particularly interested in keeping the old check, because while the required dimensions are the ones brought in by a dataset join, the full dimensions are the ones brought in by the temp tables created by materialize, and I'm not sure we have enough test coverage of QG generation edge cases to spot any regressions this change might introduce there otherwise.

TallJimbo (approved):
I've made a suggestion for how to make this a bit more general, but overall I'm very much in favor of doing just this much and no more on the old query system.

PR #1044, opened 2024-07-31
DM-45517: No longer check parent when checking compatibility

parejkoj:
## Checklist

- [X] ran Jenkins
- [??] added a release note for user-visible changes to `doc/changes`

I'm not sure whether this is a "user-visible" change, and if so, I don't know where to put it in the docs.

### DIFF
--- a/python/lsst/daf/butler/_dataset_type.py
+++ b/python/lsst/daf/butler/_dataset_type.py
@@ -323,6 +323,11 @@ class DatasetType:
             or the storage class associated with the other can be converted to
             this.
         """
+        # Support conversion when the components have the same name and
+        # convertible parents, e.g. Parquet storage classes.
+        if self.component() is not None and self.component() == other.component():
+            return self.makeCompositeDatasetType().is_compatible_with(other.makeCompositeDatasetType())
+
         mostly_equal = self._equal_ignoring_storage_class(other)
         if not mostly_equal:
             return False
--- a/python/lsst/daf/butler/_dataset_type.py
+++ b/python/lsst/daf/butler/_dataset_type.py
@@ -284,10 +284,7 @@ class DatasetType:
             return False
         if self._isCalibration != other._isCalibration:
             return False
-        if self._parentStorageClass is not None and other._parentStorageClass is not None:
-            return self._parentStorageClass == other._parentStorageClass
-        else:
-            return self._parentStorageClassName == other._parentStorageClassName
+        return True
 
     def __eq__(self, other: Any) -> bool:
         mostly_equal = self._equal_ignoring_storage_class(other)
@@ -296,6 +293,13 @@ class DatasetType:
 
         # Be careful not to force a storage class to import the corresponding
         # python code.
+        if self._parentStorageClass is not None and other._parentStorageClass is not None:
+            if self._parentStorageClass != other._parentStorageClass:
+                return False
+        else:
+            if self._parentStorageClassName != other._parentStorageClassName:
+                return False
+
         if self._storageClass is not None and other._storageClass is not None:
             if self._storageClass != other._storageClass:
                 return False
@@ -309,7 +313,8 @@ class DatasetType:
 
         Compatibility requires a matching name and dimensions and a storage
         class for this dataset type that can convert the python type associated
-        with the other storage class to this python type.
+        with the other storage class to this python type. Parent storage class
+        compatibility is not checked at all for components.
 
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


tests/test_parquet.py
```diff
@@ -642,6 +642,13 @@ def testWriteSingleIndexDataFrameReadAsArrowTable(self):
             if schema.field(name).type not in (pa.string(), pa.binary()):
                 self.assertEqual(schema.field(name).type, schema2.field(name).type)
 
+        # Check getting a component while overriding the storage class via
```
timj: Can you move the test up to line 631 because this test is doing the same test but rather than specifying a storage class override on the get() it's doing the override in the DatasetType.

timj:
```suggestion
        # Check getting a component while overriding the parent storage class via
```
This is an important distinction because what's happening is that you are changing from DataFrame to ArrowAstropy and then using columns which ends up with a different definition automatically, whereas above the test is using DataFrame.columns and then changing the storage class.

This code works already:
```python
columnsType = self.datasetType.makeComponentDatasetType("columns").overrideStorageClass("ArrowColumnList")
self.assertEqual(columns2, self.butler.get(columnsType))
```
but could you add that test as well to make it clear that the order of the component making / overriding matters?

python/lsst/daf/butler/_dataset_type.py
```diff
@@ -323,6 +323,11 @@ class for this dataset type that can convert the python type associated
             or the storage class associated with the other can be converted to
             this.
         """
+        # Support conversion when the components have the same name and
+        # convertible parents, e.g. Parquet storage classes.
+        if self.component() is not None and self.component() == other.component():
+            return self.makeCompositeDatasetType().is_compatible_with(other.makeCompositeDatasetType())
```
timj: I don't think this is the right fix. In my testing if I have two composite dataset types that have components that are incompatible this says that everything is fine. We have to test somewhere that the components are compatible.

The parquet test is failing for a different reason. It's failing because `_equal_ignoring_storage_class` is not ignoring parent storage class and so this returns False before the actual storage class is checked.  I think the fix here is to think harder about the parent storage class in this context. Maybe is_compatible_with has to check parent compatibility and component compatibility or maybe it doesn't care about the parent at all.

timj (changes requested):
This needs discussion. The fix is not correct.

python/lsst/daf/butler/_dataset_type.py
TallJimbo: I was thinking that we could declare that two component dataset types are compatible if either of the following is true:
 - if the component storage classes are directly compatible (this is the only supported case on `main`);
 - if the parent storage classes are compatible and the component names are the same (the condition added on this branch, but I think it's implied to already work by the other way the same `get` is already written on the test).

The specific situation here is that as long as `is_compatible_with` doesn't return `False` for `ArrowAstropy.columns` vs. `DataFrame.columns`, the way we call the `Formatter` already works, and we'd rather let the `Formatter` do the conversion anyway.  So I'd rather not have to define explicit conversions between the two component storage classes in this case, since that wouldn't really reflect how the conversion works.

I am somewhat worried that this would lead us to declare two storage classes as convertible when a conversion doesn't actually exist (except in the formatter), but I don't see that coming up in practice in this case so I wasn't sure how much to worry about it.

TallJimbo: Maybe it would be better if the new check here moved up to the place where we call `is_compatible_with` in `Butler.get`, so that my proposed new rule on component compatibility is not a general one, but rather a chance to let the formatter handle a request where the `Butler` cannot rule out a conversion working.

timj: The relevant `columns` storage class definitions already register converters and are compatible (that's the reason why the test above the new test was working since the storage class override on the `butler.get()` was specifying the column storage class).

I think if we start saying that conversions for components don't have to work if the parents can convert that might break the `storageClass=Override` option for components. I don't like the idea of not checking component compatibility at all.

timj: I think `is_compatible_with` not checking parent storage class but requiring that components are convertible is good enough isn't it? The test passes now if I remove the parent storage class check. We can assume that if you are comparing components you don't care about the parent.

TallJimbo (approved):
Looks good.  I'm surprised but happen that the component storage classes we care about here are already compatible without any extra effort.

timj:
>  I'm surprised but happy that the component storage classes we care about here are already compatible 

If they weren't then many of the tests that already existed for parquet would have failed since that's how `storageClass=` overrides worked if you asked for columns. @erykoff added the converters up front.

timj (approved):

timj:
@parejkoj I was expecting to merge this once my Jenkins run completed.

parejkoj:
Oops, sorry! We can revert if jenkins fails.

timj:
Jenkins passed.

PR #1045, opened 2024-08-01
DM-45556: Switch command-line tools to new query interface

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/script/certifyCalibrations.py
+++ b/python/lsst/daf/butler/script/certifyCalibrations.py
@@ -78,8 +78,13 @@ def certifyCalibrations(
     if not search_all_inputs and registry.getCollectionType(input_collection) is CollectionType.CHAINED:
         input_collection = next(iter(registry.getCollectionChain(input_collection)))
 
-    refs = set(registry.queryDatasets(dataset_type_name, collections=[input_collection]))
-    if not refs:
-        raise RuntimeError(f"No inputs found for dataset {dataset_type_name} in {input_collection}.")
+    with butler._query() as query:
+        results = query.datasets(dataset_type_name, collections=input_collection)
+        refs = set(results)
+        if not refs:
+            explanation = "\n".join(results.explain_no_results())
+            raise RuntimeError(
+                f"No inputs found for dataset {dataset_type_name} in {input_collection}. {explanation}"
+            )
     registry.registerCollection(output_collection, type=CollectionType.CALIBRATION)
     registry.certify(output_collection, refs, timespan)
--- a/python/lsst/daf/butler/script/exportCalibs.py
+++ b/python/lsst/daf/butler/script/exportCalibs.py
@@ -29,6 +29,7 @@ from __future__ import annotations
 import logging
 import os
 from collections.abc import Iterable
+from operator import attrgetter
 from typing import TYPE_CHECKING
 
 from astropy.table import Table
@@ -170,7 +171,7 @@ def exportCalibs(
         log.info("Saving %d dataset(s)", len(datasetsToExport))
         export.saveDatasets(datasetsToExport)
 
-    sortedDatasets = sorted(datasetsToExport, key=lambda x: x.datasetType.name)
+    sortedDatasets = sorted(datasetsToExport, key=attrgetter("datasetType.name", "dataId"))
 
     requiredDimensions: set[str] = set()
     for ref in sortedDatasets:
--- a/python/lsst/daf/butler/script/exportCalibs.py
+++ b/python/lsst/daf/butler/script/exportCalibs.py
@@ -171,7 +171,7 @@ def exportCalibs(
         requiredDimensions.update(ref.dimensions.names)
     dimensionColumns = {
         dimensionName: [ref.dataId.get(dimensionName, "") for ref in sortedDatasets]
-        for dimensionName in requiredDimensions
+        for dimensionName in sorted(requiredDimensions)
     }
 
     return Table(
--- a/python/lsst/daf/butler/script/exportCalibs.py
+++ b/python/lsst/daf/butler/script/exportCalibs.py
@@ -38,20 +38,20 @@ from .._butler import Butler
 from ..registry import CollectionType
 
 if TYPE_CHECKING:
-    from lsst.daf.butler import DatasetRef, DatasetType, Registry
+    from lsst.daf.butler import DatasetRef, DatasetType
 
 log = logging.getLogger(__name__)
 
 
-def parseCalibrationCollection(
-    registry: Registry, collection: str, datasetTypes: Iterable[DatasetType]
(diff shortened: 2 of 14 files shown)


tests/test_cliCmdQueryDataIds.py
```diff
@@ -112,7 +112,7 @@ def testNoResultsEasy(self):
             where="instrument='Cam1' AND skymap='SkyMap1' AND visit=1 AND tract=1",
         )
         self.assertIsNone(res, msg)
-        self.assertIn("yields no results when applied to", msg)
```
timj, line 115: This message comes from daf_relation and new query system doesn't report any failure reason.

TallJimbo: :+1:.  Explaining no results is something the new query system does not do as well yet, though the old system was actually pretty hit or miss and the tests kind of disguised that.  The old system walked back through the tree of joins and WHERE constraints, executing LIMIT 1 queries to see where the result set is first empty, but it never tried applying those joins and WHERE constraints in a different order.

python/lsst/daf/butler/script/exportCalibs.py
```diff
@@ -170,7 +171,7 @@ def exportCalibs(
         log.info("Saving %d dataset(s)", len(datasetsToExport))
         export.saveDatasets(datasetsToExport)
 
-    sortedDatasets = sorted(datasetsToExport, key=lambda x: x.datasetType.name)
+    sortedDatasets = sorted(datasetsToExport, key=attrgetter("datasetType.name", "dataId"))
```
TallJimbo, line 167: :exploding_head: I had no idea `attrgetter` handled dot-separated names.

python/lsst/daf/butler/script/exportCalibs.py
```diff
@@ -69,23 +70,23 @@ def parseCalibrationCollection(
...
+        with butler._query() as query:
+            results = query.datasets(calibType, collections=collection, find_first=False)
+
+            try:
+                refs = list(results.with_dimension_records())
+            except Exception as e:
+                e.add_note(f"Error from querying dataset type {calibType} and collection {collection}")
+                raise
+            exportDatasets.extend(refs)
+            exportCollections.update([ref.run for ref in refs])
```
TallJimbo: I don't think this is necessary; `RUN` collections are always and only exported implicitly when datasets from them are exported.

python/lsst/daf/butler/script/exportCalibs.py
```diff
@@ -37,20 +38,20 @@
...
+    from lsst.daf.butler import DatasetRef, DatasetType
 
 log = logging.getLogger(__name__)
 
 
 def parseCalibrationCollection(
-    registry: Registry, collection: str, datasetTypes: Iterable[DatasetType]
-) -> tuple[list[str], list[DatasetRef]]:
+    butler: Butler, collection: str, datasetTypes: Iterable[DatasetType]
+) -> tuple[set[str], list[DatasetRef]]:
```
TallJimbo: Might be a good time to rename this function, since it's not really doing any "parsing".

python/lsst/daf/butler/script/queryDatasets.py
```diff
@@ -214,12 +205,26 @@ def getTables(self) -> list[AstropyTable]:
...
 
-    def getDatasets(self) -> DatasetQueryResults:
-        """Get the datasets as a list of ``DatasetQueryResults``.
+    def getDatasets(self) -> Iterator[DatasetRef]:
+        """Get the datasets as a list.
 
         Returns
         -------
-        refs : `lsst.daf.butler.registry.queries.DatasetQueryResults`
+        refs : `list`
```
TallJimbo: Now it's an iterator.

python/lsst/daf/butler/script/removeRuns.py
```diff
@@ -96,17 +95,20 @@ def _getCollectionInfo(
...
-        all_results = butler.registry.queryDatasets(..., collections=collectionName)
-        assert isinstance(all_results, DatasetQueryResults)
-        for r in all_results.byParentDatasetType():
-            if r.any(exact=False, execute=False):
-                datasets[r.parentDatasetType.name] += r.count(exact=False)
+        with butler._query() as query:
+            for dt in dataset_types:
+                results = query.datasets(dt, collections=collectionName)
+                if results.any(exact=False, execute=False):
+                    datasets[dt.name] += results.count(exact=False)
```
TallJimbo: It'll be slightly faster to just execute `count` and ignore the result if it is zero.

python/lsst/daf/butler/script/certifyCalibrations.py
```diff
@@ -78,8 +78,13 @@ def certifyCalibrations(
     if not search_all_inputs and registry.getCollectionType(input_collection) is CollectionType.CHAINED:
         input_collection = next(iter(registry.getCollectionChain(input_collection)))
 
-    refs = set(registry.queryDatasets(dataset_type_name, collections=[input_collection]))
-    if not refs:
-        raise RuntimeError(f"No inputs found for dataset {dataset_type_name} in {input_collection}.")
+    with butler._query() as query:
+        results = query.datasets(dataset_type_name, collections=input_collection)
```
TallJimbo, line 82: The default for find-first has changed, so this is a behavior change.  But I think it's an unambiguous improvement, because in the past, if the query had returned multiple results for the same data ID, the `certify` call later would have choked on them.

TallJimbo (approved):

python/lsst/daf/butler/script/exportCalibs.py
timj: Neither did I until I read the documentation...

python/lsst/daf/butler/script/exportCalibs.py
timj: Ok. I will remove it completely.

python/lsst/daf/butler/script/queryDatasets.py
timj: Yes. Thanks. In my first pass it was a list and then I had to reorganize because of the prune-datasets command line mocking.

python/lsst/daf/butler/script/removeRuns.py
timj: Thanks. That crossed my mind. I was wondering whether the `execute=False` on the `any` vs the `execute=True` on the `count` would be a performance difference but I didn't run any tests.

TallJimbo: The `any(execute=False)` summary check is going to be run internally by `count` (or any other query execution method) automatically, so in this case it's just doing that check twice.  It should be a fast check the second time (once the cache is populated), so it's not a big deal, but it is unnecessary.

PR #1047, opened 2024-08-05
DM-45429: Initial implementation of a general query result

andy-slac:
Registry method `queryDatasetAssociations` is reimplemented (for both
direct and remote butler) to use new query system and new general query
result class.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/column_spec.py
+++ b/python/lsst/daf/butler/column_spec.py
@@ -53,6 +53,7 @@ from lsst.sphgeom import Region
 
 from . import arrow_utils, ddl
 from ._timespan import Timespan
+from .pydantic_utils import SerializableRegion, SerializableTime
 
 if TYPE_CHECKING:
     from .name_shrinker import NameShrinker
@@ -134,6 +135,18 @@ class _BaseColumnSpec(pydantic.BaseModel, ABC):
         """
         raise NotImplementedError()
 
+    @abstractmethod
+    def type_adapter(self) -> pydantic.TypeAdapter:
+        """Return pydantic type adapter that converts values of this column to
+        or from serializable format.
+
+        Returns
+        -------
+        type_adapter : `pydantic.TypeAdapter`
+            A converter instance.
+        """
+        raise NotImplementedError()
+
     def display(self, level: int = 0, tab: str = "  ") -> list[str]:
         """Return a human-reader-focused string description of this column as
         a list of lines.
@@ -178,6 +191,10 @@ class IntColumnSpec(_BaseColumnSpec):
         # Docstring inherited.
         return arrow_utils.ToArrow.for_primitive(self.name, pa.uint64(), nullable=self.nullable)
 
+    def type_adapter(self) -> pydantic.TypeAdapter:
+        # Docstring inherited.
+        return pydantic.TypeAdapter(self.pytype)
+
 
 @final
 class StringColumnSpec(_BaseColumnSpec):
@@ -197,6 +214,10 @@ class StringColumnSpec(_BaseColumnSpec):
     def to_arrow(self) -> arrow_utils.ToArrow:
         # Docstring inherited.
         return arrow_utils.ToArrow.for_primitive(self.name, pa.string(), nullable=self.nullable)
+
+    def type_adapter(self) -> pydantic.TypeAdapter:
+        # Docstring inherited.
+        return pydantic.TypeAdapter(self.pytype)
 
 
 @final
@@ -224,6 +245,10 @@ class HashColumnSpec(_BaseColumnSpec):
             nullable=self.nullable,
         )
 
+    def type_adapter(self) -> pydantic.TypeAdapter:
+        # Docstring inherited.
+        return pydantic.TypeAdapter(self.pytype)
+
 
 @final
 class FloatColumnSpec(_BaseColumnSpec):
@@ -238,6 +263,10 @@ class FloatColumnSpec(_BaseColumnSpec):
         assert self.nullable is not None, "nullable=None should be resolved by validators"
         return arrow_utils.ToArrow.for_primitive(self.name, pa.float64(), nullable=self.nullable)
 
+    def type_adapter(self) -> pydantic.TypeAdapter:
+        # Docstring inherited.
+        return pydantic.TypeAdapter(self.pytype)
+
 
 @final
 class BoolColumnSpec(_BaseColumnSpec):
@@ -250,6 +279,10 @@ class BoolColumnSpec(_BaseColumnSpec):
     def to_arrow(self) -> arrow_utils.ToArrow:
         # Docstring inherited.
         return arrow_utils.ToArrow.for_primitive(self.name, pa.bool_(), nullable=self.nullable)
+
+    def type_adapter(self) -> pydantic.TypeAdapter:
+        # Docstring inherited.
+        return pydantic.TypeAdapter(self.pytype)
 
 
 @final
@@ -264,6 +297,10 @@ class UUIDColumnSpec(_BaseColumnSpec):
         # Docstring inherited.
         assert self.nullable is not None, "nullable=None should be resolved by validators"
         return arrow_utils.ToArrow.for_uuid(self.name, nullable=self.nullable)
+
+    def type_adapter(self) -> pydantic.TypeAdapter:
+        # Docstring inherited.
+        return pydantic.TypeAdapter(self.pytype)
 
 
 @final
@@ -284,6 +321,10 @@ class RegionColumnSpec(_BaseColumnSpec):
         assert self.nullable is not None, "nullable=None should be resolved by validators"
         return arrow_utils.ToArrow.for_region(self.name, nullable=self.nullable)
 
+    def type_adapter(self) -> pydantic.TypeAdapter:
+        # Docstring inherited.
+        return pydantic.TypeAdapter(SerializableRegion)
+
 
 @final
 class TimespanColumnSpec(_BaseColumnSpec):
@@ -298,6 +339,10 @@ class TimespanColumnSpec(_BaseColumnSpec):
     def to_arrow(self) -> arrow_utils.ToArrow:
         # Docstring inherited.
         return arrow_utils.ToArrow.for_timespan(self.name, nullable=self.nullable)
+
+    def type_adapter(self) -> pydantic.TypeAdapter:
+        # Docstring inherited.
+        return pydantic.TypeAdapter(self.pytype)
 
 
 @final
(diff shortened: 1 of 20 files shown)


andy-slac:
@TallJimbo, I have an initial version of a general query result calss and I reimplemented `queryDatasetAssociations` to use that to make sure it can work. Could you check that what I've done so far matches your model for it?

python/lsst/daf/butler/queries/_general_query_results.py
```diff
@@ -0,0 +1,125 @@
...
+            Result row as dictionary, the keys are `ResultColumn` instances.
+        """
+        for page in self._driver.execute(self._spec, self._tree):
+            columns = tuple(page.spec.get_result_columns())
+            for row in page.rows:
+                yield dict(zip(columns, row))
+
+    def iter_refs(self, dataset_type: DatasetType) -> Iterator[tuple[DatasetRef, dict[ResultColumn, Any]]]:
+        """Iterate over result rows and return DatasetRef constructed from each
+        row and an original row.
```
TallJimbo: There's one place in QG generation where we'd want a method sort of like this that yields tuples of `DatasetRef` and `DataCoordinate` (it currently calls `DataCoordinateQueryResults.findRelatedDatasets`).

I had originally planned to do that via a special dict-like row object that had single-row `get_ref`, `get_data_id` etc. methods, but I think I like your iteration adapter better as it allows for doing work common to all rows up front.  But I'd like to extend this to be able to support at least `DatasetRef` + `DataCoordinate` + [extra stuff] somehow, and maybe even support `DatassetRef` objects for multiple dataset.  Maybe something like this?
```
def iter_tuples(self, *dataset_type: DatasetType) -> Iterator[tuple[DataCoordinate, dict[str, DatasetRef], dict[ResultColumn, Any]]
```
where it always returns the full data ID, a dict of `DatasetRef` keyed by dataset type name, and then whatever columns aren't included in one of those?  To be honest I'm not thrilled with this, either, but maybe it'll inspire you to find something better?

python/lsst/daf/butler/queries/_general_query_results.py
```diff
@@ -0,0 +1,125 @@
...
+        super().__init__(driver, tree)
+        self._spec = spec
+
+    def __iter__(self) -> Iterator[dict[ResultColumn, Any]]:
+        """Iterate over result rows.
+
+        Yields
+        ------
+        row_dict : `dict` [`ResultColumn`, `Any`]
+            Result row as dictionary, the keys are `ResultColumn` instances.
```
TallJimbo: Another possibility for the key type here is a `<dimension-or-dataset-type-name>.<column>` string; that might be better if we make this type more publicly accessible.  In either case it should probably match however we let the user specify the desired result columns in any public interface that returns one of these.

TallJimbo (commented):
As I mentioned in the stand-up, I had imagined adding a to-be-public `Query.general` method rather than `Query.dataset_associations`, as sort of a power-user escape valve.  I'm not opposed to also having `Query.dataset_associations`.

python/lsst/daf/butler/queries/_general_query_results.py
andy-slac: I switched to `str` as a key, it is indeed more natural as API uses strings as parameters. I also replaced `dataset_associations` with `general` which takes the same parameters as `GeneralResultSpec`.

python/lsst/daf/butler/queries/_general_query_results.py
andy-slac: The return annotation is hard to read. Would it make sense to introduce a helper NamedTuple like this:
```
class GeneralResultTuple(typing.NamedTuple):

    data_id: DataCoordinate
    refs: dict[str, DtasetRef]  # or maybe just list[DatasetRef]?
    raw_row: dict[str, Any]

class GeneralQueryResults(QueryResultsBase):

    def iter_tuples(self, *dataset_type: DatasetType) -> Iterator[GeneralResultTuple]:
        ...
```

TallJimbo: :+1: I do like `NamedTuple` for this; you can still unpack it if you want to, but it's also very clear what's in it and you can hang documentation on it.  And I like `list` rather than `dict` for the `DatasetRefs`, too, since we can make that ordering consistent with the argument rather than arbitrary.

python/lsst/daf/butler/queries/_query.py
```diff
@@ -292,6 +298,45 @@ def dimension_records(self, element: str) -> DimensionRecordQueryResults:
         result_spec = DimensionRecordResultSpec(element=self._driver.universe[element])
         return DimensionRecordQueryResults(self._driver, tree, result_spec)
 
+    def general(
+        self,
+        dimensions: DimensionGroup,
+        dimension_fields: Mapping[str, set[str]] = {},
+        dataset_fields: Mapping[str, set[DatasetFieldName]] = {},
```
TallJimbo: Since we don't need to mutate them, better to make the value types `collections.abc.Set` rather than `set`.

What do you think about making all of these keyword-only arguments, and adding a `*args: str`  variadic argument that we'd filter though `interpret_identifier` (in `queries/_identifiers.py`) and use to update the existing categorized values?

It might also be nice to have a simple way to say "all of the fields necessary for `DatasetRefs` of this dataset type", maybe `datasets={name: ...}` or something?

python/lsst/daf/butler/remote_butler/_registry.py
```diff
@@ -513,7 +513,19 @@ def queryDatasetAssociations(
...
+        resolved_collections = self.queryCollections(
+            collections, datasetType=datasetType, collectionTypes=collectionTypes, flattenChains=flattenChains
+        )
+        with self._butler._query() as query:
+            query = query.join_dataset_search(datasetType, resolved_collections)
+            result = query.general(
+                datasetType.dimensions,
+                dataset_fields={datasetType.name: {"dataset_id", "run", "collection", "timespan"}},
+            )
+            yield from DatasetAssociation.from_query_result(result, datasetType)
```
andy-slac, line 528: @timj, this is how replacement code for `queryDatasetAssociations` is going to look like. I checked `RepoExportContext` class to see if I can replace `queryDatasetAssociations` usage there, but that class does not even have butler instance now, and it uses registry for everything. We can do it later and keep `queryDatasetAssociations` in registry for now. Both direct and remote registries use new query system.

timj: Great. I had forgotten that we need to change the export code so that it doesn't use registry.

python/lsst/daf/butler/queries/_general_query_results.py
```diff
@@ -83,33 +102,36 @@ def __iter__(self) -> Iterator[dict[str, Any]]:
...
+            id_key = f"{dataset_type.name}.dataset_id"
+            run_key = f"{dataset_type.name}.run"
+            dataset_keys.append((dimensions, id_key, run_key))
         for row in self:
-            values = tuple(row[key] for key in data_id_keys)
-            data_id = DataCoordinate.from_required_values(dimensions, values)
-            ref = DatasetRef(dataset_type, data_id, row[run_key], id=row[id_key])
-            yield ref, row
+            values = tuple(row[key] for key in all_dimensions.required)
+            data_coordinate = DataCoordinate.from_required_values(all_dimensions, values)
```
TallJimbo: `GeneralResultSpec` guarantees that the full dimensions are available.  So we can and probably should extract the full data IDs here (and below for the dataset data IDs).

python/lsst/daf/butler/queries/_query.py
```diff
@@ -329,10 +346,32 @@ def general(
...
+        # Parse all names.
+        context = IdentifierContext(dimensions, set(self._tree.datasets))
+        for name in names:
+            identifier = interpret_identifier(context, name)
+            match identifier:
+                case DimensionKeyReference(dimension=dimension):
+                    # Could be because someone asked for the key field.
+                    # Potentially expand dimensions (but tree needs to have
+                    # them already or there will be an error).
+                    dimensions = dimensions.union(dimension.minimal_group)
```
TallJimbo: Instead of updating a `DimensionGroup` variable as you go, I'd recommend accumulating a `set` of dimension names and doing `universe.conform(names | dimension_fields_dict.keys())` at the end.  And then you can do something like https://github.com/lsst/daf_butler/blob/8e576370746fa86e5d065499e4f114fb97b83289/python/lsst/daf/butler/queries/_query.py#L200-L201 to join in the missing dimensions.  That will make this a lot easier to use, as well as consistent with how `where` handles dimensions that aren't in the query yet.

python/lsst/daf/butler/column_spec.py
```diff
@@ -88,6 +89,102 @@
...
+            return value.isoformat()
+        else:
+            return self._astropy_adapter.dump_python(value)
+
+    def deserialize(self, value: Any) -> Any:
+        # Docstring inherited.
+        if value is None:
+            return None
+        elif isinstance(value, str):
+            return datetime.datetime.fromisoformat(value)
```
TallJimbo, line 183: I understand wanting to accept `datetime.datetime` and `str` from users and databases for backwards compatibility, but I don't understand why we'd want to return a `datetime.datetime` to users or serialize to `str` going forward.

My recollection is that we wanted to eventually switch to astropy/nsec for ingest_time everywhere, but it's a tricky migration due to the timezone stuff and we don't really have a sound plan for doing it.  But I'd like to get the in-memory Python types consistent.

If that's all true, I think we could definitely support coercion from `str` or `datetime.datetime` in `pydantic_utils._deserialize_time`, and that might be enough to replace the new serialization dispatch logic here.  But it seems likely I'm not thinking about at least one of the places where we need to do the time type conversions.

TallJimbo (approved):

python/lsst/daf/butler/column_spec.py
andy-slac: I chose this for compatibility between direct butler and remote butler. Queries with direct butler return whatever sqlalchemy does for that column, which is `datetime` in current schema. It may be surprising for clients if remote butler returned astropy time instead for the same query. We could convert all `datetime` instances returned from queries to astropy, then serialization would need to handle only one type of time. 
These new methods are only used to (de)serialize query results, I'm not sure what happens when the query itself includes time instances in `bind`, need to check.

TallJimbo: Thus far it hasn't actually been possible to get the `ingest_date` back from a query - there's no interface in the Registry query system that can do it, and the "general" result type is the only way to do it in the new one.  So I think we've still got until we open up `butler._query*` as a public interface to get this this fixed, and IMO that means coercing to astropy.  I'm happy for that to be another ticket, though.

andy-slac: OK, I'll make a new ticket for that.

PR #1049, opened 2024-08-06
DM-45432: Add explicit tests for regions and timespans with query_datasets

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_simpleButler.py
+++ b/tests/test_simpleButler.py
@@ -32,6 +32,7 @@ import os
 import re
 import tempfile
 import unittest
+from operator import attrgetter
 from typing import Any
 
 try:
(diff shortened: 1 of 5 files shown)


tests/test_simpleButler.py
```diff
@@ -794,6 +798,85 @@ def makeButler(self, writeable: bool = False) -> Butler:
         DatastoreMock.apply(butler)
         return butler
 
+    def test_region_queries(self):
```
dhirving: The rest of the query tests are in `lsst/daf/butler/tests/butler_queries.py`.  That might be a better place for these because it runs against all database configurations (sqlite, postgres, remote).  `test_simplebutler.py` doesn't cover postgres.

dhirving (approved):

tests/test_simpleButler.py
timj: Thanks for pointing that out. I moved them.

PR #1053, opened 2024-08-13
DM-45738: Experiments with extending butler collections API

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler_collections.py
+++ b/python/lsst/daf/butler/_butler_collections.py
@@ -210,7 +210,6 @@ class ButlerCollections(ABC, Sequence):
         """
         raise NotImplementedError()
 
-    @abstractmethod
     def x_query(
         self,
         expression: str | Iterable[str],
@@ -219,6 +218,8 @@ class ButlerCollections(ABC, Sequence):
         include_chains: bool | None = None,
     ) -> Sequence[str]:
         """Query the butler for collections matching an expression.
+
+        **This is an experimental interface that can change at any time.**
 
         Parameters
         ----------
@@ -238,6 +239,61 @@ class ButlerCollections(ABC, Sequence):
         Returns
         -------
         collections : `~collections.abc.Sequence` [ `str` ]
+            The names of collections that match ``expression``.
+
+        Notes
+        -----
+        The order in which collections are returned is unspecified, except that
+        the children of a `~CollectionType.CHAINED` collection are guaranteed
+        to be in the order in which they are searched.  When multiple parent
+        `~CollectionType.CHAINED` collections match the same criteria, the
+        order in which the two lists appear is unspecified, and the lists of
+        children may be incomplete if a child has multiple parents.
+
+        The default implementation is a wrapper around `x_query_info`.
+        """
+        collections_info = self.x_query_info(
+            expression,
+            collection_types=collection_types,
+            flatten_chains=flatten_chains,
+            include_chains=include_chains,
+        )
+        return [info.name for info in collections_info]
+
+    @abstractmethod
+    def x_query_info(
+        self,
+        expression: str | Iterable[str],
+        collection_types: Set[CollectionType] | CollectionType | None = None,
+        flatten_chains: bool = False,
+        include_chains: bool | None = None,
+        include_parents: bool = False,
+    ) -> Sequence[CollectionInfo]:
+        """Query the butler for collections matching an expression and
+        return detailed information about those collections.
+
+        **This is an experimental interface that can change at any time.**
+
+        Parameters
+        ----------
+        expression : `str` or `~collections.abc.Iterable` [ `str` ]
+            One or more collection names or globs to include in the search.
+        collection_types : `set` [`CollectionType`], `CollectionType` or `None`
+            Restrict the types of collections to be searched. If `None` all
+            collection types are searched.
+        flatten_chains : `bool`, optional
+            If `True` (`False` is default), recursively yield the child
+            collections of matching `~CollectionType.CHAINED` collections.
+        include_chains : `bool` or `None`, optional
+            If `True`, yield records for matching `~CollectionType.CHAINED`
+            collections. Default is the opposite of ``flatten_chains``:
+            include either CHAINED collections or their children, but not both.
+        include_parents : `bool`, optional
+            Whether the returned information includes parents.
+
+        Returns
+        -------
+        collections : `~collections.abc.Sequence` [ `CollectionInfo` ]
             The names of collections that match ``expression``.
 
         Notes
@@ -301,6 +357,8 @@ class ButlerCollections(ABC, Sequence):
     @abstractmethod
     def x_remove(self, name: str) -> None:
         """Remove the given collection from the registry.
+
+        **This is an experimental interface that can change at any time.**
 
         Parameters
         ----------
--- a/python/lsst/daf/butler/_butler_collections.py
+++ b/python/lsst/daf/butler/_butler_collections.py
@@ -27,11 +27,30 @@
 
 from __future__ import annotations
 
-__all__ = ("ButlerCollections",)
+__all__ = ("ButlerCollections", "CollectionInfo")
 
 from abc import ABC, abstractmethod
-from collections.abc import Iterable, Sequence
+from collections.abc import Iterable, Sequence, Set
 from typing import Any, overload
+
+from pydantic import BaseModel
+
+from ._collection_type import CollectionType
+
+
+class CollectionInfo(BaseModel):
+    """Information about a single Butler collection."""
+
+    name: str
+    """Name of the collection."""
+    type: CollectionType
+    """Type of the collection."""
+    doc: str = ""
+    """Documentation string associated with this collection."""
+    children: tuple[str, ...] = tuple()
+    """Children of this collection (only if CHAINED)."""
+    parents: frozenset[str] = frozenset()
+    """Any parents of this collection."""
 
 
 class ButlerCollections(ABC, Sequence):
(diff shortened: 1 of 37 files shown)


python/lsst/daf/butler/_butler_collections.py
```diff
@@ -27,15 +27,59 @@
...
+    @overload
+    def __getitem__(self, index: slice) -> Sequence[str]: ...
+
+    def __getitem__(self, index: int | slice) -> str | Sequence[str]:
+        return self.defaults[index]
+
+    def __len__(self) -> int:
+        return len(self.defaults)
+
+    def __eq__(self, other: Any) -> bool:
```
timj: This is only here because we have some tests that compare collections directly. It's probably best to remove this method and change the tests to compare `butler.collections.defaults` directly.

python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -2154,7 +2154,12 @@ def collection_chains(self) -> DirectButlerCollections:
         return DirectButlerCollections(self._registry)
 
     @property
-    def collections(self) -> Sequence[str]:
+    def collections(self) -> DirectButlerCollections:
+        """Object with methods for modifying collection chains."""
+        return DirectButlerCollections(self._registry)
+
+    @property
+    def _collections(self) -> Sequence[str]:
```
timj: I will remove the `_collections` variants before merging.

python/lsst/daf/butler/_butler_collections.py
```diff
@@ -27,15 +27,59 @@
...
+    name: str
+    """Name of the collection."""
+    type: CollectionType
+    """Type of the collection."""
+    doc: str = ""
+    """Documentation string associated with this collection."""
+    children: tuple[str, ...] = tuple()
+    """Children of this collection (only if CHAINED)."""
+    parents: frozenset[str] = frozenset()
+    """Any parents of this collection."""
```
TallJimbo: We should probably make this `frozenset[str] | None` so we can distinguish between "has no parents" and "we haven't looked up the parents".

python/lsst/daf/butler/_butler_collections.py
TallJimbo: Yes, please change those tests, and they probably need to wrap `butler.collections.defaults` in `list` or something first, too.  I think anything doing `==` tests on this in the past was already asking for trouble, because it was annotated as `Sequence[str]` and different sequence types (e.g. `list` and `tuple`) do not compare as equal even when they have the same contents.

python/lsst/daf/butler/_butler_collections.py
```diff
@@ -165,3 +209,124 @@ def remove_from_chain(
...
+    @abstractmethod
+    def get_info(self, name: str, include_doc: bool = False, include_parents: bool = False) -> CollectionInfo:
+        """Obtain information for a specific collection.
+
+        Parameters
+        ----------
+        name : `str`
+            The name of the collection of interest.
+        include_doc : `bool`, optional
+            If `True` any documentation about this collection will be included.
```
TallJimbo: I'd at least default `include_doc` to `True`, and maybe even remove the option.

(`include_parents` should stay and default to `False`; that's potentially a much more expensive lookup).

python/lsst/daf/butler/_butler_collections.py
```diff
@@ -165,3 +209,124 @@ def remove_from_chain(
...
+        include_parents : `bool`, optional
+           If `True` any parents of this collection will be included.
+
```
TallJimbo, line 337: This might be something best deferred to DM-43157, but I wonder if adding a
```
include_dataset_types: bool = False
```
argument and a
```
dataset_types: frozenset[str] | None
```
attribute to `CollectionInfo` would be a good convenience interface for the summary information.  I don't think it's the right interface for similar queries that would go beyond what's in the summaries (e.g. counts of datasets by type), since we wouldn't want to query for more info one collection at a time, but in the spirit of this `butler.collections` being the easy way to do easy things, I think it'd fit nicely.

python/lsst/daf/butler/_butler_collections.py
```diff
@@ -165,3 +209,124 @@ def remove_from_chain(
...
+        Notes
+        -----
+        The order in which collections are returned is unspecified, except that
+        the children of a `~CollectionType.CHAINED` collection are guaranteed
+        to be in the order in which they are searched.  When multiple parent
+        `~CollectionType.CHAINED` collections match the same criteria, the
+        order in which the two lists appear is unspecified, and the lists of
+        children may be incomplete if a child has multiple parents.
+        """
+        raise NotImplementedError()
```
TallJimbo, line 320: With the interfaces here now, I think we should expect this pattern to be pretty common:
```
c = butler.collections
for collection in c.query(<args>):
    info = c.get_info(c)
    <do something with info>
```
That's okay, but we should think about whether we'd also want to support `query_info` or `query().info()` or something, to iterate over `CollectionInfo` objects directly (querying for collection names could even just delegate to that).  That's not just about convenience: it also gives us the opportunity to fetch the optional bits of `CollectionInfo` once up front rather than in the loop body, and it makes caching of the non-optional bits less reliant on the user knowing to pull out and reuse the property value.

The fact that this is a specific case of a more a more general question - how we want users to iterate over collection name + something else (where the "something else" might involve COUNT queries on datasets, which we definitely don't want to do one collection or dataset type at a time) is the main reason I'd like to leave the `x_` prefix in place for now.

python/lsst/daf/butler/_butler_collections.py
```diff
@@ -165,3 +209,124 @@ def remove_from_chain(
...
+
+    @abstractmethod
+    def x_remove(self, name: str) -> None:
+        """Remove the given collection from the registry.
+
+        Parameters
+        ----------
+        name : `str`
+            The name of the collection to remove.
```
TallJimbo, line 381: If we make this variadic, and change the behavior for `RUN` collections to delete their datasets artifacts from storage as well (something the `Registry` variant couldn't do, inconveniently), then this can replace `Butler.removeRuns` as well.

python/lsst/daf/butler/_butler_collections.py
```diff
@@ -165,3 +209,124 @@ def remove_from_chain(
...
+        Raises
+        ------
+        lsst.daf.butler.registry.MissingCollectionError
+            Raised if no collection with the given name exists.
+        sqlalchemy.exc.IntegrityError
+            Raised if the database rows associated with the collection are
+            still referenced by some other table, such as a dataset in a
+            datastore (for `~CollectionType.RUN` collections only) or a
+            `~CollectionType.CHAINED` collection of which this collection is
+            a child.
```
TallJimbo, line 392: We need to catch and reraise the SQLAlchemy error in order to have any chance at consistency between `DirectButler` and `RemoteButler`.

In fact, this seems like a good place to pull some of the CLI script logic into the Butler itself, by providing options for breaking parent relationships where necessary instead of just raising.

python/lsst/daf/butler/script/transferDatasets.py
```diff
@@ -79,12 +78,12 @@ def transferDatasets(
     dest_butler = Butler.from_config(dest, writeable=True)
 
     dataset_type_expr = dataset_type or ...
-    collections_expr: tuple[str, ...] | EllipsisType = collections or ...
+    collections_expr: tuple[str, ...] = collections or tuple("*")
```
TallJimbo:
```suggestion
    collections_expr: tuple[str, ...] = collections or ("*",)
```
It amounts to the same thing here, because it's a single character, but this makes it look less like a mistake.

TallJimbo (approved):

python/lsst/daf/butler/_butler_collections.py
timj: `x_query_info` helps a lot -- it simplified the code in the scripts. Good idea.

python/lsst/daf/butler/_butler_collections.py
timj: Removed include_doc.

python/lsst/daf/butler/_butler_collections.py
timj: Changed default to None.

python/lsst/daf/butler/_butler_collections.py
timj: I think I can come back to that on a later ticket. Currently ButlerCollections doesn't always have a butler. I do like the idea of moving removeRuns into collections.

python/lsst/daf/butler/_butler_collections.py
```diff
@@ -27,15 +27,73 @@
...
+    """Any parents of this collection.
+
+    `None` if the parents were not requested.
+    """
+    dataset_types: frozenset[str] | None = None
+    """Names of any dataset types associated with datasets in this collection.
+
+    `None` if no dataset type information was requested
+    """
+    governors: dict[str, frozenset[str]] | None = None
```
timj: @TallJimbo  is there a reason why governors should not be included since it comes "for free"?

python/lsst/daf/butler/script/queryDatasets.py
```diff
@@ -214,16 +213,32 @@ def getDatasets(self) -> Iterator[DatasetRef]:
...
+            # the summary information so we can pre-filter. With only one
+            # let query.datasets work it out.
+            include_summary = True if len(dataset_types) > 1 else False
+            query_collections_info = self.butler.collections.x_query_info(
+                query_collections, include_summary=include_summary
+            )
+            query_collections = [c.name for c in query_collections_info]
+
+            # Only iterate over dataset types that are relevant for the query.
+            if include_summary:
```
timj: @TallJimbo This change makes a huge difference to the "all dataset types in collection X" variant for `butler query-dataset-types`. Easily noticeable in ci_hsc_gen3 comparing the original query system version with this implementation. It doesn't seem to take long for the collection summaries to be calculated for ci_hsc so I'm not sure if "n > 1" dataset types is a reasonable cutoff. The code is in a position to report a reason for no results now for dataset types at least.

python/lsst/daf/butler/_butler_collections.py
timj: Added. Also governors, although maybe you were explicitly deciding against governors being something to worry about.

python/lsst/daf/butler/script/queryDatasets.py
timj: This logic is needed elsewhere so I feel like I'm going to have to have an API somewhere that does:

```python
def filter_dataset_types(dataset_types: Iterable[str | DatasetType], collection_info: Iterable[CollectionInfo]) -> set[str]:
    ...
```
Although I can't think where that would go since both `Butler` and `ButlerCollections` seems wrong. Or else we add `collections` to the new `Butler.query_dataset_types` API and filter that way -- because, now that I think about it, another approach here is to ask the collections first for all the possible answers, then convert the dataset type expressions to regexes and filter the collections dataset types directly and then pass the filtered non-regex iterables to the call that converts them to DatasetType (it's unfortunate that the DatasetType objects are dropped on the floor by CollectionInfo in this case though).

python/lsst/daf/butler/_butler_collections.py
TallJimbo: They just haven't proven very useful, except in the context of guessing default instrument values from collections at butler construction.

python/lsst/daf/butler/script/queryDatasets.py
TallJimbo: if it's not much of performance hit, it sounds like it makes sense to pull down summaries even for just one dataset type, for consistency.  I'm surprised that the new logic has a big performance impact - it suggests my intuition for was was expensive in the old system was off.

I do think we'll want something _like_ a `collections` argument to `query_dataset_types`, but for now it sounds like it'd be best to just put this in a package-private free function somewhere until we nail down the rest of the DM-43157 interfaces.

python/lsst/daf/butler/_butler_collections.py
timj: Ok. I was thinking that it might be useful for immediately returning with no results and single reason (for the loop over dataset types case) if it was possible that after doing `results.where("instrument = 'HsC'")` it would be possible to ask the results object to return the governor dimensions involved in the query such that we could immediately say HsC isn't an instrument in this collection. Since we don't have that capability at the moment I will remove governors from the CollectionInfo for now and we can easily add it back later (the information is immediately available without an extra API call).

TallJimbo: Yes, that was my original thinking, too, and it's just that in practice I think the collection sequences we get passed have either the one instrument the user cares about, or none, or all of them, and hence the filter doesn't affect the query.

python/lsst/daf/butler/script/queryDatasets.py
timj: Done. I changed the scripts to use it and even added support for `butler query-dataset-types` command line.

PR #1054, opened 2024-08-13
DM-45752: Support region overlap against points

dhirving:
Region overlap queries can now use points as regions.  Points can be specified
as `region OVERLAPS POINT(ra, dec)`, or by binding an `lsst.sphgeom.LonLat` or
`astropy.coordinates.SkyCoord` value.  (At the moment, this feature is only
available when using client/server Butler.)

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/queries/_expression_strings.py
+++ b/python/lsst/daf/butler/queries/_expression_strings.py
@@ -31,13 +31,14 @@ from collections.abc import Set
 from typing import Literal, NamedTuple, TypeAlias
 
 import astropy.time
+import lsst.sphgeom
 
 from .._exceptions import InvalidQueryError
 from .._timespan import Timespan
 from ..column_spec import ColumnType
 from ..dimensions import DimensionUniverse
 from ..registry.queries.expressions.categorize import ExpressionConstant, categorizeConstant
-from ..registry.queries.expressions.parser import Node, RangeLiteral, TreeVisitor, parse_expression
+from ..registry.queries.expressions.parser import Node, PointNode, RangeLiteral, TreeVisitor, parse_expression
 from ._identifiers import IdentifierContext, interpret_identifier
 from .tree import (
     BinaryExpression,
@@ -239,8 +240,12 @@ class _ConversionVisitor(TreeVisitor[_VisitorResult]):
     def visitParens(self, expression: _VisitorResult, node: Node) -> _VisitorResult:
         return expression
 
-    def visitPointNode(self, ra: _VisitorResult, dec: _VisitorResult, node: Node) -> _VisitorResult:
-        raise NotImplementedError("POINT() function is not supported yet")
+    def visitPointNode(self, ra: _VisitorResult, dec: _VisitorResult, node: PointNode) -> _VisitorResult:
+        ra_value = _get_float_literal_value(ra, node.ra)
+        dec_value = _get_float_literal_value(dec, node.dec)
+
+        lon_lat = lsst.sphgeom.LonLat.fromDegrees(ra_value, dec_value)
+        return _make_literal(lon_lat)
 
     def visitRangeLiteral(
         self, start: int, stop: int, stride: int | None, node: RangeLiteral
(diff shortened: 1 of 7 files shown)


python/lsst/daf/butler/queries/tree/_column_literal.py
```diff
@@ -374,6 +384,23 @@ def make_column_literal(value: LiteralValue) -> ColumnLiteral:
             return DateTimeColumnLiteral.from_value(astropy.time.Time(value))
         case Timespan():
             return TimespanColumnLiteral.from_value(value)
-        case Region():
+        case lsst.sphgeom.Region():
             return RegionColumnLiteral.from_value(value)
+        case lsst.sphgeom.LonLat():
+            return _make_region_literal_from_lonlat(value)
+        case astropy.coordinates.SkyCoord():
+            icrs = value.transform_to("icrs")
```
dhirving, line 392: Is ICRS the right coordinate system?

python/lsst/daf/butler/queries/_expression_strings.py
```diff
@@ -239,8 +240,12 @@ def visitNumericLiteral(self, value: str, node: Node) -> _VisitorResult:
     def visitParens(self, expression: _VisitorResult, node: Node) -> _VisitorResult:
         return expression
 
-    def visitPointNode(self, ra: _VisitorResult, dec: _VisitorResult, node: Node) -> _VisitorResult:
-        raise NotImplementedError("POINT() function is not supported yet")
+    def visitPointNode(self, ra: _VisitorResult, dec: _VisitorResult, node: PointNode) -> _VisitorResult:
+        ra_value = _get_float_literal_value(ra, node.ra)
+        dec_value = _get_float_literal_value(dec, node.dec)
+
+        lon_lat = lsst.sphgeom.LonLat.fromDegrees(ra_value, dec_value)
```
TallJimbo, line 247: Did you notice whether those preexisting docs specifically said the values were supposed to be decimal degrees?  If we don't want to get into handling multiple angular units (to say nothing of `H:M:S` angles), decimal degrees is probably the best choice, but it definitely needs to be documented.

python/lsst/daf/butler/queries/tree/_column_literal.py
```diff
@@ -374,6 +383,12 @@ def make_column_literal(value: LiteralValue) -> ColumnLiteral:
             return DateTimeColumnLiteral.from_value(astropy.time.Time(value))
         case Timespan():
             return TimespanColumnLiteral.from_value(value)
-        case Region():
+        case lsst.sphgeom.Region():
             return RegionColumnLiteral.from_value(value)
+        case lsst.sphgeom.LonLat():
+            vec = lsst.sphgeom.UnitVector3d(value)
+            # Convert the point to a Region by representing it as a zero-radius
+            # Circle.
```
TallJimbo: I wasn't sure whether zero-radius would work, but if it does, great!

python/lsst/daf/butler/queries/tree/_column_literal.py
```diff
@@ -386,9 +387,20 @@ def make_column_literal(value: LiteralValue) -> ColumnLiteral:
...
-            # Convert the point to a Region by representing it as a zero-radius
-            # Circle.
-            region = lsst.sphgeom.Circle(vec)
-            return RegionColumnLiteral.from_value(region)
+            return _make_region_literal_from_lonlat(value)
+        case astropy.coordinates.SkyCoord():
+            icrs = value.transform_to("icrs")
+            ra = icrs.ra.degree
+            dec = icrs.dec.degree
+            lon_lat = lsst.sphgeom.LonLat.fromDegrees(ra, dec)
```
TallJimbo, line 401: `astropy.coordinates.SkyCoord` can be array-valued (which I think means that `icrs.ra.degree` and `icrs.dec.degree` will come back as something numpy-based.  We should probably guard against this to make the exception raised friendlier.

TallJimbo (approved):

python/lsst/daf/butler/queries/_expression_strings.py
dhirving: The only thing it previously said about units was "similar to ADQL syntax", which implies [decimal degrees](https://www.ivoa.net/documents/ADQL/20180112/PR-ADQL-2.1-20180112.html#sec:functions.geom.point).

I had already tweaked the docs to say this:

> You can check for overlap of a region with a point using the ``POINT(ra, dec)`` syntax,
where ``ra`` and ``dec`` are specified as an ICRS sky position in degrees.

Is that a good way to phrase it?

python/lsst/daf/butler/queries/tree/_column_literal.py
dhirving: It seems to, and from the docs it seems to be intended to with slight caveats (e.g. they go into a discussion of an edge case where two zero-radius circles that are very close but not exact might still be considered overlapping.)

python/lsst/daf/butler/queries/_expression_strings.py
TallJimbo: :+1:

PR #1057, opened 2024-08-14
DM-45775: Enable datastore cache for client/server butler

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/remote_butler/_remote_butler.py
+++ b/python/lsst/daf/butler/remote_butler/_remote_butler.py
@@ -195,7 +195,7 @@ class RemoteButler(Butler):  # numpydoc ignore=PR02
             return cache.dimensions
 
     @property
-    def datastore_cache_manager(self) -> AbstractDatastoreCacheManager:
+    def _cache_manager(self) -> AbstractDatastoreCacheManager:
         """Cache manager to use when reading files from the butler."""
         # RemoteButler does not get any cache configuration from the server.
         # Read the Datastore default config (which is a FileDatastore)
@@ -204,7 +204,6 @@ class RemoteButler(Butler):  # numpydoc ignore=PR02
         # defaults for DatastoreConfig no longer include the cache.
         if self._datastore_cache_manager is None:
             datastore_config = DatastoreConfig()
-            self._datastore_cache_manager: AbstractDatastoreCacheManager
             if "cached" in datastore_config:
                 self._datastore_cache_manager = DatastoreCacheManager(
                     datastore_config["cached"], universe=self.dimensions
@@ -295,7 +294,7 @@ class RemoteButler(Butler):  # numpydoc ignore=PR02
             ref,
             _to_file_payload(model),
             parameters=parameters,
-            cache_manager=self.datastore_cache_manager,
+            cache_manager=self._cache_manager,
         )
 
     def _get_file_info(
--- a/python/lsst/daf/butler/remote_butler/_remote_butler.py
+++ b/python/lsst/daf/butler/remote_butler/_remote_butler.py
@@ -56,11 +56,7 @@ from .._exceptions import DatasetNotFoundError
 from .._storage_class import StorageClass, StorageClassFactory
 from .._utilities.locked_object import LockedObject
 from ..datastore import DatasetRefURIs, DatastoreConfig
-from ..datastore.cache_manager import (
-    AbstractDatastoreCacheManager,
-    DatastoreCacheManager,
-    DatastoreDisabledCacheManager,
-)
+from ..datastore.cache_manager import AbstractDatastoreCacheManager, DatastoreCacheManager
 from ..dimensions import DataIdValue, DimensionConfig, DimensionUniverse, SerializedDataId
 from ..queries import Query
 from ..registry import CollectionArgType, NoDefaultCollectionError, Registry, RegistryDefaults
@@ -100,6 +96,12 @@ class RemoteButler(Butler):  # numpydoc ignore=PR02
     cache : `RemoteButlerCache`
         Cache of data shared between multiple RemoteButler instances connected
         to the same server.
+    use_disabled_datastore_cache : `bool`, optional
+        If `True`, a datastore cache manager will be created with a default
+        disabled state which can be enabled by the environment. If `False`
+        a cache manager will be constructed from the default local
+        configuration, likely caching by default but only specific storage
+        classes.
 
     Notes
     -----
@@ -112,6 +114,7 @@ class RemoteButler(Butler):  # numpydoc ignore=PR02
     _cache: RemoteButlerCache
     _registry: Registry
     _datastore_cache_manager: AbstractDatastoreCacheManager | None
+    _use_disabled_datastore_cache: bool
 
     # This is __new__ instead of __init__ because we have to support
     # instantiation via the legacy constructor Butler.__new__(), which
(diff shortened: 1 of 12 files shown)


python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -171,6 +178,25 @@ def dimensions(self) -> DimensionUniverse:
...
+        # defaults for DatastoreConfig no longer include the cache.
+        if self._datastore_cache_manager is None:
+            datastore_config = DatastoreConfig()
+            self._datastore_cache_manager: AbstractDatastoreCacheManager
+            if "cached" in datastore_config:
+                self._datastore_cache_manager = DatastoreCacheManager(
+                    datastore_config["cached"], universe=self.dimensions
+                )
+            else:
+                self._datastore_cache_manager = DatastoreDisabledCacheManager("", universe=self.dimensions)
```
dhirving: What are the mechanics by which the tests are always going through the `"cached" in datastore_config` branch and not going through this `else`... where is the `cached` key coming from?  I don't see it in `configs/datastore.yaml` or `DatastoreConfig`.

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -171,6 +178,25 @@ def dimensions(self) -> DimensionUniverse:
                 cache.dimensions = universe
             return cache.dimensions
 
+    @property
+    def datastore_cache_manager(self) -> AbstractDatastoreCacheManager:
```
dhirving: This should probably be private since there isn't a good reason for a user to touch this and there isn't an equivalent interface in `DirectButler`.

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -171,6 +178,25 @@ def dimensions(self) -> DimensionUniverse:
...
+        # RemoteButler does not get any cache configuration from the server.
+        # Read the Datastore default config (which is a FileDatastore)
+        # and obtain the default cache manager configuration. Overrides will
+        # come from environment variables. This will not work properly if the
+        # defaults for DatastoreConfig no longer include the cache.
+        if self._datastore_cache_manager is None:
+            datastore_config = DatastoreConfig()
+            self._datastore_cache_manager: AbstractDatastoreCacheManager
+            if "cached" in datastore_config:
+                self._datastore_cache_manager = DatastoreCacheManager(
```
dhirving, line 213: Won't this cause out-of-disk kills in services using RemoteButler?  Because this creates a new cache manager for every RemoteButler instance, by default this will create a new cache directory for every request and never clean up.

Even if it chose a single cache directory, I think it's undesirable for services to have caching enabled by default since they usually have very little `/tmp` available, generally won't be looking at the same file twice, and have many Butler instances operating simultaneously in different threads.

One idea I had (in [DM-42066](https://rubinobs.atlassian.net/issues/DM-42066)) was to have instantiations via `Butler()` enable the cache by default, and instantiatations via `LabeledButlerFactory`disable it by default... since the difference in intended use cases between these constructors should map pretty well to when it's a good idea to have the cache or not.

python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -171,6 +178,25 @@ def dimensions(self) -> DimensionUniverse:
...
             return cache.dimensions
 
+    @property
+    def datastore_cache_manager(self) -> AbstractDatastoreCacheManager:
+        """Cache manager to use when reading files from the butler."""
+        # RemoteButler does not get any cache configuration from the server.
+        # Read the Datastore default config (which is a FileDatastore)
+        # and obtain the default cache manager configuration. Overrides will
+        # come from environment variables. This will not work properly if the
+        # defaults for DatastoreConfig no longer include the cache.
```
dhirving: Given this, it might be a good idea to add a test showing that the cache is getting used.

python/lsst/daf/butler/remote_butler/_remote_butler.py
timj: Yes. I had it private to start with but then learned that from `__init__` you can't ask for the dimension universe. I'll come up with two internal names.

python/lsst/daf/butler/remote_butler/_remote_butler.py
timj: It comes from `datastores/fileDatastore.yaml`, hence the comment about the default datastore config has to be a FileDatastore config.

dhirving: Oh that makes sense.  I didn't understand that the individual configs load their defaults starting from the top-level default butler config.

python/lsst/daf/butler/remote_butler/_remote_butler.py
timj: I've been assuming that services will set the relevant environment variables like we are going to be doing for user's notebooks. Turning off caching through environment variables can be done. We have the option of defining a separate cache manager config model that we configure to be disabled (ie keep 0 datasets) which would allow environment variables to automatically turn it on if needed. If we always use the disabled cache manager for the factory we make it impossible for anyone to use the environment to tune.

python/lsst/daf/butler/remote_butler/_remote_butler.py
timj: I don't think FileDatastore tests that the cache has really been populated either. All the tests of the cache manager are tests of the cache manager.  Should be possible to do a butler.get and ask the remote butler for its cache manager to work out the directory it's using and see if there is a file in there (or use an environment variable to override it).

python/lsst/daf/butler/remote_butler/_remote_butler.py
dhirving: I didn't see an environment variable that could be used to disable the cache... which one is it?

It doesn't need to be force-disabled when you instantiate via the factory -- just default disabled if there is no explicit configuration provided.

I don't think it's good to require all consumers to remember to set an environment variable in order for their stuff not to crash out... it would be better if it did the safe thing by default.  If we do want to require everyone to set an envvar then it would be good for this to be accompanied by a Phalanx PR setting it everywhere it is required.

timj: Currently asking for 0 datasets will effectively turn it off but I do need to add a "disabled" mode that can be used to make it explicitly disabled regardless of other settings. 

Sounds like we want to make the RemoteButler take the cache manager as a parameter and in the factory I can create it with a default disabled config but using DatastoreCacheManager so that environment variables can still be configured and respected. If it's not passed in by the constructor the code we have here can trigger so that client users can benefit from the cache.

dhirving: Yeah that sounds OK to me, though I do wonder if we should wait to add the option to enable it for factory-instantiated RemoteButlers until someone asks for it.  I can't think of a situation where this cache in its current form would be useful for a service -- I suspect we might end up writing a different implementation anyway, if someone needs it.

python/lsst/daf/butler/remote_butler/_remote_butler.py
dhirving: Maybe just add a mock to check that the expected cache manager is getting passed down to `get_dataset_as_python_object`?

dhirving (approved):
Looks OK to me

PR #1059, opened 2024-08-16
DM-45431: Update parquet formatter to use can_accept()

erykoff:
This PR additionally adds `can_accept()` support to delegates and the InMemoryDatastore.

## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/datastores/inMemoryDatastore.py
```diff
@@ -399,9 +400,19 @@ def put(self, inMemoryDataset: Any, ref: DatasetRef) -> None:
...
-        self._validate_put_parameters(inMemoryDataset, ref)
+        try:
+            delegate = ref.datasetType.storageClass.delegate()
+        except TypeError:
+            # TypeError is raised when a storage class doesn't have a delegate.
+            delegate = None
+        if not delegate or not delegate.can_accept(inMemoryDataset):
+            inMemoryDataset = ref.datasetType.storageClass.coerce_type(inMemoryDataset)
+
+        if not self.constraints.isAcceptable(ref):
```
timj: Might be a bit more efficient to do this check before the coerce step.

python/lsst/daf/butler/delegates/arrowtable.py
```diff
@@ -28,24 +28,29 @@
...
+from typing import TYPE_CHECKING, Any
 
 import pyarrow as pa
 from lsst.daf.butler import StorageClassDelegate
 from lsst.utils.introspection import get_full_type_name
 from lsst.utils.iteration import ensure_iterable
 
-__all__ = ["ArrowTableDelegate"]
+if TYPE_CHECKING:
+    import pandas
```
timj, line 42: We disable warnings from pandas bad imports in the mypy config so I think this will not be used in the mypy github action check because pandas isn't listed in types.txt.

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -77,6 +77,35 @@ class ParquetFormatter(FormatterV2):
     default_extension = ".parq"
     can_read_from_local_file = True
 
+    def can_accept(self, in_memory_dataset: Any) -> bool:
```
timj, line 81: maybe the formatter can import the storage delegate so this code isn't duplicated? (I haven't checked whether it's completely identical though).

tests/test_parquet.py
```diff
@@ -2212,5 +2162,81 @@ def testRowGroupSizeDataFrameWithLists(self):
         self.assertGreater(row_group_size, 1_000_000)
 
 
+def _checkAstropyTableEquality(table1, table2, skip_units=False, has_bigendian=False):
```
timj, line 2165: Now I wonder if this should replace `assertAstropyTablesEqual` in lsst.daf.butler.tests.utils.

timj (approved):
This looks great. I assume the tests demonstrate that the coercion is bypassed correctly and accuracy is retained?

tests/test_parquet.py
erykoff: I hadn't noticed this!  Is it worth looking into merging these now?

timj: I'm sure your checks are more complete and so we may as well sync up the utility code. Not required now though.

PR #1062, opened 2024-08-21
DM-43157: new query interfaces for dataset types and datasets of multiple types

TallJimbo:
## Checklist

- [ ] ran Jenkins


python/lsst/daf/butler/_butler_dataset_types.py
```diff
@@ -0,0 +1,244 @@
...
+from ._dataset_type import DatasetType
+from ._storage_class import StorageClass
+from .dimensions import DimensionGroup
+
+
+class ButlerDatasetTypes(ABC, Sequence):
+    """Methods for working with the dataset types known to the Butler."""
+
+    @abstractmethod
+    def get(self, name: str) -> DatasetType:
```
timj, line 44: Are we going to immediately deprecate `butler.get_dataset_type` and move to `butler.dataset_type.get()` ?

TallJimbo: I think that'd be best, yes, if we all like `butler.dataset_types` more generally.

python/lsst/daf/butler/queries/_heterogeneous_dataset_results.py
```diff
@@ -0,0 +1,101 @@
...
+    # - 'dimensions' isn't well-defined in general.
+    #
+    # - 'order_by' and 'limit' are hard to implement in the common case where
+    #   we have to run one query for each dimension group.
+    #
+    # - 'where' exists on other result objects because the way they are
+    #   constructed adds context (a dataset search join, some dimensions) that
+    #   can help interpret arguments to 'where'.  That's not generally true
+    #   here, so calling `Query.where(...).all_datasets()` can do anything that
+    #   `Query.all_datasets().where(...)` might be able to do.
```
dhirving, line 101: We might want to have `where` just for consistency.

python/lsst/daf/butler/queries/_heterogeneous_dataset_results.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        raise NotImplementedError()
+
+    # This class intentionally lacks some attributes that are defined on other
+    # QueryResults objects:
+    #
+    # - 'dimensions' isn't well-defined in general.
+    #
+    # - 'order_by' and 'limit' are hard to implement in the common case where
+    #   we have to run one query for each dimension group.
```
dhirving, line 95: I think `limit` is probably desirable since these queries can easily return huge numbers of results.  Should be relatively straightforward to implement if we're executing the queries serially -- you just start with the original limit  and subtract from it for each subsequent query.

dhirving (commented):
Looks OK to me.

andy-slac:
Rebased to current main.

timj:
To avoid future confusion with this open PR, the `_query_all_datasets` method described here was added in #1109.

PR #1063, opened 2024-08-23
DM-45919: Make removeRuns much faster

timj:
## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -1391,11 +1391,19 @@ def removeRuns(self, names: Iterable[str], unstore: bool = True) -> None:
...
             collectionType = self._registry.getCollectionType(name)
             if collectionType is not CollectionType.RUN:
                 raise TypeError(f"The collection type of '{name}' is {collectionType.name}, not RUN.")
-            refs.extend(self._registry.queryDatasets(..., collections=name, findFirst=True))
+            # Work out the dataset types that are relevant.
+            collections_info = self.collections.x_query_info(name, include_summary=True)
+            filtered_dataset_types = self.collections._filter_dataset_types(
+                all_dataset_types, collections_info
+            )
+            with self._query() as query:
```
timj, line 1399: I have no idea whether it helps for this context to be moved up to the loop over `names`.

TallJimbo (approved):
Looks good!  I was very curious what was going on here.

I wonder if stuff like this used to be reasonably fast back when we were aggressively fetching all the summaries at butler startup, and regressed when we changed that.

timj:
The good news is that there doesn't seem to be any other code doing `queryDatasets(...,` and this block can be replaced by your new API later on very easily.

python/lsst/daf/butler/direct_butler/_direct_butler.py
TallJimbo: I don't think that would help, since we're pulling down collection records and summaries by collection already.  There might be some small speedup from putting the `x_query_info` call inside the context manager, if that lets those summaries be shared by both that call and the dataset query.

PR #1065, opened 2024-08-29
DM-45908: Fix client-side HTTP timeouts when communicating with Butler server

dhirving:
Fixed an issue where slow queries would result in `httpx` timeouts on the client side.  The server now sends keep-alives from slow-running queries to prevent a timeout.  The timeout duration has been increased to 2 minutes on the client side -- previously it was 5 seconds.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/remote_butler/server/_server.py
+++ b/python/lsst/daf/butler/remote_butler/server/_server.py
@@ -33,7 +33,6 @@ from collections.abc import Awaitable, Callable
 
 import safir.dependencies.logger
 from fastapi import FastAPI, Request, Response
-from fastapi.middleware.gzip import GZipMiddleware
 from fastapi.staticfiles import StaticFiles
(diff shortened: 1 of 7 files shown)


python/lsst/daf/butler/remote_butler/server/_server.py
```diff
@@ -54,7 +53,6 @@ def create_app() -> FastAPI:
     config = load_config()
 
     app = FastAPI()
-    app.add_middleware(GZipMiddleware, minimum_size=1000)
```
dhirving, line 57: This was preventing the keep-alives from working because it would buffer/compress them and never send them.

I tested it and it turns out that gzipping was slowing down queries by 10% on the RSP -- there's oodles of bandwidth there and apparently the compression overhead is non-trivial.  So I just removed this, since there are  no configuration options to make it only apply to certain content-types or routes.

timj (approved):
Nice. Great test coverage.

PR #1066, opened 2024-08-29
DM-45860: Add property to DimensionGroup to make it easy to get best time/region dim for query

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/dimensions/_group.py
+++ b/python/lsst/daf/butler/dimensions/_group.py
@@ -449,6 +449,31 @@ class DimensionGroup:  # numpydoc ignore=PR02
         order.extend(element for element in self.elements if element not in done)
         return tuple(order)
 
+    def _choose_dimension(self, families: NamedValueAbstractSet[TopologicalFamily]) -> str | None:
+        if len(families) != 1:
+            return None
+        return list(families)[0].choose(self.elements, self.universe).name
+
+    @property
+    def region_dimension(self) -> str | None:
+        """Return the most appropriate spatial dimension to use when looking
+        up a region.
+
+        Returns `None` if there are no appropriate dimensions or more than one
+        spatial family.
+        """
+        return self._choose_dimension(self.spatial)
+
+    @property
+    def timespan_dimension(self) -> str | None:
+        """Return the most appropriate temporal dimension to use when looking
+        up a time span.
+
+        Returns `None` if there are no appropriate dimensions or more than one
+        temporal family.
+        """
+        return self._choose_dimension(self.temporal)
+
     @property
     def spatial(self) -> NamedValueAbstractSet[TopologicalFamily]:
         """Families represented by the spatial elements in this graph."""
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/registry/obscore/_records.py
```diff
@@ -188,6 +189,9 @@ def __call__(self, ref: DatasetRef, context: SqlQueryContext) -> Record | None:
             if timespan.end is not None:
                 t_max = cast(astropy.time.Time, timespan.end)
                 record["t_max"] = t_max.mjd
+        else:
+            record["t_min"] = -math.inf
+            record["t_max"] = math.inf
```
andy-slac: I don't think that SQL in general supports +/-infinity. It might work with the backends that we support, but we probably should add a unit test for it to be sure

python/lsst/daf/butler/dimensions/_group.py
```diff
@@ -449,6 +450,24 @@ def add_to_order(element: DimensionElement) -> None:
         order.extend(element for element in self.elements if element not in done)
         return tuple(order)
 
+    @property
+    def region_dim(self) -> TopologicalRelationshipEndpoint | None:
```
TallJimbo: I'm not a huge fan of "dim" here as I don't think we've used that abbreviation (outside of local variable names) before, but I don't have anything better.  Technically it's a `DimensionElement` that's returned, but `region_dimension_element` is pretty verbose and `region_element` would expect the user to know what "element" in this context, just like `region_endpoint` would expect them to understand `TopologicalRelationshipEndpoint` (even worse).  Maybe `region_provider`?  Up to you; I clearly don't have any solutions.

I do think it might be good to switch this to return the `.name` attribute (i.e. a `str`) for consistency with the rest of the `DimensionGroup` API, and our overall lean towards minimizing appearance of the `DimensionElement` objects in the public API.

python/lsst/daf/butler/dimensions/_group.py
```diff
@@ -449,6 +450,24 @@ def add_to_order(element: DimensionElement) -> None:
...
+        """Return the most appropriate spatial dimension to use when looking
+        up a region.
+        """
+        for family in self.spatial:
+            return family.choose(self.elements, self.universe)
```
TallJimbo: This is almost never going to matter because most dimension groups of interest have only one spatial or temporal family, but I'm a little uncomfortable just returning the first one when there is more than one.  The relevant real-world case is warps, which have `{visit, patch}` dimensions, and I don't know which this would pick in this case (I'm not even positive it will be deterministic; it might be set-iteration-order-dependent).

There is analogous logic in the `DataCoordinate.region` property that computes the intersection of the regions for the two families, which I think makes sense for warps but might not for more unusual cases, and it's not really an option here (returning a sequence of names whose regions need to be intersected later acts against making it easy to use).  I think it might make it best to just have this return `None` if there is more than one family.

TallJimbo (approved):
Lots of text about very minor points, but it's all advisory: I don't have any super strong opinions because there are tradeoffs on all of these points.

python/lsst/daf/butler/dimensions/_group.py
timj: Adopted `region_dimension(self) -> str`

PR #1068, opened 2024-09-04
DM-45872: Make the new query system public

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1465,7 +1465,7 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
         raise NotImplementedError()
 
     @abstractmethod
-    def _query(self) -> AbstractContextManager[Query]:
+    def query(self) -> AbstractContextManager[Query]:
         """Context manager returning a `Query` object used for construction
         and execution of complex queries.
         """
@@ -1549,7 +1549,7 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
         """
         if data_id is None:
             data_id = DataCoordinate.make_empty(self.dimensions)
-        with self._query() as query:
+        with self.query() as query:
             result = (
                 query.where(data_id, where, bind=bind, **kwargs)
                 .data_ids(dimensions)
@@ -1654,7 +1654,7 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
         """
         if data_id is None:
             data_id = DataCoordinate.make_empty(self.dimensions)
-        with self._query() as query:
+        with self.query() as query:
             result = query.where(data_id, where, bind=bind, **kwargs).datasets(
                 dataset_type,
                 collections=collections,
@@ -1738,7 +1738,7 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
         """
         if data_id is None:
             data_id = DataCoordinate.make_empty(self.dimensions)
-        with self._query() as query:
+        with self.query() as query:
             result = (
                 query.where(data_id, where, bind=bind, **kwargs)
                 .dimension_records(element)
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1549,6 +1549,8 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
         """
         if data_id is None:
             data_id = DataCoordinate.make_empty(self.dimensions)
+        if order_by is None:
+            order_by = []
         with self.query() as query:
             result = (
                 query.where(data_id, where, bind=bind, **kwargs)
@@ -1573,6 +1575,8 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
         where: str = "",
         bind: Mapping[str, Any] | None = None,
         with_dimension_records: bool = False,
+        order_by: Iterable[str] | str | None = None,
+        limit: int = 20_000,
         explain: bool = True,
         **kwargs: Any,
     ) -> list[DatasetRef]:
@@ -1584,7 +1588,7 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
             Dataset type object or name to search for.
         collections : collection expression, optional
             A collection name or iterable of collection names to search. If not
-            provided, the default collections are used.  See
+            provided, the default collections are used. Can be a wildcard. See
             :ref:`daf_butler_collection_expressions` for more information.
         find_first : `bool`, optional
             If `True` (default), for each result data ID, only yield one
@@ -1609,6 +1613,12 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
         with_dimension_records : `bool`, optional
             If `True` (default is `False`) then returned data IDs will have
             dimension records.
+        order_by : `~collections.abc.Iterable` [`str`] or `str`, optional
+            Names of the columns/dimensions to use for ordering returned data
+            IDs. Column name can be prefixed with minus (``-``) to use
+            descending ordering.
+        limit : `int`, optional
+            Upper limit on the number of returned records.
         explain : `bool`, optional
             If `True` (default) then `EmptyQueryResultError` exception is
             raised when resulting list is empty. The exception contains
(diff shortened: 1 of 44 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -1584,7 +1588,7 @@ def query_datasets(
             Dataset type object or name to search for.
         collections : collection expression, optional
             A collection name or iterable of collection names to search. If not
-            provided, the default collections are used.  See
+            provided, the default collections are used. Can be a wildcard. See
```
TallJimbo: Since `find_first=True` is now the default, might want to add that this can be a wildcard only if `find_first=False` is passed as well.

python/lsst/daf/butler/_butler.py
```diff
@@ -1573,6 +1575,8 @@ def _query_datasets(
         where: str = "",
         bind: Mapping[str, Any] | None = None,
         with_dimension_records: bool = False,
+        order_by: Iterable[str] | str | None = None,
+        limit: int = 20_000,
```
TallJimbo: The default here should probably be the same as the CLI default, and I think it'd probably be best to make it negative and warn here, too, if a negative limit is hit.  As it is it's too easy to hit the limit and not know it.

python/lsst/daf/butler/script/queryDatasets.py
```diff
@@ -46,21 +45,15 @@
...
-    uri: str | None
-
-
 class _Table:
     """Aggregates rows for a single dataset type, and creates an astropy table
     with the aggregated data. Eliminates duplicate rows.
     """
 
-    datasetRefs: set[_RefInfo]
+    datasetRefs: dict[DatasetRef, str | None]
```
TallJimbo, line 53: Is the intent here to drop rows where a ref is associated with more than one URI?  I think that's the effect of this change.  If that's undesirable I think we could safely switch to `list[_RefInfo]` instead of `set[_RefInfo]` now.

python/lsst/daf/butler/script/queryDatasets.py
```diff
@@ -111,26 +104,29 @@ def getAstropyTable(self, datasetTypeName: str) -> AstropyTable:
...
-            if refInfo.uri:
-                row.append(refInfo.uri)
+            if uri:
+                row.append(uri)
             rows.append(row)
 
         dataset_table = AstropyTable(np.array(rows), names=columnNames, dtype=columnTypes)
-        return sortAstropyTable(dataset_table, dimensions, ["type", "run"])
+        if sort:
+            return sortAstropyTable(dataset_table, dimensions, ["type", "run"])
```
TallJimbo, line 127: I had no idea we had a separate sorting code path for `order_by` on the CLI that did it all in memory, and I'm not thrilled by that, in terms of the maintenance load and keeping behavior consistent across the two code paths so we have the freedom to change the implementation behind the scenes (it already looks inconsistent in at least not supporting descending sorts with `-identifier` syntax).  I also have no idea why it's trying to sort by spatial columns (what does that even mean?) and the list of dimensions it takes is not the right way to work out spatial/temporal things (which may be dimension elements, not dimensions).

I do think it might make sense to do some in-memory sorting to make output order deterministic when `order_by` is not given, but that's a lot simpler that reimplementing `order_by` in Python.

Fixing all that in the existing CLI commands is out of scope here, but I think I'd prefer we not add new functionality that depends on it here.  For now, could we support `--order-by` in `query-datasets` only on single-dataset-type queries (which we can do in the DB), and revisit it for multiple-dataset-type queries when we have the new Python methods for that?

python/lsst/daf/butler/script/queryDatasets.py
```diff
@@ -225,30 +239,59 @@ def getDatasets(self) -> Iterator[DatasetRef]:
...
+                **kwargs,
+            )
+            if not unlimited:
+                limit -= len(results)
+                if warn_limit and limit == 0 and results:
+                    # We asked for one too many so must remove that from
+                    # the list.
+                    results.pop(-1)
+            _LOG.debug("Got %d results for dataset type %s", len(results), dt)
+            yield from results
```
TallJimbo: Are we actually getting anything out of lazy iteration here, or do all callers eventually just put the results into regular in-memory container anyway?  For at least the case where we turn it into an Astropy table for formatting, I'm sure it'd be better to pass a `list[DatasetRef]` so Astropy can know the number of rows in advance and allocate up front accordingly.

python/lsst/daf/butler/script/queryDatasets.py
```diff
@@ -225,30 +239,59 @@ def getDatasets(self) -> Iterator[DatasetRef]:
...
+
+            if not unlimited and limit == 0:
+                if warn_limit:
+                    _LOG.warning(
+                        "Requested limit of %d hit for number of datasets returned. "
+                        "Use --limit to increase this limit.",
+                        self._limit,
+                    )
```
TallJimbo, line 312: I think we should move this warn-if-limit-hit logic into the Python API, even if we have to do it here too (at least until we implement multiple dataset type queries as a new Python interface).  That could be another ticket, of course.

TallJimbo (approved):

python/lsst/daf/butler/_butler.py
timj: Ok. I can move the negative limit support into the wrappers and out of the script (and also use -20_000 in both places). At some point we can have the conversation as to whether the actual limit support in the advanced query system should understand negative limit.

python/lsst/daf/butler/_butler.py
timj: Although as written it doesn't do that because I expand the collections before passing to the query system. I need to take another look. Thanks for the reminder.

python/lsst/daf/butler/script/queryDatasets.py
timj: Good point. I thought it was for handling the deduplication of results (and I now need to retain the returned sorting order) but it also needs to support disassembled composites so I need to check that by querying pipelines_check.

python/lsst/daf/butler/script/queryDatasets.py
timj: It's only ever sorting each table by dataset type and not trying to sort across dataset types. Nate P was trying to make it so that the results were returned in a consistent way each time and it didn't seem to be a huge issue given people shouldn't be querying for millions of rows on the command line.  If the user specifies `--order-by` no sorting happens here so I think that's what you are asking for and if someone asks to order by with a dimension that is not part of the dataset type the query breaks before any of this code is hit.

python/lsst/daf/butler/script/queryDatasets.py
timj: Yielding seemed nice given the looping over dataset types meaning that we don't require we keep everything in memory. I have been wondering why the command line doesn't try to dump the tables for each dataset type as they arrive from the system rather than waiting until the queries complete. If it yielded `list[DatasetRef]` for each dataset type then that would be more flexible.

python/lsst/daf/butler/script/queryDatasets.py
timj: Ok. I can remove this warning completely here if the python API is going to handle it.

python/lsst/daf/butler/_butler.py
timj: I just tried it without me expanding the collections and the new query system doesn't look like it handles wildcards at all. I'm going to have to query the collections I'm given and then check if the query returns something different to what I started with and then complain if find_first.

python/lsst/daf/butler/script/queryDatasets.py
timj: It was working already because the dataset type name included the component and there was a table per component. I've rewritten the other code that calls this to generate a table as data arrive rather than accumulating in memory and now `butler query-datasets` starts sending output as soon as the first dataset type results arrives.

python/lsst/daf/butler/script/queryDatasets.py
timj: I made this change to yield the results of each dataset type and then yield each table immediately. Much more responsive.

python/lsst/daf/butler/script/queryDatasets.py
timj: I added the warning into `Butler.query_datasets` but had to leave the warning here because the context is different. If you ask for 1000 results spread over 20 dataset types you want to know the limit is 1000 but the internal query only thinks the limit was 100 or whatever.

python/lsst/daf/butler/script/queryDatasets.py
TallJimbo: What about multiple refs from chained datastores?

timj: Ok. I have no means of testing this but I do have a fix that I made that stored lists of URIs that I can resurrect (I deleted it when I convinced myself this was only disassembled composites).

timj: Ok. I  took a deeper look at this:

* `butler.get_many_uris` only returns one URI per ref so there can't be multiple URIs in this table.
* There is a chaining test that was working by luck and assuming the second datastore would return results. This was luck because there was a bug in the test in that the in-memory datastore was not being used because the butler instance got remade -- fixing that bug causes in-memory dataset URIs to be reported.
* I need to fix get_many_uris so that it won't return ephemeral URIs if permanent ones exist.
* I probably can remove my patch to support multiple URIs since it won't happen.

PR #1069, opened 2024-09-04
DM-45791: Put ingest reference time back in the correct place in test

timj:
Commit 4dbfaa989 moved the reference time calculation to after the ingest.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/tests/_registry.py
+++ b/python/lsst/daf/butler/registry/tests/_registry.py
@@ -2722,8 +2722,8 @@ class RegistryTests(ABC):
     def testIngestTimeQuery(self):
         butler = self.make_butler()
         registry = butler._registry
-        self.load_data(butler, "base.yaml", "datasets.yaml")
         dt0 = datetime.datetime.now(datetime.UTC)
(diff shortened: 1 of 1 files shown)


python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -2722,8 +2722,8 @@ def testSkipCalibs(self):
     def testIngestTimeQuery(self):
         butler = self.make_butler()
         registry = butler._registry
-        self.load_data(butler, "base.yaml", "datasets.yaml")
         dt0 = datetime.datetime.now(datetime.UTC)
+        self.load_data(butler, "base.yaml", "datasets.yaml")
```
timj, line 2726: I could split the load_data up into two calls but I'm not sure it makes a difference for this test.

andy-slac: I think it's OK for this test to have `dt0` before loading everuthing.

andy-slac (approved):
Looks great.

PR #1072, opened 2024-09-06
DM-46173: Allow negative values for ra and dec

dhirving:
Fix an issue where ra and dec values in POINT() query expressions could not be negative.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/queries/_expression_strings.py
+++ b/python/lsst/daf/butler/queries/_expression_strings.py
@@ -365,5 +365,7 @@ def _get_float_literal_value(value: _VisitorResult, node: Node) -> float:
             return expr.value
         elif expr.expression_type == "int":
             return float(expr.value)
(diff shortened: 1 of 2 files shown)


timj (approved):
Looks good. Thanks. Should we add a small test to `tests/test_exprParserYacc.py`?

dhirving:
> Should we add a small test to `tests/test_exprParserYacc.py`?

There's already coverage for this syntax in there: https://github.com/lsst/daf_butler/blob/tickets/DM-45933/tests/test_exprParserYacc.py#L165

timj:
Ok. I wasn't sure if that part of the code testing for POINT syntax was also responsible for checking for the negative value.

dhirving:
> Ok. I wasn't sure if that part of the code testing for POINT syntax was also responsible for checking for the negative value.

As far as the grammar is concerned, POINT() is an arbitrary function and its parameters can be any expressions.

PR #1073, opened 2024-09-07
DM-45860: Do not do the collections query again unless we have to

timj:
@andy-slac does this help?

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/utils.py
+++ b/python/lsst/daf/butler/utils.py
@@ -35,7 +35,7 @@ import fnmatch
 import functools
 import logging
 import re
-from collections.abc import Callable
+from collections.abc import Callable, Iterable
 from re import Pattern
 from types import EllipsisType
 from typing import Any, TypeVar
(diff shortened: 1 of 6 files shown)


timj:
@andy-slac I also realized that I had forgotten to add negative limit support to the dimension-records and data-ids command-lines so I've made that change here. Did the collections fix help you?

python/lsst/daf/butler/utils.py
```diff
@@ -134,6 +134,33 @@ def globToRegex(expressions: str | EllipsisType | None | list[str]) -> list[str
     return results
 
 
+def has_globs(expressions: str | EllipsisType | None | Iterable[str]) -> bool:
+    """Determine if the expressions have any glob wildcard characters.
+
+    If a stand-alone '``*``' is found in ``expressions``, or expressions is
+    empty or `None`, then the special value ``...`` will be returned,
+    indicating that any string will match.
```
andy-slac: This looks like it's copied from `globToRegex` but it probably does not apply here?

python/lsst/daf/butler/utils.py
```diff
@@ -134,6 +134,33 @@ def globToRegex(expressions: str | EllipsisType | None | list[str]) -> list[str
...
+
+    Parameters
+    ----------
+    expressions : `str` or `list` [`str`]
```
andy-slac: `Iterable` instead of `list`?

andy-slac (approved):
Looks great.

andy-slac:
> Did the collections fix help you?

Not really, there is a deeper issue when you pass a list of all collection names (mixing all chained and non-chained) to `resolve_wildcard` method, it blows up by flattening all chains again.

PR #1074, opened 2024-09-09
DM-46129: Make collections.query_info a single HTTP call

dhirving:
Previously `RemoteButlerCollections.query_info` would make an HTTP call for each resolved collection.  It now does everything in a single call to the server instead.

`RemoteButlerCollections.get_info` and other collection methods in `RemoteButlerRegistry` have also been updated to re-use the new endpoint.

Also fixed an issue where the `summary_datasets` parameter to `DirectButlerCollections.query_info` was not doing anything.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler_collections.py
+++ b/python/lsst/daf/butler/_butler_collections.py
@@ -45,6 +45,9 @@ if TYPE_CHECKING:
 class CollectionInfo(BaseModel):
     """Information about a single Butler collection."""
 
+    # This class is serialized for the server API -- any new properties you add
+    # must have default values provided to preserve backwards compatibility.
+
     name: str
     """Name of the collection."""
     type: CollectionType
--- a/python/lsst/daf/butler/_butler_collections.py
+++ b/python/lsst/daf/butler/_butler_collections.py
@@ -283,7 +283,7 @@ class ButlerCollections(ABC, Sequence):
         include_parents: bool = False,
         include_summary: bool = False,
         include_doc: bool = False,
-        summary_datasets: Iterable[DatasetType] | None = None,
+        summary_datasets: Iterable[DatasetType] | Iterable[str] | None = None,
     ) -> Sequence[CollectionInfo]:
         """Query the butler for collections matching an expression and
         return detailed information about those collections.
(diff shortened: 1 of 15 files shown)


andy-slac:
DM-45993 was merged, please rebase and add forwarding of new `query_info` parameters to remote server.

python/lsst/daf/butler/_dataset_type.py
```diff
@@ -796,3 +796,21 @@ def _unpickle_via_factory(factory: Callable, args: Any, kwargs: Any) -> DatasetT
     arguments as well as positional arguments.
     """
     return factory(*args, **kwargs)
+
+
+def get_dataset_type_name(datasetTypeOrName: DatasetType | str) -> str:
```
andy-slac, line 801: Maybe make it a classmethod of `DatasetType` to avoid extra imports?

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -12,6 +12,7 @@
 import sqlalchemy
 
 from ...._dataset_ref import DatasetId, DatasetIdGenEnum, DatasetRef, DatasetType
+from ...._dataset_type import get_dataset_type_name
```
andy-slac: `DatasetType` above should also be coming from `_dataset_type`.

python/lsst/daf/butler/_butler_collections.py
```diff
@@ -280,7 +283,7 @@ def query_info(
         include_parents: bool = False,
         include_summary: bool = False,
         include_doc: bool = False,
-        summary_datasets: Iterable[DatasetType] | None = None,
+        summary_datasets: Iterable[DatasetType] | Iterable[str] | None = None,
```
andy-slac, line 286: Should we allow `Iterable[str]` only to avoid all complications? I think that clients of this method that actually pass `summary_datasets` should already have a list of names for other purposes. I guess this also means that datasets manager has to accept `Iterable[str]` instead of `Iterable[DatasetType]`?

python/lsst/daf/butler/_dataset_type.py
```diff
@@ -796,3 +796,21 @@ def _unpickle_via_factory(factory: Callable, args: Any, kwargs: Any) -> DatasetT
...
+    """Given a `DatasetType` object or a dataset type name, return a dataset
+    type name.
+
+    Parameters
+    ----------
+    datasetTypeOrName : `DatasetType` | `str`
+        A DatasetType, or the name of a DatasetType. This union is a common
+        parameter in many `Butler` methods.
+    """
```
andy-slac, line 814: Docstring also needs `Returns`.

andy-slac (approved):
Looks great. Thanks for adding a unit test for new `query_info` parameters! Maybe we should try to simplify things by passing dataset type names instead of dataset types, but that is up to you.

python/lsst/daf/butler/_butler_collections.py
dhirving: I was thinking the same thing, but the convention in the rest of the new query system is that dataset types can be specified as either the name or `DatasetType` instance so it makes sense to follow that.  It's also common for this parameter to come from the result of a queryDatasetTypes call.

PR #1075, opened 2024-09-09
DM-45993: Optimize DirectButlerCollections.query_info to avoid too many queries

andy-slac:
Direct butler reimplements `query_info` method to avoid multiple queries, which makes it significantly faster. This patch also adds two optional parameters to `query_info` to allow further optimizations. There is still an inefficiency in `fetch_summaries` method when the number of potential collections is very large (when collections are `*`). Further optimization would probably need more work (and I think that we'll have to optimize it as the number of collections grows every day).

@dhirving, I added the same parameters to remote butler interface, but they are not used for now. I know you are working on DM-46129, maybe you can add forwarding of those parameters to remote server?

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler_collections.py
+++ b/python/lsst/daf/butler/_butler_collections.py
@@ -31,11 +31,14 @@ __all__ = ("ButlerCollections", "CollectionInfo")
 
 from abc import ABC, abstractmethod
 from collections.abc import Iterable, Sequence, Set
-from typing import Any, overload
+from typing import TYPE_CHECKING, Any, overload
 
 from pydantic import BaseModel
 
 from ._collection_type import CollectionType
+
+if TYPE_CHECKING:
+    from ._dataset_type import DatasetType
 
 
 class CollectionInfo(BaseModel):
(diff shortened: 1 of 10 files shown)


dhirving:
Sounds good.  I'll rebase #1074 after this merges and add the new parameters.

andy-slac:
@timj, I have added a new private method to `Butler.collections` and changed `query-data-ids`, `query-datasets`, and `query-dimension-records` to use it.

python/lsst/daf/butler/_butler_collections.py
```diff
@@ -411,3 +425,30 @@ def _filter_dataset_types(
...
+    ) -> Mapping[str, list[str]]:
+        """Filter dataset types and collections names based on summary in
+        collecion infos.
+
+        Parameters
+        ----------
+        dataset_types : `~collections.abc.Set` [`str`]
+            Set of dataset type names to extract.
+        collection_infos : `~collections.abc.Iterable` [`CollectionInfo`]
+            Colelction infos, must contain dataset type summary.
```
timj:
```suggestion
            Collection infos, must contain dataset type summary.
```

timj (approved):
Looks great. Thanks for unifying the logic in query-dimension-record and data-ids as well.
It makes sense not to include the doc strings in queries by default.

PR #1077, opened 2024-09-12
DM-46298: Make Butler.clone public

dhirving:
`Butler._clone` is now public as `Butler.clone()`.  Unlike the original private version, it copies default values from the original instance unless they are explicitly overridden.

## Checklist

- [ ] ran Jenkins


dhirving:
wrong branch name

PR #1079, opened 2024-09-13
DM-46340: Fix query datasets failing when run collection specified

dhirving:
## Checklist

- [x] ran Jenkins


TallJimbo (approved):
Looks great.  I'm glad casting the literals worked, because it would have been a pain to try to make the code that applies `any_value` aware of what is and is not a literal.

PR #1080, opened 2024-09-16
DM-46339: fix regression in glob detection in query-datasets

TallJimbo:
CHAINED collections could masquerade as globs with the previous logic. With the new logic, we're probably trying to turn each string into a regex more than one time, but that shouldn't be very expensive.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/script/queryDatasets.py
+++ b/python/lsst/daf/butler/script/queryDatasets.py
@@ -36,6 +36,7 @@ from astropy.table import Table as AstropyTable
 
 from .._butler import Butler
 from ..cli.utils import sortAstropyTable
+from ..utils import has_globs
 
 if TYPE_CHECKING:
     from lsst.daf.butler import DatasetRef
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/script/queryDatasets.py
```diff
@@ -261,7 +262,7 @@ def getDatasets(self) -> Iterator[list[DatasetRef]]:
             summary_datasets=dataset_types,
         )
         expanded_query_collections = [c.name for c in query_collections_info]
-        if self._find_first and set(query_collections) != set(expanded_query_collections):
+        if self._find_first and any(has_globs(c) for c in query_collections):
```
timj:
```suggestion
        if self._find_first and has_globs(query_collections):
```
has_globs takes an iterable.

timj (approved):
One real change requested but I'll make approved anyhow on the assumption it will be changed.

PR #1084, opened 2024-09-25
DM-45844: Avoid creating identical parsers

andy-slac:
Building parsers is expensive, now `ParserYacc` keeps a cache of parsers,
there is one parser created per combination of keyword parameters. In
reality there will be just one parser because we do not pass any non-default
parameters to constructor. This makes repeated construction or `ParserYacc`
significantly faster, timing shows reduction from 6.6 sec to 0.3 msec per
1000 instantiations.

This commit removes option for identifier substitution at parser level (`idMap`
parameter for `ParserYacc`). We never used that option, its presence made
reusable parsers hard to implement. We handle identifiers in our visitor classes,
so there is no impact on our code. The only use of that option in the unit tests
was also removed.

Also added type annotations to parserLex and parserYacc modules.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/queries/expressions/parser/parser.py
+++ b/python/lsst/daf/butler/registry/queries/expressions/parser/parser.py
@@ -28,7 +28,7 @@
 from __future__ import annotations
 
 from .exprTree import Node
-from .parserYacc import ParserYacc  # type: ignore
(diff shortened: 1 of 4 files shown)


dhirving (approved):
Looks good to me

PR #1085, opened 2024-09-30
DM-46347: Fix issue where default data IDs were not being applied

dhirving:
Fixed two issues where the default data ID was not constraining the result of queries:

* For default data IDs required because of references in the `where` string, it was not actually using the data ID to constrain the result  -- it was only being applied to pre-filtering of dataset searches based on summaries.
* For cases where the governor dimension was included in the query result without a reference in the `where` clause, it was not applying the default data ID at all.

## Checklist

- [ ] ran Jenkins


python/lsst/daf/butler/tests/butler_queries.py
```diff
@@ -1892,6 +1892,79 @@ def test_multiple_instrument_queries(self) -> None:
...
+        # constrain the instrument.
+        # In this case, 'instrument' is not part of the dimensions returned by
+        # the query, so there is extra logic needed to detect the need for the
+        # default data ID.
+        data_ids = butler.query_data_ids(["band"], where="physical_filter='Cam1-G'")
+        self.assertEqual([x["band"] for x in data_ids], ["g"])
+        # Default data ID instrument=Cam1 does not match Cam2, so there are no
+        # results.
+        data_ids = butler.query_data_ids(["band"], where="physical_filter='Cam2-G'", explain=False)
+        self.assertEqual(data_ids, [])
```
dhirving, line 1939: I think it might be a bit surprising that no results are returned here, but it's consistent with "always apply the default data ID unless it's explicitly overridden by the user".

python/lsst/daf/butler/direct_query_driver/_predicate_constraints_summary.py
```diff
@@ -0,0 +1,221 @@
...
+
+    messages: list[str]
+    """Diagnostic messages that report reasons the query may not return any
+    rows.
+    """
+
+    _governors_referenced: set[str]
+    """Governor dimensions referenced directly in the predicate, but not
+    necessarily constrained to the same value in all logic branches.
+    """
```
TallJimbo: I find it a little strange stylistically to annotate and document a private attribute out here, rather than just annotating it in `__init__`.

TallJimbo (approved):

PR #1086, opened 2024-09-30
DM-46563: Ensure that arrow tables with null values return masked numpy array

timj:
## Checklist

- [ ] ran Jenkins
- [ ] added a release note for user-visible changes to `doc/changes`
- [ ] (if changing dimensions.yaml) make a copy of dimensions.yaml in `configs/old_dimensions`

PR #1087, opened 2024-09-30
DM-46563: Use masked record arrays in arrow to numpy conversion

timj:
This is #1086 with the right branch name. 

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/formatters/parquet.py
+++ b/python/lsst/daf/butler/formatters/parquet.py
@@ -219,7 +219,7 @@ def arrow_to_astropy(arrow_table: pa.Table) -> atable.Table:
     return astropy_table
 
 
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -231,21 +231,28 @@ def arrow_to_numpy(arrow_table: pa.Table) -> np.ndarray:
     -------
     array : `numpy.ndarray` (N,)
         Numpy array table with N rows and the same column names
-        as the input arrow table.
+        as the input arrow table. Will be masked records if any values
+        in the table are null.
```
erykoff, line 235: I guess the return type should be `np.ndarray` or `np.ma.MaskedArray`?  Same for the mypy type?

erykoff (approved):
Looks good with test, but a question about the typing.

PR #1089, opened 2024-10-01
DM-46575: Update parquet formatter to use fsspec.

erykoff:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/formatters/parquet.py
+++ b/python/lsst/daf/butler/formatters/parquet.py
@@ -50,6 +50,7 @@ __all__ = (
 import collections.abc
 import itertools
 import json
+import logging
 import re
 from collections.abc import Iterable, Sequence
 from typing import TYPE_CHECKING, Any, cast
@@ -62,10 +63,19 @@ from lsst.resources import ResourcePath
 from lsst.utils.introspection import get_full_type_name
 from lsst.utils.iteration import ensure_iterable
 
+log = logging.getLogger(__name__)
+
 if TYPE_CHECKING:
     import astropy.table as atable
     import numpy as np
     import pandas as pd
+
+    try:
+        import fsspec
+        from fsspec.spec import AbstractFileSystem
+    except ImportError:
+        fsspec = None
+        AbstractFileSystem = type
 
 TARGET_ROW_GROUP_BYTES = 1_000_000_000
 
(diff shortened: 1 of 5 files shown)


tests/test_parquet.py
```diff
@@ -2123,6 +2144,131 @@ class InMemoryArrowSchemaDelegateTestCase(ParquetFormatterArrowSchemaTestCase):
     configFile = os.path.join(TESTDIR, "config/basic/butler-inmemory.yaml")
 
 
+@unittest.skipUnless(pa is not None, "Cannot test S3 without pyarrow.")
+@unittest.skipUnless(boto3 is not None, "Cannot test S3 without boto3.")
+@unittest.skipUnless(fsspec is not None, "Cannot test S3 with fsspec.")
```
timj:
```suggestion
@unittest.skipUnless(fsspec is not None, "Cannot test S3 without fsspec.")
```
Technically it will also need s3fs for the s3 test to work...

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -76,15 +86,37 @@ class ParquetFormatter(FormatterV2):
...
     def can_accept(self, in_memory_dataset: Any) -> bool:
         # Docstring inherited.
         return _checkArrowCompatibleType(in_memory_dataset) is not None
 
+    def read_from_uri(self, uri: ResourcePath, component: str | None = None, expected_size: int = -1) -> Any:
+        # Docstring inherited from Formatter.read.
+        try:
+            fs, path = uri.to_fsspec()
+        except ImportError:
+            log.debug("fsspec not available; falling back to file access.")
```
timj:
```suggestion
            log.debug("fsspec not available; falling back to local file access.")
```
I'm pondering how we can convince the test to run the fallback path. I assume you tested that it does work.

timj (approved):
Looks great. Sorry about the duplication of the s3 test infrastructure.

python/lsst/daf/butler/formatters/parquet.py
erykoff: So it is my understanding that the docker build does not have fsspec or s3fs, and I didn't update that build action.  So I think that does test the fallback path, even if coverage doesn't know.  Also, after I made these changes I pushed to get the GHA actions to pass before I added fsspec to the build yaml.  (But that was a one-off obviously).  I'm not sure how to regularly test both code paths without a lot of additional complications that don't seem warranted.

timj: Yes. That's fine. Good point about the docker tests passing without fsspec.

PR #1090, opened 2024-10-02
DM-46581: Speed up TimeConverter

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_timespan.py
+++ b/python/lsst/daf/butler/_timespan.py
@@ -335,14 +335,23 @@ class Timespan(pydantic.BaseModel):
                 tail = f"{self.end.tai.strftime(fmt)})"
         return head + tail
 
+    def __repr_astropy__(self, t: astropy.time.Time | None) -> str:
+        # Provide our own repr for astropy time.
+        # For JD times we want to use jd1 and jd2 to maintain precision.
+        if isinstance(t, astropy.time.Time):
+            if t.format == "jd":
+                return f"astropy.time.Time({t.jd1}, {t.jd2}, scale='{t.scale}', format='{t.format}')"
+            else:
+                return f"astropy.time.Time('{t}', scale='{t.scale}', format='{t.format}')"
+        return str(t)
+
     def __repr__(self) -> str:
         # astropy.time.Time doesn't have an eval-friendly __repr__, so we
         # simulate our own here to make Timespan's __repr__ eval-friendly.
         # Interestingly, enum.Enum has an eval-friendly __str__, but not an
         # eval-friendly __repr__.
-        tmpl = "astropy.time.Time('{t}', scale='{t.scale}', format='{t.format}')"
-        begin = tmpl.format(t=self.begin) if isinstance(self.begin, astropy.time.Time) else str(self.begin)
-        end = tmpl.format(t=self.end) if isinstance(self.end, astropy.time.Time) else str(self.end)
+        begin = self.__repr_astropy__(self.begin)
+        end = self.__repr_astropy__(self.end)
         return f"Timespan(begin={begin}, end={end})"
 
     def __eq__(self, other: Any) -> bool:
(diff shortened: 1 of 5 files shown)


timj:
This change almost halves the time of `nsec_to_astropy` (and makes `astropy_to_nsec` 4% faster).

python/lsst/daf/butler/_timespan.py
```diff
@@ -335,14 +335,23 @@ def __str__(self) -> str:
                 tail = f"{self.end.tai.strftime(fmt)})"
         return head + tail
 
+    def __repr_astropy__(self, t: astropy.time.Time | None) -> str:
+        # Provide our own repr for astropy time.
+        # For JD times we want to use jd1 and jd2 to maintain precision.
+        if isinstance(t, astropy.time.Time):
+            if t.format == "jd":
+                return f"astropy.time.Time({t.jd1}, {t.jd2}, scale='{t.scale}', format='{t.format}')"
```
timj, line 343: This doesn't trigger with the new code because for Timespan we always convert the Time to nsec internally and then convert back to Time on request and so return a format=unix_tai_fast variant for repr. If we ever decide to simply keep the Timespan we were given then this would trigger again.

python/lsst/daf/butler/time_utils.py
```diff
@@ -136,10 +177,10 @@ def nsec_to_astropy(self, time_nsec: int) -> astropy.time.Time:
         that the number falls in the supported range and can produce output
         time that is outside of that range.
         """
-        jd1, jd2 = divmod(time_nsec, self._NSEC_PER_DAY)
-        delta = astropy.time.TimeDelta(float(jd1), float(jd2) / self._NSEC_PER_DAY, format="jd", scale="tai")
-        value = self.epoch + delta
-        return value
+        jd1, jd2 = divmod(time_nsec, 1_000_000_000)
+        return astropy.time.Time(
+            float(jd1), jd2 / 1_000_000_000, format="unix_tai_fast", scale="tai", precision=6
```
timj, line 182: @andy-slac  We have to pass in seconds here for jd1 rather than days (and so require the divmod call again) because we are subclassing `TimeUnixTai`?

andy-slac: Right, I decided that we want to use TimeUnixTai instead of TimeJD format as a base because this is exactly what we do, and TimeUnix takes seconds. It would still be possible to pass days here, but I was worried it could cause confusion somewhere else.

python/lsst/daf/butler/_timespan.py
```diff
@@ -335,14 +335,23 @@ def __str__(self) -> str:
...
+            else:
+                return f"astropy.time.Time('{t}', scale='{t.scale}', format='{t.format}')"
```
andy-slac, line 345: Hmm, this will now render as:
```
astropy.time.Time('1727969332.139023', scale='tai', format='unix_tai_fast')
```
with value appearing as string. Also I am a bit worried that we expose `unix_tai_fast` which is supposed to be internal.

timj: And yet the tests pass using it. It worked "correctly" before because we were using an isot string format by default by defining `self.epoch` as that in `TimeConverter`.

I was also worried about `unix_tai_fast` leaking out but I tried adding a `.replicate(format="jd")` to the `nsec_to_astropy` method and whilst that worked, it did slow things down by factor 2 from 7 to 14 microseconds. Not sure what to do really. We could force to JD or ISOT format in this repr by calling `.replicate()`. The format will still leak out for anyone who decides to print `ts.begin` anywhere.

timj: Oh, and that quoting isn't new. It was always like that.

andy-slac: Maybe instead of replicate() we can do `time.format = "jd"` inside nsec_to_astropy, this makes it about 25% slower.

timj: Ok. I will push that as a compromise.

timj: Slowed down from 8us to 10us.

PR #1092, opened 2024-10-05
DM-46259: Prevent confusing error when deriving dataId from bad exposure/detector combination

timj:
* Use new query system in direct butler for dimension records.
* When querying for a detector, say, do not include the exposure in the query (if the exposure is bad the detector query will fail with a confusing error message).
* Improve some other error messages if you have "detector=16, raft='0'" and the raft is inconsistent with the detector or the detector is not known at all.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_butler/_direct_butler.py
+++ b/python/lsst/daf/butler/direct_butler/_direct_butler.py
@@ -751,10 +751,9 @@ class DirectButler(Butler):  # numpydoc ignore=PR02
                                         # Use bind to allow zero results.
                                         # This is a fully-specified query.
                                         "visit_system_membership",
-                                        where="instrument = inst AND visit_system = system AND visit = v",
-                                        bind=dict(
-                                            inst=instrument_records[0].name, system=visit_system, v=rec.id
-                                        ),
+                                        instrument=instrument_records[0].name,
+                                        visit_system=visit_system,
+                                        visit=rec.id,
                                         explain=False,
                                     )
                                     if membership:
--- a/python/lsst/daf/butler/direct_butler/_direct_butler.py
+++ b/python/lsst/daf/butler/direct_butler/_direct_butler.py
@@ -61,7 +61,7 @@ from .._dataset_existence import DatasetExistence
 from .._dataset_ref import DatasetRef
 from .._dataset_type import DatasetType
 from .._deferredDatasetHandle import DeferredDatasetHandle
-from .._exceptions import DatasetNotFoundError, DimensionValueError, ValidationError
+from .._exceptions import DatasetNotFoundError, DimensionValueError, EmptyQueryResultError, ValidationError
 from .._limited_butler import LimitedButler
 from .._registry_shim import RegistryShim
 from .._storage_class import StorageClass, StorageClassFactory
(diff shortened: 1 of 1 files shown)


python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -2066,7 +2085,7 @@ def validateConfiguration(
         # Find all the registered instruments (if "instrument" is in the
         # universe).
         if "instrument" in self.dimensions:
-            instruments = {record.name for record in self._registry.queryDimensionRecords("instrument")}
+            instruments = set(rec.name for rec in self.query_dimension_records("instrument", explain=False))
```
andy-slac: Why `set()` instead of `{}`?

python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -728,16 +747,15 @@ def _rewrite_data_id(
...
-                                            where="instrument = inst AND visit_system = system AND visit = v",
-                                            bind=dict(
-                                                inst=instrument_records[0].name, system=visit_system, v=rec.id
-                                            ),
-                                        )
+                                    membership = self.query_dimension_records(
+                                        # Use bind to allow zero results.
+                                        # This is a fully-specified query.
+                                        "visit_system_membership",
+                                        where="instrument = inst AND visit_system = system AND visit = v",
```
andy-slac: Looks somewhat unfortunate that we have to use string `where` which has to be parsed again slowing things down. I think it would be more efficient to use `Predicate` from new query system instead. But can't we in this particular case use `data_id={"instrument": instrument_records[0].name, "system"=visit_system, "visit"=rec.id}` instead of `where`?

andy-slac (approved):
Looks great, a couple of minor questions.

python/lsst/daf/butler/direct_butler/_direct_butler.py
timj: because it was `set(records)` to begin with and then I realized I wanted the names...

python/lsst/daf/butler/direct_butler/_direct_butler.py
timj: Yes. You are right. I will change it over to the data_id since this part is all dimensions (the code above involves dimension metadata). I was also wondering about Predicates but hadn't looked into it (and the SIAv2 code might want to use it as well).

python/lsst/daf/butler/direct_butler/_direct_butler.py
andy-slac: Well, I was asking about change from `{...}` to `set(...)`, they do the same thing?

timj: Yes. I know. Looking at the diff it's obvious that I should go back to how it was before to minimize unnecessary churn.

python/lsst/daf/butler/direct_butler/_direct_butler.py
TallJimbo: Rather than using `Predicate` directly, check out `Query.expression_factory`.

timj: Thanks. I'll take a look for dax_obscore, although I might need a bit of help because at the moment I create the WHERE strings and bind parameters in one place when iterating over all the SIAv2 parameters (and have schemes for combining things with OR and AND), and then pass those to the code that is actually inside the Query context. If the expression factory doesn't have to be inside a query context and just needs a universe that will probably be okay.

PR #1094, opened 2024-10-11
DM-46794: Fix problem with data ID rewriting

timj:
Fixes bug introduced in #1092. Now uses the implied and required dimensions for filtering out unnecessary dimensions rather than just required dimensions. This was preventing the standard day_obs+seq_num queries from working.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_butler/_direct_butler.py
+++ b/python/lsst/daf/butler/direct_butler/_direct_butler.py
@@ -708,21 +708,23 @@ class DirectButler(Butler):  # numpydoc ignore=PR02
                     for k, v in newDataId.items()
                     if k in self.dimensions[dimensionName].minimal_group.names
                 }
-                # Build up a WHERE expression
-                bind = dict(values.items())
-                where = " AND ".join(f"{dimensionName}.{k} = {k}" for k in bind)
 
-                # Hopefully we get a single record that matches
-                records = set(
-                    self.query_dimension_records(
-                        dimensionName,
-                        data_id=filtered_data_id,
-                        where=where,
-                        bind=bind,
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -708,21 +709,22 @@ def _rewrite_data_id(
...
-                        bind=bind,
-                        explain=False,
-                        **kwargs,
-                    )
-                )
+                def _get_attr(obj: Any, attr: str) -> Any:
+                    # Used to implement x.exposure.seq_num when given
+                    # x and "exposure.seq_num".
+                    f = attrgetter(attr)
+                    return f(obj)
```
TallJimbo: Why `attrgetter` instead of the `getattr` free function?  I think of `attrgetter` as being an optimization for the case where you want to get the same attribute on many different objects.  But it's entirely possible there's some other advantage I'm just not aware of.

TallJimbo (approved):

python/lsst/daf/butler/direct_butler/_direct_butler.py
timj: `getattr` doesn't work for `exposure.seq_num` does it? I think I'd have to split on dot and iterate so it seemed simpler to use `attrgetter`. I just did a timing test between the two and a split plus loop with `getattr` took the same amount of time as `attregetter`.

timj: I'll use nested getattr instead on the basis that it's more obvious to people reading the code.

PR #1096, opened 2024-10-11
DM-46599: Deprecate old query system behavior

dhirving:
Deprecate behavior from the old query system that will no longer be supported when we switch to the new implementation.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-46599.removal.md
+++ b/doc/changes/DM-46599.removal.md
@@ -1,0 +1,19 @@
+Regular expressions in collection and dataset type patterns are now deprecated. (Shell-like globs will continue to be supported.)
+
+Materializing dataset queries into temporary tables is now deprecated.  (Materializing data ID queries will continue to be supported.)
+
+The `datasetTypes` argument to `Registry.queryCollections` is now deprecated. (This parameter has never had any effect.)
+
+We will soon stop raising `DataIdValueError` exceptions for typos and other bad values in query expressions like `instrument='HsC'` for typos and other bad values in query expressions.  Instead, these queries will return an empty iterable of results.
+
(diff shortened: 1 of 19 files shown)


doc/changes/DM-46599.removal.md
```diff
@@ -0,0 +1,19 @@
...
+
+We will soon stop raising `DataIdValueError` exceptions for typos and other bad values in query expressions like `instrument='HsC'` for typos and other bad values in query expressions.  Instead, these queries will return an empty iterable of results.
+
+Using HTM and HEALPix spatial dimensions like `htm11` or `healpix10` in data ID constraints passed to queries is now deprecated. The exception is `htm7`, which will continue to work.
+
+The `--no-check` parameter to `butler query-dimension-records` is now deprecated.
+
+The `offset` argument to `limit()` for `Registry.queryDataIds` and `Registry.queryDimensionRecords` result objects is now deprecated.
+
+The `--offset` option for `butler query-data-ids` and `butler-query-datasets` is now deprecated.
```
timj: It looks like `butler query-data-ids` raises when `--offset` is used and `query-dimension-records` ignores it. Maybe we add a raise to the latter as well? Or remove the raise from the former and add deprecation warnings instead noting that the option is a no-op? I think this happened during the migration to the new query system.

timj (approved):
Looks great.

doc/changes/DM-46599.removal.md
dhirving: I didn't realize that these had been switched over to the new query system... theoretically shouldn't have happened until after this deprecation.

Probably better to raise than silently ignore the option, so I'll add an exception to `query-dimension-records`.

PR #1099, opened 2024-10-16
DM-46496: Include advanced query system objects in docs

dhirving:
`Butler.query` has been public for a while, but the documentation for the `Query` object it returns was not included in the documentation.

This change pulls in the user-facing portions of the interface without any of the internal objects.  (The internal objects are sometimes referenced from constructors, but those constructors are documented as "do not use".)

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/queries/tree/__init__.py
+++ b/python/lsst/daf/butler/queries/tree/__init__.py
@@ -35,6 +35,5 @@ from ._predicate import LogicalNot
 from ._query_tree import *
 
 LogicalNot.model_rebuild()
-del LogicalNot
 
 Predicate.model_rebuild()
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/queries/tree/__init__.py
```diff
@@ -35,6 +35,5 @@
 from ._query_tree import *
 
 LogicalNot.model_rebuild()
-del LogicalNot
```
dhirving, line 38: This `del` was crashing sphinx.  This is an internal module that isn't re-exported anywhere or referenced from the documentation so it doesn't super matter if there's an extra symbol in here.

timj: I think it was being deleted so as not to be lifted into the parent on import. Another way to do it might be:

```python
from ._predicated import LogicalNot as _LogicalNot
...
_LogicalNot.model_rebuild()
```
which should have the same effect of not affecting the exports.

timj (approved):

PR #1104, opened 2024-10-22
DM-46479: add low-level support for union queries over multiple dataset types

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_query_driver/_driver.py
+++ b/python/lsst/daf/butler/direct_query_driver/_driver.py
@@ -209,26 +209,26 @@ class DirectQueryDriver(QueryDriver):
         # Docstring inherited.
         if self._exit_stack is None:
             raise RuntimeError("QueryDriver context must be entered before queries can be executed.")
-        _, builder = self.build_query(
+        _, builder, postprocessing = self.build_query(
             tree,
             final_columns=result_spec.get_result_columns(),
             order_by=result_spec.order_by,
             find_first_dataset=result_spec.find_first_dataset,
         )
-        sql_select = builder.select()
+        sql_select = builder.select(postprocessing)
         if result_spec.order_by:
             visitor = SqlColumnVisitor(builder.joiner, self)
             sql_select = sql_select.order_by(*[visitor.expect_scalar(term) for term in result_spec.order_by])
         if result_spec.limit is not None:
-            if builder.postprocessing:
-                builder.postprocessing.limit = result_spec.limit
+            if postprocessing:
+                postprocessing.limit = result_spec.limit
             else:
                 sql_select = sql_select.limit(result_spec.limit)
-        if builder.postprocessing.limit is not None:
+        if postprocessing.limit is not None:
             # We might want to fetch many fewer rows than the default page
             # size if we have to implement limit in postprocessing.
             raw_page_size = min(
-                self._postprocessing_filter_factor * builder.postprocessing.limit,
+                self._postprocessing_filter_factor * postprocessing.limit,
                 self._raw_page_size,
             )
         else:
@@ -238,7 +238,7 @@ class DirectQueryDriver(QueryDriver):
         cursor = _Cursor(
             self.db,
             sql_select,
-            postprocessing=builder.postprocessing,
+            postprocessing=postprocessing,
             raw_page_size=raw_page_size,
             page_converter=self._create_result_page_converter(result_spec, builder),
         )
@@ -292,7 +292,7 @@ class DirectQueryDriver(QueryDriver):
         # Docstring inherited.
         if self._exit_stack is None:
             raise RuntimeError("QueryDriver context must be entered before 'materialize' is called.")
-        _, builder = self.build_query(tree, qt.ColumnSet(dimensions))
+        _, builder, postprocessing = self.build_query(tree, qt.ColumnSet(dimensions))
         # Current implementation ignores 'datasets' aside from remembering
         # them, because figuring out what to put in the temporary table for
         # them is tricky, especially if calibration collections are involved.
@@ -306,12 +306,14 @@ class DirectQueryDriver(QueryDriver):
         #   search is straightforward and definitely well-indexed, and not much
         #   (if at all) worse than joining back in on a materialized UUID.
         #
-        sql_select = builder.select()
-        table = self._exit_stack.enter_context(self.db.temporary_table(builder.make_table_spec()))
+        sql_select = builder.select(postprocessing)
+        table = self._exit_stack.enter_context(
+            self.db.temporary_table(builder.make_table_spec(postprocessing))
+        )
         self.db.insert(table, select=sql_select)
         if key is None:
             key = uuid.uuid4()
-        self._materializations[key] = _MaterializationState(table, datasets, builder.postprocessing)
+        self._materializations[key] = _MaterializationState(table, datasets, postprocessing)
         return key
 
     def upload_data_coordinates(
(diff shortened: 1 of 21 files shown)


python/lsst/daf/butler/nonempty_mapping.py
```diff
@@ -97,3 +98,13 @@ def get(self, key: _K, default: Any = None) -> Any:
...
         return default
+
+    def copy(self) -> NonemptyMapping[_K, _V]:
+        """Return a copy of the mapping.
+
+        This deep-copies values while referencing the ``default_factory``
+        callback and keys.
+        """
+        result = NonemptyMapping[_K, _V](self._default_factory)
+        result._mapping = copy.deepcopy(self._mapping)
```
dhirving: This `deepcopy` seems hazardous if the inner value type is non-trivial.  e.g. this gets called on a  `NonemptyMapping[str, list[sqlalchemy.ColumnElement]]`.

I dunno what the behavior of calling `deepcopy` on a sqlalchemy object is and I'm not sure I want to find out :D  The author of sqlalchemy [has been quoted as saying](https://github.com/sqlalchemy/sqlalchemy/issues/4603#issuecomment-481422077):

>  I think deepcopy is kind of crazy to use ever.

and I can't find any documentation saying it's intended to work.

Maybe we should add more-concrete subclasses for `NonemptyMapping[T, list[S]]`, or just shallow-copy the values to limit the blast radius.

python/lsst/daf/butler/direct_query_driver/_query_analysis.py
```diff
@@ -28,29 +28,37 @@
...
 @dataclasses.dataclass
-class ResolvedDatasetSearch:
+class ResolvedDatasetSearch(Generic[_T]):
     """A struct describing a dataset search joined into a query, after
     resolving its collection search path.
     """
 
-    name: str
-    """Name of the dataset type."""
+    name: _T
```
dhirving, line 59: It took me a while to figure out which types this could have, and under which circumstances it was which, and how that interacted with the function overloads associated with this generic.

I wonder if this could just be `list[str]` instead of needing the generic?

Or maybe two classes like:
```
class SingleResolvedDatasetSearch:
    name: str
    search: ResolvedDatasetSearch
class UnionResolvedDatasetSearch:
    names: list[str]
    search: ResolvedDatasetSearch
```
since it seems like usually we want to know which of the generic cases we have, and when we don't we can pass in the `search` object instead of the whole thing.

python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -454,140 +460,304 @@ def build_query(
...
+                        collection_analysis.collection_records[collection_name]
+                        for collection_name in collection_path
+                    ],
+                    messages=[],
+                )
+                resolved_search.is_calibration_search = any(
+                    r.type is CollectionType.CALIBRATION for r in resolved_search.collection_records
+                )
+                dataset_searches_by_collections[key] = resolved_search
+            resolved_search.name.append(dataset_type_name)
```
dhirving: It can help readability to pull out independent chunks like this to a separate function.   I find that it makes it easier to see the high-level structure of the logic without getting bogged down in the details.  I like to use Right click -> Refactor -> Extract Method in VSCode to experimentally check candidate chunks -- in this case it comes out cleanly as a function of 2 parameters.

python/lsst/daf/butler/direct_query_driver/_sql_builders.py
```diff
@@ -0,0 +1,678 @@
...
+        select0 = self.select(postprocessing)
+        other_selects = [other.select(postprocessing) for other in others]
+        return SqlJoinsBuilder(
+            db=self.joins.db,
+            from_clause=select0.union(*other_selects).subquery(),
+        ).extract_columns(self.columns, postprocessing)
+
+
+@dataclasses.dataclass(kw_only=True)
+class SqlColumns:
```
dhirving, line 300: I like the refactor to pull this stuff out into its own class -- these are logically separate from some of the things they were mixed in with before.  I'd like it even better if SqlJoinsBuilder consumed it by composition instead of inheritance.  (Looks like maybe you were moving that direction and didn't get that far?)

python/lsst/daf/butler/direct_query_driver/_query_builder.py
```diff
@@ -27,536 +27,684 @@
...
-            [
-                self.columns.get_column_spec(logical_table, field).to_sql_spec(
-                    name_shrinker=self.joiner.db.name_shrinker
-                )
-                for logical_table, field in self.columns
-            ]
+        raise NotImplementedError()
+
+
+class SingleSelectQueryBuilder(QueryBuilder):
```
dhirving, line 287: It seems like it might be possible to unify more of this near-duplicate code with `UnionQueryBuilder` by treating this is a union with only one term, though there's a few corners that will make that difficult.

tests/test_pydantic_utils.py
```diff
@@ -137,6 +142,24 @@ def test_time(self) -> None:
         with self.assertRaises(ValueError):
             adapter.validate_json({})
 
+    def test_ellipsis(self) -> None:
+        adapter = pydantic.TypeAdapter(SerializableEllipsis)
+        json_roundtripped = adapter.validate_json(adapter.dump_json(...))
+        self.assertIs(json_roundtripped, ...)
+        python_roundtripped = adapter.validate_python(adapter.dump_python(...))
+        self.assertIs(python_roundtripped, ...)
```
dhirving: There's a missing test case here for `adapter.validate_python(...)` which I think would fail.  This will make it hard to instantiate models using SerializableEllipsis from inside Python rather than from JSON.

dhirving (approved):
I think this is good to merge as-is so we can unblock Andy and I can start working with it.  The refactors all look good and the new code seems to be mostly segregated from existing code paths.

The only comment that I think is a major issue is the `deepcopy` thing, but that code is not currently being executed so it's not immediately harmful.  I can fix it up when I go in there to start using this new functionality.

We'll have to see what happens with performance -- I'd guess that Postgres's query planner is going to treat each member of the union as an independent query.  I'm also a little concerned that these queries are going to end up complex and difficult to debug.

python/lsst/daf/butler/nonempty_mapping.py
TallJimbo: I think we could get way with versions of `NonemptyMapping` whose values either have a `.copy()` (which should be used) or do not (in which case we assume the result is immutable).  Might be cleaner overall to just dispatch on that `hasattr` check than have subclasses.

I figured the worst-case scenario would be unnecessary copies, and I'd have hoped library authors where consequences would be more severe would just implement `__deepcopy__` or `__reduce__` to fix it. But it sounds like that's not the case.

python/lsst/daf/butler/direct_query_driver/_query_analysis.py
TallJimbo: There _might_ be a case where `T` can be `...`, too; I'm not sure.  I don't really have much of an opinion in the abstract here; it'd come down to what the options actually look like in practice.

python/lsst/daf/butler/direct_query_driver/_driver.py
TallJimbo: Yeah, it's mostly the number of parameters (especially modified-in-place parameters) that makes me lean towards doing this a bit less than you, but if this one really is just two parameters then I agree it's a good idea.

python/lsst/daf/butler/direct_query_driver/_sql_builders.py
TallJimbo: The need to pull this out happened pretty late in the refactor and I just wasn't looking for further refinements at that stage.  I agree composition at least intuitively feels a little better, though I could also imagine that just adding a lot of `.columns` visual noise in practice, since the usage of the base class on its own is pretty niche.

python/lsst/daf/butler/direct_query_driver/_query_builder.py
TallJimbo: I can't claim I tried super hard to make that work, and I may have been overly concerned about pessimizing the common case, but there were indeed a lot of small differences that seemed hard to iron out.

tests/test_pydantic_utils.py
TallJimbo: Oh, bleh.

I tried rewriting this with `__get_pydantic_core_schema__` (since I understand that better), but any solution that fixes this problem seems to fail to work with the union.  I'm inclined to go replace ellipsis usage here with a single-variant `enum`; I had that idea later in the ticket and wondered if it might be better but didn't see a need, but I think this problem puts me on the other side of that fence.

dhirving: Yeah the enum thing will probably work, or one thing that I often see in JSON is using an object like `{ "special": "any_type" }`, which can't be confused with a string.  (Better if you know you need it up front so you end up with a discriminated union instead of an awkward union of string and object, but `str | object` ends up happening a lot.)

python/lsst/daf/butler/nonempty_mapping.py
dhirving: The actual worst-case scenario is more like double-freeing of pointers, accidental sharing of sockets, duplicate copies of large caches that are meant to be shared, or spooky action at a distance involving accidental sharing of state between separate copies.

I think the average library author isn't that interested in exploring all the possible interactions of their system with different weird Python double-underscore features -- if everyone dealt with every possible thing that anyone could do to their objects in Python, every library would be 10x bigger.  Same reason you wouldn't assume you can subclass an arbitrary library type or use it from multiple threads simultaneously.

TallJimbo: Turns out this was even easier than I thought: we're already only using it with `set`, `list`, or `dict` values, and I think it's only useful when the value type is a mutable containers.  So I just added a `Protocol` bound on the value type that requires it to have a `copy` method, and used that.

tests/test_pydantic_utils.py
TallJimbo: Redone with an enum.  Turns out it's important to set the enum value to something that a string (I used `-1`) since that's what Pydantic uses for serialization, and it's not able to handle unions whose members are overlapping in the JSON space.

python/lsst/daf/butler/direct_query_driver/_driver.py
TallJimbo: Done.

python/lsst/daf/butler/direct_query_driver/_query_analysis.py
TallJimbo: Punting this one to future tickets.

python/lsst/daf/butler/direct_query_driver/_sql_builders.py
TallJimbo: Punting this one to future tickets.

python/lsst/daf/butler/direct_query_driver/_query_builder.py
TallJimbo: Punting this one to future tickets.

PR #1106, opened 2024-10-23
DM-46631: fix replace=True on dimensions (etc) with only primary key columns

TallJimbo:
## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/registry/tests/_database.py
```diff
@@ -685,6 +685,26 @@ def testReplace(self):
             [r._asdict() for r in self.query_list(db, tables.a.select())], [row1, row2a, row3]
         )
 
+    def test_replace_pkey_only(self):
+        """Test `Database.replace` on a table only has primary key columns,
```
andy-slac:
```suggestion
        """Test `Database.replace` on a table that only has primary key columns,
```

andy-slac (approved):
Looks great.

PR #1107, opened 2024-10-25
DM-47143: Allow entry points to be used to discover cli plugins

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/cli/butler.py
+++ b/python/lsst/daf/butler/cli/butler.py
@@ -40,6 +40,7 @@ import os
 import traceback
 import types
 from collections import defaultdict
+from functools import cache
 from typing import Any
 
 import click
@@ -303,6 +304,7 @@ class LoaderCLI(click.MultiCommand, abc.ABC):
             cls._mergeCommandLists(commands, defaultdict(list, pluginCommands))
         return commands
 
+    @cache
     def _getCommands(self) -> defaultdict[str, list[str]]:
         """Get the commands offered by daf_butler and plugin packages.
 
@@ -370,6 +372,21 @@ class ButlerCLI(LoaderCLI):
         if commandName == "import":
             return "butler_import"
         return super()._cmdNameToFuncName(commandName)
+
+
+class UncachedButlerCLI(ButlerCLI):
+    """ButlerCLI that can be used where caching of the commands is disabled."""
+
+    def _getCommands(self) -> defaultdict[str, list[str]]:  # type: ignore[override]
+        """Get the commands offered by daf_butler and plugin packages.
+
+        Returns
+        -------
+        commands : `defaultdict` [`str`, `list` [`str`]]
+            The key is the command name. The value is a list of package(s) that
+            contains the command.
+        """
+        return self._mergeCommandLists(self.getLocalCommands(), self._getPluginCommands())
 
 
 @click.command(cls=ButlerCLI, context_settings=dict(help_option_names=["-h", "--help"]))
--- a/python/lsst/daf/butler/cli/butler.py
+++ b/python/lsst/daf/butler/cli/butler.py
@@ -33,7 +33,6 @@ __all__ = (
     "main",
 )
 
-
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/cli/butler.py
```diff
@@ -86,6 +91,12 @@ def _importPlugin(pluginName: str) -> types.ModuleType | type | None | click.Com
         return None
 
 
+@dataclasses.dataclass(frozen=True)
+class PluginCommand:
```
andy-slac, line 95: A docstring would be nice to have, either for the whole class or for attributes.

python/lsst/daf/butler/cli/butler.py
```diff
@@ -255,14 +277,14 @@ def _cmdNameToFuncName(cls, commandName: str) -> str:
...
-    ) -> defaultdict[str, list[str]]:
+        a: defaultdict[str, list[str | PluginCommand]], b: defaultdict[str, list[str | PluginCommand]]
+    ) -> defaultdict[str, list[str | PluginCommand]]:
         """Combine two dicts whose keys are strings (command name) and values
         are list of string (the package(s) that provide the named command).
 
         Parameters
         ----------
-        a : `defaultdict` [`str`, `list` [`str`]]
+        a : `defaultdict` [`str`, `list` [`str` | `PluginCommand` ]]
```
andy-slac, line 291: (Can't comment on `Returns`) `commands` in `Returns` needs to be updated. Can implementation be extended so that duplicated can be merged without duplicates in output?

python/lsst/daf/butler/cli/butler.py
```diff
@@ -278,20 +300,24 @@ def _mergeCommandLists(
...
         Returns
         -------
         commands : `defaultdict` [`str`, `list` [`str`]]
             The key is the command name. The value is a list of package(s) that
             contains the command.
+
+        Notes
+        -----
+        Assumes that if entry points are defined, the plugin environment
+        variable will not be defined for that same package.
```
andy-slac, line 319: Not sure if I read it right, but bot `pie_base` and `daf_butler_migrate` add entry point while keeping envvar too?

andy-slac (approved):
Looks great, a couple of questions.

python/lsst/daf/butler/cli/butler.py
timj: Correct. Because scons does not create the metadata file so python can't find the entry points. There is another ticket to see if scons can be modified to write the correct files so that this all works without involving EUPS. For now we assume that either the package is pip installed or EUPS is setting the env var.

python/lsst/daf/butler/cli/butler.py
timj: It's specifically designed to complain if two packages define a subcommand of the same name so the `list` here is deliberate. There is a later check that makes sure that there is only one entry in the list.

andy-slac: But if two packages are the same package? I'm thinking about the case when a package has both an entry point and envar?

timj: 1. If people don't need to set the envvar if they are using pip and they can't use entry points if they are using EUPS so I don't think it's a problem as such (it will mostly be for people who have been using pip and setting the envvar and it's better if they stop doing that anyhow rather than continuing to think they need the envvar.
2. Currently you get two strings in the comparison: lsst.pipe.base.cli.cmd and lsst.pipe.base.cli.get_cli_subcommands. If you know that the plugin is going to be in `lsst.pipe.base.cli` then you can work it out but in the general case I think you would need to load all the commands and do `is` check and then warn if they are the same -- I'm not sure it's worth it given point 1. We don't want pip users to set the env vars.

andy-slac: OK, I don't want to overcomplicate it, fine to leave it as it is 👍

PR #1109, opened 2024-10-29
DM-45873: Initial query_all_datasets implementation

dhirving:
## Checklist

Extracted the query logic from the `query-datasets` CLI and used it to provide an initial implementation of query_all_datasets.  

This gets the interface and unit tests in place for a future PR that will run the whole query on the server side for RemoteButler.

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/script/queryDatasets.py
+++ b/python/lsst/daf/butler/script/queryDatasets.py
@@ -259,7 +259,7 @@ class QueryDatasets:
             Dataset references matching the given query criteria grouped
             by dataset type.
         """
-        query_collections: Iterable[str] = self._collections_wildcard or ["*"]
+        query_collections = self._collections_wildcard or ["*"]
 
         warn_limit = False
         if self._limit < 0:
--- a/python/lsst/daf/butler/script/queryDatasets.py
+++ b/python/lsst/daf/butler/script/queryDatasets.py
@@ -263,6 +263,7 @@ class QueryDatasets:
         query_collections = self._collections_wildcard or ["*"]
 
         warn_limit = False
+        limit_reached = False
         if self._limit < 0:
             # Negative limit means we should warn if the limit is exceeded.
             warn_limit = True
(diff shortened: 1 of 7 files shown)


python/lsst/daf/butler/script/queryDatasets.py
```diff
@@ -258,83 +260,50 @@ def getDatasets(self) -> Iterator[list[DatasetRef]]:
...
+        else:
+            limit = self._limit
+
+        try:
+            pages = query_all_datasets(
+                self.butler,
+                collections=query_collections,
                 find_first=self._find_first,
+                name=self._dataset_type_glob,
                 with_dimension_records=True,
```
timj, line 283: Following on from a Slack comment, it looks like there are other scripts using the `QueryDatasets` class that do need dimensions so if you want to be able to turn this off for `query-datasets` but turn it on for `transfer-datasets` then you will need to add a parameter higher up.

tests/test_cliCmdQueryDatasets.py
```diff
@@ -357,6 +357,13 @@ def test_limit_order(self):
         ]
         self.assertAstropyTablesEqual(tables, expectedTables, filterColumns=True)
 
+        # Same as previous test, but with positive limit so no warning is
+        # issued.
+        tables = self._queryDatasets(
+            repo=testRepo.butler, limit=1, order_by=("visit"), collections="*", glob="*"
```
timj: Maybe test that no warning was issued (by issuing your own warning and then making sure you only got the one warning)?

timj (approved):
Thanks. I've tested it and it does seem to work as expected. It looks like the restriction of order-by only working for a single dataset type is not part of this PR yet.

python/lsst/daf/butler/script/queryDatasets.py
dhirving: Thanks -- I failed to notice that `QueryDatasets` was shared by multiple scripts.  I had removed order-by and the dimension records in the follow-up PR to this one but I'll have to put the dimension records back.

PR #1111, opened 2024-11-01
DM-47328: Add QuantumBackedButler.retrieveArtifacts

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_quantum_backed.py
+++ b/python/lsst/daf/butler/_quantum_backed.py
@@ -531,6 +531,53 @@ class QuantumBackedButler(LimitedButler):
         """
         return retrieve_and_zip(refs, destination, self._datastore.retrieveArtifacts, overwrite)
 
+    def retrieve_artifacts(
+        self,
+        refs: Iterable[DatasetRef],
+        destination: ResourcePathExpression,
+        transfer: str = "auto",
+        preserve_path: bool = True,
+        overwrite: bool = False,
+    ) -> list[ResourcePath]:
+        """Retrieve the artifacts associated with the supplied refs.
+
+        Parameters
+        ----------
+        refs : iterable of `DatasetRef`
+            The datasets for which artifacts are to be retrieved.
+            A single ref can result in multiple artifacts. The refs must
+            be resolved.
+        destination : `lsst.resources.ResourcePath` or `str`
+            Location to write the artifacts.
+        transfer : `str`, optional
+            Method to use to transfer the artifacts. Must be one of the options
+            supported by `~lsst.resources.ResourcePath.transfer_from()`.
+            "move" is not allowed.
+        preserve_path : `bool`, optional
+            If `True` the full path of the artifact within the datastore
+            is preserved. If `False` the final file component of the path
+            is used.
+        overwrite : `bool`, optional
+            If `True` allow transfers to overwrite existing files at the
+            destination.
+
+        Returns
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/_quantum_backed.py
```diff
@@ -531,6 +531,53 @@ def retrieve_artifacts_zip(
...
+        destination : `lsst.resources.ResourcePath` or `str`
+            Location to write the artifacts.
+        transfer : `str`, optional
+            Method to use to transfer the artifacts. Must be one of the options
+            supported by `~lsst.resources.ResourcePath.transfer_from()`.
+            "move" is not allowed.
+        preserve_path : `bool`, optional
+            If `True` the full path of the artifact within the datastore
+            is preserved. If `False` the final file component of the path
+            is used.
```
andy-slac, line 559: What happens if there are duplicates after stripping directories, will it raise or overwrite?

python/lsst/daf/butler/_quantum_backed.py
```diff
@@ -531,6 +531,53 @@ def retrieve_artifacts_zip(
         """
         return retrieve_and_zip(refs, destination, self._datastore.retrieveArtifacts, overwrite)
 
+    def retrieveArtifacts(
```
andy-slac: If we want to eventually make all methods snake-case, maybe call it `retrieve_artifacts` instead? I guess this is for consistency with `Butler`, but `LimitedButler` does not have the abstract method for this?

python/lsst/daf/butler/_quantum_backed.py
```diff
@@ -531,6 +531,53 @@ def retrieve_artifacts_zip(
...
+            is used.
+        overwrite : `bool`, optional
+            If `True` allow transfers to overwrite existing files at the
+            destination.
+
+        Returns
+        -------
+        targets : `list` of `lsst.resources.ResourcePath`
+            URIs of file artifacts in destination location. Order is not
+            preserved.
```
andy-slac, line 568: Maybe mention that the list will also include ZipIndex URI, so the targets length will be longer that input refs?

andy-slac (approved):
Looks great.

python/lsst/daf/butler/_quantum_backed.py
timj: It seems vaguely confusing to me to have `QBB.retrieve_artifacts` and `Butler.retrieveArtifacts`. @TallJimbo ?

python/lsst/daf/butler/_quantum_backed.py
timj: It will depend on the overwrite flag. The way we set up our file templates it *should* be impossible.

python/lsst/daf/butler/_quantum_backed.py
timj: It doesn't return the index JSON uri.

python/lsst/daf/butler/_quantum_backed.py
TallJimbo: Let's add the snake case version everywhere, even if we don't drop the camelCase one.

timj: I'll do that on a different ticket since I think I need to think about it a little before adding all the butler alternative APIs with different parameters.

PR #1112, opened 2024-11-01
DM-47335: Fix dataset query missing implied dimensions

dhirving:
Revert serialization change from DM-46776 in SerializedDataCoordinate, which caused it to drop values for implied dimensions.

That change broke the transfer of DatasetRefs from server to client, since it drops the values for implied dimensions in the data ID.  Per the documented behavior of dataset queries, the returned refs will have `hasFull() = True` and this behavior is relied on in tutorial notebooks.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/dimensions/_coordinate.py
+++ b/python/lsst/daf/butler/dimensions/_coordinate.py
@@ -709,7 +709,7 @@ class DataCoordinate:
         else:
             records = None
 
-        return SerializedDataCoordinate(dataId=dict(self.required), records=records)
+        return SerializedDataCoordinate(dataId=dict(self.mapping), records=records)
 
     @classmethod
     def from_simple(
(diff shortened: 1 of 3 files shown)


timj (approved):

TallJimbo:
I can see how this broke `RemoteButler` query results, but I'm curious how this could have changed dataset type definitions, since initializing a `DimensionGroup` with the required subset and the full set should always yield the same result, unless the dimension universe has changed.

I'm not at all against the revert that was just merged, but I think long term this is a good reason to have the wire format for queries differ from long-term persistence and other cases where the consumer is another butler, because I do still think the latter should be only saving the required subset.

dhirving:
Looks like you're right -- I didn't realize that the DatasetType constructor would automatically add all implied dimensions.  (The unit test data that is used everywhere explicitly adds in `band` even though it's theoretically implied by `physical_filter`.)

So I can probably un-revert the `SerializedDatasetType` change.

PR #1113, opened 2024-11-01
DM-47325: Add API for parsing butler dataset URIs (butler and ivo)

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -568,7 +568,8 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
         in the ``ivo`` URI query component.
         """
         parsed = urllib.parse.urlparse(uri)
-        if parsed.scheme == "ivo":
+        parsed_scheme = parsed.scheme.lower()
+        if parsed_scheme == "ivo":
             # Do not validate the netloc or the path values.
             qs = urllib.parse.parse_qs(parsed.query)
             if "repo" not in qs or "id" not in qs:
@@ -577,8 +578,8 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
                 raise ValueError(f"Butler IVOID only supports a single value of repo and id, got {uri}")
             label = qs["repo"][0]
             id_ = qs["id"][0]
-        elif parsed.scheme == "butler":
-            label = parsed.netloc
+        elif parsed_scheme == "butler":
+            label = parsed.netloc  # Butler label is case sensitive.
             # Need to strip the leading /.
             id_ = parsed.path[1:]
         else:
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -29,6 +29,7 @@ from __future__ import annotations
 
 __all__ = ["Butler"]
 
+import dataclasses
 import urllib.parse
 import uuid
 from abc import abstractmethod
@@ -72,6 +73,19 @@ if TYPE_CHECKING:
     from .transfers import RepoExportContext
 
 _LOG = getLogger(__name__)
+
+
+@dataclasses.dataclass
+class ParsedButlerDatasetURI:
+    label: str
+    dataset_id: uuid.UUID
+    uri: str
+
+
+@dataclasses.dataclass
+class SpecificButlerDataset:
+    butler: Butler
+    dataset: DatasetRef | None
 
 
 class Butler(LimitedButler):  # numpydoc ignore=PR02
@@ -530,7 +544,7 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
         return ButlerRepoIndex.get_known_repos()
 
     @classmethod
-    def parse_dataset_uri(cls, uri: str) -> tuple[str, DatasetId]:
+    def parse_dataset_uri(cls, uri: str) -> ParsedButlerDatasetURI:
         """Extract the butler label and dataset ID from a dataset URI.
 
         Parameters
@@ -540,11 +554,9 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
 
         Returns
         -------
-        label : `str`
+        parsed : `ParsedButlerDatasetURI`
             The label associated with the butler repository from which this
-            dataset originates.
-        dataset_id : `DatasetId`
-            The ID of the dataset.
+            dataset originates and the ID of the dataset.
 
         Notes
         -----
@@ -594,12 +606,12 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
             e.add_note(f"Error extracting dataset ID from uri {uri!r} with dataset ID string {id_!r}")
             raise
 
-        return label, dataset_id
+        return ParsedButlerDatasetURI(label=label, dataset_id=dataset_id, uri=uri)
 
     @classmethod
     def get_dataset_from_uri(
         cls, uri: str, factory: LabeledButlerFactoryProtocol | None = None
-    ) -> tuple[Butler, DatasetRef | None]:
+    ) -> SpecificButlerDataset:
         """Get the dataset associated with the given dataset URI.
 
         Parameters
@@ -613,23 +625,22 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
 
         Returns
         -------
-        butler : `Butler`
-            Butler object associated with this URI.
-        ref : `DatasetRef` or `None`
-            The dataset associated with that URI, or `None` if the UUID
-            is valid but the dataset is not known to this butler.
+        result : `SpecificButlerDataset`
+            The butler associated with this URI and the dataset itself.
+            The dataset can be `None` if the UUID is valid but the dataset
+            is not known to this butler.
         """
-        label, dataset_id = cls.parse_dataset_uri(uri)
+        parsed = cls.parse_dataset_uri(uri)
         butler: Butler | None = None
         if factory is not None:
             # If the label is not recognized, it might be a path.
             try:
-                butler = factory(label)
+                butler = factory(parsed.label)
             except KeyError:
                 pass
         if butler is None:
-            butler = cls.from_config(label)
-        return butler, butler.get_dataset(dataset_id)
+            butler = cls.from_config(parsed.label)
+        return SpecificButlerDataset(butler=butler, dataset=butler.get_dataset(parsed.dataset_id))
 
     @abstractmethod
     def _caching_context(self) -> AbstractContextManager[None]:
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -63,6 +63,7 @@ if TYPE_CHECKING:
     from ._dataset_type import DatasetType
     from ._deferredDatasetHandle import DeferredDatasetHandle
     from ._file_dataset import FileDataset
+    from ._labeled_butler_factory import LabeledButlerFactoryProtocol
     from ._storage_class import StorageClass
     from ._timespan import Timespan
     from .datastore import DatasetRefURIs
@@ -583,27 +584,35 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
         return label, dataset_id
 
     @classmethod
-    def get_dataset_from_uri(cls, uri: str) -> DatasetRef | None:
+    def get_dataset_from_uri(
+        cls, uri: str, factory: LabeledButlerFactoryProtocol | None = None
+    ) -> DatasetRef | None:
         """Get the dataset associated with the given dataset URI.
 
         Parameters
         ----------
         uri : `str`
             The URI associated with a dataset.
+        factory : `LabeledButlerFactoryProtocol` or `None`, optional
+            Bound factory function that will be given the butler label
+            and receive a `Butler`.
 
         Returns
         -------
         ref : `DatasetRef` or `None`
             The dataset associated with that URI, or `None` if the UUID
             is valid but the dataset is not known to this butler.
-
-        Notes
-        -----
-        It might be possible to pass in an optional ``LabeledButlerFactory``
-        but how would a caller know the right access token to supply?
         """
         label, dataset_id = cls.parse_dataset_uri(uri)
-        butler = cls.from_config(label)
+        butler: Butler | None = None
+        if factory is not None:
+            # If the label is not recognized, it might be a path.
+            try:
+                butler = factory(label)
+            except KeyError:
+                pass
+        if butler is None:
+            butler = cls.from_config(label)
         return butler.get_dataset(dataset_id)
 
     @abstractmethod
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -549,12 +549,13 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
 
         Notes
         -----
-        Supports dataset URIs of the forms ``ivo://rubin/butler_label/UUID``
+        Supports dataset URIs of the forms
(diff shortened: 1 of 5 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -525,6 +528,83 @@ def get_known_repos(cls) -> set[str]:
...
+        try:
+            dataset_id = uuid.UUID(hex=id_)
+        except Exception as e:
+            e.add_note(f"Error extracting dataset ID from uri {uri!r} with dataset ID string {id_!r}")
+            raise
+
+        return label, dataset_id
+
+    @classmethod
+    def get_dataset_from_uri(cls, uri: str) -> DatasetRef | None:
```
timj: @dhirving I was thinking that this should take a `LabeledButlerFactory` as an optional parameter (so that a service like vo-cutouts can call it directly with the URI) but do you have any thoughts on how to pass in the token? In the general case the caller won't know which butler the URI is referring to. Can I assume that `**kwargs` should be forwarded as `factory.create_butler(label=label, **kwargs)` and assume the token would be shared? Or should there be a dict indexed by label that has the additional parameters that the service can set up once and pass in?

python/lsst/daf/butler/_butler.py
```diff
@@ -525,6 +528,83 @@ def get_known_repos(cls) -> set[str]:
...
+            The ID of the dataset.
+
+        Notes
+        -----
+        Supports dataset URIs of the forms ``ivo://rubin/butler_label/UUID``
+        and ``butler://butler_label/UUID``. In ``ivo`` URIs the butler label
+        can include ``/`` and the leading ``/`` is always stripped. If the
+        repository label starts with ``/`` then it must be doubled up. e.g.,
+
+            ivo://rubin//repo/main/82d79caa-0823-4300-9874-67b737367ee0
```
timj: @gpdf do you have any thoughts as to what the netloc field for our IVO IDs might look like?

python/lsst/daf/butler/_butler.py
dhirving: I like your idea of a config object that could be set up by the caller and passed through.

One other idea would be a dependency injection pattern.  Maybe declare a protocol like
```
class ButlerFactoryProtocol(Protocol):
    create(label: str) -> Butler

# (or just `Callable[[str], Butler]
```
which `get_dataset_from_uri` could take as a parameter.  Then `LabeledButlerFactory` could grow a method `bind(access_token: str) -> ButlerFactoryProtocol` to create a factory with the access token already bound in.  That would let callers do whatever they want to customize the instance creation.

dhirving: At the moment the access token is in fact shared since it comes from Gafaelfawr and just represents the user's identity.  Though no guarantees it will stay they way if we end up needing to delegate authentication between IDF and USDF, or some other unfortunate thing.

timj: Trying to think through what you said above. For:

```python
my_factory = factory.bind(access_token=token) 
ref = Butler.get_dataset_from_uri(dataset_uri, my_factory)
```
doesn't this mean that the access_token is required to be the same for every butler label?

Don't we need:

```python
my_factory = factory.bind("dp02": {"access_token": token}, "dp02-direct": {"access_token": None})
```
?

dhirving: There is only one access token per HTTP request -- Gafaelfawr generates it to identify the user.  It is not associated with the repository in any way.  So at least at the moment there's only one access token.  If you pass an access token to `LabeledButlerFactory` for a DirectButler label it just ignores it.

The advantage to the dependency injection pattern is that this code wouldn't have to know anything at all about authentication or where Butlers come from -- it makes that the caller's responsibility.  This code just knows that if it has a label, the caller will work out how to get it the matching Butler.  So if things get more complicated later, this interface doesn't have to change.  (One example might be caching the Butler if someone wants to call `get_dataset_from_uri` multiple times in the same request -- though that might be better solved by wrapping this function in a separate object.)

Your config object idea would work well too, allowing us to add more options later if `LabeledButlerFactory` gets more complicated.  It's just a little less flexible and it couples this function tightly to `LabeledButlerFactory`.

python/lsst/daf/butler/_butler.py
gpdf: Following discussion with @timj today, the suggestion is that the netloc be treated leniently -- it will have a standard value (e.g., "rubin", still to be confirmed with the IVOA), and then the path will be:

butler/(butler-identity)/(uuid)

If the *butler-identity* begins with a "/", e.g., "/repo/main", then an extra "/" is required in the pathname, e.g., 

ivo://rubin/butler//repo/main/82d79caa-0823-4300-9874-67b737367ee0

To be explicit the *butler-identity* is everything between the first and last slashes in the path.

gpdf (changes requested):
Responded to request for feedback on ivo: URIs.

python/lsst/daf/butler/_butler.py
timj: After discussing this with Markus it was noted that the variable part of the IVO ID has to be a query parameter. ie something like

```
ivo://lsst.org/butler?/repo/main/82d79caa-0823-4300-9874-67b737367ee0
```
or I suppose:
```
ivo://lsst.org/butler//repo/main?82d79caa-0823-4300-9874-67b737367ee0
```
Depending on whether we have a look up service per butler concept or per butler repository.
Maybe it's worth pondering what it would look like for an Object/Source ID.
```
ivo://lsst.org/object?1234567890
```
where in theory we can work out the data release number from the object ID. Or
```
ivo://lsst.org/object/dr1?1234567890
```

timj: or even

```
ivo://lsst.org/dr1?1234567890
```
and
```
ivo://lsst.org/dr1?butler/82d79caa-0823-4300-9874-67b737367ee0
```

timj: I am working through ideas for IVO IDs on [DMTN-302](https://dmtn-302.lsst.io).

tests/test_simpleButler.py
```diff
@@ -882,10 +883,61 @@ def makeButler(self, writeable: bool = False) -> Butler:
...
+        config_dir = butler._config["root"]
+
+        # Read it via a repo label and a path.
+        with tempfile.NamedTemporaryFile(mode="w", suffix=".yaml") as index_file:
+            label = "test_repo"
+            index_file.write(f"{label}: {config_dir}\n")
+            index_file.flush()
+            with mock_env({"DAF_BUTLER_REPOSITORY_INDEX": index_file.name}):
+                butler_factory = LabeledButlerFactory()
+                factory = butler_factory.bind(access_token=None)
```
timj, line 912: @rra does this approach work for you in the cutout service?

```python
factory = butler_factory.bind(access_token=token)
...
ref = Butler.get_dataset_from_uri(dataset_uri, factory=factory)
```
?

rra: The cutout service passes a Butler instance into the backend, so the pattern looks like:
```python
butler_factory = LabeledButlerFactory()

def _get_backend(label: str, token: str) -> ImageCutoutBackend:
    # Called for each cutout
    factory = butler_factory.bind(access_token=token)
    butler = factory.create_butler(label=label)
    # ...
    return ImageCutoutBackend(butler, projection_finder, output, tmpdir)
```
Is that what you had in mind? Basically move the access token parameter to `create_butler` to an intermediate step to create a factory with a bound token? If so, that would be fine here.

timj: Ah. No. It's probably the wrong API for you. Somewhere you are parsing the `butler://` URI and I'm trying to provide code here that will hide the URI structure from you (so that we can also support the new `ivo://` URIs). This PR creates two APIs: one that parses the URI and returns the butler repo label and the UUID, and another API (that is the one I talk about here) that lets you retrieve the DatasetRef directly from the URI and a butler factory. Maybe `get_dataset_from_uri` should return the `Butler` instance along with the `DatasetRef`?

rra: Maybe I could pass the return value of `bind` into the backend along with the URIs as-is without parsing them and then the backend can do whatever it needs to do? That would be even more convenient. In other words, have the constructor of `ImageCutoutBackend` take a Butler factor instead of a Butler instance.

timj: I think passing the bound factory and URI into whatever is wanting to know the DatasetRef is what we want here.

python/lsst/daf/butler/_butler.py
timj: For completeness we now parse `ivo://org.rubinobs/usdac/lsst-dr1?repo=dr1id=UUID`.

The parser does not attempt to look at anything other than the query component.

python/lsst/daf/butler/_butler.py
```diff
@@ -526,6 +529,107 @@ def get_known_repos(cls) -> set[str]:
...
+        This method does not attempt to check that the dataset exists in the
+        labeled butler.
+
+        Since the IVOID can be issued by any publisher to represent a Butler
+        dataset there is no validation of the path or netloc component of the
+        URI. The only requirement is that there are ``id`` and ``repo`` keys
+        in the ``ivo`` URI query component.
+        """
+        parsed = urllib.parse.urlparse(uri)
+        if parsed.scheme == "ivo":
```
stvoutsin:
```suggestion
        scheme_lower = parsed.scheme.lower()
        if scheme_lower == "ivo":
```

python/lsst/daf/butler/_butler.py
```diff
@@ -526,6 +529,107 @@ def get_known_repos(cls) -> set[str]:
...
+            # Do not validate the netloc or the path values.
+            qs = urllib.parse.parse_qs(parsed.query)
+            if "repo" not in qs or "id" not in qs:
+                raise ValueError(f"Missing 'repo' and/or 'id' query parameters in IVOID {uri}.")
+            if len(qs["repo"]) != 1 or len(qs["id"]) != 1:
+                raise ValueError(f"Butler IVOID only supports a single value of repo and id, got {uri}")
+            label = qs["repo"][0]
+            id_ = qs["id"][0]
+        elif parsed.scheme == "butler":
```
stvoutsin:
```suggestion
        elif scheme_lower == "butler":
```

python/lsst/daf/butler/_butler.py
stvoutsin: I think for the IVOIDs the registry part (everything before ? or #) needs to be case insensitive

python/lsst/daf/butler/_labeled_butler_factory.py
```diff
@@ -83,6 +90,12 @@ def __init__(self, repositories: Mapping[str, str] | None = None) -> None:
         # This may be overridden by unit tests.
         self._preload_direct_butler_cache = True
 
+    def bind(self, access_token: str | None) -> LabeledButlerFactoryProtocol:
```
stvoutsin, line 93: Perhaps a docstring here for clarity?

```suggestion
    def bind(self, access_token: str | None) -> LabeledButlerFactoryProtocol:
        """Create a callable factory function for generating Butler instances.

        Parameters
        ----------
        access_token : str | None
            An optional access token to authenticate with the Butler.

        Returns
        -------
        LabeledButlerFactoryProtocol
            A callable that takes a label as input and returns a Butler
            instance.
        """
```

tests/test_simpleButler.py
```diff
@@ -882,10 +883,75 @@ def makeButler(self, writeable: bool = False) -> Butler:
...
+            "butler://1234",  # No UUID.
+            "butler:///1234",  # No label.
+            "ivo://rubin/1234",  # No query part and bad UUID and no label.
+            "ivo://rubin/datasets/dr1/82d79caa-0823-4300-9874-67b737367ee0",  # No query part.
+            "ivo://org.rubinobs/datasets?repo=dr1&id=1234",  # Bad UUID.
+            "ivo://org.rubinobs/butler?release=dr1&id=82d79caa-0823-4300-9874-67b737367ee0",  # No repo key.
+            "ivo://org.rubinobs/butler?repo=dr1&repo=dr2&id=82d79caa-0823-4300-9874-67b737367ee0",  # 2 vals.
+            "ivo://org.rubinobs/something?repo=%20&id=82d79caa-0823-4300-9874-67b737367ee0",  # no repo.
+            "https://something.edu/1234",  # Wrong scheme.
+        ):
```
stvoutsin, line 951: Might it also be useful to test percent-encoded params?

python/lsst/daf/butler/_butler.py
```diff
@@ -526,6 +529,107 @@ def get_known_repos(cls) -> set[str]:
...
+        except Exception as e:
+            e.add_note(f"Error extracting dataset ID from uri {uri!r} with dataset ID string {id_!r}")
+            raise
+
+        return label, dataset_id
+
+    @classmethod
+    def get_dataset_from_uri(
+        cls, uri: str, factory: LabeledButlerFactoryProtocol | None = None
+    ) -> tuple[Butler, DatasetRef | None]:
```
stvoutsin: I think this is fine if the result tuple is unlikely to change in the future, If you think it may though, I'd also consider a dataclass/named tuple result which is a bit cleaner to change and generally more explicit.

python/lsst/daf/butler/_butler.py
```diff
@@ -526,6 +529,107 @@ def get_known_repos(cls) -> set[str]:
...
+        Since the IVOID can be issued by any publisher to represent a Butler
+        dataset there is no validation of the path or netloc component of the
+        URI. The only requirement is that there are ``id`` and ``repo`` keys
+        in the ``ivo`` URI query component.
+        """
+        parsed = urllib.parse.urlparse(uri)
+        if parsed.scheme == "ivo":
+            # Do not validate the netloc or the path values.
+            qs = urllib.parse.parse_qs(parsed.query)
+            if "repo" not in qs or "id" not in qs:
```
stvoutsin, line 587: I'm not sure what is expected of query param keys in terms of case sensitivity. But I think no harm in treating those as case-insensitive as well?

stvoutsin (approved):
Overall looks good to me, I think the new methods make sense and are useful. Only a couple thoughts/suggestions here

python/lsst/daf/butler/_labeled_butler_factory.py
timj: Thanks. I'm confused as to how it is that my numpydoc linter didn't complain.

python/lsst/daf/butler/_butler.py
timj: I will think about it. I can't currently think of any information that could be returned given the limited content of the URI itself.

tests/test_simpleButler.py
timj: The `%20` is converted to a space for the failing test here but I added a `%2F` for a `/` in the URIs that work.

python/lsst/daf/butler/_butler.py
timj: In theory we are generating these IVO IDs so I don't really want the added complication of converting the dict to a dict with case insensitive keys (I have a recollection of using such a special dict at some point but I'm not sure where it was).

python/lsst/daf/butler/_butler.py
timj: Switched to data class.

gpdf (approved):
I'm fine with this now, though we are still discussion what will go into the `authority` field in the IVOID versus what will go in the `resource key` field (hostname and path fields, colloquially speaking).

I am assured that that's not going to require additional work on the code affected by this PR.

PR #1114, opened 2024-11-04
DM-47375: Run query_all_datasets in a single request for RemoteButler

dhirving:
Added a server-side endpoint to handle `query_all_datasets` in a single request.  `query_all_datasets` can potentially involve hundreds or thousands of separate dataset queries, and we don't want clients slamming the server with that many HTTP requests.

The new endpoint streams results in the same manner as the existing query endpoints used by `QueryDriver`, but it is separate from the `Query`/`QueryDriver` framework.

This is not yet used in the CLI tools and `Butler._query_all_datasets` is still private -- we need to deploy an updated server with this change before we can release the client side.

`--order-by` in the CLI tools is now restricted to queries for a single dataset type -- future implementations of `query_all_datasets` may not support it.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/remote_butler/server/handlers/_external.py
+++ b/python/lsst/daf/butler/remote_butler/server/handlers/_external.py
@@ -27,17 +27,19 @@
 
 from __future__ import annotations
 
-from lsst.daf.butler.remote_butler.server.handlers._utils import set_default_data_id
-
 __all__ = ()
 
 import uuid
 from typing import Any
 
 from fastapi import APIRouter, Depends
-from lsst.daf.butler import Butler, CollectionType, DatasetRef
-from lsst.daf.butler.registry.interfaces import ChainedCollectionRecord
-from lsst.daf.butler.remote_butler.server_models import (
+
+from ...._butler import Butler
+from ...._collection_type import CollectionType
+from ...._dataset_ref import DatasetRef
+from ...._exceptions import DatasetNotFoundError
+from ....registry.interfaces import ChainedCollectionRecord
+from ...server_models import (
     ExpandDataIdRequestModel,
     ExpandDataIdResponseModel,
     FindDatasetRequestModel,
@@ -55,10 +57,9 @@ from lsst.daf.butler.remote_butler.server_models import (
     QueryDatasetTypesRequestModel,
     QueryDatasetTypesResponseModel,
 )
-
-from ...._exceptions import DatasetNotFoundError
 from .._dependencies import factory_dependency
 from .._factory import Factory
+from ._utils import set_default_data_id
 
 external_router = APIRouter()
 
--- a/python/lsst/daf/butler/remote_butler/server/handlers/_external.py
+++ b/python/lsst/daf/butler/remote_butler/server/handlers/_external.py
@@ -27,6 +27,8 @@
 
 from __future__ import annotations
 
+from lsst.daf.butler.remote_butler.server.handlers._utils import set_default_data_id
+
 __all__ = ()
 
 import uuid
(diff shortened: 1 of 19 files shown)


python/lsst/daf/butler/script/queryDatasets.py
```diff
@@ -275,28 +276,30 @@ def getDatasets(self) -> Iterator[list[DatasetRef]]:
...
+                    limit=limit,
+                    order_by=self._order_by,
+                )
+                datasets_found = 0
+                for dataset_type, refs in _chunk_by_dataset_type(pages):
+                    datasets_found += len(refs)
+                    if warn_limit and limit is not None and datasets_found >= limit:
+                        # We asked for one too many so must remove that from
+                        # the list.
+                        refs = refs[0:-1]
```
TallJimbo, line 307: This looks like an O(N) way to do a deletion from a list, while `del refs[-1]` would be O(1).  Is there some reason the original list shouldn't be modified?

python/lsst/daf/butler/remote_butler/server/handlers/_external_query.py
```diff
@@ -64,16 +64,60 @@
...
+        """Execute the database query and and return pages of results.
+
+        Parameters
+        ----------
+        context : generic
+            Value returned by the call to ``setup()``.
+        """
+
+
+class _StreamQueryDriverExecute(StreamingQuery):
```
TallJimbo: I think you might need to repeat the `[_TContext]` generic parameterization here to avoid having it implicitly interpreted as `StreamingQuery[Any]`, but I'm not sure.

python/lsst/daf/butler/remote_butler/server/handlers/_external.py
```diff
@@ -27,6 +27,8 @@
 
 from __future__ import annotations
 
+from lsst.daf.butler.remote_butler.server.handlers._utils import set_default_data_id
```
TallJimbo: This import should move down into the main section (I bet VSCode added it for you).

I'm also curious why this and many other imports here are absolute rather than relative.

TallJimbo (approved):
Looks good.

I'm a little worried some of this will need refactoring when we add planned query interfaces like counts of datasets by type, but that may reflect my own tendency to try to prototype everything out before implementing anything rather than a concrete concern.

I'm glad you kept the restoration of dimension records in `query_all_datasets` private, because I suspect what `transferDatasets` _really_ wants is some sort of normalized batch of all dimension records relevant for a bunch of datasets, not actual expanded data IDs; that's also what QG generation really wants, and eventually I hope we can provide it rather than creating duplicate copies of a lot of those dimension records by denormalizing them into the result rows.

python/lsst/daf/butler/script/queryDatasets.py
dhirving: That line is unchanged from the previous revision.  But the full history of it is that Tim had it as `.pop()` originally, which would be O(1).  But when I pulled out `query_all_datasets` that introduced a bug, because  `query_all_datasets` is a coroutine and this function was mutating the list before `query_all_datasets` was done with it.  So since this list is at most a few hundred thousand items and is sandwiched in between network I/O and console I/O I thought it was better to prevent similar bugs in the future by avoiding the mutation than to worry about like 1ms in a CLI command that takes 2 seconds to even start up.

python/lsst/daf/butler/remote_butler/server/handlers/_external.py
dhirving: I think at the time they were added, there was a theory that the Butler server code was supposed to move into a separate repository at some point.  Made them all relative since I don't think we want to do that anymore.

PR #1116, opened 2024-11-13
DM-47543: Reduce likelihood of collection deadlocks

dhirving:
Fix an issue where concurrent calls to transfer_from() in Prompt Processing were causing Postgres to abort with a deadlock error.

We now register collection names in alphabetical order when multiple collections are registered inside a transaction in `transfer_from()`.  Because `ButlerCollections.register()` inserts rows into a table with a unique index, Postgres takes a lock when the row is inserted and holds it until the end of the transaction.  If there are two processes with concurrent transactions attempting to create the same collections, it is possible for the locks to be acquired in different orders, causing a deadlock.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_butler/_direct_butler.py
+++ b/python/lsst/daf/butler/direct_butler/_direct_butler.py
@@ -46,7 +46,7 @@ import warnings
 from collections import Counter, defaultdict
 from collections.abc import Iterable, Iterator, MutableMapping, Sequence
 from types import EllipsisType
-from typing import TYPE_CHECKING, Any, ClassVar, TextIO, cast
+from typing import TYPE_CHECKING, Any, ClassVar, NamedTuple, TextIO, cast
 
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -2351,3 +2362,12 @@ def _preload_cache(self) -> None:
...
     the ``registry`` property.
     """
+
+
+class _RefGroup(NamedTuple):
+    """Key identifying a batch of DatasetRefs to be inserted in
+    `Butler.transfer_from`.
+    """
+
+    dataset_type: DatasetType
```
TallJimbo: This is a somewhat expensive equality key, and I think the `str` dataset type name would work as well.

TallJimbo (approved):

PR #1124, opened 2024-12-03
DM-47889: Prevent DB connection pool exhaustion in Butler server

dhirving:
Adjusted the database connection pool parameters and added a limit to the number of long-running queries that the server will execute simultaneously.

If a client tries to run a long-running query while the server is overloaded, the request will be rejected with an HTTP 503 and the client will retry later.  This also sets up the retry logic for HTTP 429, which will be used in the future for other kinds of rate limiting.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/remote_butler/server/handlers/_query_streaming.py
+++ b/python/lsst/daf/butler/remote_butler/server/handlers/_query_streaming.py
@@ -123,8 +123,7 @@ async def execute_streaming_query(query: StreamingQuery) -> StreamingResponse:
         await _block_retry_for_unit_test()
         raise HTTPException(
             status_code=503,  # service temporarily unavailable
-            detail="The Butler Server is currently overloaded with requests."
-            f" Try again in {_QUERY_RETRY_SECONDS} seconds.",
+            detail="The Butler Server is currently overloaded with requests.",
             headers={"retry-after": str(_QUERY_RETRY_SECONDS)},
         )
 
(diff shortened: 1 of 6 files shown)


python/lsst/daf/butler/remote_butler/server/handlers/_query_streaming.py
```diff
@@ -43,11 +44,26 @@
...
+# simultaneously, to prevent the database connection pool and the thread pool
+# from being tied up indefinitely.  Beyond this number, the server will return
+# an HTTP 503 Service Unavailable with a Retry-After header.  We are currently
+# using the default FastAPI thread pool size of 40 (total) and have 40 maximum
+# database connections (per Butler repository.)
+_MAXIMUM_CONCURRENT_STREAMING_QUERIES = 25
+# How long we ask callers to wait before trying their query again.
+# The hope is that they will bounce to a less busy replica, so we don't want
+# them to wait too long.
+_QUERY_RETRY_SECONDS = 5
```
timj, line 57: Are we planning to do randomized exponential backoff at some point in the future? Can we tell if this is the 100th time a client has attempted a retry?

timj (approved):

python/lsst/daf/butler/remote_butler/server/handlers/_query_streaming.py
dhirving: We don't really want to go exponential (at least by default), since most consumers of the Butler can't tolerate a request hanging out indefinitely.  This is just supposed to buy us enough time for auto-scaling to get some more replicas booted up.  When this case is triggered the scarce resource is threads and database connections, so I'm not really worried about the minimal amount of CPU and network needed to tell the caller to go away once every 5 seconds.

I added a cap on the retry time on the client side.  Currently we don't have a way to track what a client is up to on the server side, but we will eventually have a redis instance (or Russ will) for more elaborate rate limiting.

PR #1127, opened 2024-12-06
DM-47976: Support conversion field in file template format strings

tgoldina:
A fix to support type conversion in file template format strings. For example, the conversion specifier `!s` in a format string like `{healpix!s:.2s}` converts a numeric field healpix into a string before applying additional formatting.

The modified unit test, which previously produced a `ValueError: Unknown format code 's' for object of type 'int'`, now passes.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/doc/changes/DM-47976.bugfix.md
+++ b/doc/changes/DM-47976.bugfix.md
@@ -1,0 +1,1 @@
+Support type conversion field in file template format strings.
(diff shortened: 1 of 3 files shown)


timj:
Can you add a test please?

andy-slac (approved):
Looks great.

PR #1132, opened 2024-12-10
DM-47770: Fix threadsafety of sqlalchemy MetaData access

dhirving:
Wrap all accesses to sqlalchemy.MetaData with a lock to avoid concurrency issues.

sqlalchemy.MetaData is documented to be threadsafe for reads, but not with concurrent modifications.  We add tables dynamically at runtime, and the MetaData object is shared by all Database instances sharing the same connection pool.

Prior to adding the lock, Butler server database calls that added table definitions dynamically were sometimes failing with InvalidRequestError exceptions complaining about inconsistency of table definitions.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/interfaces/_database.py
+++ b/python/lsst/daf/butler/registry/interfaces/_database.py
@@ -856,30 +856,6 @@ class Database(ABC):
         """
         return shrunk
 
-    def _mangleTableName(self, name: str) -> str:
-        """Map a logical, user-visible table name to the true table name used
-        in the database.
-
-        The default implementation returns the given name unchanged.
-
-        Parameters
-        ----------
-        name : `str`
-            Input table name.  Should not include a namespace (i.e. schema)
-            prefix.
-
-        Returns
-        -------
-        mangled : `str`
-            Mangled version of the table name (still with no namespace prefix).
-
-        Notes
-        -----
-        Reimplementations of this method must be idempotent - mangling an
-        already-mangled name must have no effect.
-        """
-        return name
-
     def _makeColumnConstraints(self, table: str, spec: ddl.FieldSpec) -> list[sqlalchemy.CheckConstraint]:
         """Create constraints based on this spec.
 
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -164,13 +165,9 @@ def addTable(self, name: str, spec: ddl.TableSpec) -> sqlalchemy.schema.Table:
         to be declared in any order even in the presence of foreign key
         relationships.
         """
-        name = self._db._mangleTableName(name)
```
andy-slac, line 167: I am confused, do you want `DatabaseMetadata` to handle table name mangling internally? I do not think it is done consistently now, you call `add_table` with the original name, but `get_table` is called with mangled name?

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -1192,7 +1189,7 @@ def getExistingTable(self, name: str, spec: ddl.TableSpec) -> sqlalchemy.schema.
         """
         assert self._metadata is not None, "Static tables must be declared before dynamic tables."
         name = self._mangleTableName(name)
```
andy-slac: Maybe this one needs to be dropped too?

python/lsst/daf/butler/registry/interfaces/_database.py
```diff
@@ -1206,10 +1203,7 @@ def getExistingTable(self, name: str, spec: ddl.TableSpec) -> sqlalchemy.schema.
             )
             if name in inspector.get_table_names(schema=self.namespace):
```
andy-slac, line 1176: Hmm, but this won't work if name is not mangled?

andy-slac (approved):
Looks good in general, but I'm not sure that name mangling is done consistently.

python/lsst/daf/butler/registry/interfaces/_database.py
dhirving: I am just going to remove it entirely.  It's a no-op -- the only implementation returns the name unchanged.  I think it is a holdover from the Oracle implementation.  

The existing implementation was already incorrect because there were several places where name mangling was applied multiple times.  I noticed that the two cases I already removed were double-applying the mangling (because it is done internally in _convertTableSpec) so I had removed those, but apparently the logic is also wrong in a lot of other places.

TallJimbo: My recollection is that it was PostgreSQL, not Oracle, that needed name shrinking.  It may have been index and constraint names more than table names where it was important, but I'm pretty sure PostgreSQL is the one with a 64-char limit.

dhirving: Two different things.  The name shrinking logic is still there.

TallJimbo: Ah, sorry for the noise.

PR #1135, opened 2024-12-14
DM-47980: Add an option to include dimension records into general query result

andy-slac:
The `GeneralQueryResults.iter_tuples` method returned DataIds without dimension records. In some cases (e.g. for obscore export) it would be very useful to include records in the same result to avoid querying them separately. New method `with_dimension_records` is added to the class to trigger adding fields from all dimension records into returned page. This will produce many duplicates for some dimensions (e.g. `instrument`) but it keeps page structure simple.

This adds one attribute to the `GeneralResultSpec` class, will need some care with Butler server compatibility.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_query_driver/_result_page_converter.py
+++ b/python/lsst/daf/butler/direct_query_driver/_result_page_converter.py
@@ -330,25 +330,14 @@ class GeneralResultPageConverter(ResultPageConverter):  # numpydoc ignore=PR01
         # columns returned by the query we have to add columns from dimension
         # records that are not returned by the query. These columns belong to
         # either cached or skypix dimensions.
-        query_result_columns = set(spec.get_result_columns())
-        output_columns = spec.get_all_result_columns()
+        columns = spec.get_result_columns()
         universe = spec.dimensions.universe
         self.converters: list[_GeneralColumnConverter] = []
-        for column in output_columns:
+        self.record_converters: dict[DimensionElement, _DimensionRecordRowConverter] = {}
+        for column in columns:
             column_name = qt.ColumnSet.get_qualified_name(column.logical_table, column.field)
             converter: _GeneralColumnConverter
-            if column not in query_result_columns and column.field is not None:
-                # This must be a field from a cached dimension record or
-                # skypix record.
-                assert isinstance(column.logical_table, str), "Do not expect AnyDatasetType here"
-                element = universe[column.logical_table]
-                if isinstance(element, SkyPixDimension):
-                    converter = _SkypixRecordGeneralColumnConverter(element, column.field)
-                else:
-                    converter = _CachedRecordGeneralColumnConverter(
-                        element, column.field, ctx.dimension_record_cache
-                    )
-            elif column.field == TimespanDatabaseRepresentation.NAME:
+            if column.field == TimespanDatabaseRepresentation.NAME:
                 converter = _TimespanGeneralColumnConverter(column_name, ctx.db)
             elif column.field == "ingest_date":
                 converter = _TimestampGeneralColumnConverter(column_name)
@@ -356,9 +345,29 @@ class GeneralResultPageConverter(ResultPageConverter):  # numpydoc ignore=PR01
                 converter = _DefaultGeneralColumnConverter(column_name)
             self.converters.append(converter)
 
+        if spec.include_dimension_records:
+            universe = self.spec.dimensions.universe
+            for element_name in self.spec.dimensions.elements:
+                element = universe[element_name]
+                if isinstance(element, SkyPixDimension):
+                    self.record_converters[element] = _SkypixDimensionRecordRowConverter(element)
+                elif element.is_cached:
+                    self.record_converters[element] = _CachedDimensionRecordRowConverter(
+                        element, ctx.dimension_record_cache
+                    )
+
     def convert(self, raw_rows: Iterable[sqlalchemy.Row]) -> GeneralResultPage:
-        rows = [tuple(cvt.convert(row) for cvt in self.converters) for row in raw_rows]
-        return GeneralResultPage(spec=self.spec, rows=rows)
+        rows = []
+        dimension_records = None
+        if self.spec.include_dimension_records:
+            dimension_records = {element: DimensionRecordSet(element) for element in self.record_converters}
+        for row in raw_rows:
+            rows.append(tuple(cvt.convert(row) for cvt in self.converters))
+            if dimension_records:
+                for element, converter in self.record_converters.items():
+                    dimension_records[element].add(converter.convert(row))
+
+        return GeneralResultPage(spec=self.spec, rows=rows, dimension_records=dimension_records)
 
 
 class _GeneralColumnConverter:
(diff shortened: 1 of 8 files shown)


python/lsst/daf/butler/direct_query_driver/_result_page_converter.py
```diff
@@ -326,17 +326,35 @@ class GeneralResultPageConverter(ResultPageConverter):  # numpydoc ignore=PR01
 
     def __init__(self, spec: GeneralResultSpec, ctx: ResultPageConverterContext) -> None:
         self.spec = spec
-
-        result_columns = spec.get_result_columns()
+        # In case `spec.include_dimension_records` is True then in addition to
+        # columns returned by the query we have to add columns from dimension
+        # records that are not returned by the query. These columns belong to
+        # either cached or skypix dimensions.
```
TallJimbo, line 332: Why do you need to add the cached/skypix columns to the list of column converters?  I'd have thought you could just load entire records from the cache (or construct them on-the fly for skypix) when you expand the data IDs.  That is what the `DataCoordinate` results converter does, and by extension the `DatasetRef` one as well.

Is there a need to have the cached/skypix dimension record fields accessible outside of `iter_tuples`?

python/lsst/daf/butler/queries/_general_query_results.py
```diff
@@ -125,14 +125,10 @@ def iter_tuples(self, *dataset_types: DatasetType) -> Iterator[GeneralResultTupl
...
-            values = tuple(
-                row[key] for key in itertools.chain(all_dimensions.required, all_dimensions.implied)
-            )
-            data_coordinate = DataCoordinate.from_full_values(all_dimensions, values)
+            data_coordinate = self._make_data_id(row, all_dimensions)
             refs = []
             for dataset_type, dimensions, id_key, run_key in dataset_keys:
-                values = tuple(row[key] for key in itertools.chain(dimensions.required, dimensions.implied))
-                data_id = DataCoordinate.from_full_values(dimensions, values)
+                data_id = self._make_data_id(row, dimensions)
```
TallJimbo: Instead of making a brand new data ID, it should be a bit more efficient (especially in memory use) to do:
```
data_id = data_coordinate.subset(dimensions)
```
since we know we've already got a data ID with a superset of these dimensions.

python/lsst/daf/butler/remote_butler/server/handlers/_query_serialization.py
```diff
@@ -79,11 +79,12 @@ def convert_query_page(spec: ResultSpec, page: ResultPage) -> QueryExecuteResult
...
     serializers = [
         columns.get_column_spec(column.logical_table, column.field).serializer() for column in columns
     ]
     rows = [
-        tuple(serializer.serialize(value) for value, serializer in zip(row, serializers)) for row in page.rows
+        tuple(serializer.serialize(value) for value, serializer in zip(row, serializers, strict=True))
+        for row in page.rows
     ]
-    return GeneralResultModel(rows=rows)
+    return GeneralResultModel(rows=rows, columns=[str(column) for column in columns])
```
TallJimbo: Oh, are you getting all result columns (including the cached/skypix ones) because you don't want/expect the `RemoteButler` client to have those caches populated?  I think we do...that's a lot of network traffic we can just avoid, and the dimension record cache is the least problematic one we've got in terms of how to manage expiration.

TallJimbo (approved):
I think you've made this more difficult and less memory-efficient than it needed to be by making the cached/skypix columns appear like regular result columns (a couple of line comments on that subject), but it's possible I just didn't see the motivation.  Looks good aside from that question.

python/lsst/daf/butler/direct_query_driver/_result_page_converter.py
andy-slac: My idea was that when a client explicitly asks for dimension records then all their columns for all dimensions need to be present in the returned rows.  Handling it differently between cached and non-cached dimensions will probably complicate client code when that code wants to look at individual fields (as opposed to the DatasetRefs returned from iter_tuples). Potentially we can avoid adding those columns to the GeneralResultPage, and make GeneralResult to add them during iteration, but that would require additional round trips for filling caches in case of remote butler.

python/lsst/daf/butler/remote_butler/server/handlers/_query_serialization.py
andy-slac: Indeed, I though avoiding the need for client-side cache would be nice 🙂 But it definitely increases data size. Let me think about it again, I'll probably revert to getting those records from cache.

andy-slac: OK, I remember the reason now why I did not want to use cached dimension records inside `GeneralResult` - the cache does not exist at that level. `DimensionRecordCache` is instantiated now only for direct butler (SqlRegistry more specifically) and is passed to `DirectQueryDriver`. Remote butler does not make `DimensionRecordCache` and remote query driver does not do it either. I guess none of the result classes so far needed that cache. @dhirving, what are your thoughts on that, should we add dimension records cache to remote butler?

dhirving: I prefer not to have that cache on the client side.  Even if the entire contents of the cache had to be sent on every request, it is less than 4k gzipped.  So we're not exactly talking about a ton of data.  To me the added complexity of maintaining the cache and embedding the concept of governor dimensions into the client/server protocol doesn't seem worth that small of a savings.  

It's not so bad having that cache on our server because we can restart the server if it changes, but I think we'd want some way to invalidate it if we had it on the client side.  If we didn't, we'd have to track down and restart every service using the client too.  This wouldn't be that hard -- we could send a hash with every response that implies a particular set of cached dimension records.  But there are lots of ways to mess that up, so why open that can of worms to save (at worst) 4k.

For the other result specs, my plan was to use the containers from DM-41114 to pull out the repeated dimension records.  I think the governor dimensions could be treated exactly the same as other dimension records within those containers, and the client side doesn't need to know whether they came from the cache or the DB.  Not sure how well that will work in the context of GeneralResultSpec, though.

TallJimbo: The potential savings here is not the same as the size of the cache - in the cache, the data is normalized, while pulling the values from the cached dimension records out into the main rows duplicates them for each row.  And that can be a really big factor, like repeating an instrument or filter name `n_detectors * n_visits` times.

I may still be assuming the cache is more important than it is, because my intuition on this is based on stuff in QG serialization that isn't quite like-for-like.

TallJimbo: Ah, and David already mentioned the possibility of sending the cache down on every request - so maybe we should just do that, at least for now.

dhirving: Right, that's why I mentioned the container classes that normalize the records.  It gives you most of the benefit without the need for an explicit cache (or in some cases a greater benefit, since it would also help with non-governor dimensions).  Not clear how that would fit in with GeneralResultSpec where you are basically trying to ship a lightly transformed version of the raw database row, though.

In my opinion it wouldn't be wise to go too wild optimizing this right now.  My guess would be we're going to see much more of a slowdown in SIA from the region queries than we are from serialization overhead, though a profile might prove me very wrong.

dhirving: I guess to give a more complete top-down answer:

This feature is being added to support an experimental service which is not planned for public release in DP1 -- and which may not survive the experiment phase.  Given that it's an experiment, doing the simplest thing that could possibly work is reasonable -- and I think that's what Andy's existing code does.

Any speedup from the dimension cache is probably linear with the number of results, so there's not an obvious scalability benefit to the cache.  And in any case this is orders of magnitude faster than the previous iteration, which downloaded half the database on every query.

I also think that once this merges down and we get to the next step, it is likely to reveal additional problems that may end up on Andy's plate.  So I personally would wait a minute before spending time optimizing this.  

We might want to wait on making the new API public to avoid needing to think much about backwards compatibility, though.  It is very easy to ensure a simultaneous release of SIA and Butler server since they're deployed through the same system and Stelios' desk is 20 feet from mine.

TallJimbo: I'm fine with punting any optimization of this and going with the code as is, especially if we can try to avoid establishing an interface that limits our options in the future.

TallJimbo (approved):

PR #1138, opened 2025-01-03
DM-48288: Do not cache collection summaries by default in queries

dhirving:
We now enable the collection summary cache separately from the collection record cache.  In the Butler.query() context, only the collection record cache is now enabled by default.  When Butler.registry.caching_context() is called, both caches are enabled (as before.)

More often than not, the query context manager is used to execute one query and then exited.  The collection summary cache was often a pessimization in this case -- fetching tens of thousands of rows when only a few were needed.

Also added explicit caching context calls to a few places in the Butler internals where we are doing repeated dataset queries that will benefit from the cache.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/_caching_context.py
+++ b/python/lsst/daf/butler/registry/_caching_context.py
@@ -27,6 +27,10 @@
 
 from __future__ import annotations
 
+from collections.abc import Callable, Iterator
+from contextlib import AbstractContextManager, contextmanager
+from typing import Generic, TypeVar
+
 __all__ = ["CachingContext"]
 
 from ._collection_record_cache import CollectionRecordCache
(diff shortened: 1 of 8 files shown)


python/lsst/daf/butler/registry/_caching_context.py
```diff
@@ -48,46 +52,70 @@ class is passed to the relevant managers that can use it to query or
...
         """
-        if self._depth == 1:
-            self._collection_records = None
-            self._collection_summaries = None
-            self._depth = 0
-        elif self._depth > 1:
-            self._depth -= 1
-        else:
-            raise AssertionError("Bad caching context management detected.")
+        return self._collection_records.enable()
```
andy-slac: Should it be `_collection_summaries`?

andy-slac (approved):
Looks good, one comment.

python/lsst/daf/butler/registry/_caching_context.py
dhirving: .... yes it should.

PR #1142, opened 2025-01-15
DM-48282: Refactor obscore to allow subclasses

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/obscore/_records.py
+++ b/python/lsst/daf/butler/registry/obscore/_records.py
@@ -33,12 +33,13 @@ import logging
 import warnings
 from abc import abstractmethod
 from collections.abc import Callable, Collection, Mapping
+from importlib.metadata import entry_points
 from typing import TYPE_CHECKING, Any, cast
 from uuid import UUID
 
 import astropy.time
 from lsst.daf.butler import DataCoordinate, DatasetRef, Dimension, DimensionRecord, DimensionUniverse
-from lsst.utils.introspection import find_outside_stacklevel
+from lsst.utils.introspection import find_outside_stacklevel, get_full_type_name
 
 from ._config import ExtraColumnConfig, ExtraColumnType, ObsCoreConfig
 from ._spatial import RegionTypeError, RegionTypeWarning
@@ -46,6 +47,7 @@ from ._spatial import RegionTypeError, RegionTypeWarning
 if TYPE_CHECKING:
     from lsst.sphgeom import Region
 
+    from ._config import DatasetTypeConfig
     from ._schema import ObsCoreSchema
     from ._spatial import SpatialObsCorePlugin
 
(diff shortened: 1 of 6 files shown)


python/lsst/daf/butler/registry/obscore/_records.py
```diff
@@ -96,7 +99,7 @@ class RecordFactory:
         Registry dimensions universe.
     spatial_plugins : `~collections.abc.Collection` of `SpatialObsCorePlugin`
         Spatial plugins.
-    exposure_region_factory : `ExposureRegionFactory`, optional
+    derived_region_factory : `DerivedRegionFactory`, optional
         Manager for Registry dimensions.
```
andy-slac: Looks like I copy-pasted this line long time ago and did not update, can you add a better description?

python/lsst/daf/butler/registry/obscore/_records.py
```diff
@@ -120,15 +123,122 @@ def __init__(
         self.visit = universe["visit"]
         self.physical_filter = cast(Dimension, universe["physical_filter"])
 
+    @abstractmethod
+    def region_dimension(self, dimensions: DimensionGroup) -> tuple[str | None, str | None]:
```
andy-slac: Could just one of them be `None`? If not then `tuple[str, str] | tuple[None, None]` could be better?

python/lsst/daf/butler/registry/obscore/_records.py
```diff
@@ -160,15 +269,127 @@ def __call__(self, ref: DatasetRef) -> Record | None:
...
+                # Clear the record.
+                record = Record()
+            else:
+                if msg:
+                    exc.add_note(msg)
+                raise
+        return record
+
+    @classmethod
+    def get_record_type_from_universe(cls, universe: DimensionUniverse) -> type[RecordFactory]:
```
andy-slac, line 330: We could potentially specify the type in the obscore config itself?

python/lsst/daf/butler/registry/obscore/_records.py
```diff
@@ -160,15 +269,127 @@ def __call__(self, ref: DatasetRef) -> Record | None:
...
+    config : `ObsCoreConfig`
+        Complete configuration specifying conversion options.
+    schema : `ObsCoreSchema`
+        Description of obscore schema.
+    universe : `DimensionUniverse`
+        Registry dimensions universe.
+    spatial_plugins : `~collections.abc.Collection` of `SpatialObsCorePlugin`
+        Spatial plugins.
+    derived_region_factory : `DerivedRegionFactory`, optional
+        Manager for Registry dimensions.
```
andy-slac: This too wants an update.

python/lsst/daf/butler/registry/obscore/_records.py
```diff
@@ -160,15 +269,127 @@ def __call__(self, ref: DatasetRef) -> Record | None:
...
+            return DafButlerRecordFactory
+        # Check for entry points.
+        plugins = {p.name: p for p in entry_points(group="butler.obscore_factory")}
+        if namespace in plugins:
+            func = plugins[namespace].load()
+            # The entry point function is required to return the class that
+            # should be used for the RecordFactory for this universe.
+            record_factory_type = func()
+            if not issubclass(record_factory_type, RecordFactory):
+                raise ValueError(
```
andy-slac: Maybe `TypeError`?

andy-slac (approved):
Looks good. One thing that I'm not sure about is whether it's easier to specify record factory type in obscore config vs. entry point.

python/lsst/daf/butler/registry/obscore/_records.py
timj: This was discussed on Jira but replying here for completeness. Config class description works well if there are many choices that are effectively independent of universe. For a case where we know that there is going to be one RecordFactory per universe it makes sense to me to use entry points rather than requiring that every configuration file specify the class name.

PR #1146, opened 2025-01-23
DM-46248: Few small improvements for the new query system

andy-slac:
## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_query_driver/_driver.py
+++ b/python/lsst/daf/butler/direct_query_driver/_driver.py
@@ -217,6 +217,7 @@ class DirectQueryDriver(QueryDriver):
             final_columns=result_spec.get_result_columns(),
             order_by=result_spec.order_by,
             find_first_dataset=result_spec.find_first_dataset,
+            allow_duplicate_overlaps=result_spec.allow_duplicate_overlaps,
         )
         sql_select, sql_columns = builder.finish_select()
         if result_spec.order_by:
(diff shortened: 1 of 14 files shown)


dhirving (commented):
The client-server stuff and model changes look fine to me.

python/lsst/daf/butler/direct_query_driver/_driver.py
```diff
@@ -311,7 +311,9 @@ def materialize(
         #
         sql_select, _ = plan.finish_select(return_columns=False)
         table = self._exit_stack.enter_context(
-            self.db.temporary_table(make_table_spec(plan.final_columns, self.db, plan.postprocessing))
+            self.db.temporary_table(
+                make_table_spec(plan.final_columns, self.db, plan.postprocessing, make_indices=True)
+            )
```
TallJimbo, line 320: I recall you saying earlier that adding the indexes before doing the inserts sometimes made the inserts much more expensive.  Did that get resolved by other changes, or is it just an overall win even if it's occasionally worse.

And would it be worth it to try to move index creation to just after insert in this method?

TallJimbo (approved):

python/lsst/daf/butler/direct_query_driver/_driver.py
andy-slac: That was a false alarm - I have added indices and filling temp table became terribly slow, the real reason was that query planner went into a dumb state on the same exact query (this was before I dropped DISTINCT). With disabled DISTINCT I have not noticed any significant slow down due to indices. 
It is true that for bulk data insert it is better to load the data before adding indices. I tried to implement it, but it is not trivial with our code now, the simple idea did not work, and I did not want to start restructuring a bunch of code just for that. Still, if we see it later, it should be possible to delay index creation until after filling the data.

andy-slac: Just in case we want to do index creation separately later - my initial idea was that I run the same `_convertTableSpec` but do not add `Index` instances to `Table` constructor arguments, instead the indices are returned as a separate list so you can later iterate and create them manually. It happens, that even if you do not add an `Index` to the `Table()` arguments, the metadata already knows about all indices for that table (as `Index` is created for a specific table) and will create that index anyways. So delaying just did not work.

PR #1147, opened 2025-01-23
DM-35396: Add provenance hooks for datastore to use

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_dataset_provenance.py
+++ b/python/lsst/daf/butler/_dataset_provenance.py
@@ -1,0 +1,67 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This software is dual licensed under the GNU General Public License and also
+# under a 3-clause BSD license. Recipients may choose which of these licenses
+# to use; please see the files gpl-3.0.txt and/or bsd_license.txt,
+# respectively.  If you choose the GPL option then the following text applies
+# (but note that there is still no warranty even if you opt for BSD instead):
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import annotations
+
+__all__ = ("DatasetProvenance",)
+
+import typing
+import uuid
+
+import pydantic
+
+from ._dataset_ref import DatasetRef, SerializedDatasetRef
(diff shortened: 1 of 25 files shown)


TallJimbo (approved):

python/lsst/daf/butler/_quantum_backed.py
```diff
@@ -453,11 +454,11 @@ def dimensions(self) -> DimensionUniverse:
         # Docstring inherited.
         return self._dimensions
 
-    def put(self, obj: Any, ref: DatasetRef, /) -> DatasetRef:
+    def put(self, obj: Any, ref: DatasetRef, /, provenance: DatasetProvenance | None = None) -> DatasetRef:
```
andy-slac:
```suggestion
    def put(self, obj: Any, ref: DatasetRef, /, *, provenance: DatasetProvenance | None = None) -> DatasetRef:
```
to make it consistent with LimitedButler?

python/lsst/daf/butler/datastore/_datastore.py
```diff
@@ -1449,7 +1455,9 @@ def get(
     ) -> Any:
         raise FileNotFoundError("This is a no-op datastore that can not access a real datastore")
 
-    def put(self, inMemoryDataset: Any, datasetRef: DatasetRef) -> None:
+    def put(
+        self, inMemoryDataset: Any, datasetRef: DatasetRef, provenance: DatasetProvenance | None = None
+    ) -> None:
         raise NotImplementedError("This is a no-op datastore that can not access a real datastore")
 
     def put_new(self, in_memory_dataset: Any, ref: DatasetRef) -> Mapping[str, DatasetRef]:
```
andy-slac, line 1463: Hmm, we should probably drop `put_new` if we are not using it. It's not tested and is probably broken already. Not on this ticket, of course.

python/lsst/daf/butler/_formatter.py
```diff
@@ -169,6 +174,7 @@ def __init__(
         ref: DatasetRef,
         write_parameters: Mapping[str, Any] | None = None,
         write_recipes: Mapping[str, Any] | None = None,
+        provenance: DatasetProvenance | None = None,
```
andy-slac: Does it need to be a constructor argument, could you pass it to `write` instead?

python/lsst/daf/butler/_dataset_provenance.py
```diff
@@ -65,3 +67,20 @@ def add_input(self, ref: DatasetRef) -> None:
...
+            The ID of the dataset to receive this provenance.
+        extra : `dict` [ `str`, `typing.Any` ]
+            The extra provenance information as a dictionary. The values
+            must be simple Python scalars.
+        """
+        if dataset_id not in self._uuids:
+            raise ValueError(f"The given dataset ID {dataset_id} is not known to this provenance instance.")
+        if dataset_id not in self.extras:
+            self.extras[dataset_id] = {}
+        self.extras[dataset_id].update(extra)
```
andy-slac: Maybe 

    self.extras.setdefault(dataset_id, {}).update(extra)

?

andy-slac (approved):
Looks great, a couple of questions.

python/lsst/daf/butler/_formatter.py
timj: I think when I wrote it originally I was thinking that provenance was similar to the `ref` parameter. Also that it would be quite disruptive to pass an extra parameter through the write APIs. Now that I ended up with `Formatter.write()` calling `add_provenance` directly without it needing to be passed down through to the user-supplied write methods it is a fairly simple modification to pass it to `write()`.

PR #1149, opened 2025-01-29
DM-46178: Documentation fixes for new query system

kfindeisen:
This PR addresses various documentation issues that came up while working on lsst-dm/prompt_processing#249.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/queries/_base.py
+++ b/python/lsst/daf/butler/queries/_base.py
@@ -138,10 +138,10 @@ class QueryBase(ABC):
 
         Notes
         -----
-        If an expression references a dimension or dimension element that is
-        not already present in the query, it will be joined in, but dataset
-        searches must already be joined into a query in order to reference
-        their fields in expressions.
+        Expressions referring to dimensions or dimension elements are resolved
+        automatically. References to dataset fields (see `expression_factory`
+        for the distinction) may or may not be resolvable, depending on the
+        implementation class.
 
         Data ID values are not checked for consistency; they are extracted from
         ``args`` and then ``kwargs`` and combined, with later values overriding
--- a/python/lsst/daf/butler/queries/_base.py
+++ b/python/lsst/daf/butler/queries/_base.py
@@ -41,16 +41,17 @@ from .tree import OrderExpression, Predicate, QueryTree
 
 
 class QueryBase(ABC):
-    """Common base class for `Query` and all `QueryResult` objects.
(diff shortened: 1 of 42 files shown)


python/lsst/daf/butler/queries/_base.py
```diff
@@ -242,7 +246,7 @@ def where(
 
     @abstractmethod
     def _get_datasets(self) -> Set[str]:
-        """Return all dataset types included in the query's result rows."""
+        """Return all datasets included in the query's result rows."""
```
TallJimbo:
```suggestion
        """Return all dataset type names included in the query's result rows."""
```
Method name notwithstanding, these are more dataset types (they're the names of them) than datasets.

TallJimbo (approved):
For future reference, I think I prefer Sphinx's relative-name syntax (leading `.` for each parent level to move up) to qualifying everything with absolute module names, and I don't consider being able to reference a name in the exact same package without qualification to be lucky; it's correctly concise and just about the only part of Sphinx's name resolution rules/syntax that works like it should.

But this is a big improvement over the previous situation that you didn't have to contribute to get your work done, so I'm certainly not going to ask you to redo it here!

This might also be a good time to see if @timj, @dhirving, and @andy-slac have different opinions on how to handle namespaced references in docs.

kfindeisen:
> For future reference, I think I prefer Sphinx's relative-name syntax (leading . for each parent level to move up) to qualifying everything with absolute module names, and I don't consider being able to reference a name in the exact same package without qualification to be lucky; it's correctly concise and just about the only part of Sphinx's name resolution rules/syntax that works like it should.

The problem is that this is not a matter of "preference" -- the Sphinx links simply do not, and _cannot_, work the way you seem to think they do. The Sphinx parser does not have access to the source code, and evaluates relative links based on the package in which the component is _documented_. If a component is lifted into a higher-level namespace (a decision that often happens independently of writing the docstrings), links are relative to that. If it is documented in both high-level and low-level namespaces (as is common in MW), then any relative links will work in at most one of those places.

So I stand by my statement that the working links were such only through luck: many, if not most, of the links I fixed look like the writer assumed that the referenced component was in the same package when, in the context of the documentation, it wasn't.

TallJimbo:
I'm aware the syntax and lookup behavior is not the same as Python's actual import system (which is mind-bogglingly frustrating), but there _is_ a relative syntax that does work, if you take into account where things are documented rather than where they are declared.

PR #1151, opened 2025-02-06
DM-47201: Fix duplicates in non-find-first dataset search

dhirving:
Fix an issue where duplicate results could appear in a non-find-first dataset search, if the same dataset appeared in multiple collections in a chain.

This was occurring because we were forcing the addition of the collection key field to make the rows distinct.  But on a non-find-first search, we don't have the window function to de-duplicate the rows by dataset ID.  So we need to:
1.  Keep the collection key out of the rows (because this is preventing rows from being de-duplicated) 
2. Treat dataset ID as a unique key instead (so that we don't drop rows with the same data ID but different dataset IDs).

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_query_driver/_query_builder.py
+++ b/python/lsst/daf/butler/direct_query_driver/_query_builder.py
@@ -78,6 +78,10 @@ class QueryBuilder(ABC):
         or DISTINCT may be performed.
     final_columns : `.queries.tree.ColumnSet`
         Columns to include in the final query.
+    find_first_dataset : `str` or ``...`` or None
+        Name of the dataset type that needs a find-first search.  ``...``
+        is used to indicate the dataset types in a union dataset query.
+        `None` means find-first is not used.
     """
 
     def __init__(
@@ -86,13 +90,14 @@ class QueryBuilder(ABC):
         *,
         projection_columns: qt.ColumnSet,
         final_columns: qt.ColumnSet,
+        find_first_dataset: str | qt.AnyDatasetType | None,
     ):
         self.joins_analysis = tree_analysis.joins
         self.postprocessing = tree_analysis.postprocessing
         self.projection_columns = projection_columns
         self.final_columns = final_columns
         self.needs_dimension_distinct = False
-        self.find_first_dataset = None
+        self.find_first_dataset = find_first_dataset
 
     joins_analysis: QueryJoinsAnalysis
     """Description of the "joins" stage of query construction."""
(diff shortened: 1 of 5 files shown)


dhirving:
@TallJimbo I'm not quite done with this, but could you take a quick look and let me know if this seems like I'm solving the right problem?  

It seems excessively complicated to me but I haven't thought of a better way to handle this yet -- I'll take a closer look in the morning with fresh eyes.

(I will also be adding more tests to this -- we don't have sufficient coverage around non-trivial non-find-first queries.)

dhirving:
By the way, I had mentioned that dataset fields were missing from the joins stage... it turned out to not be relevant because `needs_dataset_distinct` wasn't actually causing the issue.  The fields were missing in `analyze_projection` because in `DirectQueryDriver.build_query()`, we don't add the fields from the projection stage until after we analyze the projection:

```
         # Finish setting up the projection part of the builder.
        builder.analyze_projection()
        # The joins-stage query also needs to include all columns needed by the
        # downstream projection query.  Note that this:
        # - never adds new dimensions to the joins stage (since those are
        #   always a superset of the projection-stage dimensions);
        # - does not affect our previous determination of
        #   needs_dataset_distinct, because any dataset fields being added to
        #   the joins stage here are already in the projection.
        builder.joins_analysis.columns.update(builder.projection_columns)
```

python/lsst/daf/butler/direct_query_driver/_query_builder.py
```diff
@@ -284,6 +283,47 @@ def finish_nested(self, cte: bool = False) -> SqlSelectBuilder:
...
+                # We might have multiple calibration collections containing the
+                # same dataset with the same timespan.
+                or "timespan" in fields_for_dataset
+                # The user specifically asked for a row for each collection we
+                # found the dataset in.
+                or "collection" in fields_for_dataset
+            ):
+                return True
+
+        return False
```
TallJimbo, line 325: Conceptually this method feels like it would fit well as something we would call to set an attribute on `ResolvedDatasetSearch`, but I think to do that it'd have to be near the end of the "analysis" phase rather than the beginning, where the `ResolvedDatasetSearch` is constructed, so I'm not sure it's worth doing.

TallJimbo (approved):
This all looks good to me (sorry I lost track of it so long; Tim just reminded me about it yesterday); reasoning about what's going on makes sense and I don't see an obvious simpler solution.

python/lsst/daf/butler/direct_query_driver/_query_builder.py
dhirving: Good timing... I was planning to circle back and finish this up in the next week here.  I am going to add some more tests before merging this because we don't have much coverage on non-find-first queries (especially around things like calibration timespans) and I'm concerned this might interact in weird ways.

PR #1154, opened 2025-02-11
DM-48880: avoid too-coarse spatial joins in queries

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/queries/overlaps.py
+++ b/python/lsst/daf/butler/queries/overlaps.py
@@ -278,7 +278,7 @@ class OverlapsVisitor(SimplePredicateVisitor):
         association between non-overlapping things that we'd want to respect by
         *not* adding a more restrictive automatic join.
         """
-        for a_family, b_family in itertools.pairwise(families):
+        for a_family, b_family in itertools.combinations(families, 2):
             a_element = a_family.choose(self.dimensions)
             b_element = b_family.choose(self.dimensions)
             if (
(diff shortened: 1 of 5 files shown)


python/lsst/daf/butler/queries/overlaps.py
```diff
@@ -259,10 +277,14 @@ def add_join_operand_connections(self, operand_dimensions: DimensionGroup) -> No
         association between non-overlapping things that we'd want to respect by
         *not* adding a more restrictive automatic join.
         """
-        for a_family, b_family in itertools.pairwise(operand_dimensions.spatial):
-            self._spatial_connections.merge(a_family, b_family)
-        for a_family, b_family in itertools.pairwise(operand_dimensions.temporal):
-            self._temporal_connections.merge(a_family, b_family)
+        for a_family, b_family in itertools.pairwise(getattr(operand_dimensions, kind)):
```
andy-slac: Can you pass `operand_dimensions.spatial` or `.temporal` into this method instead of doing `getattr`? It might help `mypy` to figure out types of those?

andy-slac (approved):
Looks great.

python/lsst/daf/butler/queries/overlaps.py
TallJimbo: Done.  I also noticed along the way that we've been using `pairwise` here but it really should be `combinations`.

PR #1155, opened 2025-02-12
DM-48869: Add method to return dataset provenance in a flat dict

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_dataset_provenance.py
+++ b/python/lsst/daf/butler/_dataset_provenance.py
@@ -110,7 +110,8 @@ class DatasetProvenance(pydantic.BaseModel):
             A prefix to use for each key in the provenance dictionary.
         sep : `str`, optional
             Separator to use to represent hierarchy. Must be a single
-            character.
+            character. Can not be a number, letter, or underscore (to avoid
+            confusion with provenance keys themselves).
         simple_types : `bool`, optional
             If `True` only simple Python types will be used in the returned
             dictionary, specifically UUIDs will be returned as `str`. If
@@ -169,6 +170,12 @@ class DatasetProvenance(pydantic.BaseModel):
         ValueError
             Raised if the separator is not a single character.
         """
+        if len(sep) != 1:
+            raise ValueError(f"Separator for provenance keys must be a single character. Got {sep!r}.")
+        if re.match(r"[_\w\d]$", sep):
+            raise ValueError(
+                f"Separator for provenance keys can not be word character or underscore. Got {sep!r}."
+            )
 
         def _make_key(*keys: str | int) -> str:
             """Make the key in the correct form with simpler API."""
@@ -221,8 +228,6 @@ class DatasetProvenance(pydantic.BaseModel):
             prefix (defaulting to lower case if the first character of
             prefix has no case).
         """
-        if len(sep) != 1:
-            raise ValueError(f"Separator for provenance keys must be a single character. Got {sep!r}.")
         use_upper = prefix[0].isupper() if prefix else False
         if prefix:
             prefix += sep
@@ -348,16 +353,19 @@ class DatasetProvenance(pydantic.BaseModel):
 
         core_provenance = tuple(f"{prefix}{k}".lower() for k in ("run", "id", "datasettype", "quantum"))
 
+        # Need to escape the prefix and separator for regex usage.
+        esc_sep = re.escape(sep)
+        esc_prefix = re.escape(prefix)
         prov_keys: dict[str, str] = {}
         for k in list(prov_dict):
             # the input provenance can include extra keys that we cannot
             # know so have to match solely on INPUT N.
             found_key = False
-            if re.match(rf"{prefix}input{sep}(\d+){sep}(.*)$", k, flags=re.IGNORECASE):
+            if re.match(rf"{esc_prefix}input{esc_sep}(\d+){esc_sep}(.*)$", k, flags=re.IGNORECASE):
                 found_key = True
             elif k.lower() in core_provenance:
                 found_key = True
-            elif re.match(f"{prefix}dataid{sep}[a-z_]+$", k, flags=re.IGNORECASE):
+            elif re.match(f"{esc_prefix}dataid{esc_sep}[a-z_]+$", k, flags=re.IGNORECASE):
                 found_key = True
 
             if found_key:
--- a/python/lsst/daf/butler/_dataset_provenance.py
+++ b/python/lsst/daf/butler/_dataset_provenance.py
@@ -29,12 +29,15 @@ from __future__ import annotations
 
 __all__ = ("DatasetProvenance",)
 
-import typing
 import uuid
+from typing import TYPE_CHECKING, Any, Self
 
 import pydantic
 
 from ._dataset_ref import DatasetRef, SerializedDatasetRef
+
+if TYPE_CHECKING:
+    from .dimensions import DataIdValue
 
 
 class DatasetProvenance(pydantic.BaseModel):
@@ -44,12 +47,12 @@ class DatasetProvenance(pydantic.BaseModel):
     """The input datasets."""
     quantum_id: uuid.UUID | None = None
     """Identifier of the Quantum that was executed."""
-    extras: dict[uuid.UUID, dict[str, int | float | str | bool]] = pydantic.Field(default_factory=dict)
+    extras: dict[uuid.UUID, dict[str, Any]] = pydantic.Field(default_factory=dict)
     """Extra provenance information associated with a particular dataset."""
     _uuids: set[uuid.UUID] = pydantic.PrivateAttr(default_factory=set)
 
     @pydantic.model_validator(mode="after")
-    def populate_cache(self) -> typing.Self:
+    def populate_cache(self) -> Self:
         for ref in self.inputs:
             self._uuids.add(ref.id)
         return self
@@ -68,7 +71,7 @@ class DatasetProvenance(pydantic.BaseModel):
         self._uuids.add(ref.id)
         self.inputs.append(ref.to_simple())
 
-    def add_extra_provenance(self, dataset_id: uuid.UUID, extra: dict[str, int | float | str | bool]) -> None:
+    def add_extra_provenance(self, dataset_id: uuid.UUID, extra: dict[str, Any]) -> None:
         """Attach extra provenance to a specific dataset.
 
         Parameters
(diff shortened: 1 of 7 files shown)


python/lsst/daf/butler/_dataset_provenance.py
```diff
@@ -82,3 +89,352 @@ def add_extra_provenance(self, dataset_id: uuid.UUID, extra: dict[str, int | flo
...
+        Keys from the given dataset (all optional if no dataset is given):
+
+        :id: UUID of the given dataset.
+        :run: Run of the given dataset.
+        :datasettype: Dataset type of the given dataset.
+        :dataid x: An entry for each required dimension in the data ID.
+
+        Each input dataset will have the ``id``, ``run``, and ``datasettype``
+        keys as defined above (but no ``dataid`` key) with an ``input N``
+        prefix where ``N`` starts counting at 0.
```
erykoff, line 158: Can you give a fully populated example here?  I'm having trouble wrapping my head around this sentence.

tests/test_butler.py
```diff
@@ -2151,10 +2151,31 @@ def test_provenance(self):
...
+        self.assertEqual(ref_from_prov, metric_ref2)
+        # Direct __eq__ of the provenance does not work because one side
+        # includes dimension records.
+        self.assertEqual({ref.id for ref in prov_from_prov.inputs}, {ref.id for ref in prov.inputs})
+        self.assertEqual(prov_from_prov.quantum_id, prov.quantum_id)
+        self.assertEqual(prov_from_prov.extras, prov.extras)
+
+        # Check that simple types can be reconstructed with non-standard
+        # separators.
+        prov_dict = prov.to_flat_dict(metric_ref2, prefix="XYZ", sep="😎", simple_types=True)
```
erykoff, line 2180: Do we have any other emoji in the codebase?

timj: Yes. Butler uses some for collection names in tests. The developer guide does recommend using unicode in tests.

timj:
@andy-slac thanks for agreeing to review.

python/lsst/daf/butler/_dataset_provenance.py
```diff
@@ -82,3 +85,94 @@ def add_extra_provenance(self, dataset_id: uuid.UUID, extra: dict[str, int | flo
...
+            """Make the key in the correct form."""
+            start = prefix
+            if start:
+                start += sep
+            k = sep.join(str(kk) for kk in keys)
+            if use_upper:
+                k = k.upper()
+            return f"{start}{k}"
+
+        prov: dict[str, int | float | str | bool | uuid.UUID | DataIdValue] = {}
```
andy-slac: `self.extras` value type is `Any`, does that mean that `prov` can contain `Any` as well if `simple_types` is `False`? (And return type of this method too.)

python/lsst/daf/butler/_dataset_provenance.py
```diff
@@ -82,3 +85,94 @@ def add_extra_provenance(self, dataset_id: uuid.UUID, extra: dict[str, int | flo
...
+        prov : `dict`
+            Dictionary representing the provenance. The keys are defined
+            in the notes below.
+
+        Notes
+        -----
+        The provenance keys created by this method are defined below. The
+        case of the keys will match the case of the first character of the
+        prefix (defined by whether `str.isupper()` returns true, else they will
+        be lower case).
```
andy-slac: That seems to be a little ad hoc. What if people want upper case keys, but do not want to specify prefix?

python/lsst/daf/butler/_dataset_provenance.py
```diff
@@ -82,3 +89,372 @@ def add_extra_provenance(self, dataset_id: uuid.UUID, extra: dict[str, int | flo
...
+            prefix has no case).
+        """
+        if len(sep) != 1:
+            raise ValueError(f"Separator for provenance keys must be a single character. Got {sep!r}.")
+        use_upper = prefix[0].isupper() if prefix else False
+        if prefix:
+            prefix += sep
+        k = sep.join(str(kk) for kk in keys)
+        if use_upper:
+            k = k.upper()
```
andy-slac, line 254: This will also uppercase the separator character? I know that it's very likely to be some sort of punctuation, but people are allowed to use any string for separator 🙂

python/lsst/daf/butler/_dataset_provenance.py
```diff
@@ -82,3 +89,372 @@ def add_extra_provenance(self, dataset_id: uuid.UUID, extra: dict[str, int | flo
...
+            if match := re.match("(.*)input(.)0(.)(?:id|datasettype|run)$", k, flags=re.IGNORECASE):
+                _update_matches(match, prefixes, separators)
+            elif match := re.match(
+                # Data coordinates are a-z or underscore.
+                "(.*)dataid(.)[a-z_]+$",
+                k,
+                flags=re.IGNORECASE,
+            ):
+                _update_matches(match, prefixes, separators)
+            elif match := re.match(r"(.*)\b(id|datasettype|run)$", k, flags=re.IGNORECASE):
```
andy-slac, line 316: This assumes that separator should be a non-word character, so underscore cannot be used as separator. Maybe `to_flat_dict` should check that separator is not `\w`?

python/lsst/daf/butler/_dataset_provenance.py
```diff
@@ -82,3 +89,372 @@ def add_extra_provenance(self, dataset_id: uuid.UUID, extra: dict[str, int | flo
...
+        use_upper = prefix[0].isupper() if prefix else False
+        if prefix:
+            prefix += sep
+        k = sep.join(str(kk) for kk in keys)
+        if use_upper:
+            k = k.upper()
+        return f"{prefix}{k}"
+
+    @staticmethod
+    def _find_prefix_and_sep(prov_dict: Mapping) -> tuple[str, str] | tuple[None, None]:
```
andy-slac: `Mapping[str, Any]` maybe?

python/lsst/daf/butler/_dataset_provenance.py
```diff
@@ -82,3 +89,372 @@ def add_extra_provenance(self, dataset_id: uuid.UUID, extra: dict[str, int | flo
...
+        if len(separators) > 1:
+            raise ValueError(
+                f"Inconsistent values found for separators in provenance header. Got {separators}."
+            )
+        if len(prefixes) > 1:
+            raise ValueError(f"Inconsistent values for prefix found in provenance headers. Got {prefixes}.")
+        return prefixes.pop(), separators.pop()
+
+    @classmethod
+    def _find_provenance_keys_in_flat_dict(cls, prov_dict: Mapping) -> dict[str, str]:
```
andy-slac: Also `Mapping[str, Any]`?

python/lsst/daf/butler/_dataset_provenance.py
```diff
@@ -82,3 +89,372 @@ def add_extra_provenance(self, dataset_id: uuid.UUID, extra: dict[str, int | flo
...
+            prefix += sep
+
+        core_provenance = tuple(f"{prefix}{k}".lower() for k in ("run", "id", "datasettype", "quantum"))
+
+        prov_keys: dict[str, str] = {}
+        for k in list(prov_dict):
+            # the input provenance can include extra keys that we cannot
+            # know so have to match solely on INPUT N.
+            found_key = False
+            if re.match(rf"{prefix}input{sep}(\d+){sep}(.*)$", k, flags=re.IGNORECASE):
```
andy-slac: Hmm, what if separator is a backslash?

python/lsst/daf/butler/_dataset_provenance.py
```diff
@@ -82,3 +89,372 @@ def add_extra_provenance(self, dataset_id: uuid.UUID, extra: dict[str, int | flo
...
+            if input.run is not None:  # for mypy
+                prov[_make_key("input", i, "run")] = input.run
+            if input.datasetType is not None:  # for mypy
+                prov[_make_key("input", i, "datasettype")] = input.datasetType.name
+
+            if input.id in self.extras:
+                for xk, xv in self.extras[input.id].items():
+                    if simple_types and not isinstance(xv, str | float | int | bool):
+                        xv = str(xv)
+                    prov[_make_key("input", i, xk)] = xv
```
andy-slac, line 220: Could `self.extras` contain keys which conflict with other keys, e.g. `datasettype`?

andy-slac (approved):
Looks good, a small bunch of comments.

python/lsst/daf/butler/_dataset_provenance.py
timj: Seemed like a highly unlikely scenario but I can add a new parameter that can be used to override the default of using the prefix.

python/lsst/daf/butler/_dataset_provenance.py
timj: Yes. I should have changed that to Any when I changed the definition of extras up top (although, it does mean if someone tries to store something in extras that can't be serialized to JSON by Pydantic then things could break). Maybe I should walk back the `extras` scope and use the definition on this line for now (but simple_types is still useful for uuid).

python/lsst/daf/butler/_dataset_provenance.py
timj: Sigh. Yes. Hmm. This is why I ask you for reviews... I think the fix is to validate add_extra_provenance and disallow run, id, and datasettype as keys.

python/lsst/daf/butler/_dataset_provenance.py
timj: I will make sure that separator does not match `[_\w]`.

python/lsst/daf/butler/_dataset_provenance.py
timj: Will add a check.

python/lsst/daf/butler/_dataset_provenance.py
timj: Thank you.

tests/test_parquet.py
```diff
@@ -771,12 +773,69 @@ def testBadDataFrameColumnParquet(self):
...
+        put_ref3 = self.butler.put(tab2, self.datasetType, dataId={}, run="new_run")
+
+        # tab2 will have been updated in place.
+        expected = {
+            "lsst.butler.id": str(put_ref3.id),
+            "lsst.butler.run": "new_run",
+            "lsst.butler.datasettype": "data",
+        }
+        self.assertEqual(tab2.meta, expected)
+        null_prov, prov_ref = DatasetProvenance.from_flat_dict(tab2.meta, self.butler)
```
erykoff, line 835: So if anybody were to set a metadata key that was `lsst.butler.something` would this break?  Just curious; it seems fair to have these as reserved keys but I don't know how we could enforce that.

erykoff (approved):
Parquet code looks good.

python/lsst/daf/butler/_dataset_provenance.py
timj: Backslash now works (now that I escape sep and prefix before using in the regex.

tests/test_parquet.py
timj: It seems fair to me that only butler can set `lsst.butler.` keys (that's the point of the namespacing). If someone added a `lsst.butler.input 5 id` or something then this would break because the separators would be confused. I think so long as `something` was not `run`, `dataid`, `id`, `datasettype` or `input` then this code wouldn't notice.

python/lsst/daf/butler/_dataset_provenance.py
timj: Validation added.

PR #1159, opened 2025-02-24
DM-46023: Use Mapping instead of dict to fix downsteam type complaints.

TallJimbo:
With this change, downstream code can list just what types they might actually provide, rather than all of the ones the method could accept.

## Checklist

- [ ] ran Jenkins


timj (approved):
Sorry about that.

TallJimbo:
Merging this ahead of the `pipe_base` branch for this ticket to unblock https://github.com/lsst/pipe_base/pull/471.

PR #1160, opened 2025-02-26
DM-49197: Fix astropy pformat_all deprecation

dhirving:
Astropy's Table.pformat_all() is now deprecated -- Table.pformat() is an exact replacement as of Astropy v7.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
 ]
 keywords = ["lsst"]
 dependencies = [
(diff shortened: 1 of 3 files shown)


timj:
I think we may need to set our min astropy version to >=7 because in v6 the `pformat` had different defaults so we will get truncated tables in astropy 6 unless we for the max_width/max_liine parameters here.

timj (approved):

PR #1161, opened 2025-02-27
DM-49180: Check universe version before registering dataset type

andy-slac:
A mismatch in the universe versions causes some unexpected issues, sometimes silently. A simple version check before new dataset type registration should solve some of the cases.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
+++ b/python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
@@ -317,6 +317,16 @@ class ByDimensionsDatasetRecordStorageManagerUUID(DatasetRecordStorageManager):
             raise ValueError(
                 f"Component dataset types can not be stored in registry. Rejecting {dataset_type.name}"
             )
+
+        # If database universe and dimension group universe are different it
+        # can cause unexpected effects.
+        if dataset_type.dimensions.universe is not self._dimensions.universe:
+            raise ValueError(
+                "Incompatible dimension universe versions - "
+                f"database universe: {self._dimensions.universe}, "
+                f"dataset type universe: {dataset_type.dimensions.universe}."
+            )
+
         record = self._fetch_dataset_type_record(dataset_type.name)
         if record is None:
             if (dynamic_tables := self._cache.get_by_dimensions(dataset_type.dimensions)) is None:
--- a/python/lsst/daf/butler/registry/tests/_registry.py
+++ b/python/lsst/daf/butler/registry/tests/_registry.py
@@ -75,7 +75,7 @@ from ..._exceptions import (
 from ..._exceptions_legacy import DatasetTypeError
 from ..._storage_class import StorageClass
 from ..._timespan import Timespan
-from ...dimensions import DataCoordinate, DataCoordinateSet, SkyPixDimension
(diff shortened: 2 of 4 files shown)


timj (approved):
Aha. Was this the cause of the NERSC problem?

andy-slac:
> Aha. Was this the cause of the NERSC problem?

Indeed. Kind of non-trivial to figure out the cause.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -317,6 +317,21 @@ def register_dataset_type(self, dataset_type: DatasetType) -> bool:
...
             )
+
+        # If database universe and dimension group universe are different it
+        # can cause unexpected effects.
+        db_universe_version = (self._dimensions.universe.namespace, self._dimensions.universe.version)
+        dataset_type_universe_version = (
+            dataset_type.dimensions.universe.namespace,
+            dataset_type.dimensions.universe.version,
+        )
+        if dataset_type_universe_version != db_universe_version:
```
TallJimbo: It should also work to test `self._dimensions.universe is dataset_type.dimensions`, since `DimensionUniverse` construction works very hard to never return new instances for the same universe.  This has caused trouble in some other contexts, so maybe we don't want to add more code that depends on that behavior, but I think we've got a lot of code that does assume it.

TallJimbo (approved):
It might be a little bit safe to put similar guards in `DimensionRecordStorageManager.save_dimension_group`.  I'm not sure if that would avoid the need for the guard you've added here or not.

andy-slac:
> It might be a little bit safe to put similar guards in `DimensionRecordStorageManager.save_dimension_group`. I'm not sure if that would avoid the need for the guard you've added here or not.

I had this check in `save_dimension_group` originally, but there seem to be other failure modes in `register_dataset_type` that it does not handle. `save_dimension_group` is only called from `register_dataset_type`, I think a single check there should be enough.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
andy-slac: Hmm, identity check breaks some unit tests. I think this is because of [this code](https://github.com/lsst/daf_butler/blob/main/python/lsst/daf/butler/remote_butler/_remote_butler.py#L192-L196) in RemoteButler:

        universe = DimensionUniverse(
            config,
            # The process-global cache internal to DimensionUniverse can mask
            # problems in unit tests, since client and server live in the same
            # process for these tests.  We are doing our own caching, so we
            # don't benefit from it.  So just disable it.
            use_cache=False,
        )

andy-slac: I think I found a reasonable workaround that allows identity check to work. @TallJimbo, could you re-review it?

TallJimbo: Looks good.

And while I don't know much about the multiple-remote-butlers scenario, I suspect that anyone who trips this tighter check by passing an equivalent but different universe instance is probably doing something less than ideal anyway.

TallJimbo (approved):

PR #1162, opened 2025-03-04
DM-31824: Use ResourcePath.mtransfer for Butler.transfer_from

timj:
Requires lsst/resources#107

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -2748,6 +2748,9 @@ class FileDatastore(GenericBaseDatastore[StoredFileInfo]):
         # Record each time we have done a "direct" transfer.
         direct_transfers = []
 
+        # Keep track of all the file transfers that are required.
+        from_to: list[tuple[ResourcePath, ResourcePath]] = []
+
         # Now can transfer the artifacts
         log.verbose("Transferring artifacts")
         for ref in refs:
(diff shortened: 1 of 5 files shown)


python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -2158,10 +2158,19 @@ def retrieveArtifacts(
...
-        # In theory can now parallelize the transfer
+        # Parallelize the transfer. Re-raise as a single exception if
+        # a FileExistsError is encountered anywhere.
         log.debug("Number of artifacts to transfer to %s: %d", str(destination), len(to_transfer))
-        for source_uri, target_uri in to_transfer.items():
-            target_uri.transfer_from(source_uri, transfer=transfer, overwrite=overwrite)
+        try:
+            ResourcePath.mtransfer(
+                transfer,
+                ((from_uri, to_uri) for from_uri, to_uri in to_transfer.items()),
```
dhirving: Isn't this equivalent to `to_transfer.items()`?

python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -2796,21 +2808,28 @@ def transfer_from(
...
+        if dry_run:
+            log.info("Would be copying %d file artifacts", len(from_to))
+        else:
+            log.verbose("Copying %d file artifacts", len(from_to))
+            with time_this(log, msg="Transferring datasets into datastore", level=VERBOSE):
+                ResourcePath.mtransfer(
+                    transfer,
+                    from_to,
+                    overwrite=True,
+                    transaction=self._transaction,
```
dhirving, line 2826: I _think_ `DatastoreTransaction` is currently threadsafe since it does nothing except append values to a list, but you might want to add some comments in `DatastoreTransaction` to make future editors aware that it is has to stay thread-safe.

dhirving (approved):
Looks good to me

python/lsst/daf/butler/datastores/fileDatastore.py
timj: Yes 🙂 I think that before the from and to were backwards in the original.

PR #1163, opened 2025-03-05
DM-49316: Add more logging to FileDatastore.transfer_from()

dhirving:
Add enough logs in FileDatastore.transfer_from() to distinguish the timing of database reads, database writes, and the actual file transfer.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -2649,6 +2649,7 @@ class FileDatastore(GenericBaseDatastore[StoredFileInfo]):
         artifact_existence: dict[ResourcePath, bool] | None = None,
         dry_run: bool = False,
     ) -> tuple[set[DatasetRef], set[DatasetRef]]:
+        log.verbose("transfer_from %s to %s", source_datastore.name, self.name)
         # Docstring inherited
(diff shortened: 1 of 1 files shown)


python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -2828,6 +2831,7 @@ def transfer_from(
                 "" if n_skipped == 1 else "s",
             )
 
+        log.verbose("Finished transfer with %d accepted, %d rejected", len(accepted), len(rejected))
```
timj: Can you put this in the butler.transfer_from code since there is already logic there to print the number of rejected and so if you get rejected datasets it will tell you twice in verbose mode (since the other is a warning).

timj: Also, in chained datastore context there could be message from multiple file datastores so we should consider putting the datastore name in the log message (which does make an argument for putting a global verbose message in the butler.transfer_from and a datastore specific message here with a name.

dhirving: OK,  I added the datastore names in and some additional logs to bracket the transfer by datastore.  Also realized that I had missed an important log for one of the database regions.

timj (approved):

PR #1164, opened 2025-03-07
DM-49296: Set SQLite constant_rows_max, and use it in join_data_coordinates.

TallJimbo:
With QG generation now relying heavily on join_data_coordinates, we were hitting this limit sometimes.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/databases/sqlite.py
+++ b/python/lsst/daf/butler/registry/databases/sqlite.py
@@ -398,6 +398,12 @@ class SqliteDatabase(Database):
         ]
         return sqlalchemy.sql.union_all(*selects).alias(name)
 
+    def get_constant_rows_max(self) -> int:
+        # Docstring inherited.
+        # This is the default SQLITE_MAX_COMPOUND_SELECT (see
+        # https://www.sqlite.org/limits.html):
+        return 500
+
     @property
     def has_distinct_on(self) -> bool:
         # Docstring inherited.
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/registry/databases/sqlite.py
```diff
@@ -398,6 +398,12 @@ def constant_rows(
         ]
         return sqlalchemy.sql.union_all(*selects).alias(name)
 
+    def get_constant_rows_max(self) -> int:
+        # Docstring inherited.
+        # This is the default SQLITE_MAX_COMPOUND_SELECT (see
+        # https://www.sqlite.org/limits.html):
+        return 500
```
andy-slac, line 405: The default one (in `Database`) is 100, does it mean Postgres limit is lower than SQLite?

andy-slac (approved):
Looks OK, one question.

python/lsst/daf/butler/registry/databases/sqlite.py
TallJimbo: There's no limit for this specifically listed on the PostgreSQL limits doc page, so I suspect the de facto PostgreSQL limit one on total SELECT statement size or number of bind parameters.  Keeping the limit really low here means we make temporary tables more often, which might be less efficient than we could be, but it guards pretty effectively against blowing one of those limits with really wide rows.

Or, in other words, "I don't think we know what we should set it to, but performance is okay with this."

andy-slac: I'm just surprised that you are reducing `DirectQueryDriver._constant_rows_limit` to 500 for SQLite and to 100 for Postgres. Before this change it was 1000 for both. My guess is that you hit 500 limit in SQLite tests, but does it also need much lower Postgres limit?

TallJimbo: Oh, that's a good point, when you put it that way, and all of my testing with PostgreSQL was with the 1000-row limit before I made this change.  I think I will go change the default to that, as I think that's still plenty conservative enough.

PR #1169, opened 2025-03-18
DM-49513: Reduce number of queries required when importing datasets

dhirving:
Reduce the number of queries required to import datasets to the registry in `Butler.transfer_from()`.  Instead of doing separate queries for each dataset type, we now group dataset types that have the same dimensions.

This reduces the number of queries in Prompt Processing transfers back to the central repo from around 750 to 50, and reduces the number of temp tables used from 70 to 5.

## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -2131,11 +2131,6 @@ def transfer_from(
                 else:
                     imported_refs = refs_to_import
 
-                # Check all the UUIDs were imported, ignoring dataset type
-                # storage class differences.
-                imported_uuids = set(ref.id for ref in imported_refs)
-                uuids_to_import = set(ref.id for ref in refs_to_import)
-                assert set(imported_uuids) == set(uuids_to_import)
```
dhirving, line 2138: This assertion wasn't meaningful even before this change -- a complete set of refs with the same IDs is/was roundtripped in all cases.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -820,23 +826,16 @@ def _validate_import(
...
-                        raise ConflictingDefinitionError(
-                            f"Current run {existing_run!r} and new run {new_run!r} do not agree for "
-                            f"dataset {row.dataset_id}."
-                        )
-                    else:
-                        raise ConflictingDefinitionError(
-                            f"Dataset {row.dataset_id} was provided with type {storage.dataset_type.name!r} "
-                            f"in run {new_run!r}, but was already defined with type ID {row.dataset_type_id} "
-                            f"in run {run!r}."
-                        )
```
dhirving, line 834: This branch was unreachable and uncovered by tests in the original version.

doc/changes/DM-49513.perf.md
```diff
@@ -0,0 +1,3 @@
+`Butler.transfer_from()` now uses fewer database queries when inserting
+datasets of multiple dataset types, when some of the dataset types have the
+same dimensions.
```
andy-slac: I think it should be one sentence per line.

python/lsst/daf/butler/registry/sql_registry.py
```diff
@@ -1169,34 +1159,24 @@ def _importDatasets(
...
         progress = Progress("daf.butler.Registry.importDatasets", level=logging.DEBUG)
         if expand:
-            data_ids = {
-                dataset.id: self.expandDataId(dataset.dataId, dimensions=datasetType.dimensions)
-                for dataset in progress.wrap(datasets, f"Expanding {datasetType.name} data IDs")
-            }
-        else:
-            data_ids = {dataset.id: dataset.dataId for dataset in datasets}
+            datasets = [
+                dataset.expanded(self.expandDataId(dataset.dataId, dimensions=dataset.datasetType.dimensions))
```
andy-slac, line 1163: This could also generate query per ref (or even multiple queries)?

python/lsst/daf/butler/registry/interfaces/_datasets.py
```diff
@@ -424,29 +424,16 @@ def insert(
...
             The record object describing the `~CollectionType.RUN` collection
             these datasets will be associated with.
-        data_ids : `~collections.abc.Mapping`
-            Mapping from dataset ID to data ID.
-
-        Returns
-        -------
-        datasets : `list` [ `DatasetRef` ]
-            References to the inserted or existing datasets.
+        refs : `list` [ `DatasetRef ` ]
```
andy-slac:
```suggestion
        refs : `list` [ `DatasetRef` ]
```

python/lsst/daf/butler/registry/interfaces/_datasets.py
```diff
@@ -424,29 +424,16 @@ def insert(
...
+            List of datasets to be be inserted.
```
andy-slac: Maybe mention that datasets' `run` must match ``run`` parameter?

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -698,45 +698,56 @@ def insert(
...
-            for dataset_id, data_id in data_ids.items()
+            {
+                "dataset_type_id": dataset_type_storage[ref.datasetType.name].dataset_type_id,
+                collection_fkey_name: run.key,
+                "dataset_id": ref.id,
+                **ref.dataId.required,
+            }
+            for ref in refs
         ]
         with self._db.transaction(for_temp_tables=True), self._db.temporary_table(table_spec) as tmp_tags:
```
andy-slac, line 745: Would be nice if we could get rid of temp table here in case there are not very many records to insert.

andy-slac (approved):
Looks great!

python/lsst/daf/butler/registry/sql_registry.py
dhirving: Yeah, I noticed that when I was in there as well.

In the prompt processing case it's not an issue -- their `DatasetRef`s are already expanded.  But I think it's worth following up on sometime soon, there are several places in the code base that do `expandDataId` in a loop over a potentially large number of refs.

timj: CM team can give you some batch graphs that contain 200k datasets to be imported by `transfer-from-graph` if you want to see the effect this has on that use case.

timj:
Do you want to fix `butler.ingest()` whilst you are in there? That also has a loop over dataset type doing importDatasets.

dhirving:
> Do you want to fix `butler.ingest()` whilst you are in there? That also has a loop over dataset type doing importDatasets.

I was going to open a separate ticket for that -- I want to get this one down so PP can benefit from it because they are running weekly large scale tests now.  

There is a lot of duplicated logic between `_ingest_file_datasets` and `transfer_from` that should be pushed down into `SqlRegistry.import_datasets` because this whole dimension group thing is a weird implementation detail that doesn't need to be duplicated multiple places.

I really don't want to get involved in totally optimizing every facet of this situation right now -- my goal was just to get PP to a temporarily workable state so I have a minute to help them with their architecture change, and this whole PP thing is an interrupt that I am spending more time on than I really have right now.

PR #1173, opened 2025-03-21
DM-49622: add Pydantic serialization for DimensionRecordSet

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/column_spec.py
+++ b/python/lsst/daf/butler/column_spec.py
@@ -125,18 +125,6 @@ class ColumnValueSerializer(ABC):
         raise NotImplementedError
 
 
-class _DefaultColumnValueSerializer(ColumnValueSerializer):
-    """Default implementation of serializer for basic types."""
-
-    def serialize(self, value: Any) -> Any:
-        # Docstring inherited.
-        return value
-
-    def deserialize(self, value: Any) -> Any:
-        # Docstring inherited.
-        return value
-
-
 class _TypeAdapterColumnValueSerializer(ColumnValueSerializer):
     """Implementation of serializer that uses pydantic type adapter."""
 
@@ -258,7 +246,7 @@ class IntColumnSpec(_BaseColumnSpec):
 
     def serializer(self) -> ColumnValueSerializer:
         # Docstring inherited.
-        return _DefaultColumnValueSerializer()
+        return _TypeAdapterColumnValueSerializer(pydantic.TypeAdapter(int))
 
 
 @final
@@ -282,7 +270,7 @@ class StringColumnSpec(_BaseColumnSpec):
 
     def serializer(self) -> ColumnValueSerializer:
         # Docstring inherited.
-        return _DefaultColumnValueSerializer()
+        return _TypeAdapterColumnValueSerializer(pydantic.TypeAdapter(str))
 
 
 @final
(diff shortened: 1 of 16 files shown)


python/lsst/daf/butler/dimensions/_records.py
```diff
@@ -490,6 +538,86 @@ def toDict(self, splitTimespan: bool = False) -> dict[str, Any]:
                 results["datetime_end"] = timespan.end
         return results
 
+    def serialize_key_value(self) -> SerializedKeyValueDimensionRecord:
+        """Serialize this record to a `list` that can be sliced into a key
+        (data ID values) / value (everything else) pair.
```
TallJimbo, line 538: If anybody has thoughts on better nomenclature for this new `DimensionRecord` serialization mode (it needs to be distinguished from the old one somehow), I'd love to hear it, because I'm not thrilled with what I've come up with so far.

timj: It's still a single DimensionRecord? not effectively `list[DimensionRecord]`?

TallJimbo: Yes, here.  Serializing a collection of dimension records is over on `DimensionRecordSet.serialize_records` and it delegates to this.

python/lsst/daf/butler/column_spec.py
```diff
@@ -243,6 +250,157 @@ def display(self, level: int = 0, tab: str = "  ") -> list[str]:
...
+        serializer or JSON schema.
+    """
+    return pydantic_core.core_schema.typed_dict_schema(
+        {
+            spec.name: pydantic_core.core_schema.TypedDictField(
+                schema=spec.pydantic_core_schema, type="typed-dict-field"
+            )
+            for spec in columns
+        }
+    )
```
dhirving: There's a functional form of TypedDict that could accomplish this without resorting to pydantic-core, e.g. [from the docs](https://docs.python.org/3/library/typing.html#typing.TypedDict)
```
Point2D = TypedDict('Point2D', {'x': int, 'y': int, 'label': str})
```

python/lsst/daf/butler/column_spec.py
```diff
@@ -243,6 +250,157 @@ def display(self, level: int = 0, tab: str = "  ") -> list[str]:
...
+    columns : `~collections.abc.Iterable` [ `ColumnSpec` ]
+        Iterable of column specifications.
+
+    Returns
+    -------
+    schema : `pydantic_core.core_schema.CoreSchema`
+        A low-level Pydantic schema that can be used to construct a validator,
+        serializer or JSON schema.
+    """
+    return pydantic_core.core_schema.tuple_schema([spec.pydantic_core_schema for spec in columns])
```
dhirving: You can create tuple types dynamically without pydantic-core as well:
```
types = [int, str, str]
tuple[*types]

dhirving (approved):
Looks good to me.  Not sure it's necessary to resort to pydantic-core here but it's also used in a relatively straightforward way so it's not necessarily a problem either.  

The changes to general query result serialization are fine with me -- nothing is really using that except for Stelio's SIA service, which is not widely used yet, so if it breaks something we'll fix it.

python/lsst/daf/butler/column_spec.py
TallJimbo: Done (for `tuple` as well).  I considered this before but didn't really give it a try because I'm uncomfortable using the static-typing constructs as runtime objects.  But I think I agree that in this case it's a bit better when you factor in the vague stability guarantees on `pydantic_core`.

PR #1180, opened 2025-04-01
DM-49537: Implement ScarletModelFormatter for scarlet models

fred3m:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -409,6 +409,9 @@ storageClasses:
     pytype: spectractor.fit.fitter.FitParameters
   ScarletModelData:
     pytype: lsst.scarlet.lite.io.ScarletModelData
+    parameters:
+      - blend_id
+    delegate: lsst.meas.extensions.scarlet.io.ScarletModelDelegate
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -409,6 +409,9 @@ storageClasses:
     pytype: spectractor.fit.fitter.FitParameters
   ScarletModelData:
     pytype: lsst.scarlet.lite.io.ScarletModelData
+    parameters:
+      - blend_id
```
timj, line 413: One issue is going to be that if someone asks for a blend ID from an old JsonFormatter dataset it's going to break. I'm not sure if there is anyway to fix that and I don't know how important it is for it to work.

timj (approved):

python/lsst/daf/butler/configs/storageClasses.yaml
fred3m: Yes, I think that it's a good idea to include a comment in the (not yet written) deblending notebooks for DP1 that this feature will be available in DR1, and give an example of the API, but note that it will not work with DP1.

PR #1184, opened 2025-04-17
DM-50044: Unify ingest and transfer_from ref import logic

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_butler/_direct_butler.py
+++ b/python/lsst/daf/butler/direct_butler/_direct_butler.py
@@ -1607,10 +1607,6 @@ class DirectButler(Butler):  # numpydoc ignore=PR02
 
         # And the nested loop that populates it:
         for dataset in progress.wrap(datasets, desc="Grouping by dataset type"):
-            # Somewhere to store pre-existing refs if we have an
-            # execution butler.
-            existingRefs: list[DatasetRef] = []
-
             for ref in dataset.refs:
                 group_key = (ref.datasetType, ref.run)
 
@@ -1623,21 +1619,7 @@ class DirectButler(Butler):  # numpydoc ignore=PR02
                     )
 
                 groupedDataIds[group_key][ref.dataId] = dataset
-
-            if existingRefs:
-                if len(dataset.refs) != len(existingRefs):
-                    # Keeping track of partially pre-existing datasets is hard
-                    # and should generally never happen. For now don't allow
-                    # it.
-                    raise ConflictingDefinitionError(
-                        f"For dataset {dataset.path} some dataIds already exist"
-                        " in registry but others do not. This is not supported."
-                    )
-
-                # Store expanded form in the original FileDataset.
-                dataset.refs = existingRefs
-            else:
-                groupedData[group_key].append(dataset)
+            groupedData[group_key].append(dataset)
 
         # Now we can bulk-insert into Registry for each DatasetType.
         for (datasetType, this_run), grouped_datasets in progress.iter_item_chunks(
--- a/python/lsst/daf/butler/direct_butler/_direct_butler.py
+++ b/python/lsst/daf/butler/direct_butler/_direct_butler.py
@@ -1587,17 +1587,13 @@ class DirectButler(Butler):  # numpydoc ignore=PR02
     def _ingest_file_datasets(
         self,
         datasets: Sequence[FileDataset],
+        *,
+        dry_run: bool = False,
     ) -> None:
-        # Docstring inherited.
         if not datasets:
             return
 
         progress = Progress("lsst.daf.butler.Butler.ingest", level=logging.DEBUG)
-
-        # We need to reorganize all the inputs so that they are grouped
-        # by dataset type and run. Multiple refs in a single FileDataset
-        # are required to share the run and dataset type.
-        groupedData: MutableMapping[tuple[DatasetType, str], list[FileDataset]] = defaultdict(list)
 
         # Track DataIDs that are being ingested so we can spot issues early
         # with duplication. Retain previous FileDataset so we can report it.
@@ -1605,8 +1601,10 @@ class DirectButler(Butler):  # numpydoc ignore=PR02
             defaultdict(dict)
         )
 
-        # And the nested loop that populates it:
-        for dataset in progress.wrap(datasets, desc="Grouping by dataset type"):
+        # All the refs we need to import.
+        refs: list[DatasetRef] = []
+
+        for dataset in progress.wrap(datasets, desc="Validating dataIDs"):
             for ref in dataset.refs:
                 group_key = (ref.datasetType, ref.run)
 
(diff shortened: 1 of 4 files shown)


timj:
The diff is hard to read but what I did was split the logic in transfer_from into 3 private methods -- one for working out and registering dataset types and collating dimension record information, one for writing those dimension records, and one for doing the grouped importDatasets call.  This was necessary because the ingest and transfer_from datastore logic is slightly different so a single "do everything" method didn't quite work. The ingest logic needs to reattach expanded dataIds into the FileDataset before passing to datastore but transfer_from is working on refs.

python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -1587,30 +1587,24 @@ def ingest_zip(self, zip_file: ResourcePathExpression, transfer: str = "auto") -
     def _ingest_file_datasets(
         self,
         datasets: Sequence[FileDataset],
+        dry_run: bool = False,
     ) -> None:
         # Docstring inherited.
```
andy-slac: The comment seems to be wrong, the method is not declared in any of the base classes (I know it does not come from this PR, just caught my attention).

python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -1623,54 +1617,29 @@ def _ingest_file_datasets(
...
+            register_dataset_types=True,
+            dry_run=dry_run,
+            transfer_dimensions=False,
+        )
 
-            # Import the refs and expand the DataCoordinates since we can't
-            # guarantee that they are expanded and Datastore will need
-            # the records.
-            imported_refs = self._registry._importDatasets(refs_to_import, expand=True)
+        self._import_dimension_records(dimension_records, dry_run)
```
andy-slac: Maybe for uniformity with other methods `dry_run=dry_run`?

python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -1623,54 +1617,29 @@ def _ingest_file_datasets(
...
+        imported_refs = self._import_grouped_refs(
+            grouped_refs, self, progress, dry_run=dry_run, expand_refs=True
```
andy-slac: `progress` here feels a bit invasive, in a sense that it is not essential for what the method's primary purpose is. Could it be made optional and keyword-only?

python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -1933,28 +1902,30 @@ def _extract_dimension_records_from_data_ids(
...
         skip_missing: bool = True,
         register_dataset_types: bool = False,
         transfer_dimensions: bool = False,
         dry_run: bool = False,
-    ) -> collections.abc.Collection[DatasetRef]:
+    ) -> tuple[
+        defaultdict[_RefGroup, list[DatasetRef]],
+        dict[DimensionElement, dict[DataCoordinate, DimensionRecord]],
+        dict[ResourcePath, bool],
+    ]:
```
andy-slac: This feels somewhat uncomprehensible, could a separate `dataclass` describe it better?

python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -2076,86 +2047,126 @@ def transfer_from(
...
 
+    def _import_dimension_records(
+        self, dimension_records: dict[DimensionElement, dict[DataCoordinate, DimensionRecord]], dry_run: bool
+    ) -> None:
+        """Import dimension records collected during import pre-process."""
+        if dimension_records and not dry_run:
+            _LOG.verbose("Ensuring that dimension records exist for transferred datasets.")
+            # Order matters.
+            for element in self.dimensions.sorted(dimension_records.keys()):
+                records = [r for r in dimension_records[element].values()]
```
andy-slac: Could be just `list(dimension_records[element].values())`?

python/lsst/daf/butler/tests/hybrid_butler.py
```diff
@@ -72,6 +72,9 @@ def __new__(cls, remote_butler: RemoteButler, direct_butler: DirectButler) -> Hy
         self = cast(HybridButler, super().__new__(cls))
         self._remote_butler = remote_butler
         self._direct_butler = direct_butler
+        # Force the dimension universes to match remote and direct.
+        with self._remote_butler._cache.access() as cache:
+            cache.dimensions = self._direct_butler.dimensions
```
andy-slac: I guess this should work OK for tests, but does this expect that local universe and remote universe are identical? What happens if we add a test when they are different? Would it be better to test here that they are "equal" (namespace and version number)?

andy-slac (approved):
Looks good, a few minor questions.

python/lsst/daf/butler/tests/hybrid_butler.py
timj: hybrid butler is specifically designed for testing but remote butler is specifically designed not to use the main universe cache so this didn't seem like a problem. I could add a `DimesnsionUniverse.__eq__` method instead -- maybe we didn't have one originally because we didn't have formal versioning but now we have decided that if someone edits the universe yaml without changing the version number that we are going to have bigger problems anyhow so just assume it's enough.

python/lsst/daf/butler/direct_butler/_direct_butler.py
timj: Yes. Thanks. This was old code that was moved.

python/lsst/daf/butler/direct_butler/_direct_butler.py
timj: Ok. Will take another look. I was mostly trying to split the code up without changing it much. Maybe I could pass in the progress bar name instead.

python/lsst/daf/butler/direct_butler/_direct_butler.py
timj: Yes. Not sure why it says that.

python/lsst/daf/butler/direct_butler/_direct_butler.py
timj: I was wondering whether you would comment on that. I thought I would get away with it given that it's an internal API... 😉

python/lsst/daf/butler/direct_butler/_direct_butler.py
timj: Looking at the code the progress is used in the file ingest and passed into the import prepare so passing in the progress bar name doesn't seem right. This is an internal API so I'm not sure what I gain by making it optional if I'm going to define it in the two places that call the method.

python/lsst/daf/butler/tests/hybrid_butler.py
timj: Removed remote butler's standalone universe cache instead.

PR #1185, opened 2025-04-18
DM-50310: add dimension record serialization helpers

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_dataset_ref.py
+++ b/python/lsst/daf/butler/_dataset_ref.py
@@ -985,8 +985,7 @@ class SerializedDatasetRefContainerV1(SerializedDatasetRefContainer):
         # Extract dimension record metadata if present.
         dimension_records = None
         if data_ids and len(compact_refs) == len(data_ids):
-            dimension_group = dimensions.pop()
-            dimension_group = dimension_group.union(*dimensions)
+            dimension_group = DimensionGroup.union(*dimensions, universe=universe)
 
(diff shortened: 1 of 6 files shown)


TallJimbo:
@timj, I haven't implemented the key methods yet, but I'd like to get early feedback on whether the interfaces here look like they'd do what you need for the zip files.

python/lsst/daf/butler/_dataset_ref.py
```diff
@@ -1019,14 +1046,27 @@ def to_refs(self, universe: DimensionUniverse) -> list[DatasetRef]:
...
+                deserializers=self.dimension_records.make_deserializers(universe)
+            )
+
         refs: list[DatasetRef] = []
         for id_, minimal in self.compact_refs.items():
+            dataset_type = dataset_types[minimal.dataset_type_name]
             simple_data_id = SerializedDataCoordinate(dataId=minimal.data_id)
             data_id = DataCoordinate.from_simple(simple=simple_data_id, universe=universe)
+            if attacher:
+                data_ids = attacher.attach(dataset_type.dimensions, [data_id])
```
TallJimbo, line 1064: At first I was a little worried this would be less efficient than calling it with a lot of data IDs at once, but I think that's only true if you pass a `Query` in.  So for now just something to keep in mind; I don't think it's worth it to try to work this code to be able to pass in multiple data IDs at once.

PR #1187, opened 2025-04-25
DM-50491: Instrument butler.get and put calls with new ButlerMetrics class

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler_metrics.py
+++ b/python/lsst/daf/butler/_butler_metrics.py
@@ -1,0 +1,117 @@
+# This file is part of daf_butler.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (http://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This software is dual licensed under the GNU General Public License and also
+# under a 3-clause BSD license. Recipients may choose which of these licenses
+# to use; please see the files gpl-3.0.txt and/or bsd_license.txt,
+# respectively.  If you choose the GPL option then the following text applies
+# (but note that there is still no warranty even if you opt for BSD instead):
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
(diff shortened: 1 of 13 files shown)


python/lsst/daf/butler/_butler_metrics.py
```diff
@@ -0,0 +1,117 @@
...
+
+from lsst.utils.logging import LsstLoggers
+from lsst.utils.timer import time_this
+
+
+class ButlerMetrics(BaseModel):
+    """Metrics collected during Butler operations."""
+
+    time_in_put: float = 0.0
+    """Time, in seconds, spent in put()."""
```
TallJimbo: Please clarify wall-clock vs. CPU time.

python/lsst/daf/butler/_butler_metrics.py
```diff
@@ -0,0 +1,117 @@
...
+from collections.abc import Callable, Iterator
+from contextlib import contextmanager
+
+from pydantic import BaseModel
+
+from lsst.utils.logging import LsstLoggers
+from lsst.utils.timer import time_this
+
+
+class ButlerMetrics(BaseModel):
```
TallJimbo, line 39: We might want to define a separate ABC or `typing.Protocol` for the type annotations on `Butler`.

TallJimbo (approved):

python/lsst/daf/butler/_butler_metrics.py
timj: You mean something that combines multiple butler parameters so we don't need to add new ones in many places?

TallJimbo: No, I mean a `ButlerMetricsInterface` with no state and with methods that `raise NotImplementedError` or do nothing, and then the current `ButlerMetrics` would be a concrete implementation of that interface.

But I'm not really sure we need it; it's hard to imagine a scenario where the default one is too onerous to inherit from, unless we add a whole lot more to the default one.

timj: I think it's just as easy to have a `ButlerMetricsNoOp` subclass that disables the instrumentation context managers and leaves all the values at 0.0 (but presumably we could also make it increment the counters but not increment the durations).

PR #1189, opened 2025-04-28
DM-50191: Add GLOB() function to user expression language

andy-slac:
Only the new query system supports `GLOB()`, old query system will raise `NotImplementedError` for it.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/doc/lsst.daf.butler/queries.rst
+++ b/doc/lsst.daf.butler/queries.rst
@@ -229,7 +229,7 @@ Comparison operators
 
 Language supports set of regular comparison operators: ``=``, ``!=``, ``<``,
 ``<=``, ``>``, ``>=``. This can be used on operands that evaluate to a numeric
-values or timestamps.
+values, timestamps, or strings.
 
 .. note :: The equality comparison operator is a single ``=`` like in SQL, not
     double ``==`` like in Python or C++.
@@ -272,7 +272,7 @@ as are these:
 
 Another usage of ``IN`` operator is for checking whether a timestamp or a time
 range is contained wholly in other time range. Time range in this case can be
-specified as a tuple of two time literals or identifers each representing a
+specified as a tuple of two time literals or identifiers each representing a
 timestamp, or as a single identifier representing a time range. In case a
 single identifier appears on the right side of ``IN`` it has to be enclosed
 in parentheses.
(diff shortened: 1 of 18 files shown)


doc/lsst.daf.butler/queries.rst
```diff
@@ -352,11 +352,17 @@ Function call
 Function call syntax is similar to other languages, expression for call
 consists of an identifier followed by zero or more comma-separated arguments
 enclosed in parentheses (e.g. ``func(1, 2, 3)``). An argument to a function
-can be any expression.
+can be any expression. Names of the functions are not case-sensitive.
 
-Presently there only one construct that uses this syntax, ``POINT(ra, dec)``
-is function which declares (or returns) sky coordinates similarly to ADQL
-syntax. Name of the ``POINT`` function is not case-sensitive.
+Presently there only few functions implemented in query language:
```
timj:
```suggestion
Presently the following functions are implemented in the query language:
```
Does make me wonder whether we should add the IVOA `POS` syntax since sphgeom already supports that.

python/lsst/daf/butler/queries/_expression_strings.py
```diff
@@ -290,6 +290,19 @@ def visitUnaryOp(self, operator: str, operand: _VisitorResult, node: Node) -> _V
             f"Unary operator {operator!r} is not valid for operand of type {operand.column_type} in {node!s}."
         )
 
+    def visitGlobNode(
+        self, expression: _VisitorResult, pattern: _VisitorResult, node: Node
+    ) -> _VisitorResult:
+        # Docstring inherited.
+        if isinstance(expression, _ColExpr) and expression.value.is_column_reference:
+            if expression.value.column_type != "string":
+                raise InvalidQueryError(f"glob() first argument must be a string column (in node {node})")
```
timj, line 299: Would it be quick to add tests for these two failure conditions? Or is this a case where the parser has already errored out before it gets here?

python/lsst/daf/butler/queries/_expression_strings.py
```diff
@@ -317,6 +330,8 @@ def _convert_comparison_operator(value: str) -> ComparisonOperator:
             return "=="
         case "OVERLAPS":
             return "overlaps"
+        case "GLOB":
+            return "glob"
```
timj: Did you expect this not to have test coverage?

python/lsst/daf/butler/registry/queries/expressions/parser/exprTree.py
```diff
@@ -73,6 +73,14 @@
 
     from .treeVisitor import TreeVisitor
 
+
+def _strip_parens(expression: Node) -> Node:
```
timj, line 77: I'm missing something here because it looks like this function only calls itself and does nothing to actually strip any parens...

timj (approved):
Looks okay. I might have missed it but I can't see a test for a glob string of a dimension itself (such as "instrument" or "group"), only dimension metadata.

python/lsst/daf/butler/queries/_expression_strings.py
andy-slac: There is no test for this exception, but there is a test for the second one. Second one is triggered by the parser because it must be a string literal. I'll try to add a test for this one.

python/lsst/daf/butler/queries/_expression_strings.py
andy-slac: Hmm, it turns out I'm not using this method for `glob`, but I wanted to do it initially. Will drop this `case`.

python/lsst/daf/butler/registry/queries/expressions/parser/exprTree.py
andy-slac: It should strip parens. If the node is not `Parens` it returns the node itself, but if the node is `Parens` then it returns the child of the node (recursively) effectively removing parens. This works on a parsed expression tree, this is why you do not see any `()` here.

andy-slac:
Added a few more unit tests, including a test for glob on `instrument`.

PR #1191, opened 2025-04-30
DM-50451: Add support for UUID literals to user expressions

andy-slac:
Queries on `dataset_id` were already working, but the only way to use
them was to provide UUID via bind parameters. This patch adds new
UUID literal type to parser which is specified in expression string
via function call `UUID('hex-string')`.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/doc/lsst.daf.butler/queries.rst
+++ b/doc/lsst.daf.butler/queries.rst
@@ -136,6 +136,16 @@ Time literals
     time scale. For detailed description of supported time specification
     check section :ref:`time-literals-syntax`.
 
+UUID literals
+    UUID values are produced by ``UUID()`` function which accepts a string
+    representation of the UUID. The string can have any format acceptable
+    by Python ``uuid.UUID()`` method.
+
+    Examples:
+
+    - ``UUID('12345678-1234-5678-1234-567812345678')``
+    - ``UUID('12345678123456781234567812345678')``
+
 Range literals
     This sort of literal is allowed inside ``IN`` expressions only. It consists
     of two integer literals separated by double dots and optionally followed by
@@ -441,6 +451,8 @@ Few examples of valid expressions using some of the constructs:
     visit.timespan.end < T'mjd/58938.515/tai'
 
     ingest_date < T'2020-11-06 21:10:00'
+
+    flat.dataset_id = UUID('e15ab039-bc8b-4135-87c5-90902a7c0b22')
 
 
 .. _daf_butler_query_ordering:
--- a/python/lsst/daf/butler/registry/queries/expressions/parser/exprTree.py
+++ b/python/lsst/daf/butler/registry/queries/expressions/parser/exprTree.py
@@ -136,7 +136,7 @@ class BinaryOp(Node):
     """
 
     def __init__(self, lhs: Node, op: str, rhs: Node):
-        Node.__init__(self, (lhs, rhs))
+        super().__init__((lhs, rhs))
         self.lhs = lhs
         self.op = op
         self.rhs = rhs
@@ -166,7 +166,7 @@ class UnaryOp(Node):
     """
 
     def __init__(self, op: str, operand: Node):
-        Node.__init__(self, (operand,))
+        super().__init__((operand,))
         self.op = op
         self.operand = operand
 
@@ -189,7 +189,7 @@ class StringLiteral(LiteralNode):
     """
 
     def __init__(self, value: str):
-        Node.__init__(self)
+        super().__init__()
         self.value = value
 
     def visit(self, visitor: TreeVisitor) -> Any:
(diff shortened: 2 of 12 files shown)


python/lsst/daf/butler/registry/queries/expressions/parser/exprTree.py
```diff
@@ -240,11 +245,32 @@ def __str__(self) -> str:
...
+    """Node representing a UUID literal.
+
+    Parameters
+    ----------
+    value : `UUID`
+        Literal value.
+    """
+
+    def __init__(self, value: UUID):
+        Node.__init__(self)
```
TallJimbo:
```suggestion
        super().__init__()
```
?  Or maybe `LiteralNode.__init__(self)`?  Or is using an indirect base-class `__init__` intentional?

python/lsst/daf/butler/registry/queries/expressions/parser/exprTree.py
```diff
@@ -240,11 +245,32 @@ def __str__(self) -> str:
...
+        self.value = value
+
+    def visit(self, visitor: TreeVisitor) -> Any:
+        # Docstring inherited from Node.visit
+        return visitor.visitUuidLiteral(self.value, self)
+
+    def __str__(self) -> str:
+        return "{value}".format(**vars(self))
```
TallJimbo:
```suggestion
        return str(self.value)
```
seem like a much simpler way to write this.

python/lsst/daf/butler/registry/queries/expressions/parser/exprTree.py
```diff
@@ -335,6 +361,12 @@ class IsIn(Node):
     """
 
     def __init__(self, lhs: Node, values: list[Node], not_in: bool = False):
+        # All values must be literals or binds (and we allow simple identifiers
```
TallJimbo, line 365: Does this need to delegate to `super()`?  I see that another derived class did.

python/lsst/daf/butler/registry/queries/expressions/parser/parserYacc.py
```diff
@@ -365,22 +365,39 @@ def p_predicate(cls, p: YaccProduction) -> None:
...
-        | identifier
+        | simple_id
         | bind_name
+        | function_call
         """
+        # This expression is only used in IN() operator and it is supposed to
+        # include only literals and bind names (and identifiers as we still
+        # allow simple identifiers as bind names). UUID literal is implemented
+        # via UUID() function call, so we need to allow function calls here
+        # too. IsIn will check that all operands are literlas or binds.
```
TallJimbo:
```suggestion
        # too. IsIn will check that all operands are literals or binds.
```

TallJimbo (approved):

doc/lsst.daf.butler/queries.rst
```diff
@@ -442,6 +452,8 @@ Few examples of valid expressions using some of the constructs:
 
     ingest_date < T'2020-11-06 21:10:00'
 
+    flat.dataset_id = UUID('e15ab039-bc8b-4135-87c5-90902a7c0b22')
```
timj, line 455: You need to specify the dataset type? We don't have any documentation on the `bind` form but if I have a list of arbitrary UUIDs and I want a query that retrieves them, do I do `where="dataset_id IN (ids)", bind={"ids": list_of_UUID}`? Where that is `list[uuid.UUID]`?

TallJimbo: If you have a query that only has one dataset type in it, you don't have to specify the dataset type.  The problem of querying for all datasets given UUIDs is tricky because they might have arbitrary dimensions, and our half-completed effort to query for multiple dataset types in batches with the same dimensions helps with that but doesn't completely solve it.  We'd probably have do that kind of query in two phases: first look up all of the dataset types and run collections from the UUIDs, and then do one query for each dimension group.

timj: Ah. So none of this solves my SIAv2 problem where I can be given arbitrary IVOIDs and know nothing about where they come from. Sounds like I will have to do the loop over `Butler.get_dataset`. I was hoping I could do a WHERE of `dataset_id in (ids)` potentially with some more filtering clauses.

TallJimbo: Unfortunately correct.  We probably want to look into a dedicated vectorized version of that that could do its own query against the `dataset` table and then delegate to the rest of the query system to get `DatasetRefs`.

python/lsst/daf/butler/registry/queries/expressions/parser/exprTree.py
andy-slac: That was copy-paste, need to use `super()` indeed.

python/lsst/daf/butler/registry/queries/expressions/parser/exprTree.py
andy-slac: It also calls `Node.__init__`, I need to change all subclasses to use `super()`.

andy-slac: You probably looked at another class - `RangeLiteral` was indeed missing call to base class `__init__`, I fixed that.

PR #1193, opened 2025-05-02
DM-50313: Allow zip ingest to register dimension records

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler.py
+++ b/python/lsst/daf/butler/_butler.py
@@ -1374,7 +1374,14 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
         raise NotImplementedError()
 
     @abstractmethod
-    def ingest_zip(self, zip_file: ResourcePathExpression, transfer: str = "auto") -> None:
+    def ingest_zip(
+        self,
+        zip_file: ResourcePathExpression,
+        transfer: str = "auto",
+        *,
+        transfer_dimensions: bool = False,
+        dry_run: bool = False,
+    ) -> None:
         """Ingest a Zip file into this butler.
 
         The Zip file must have been created by `retrieve_artifacts_zip`.
@@ -1385,10 +1392,17 @@ class Butler(LimitedButler):  # numpydoc ignore=PR02
             Path to the Zip file.
         transfer : `str`, optional
             Method to use to transfer the Zip into the datastore.
+        transfer_dimensions : `bool`, optional
+            If `True`, dimension record data associated with the new datasets
+            will be transferred from the Zip file, if present.
+        dry_run : `bool`, optional
+            If `True` the ingest will be processed without any modifications
+            made to the target butler and as if the target butler did not
+            have any of the datasets.
 
         Notes
         -----
-        Run collections are created as needed.
+        Run collections and dataset types are created as needed.
         """
         raise NotImplementedError()
 
--- a/python/lsst/daf/butler/direct_butler/_direct_butler.py
+++ b/python/lsst/daf/butler/direct_butler/_direct_butler.py
@@ -38,7 +38,6 @@ import collections.abc
 import contextlib
 import io
 import itertools
-import logging
 import numbers
 import os
 import uuid
@@ -1563,8 +1562,14 @@ class DirectButler(Butler):  # numpydoc ignore=PR02
             ):
                 self._datastore.emptyTrash(refs=refs)
 
-    @transactional
-    def ingest_zip(self, zip_file: ResourcePathExpression, transfer: str = "auto") -> None:
+    def ingest_zip(
+        self,
+        zip_file: ResourcePathExpression,
+        transfer: str = "auto",
+        *,
+        transfer_dimensions: bool = False,
+        dry_run: bool = False,
+    ) -> None:
         """Ingest a Zip file into this butler.
 
         The Zip file must have been created by `retrieve_artifacts_zip`.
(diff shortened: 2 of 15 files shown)


python/lsst/daf/butler/_butler.py
```diff
@@ -1385,10 +1392,17 @@ def ingest_zip(self, zip_file: ResourcePathExpression, transfer: str = "auto") -
             Path to the Zip file.
         transfer : `str`, optional
             Method to use to transfer the Zip into the datastore.
+        transfer_dimensions : `bool`, optional
+            If `True`, dimension record data associated with the new datasets
+            will be transferred from the Zip file, if present.
+        dry_run : `bool`, optional
+            If `True` the ingest will be processed without any modifications
+            made to the target butler and as if the target butler did not
+            have any of the datasets.
```
andy-slac, line 1401: Not clear to me how this is useful. I'd imagine that dry-run would try to detect any inconsistencies between data in butler and new data, but "as if the target butler did not have any of the dataset" implies that it will say OK to anything?

python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -1618,25 +1630,30 @@ def ingest_zip(self, zip_file: ResourcePathExpression, transfer: str = "auto") -
             if registered:
                 _LOG.verbose("Created RUN collection %s as part of zip ingest", run)
 
-        # Do not need expanded dataset refs so can ignore the return value.
-        self._ingest_file_datasets(datasets)
+        progress = Progress("lsst.daf.butler.Butler.ingest", level=VERBOSE)
+        import_info = self._prepare_ingest_file_datasets(
+            datasets, progress, dry_run=dry_run, transfer_dimensions=transfer_dimensions
+        )
```
andy-slac, line 1636: Should this be done in the same transaction below? I think it modifies the database, and if transaction below fails, you will end up with partial modifications.

python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -1661,29 +1678,39 @@ def _ingest_file_datasets(
...
-        )
+    def _ingest_file_datasets(
+        self,
+        datasets: Sequence[FileDataset],
+        import_info: _ImportDatasetsInfo,
+        progress: Progress,
+        *,
+        dry_run: bool = False,
+    ) -> None:
+        with self.transaction():
```
andy-slac: `ingest_zip()` already has a transaction open when calling this method, do we want a nested transaction here?

python/lsst/daf/butler/script/ingest_zip.py
```diff
@@ -46,7 +48,14 @@ def ingest_zip(
         URI string of the location of the Zip file.
     transfer : `str`, optional
         Transfer mode to use for ingest.
+    transfer_dimensions : `bool`
+        Indicate whether dimensions should be transferred along with
+        datasets. It can be more efficient to disable this if it is known
+        that all dimensions exist.
```
andy-slac: But default is already `False`? Or is default for parameter different from CLI default?

python/lsst/daf/butler/tests/hybrid_butler.py
```diff
@@ -268,9 +268,16 @@ def _exists_many(
...
+        self,
+        zip_file: ResourcePathExpression,
+        transfer: str = "auto",
+        *,
+        transfer_dimensions: bool = False,
+        dry_run: bool = False,
+    ) -> None:
         # Docstring inherited.
-        return self._direct_butler.ingest_zip(zip_file, transfer=transfer)
+        return self._direct_butler.ingest_zip(zip_file, transfer=transfer, dry_run=dry_run)
```
andy-slac: Pass `transfer_dimensions` too?

andy-slac (approved):

python/lsst/daf/butler/_butler.py
timj: That is something to consider for the future. At the moment it's matching the `transfer_from` behavior (Since it's passing the parameter straight through).

python/lsst/daf/butler/direct_butler/_direct_butler.py
timj: This was part of the reorganization of the transfer_from code so it can be called from ingest. The prepare step cannot be in a transaction because it creates a dataset type and that is not allowed to be in a transaction. No other database changes happen. It's one of the reasons why it's a distinct piece of code. The dimension record writing and file ingest are in the transaction.

python/lsst/daf/butler/direct_butler/_direct_butler.py
timj: Good spot. That transaction is irrelevant.

python/lsst/daf/butler/script/ingest_zip.py
timj: The script interface gets called with every parameter from the CLI. I will clarify the text.

python/lsst/daf/butler/tests/hybrid_butler.py
timj: Clearly my test coverage is terrible...

python/lsst/daf/butler/script/ingest_zip.py
timj: I realized that I didn't really want the CLI default to be True so I just fixed that (people should be intentional about asking for dimension records to be inserted)

PR #1194, opened 2025-05-06
DM-50724: Use new mremove bulk API in emptyTrash

timj:
Needs cleaning up for exception groups.

Requires lsst/resources#117

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -422,6 +422,32 @@ class FileDatastore(GenericBaseDatastore[StoredFileInfo]):
             raise
         log.debug("Successfully deleted file: %s", location.uri)
 
+    def _delete_artifacts(self, locations: list[Location]) -> None:
+        """Delete multiple artifacts from the datastore."""
+        # Filter out all artifacts that are direct.
+        n_locations = len(locations)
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -2609,42 +2626,47 @@ def emptyTrash(self, ignore_errors: bool = True) -> None:
...
-                    else:
-                        raise
+                artifacts_to_delete.append(location)
+
+            # Now do the deleting.
+            log.debug("Now removing %d file artifacts from datastore %s", len(artifacts_to_delete), self.name)
+
+            try:
+                self._delete_artifacts(artifacts_to_delete)
+            except* FileNotFoundError:
```
andy-slac, line 2652: But this is handled already in `_delete_artifacts`?

andy-slac (approved):
Looks great.

python/lsst/daf/butler/datastores/fileDatastore.py
timj: Yes. But this was a nice comment I didn't want to delete and I didn't think it would do any harm. Maybe I should move the `_delete_artifact` code into `emptyTrash` instead...

PR #1195, opened 2025-05-07
DM-50727: Change pruneDatasets to only prune the requested datasets

timj:
Rather than emptyTrash always trying to clear the entire trash table, add the ability to constrain the removal by a specific set of refs (that generally are the refs that were just pruned). Also take the opportunity to chunk deletes into 50,000 files to reduce the size of transactions that could involve a million files.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/bridge/monolithic.py
+++ b/python/lsst/daf/butler/registry/bridge/monolithic.py
@@ -246,15 +246,16 @@ class MonolithicDatastoreRegistryBridge(DatastoreRegistryBridge):
         # table that is not listed in the records table. Such an
         # inconsistency would be missed by this query.
         info_in_trash = join_records(records_table._table.select(), self._tables.dataset_location_trash)
+        if selected_ids:
+            info_in_trash = info_in_trash.where(
+                self._tables.dataset_location_trash.columns["dataset_id"].in_(selected_ids)
+            )
+        info_in_trash = info_in_trash.with_for_update(skip_locked=True)
 
         # Run query, transform results into a list of dicts that we can later
         # use to delete.
         with self._db.query(info_in_trash) as sql_result:
             rows = [dict(row, datastore_name=self.datastoreName) for row in sql_result.mappings()]
-
-        # Only keep rows that match the external override selection.
-        if selected_ids is not None:
-            rows = [row for row in rows if row["dataset_id"] in selected_ids]
 
         # It is possible for trashed refs to be linked to artifacts that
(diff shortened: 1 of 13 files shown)


python/lsst/daf/butler/registry/bridge/monolithic.py
```diff
@@ -249,6 +252,10 @@ def join_records(
         with self._db.query(info_in_trash) as sql_result:
             rows = [dict(row, datastore_name=self.datastoreName) for row in sql_result.mappings()]
 
+        # Only keep rows that match the external override selection.
+        if selected_ids is not None:
+            rows = [row for row in rows if row["dataset_id"] in selected_ids]
```
andy-slac: I am worried that this still leaves a chance for a deadlock in case one uses `admin empty-trash` concurrently with regular removal. We probably need to use explicit locking to avoid that. What I would do is something like (line 248 above):

    info_in_trash = join_records(records_table._table.select(), self._tables.dataset_location_trash)
    if selected_ids:
        info_in_trash = info_in_trash.where(self._tables.dataset_location_trash.columns["dataset_id"].in_(selected_ids))
    info_in_trash = info_in_trash.with_for_update(skip_locked=True)

And I think that we want similar `.where()` for `items_in_trash` below too.

andy-slac (approved):
Looks good, though I suspect it may still deadlock under rare circumstances. I left one suggestion if you want to try it.

python/lsst/daf/butler/registry/bridge/monolithic.py
```diff
@@ -243,6 +246,11 @@ def join_records(
         # table that is not listed in the records table. Such an
         # inconsistency would be missed by this query.
         info_in_trash = join_records(records_table._table.select(), self._tables.dataset_location_trash)
+        if selected_ids:
+            info_in_trash = info_in_trash.where(
+                self._tables.dataset_location_trash.columns["dataset_id"].in_(selected_ids)
+            )
+        info_in_trash = info_in_trash.with_for_update(skip_locked=True)
```
timj, line 253: @dhirving are you thinking that there should also be an order_by clause here to sort the results (eg by path if sorting by UUID is less efficient)?  cc/ @andy-slac

dhirving: I don't know enough about the internals of how Postgres takes row locks.  But I doubt an `ORDER BY` will give us any extra guarantees here, and it will probably hurt performance.

PR #1198, opened 2025-05-12
DM-50841: Butler server authorization using Gafaelfawr groups

dhirving:
Butler server can now authorize access to specific repositories based on the user's Gafaelfawr group memberships.  This will be used to restrict access to Data Preview 1 ahead of its public release.

Prior to this PR, there was no concept of configuration specific to Butler server -- everything was shared with the `DirectButler` configuration.  `pydantic-settings` is now used to configure the server, adding configuration variables for group membership restrictions and the Gafaelfawr endpoint.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/remote_butler/server/_gafaelfawr.py
+++ b/python/lsst/daf/butler/remote_butler/server/_gafaelfawr.py
@@ -44,10 +44,12 @@ class GafaelfawrClient:
     base_url : `str`
         The top-level HTTP path where Gafaelfawr can be found (e.g.
         ``"https://data-int.lsst.cloud"``).
+    transport : ``httpx.AsyncBaseTransport``, optional
+        Override the HTTP client's transport.  (For unit tests).
     """
 
-    def __init__(self, base_url: str) -> None:
-        self._client = httpx.AsyncClient(base_url=base_url)
+    def __init__(self, base_url: str, *, transport: httpx.AsyncBaseTransport | None = None) -> None:
+        self._client = httpx.AsyncClient(base_url=base_url, transport=transport)
 
     async def get_groups(self, user_token: str) -> list[str]:
         response = await self._client.get(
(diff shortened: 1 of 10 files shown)


python/lsst/daf/butler/remote_butler/server/_gafaelfawr.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+
+    def __init__(self, base_url: str, *, transport: httpx.AsyncBaseTransport | None = None) -> None:
+        if transport is None:
+            transport = httpx.AsyncHTTPTransport(retries=3)
+        self._client = httpx.AsyncClient(base_url=base_url, transport=transport, timeout=20.0)
+
+    async def get_groups(self, user_token: str) -> list[str]:
+        response = await self._client.get(
+            "/auth/api/v1/user-info", headers=get_authentication_headers(user_token)
```
rra: It's probably slightly more future-proof to move `/auth` here into the base URL. I think the `/auth` prefix will only be used when everything is running in the same domain. The long-term intent is to have Gafaelfawr run under auth.data.lsst.cloud (for example), and in that configuration we'll probably drop the `/auth` prefix.

python/lsst/daf/butler/remote_butler/server/_gafaelfawr.py
```diff
@@ -0,0 +1,127 @@
...
+        return [group.name for group in info.groups]
+
+
+class _GafaelfawrUserInfo(pydantic.BaseModel):
+    username: str
+    groups: list[_GafaelfawrGroup] | None = None
+
+
+class _GafaelfawrGroup(pydantic.BaseModel):
+    name: str
```
rra, line 71: Apologies for having to duplicate these models; will definitely fix this once I have a moment to breathe and make a proper Gafaelfawr client library.

rra (approved):
Gafaelfawr parts look good with one minor comment.

PR #1200, opened 2025-05-13
DM-50855: Reduce blocking in remove_from_chain

dhirving:
Allow multiple concurrent calls to `Butler.collections.remove_from_chain`, to fix an issue where `butler remove-runs` can block indefinitely while waiting for long-lived parallel transactions to complete.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/collections/_base.py
+++ b/python/lsst/daf/butler/registry/collections/_base.py
@@ -570,7 +570,16 @@ class DefaultCollectionManager(CollectionManager[K]):
             child_collection_names,
             # Removing members from a chain can't create collection cycles
             skip_cycle_check=True,
+            # It is OK for multiple instances of `remove_from_collection_chain`
+            # to run concurrently on the same collection, because it doesn't
+            # read/modify the position numbers of the children -- it only
+            # deletes existing rows.
+            #
+            # However, other chain modification operations must still be
+            # blocked to avoid consistency issues.
+            exclusive_lock=False,
         ) as c:
+            self._block_for_concurrency_test()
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -570,7 +570,16 @@ def remove_from_collection_chain(
             child_collection_names,
             # Removing members from a chain can't create collection cycles
             skip_cycle_check=True,
+            # It is OK for multiple instances of `remove_from_collection_chain`
+            # to run concurrently on the same collection, because it doesn't
+            # read/modify the position numbers of the children -- it only
+            # deletes existing rows.
+            #
+            # However, other chain modification operations must still be
+            # blocked to avoid consistency issues.
```
andy-slac, line 579: I wonder if it could be done better by exclusive lock on the collection_chain rows instead of parent collection record? Locking the row that you are not going to modify feels a little odd.

andy-slac (approved):
Looks good.

python/lsst/daf/butler/registry/collections/_base.py
dhirving: When I was writing this originally it wasn't totally obvious what set of rows that would be.  For `prepend` and `append` you don't modify the existing `collection_chain` rows... you effectively need a reservation on "position one greater/lower than the lowest/highest existing row" (with some additional quirks if the collection is already present in the chain.)

So it seemed more straightforward to use a single lock for the parent collection than trying to reason about the effect of locking some set of rows in the chain table.

(If it was just `remove` we wouldn't need the lock at all -- `DELETE` will already take a lock on all the relevant rows internally.)

PR #1203, opened 2025-05-19
DM-50996: Add unlink_from_chains parameter to removeRuns and reorganize transactions

timj:
Have separate transaction for the registry manipulation and the trash emptying. Previously the remove-runs CLI created its own transaction which included emptyTrash.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_butler/_direct_butler.py
+++ b/python/lsst/daf/butler/direct_butler/_direct_butler.py
@@ -2339,6 +2339,7 @@ class DirectButler(Butler):  # numpydoc ignore=PR02
 
         # Find all the registered instruments (if "instrument" is in the
         # universe).
+        instruments: set[str] = set()
         if "instrument" in self.dimensions:
             instruments = {rec.name for rec in self.query_dimension_records("instrument", explain=False)}
 
--- a/python/lsst/daf/butler/direct_butler/_direct_butler.py
+++ b/python/lsst/daf/butler/direct_butler/_direct_butler.py
@@ -38,6 +38,7 @@ import collections.abc
 import contextlib
 import io
 import itertools
+import math
 import numbers
 import os
 import uuid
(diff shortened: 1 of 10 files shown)


python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -1443,27 +1443,40 @@ def _exists_many(
...
+                        raise AssertionError("Internal error: Collection parents required but not received")
+                    for parent in info.parents:
+                        parents_to_children[parent].add(info.name)
+
+            # Get all the datasets from these runs.
+            refs = self._query_all_datasets(
+                [info.name for info in collections_info], find_first=False, limit=None
+            )
+
         with self._datastore.transaction(), self._registry.transaction():
```
timj: Question is whether this should do something like:

```python
for chunk in chunk_iterable(refs, chunk_size=500_000):
    with self._datastore.transaction(), self._registry.transaction():
        unlink _from chains
        datastore.trash(chunk)
        pruneDatasets(chunk, ...)

    datastore.emptyTrash(chunk)
registry.removeCollection(name)
```
on the basis that if it fails inside a chunk, rerunning the command should work just fine and it shouldn't be a problem if some datasets were deleted -- you wanted them deleted anyhow.

python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -1443,47 +1445,99 @@ def _exists_many(
...
+        with self._registry.transaction():
+            # This will fail if caller is not unlinking from chains but the
+            # RUN is in a chain -- but we have already deleted all the datasets
+            # by this point.
+            if unlink_from_chains:
+                # Use deterministic order for deletions to attempt to minimize
+                # risk of deadlocks for parallel deletes.
+                for parent in sorted(parents_to_children):
+                    self.collections.remove_from_chain(parent, sorted(parents_to_children[parent]))
+            for name in names:
```
dhirving: I think you'll want to sort the run names in this loop, too, for the same reason as the parents.  (We don't explicitly take a lock when removing a run collection, but Postgres implicitly takes a row lock when doing the delete.)

dhirving (approved):
Looks good.  One minor tweak that might help avoid deadlock.

PR #1206, opened 2025-05-20
DM-50958: Change column order in PK of dataset_location_trash

andy-slac:
Also avoids running a separate query to check which dataset exists when moving them to trash.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/bridge/monolithic.py
+++ b/python/lsst/daf/butler/registry/bridge/monolithic.py
@@ -38,6 +38,8 @@ from typing import TYPE_CHECKING, cast
 
 import sqlalchemy
 
+from lsst.utils.iteration import chunk_iterable
+
 from ..._dataset_ref import DatasetId
 from ..._named import NamedValueSet
 from ...datastore.stored_file_info import StoredDatastoreItemInfo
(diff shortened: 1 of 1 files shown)


python/lsst/daf/butler/registry/bridge/monolithic.py
```diff
@@ -91,30 +91,36 @@ def _makeTableSpecs(datasets: type[DatasetRecordStorageManager]) -> _TablesTuple
     # We want the dataset_location and dataset_location_trash tables
     # to have the same definition, aside from the behavior of their link
     # to the dataset table: the trash table has no foreign key constraint.
-    dataset_location_spec = ddl.TableSpec(
+    # The order of columns in dataset_location_trash is reversed, it is more
```
timj, line 94: Are we sure it's not fine to also reverse it for dataset_location on the basis that for a small butler this will also result in a better query plan before a sufficient number of rows exist for ANALYZE?

python/lsst/daf/butler/registry/bridge/monolithic.py
```diff
@@ -168,23 +174,36 @@ def insert(self, refs: Iterable[DatasetIdRef]) -> None:
 
     def forget(self, refs: Iterable[DatasetIdRef]) -> None:
         # Docstring inherited from DatastoreRegistryBridge
-        rows = self._refsToRows(self.check(refs))
-        self._db.delete(self._tables.dataset_location, ["datastore_name", "dataset_id"], *rows)
+        with self._db.transaction():
+            for refs_chunk in chunk_iterable(refs, 50_000):
```
timj, line 180: What's the motivation for the chunking here? Can there be a comment explaining why? Is it because the performance goes out of control if the chunk is too large now that there is a deleteWhere instead of a simple delete?

python/lsst/daf/butler/registry/bridge/monolithic.py
```diff
@@ -168,23 +174,36 @@ def insert(self, refs: Iterable[DatasetIdRef]) -> None:
...
-        #         WHERE datastore_name=? AND dataset_id IN (?);
-        # ...but the Database interface doesn't support those kinds of queries
-        # right now.
-        rows = self._refsToRows(self.check(refs))
+        location = self._tables.dataset_location
+        location_trash = self._tables.dataset_location_trash
         with self._db.transaction():
-            self._db.delete(self._tables.dataset_location, ["datastore_name", "dataset_id"], *rows)
-            self._db.insert(self._tables.dataset_location_trash, *rows)
+            for refs_chunk in chunk_iterable(refs, 50_000):
```
timj, line 193: Can you add a small comment explaining why deleting N rows and inserting N rows is less efficient than reusing a where clause for the insert and delete?

timj (approved):
Thanks.

python/lsst/daf/butler/registry/bridge/monolithic.py
andy-slac: Theoretically I think it is better to have datastore name as leading part of the key, but in practice I don't think we'll see any difference. With small number of datasets I would not care about performance too much, and more important question would be how frequently ANALYZE is going to run.

python/lsst/daf/butler/registry/bridge/monolithic.py
andy-slac: I am worried that with many millions parameters the query becomes too large. I saw errors in my case which looked like there is a limit somewhere on query size (at least with EXPLAIN). The previous code used `Database.delete` and `Database.insert` which do 1k-chunks internally (either in Database or in sqlalchemy). Here I'm constructing WHERE subquery myself, so I need to limit it to a reasonable size. I'll add a comment.

PR #1209, opened 2025-05-22
DM-51060: Move chunking from removeRuns to pruneDatasets

timj:
## Checklist

- [ ] ran Jenkins


dhirving (approved):
Looks good to me.  This does change the contract of `pruneDatasets`: the registry changes are no longer all-or-nothing.  But from earlier discussions I think that's what you wanted.

PR #1212, opened 2025-05-30
DM-51075: Implement transfer_from RemoteButler

dhirving:
RemoteButler can now be used as the source Butler in `transfer_from()`.

This required re-factoring `FileDatastore.transfer_from()` to break it up into smaller pieces, so that execution could be divided between Butler server and the client side.  A `FileTransferSource` protocol was added to define the methods required for this.  This protocol is now implemented by `FileDatastore`, `ChainedDatastore`, and the new `RemoteFileTransferSource`.  Some `ChainedDatastore`-specific code was moved out of `FileDatastore` into `ChainedDatastore` itself as a result.

There are two known issues with this initial release, see [DM-51301](https://rubinobs.atlassian.net/browse/DM-51301) and [DM-51302](https://rubinobs.atlassian.net/browse/DM-51302).  Both will require additional re-organization of the existing code to resolve.

This PR will probably be easiest to review commit-by commit, since a lot of code was shuffled between files and functions.

## Checklist

- [x] ran Jenkins

[DM-51301]: https://rubinobs.atlassian.net/browse/DM-51301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DM-51302]: https://rubinobs.atlassian.net/browse/DM-51302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/lsst/daf/butler/datastore/_datastore.py
+++ b/python/lsst/daf/butler/datastore/_datastore.py
@@ -54,7 +54,7 @@ from .._config import Config, ConfigSubset
 from .._exceptions import DatasetTypeNotSupportedError, ValidationError
 from .._file_dataset import FileDataset
 from .._storage_class import StorageClassFactory
-from ._transfer import FileTransferInfo
+from ._transfer import FileTransferInfo, FileTransferSource
 from .constraints import Constraints
 
 if TYPE_CHECKING:
@@ -284,7 +284,7 @@ class DatasetRefURIs(abc.Sequence):
         return f"DatasetRefURIs({repr(self.primaryURI)}, {repr(self.componentURIs)})"
 
 
-class Datastore(metaclass=ABCMeta):
+class Datastore(FileTransferSource, metaclass=ABCMeta):
     """Datastore interface.
 
     Parameters
@@ -869,7 +869,7 @@ class Datastore(metaclass=ABCMeta):
 
     def transfer_from(
         self,
-        source_datastore: Datastore,
+        source_datastore: FileTransferSource,
         refs: Collection[DatasetRef],
         transfer: str = "auto",
         artifact_existence: dict[ResourcePath, bool] | None = None,
@@ -1421,7 +1421,7 @@ class Datastore(metaclass=ABCMeta):
         """
         raise NotImplementedError()
 
-    def _get_file_info_for_transfer(
+    def get_file_info_for_transfer(
         self, refs: Iterable[DatasetRef], artifact_existence: dict[ResourcePath, bool]
     ) -> dict[DatasetId, list[FileTransferInfo]]:
         raise NotImplementedError(f"Transferring files is not supported by datastore {self}")
@@ -1500,7 +1500,7 @@ class NullDatastore(Datastore):
 
     def transfer_from(
         self,
-        source_datastore: Datastore,
+        source_datastore: FileTransferSource,
         refs: Iterable[DatasetRef],
         transfer: str = "auto",
         artifact_existence: dict[ResourcePath, bool] | None = None,
--- a/python/lsst/daf/butler/datastore/_datastore.py
+++ b/python/lsst/daf/butler/datastore/_datastore.py
@@ -54,7 +54,7 @@ from .._config import Config, ConfigSubset
 from .._exceptions import DatasetTypeNotSupportedError, ValidationError
 from .._file_dataset import FileDataset
 from .._storage_class import StorageClassFactory
-from ._transfer import FileTransferInfo, FileTransferSource
+from ._transfer import FileTransferMap, FileTransferSource
 from .constraints import Constraints
 
 if TYPE_CHECKING:
@@ -1421,9 +1421,12 @@ class Datastore(FileTransferSource, metaclass=ABCMeta):
         """
         raise NotImplementedError()
 
-    def get_file_info_for_transfer(
+    def get_file_info_for_transfer(self, dataset_ids: Iterable[DatasetId]) -> FileTransferMap:
+        raise NotImplementedError(f"Transferring files is not supported by datastore {self}")
+
+    def locate_missing_files_for_transfer(
         self, refs: Iterable[DatasetRef], artifact_existence: dict[ResourcePath, bool]
-    ) -> dict[DatasetId, list[FileTransferInfo]]:
+    ) -> FileTransferMap:
         raise NotImplementedError(f"Transferring files is not supported by datastore {self}")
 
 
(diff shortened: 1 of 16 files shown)


python/lsst/daf/butler/datastore/_datastore.py
```diff
@@ -283,7 +284,7 @@ def __repr__(self) -> str:
         return f"DatasetRefURIs({repr(self.primaryURI)}, {repr(self.componentURIs)})"
 
 
-class Datastore(metaclass=ABCMeta):
+class Datastore(FileTransferSource, metaclass=ABCMeta):
```
timj, line 287: This pains me a little since Datastore hasn't needed to know about files as a concept before (eg Sasquatch and the proposed rows in database table as datasets) other than for ingest. I understand that from your point of view files are the only things that we care about at the moment. Maybe we should give up the idea that Datastore is anything other than files.

python/lsst/daf/butler/datastore/stored_file_info.py
```diff
@@ -395,3 +395,36 @@ class SerializedStoredFileInfo(pydantic.BaseModel):
 
     file_size: int
     """Size of the serialized dataset in bytes."""
+
+
+def make_datastore_path_relative(path: str) -> str:
```
timj, line 400: Did you consider making this a method on StoredFileInfo? You use it as if it would be:

```python
file_info.path = make_datastore_path_relative(file_info.path)
```
vs
```python
new_file_info = file_info.make_datastore_path_relative()
```

Although I don't really understand how this works. If you have a direct ingest file how does it end up working at the other end if the URI scheme and bucket are removed from it. Don't you end up with a path with a stripped leading `/` that has no idea where it came from?

python/lsst/daf/butler/datastore/stored_file_info.py
```diff
@@ -395,3 +395,36 @@ class SerializedStoredFileInfo(pydantic.BaseModel):
...
+    Parameters
+    ----------
+    path : `str`
+        The file path from a `StoredFileInfo`.
+
+    Return
+    ------
+    normalized_path : `str`
+        The original path, if it was relative. Otherwise, a version of it that
+        was converted to a relative path.
```
timj: and maybe say, "stripping URI scheme and netloc from it in the process"?

python/lsst/daf/butler/remote_butler/_remote_file_transfer_source.py
```diff
@@ -0,0 +1,92 @@
...
+        missing_ids = {ref.id for ref in refs}
+        raise ValueError(f"Some datasets could not be found in {self.name}: {missing_ids}")
+
+    def mexists(
+        self, refs: Iterable[DatasetRef], artifact_existence: dict[ResourcePath, bool]
+    ) -> dict[DatasetRef, bool]:
+        # This is a stub.  It's not clear that this function makes much
+        # sense in the context of Butler Server -- our data releases should
+        # never include files that are accidentally missing.
+        # TODO DM-51302: Rework the transfer_from process so that we can use
```
timj, line 82: I agree that if files are missing when client/server is being used then something is broken. The "missing file artifact" functionality is entirely for batch production and prompt processing where a quantum can fail to write a dataset.

python/lsst/daf/butler/remote_butler/server_models.py
```diff
@@ -416,3 +417,17 @@ class QueryAllDatasetsRequestModel(pydantic.BaseModel):
     bind: dict[str, ColumnLiteral]
     limit: int | None
     with_dimension_records: bool
+
+
+class GetFileTransferInfoRequestModel(pydantic.BaseModel):
+    MAX_ITEMS_PER_REQUEST: ClassVar[int] = 10000
```
timj:
```suggestion
    MAX_ITEMS_PER_REQUEST: ClassVar[int] = 10_000
```
Slightly more readable at a glance if we use the underscore separators.

tests/test_butler.py
```diff
@@ -2582,133 +2576,17 @@ def setUp(self) -> None:
     def tearDown(self) -> None:
         removeTestTempDir(self.root)
 
-    def create_butler(self, manager: str, label: str, config_file: str | None = None) -> Butler:
+    def create_butler(self, manager: str | None, label: str, config_file: str | None = None) -> Butler:
+        if manager is None:
+            manager = (
+                "lsst.daf.butler.registry.datasets.byDimensions.ByDimensionsDatasetRecordStorageManagerUUID"
```
timj, line 2582: At some point we will clear out the code that was trying to be explicit about UUIDs. We aren't going back 😄

timj (approved):
Looks great.

python/lsst/daf/butler/datastore/_datastore.py
dhirving: I could have it return a separate `FileTransferSource` object via like a `Datastore.get_file_transfer_source()` instead, similar to what the RemoteButler version does.  Then the other Datastores can just raise NotImplemented in one place instead of multiple.

python/lsst/daf/butler/datastore/stored_file_info.py
dhirving: I did consider that, but `StoredFileInfo` is some sort of weird non-frozen frozen dataclass (it calls `__setattr__` to override the frozen-ness?), I had no idea how to safely "copy with replacement", it would have just made the unit tests harder to write, and I'm actually manipulating the data in the serialized version rather than `StoredFileInfo` itself.

The point of this is that we don't want the other end to see the server's URI scheme and bucket -- the internal S3 URIs are none of the client's business.  There is a concept of `Location` in `Datastore` -- everywhere that manipulates a `StoredFileInfo` has a `Location` attached to it with the actual full path to the file (this is a thing that pre-exists me.)  So for the client side, the `Location` is a presigned HTTP URL and the `StoredFileInfo.path` is a relative path.

But then there are places that want to know a relative name to put a file on disk... (mostly `retrieveArtifacts`, and then apparently the internals of `transfer_from` care whether this path is relative or absolute.)  So we need to give the client some path name it can use, without leaking internal details.

I expect that going forward all the datastore records at IDF will always contain relative paths, but `dp02` still has a lot of absolute URLs.

python/lsst/daf/butler/datastore/_datastore.py
dhirving: I think if we did a 'database Datastore' thing this general structure would still work, but there would need to be some extra levels of indirection for 'loading a thing from the DB and converting it to a byte array'.

PR #1215, opened 2025-06-10
DM-51302: Handle missing transfer datastore records for Butler server

dhirving:
When using RemoteButler as the source for `transfer_from`, `skip_missing=True` will now skip importing refs that do not have datastore records available from the server.

There are two side-effects from this re-organization:
1. "Accidentally" missing files, for which the datastore has a database record but the file is gone, will now raise an error when they are transferred.
2. The number of database queries for the datastore is cut in half for transfer_from, since mexists is no longer called.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastores/chainedDatastore.py
+++ b/python/lsst/daf/butler/datastores/chainedDatastore.py
@@ -54,7 +54,7 @@ from lsst.utils.introspection import get_full_type_name
 from lsst.utils.logging import getLogger
 
 from .._dataset_ref import DatasetId
-from ..datastore import FileTransferMap, FileTransferSource
+from ..datastore import FileTransferMap
 from .fileDatastore import FileDatastore
 
 if TYPE_CHECKING:
@@ -1297,7 +1297,7 @@ class ChainedDatastore(Datastore):
 
     def transfer_from(
         self,
-        source_datastore: FileTransferSource,
+        source_records: FileTransferMap,
         refs: Collection[DatasetRef],
         transfer: str = "auto",
         artifact_existence: dict[ResourcePath, bool] | None = None,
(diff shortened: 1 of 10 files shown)


python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -1239,65 +1238,47 @@ def export(
...
+                found = datastore.get_file_info_for_transfer(unassigned_ids)
+                found_acceptable_datastore = True
+                output.update(found)
+                unassigned_ids -= found.keys()
+            except NotImplementedError:
+                pass
+
+        if not found_acceptable_datastore:
+            types = {get_full_type_name(d) for d in self.datastores}
+            raise TypeError(f"ChainedDatastore encountered that had no FileDatastores. Had {','.join(types)}")
```
timj: Maybe the message should say "had no datastores able to provide file transfer information" ?

python/lsst/daf/butler/datastores/file_datastore/transfer.py
```diff
@@ -0,0 +1,104 @@
...
+
+log = getLogger(__name__)
+
+
+def retrieve_file_transfer_records(
+    source_datastore: FileTransferSource,
+    refs: Iterable[DatasetRef],
+    artifact_existence: dict[ResourcePath, bool],
+) -> FileTransferMap:
+    """Look up the datastore records corresponding to the given `DatasetRef`s.
```
timj:
```suggestion
    r"""Look up the datastore records corresponding to the given `DatasetRef`\ s.
```
I don't think restructured text likes a pluralized form. Rephrasing as "to the given datasets" might be simpler.

timj (approved):
This looks okay but I would like to get feedback from PP. Will comment on Jira.

PR #1216, opened 2025-06-10
DM-51269: Add fallback instrument to ObsCore configuration

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/obscore/_records.py
+++ b/python/lsst/daf/butler/registry/obscore/_records.py
@@ -394,12 +394,14 @@ class DafButlerRecordFactory(RecordFactory):
         dataId = ref.dataId
         record: dict[str, str | int | float | UUID | None] = {}
 
-        instrument_name = cast(str, dataId.get("instrument"))
+        instrument_name = cast(str | None, dataId.get("instrument", self.config.fallback_instrument))
         record["instrument_name"] = instrument_name
         if self.schema.dataset_fk is not None:
             record[self.schema.dataset_fk.name] = ref.id
 
-        record["facility_name"] = self.config.facility_map.get(instrument_name, self.config.facility_name)
+        record["facility_name"] = self.config.facility_map.get(
+            instrument_name or "", self.config.facility_name
+        )
 
         timespan = dataId.timespan
         if timespan is not None:
--- a/doc/changes/DM-51269.misc.rst
+++ b/doc/changes/DM-51269.misc.rst
@@ -1,0 +1,2 @@
+Added a ``fallback_instrument`` to ObsCore configurations.
+This instrument is used when an ObsCore record is constructed from a dataset type that has no instrument defined.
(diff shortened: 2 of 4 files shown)


python/lsst/daf/butler/registry/obscore/_records.py
```diff
@@ -394,12 +394,16 @@ def make_universe_records(self, ref: DatasetRef) -> Record:
         dataId = ref.dataId
         record: dict[str, str | int | float | UUID | None] = {}
 
-        instrument_name = cast(str, dataId.get("instrument"))
+        instrument_name = cast(str | None, dataId.get("instrument"))
```
timj: mypy starts getting annoyed if I try to use a default fallback instrument here because dataId.get doesn't allow a None default.

TallJimbo: I think `DataCoordinate.get` probably needs some `@typing.override` work; that's the usual pattern for `Mapping.get`-likes, and I don't think I realized that back when I first wrote `DataCoordinate`.

andy-slac: `DataCoordinate.get` already has `@overloads`. I think you can just say:

    instrument_name: str | None = dataId.get("instrument", "")

and then check for an empty string (`None` is needed because `fallback_instrument` can be `None`)

andy-slac (approved):
Looks good.

python/lsst/daf/butler/registry/obscore/_records.py
timj: I initially had:
```python
instrument_name = cast(str | None, dataId.get("instrument", self.config.fallback_instrument))
```
but mypy definitely did not like that:
```
 Argument 2 to "get" of "DataCoordinate" has incompatible type "str | None"; expected "str"  [arg-type]
```
BUT, thank you for the hint because the real fix is to add:
```python
@overload
def get(self, key: str, default: None) -> None: ...
```
to DataCoordinate. That simplifies everything so I'm going to do that.

andy-slac: Hmm, with `default: None` it should always return `None`? There is already this method:

    def get(self, key: str, default: DataIdValue | None = None) -> DataIdValue | None:

which I think should cover it.

timj: Except it doesn't cover it. Maybe mypy only looks at the overloads and not the implementation itself?

timj: I agree that the default controlling the return value is a bit odd because it only triggers for the case where the key wasn't found but we have this:

```python
    @overload
    def get(self, key: str) -> DataIdValue | None: ...

    @overload
    def get(self, key: str, default: int) -> int: ...

    @overload
    def get(self, key: str, default: str) -> str: ...
```
already which I guess is saying that if you are specifying a default str you expect the key to have a str. Whereas None is not like that.

timj: Okay. If I explicitly add the `DataIdValue | None` option as an overload things do work. @andy-slac does this look okay to you?

andy-slac: 👍 Looks good.

PR #1218, opened 2025-06-12
DM-51363: move some test data into Python tree for downstream access

TallJimbo:
## Checklist

- [x] ran Jenkins


timj (approved):
Nice cleanup. It does mean we have to be a bit more careful about changing the contents of these files since now they are a public interface (so I think this will need a news fragment).

PR #1225, opened 2025-06-18
DM-51373: Limit maximum concurrent queries per user

dhirving:
Limit the maximum number of concurrent long-running queries to 2 per user per Butler server replica.  The fact that this limit is applied on a per-replica basis makes it somewhat nondeterministic, but for now it is hopefully sufficient to stop simple errors in end-user code from consuming all available database resources.

## Checklist

- [x] ran Jenkins


dhirving:
The jenkins failure on linux_aarch64 only appears to be spurious -- it was in `meas_extensions_trailedSources` which this change couldn't possibly affect.

timj (approved):

PR #1226, opened 2025-06-23
DM-51495: Place exposure in visit field for obs core records

timj:
This allows the exposure ID to turn up in the lsst_visit record rather than being empty and saves us from needing and lsst_exposure column when in almost all cases the exposure ID is the same as the visit ID.

## Checklist

- [x] ran Jenkins


andy-slac (approved):
Looks great.

PR #1228, opened 2025-06-23
DM-51515: Use UTC in ObsCore t_min/t_max fields

timj:
Previously used TAI

## Checklist

- [ ] ran Jenkins


gpdf (approved):
Confirming that this is what ObsCore expects.  This is independent of the LSST-requirements-driven use of TAI for time stamps for individual observations is in catalogs.

PR #1229, opened 2025-06-28
DM-51593: Eliminate expandDataId loops

dhirving:
There were a number of places in the Butler internals that looped over `registry.expandDataId`, issuing a DB query per dataset per dimension (potentially millions).  These are now replaced by `DimensionDataAttacher`, which uses a query per dimension instead.

`DimensionDataAttacher` was also reworked so that it can operate on `DataCoordinate` values with `hasFull() = False`.  It now looks up the missing 'implied' data ID values from the dimension records it is attaching.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/dimensions/_record_set.py
+++ b/python/lsst/daf/butler/dimensions/_record_set.py
@@ -809,8 +809,7 @@ class DimensionDataAttacher:
             been referenced in at least one of the constructor arguments.
         data_ids : `~collections.abc.Iterable` [ `DataCoordinate` ]
             Data IDs to attach dimension records to (not in place; data
-            coordinates are immutable).  Must have full values (i.e. implied
-            as well as required dimensions).
+            coordinates are immutable).
         query : `.queries.Query`, optional
             A butler query that can be used to look up missing dimension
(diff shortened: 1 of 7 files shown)


dhirving:
This will probably be easier to review if you look at the first few commits separately, since that's where the internals of DimensionDataAttacher get shuffled around.

python/lsst/daf/butler/dimensions/_record_set.py
```diff
@@ -899,8 +898,28 @@ def lookup(self, records: _InProgressRecordDicts) -> None:
...
+                value = records.data_id.get(dimension)
+                if value is None:
+                    value = self._find_implied_value(dimension, records)
+                values.append(value)
+            return tuple(values)
+
+    def _find_implied_value(self, implied_dimension: str, records: _InProgressRecordDicts) -> DataIdValue:
+        for rec in records.done.values():
+            if implied_dimension in rec.definition.implied:
+                return rec.toDict()[implied_dimension]
```
TallJimbo: `getattr(rec, implied_dimension)` should be a lot more efficient.

TallJimbo (approved):

PR #1230, opened 2025-07-02
DM-51454: Add support for more geometry types to user expressions

andy-slac:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-51454.feature.md
+++ b/doc/changes/DM-51454.feature.md
@@ -1,1 +1,1 @@
-User query language adds support for functions which define sky regions - `CIRCLE`, `BOX`, `POLYGON`, and `REGION`.
+Added support to the query language to allow sky regions to be specified directly in an ADQL-like way. This includes `CIRCLE`, `BOX`, `POLYGON`, and `REGION`.
--- a/doc/changes/DM-51454.feature.md
+++ b/doc/changes/DM-51454.feature.md
@@ -1,0 +1,1 @@
+User query language adds support for functions which define sky regions - `CIRCLE`, `BOX`, `POLYGON`, and `REGION`.
--- a/doc/lsst.daf.butler/queries.rst
+++ b/doc/lsst.daf.butler/queries.rst
@@ -391,9 +391,13 @@ either numeric literals or bind identifiers (``:name`` syntax):
   sky region, all arguments are in degrees.
 - ``POLYGON(ra1, dec1, ra2, dec2, ...)`` - function which returns polygonal
   sky region, all arguments are in degrees, at least three pairs of coordinates
-  have to be provided.
+  have to be provided. Note that implementation for this function is based on
+  ``lsst.sphgeom.ConvexPolygon`` class which builds a convex hull out of all
+  vertices, thus ignoring their order. This is not strictly consistent with
+  ADQL definition that relies on ordering of the vertices and allows non-convex
+  polygons.
 - ``REGION('string')`` - function returning sky region, string argument uses
-  `IVOA SIAv2 POS`_ syntax.
+  `IVOA SIAv2 POS`_ syntax. Same caveat applies as for ``POLYGON`` function.
 
 .. _IVOA SIAv2 POS: https://ivoa.net/documents/SIA/20151223/REC-SIA-2.0-20151223.html#toc12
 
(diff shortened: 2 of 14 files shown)


doc/changes/DM-51454.feature.md
```diff
@@ -0,0 +1 @@
+User query language adds support for functions which define sky regions - `CIRCLE`, `BOX`, `POLYGON`, and `REGION`.
```
timj:
```suggestion
Added support to the query language to allow sky regions to be specified directly in an ADQL-like way. This includes `CIRCLE`, `BOX`, `POLYGON`, and `REGION`.
```

doc/lsst.daf.butler/queries.rst
```diff
@@ -364,16 +370,33 @@ consists of an identifier followed by zero or more comma-separated arguments
...
+Functions for defining sky coordinates or regions. All these functions accept
+either numeric literals or bind identifiers (``:name`` syntax):
+
+- ``POINT(ra, dec)`` - function which declares (or returns) sky coordinates
+  similarly to ADQL syntax, arguments are in degrees.
+- ``CIRCLE(ra, dec, radius)`` - function which returns circular sky region,
+  all arguments are in degrees.
+- ``BOX(center_ra, center_dec, width, height)`` - function which returns box
+  sky region, all arguments are in degrees.
+- ``POLYGON(ra1, dec1, ra2, dec2, ...)`` - function which returns polygonal
```
timj, line 392: Does the order of vertices matter?

python/lsst/daf/butler/registry/queries/expressions/parser/exprTree.py
```diff
@@ -488,6 +492,115 @@ def __str__(self) -> str:
...
+
+    def visit(self, visitor: TreeVisitor) -> Any:
+        # Docstring inherited from Node.visit
+        ra = self.ra.visit(visitor)
+        dec = self.dec.visit(visitor)
+        radius = self.radius.visit(visitor)
+        return visitor.visitCircleNode(ra, dec, radius, self)
+
+    def __str__(self) -> str:
+        return f"CIRCLE({self.ra}, {self.dec}, {self.radius})"
```
timj, line 522: I see none of the stringifications are tested -- is it easy to force them to be stringified somewhere? I assume you added the string form for a reason.

timj (approved):
Looks great. Thanks.

doc/lsst.daf.butler/queries.rst
andy-slac: It should matter. ADQL says

> ...each vertex connected to the next along a great circle and the last vertex implicitly connected to the first vertex.

And sphgeom only supports convex polygons. I do not know what happens when order is messed up and polygon is not convex, I probably need to add a test for that.

Or do you mean clock-wise vs counter-clock-wise?

python/lsst/daf/butler/registry/queries/expressions/parser/exprTree.py
andy-slac: String form is used only in diagnostics. I can add stringification to unit test for all nodes.

doc/lsst.daf.butler/queries.rst
andy-slac: Actually sphgeom ConvexPolygon says that it is a convex hull of the vertices, so the order should not matter for sphgeom. And that works differently from ADQL if ADQL allows non-convex polygon. ADQL refers to IVOA STC and that is a very dense read. I'll add more details to the doc.

andy-slac: OK, I added a caveat to the doc which explains how our POLYGON is different from ADQL POLYGON.

python/lsst/daf/butler/registry/queries/expressions/parser/exprTree.py
andy-slac: Added stringification tests to unit testing, and fixed some issues with stringification.

doc/lsst.daf.butler/queries.rst
timj: Thanks. I was mainly wanting some comment in the docs as to how important it is to get the ordering correct.

PR #1232, opened 2025-07-15
Fix handling of chained collections in resolve_wildcard (DM-51819)

andy-slac:
Children of chained collections whose name matches wildcard expression were not returned if they did not match. The fix is to add children names to explicit list of strings to query. Extended unit tests to cover this sort of failure.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/collections/_base.py
+++ b/python/lsst/daf/butler/registry/collections/_base.py
@@ -427,11 +427,13 @@ class DefaultCollectionManager(CollectionManager[K]):
         cache: CollectionRecordCache | None = None
         result: list[CollectionRecord[K]] = []
         done_keys: set[K] = set()
+        explicit_names = list(wildcard.strings)
         if wildcard.patterns:
-            if wildcard.strings:
+            if explicit_names or flatten_chains:
                 # To be efficient in case both patterns and strings are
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/registry/collections/_base.py
```diff
@@ -427,23 +427,31 @@ def filter_types(records: Iterable[CollectionRecord[K]]) -> Iterator[CollectionR
...
                         result.append(record)
                         done_keys.add(record.key)
-
-        if wildcard.strings:
+            if flatten_chains:
+                # If flattening include children names of all matching chains.
+                for record in all_records:
+                    if isinstance(record, ChainedCollectionRecord):
+                        if any(p.fullmatch(record.name) for p in wildcard.patterns):
+                            explicit_names.extend(record.children)
```
TallJimbo, line 449: Does this new logic need to be recursive, or are we already in something that's recursing here?

TallJimbo (approved):

python/lsst/daf/butler/registry/collections/_base.py
andy-slac: The logic that handles `explicit_names` below is recursive, here we just need to collect children names to trigger that logic.

PR #1233, opened 2025-07-16
DM-51824: Remove explicit DROP of temporary tables

andy-slac:
DROP causes problems when a transaction or server connection aborts inside context manager. Explicit drop is not needed for temp tables as their lifetime is managed by backend and they disappear either after the end of transaction (in Postgres) or when a session finishes.

## Checklist

- [ ] ran Jenkins


andy-slac:
Closed, wrong ticket.

PR #1235, opened 2025-07-18
DM-51866: Add hooks for CADC authentication in Butler server

dhirving:
Tweak Butler server for use by CADC by adding a way to use an HTTP datastore that cannot provide pre-signed URLs.

CADC's Storage Inventory system is an HTTP service that requires bearer token `Authorization` headers.  When accessing these URLs, the client now attaches headers to the `ResourcePath` instances similar to the `transfer_from` change in [DM-51301](https://github.com/lsst/daf_butler/pull/1217).

Added a new `DAF_BUTLER_SERVER_AUTHENTICATION` environment variable for the server, and a corresponding configuration parameter in the `ButlerConfig` for RemoteButler.  This is needed on the server to select between URL-signing and non-signing logic for providing HTTP URLs.  On the client side, the bearer token for CADC will not come from the same place that it does for rubin science platform and will be sent to a different HTTP service instead of the Butler server REST service.

Some `Butler.get()`-related code and Pydantic models were moved out of `FileDatastore` into Butler server, since the behavior now depends on the server configuration.

## Checklist

- [x] ran Jenkins

[DM-51301]: https://rubinobs.atlassian.net/browse/DM-51301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/lsst/daf/butler/remote_butler/_factory.py
+++ b/python/lsst/daf/butler/remote_butler/_factory.py
@@ -36,10 +36,10 @@ from lsst.daf.butler.repo_relocation import replaceRoot
 from .._butler_config import ButlerConfig
 from .._butler_instance_options import ButlerInstanceOptions
 from ..registry import RegistryDefaults
-from ._authentication import get_authentication_token_from_environment
 from ._config import RemoteButlerConfigModel
 from ._http_connection import RemoteButlerHttpConnection
 from ._remote_butler import RemoteButler, RemoteButlerCache
+from .authentication.rubin import RubinAuthenticationProvider
 
 
 class RemoteButlerFactory:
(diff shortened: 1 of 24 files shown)


python/lsst/daf/butler/remote_butler/_config.py
```diff
@@ -25,16 +25,24 @@
...
 
     url: AnyHttpUrl
+    authentication: AuthenticationMode = "rubin_science_platform"
 
 
-class RemoteButlerConfigModel(BaseModel):
-    """Configuration representing a remote butler."""
-
-    remote_butler: RemoteButlerOptionsModel
+AuthenticationMode: TypeAlias = Literal["rubin_science_platform", "cadc"]
```
timj, line 48: What's the reason for putting this at the end of the file rather than towards the top? Reading the code down from the top needs you to go past the first usage of the type to find out what it means.

python/lsst/daf/butler/remote_butler/_factory.py
```diff
@@ -93,42 +98,61 @@ def create_factory_from_config(config: ButlerConfig) -> RemoteButlerFactory:
...
+        self,
+        access_token: str,
+        *,
+        butler_options: ButlerInstanceOptions | None = None,
+        enable_datastore_cache: bool = False,
+    ) -> RemoteButler:
+        auth: RemoteButlerAuthenticationProvider
+        if self._config.authentication == "rubin_science_platform":
+            auth = RubinAuthenticationProvider(access_token)
+        else:
```
timj:
```suggestion
        elif self._config.authentication == "cadc":
```
? and then an else to complain if it's not a recognized value? (also below).  Or are you hoping that it's impossible to get to the else because the model validation won't allow it? (and mypy will complain if you put in an else here because it will say both options were checked?)

timj (approved):

python/lsst/daf/butler/remote_butler/_config.py
dhirving: I dunno, I ordered it from the top of the config hierarchy downward.  I usually figure it's better to have the main thing at the top and the minor details at the bottom, since when looking at the main thing people can just hover their mouse over a symbol to see the details if they care.

PR #1238, opened 2025-08-04
DM-52044: miscellaneous prep work for QG replacement

TallJimbo:
## Checklist

- [x] ran Jenkins


timj (approved):
Clearly we didn't need StorageClass subclasses to be dynamically generated from a config...

PR #1240, opened 2025-08-07
DM-42272: Preserve metadata when an exception escapes ButlerMDC context.

kfindeisen:
This PR ports a feature from Prompt Processing's `RecordFactoryContextAdapter` that lets exceptions hang on to, and be logged with, metadata from the point where they were raised.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/doc/changes/DM-42272.feature.md
+++ b/doc/changes/DM-42272.feature.md
@@ -1,0 +1,2 @@
+ButlerMDC now preserves metadata for exceptions that are raised in a context but handled outside it.
+This lets exceptions be logged by high-level handlers the same way they would have been if the problem were logged immediately.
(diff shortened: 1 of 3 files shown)


kfindeisen:
The original approach is fundamentally type-unsafe, will need to revisit.

kfindeisen:
Per Jim Bosch's advice, I gave up on trying to make it type-safe and just suppressed the error.

timj (approved):
Thank you. This looks great.

kfindeisen:
Is the "Jenkins Build" action supposed to say "Build succeeded" if [2 of 3 branches failed](https://ci.lsst.cloud/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/6020/pipeline/67) (with Jenkins-side problems, but still)?

timj:
The problem is that the other architectures never even started their "rebuild" so never registered that they existed with the PR.

PR #1242, opened 2025-08-19
DM-47563: Make query_all_datasets public

dhirving:
Make `Butler.query_all_datasets()` a public function, to allow users to more efficiently search for datasets of multiple types.

When using CLI tools that search for datasets with RemoteButler, it now uses a dedicated REST endpoint for `query_all_datasets`, reducing the number of HTTP calls required to complete these queries.

## Checklist

- [x] ran Jenkins


dhirving:
The Jenkins build failures on Linux are a completely unrelated C++ linker error -- it passed on Mac.

timj (approved):

PR #1243, opened 2025-08-19
DM-51314: Ensure exceptions propagate from server to client

dhirving:
Fix a few server-side exceptions that should have been sent to RemoteButler clients as meaningful exceptions, but instead were triggering an HTTP 500 Internal Server Error.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/dimensions/_coordinate.py
+++ b/python/lsst/daf/butler/dimensions/_coordinate.py
@@ -1005,10 +1005,12 @@ class _FullTupleDataCoordinate(_BasicTupleDataCoordinate):
         dimensions = self.universe.conform(dimensions)
         if self._dimensions == dimensions:
             return self
-        return DataCoordinate.from_full_values(
-            dimensions,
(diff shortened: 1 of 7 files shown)


python/lsst/daf/butler/dimensions/_coordinate.py
```diff
@@ -1005,10 +1005,12 @@ def subset(self, dimensions: DimensionGroup | Iterable[str]) -> DataCoordinate:
...
             return self
-        return DataCoordinate.from_full_values(
-            dimensions,
-            tuple(self[k] for k in dimensions.data_coordinate_keys),
-        )
+
+        try:
+            values = tuple(self[k] for k in dimensions.data_coordinate_keys)
+        except KeyError as e:
+            raise DimensionNameError(f"Data ID is missing value for dimension {str(e)}.")
```
timj, line 1012: Doesn't f-string force the `e` to be a `str`?

timj (approved):
Looks great.

PR #1244, opened 2025-08-26
DM-52180: Add "export-in-place" function for Prompt Processing

dhirving:
Add functions that will be used by Prompt Processing to write outputs back to the central repository in their final destination, without immediately writing registry entries for the datasets.  This shares the code from `QuantumBackedButler` that is used to set up a `Datastore` instance detached from Butler/Registry.

This supports a tweak to the DMTN-310 butler writer service to avoid doubling the S3 I/O requirements for prompt processing writes.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastores/fileDatastore.py
+++ b/python/lsst/daf/butler/datastores/fileDatastore.py
@@ -3050,17 +3050,11 @@ class FileDatastore(GenericBaseDatastore[StoredFileInfo]):
         if transfer == "auto" and directory is None:
             transfer = None
 
-        if transfer is not None and directory is None:
+        if transfer is not None and transfer != "direct" and directory is None:
             raise TypeError(f"Cannot export using transfer mode {transfer} with no export directory given")
 
         if transfer == "move":
             raise TypeError("Can not export by moving files out of datastore.")
-        elif transfer == "direct":
-            # For an export, treat this as equivalent to None. We do not
-            # want an import to risk using absolute URIs to datasets owned
-            # by another datastore.
-            log.info("Treating 'direct' transfer mode as in-place export.")
-            transfer = None
 
         # Force the directory to be a URI object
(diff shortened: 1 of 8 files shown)


python/lsst/daf/butler/datastores/fileDatastore.py
```diff
@@ -3050,17 +3050,11 @@ def export(
         if transfer == "auto" and directory is None:
             transfer = None
 
-        if transfer is not None and directory is None:
+        if transfer is not None and transfer != "direct" and directory is None:
             raise TypeError(f"Cannot export using transfer mode {transfer} with no export directory given")
 
         if transfer == "move":
             raise TypeError("Can not export by moving files out of datastore.")
-        elif transfer == "direct":
```
dhirving, line 3058: This tweak was needed because `transfer=None` returns relative URLs, and we need an absolute URL so we can access the files.  (There is some previously-unused code a little further down that returns absolute URLs for transfer=direct).

Arguably `transfer=None` should return absolute URLs because a relative URL doesn't actually let you find the file.  (Due to ChainedDatastore etc, without knowing which datastore it came from you don't have a root directory to attach.)  But I wasn't sure what the use case for the existing `transfer=None` logic was so I thought it better to leave that in place, since the behavior I need is consistent with the meaning of `transfer=direct` elsewhere.

dhirving:
I'm not delighted about this but I didn't really have a better idea to fulfill the requirement that "files should be written in-place in their final destination in the datastore".

python/lsst/daf/butler/datastores/fileDatastore.py
timj: `transfer=None` is you telling the datastore that the files are already in the datastore location so don't transfer them at all and register them in place.

timj (approved):

PR #1246, opened 2025-08-30
DM-52179: Make Butler.ingest() retryable

dhirving:
`Butler.ingest()` now has an optional `skip_existing` parameter that makes it ignore datasets that already exist in the repository instead of raising an error.  This makes it easier for applications to retry ingests.

As a side-effect, this moves the test for conflicting datastore records before file copying instead of after.  This will reduce the chances of a file getting incorrectly overwritten by a failed ingest. (But not totally prevent it, in the case of concurrent writes.)

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-52179.feature.md
+++ b/doc/changes/DM-52179.feature.md
@@ -1,0 +1,1 @@
+`Butler.ingest()` now has an optional `skip_existing` parameter that makes it ignore datasets that already exist in the repository instead of raising an error.  This makes it easier for applications to retry ingests.
--- a/python/lsst/daf/butler/direct_butler/_direct_butler.py
+++ b/python/lsst/daf/butler/direct_butler/_direct_butler.py
@@ -1797,6 +1797,7 @@ class DirectButler(Butler):  # numpydoc ignore=PR02
         *datasets: FileDataset,
         transfer: str | None = "auto",
         record_validation_info: bool = True,
+        skip_existing: bool = False,
     ) -> None:
         # Docstring inherited.
         if not datasets:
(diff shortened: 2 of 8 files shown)


doc/changes/DM-52179.feature.md
```diff
@@ -0,0 +1 @@
+`Butler.ingest()` now has an optional `skip_existing` parameter that makes it ignore datasets that already exist in the repository instead of raising an error.  This makes it easier for applications to retry ingests.
```
timj, line 1: @srp3rd this seems like something you were asking for in OODS.

python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -1808,22 +1809,58 @@ def ingest(
 
         import_info = self._prepare_ingest_file_datasets(datasets, progress)
 
+        new_datasets, existing_datasets = self._partition_datasets_by_known(datasets)
```
timj, line 1810: Maybe move this chunk before the `_prepare_ingest_file_datasets` so we can fail quicker?

python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -1808,22 +1809,58 @@ def ingest(
...
+
         with self.transaction():
             self._ingest_file_datasets(datasets, import_info, progress)
 
             # Bulk-insert everything into Datastore.
             # We do not know if any of the registry entries already existed
-            # (_importDatasets only complains if they exist but differ) so
-            # we have to catch IntegrityError explicitly.
+            # (_importDatasets only complains if they exist but differ).
+            # The _partition_datasets_by_known logic above should catch most
```
timj, line 1833: If concurrent writes are an issue would it help to move this logic inside the transaction?

timj (approved):
Looks good.

python/lsst/daf/butler/direct_butler/_direct_butler.py
dhirving: Nope -- whether or not the initial check is inside the transaction, the second concurrent writer to complete will fail at the point the transaction commits.

PR #1248, opened 2025-09-04
DM-52397: Remove daf_relation from registry query methods.

dhirving:
DirectButler now uses RemoteButler's shims for the registry query methods, to forward them to the new query system.

Added shims for `findDatasets`, `findRelatedDatasets` and the `check` parameter to query methods, which RemoteButler did not previously support.

`DataCoordinateQueryResult.materialize()` has been replaced with a no-op shim.

## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -2454,19 +2340,6 @@ def testCalibrationCollections(self):
                 (registry.expandDataId(instrument="Cam1", detector=2, exposure=2), bias2a),
             },
         )
-        self.assertEqual(
```
dhirving, line 2457: I'm not sure what to make of this test (and a few similar ones below.)  I removed these because they were failing, but I don't know if that is OK or not.

Not sure if I don't understand the intended behavior of `findRelatedDatasets`, if this was capturing weird unintended behavior of the old query system, if there is a bug in my new `findRelatedDatasets` shim, or there is a bug in the new query system.

A very similar test is run on the line above -- the only difference is no detector constraint and a different set of collections.  Given that bias is `(instrument, detector)`, I don't understand how changing the collections and adding a constraint could make additional exposure IDs be returned by this query.

The removed test fails with:
```
E       AssertionError: Items in the second set but not the first:
E       ({instrument: 'Cam1', detector: 2, exposure: 0, band: 'g', day_obs: 20200101, group: 'group0', physical_filter: 'Cam1-G'}, DatasetRef(DatasetType('bias', {instrument, detector}, Exposure, isCalibration=True), {instrument: 'Cam1', detector: 2}, run='imported_r', id=87f3e68d-258d-41b7-8ea5-edf3557ccb30))
E       ({instrument: 'Cam1', detector: 2, exposure: 3, band: 'g', day_obs: 20200101, group: 'group3', physical_filter: 'Cam1-G'}, DatasetRef(DatasetType('bias', {instrument, detector}, Exposure, isCalibration=True), {instrument: 'Cam1', detector: 2}, run='imported_r', id=87f3e68d-258d-41b7-8ea5-edf3557ccb30)
```
... meaning that it returned the identical result to the test above.  But this test expects an extra dataset from the added collection, with two new exposure IDs `0` and `3` that didn't appear above.  findFirst is true and the data IDs are the same between the  two datasets, so I would expect it to just get the original dataset from the front of the collection.  If exposures `0` and `3` were going to appear at all, I would think they should have showed up in the test above too.  (And I actually don't understand why those exposures don't show up in the test above, but the behavior is consistent between the old and new code there.)

dhirving: For the test that was removed, the new implementation is returning identical results to the test above.  (As I would have expected.)

python/lsst/daf/butler/registry/queries/_query_data_coordinates.py
```diff
@@ -99,10 +102,10 @@ def __iter__(self) -> Iterator[DataCoordinate]:
 
     @contextmanager
     def materialize(self) -> Iterator[LegacyDataCoordinateQueryResults]:
-        raise NotImplementedError()
+        yield self
```
dhirving, line 105: We didn't deprecate `materialize` for data IDs previously.  I just shimmed it as a no-op here.  In the unlikely event someone is using it, it should return the same results, just more slowly.

Would it be OK to just mark this as deprecated in this PR, or does that need another RFC?

TallJimbo: I think it's fine to mark it deprecated here, once as we create/have another ticket for its removal.

python/lsst/daf/butler/registry/tests/_registry.py
TallJimbo: I think this must have been a careless test that was capturing unintended behavior in the original implementation.  Or maybe it was desirable behavior for some weird special case, but it certainly doesn't make sense or conform to the docs.  :+1: to dropping the test.

TallJimbo (approved):

dhirving:
This can't merge down until Thursday afternoon, once I have deployed the updated Butler server with the forward-compatibility fix for QueryTree serialization from DM-52635.

PR #1249, opened 2025-09-04
DM-52180b: Allow ChainedDatastore ingest with transfer=None

dhirving:
Fix an issue encountered during integration testing of Prompt Processing with the Butler writer service.  Prompt Processing's target repository uses a ChainedDatastore, and the Butler code was explicitly disallowing ingests with transfer=None.

The exception was probably added to prevent confusion -- normally it would be difficult to split up the files correctly between the multiple roots in a ChainedDatastore.  However, PP's call to `transfer_datasets_to_datastore` writes the files into the correct place and the existing code works correctly if the files are where they need to be.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/datastores/chainedDatastore.py
+++ b/python/lsst/daf/butler/datastores/chainedDatastore.py
@@ -562,9 +562,6 @@ class ChainedDatastore(Datastore):
 
     def _prepIngest(self, *datasets: FileDataset, transfer: str | None = None) -> _IngestPrepData:
         # Docstring inherited from Datastore._prepIngest.
-        if transfer is None:
-            raise NotImplementedError("ChainedDatastore does not support transfer=None.")
-
         def isDatasetAcceptable(dataset: FileDataset, *, name: str, constraints: Constraints) -> bool:
             acceptable = [ref for ref in dataset.refs if constraints.isAcceptable(ref)]
             if not acceptable:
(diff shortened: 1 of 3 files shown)


tests/test_butler.py
```diff
@@ -2294,10 +2295,72 @@ def test_specialized_file_datasets_functions(self):
...
+                  - datastore:
+                      cls: lsst.daf.butler.datastores.fileDatastore.FileDatastore
+                      root: <butlerRoot>/FileDatastore_0
+                  - datastore:
+                      cls: lsst.daf.butler.datastores.fileDatastore.FileDatastore
+                      root: <butlerRoot>/FileDatastore_1
+            """
+            )
+            target_repo_config = Butler.makeRepo(tempdir, config)
+            print(target_repo_config)
```
timj: Do we need this print left in the test.

python/lsst/daf/butler/datastores/chainedDatastore.py
```diff
@@ -562,9 +562,6 @@ def _overrideTransferMode(self, *datasets: Any, transfer: str | None = None) ->
 
     def _prepIngest(self, *datasets: FileDataset, transfer: str | None = None) -> _IngestPrepData:
         # Docstring inherited from Datastore._prepIngest.
-        if transfer is None:
```
timj, line 565: This check was probably added as a safeguard early on but then later other code was added.

timj (approved):

PR #1252, opened 2025-09-12
DM-52412: Set the ARROW_DEFAULT_MEMORY_POOL env var to jemalloc on eups set up

timj:
There is a memory leak in pyarrow 20 that occurs in butler usage in some scenarios. The current fix is to switch pyarrow to use the old memory allocator.

The pyarrow issue is at https://github.com/apache/arrow/issues/45882

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-52412.misc.rst
+++ b/doc/changes/DM-52412.misc.rst
@@ -1,0 +1,2 @@
+For EUPS users we now set the ``ARROW_DEFAULT_MEMORY_POOL`` environment variable to ``jemalloc``.
+This works around a memory leak reported in https://github.com/apache/arrow/issues/45882 for pyarrow v18 and newer.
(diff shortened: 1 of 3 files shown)


doc/changes/DM-52412.misc.rst
```diff
@@ -0,0 +1,2 @@
+For EUPS users we now set the ``ARROW_DEFAULT_MEMORY_POOL`` environment variable to ``jemalloc``.
+This works around a memory leak reported in https://github.com/apache/arrow/issues/45882 for pyarrow v20 and newer.
```
erykoff: pyarrow 18 and newer.

erykoff (approved):
This works for eups setup, but what happens for pypi installs?

timj:
> but what happens for pypi installs

We are pretty much the only users of EUPS in this context so setting that variable is something we can clearly get away with. I don't know if we want to try setting that env var in butler `__init__.py` itself so that PyPI users switch to the old malloc without them realizing.

erykoff:
Duh, that makes a lot of sense.  Thanks!

PR #1253, opened 2025-09-13
DM-52515: Add glob-style matching of column names in parquet parameters.

erykoff:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/delegates/arrowtable.py
+++ b/python/lsst/daf/butler/delegates/arrowtable.py
@@ -32,6 +32,7 @@ from __future__ import annotations
 __all__ = ["ArrowTableDelegate"]
 
 from collections.abc import Mapping
+from fnmatch import fnmatchcase
 from typing import TYPE_CHECKING, Any
 
 import pyarrow as pa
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/delegates/arrowtable.py
```diff
@@ -162,28 +163,33 @@ def handleParameters(self, inMemoryDataset: Any, parameters: Mapping[str, Any] |
...
+                for column in readColumnsIn:
                     if not isinstance(column, str):
                         raise NotImplementedError(
                             f"InMemoryDataset of a {get_full_type_name(inMemoryDataset)} only "
                             "supports string column names."
                         )
-                    if column not in allColumns:
+                    found = False
+                    for allColumn in allColumns:
+                        if fnmatchcase(allColumn, column):
```
fred3m, line 186: I wonder if it would be better to use the `re` module to match using regular expressions instead (here and in the parquet formatter). I could go either way but I know that there are some people who are quite good at regular expressions (myself _not_ included) and they might appreciate being able to use them.

fred3m (approved):
One suggestion, otherwise LGTM.

python/lsst/daf/butler/delegates/arrowtable.py
erykoff: We can either do glob-style matching or regex style matching, but we can't have both.  I chose glob-style because I think that's simpler, but it is less powerful.

fred3m: Yeah, based on Tim's comments on slack I think that's the right call.

PR #1254, opened 2025-09-23
DM-52606: Log more information for Butler server file access

dhirving:
To allow us to understand the usage patterns for data releases, log information about files accessed through the Butler server.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/remote_butler/server/handlers/_external.py
+++ b/python/lsst/daf/butler/remote_butler/server/handlers/_external.py
@@ -30,10 +30,11 @@ from __future__ import annotations
 __all__ = ()
 
 import uuid
-from typing import Any
+from typing import Annotated, Any
 
 from fastapi import APIRouter, Depends, HTTPException, Request
 from fastapi.responses import RedirectResponse
+from structlog.stdlib import BoundLogger
 
 from ...._butler import Butler
 from ...._collection_type import CollectionType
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/remote_butler/server/handlers/_external.py
```diff
@@ -220,12 +225,21 @@ def _get_file_info_response(butler: Butler, ref: DatasetRef) -> GetFileResponseM
     )
 
 
+def _log_file_access(logger: BoundLogger, ref: DatasetRef, access_method: str) -> None:
+    logger.info(
+        "file-access",
+        dataset={"id": str(ref.id), "type": ref.datasetType.name},
```
timj: If we are analyzing usage patterns, would it be helpful to include the dataId in the log message to see if everyone is using the same tract? Or is the assumption we will export the log messages and do an analysis of that type by querying the butler for this information?

timj (approved):
This looks useful.

python/lsst/daf/butler/remote_butler/server/handlers/_external.py
dhirving: Adding the data ID is a good idea, I'll throw that in there.  Will let us avoid some annoying scripting to get some useful insights when the prompt data products are first released.

PR #1257, opened 2025-09-25
DM-52653: Add QBB method for retrieving predicted datastore records

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_quantum_backed.py
+++ b/python/lsst/daf/butler/_quantum_backed.py
@@ -614,6 +614,34 @@ class QuantumBackedButler(LimitedButler):
             datastore_records=provenance_records,
         )
 
+    def export_predicted_datastore_records(
+        self, refs: Iterable[DatasetRef]
+    ) -> dict[str, DatastoreRecordData]:
+        """Export datastore records for a set of predicted output dataset
+        references.
+
+        Parameters
+        ----------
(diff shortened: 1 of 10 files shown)


python/lsst/daf/butler/_quantum_backed.py
```diff
@@ -614,6 +614,33 @@ def extract_provenance_data(self) -> QuantumProvenanceData:
...
+        -------
+        records : `dict` [ `str`, `DatastoreRecordData` ]
+            Predicted datastore records indexed by datastore name. No attempt
+            is made to ensure that the associated datasets exist on disk.
+        """
+        unknowns = [
+            ref
+            for ref in refs
+            if (ref.id not in self._predicted_outputs and ref.id not in self._predicted_inputs)
+        ]
```
TallJimbo, line 639: This check would be slightly inconvenient for me, because I'm initializing the QBB well before I read any particular `DatasetRef` from the graph, and so far I've gotten away with not telling the QBB about any of the expected datasets.

It turns out we're doing the same sort of "is this one of the outputs I'm expecteding to `put`" logic both in QBB and in `QuantumContext`, so I've been thinking we might want to just drop the version in QBB more broadly.

TallJimbo (approved):
Yes, this would be perfect.

TallJimbo (approved):

PR #1258, opened 2025-09-26
DM-52398: Implement findDataset using the new query system 

dhirving:
Replace the implementation of `Registry.findDataset()` with the new query system, to allow us to retire daf_relation.  This attempts to preserve the existing behavior, with one exception: returned DatasetRefs now have `dataId.hasFull() = True`, in line with the behavior of the new query system.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_registry_shim.py
+++ b/python/lsst/daf/butler/_registry_shim.py
@@ -36,6 +36,7 @@ from typing import TYPE_CHECKING, Any
 from ._collection_type import CollectionType
 from ._dataset_ref import DatasetId, DatasetIdGenEnum, DatasetRef
 from ._dataset_type import DatasetType
+from ._exceptions import CalibrationLookupError
 from ._storage_class import StorageClassFactory
 from ._timespan import Timespan
 from .dimensions import (
@@ -48,7 +49,9 @@ from .dimensions import (
 )
 from .registry._collection_summary import CollectionSummary
 from .registry._defaults import RegistryDefaults
+from .registry._exceptions import NoDefaultCollectionError
 from .registry._registry_base import RegistryBase
+from .registry.queries._query_common import resolve_collections
 
 if TYPE_CHECKING:
     from .direct_butler import DirectButler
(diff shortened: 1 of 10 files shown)


python/lsst/daf/butler/registry/tests/_registry.py
```diff
@@ -634,6 +634,41 @@ def testFindDataset(self):
...
+                "int",
+                universe=butler.dimensions,
+            )
+        )
+        data_id = {"instrument": "Cam1", "physical_filter": "Cam1-G"}
+        (implied_ref,) = registry.insertDatasets("dt_with_implied", dataIds=[data_id], run=run)
+        self.assertEqual(implied_ref, registry.findDataset("dt_with_implied", data_id, collections=[run]))
+        # The search ignores excess data ID values beyond the 'required' set.
+        # This is not the correct band value for this physical_filter, but
+        # the mismatch is ignored.
```
TallJimbo, line 671: I understand not wanting to do it on this ticket, but I think this behavior is bad enough it'd be okay to turn it into an exception.

python/lsst/daf/butler/_registry_shim.py
```diff
@@ -182,13 +185,79 @@ def findDataset(
...
+            if (
+                datasetType.isCalibration()
+                and timespan is not None
+                and (timespan.begin is not None or timespan.end is not None)
+            ):
+                timespan_column = query.expression_factory[dataset_type_name].timespan
+                # The 'logical_or(timespan.is_null)' allows non-calibration
+                # collections to participate in the search, with the assumption
+                # that they are valid for any time range.
+                result = result.where(timespan_column.overlaps(timespan).logical_or(timespan_column.is_null))
```
TallJimbo, line 240: This might be something we should move deeper into the new query system, since treating non-CALIBRATION collections as temporal with infinite validity ranges is something I think want generally for consistency.  Not for this ticket, though; I'm not sure exactly what else would even need to change, and it could see it being hard to reconcile that model with all of the things we _can't_ do with CALIBRATION collections.

python/lsst/daf/butler/_registry_shim.py
```diff
@@ -182,13 +185,79 @@ def findDataset(
...
+            datasets = list(result)
+            if len(datasets) == 1:
+                ref = datasets[0]
+                if dataId.hasRecords():
+                    ref = ref.expanded(dataId)
+                # Propagate storage class from user-provided DatasetType, which
+                # may not match the definition in the database.
+                ref = ref.overrideStorageClass(datasetType.storageClass_name)
+                if datastore_records:
+                    ref = self._registry.get_datastore_records(ref)
```
TallJimbo, line 251: We should really go and rip out all of the datastore-records-in-refs stuff at some point; it was a complete dead-end.

TallJimbo (approved):
Looks great; only comments are musing about scope for future tickets.

PR #1260, opened 2025-09-30
DM-52399: Remove daf_relation from registry.certify() and decertify()

dhirving:
Modify `registry.certify()` / `registry.decertify()` to use the new query system instead of daf_relation, because we plan to retire daf_relation.

## Checklist

- [x] ran Jenkins


TallJimbo (approved):
Looks great; love to see all those deletes!

PR #1261, opened 2025-10-01
DM-52719: Use UUIDv7 for dataset IDs

dhirving:
Generate v7 UUIDs for dataset IDs instead of v4 UUIDs.  v4 UUIDs create problems for Postgres insert performance, because they  scatter index updates randomly around the disk.  v7 UUIDs have a timestamp in the most significant bits, so datasets generated at the same time will be close together in the index.  Initial testing shows up to 100x improvement in insert speed with v7 UUIDs.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-52719.perf.md
+++ b/doc/changes/DM-52719.perf.md
@@ -1,0 +1,1 @@
+New dataset UUIDs are now generated using UUIDv7 instead of UUIDv4, to improve Postgres insert performance.
--- a/python/lsst/daf/butler/_dataset_ref.py
+++ b/python/lsst/daf/butler/_dataset_ref.py
@@ -64,6 +64,7 @@ from lsst.utils.classes import immutable
 from ._config_support import LookupKey
 from ._dataset_type import DatasetType, SerializedDatasetType
 from ._named import NamedKeyDict
+from ._uuid import generate_uuidv7
 from .datastore.stored_file_info import StoredDatastoreItemInfo
 from .dimensions import (
     DataCoordinate,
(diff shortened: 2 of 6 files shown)


TallJimbo:
Looks great, works great.

Note that we actually generate most of our UUIDs in practice in `pipe_base`, since that happens before we ever make a `DatasetRef` from them.  I've got some commits on pipe_base for utilizing this code there, and if you don't mind I might just make a pipe_base branch for this ticket.  It's not just a simple swap since I actually just merged some code that makes  UUID4 assumptions to speed up some lookups, but I've already got a fix.

In light of that, should we move this to a more public module in daf_butler?

dhirving:
> Looks great, works great.
> 
> Note that we actually generate most of our UUIDs in practice in `pipe_base`, since that happens before we ever make a `DatasetRef` from them. I've got some commits on pipe_base for utilizing this code there, and if you don't mind I might just make a pipe_base branch for this ticket. It's not just a simple swap since I actually just merged some code that makes UUID4 assumptions to speed up some lookups, but I've already got a fix.
> 
> In light of that, should we move this to a more public module in daf_butler?

Sure, I'll move it to the `_rubin` "internal public" package.

timj:
Are we intending to prefer the python 3.14 implementation if available? I assume the UUIDs in pipe_base are the quanta UUIDs and not the DatasetRef ones.

dhirving:
> Are we intending to prefer the python 3.14 implementation if available? I assume the UUIDs in pipe_base are the quanta UUIDs and not the DatasetRef ones.

I think we should just have one implementation -- why introduce variable behavior when we don't have to.  (The standard library is currently using a more complex variant that provides monotonicity in the generated IDs that we don't need.)   When we no longer have to support 3.13 we can get rid of this code and switch to the standard library version.

TallJimbo:
`pipe_base` is generating both dataset and quantum UUIDs in the QG builder; its internal data structures don't use `DatasetRef` until well after it needs the IDs.

PR #1264, opened 2025-10-07
DM-52360: improvements to support the provenance aggregator

TallJimbo:
The only commits actually needed by the current incarnation of the aggregator are the "Add optimized registry import" and "Add method to fetch all dataset IDs" ones; the rest are small cleanups that got in the way of earlier versions.

The "Don't existence check predicted URIs" commit is a behavior change that @timj discussed; while I don't need it for the aggregator anymore, I do think the new behavior is less surprising.  But I don't care much and would be happy to roll it back if anyone disagrees.

## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -899,6 +912,47 @@ def _validate_import(
...
+        static_rows = [
+            {
+                "id": ref.id,
+                "dataset_type_id": dataset_type_storage[ref.datasetType.name].dataset_type_id,
+                self._run_key_column: run.key,
+                "ingest_date": timestamp,
+            }
+            for ref in refs
+        ]
+        tags_rows = [
```
dhirving: You could consider pulling out a helper functions to de-duplicate the generation of `tag_rows` here, which is identical to `tmpRows` in the guarded version.

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -899,6 +912,47 @@ def _validate_import(
...
+        refs: list[DatasetRef],
+        dataset_type_storage: dict[str, _DatasetRecordStorage],
+        tags_table: sqlalchemy.Table,
+    ) -> None:
+        collection_fkey_name = self._collections.getCollectionForeignKeyName()
+        # Current timestamp, type depends on schema version. Use microsecond
+        # precision for astropy time to keep things consistent with
+        # TIMESTAMP(6) SQL type.
+        timestamp: datetime.datetime | astropy.time.Time
+        if self._use_astropy_ingest_date:
```
dhirving: This ingest_date logic could be shared between the two code paths if a shared function returned  `sqlalchemy.sql.literal` and you call `.value` on it for the case where you don't need the SQLAlchemy definition.

dhirving (approved):
Looks good to me.  A couple non-essential suggestions for some minor deduplication of code between the two import code paths.

PR #1268, opened 2025-10-15
DM-52895: Use unbounded timespan for non-calibration collections.

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/_registry_base.py
+++ b/python/lsst/daf/butler/registry/_registry_base.py
@@ -231,20 +231,28 @@ class RegistryBase(Registry):
         collectionTypes: Iterable[CollectionType] = CollectionType.all(),
         flattenChains: bool = False,
     ) -> Iterator[DatasetAssociation]:
-        # queryCollections only accepts DatasetType.
         if isinstance(datasetType, str):
             datasetType = self.getDatasetType(datasetType)
-        resolved_collections = self.queryCollections(
-            collections, datasetType=datasetType, collectionTypes=collectionTypes, flattenChains=flattenChains
-        )
         with self._butler.query() as query:
+            resolved_collections = self.queryCollections(
+                collections,
+                datasetType=datasetType,
+                collectionTypes=collectionTypes,
+                flattenChains=flattenChains,
+            )
+            # It's annoyingly difficult to just do the collection query once,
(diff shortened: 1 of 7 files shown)


python/lsst/daf/butler/registry/_registry_base.py
```diff
@@ -231,20 +231,25 @@ def queryDatasetAssociations(
         collectionTypes: Iterable[CollectionType] = CollectionType.all(),
         flattenChains: bool = False,
     ) -> Iterator[DatasetAssociation]:
-        # queryCollections only accepts DatasetType.
         if isinstance(datasetType, str):
             datasetType = self.getDatasetType(datasetType)
         resolved_collections = self.queryCollections(
             collections, datasetType=datasetType, collectionTypes=collectionTypes, flattenChains=flattenChains
         )
+        # It's annoyingly difficult to just do the collection query once, since
```
dhirving: If you moved the two collection queries down inside the query() context it would cause them to be done only once -- query() implicitly turns on collection record caching.

python/lsst/daf/butler/transfers/_context.py
```diff
@@ -354,19 +354,24 @@ def _computeDatasetAssociations(self) -> dict[str, list[DatasetAssociation]]:
...
+                query = query.join_dataset_search(datasetType, resolved_collections.keys())
                 result = query.general(
                     datasetType.dimensions,
                     dataset_fields={datasetType.name: {"dataset_id", "run", "collection", "timespan"}},
                     find_first=False,
                 )
-                for association in DatasetAssociation.from_query_result(result, datasetType):
+                for association in DatasetAssociation.from_query_result(
+                    result, datasetType, resolved_collections
+                ):
```
dhirving: I'm pretty sure this chunk of code is identical to `self._butler.registry.queryDatasetAssociations(datasetType, self._collections.key(), collectionTypes=collectionTypes, flattenChains=False)`.

dhirving (approved):
Looks good to me.

I think the `timespan.is_null` hack in findDataset can also be removed now that the underlying query system has the same behavior.

https://github.com/lsst/daf_butler/blob/9fa57067c8fa69791258539a4008329d01b61689/python/lsst/daf/butler/_registry_shim.py#L240

PR #1271, opened 2025-10-17
DM-52952: Add missing ParentDatasetQueryResults.dataIds query shim

dhirving:
Added an implementation of ParentDatasetQueryResults.dataIds using the new query system.  This method did not previously have any test coverage, so I failed to notice that it did not have an implementation shim to the new query system in DM-52397.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/queries/_query_common.py
+++ b/python/lsst/daf/butler/registry/queries/_query_common.py
@@ -58,6 +58,9 @@ class CommonQueryArguments:
     def replaceCollections(self, collections: list[str]) -> CommonQueryArguments:
         return dataclasses.replace(self, collections=collections)
 
+    def replaceDatasetTypes(self, dataset_types: list[str]) -> CommonQueryArguments:
+        return dataclasses.replace(self, dataset_types=dataset_types)
+
 
 _T = TypeVar("_T", bound=QueryResultsBase)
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/registry/queries/_query_common.py
```diff
@@ -58,6 +58,9 @@ class CommonQueryArguments:
     def replaceCollections(self, collections: list[str]) -> CommonQueryArguments:
         return dataclasses.replace(self, collections=collections)
 
+    def replaceDatasetTypes(self, dataset_types: list[str]) -> CommonQueryArguments:
+        return dataclasses.replace(self, dataset_types=dataset_types)
+
```
TallJimbo, line 63: I'm not sure these helpers actually add value; I think I'd prefer to just see the `dataclasses.replace` call and know exactly what's going on from that.

TallJimbo (approved):

python/lsst/daf/butler/registry/queries/_query_common.py
dhirving: `dataclasses.replace` isn't type-safe, and it might not be obvious to someone investigating the usage of `CommonQueryArguments` that there is code in another module manipulating its data in a way that can't be found using right-click -> find all references.

PR #1273, opened 2025-10-23
Multiplatform build

athornton:
## Checklist

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -34,14 +34,23 @@ on:
 
 jobs:
   docker:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform:
+          - arch: "amd64"
+            runs_on: "ubuntu-latest"
+          - arch: "arm64"
+            runs_on: "ubuntu-24.04-arm"
+    runs-on: ${{ matrix.platform.runs_on }}
     timeout-minutes: 30
     permissions:
       contents: read
       packages: write
+    outputs:
+      tag: ${{ steps.build.outputs.tag}}
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           # Needed to fetch tags, used by Python install process to
           # figure out version number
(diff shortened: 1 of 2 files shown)


.github/workflows/docker.yaml
```diff
@@ -50,20 +59,88 @@ jobs:
...
+    env:
+      tag: ${{ needs.docker.outputs.tag }}
+
+    steps:
+      - name: Install system dependency to build psycopg on ARM
+        if: ${{ matrix.platform.arch }} == 'arm64'
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libpq-dev
+
```
dhirving: What is this postgres setup needed for?

dhirving (commented):
I think there might be a [simpler way to solve this problem](https://github.com/lsst/daf_butler/pull/1274)

dhirving:
> I think there might be a [simpler way to solve this problem](https://github.com/lsst/daf_butler/pull/1274)

Maybe spoke too soon.  Looks like the build works just fine, but the tests run rather slow (I think it's using qemu to emulate the arm64 half.) So maybe use the `platforms` parameter to the build to get rid of the 'unify' step and avoid the random 3rd party dependency, but then matrix the test portion as you have here?

.github/workflows/docker.yaml
```diff
@@ -50,20 +59,88 @@ jobs:
...
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build multi-platform manifest
+        uses: Noelware/docker-manifest-action@v1
+        with:
+          inputs: ghcr.io/${{ github.repository }}-amd64:${{ env.tag }},ghcr.io/${{ github.repository }}-arm64:${{ env.tag }}
```
dhirving, line 86: Sorry, I didn't actually share my concerns about this `unify` step because I thought the `platforms` parameter would just completely replace all of these changes.

Because we're using `env.tag` instead of the image digest, I think there's a race condition here that can cause this to pick up an incorrect build from a different CI run, if there are multiple pushes to the same PR in a short amount of time.  Could be very confusing when trying to test/debug an image.

Also I don't think it's necessary to use this `Noelware` 3rd party GHA -- the official docker GHA tools can be used to [do the same thing](https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners)

athornton:
So, my thesis here is that I need an arm64 build of the container, because I am trying to build an arm64-only RSP cluster.

Yes, you can do this entirely with `--platform` but as you point out you are then emulating arm64, which carries a 3x-4x performance penalty (plus being interleaved, rather than run in parallel, with the amd64 build), and the test suite takes a long time to run even natively.

Yes, you can also do the multiplatform stuff directly with `docker manifest` but why on earth would you when there's already a perfectly cromulent action that does it for you?

You could also just run the test suite on amd64 by *not* matrixing that step.  I think this would be unwise, because, for instance, psycopg2 needs to build on arm64 rather than coming prebuilt as it does on amd64, so I think that just assuming they behave identically isn't the safest assumption.

I chose to use the tag rather than the digest output because I don't think there is a use case for an image digest that is no longer associated with a tag.  If your use case actually refers to digests directly, I can change it back, but everything SQuaRE does, at least, is tag-driven (the actual digest is known to and used by the nublado-controller, but it frequently refreshes itself and what is shown to the user for selection is the tag, and the controller uses the digest it found for that tag in its most recent reconciliation run).

But my *real* point here is, I need an arm64 build of the daf_butler container that I can pull.  I don't care how you get it to me, but I need it to exist.  Implement this however you like.  I created an implementation that works and that (I claim) has certain obvious advantages over the more obvious multiplatform build.  But if you want to implement it that way, I don't have a dog in that fight.  I do, however, need an arm64 build of the container that lives at some remotely-accessible registry in the next couple weeks.

PR #1274, opened 2025-10-24
Add build of ARM64 Docker container

dhirving:
## Checklist

- [ ] ran Jenkins


athornton:
This will work too, and I don't care whether you do it this way.  It's a lot slower to build and test than matrixing those steps because you're emulating arm64 and interleaving the steps rather than running two native builds in parallel, but as long as I have a working container I can download from a registry at the end of the day, I will be happy.

PR #1275, opened 2025-10-25
DM-53019: make ButlerLogRecords extensible

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_logFormatter.py
+++ b/tests/test_logFormatter.py
@@ -30,6 +30,7 @@
 import json
 import logging
 import os
+import sys
 import tempfile
 import unittest
 from logging import FileHandler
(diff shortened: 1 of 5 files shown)


tests/test_logFormatter.py
```diff
@@ -73,31 +83,127 @@ def testButlerLogRecordsFormatter(self):
...
+            extra_data = {"extra1": 1, "extra2": 2}
+            with open(tmp.name, "a") as stream:
+                ButlerLogRecords.write_streaming_extra(stream, json.dumps(extra_data))
+
+            # Now ingest the file.
+            ref = DatasetRef(self.datasetType, dataId={}, run=self.run)
+            dataset = FileDataset(path=tmp.name, refs=ref)
+            self.butler.ingest(dataset, transfer="move")
+
+            records: ButlerLogRecords = self.butler.get(ref)
```
MichelleGower, line 203: Did you mean to add the type annotation in a test?  Not seeing it in other tests and won't the function be skipped for type checking because it doesn't have type annotations?

MichelleGower (approved):
A type annotation question.  Merge approved.

tests/test_logFormatter.py
TallJimbo: Yes, I added it just because it made my editor more useful while working on the code, and I figured that might be true for others so it was worth leaving.

PR #1276, opened 2025-10-27
DM-53040: Convert ScarletModelData to LsstScarletModelData

fred3m:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -414,6 +414,8 @@ storageClasses:
     delegate: lsst.meas.extensions.scarlet.io.ScarletModelDelegate
   LsstScarletModelData:
     pytype: lsst.meas.extensions.scarlet.io.LsstScarletModelData
+    converters:
+      lsst.scarlet.lite.io.model_data.ScarletModelData: lsst.meas.extensions.scarlet.io.utils.scarlet_model_to_lsst_scarlet_model
     parameters:
       - blend_id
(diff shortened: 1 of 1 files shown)


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -414,6 +414,8 @@ storageClasses:
     delegate: lsst.meas.extensions.scarlet.io.ScarletModelDelegate
   LsstScarletModelData:
     pytype: lsst.meas.extensions.scarlet.io.LsstScarletModelData
+    converters:
+      lsst.scarlet.lite.io.model_data.ScarletModelData: lsst.meas.extensions.scarlet.io.utils.scarlet_model_to_lsst_scarlet_model
```
fred3m, line 418: Note to reviewer: check that`lsst.scarlet.lite.io.model_data.ScarletModelData` is compatible with `lsst.scarlet.lite.io.ScarletModelData`. The reason is that DM-52816 moved the io module. It will still import properly from `lsst.scarlet.lite.io.ScarletModelData` and I'm not sure if the sub modules matter for Butler lookup.

TallJimbo: Should be completely fine, especially with the import shim.  Butler stores the storage class name and the full formatter module path in the database, but not the Python type; that's always looked up from the storage class name on `get`, so as long as the type in `storageClasses.yaml` is usable, you're fine.

TallJimbo (approved):

PR #1280, opened 2025-10-28
DM-51660: Rename photo-z datasets

taranu:
## Checklist

- [x] ran Jenkins
- [x] (if changing dimensions.yaml) make a copy of dimensions.yaml in `configs/old_dimensions`

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -433,6 +433,8 @@ storageClasses:
     pytype: qp.Ensemble
   PZModel:
     pytype: rail.core.model.Model
+  PhotozModel:
+    pytype: rail.core.model.Model
   VisitBackgroundModel:
     pytype: lsst.drp.tasks.fit_visit_background.VisitBackgroundModel
   VignettingCorrection:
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -429,7 +429,7 @@ storageClasses:
     pytype: lsst.pipe.base.utils.RegionTimeInfo
   QPEnsemble:
     pytype: qp.Ensemble
-  PZModel:
```
timj, line 434: I am assuming that PZModel has never been used for anything in a real butler?

taranu: No, there are PZModel datasets in most of the butlers by now, but all in user runs (most if not all `u/echarles/*`). These are the ones you were unhappy to learn are pickles.

timj: If you delete the storage class definition all those datasets are going to break things. You can't delete a definition of a storage class once you've used it somewhere because the dataset type is defined with that storage class. You need to make a new entry but leave the old behind.

timj: (you can change the definition of the storage class but the label needs to still exist)

arunkannawadi (approved):

PR #1285, opened 2025-11-05
DM-53176: Add get_many_datasets()

dhirving:
Added a method `Butler.get_many_datasets()` to look up multiple `DatasetRef` instances from their UUIDs.  This will be used by the [DMTN-330](https://dmtn-330.lsst.io/) Prompt Publication Service when transferring groups of datasets between repositories.  

Also added support for passing dataset UUIDs as strings into the existing `get_dataset()` method, instead of requiring them to be manually converted to UUID objects.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
+++ b/python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
@@ -311,7 +311,7 @@ class ByDimensionsDatasetRecordStorageManagerUUID(DatasetRecordStorageManager):
     def register_dataset_type(self, dataset_type: DatasetType) -> bool:
         # Docstring inherited from DatasetRecordStorageManager.
         #
-        # This is one of three places where we populate the dataset type cache.
+        # This is one of the places where we populate the dataset type cache.
         # See the comment in _fetch_dataset_types for how these are related and
         # invariants they must maintain.
         #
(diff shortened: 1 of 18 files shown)


python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -422,50 +423,63 @@ def resolve_wildcard(
...
+        id_col = self._static.dataset.columns["id"]
+        run_key_column = self._static.dataset.columns[self._collections.getRunForeignKeyName()]
         sql = sqlalchemy.sql.select(
-            self._static.dataset.columns.dataset_type_id,
-            self._static.dataset.columns[self._collections.getRunForeignKeyName()],
-        ).where(self._static.dataset.columns.id == id)
+            id_col,
+            self._static.dataset.columns["dataset_type_id"],
+            run_key_column,
+        ).where(id_col.in_(ids))
```
TallJimbo, line 432: If you add `.subquery()` to the end of this, you'll get an equivalent `FromClause` to pass to `lookup_name_sql`, and you can drop the `Joinable` type alias and all of the annotation changes that use it (which otherwise would need some docstring updates to go along with them).

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
```diff
@@ -422,50 +423,63 @@ def resolve_wildcard(
...
+            assert run_name_column.key is not None
+            for data_id_row in data_id_rows:
+                id = data_id_row["dataset_id"]
+                dataset_row = dataset_rows[id]
+                # Tags table includes run collections and tagged
+                # collections.
+                # In theory the data ID for a given dataset should be the
+                # same in both, but nothing actually guarantees this.
+                # So skip any tagged collections, using the run collection
+                # as the definitive definition.
```
TallJimbo: Seems like it might be better to join to the collections table and require the collection type there to be RUN, especially if someday we have a lot of TAGGED collections (as I think we're generally planning to use them more in the future than we are now).

TallJimbo (approved):

python/lsst/daf/butler/registry/datasets/byDimensions/_manager.py
TallJimbo: Did you miss this comment?

dhirving: I adjusted the docstrings.  I don't think it's worth subjecting ourselves to the vagaries of Postgres' subquery optimization logic to avoid having a type definition.

TallJimbo: I didn't think there would be any difference in the SQL that would be generated; I was thinking it was a question of whether you add the `subquery()` or SQLAlchemy does it internally.  But maybe I'm wrong about that.

dhirving: According to the `SelectBase.subquery()` docs it renders as a parenthesized subquery:

```
        The subquery form by itself renders the same way, however when
        embedded into the FROM clause of another SELECT statement, it becomes
        a named sub-element::

            subq = stmt.subquery()
            new_stmt = select(subq)

        The above renders as:

        .. sourcecode:: sql

            SELECT anon_1.id, anon_1.name
            FROM (SELECT table.id, table.name FROM table) AS anon_1
```

Whether that's true in all cases I'm not sure, but during development I did confirm that without the `subquery()` it generates a totally normal `SELECT ... FROM ... INNER JOIN ...` query.

PR #1288, opened 2025-11-13
DM-53233: Add resource cleanup interface to Butler

dhirving:
Added `Butler.close()`, a context manager implementation for `Butler`, and `QuantumBackedButler.close()`.  These can be used to ensure that database connections are closed deterministically, rather than waiting for mark-and-sweep garbage collection.

The Butler unit tests and CLI scripts were updated to use these, ensuring all file handles are closed promptly.  This fixes an issue where the unit tests could fail due to file descriptor exhaustion.

Also removed circular references from `DirectButler`, allowing it to be garbage collected immediately when it goes out of scope.  However, the underlying `SqlAlchemy.Engine` has internal circular references, so this does not release the database connections until mark-and-sweep garbage collection is run or `Butler.close()` is called.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_butler.py
+++ b/tests/test_butler.py
@@ -711,8 +711,10 @@ class ButlerTests(ButlerPutGetTests):
                     uri = Butler.get_repo_uri("label")
                     butler = Butler.from_config(uri, writeable=False)
                     self.assertIsInstance(butler, Butler)
+                    butler.close()
                     butler = Butler.from_config("label", writeable=False)
                     self.assertIsInstance(butler, Butler)
+                    butler.close()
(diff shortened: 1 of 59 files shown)


tests/test_butler.py
```diff
@@ -698,8 +711,10 @@ def testConstructor(self) -> None:
                     uri = Butler.get_repo_uri("label")
                     butler = Butler.from_config(uri, writeable=False)
                     self.assertIsInstance(butler, Butler)
+                    butler.close()
```
timj, line 714: The logic here being that if it fails the assert above then the `close` here won't work anyhow.

timj (approved):
Impressive amount of fiddly work to get these all fixed. Thank you.

PR #1289, opened 2025-11-18
DM-53326: Add provenance cap for inputs

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-53326.misc.rst
+++ b/doc/changes/DM-53326.misc.rst
@@ -1,0 +1,2 @@
+Added new options to ``DatasetProvenance`` to prevent excessive numbers of input datasets being recorded in file metadata.
+The limit for parquet datasets was set to 2,000 inputs.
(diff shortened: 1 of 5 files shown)


erykoff (approved):
This needs a changelog fragment!

And are you sure you don't want a cap on (e.g.) fits files as well?  It's not like a fits file with 400Mb of header provenance would be something people want.

timj:
I usually add the change log fragment once I know it's not all changing.

@TallJimbo you implied that FITS outputs won't really have the tens of thousands of inputs but I can patch obs_base if we think that is a risk. I have pondered whether we can auto switch from including the dataset type and run to not including them but having 3x the number of inputs -- easy to implement but it might be confusing.

TallJimbo:
Guarding against massive provenance counts in the FITS variant seems prudent even if we don't expect to hit those limits.

But I don't think we should try to support UUID-only provenance as an alternative just to move the cap a little higher.  The fallback can just be external (QG) provenance.

PR #1291, opened 2025-11-21
DM-53370: Move butler context manager up to LimitedButler

timj:
This allows all butlers to use it but with close() defaulting to a no-op.

## Checklist

- [x] ran Jenkins


timj:
The log formatter fix is here along with the LimitedButler change because there was a failing test discovered in ctrl_mpexec during the resource cleanup that was generating an empty log file that couldn't be read.

dhirving (approved):
Looks good to me

PR #1302, opened 2025-12-02
DM-53468: Open the file handle before calling parquet to open file

timj:
We think that in some cases pyarrow is opening the fsspec file internally and then not closing it.

## Checklist

- [x] ran Jenkins


erykoff (approved):
Honestly, if I realized that pyarrow was only doing `fs.open()` inside I may have gone in this directly in the first place.  I thought it was doing something fancier.

Thanks!

PR #1308, opened 2025-12-11
Added parquet support for filters parameter

gtorrini:
## Checklist

- [ ] ran Jenkins

## Notes
- Missing release note (no ticket associated yet).
- I didn't do any semantic validation on the filters, as this is performed by [filters_to_expression](https://github.com/apache/arrow/blob/main/python/pyarrow/parquet/core.py#L134C1-L196C53) early on in [ParquetDataset initialization](https://github.com/apache/arrow/blob/main/python/pyarrow/parquet/core.py#L1383C1-L1384C69). 
- I tried to get started on some basic unit test checks, but it seems there's a wrapper/`except` block preventing the `ValueError` for invalid columns in the filter predicates from being interpreted directly. See the #FIXME comments in `test_parquet.py`. The `get` call outside the `assertRaises` context manager DOES raise a `ValueError`, but the `get` call inside the `assertRaises` context manager raises an `AssertionError` for not raising a value error.
```
====================================== short test summary info =======================================
FAILED test_parquet.py::ParquetFormatterDataFrameTestCase::testMultiIndexDataFrame - ValueError: Failure from formatter 'lsst.daf.butler.formatters.parquet.ParquetFormatter' for data...
FAILED test_parquet.py::InMemoryDataFrameDelegateTestCase::testSingleIndexDataFrame - AssertionError: ValueError not raised
```

### DIFF
--- a/python/lsst/daf/butler/formatters/parquet.py
+++ b/python/lsst/daf/butler/formatters/parquet.py
@@ -131,6 +131,8 @@ class ParquetFormatter(FormatterV2):
         component: str | None = None,
         expected_size: int = -1,
     ) -> Any:
+        import pyarrow.dataset as ds
+
         with generic_open(path, fs) as handle:
             schema = pq.read_schema(handle)
 
@@ -156,6 +158,7 @@ class ParquetFormatter(FormatterV2):
             return len(temp_table[schema.names[0]])
 
         par_columns = None
+        par_filters = None
         if self.file_descriptor.parameters:
             par_columns = self.file_descriptor.parameters.pop("columns", None)
             if par_columns:
@@ -190,6 +193,27 @@ class ParquetFormatter(FormatterV2):
                         par_columns,
                     )
 
+            par_filters = self.file_descriptor.parameters.pop("filters", None)
+            if par_filters:
+                # Validate basic filter structure
+                if not isinstance(par_filters, ds.Expression):
+                    if not isinstance(par_filters, list):
+                        raise TypeError("Filters must be a list or a pyarrow.dataset.Expression.")
+
+                    # Lists must contain tuples or lists of tuples
+                    if all(isinstance(item, tuple) for item in par_filters):
+                        par_filters = [par_filters]  # convert to DNF for column checking
+                    elif not all(isinstance(item, list) for item in par_filters):
+                        raise TypeError("List filters must contain tuples or lists of tuples.")
+
+                    # Ensure requested columns are in schema
+                    for predicate in par_filters:
+                        for col, _, _ in predicate:
+                            if col not in schema.names:
+                                raise ValueError(
+                                    f"Column {col} specified in filters not available in parquet file."
+                                )
+
             if len(self.file_descriptor.parameters):
                 raise ValueError(
                     f"Unsupported parameters {self.file_descriptor.parameters} in ArrowTable read."
@@ -202,6 +226,7 @@ class ParquetFormatter(FormatterV2):
                 columns=par_columns,
                 use_threads=False,
                 use_pandas_metadata=(b"pandas" in metadata),
+                filters=par_filters,
             )
 
         return arrow_table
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -190,6 +193,27 @@ def _read_parquet(
...
+                    elif not all(isinstance(item, list) for item in par_filters):
+                        raise TypeError("List filters must contain tuples or lists of tuples.")
+
+                    # Ensure requested columns are in schema
+                    for predicate in par_filters:
+                        for col, _, _ in predicate:
+                            if col not in schema.names:
+                                raise ValueError(
+                                    f"Column {col} specified in filters not available in parquet file."
+                                )
```
erykoff, line 215: Is the indentation correct above?  This won't do the check if it is an Expression; is that intended?

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -190,6 +193,27 @@ def _read_parquet(
...
+                # Validate basic filter structure
+                if not isinstance(par_filters, ds.Expression):
+                    if not isinstance(par_filters, list):
+                        raise TypeError("Filters must be a list or a pyarrow.dataset.Expression.")
+
+                    # Lists must contain tuples or lists of tuples
+                    if all(isinstance(item, tuple) for item in par_filters):
+                        par_filters = [par_filters]  # convert to DNF for column checking
+                    elif not all(isinstance(item, list) for item in par_filters):
+                        raise TypeError("List filters must contain tuples or lists of tuples.")
```
erykoff, line 207: I don't understand this section at all.  If it's a list of tuples you convert a list to a list of lists?  And then if it's a list of lists it raises?

tests/test_parquet.py
```diff
@@ -408,6 +408,12 @@ def testSingleIndexDataFrame(self):
         # Passing an unrecognized column should be a ValueError.
         with self.assertRaises(ValueError):
             self.butler.get(self.datasetType, dataId={}, parameters={"columns": ["e"]})
+        # Filter predicates with unrecognized column should be a ValueError.
+        # FIXME: ValueErrorWeirdness1. This raises an AssertionError,
+        #        but Ctrl+f "ValueErrorWeirdness2" below.
+        # with self.assertRaises(ValueError):
+        #     self.butler.get(self.datasetType, dataId={},
+        #                     parameters={"filters": [("e", ">", 1)]})
```
erykoff, line 416: Tests of the functionality is missing.

erykoff (changes requested):

python/lsst/daf/butler/formatters/parquet.py
gtorrini: It is not, I should fix it to perform the check if it is an Expression as well.

python/lsst/daf/butler/formatters/parquet.py
gtorrini: I converted the list of tuples to a list of list of tuples for streamlined column checking later on (par_filters in DNF). It should only raise the TypeError if it is not a list of tuples or list of lists.

tests/test_parquet.py
gtorrini: Yes, this is addressed in the third bullet points of the notes above. I couldn't figure out why it raises a ValueError outside of the context manager, but calling the context manager to check for a ValueError raises an AssertionError.

gtorrini:
After revisiting Eli's feedback and doing some more research, I think implementing parquet filter support is probably above my pay grade:

- As far as I can tell, PyArrow does not perform any _dataset-aware_ validation on the filters before reading; it only semantically validates predicates. See [`ParquetDataset` initialization](https://github.com/apache/arrow/blob/main/python/pyarrow/parquet/core.py#L1382C1-L1384C69), [`filters_to_expression`](https://github.com/apache/arrow/blob/main/python/pyarrow/parquet/core.py#L134C1-L197C1), and [`ParquetDataset.read` method](https://github.com/apache/arrow/blob/main/python/pyarrow/parquet/core.py#L1552C1-L1555C10) for more info.
- In addition, I could not find a public API exposing field references in a [`pyarrow.dataset.Expression`](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.Expression.html), which prevents me from validating against the fields in the schema.

Passing filters without any field checks to `pq.read_table`--even if they are a properly formatted `ds.Expression`, List[Tuple], or List[List[Tuple]]--could result in some very late errors raised by PyArrow. I'm not sure what degree of validation is preferred at the daf_butler API boundary, so I'll defer to someone with more expertise.

PR #1309, opened 2025-12-12
DM-53610: Fix FileDatastore.export_records for large inputs

dhirving:
Fix an issue where `FileDatastore.export_records` would fail when given a large number of input datasets due to generating a >1GB SQL query.  Also fixed a similar potential issue in `Butler.get_many_datasets()`, although it has not occurred in the wild yet.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-53610.bugfix.md
+++ b/doc/changes/DM-53610.bugfix.md
@@ -1,0 +1,2 @@
+`FileDatastore.export_records()` will no longer fail with `sqlalchemy.exc.InternalError` when given a large number of dataset IDs.
+`Butler.get_many_datasets()` will no longer fail with `sqlalchemy.exc.InternalError` when given a large number of dataset IDs.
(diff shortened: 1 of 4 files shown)


dhirving:
It turns out that Postgres supports a `column = ANY(array)` syntax, which is roughly the same but would only generate one bind parameter instead of a parameter per value.   I opened [DM-53616](https://rubinobs.atlassian.net/browse/DM-53616) if we want to mess with that at some point.

[DM-53616]: https://rubinobs.atlassian.net/browse/DM-53616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

TallJimbo (approved):
Looks great.  Thanks for the comment on the `bridge.check` behavior - I certainly didn't those details.

dhirving:
> Looks great. Thanks for the comment on the `bridge.check` behavior - I certainly didn't those details.

Yeah, I had been hoping to just remove `bridge.check()` entirely because it looked unnecessary, but on further investigation there were those weird details.

PR #1311, opened 2025-12-17
DM-53648: Move ingest-from-temporary logic to a utility class.

TallJimbo:
This also changes the name pattern for temporaries to include the full final filename, in order to make it possible to find orphaned temporaries given a predicted filename (at least when the filename is fully unique on its own).

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_rubin/temporary_for_ingest.py
+++ b/python/lsst/daf/butler/_rubin/temporary_for_ingest.py
@@ -50,6 +50,12 @@ if TYPE_CHECKING:
 class TemporaryForIngest:
     """A context manager for generating temporary paths that will be ingested
     as butler datasets.
+
+    Notes
+    -----
+    Neither this class nor its `make_path` method run ingest automatically when
+    their context manager is exited; the `ingest` method must always be called
+    explicitly.
     """
 
     butler: Butler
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/_rubin/temporary_for_ingest.py
```diff
@@ -0,0 +1,209 @@
...
+
+    ingest_on_success: bool = True
+    """Whether to actually ingest the file when the context manager exits
+    without an exception.
+    """
+
+    ingest_on_failure: bool = False
+    """Whether to actually ingest the file when the context manager exits due
+    to an exception being raised.
+    """
```
dhirving: I wonder if it would be simpler to provide an `ingest()` method for the user to call instead of having them twiddle these flags.  Would make it explicit to readers of the calling code when the ingest is occurring, anyway.

python/lsst/daf/butler/_rubin/temporary_for_ingest.py
```diff
@@ -0,0 +1,209 @@
...
+        by a hard error that prevents any cleanup code from running (e.g.
+        sudden loss of power).
+        """
+        from .._dataset_ref import DatasetRef
+
+        if butler is not None:
+            assert isinstance(target, DatasetRef), "Overload accepts a DatasetRef and a Butler."
+            target = butler.getURI(target, predict=True).replace(fragment="")
+        else:
+            assert isinstance(target, ResourcePath), "Overload accepts a ResourcePath and no Butler."
```
dhirving: Python would take care of enforcing the right number of arguments for you if these were two separate methods instead of an overload.

dhirving (approved):
Looks good to me, just a couple non-essential suggestions.

python/lsst/daf/butler/_rubin/temporary_for_ingest.py
TallJimbo: In some cases that probably would be better, but I think what I've got works better for the one use case we've got (in `LogCapture` in `pipe_base`).  I'll leave this one for now.

TallJimbo: Comments on the pipe_base PR changed my mind on this again; I'll do the `ingest` method.

PR #1314, opened 2025-12-19
DM-53622: changes in support of provenance in SeparablePipelineExecutor

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -443,3 +443,18 @@ storageClasses:
     pytype: lsst.pipe.tasks.sspAuxiliaryFile.SSPAuxiliaryFile
   VisitGeometry:
     pytype: lsst.obs.base.visit_geometry.VisitGeometry
+  ProvenanceQuantumGraph:
+    pytype: lsst.pipe.base.quantum_graph.ProvenanceQuantumGraph
+    parameters:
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -443,3 +443,18 @@ storageClasses:
...
+      - import_mode   # lsst.pipe.base.pipeline_graph.TaskImportMode
+      - quanta  # iterable of uuid.UUID; quanta to read
+      - datasets  # iterable of uuid.UUID; datasets to read
+      - read_init_quanta  # bool, defaults to True; whether to read pre-exec-init info
+    derivedComponents:
+      packages: Packages  # ignores node parameters
+
+      # UUID keys can be quantum or data IDs (whichever is passed in via
+      # parameters).  Nested lists are attempts to run the quantum (last is
+      # most recent).
```
MichelleGower, line 458: Not knowing which kind of UUID key seems like it could cause problems someday, but maybe everything is always taken care of via parameters as mentioned in the comment.

MichelleGower (approved):
One comment.  Merge approved.

python/lsst/daf/butler/configs/storageClasses.yaml
TallJimbo: UUIDs and data IDs would be specified via different parameter keys, so I think this should be fine.

PR #1320, opened 2026-01-09
DM-51681: Remove repeated queries in dimension transfer

dhirving:
Reduce the number of queries required to transfer dimension records using `Butler.transfer_from` or `Butler.transfer_dimension_records_from` from O(N) to O(1).

Also unified `Butler.export()`'s dimension record lookup code with the other `Butler.transfer*` methods, so that it exports the ancillary visit records (`visit_definition`, `visit_system_membership`, etc.) 

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/direct_butler/_direct_butler.py
+++ b/python/lsst/daf/butler/direct_butler/_direct_butler.py
@@ -2032,15 +2032,6 @@ class DirectButler(Butler):  # numpydoc ignore=PR02
                 self.dimensions[original_element.name]  # type: ignore
             )
             if populated_by:
-                # Extract only the data IDs that contain the keys required to
-                # identify the original element's rows, which will then be used
-                # to look up its populated_by children.
-                relevant_data_ids = {
-                    data_id.subset(original_element.required.names)
-                    for data_id in record_mapping.keys()
-                    if data_id.dimensions.required >= original_element.minimal_group.required
-                }
-
                 for element in populated_by:
                     if element not in allowed_elements:
                         continue
@@ -2059,7 +2050,7 @@ class DirectButler(Butler):  # numpydoc ignore=PR02
                         # have to be scanned.
                         continue
 
-                    if relevant_data_ids:
+                    if record_mapping:
                         if not isinstance(source_butler, Butler):
                             raise RuntimeError(
                                 f"Transferring populated_by records like {element.name}"
@@ -2067,7 +2058,7 @@ class DirectButler(Butler):  # numpydoc ignore=PR02
                             )
 
                         with source_butler.query() as query:
-                            records = query.join_data_coordinates(relevant_data_ids).dimension_records(
+                            records = query.join_data_coordinates(record_mapping.keys()).dimension_records(
                                 element.name
                             )
                             for record in records:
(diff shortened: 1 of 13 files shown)


python/lsst/daf/butler/direct_butler/_direct_butler.py
```diff
@@ -2025,16 +2025,22 @@ def _extract_all_dimension_records_from_data_ids(
...
             )
+            if populated_by:
+                # Extract only the data IDs that contain the keys required to
+                # identify the original element's rows, which will then be used
+                # to look up its populated_by children.
+                relevant_data_ids = {
+                    data_id.subset(original_element.required.names)
+                    for data_id in record_mapping.keys()
+                    if data_id.dimensions.required >= original_element.minimal_group.required
+                }
```
dhirving: Could you please take a close look at this bit?  Not 100% sure I got this right "in general", though it seems to work for the current dimension universe.

python/lsst/daf/butler/registry/expand_data_ids.py
```diff
@@ -0,0 +1,93 @@
...
+        cache is not available.
+
+    Returns
+    -------
+    expanded : `list` [ `DataCoordinate` ]
+        List of `DataCoordinate` instances in the same order as the input
+        values.  It is guaranteed that each `DataCoordinate` has
+        ``hasRecords()=True`` and ``hasFull()=True``.
+    """
+    output = list(data_ids)
```
dhirving, line 72: This code is unmodified beyond moving it to another file.

python/lsst/daf/butler/direct_butler/_direct_butler.py
TallJimbo: I think `relevant_data_ids == data_ids` always, because:
- A `DataCoordinate.subset` is actually going to take its argument and turn it into a `DimensionGroup` before it does anything else, and hence you end up with `data_id.subset(original_element.minimal_group)`, which is just  `data_id.subset(data_id.dimensions)`.  That's okay, because it's totally legal for a `DataCoordinate` to only have required values, and `Query.join_data_coordinates` only ever uploads the required values (for simplicity; uploading implied iff available might be faster for some queries, but we don't do that right now).
- `data_id.dimensions` and `original_element` do not change over the course of the loop comprehension, and I believe they are always equal.

TallJimbo (approved):

python/lsst/daf/butler/direct_butler/_direct_butler.py
dhirving: Oh, right.  I lost track of the fact that `_extract_dimension_records_from_data_ids` already splits up the original data IDs into their per-dimension primary keys.

So yeah, it's actually `relevant_data_ids` is always equal to `record_mapping.keys()` (the current dimension's data ID list).  `relevant_data_ids != data_ids` (the top-level function parameter) because `data_ids` can have heterogeneous dimensions.

dhirving:
The failed Jenkins build on linux-x86 was caused by a [known issue with floating point precision](https://rubin-obs.slack.com/archives/C08CANY3A21/p1768339560425169) from the latest `scipy`.

PR #1322, opened 2026-01-19
DM-52738: support code for provenance ingestion

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/registry/sql_registry.py
+++ b/python/lsst/daf/butler/registry/sql_registry.py
@@ -1375,7 +1375,7 @@ class SqlRegistry:
             records = {}
         else:
             records = dict(records)
-        if isinstance(dataId, DataCoordinate) and dataId.hasRecords():
+        if isinstance(dataId, DataCoordinate) and dataId.hasRecords() and not kwargs:
             for element_name in dataId.dimensions.elements:
                 records[element_name] = dataId.records[element_name]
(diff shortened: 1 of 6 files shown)


python/lsst/daf/butler/registry/sql_registry.py
```diff
@@ -1375,7 +1375,7 @@ def expandDataId(
             records = {}
         else:
             records = dict(records)
-        if isinstance(dataId, DataCoordinate) and dataId.hasRecords():
+        if isinstance(dataId, DataCoordinate) and dataId.hasRecords() and not kwargs:
             for element_name in dataId.dimensions.elements:
```
timj, line 1379: I understand you are saying that for `dataId, detector=2` this code was previously resetting detector back to the value in dataId, so is the point here that if any kwarg is given you don't want partial records assigned since there is no efficient way to copy the records you do have and then fill in the extra dimension records from the kwargs? Maybe add a comment saying this is copying records but not if that results in partial record content?

timj (approved):

python/lsst/daf/butler/registry/sql_registry.py
TallJimbo: Yeah, dealing with partial records is just more trouble than it's worth, especially since this non-vectorized expansion is really slow to begin with compared with the vectorized options we have now, so I don't want to micro-optimize it.  I'll add a comment to that effect.

PR #1323, opened 2026-01-22
DM-53899: Migrate documentation to sphinxutils

timj:
Tests fail with pandas 3 for the moment.

## Checklist

- [x] ran Jenkins


python/lsst/daf/butler/dimensions/_coordinate.py
```diff
@@ -35,8 +35,6 @@
 __all__ = (
     "DataCoordinate",
     "DataId",
-    "DataIdKey",
```
timj, line 38: @TallJimbo this is an actual code change that was triggered by me seeing a warning. I can't understand why these were being reexported from here when they are defined in `_records`.

TallJimbo: I'm just guessing, but I could imagine defining those first in `_coordinate.py`, moving them to `_records.py` when they needed to for module-import reasons, and forgetting to update `__all__`.

pyproject.toml
TallJimbo, line 1: FWIW, my guess is that this is a bad interaction between `sphinx-autodoc-typehints` and `numpydoc`, neither of which are explicitly designed to be compatible with each other but do mostly work together.  When I was searching for information on using type annotations with sphinx I stumbled upon an ancient numpydoc issue that's still open (https://github.com/numpy/numpydoc/issues/196).

TallJimbo (approved):

PR #1325, opened 2026-02-02
DM-54024: Add "unsafe_direct" transer mode to Butler.transfer_from

dhirving:
Add a `Butler.transfer_from(transfer="unsafe_direct")` mode that operates like `Butler.ingest(transfer="direct")`.  It records the absolute URI of the files from the source repository in the target datastore with no copying.

## Checklist

- [x] ran Jenkins


timj (approved):
Looks good, although it reminds me that we aren't validating the transfer strings.

PR #1330, opened 2026-02-06
DM-54067: Fix query concurrency errors in Butler server

dhirving:
Fix an issue where cursors were being accessed from multiple threads, and could be closed while they were still being used if cancellation occurred in the query result handler.  This was leading to intermittent segfaults in Butler server, as well as miscellaneous cursor-related exceptions.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/doc/changes/DM-54067.bugfix.md
+++ b/doc/changes/DM-54067.bugfix.md
@@ -1,0 +1,1 @@
+Fixed an intermittent segfault in Butler server caused by a concurrency error related to Postgres cursors. Butler server now requires Python 3.13 or later.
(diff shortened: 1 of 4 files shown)


doc/changes/DM-54067.bugfix.md
```diff
@@ -0,0 +1 @@
+Fixed an intermittent segfault in Butler server caused by a concurrency error related to Postgres cursors.
```
timj: Please add a note about this meaning that you are now required to use python 3.13 to run a server.

timj (approved):
I have not used these asyncio APIs before but after reading the documentation and your comments this code looks good to me.

PR #1338, opened 2026-02-11
DM-54054: Add transfer_datasets_in_place()

dhirving:
Add a new Rubin-private function `transfer_datasets_in_place()`, which will be used by the Prompt Publication Service to replicate database entries between environments while sharing the underlying files on disk.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/tests/utils.py
+++ b/python/lsst/daf/butler/tests/utils.py
@@ -309,23 +309,24 @@ class MetricTestRepo:
 
     @classmethod
     def create_from_butler(
-        cls, butler: Butler, butler_config_file: str, storageClassName: str | None = None
+        cls, butler: Butler, butler_config_file: str | Config, storageClassName: str | None = None
     ) -> MetricTestRepo:
         """Create a MetricTestRepo from an existing Butler instance.
 
         Parameters
         ----------
-        butler : `Butler`
+        butler
             `Butler` instance used for setting up the repository.
-        butler_config_file : `str`
-            Path to the config file used to set up that Butler instance.
-        storageClassName : `bool` or `None`, optional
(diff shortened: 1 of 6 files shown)


python/lsst/daf/butler/tests/utils.py
```diff
@@ -309,23 +309,24 @@ def __init__(
...
     def create_from_butler(
-        cls, butler: Butler, butler_config_file: str, storageClassName: str | None = None
+        cls, butler: Butler, butler_config_file: str | Config, storageClassName: str | None = None
     ) -> MetricTestRepo:
         """Create a MetricTestRepo from an existing Butler instance.
 
         Parameters
         ----------
-        butler : `Butler`
+        butler
```
timj, line 318: In theory we can't drop the types until pipelines.lsst.io migrates to documenteer 2. In practice no-one is going to notice this doc changing.

tests/test_butler.py
```diff
@@ -3406,6 +3407,97 @@ def _remap_transfer_url(path: HttpResourcePath) -> HttpResourcePath:
...
+            )
+            Butler.makeRepo(datastore_root, config=config, forceConfigRoot=False)
+            Butler.makeRepo(other_repo_root, config=config, forceConfigRoot=False)
+            with (
+                Butler(datastore_root, writeable=True) as source_butler,
+                Butler(other_repo_root, writeable=True) as target_butler,
+            ):
+                self._test_transfer_datasets_in_place(source_butler, target_butler)
+
+    def _test_transfer_datasets_in_place(
```
timj, line 3453: Sometimes we use a public method of `assert_x` but it doesn't really make a difference.

timj (approved):

PR #1340, opened 2026-02-20
DM-54083: add storage classes from lsst.images

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -263,7 +263,6 @@ storageClasses:
     pytype: lsst.afw.image.Exposure
     delegate: lsst.obs.base.exposureAssembler.ExposureAssembler
     parameters:
-
       # Load a subimage defined by this box (`lsst.afw.geom.Box2I`).
       - bbox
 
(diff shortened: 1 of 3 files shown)


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -446,15 +445,72 @@ storageClasses:
...
+    converters:
+      lsst.geom.Box2I: lsst.images.Box.from_legacy
+  Transform:
+    pytype: lsst.images.Transform
+  Projection:
+    pytype: lsst.images.Projection
+    converters:
+      # This converter function doesn't actually work, because it takes an
+      # extra argument.  But we need to declare it anyway to give the
+      # legacy formatter and opportunity to do the conversion on read.
```
timj:
```suggestion
      # legacy formatter an opportunity to do the conversion on read.
```

python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -446,15 +445,72 @@ storageClasses:
...
+    derivedComponents:
+      projection: Projection
+      bbox: BoxV2
+    converters:
+      lsst.afw.image.Image: lsst.images.Image.from_legacy
+  MaskV2:
+    pytype: lsst.images.Mask
+    parameters:
+      - bbox
+    derivedComponents:
```
timj, line 486: I think in theory these will need a storage class delegate to handle the in-memory case (such as InMemoryDatasetHandle) but maybe it all works magically if the derived component is a `getattr(thing, "projection")`.

python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -446,15 +445,72 @@ storageClasses:
...
+      mask: MaskV2
+      variance: ImageV2
+    converters:
+      lsst.afw.image.MaskedImage: lsst.images.MaskedImage.from_legacy
+  PointSpreadFunction:
+    pytype: lsst.images.psfs.PointSpreadFunction
+    converters:
+      # This converter function doesn't actually work, because it takes an
+      # extra argument.  But we need to declare it anyway to give the
+      # legacy formatter and opportunity to do the conversion on read.
```
timj:
```suggestion
      # legacy formatter an opportunity to do the conversion on read.
```

timj (approved):

python/lsst/daf/butler/configs/storageClasses.yaml
TallJimbo: Yes, I think `getattr(thing, "projection")` etc works across the board for these.

I do think some `InMemoryDataset` component _conversions_ are probably broken right now (i.e. the ones where the registered converter is a lie but the formatter handles it), but I'm not sure that's something even a storage class delegate can fix.  I also think we might be able to skate through anyway, since the need to convert components should be very rare.

PR #1352, opened 2026-03-30
DM-54472: Add ObservationInfo component to lsst.images Storage Classes

timj:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -470,6 +470,11 @@ storageClasses:
       # extra argument.  But we need to declare it anyway to give the
       # legacy formatter an opportunity to do the conversion on read.
       lsst.afw.image.SkyWcs: lsst.images.Projection.from_legacy
+  ObservationInfo:
+    # In theory we could have a conversion from VisitInfo but there is
+    # no need for that at the moment (and where would those 5 lines of code
+    # live).
+    pytype: astro_metadata_translator.ObservationInfo
   ImageV2:
     pytype: lsst.images.Image
     parameters:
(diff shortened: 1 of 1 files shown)


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -470,13 +470,19 @@ storageClasses:
       # extra argument.  But we need to declare it anyway to give the
       # legacy formatter an opportunity to do the conversion on read.
       lsst.afw.image.SkyWcs: lsst.images.Projection.from_legacy
+  ObservationInfo:
+    # In theory we could have a conversion from VisitInfo but there is
```
timj, line 474: @TallJimbo obs_base has the ObservationInfo to VisitInfo code so maybe it could also have the VisitInfo to ObservationInfo code that is currently in lsst.images. We can assume that obs_base is available if VisitInfo is available. I can do that on a different ticket so I don't keep this one open for too long.

TallJimbo: That makes sense in terms of discoverability, at least.

I think it's a little unclear where `lsst.images` appears in the `lsst_distrib` hierarchy; right now it's a dependency of `pipe_tasks` (since that's where I anticipate adding the tasks to do the legacy-file-rewrite), and it depends on `meas_extensions_piff` and `meas_extensions_psfex` (so we can run the legacy-conversion tests on those PSF types).  The formatter code in `obs_base` that imports `lsst.images` only imports at function scope and is not reflected in the dependencies, but I _think_ it should be okay if we do make `obs_base` depend on `lsst.images`.  I don't think we can push the dependency much further down than that without splitting off some tests into a new package, though.

PR #1364, opened 2026-05-02
DM-54834: Add export and import of datastore records in arrow format

dhirving:
Add a class `DatastoreRecordTable` to represent datastore records as an arrow table, as well as `Datastore` methods `export_table` and `import_table` to work with it.  This also adds "rubin-internal" wrapper functions to expose these datastore methods, intended for use to write parquet files from [butler_transform](https://github.com/lsst-dm/butler_transform/).  

This mirrors the functionality from the existing `Datastore.export_records` and `Datastore.import_records` methods, but with a flat table instead of a convoluted multi-level hierarchical data structure.  The existing data structure embeds too many implementation details and is difficult to work with from external code.

Putting this in `daf_butler` makes it easier to add comprehensive tests for the conversion to arrow and keeps it closer to the other code relying on the same implementation details.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/arrow_utils.py
+++ b/python/lsst/daf/butler/arrow_utils.py
@@ -41,6 +41,7 @@ __all__ = (
 
 import uuid
 from abc import ABC, abstractmethod
+from collections.abc import Callable
 from typing import Any, ClassVar, final
 
 import astropy.time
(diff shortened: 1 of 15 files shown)


python/lsst/daf/butler/arrow_utils.py
```diff
@@ -556,6 +557,63 @@ def as_py(self, **_unused: Any) -> astropy.time.Time:
         return TimeConverter().nsec_to_astropy(self.value.as_py())
 
 
+class ArrowTableUtils:
+    """Utility functions for manipulating `pyarrow.Table` instances."""
```
TallJimbo, line 561: I'm curious why you used a class here vs. free functions.

TallJimbo (approved):

python/lsst/daf/butler/arrow_utils.py
dhirving: No reason in particular.  The code patterns in `daf_butler` often make my brain think it's writing Java or C#, so I think it saw "here is a group of related functions" and the C# thing just came out.

Wouldn't be at all opposed to changing it if you find it objectionable.

TallJimbo: I think I have a slight preference for free functions, because I don't think use the class-as-namespace pattern  much (if at all), but it's very very slight preference.

PR #1366, opened 2026-05-05
DM-54837: Add ssp dimensions

Gerenjie:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/old_dimensions/daf_butler_universe8.yaml
+++ b/python/lsst/daf/butler/configs/old_dimensions/daf_butler_universe8.yaml
@@ -435,40 +435,6 @@ elements:
     storage:
       cls: lsst.daf.butler.registry.dimensions.table.TableDimensionRecordStorage
 
-  ssp_hypothesis_table:
-    doc: >
-      A partitioned set of hypothesized r, r', r'' over which the solar system
-      pipelines (SSP) search in parallel for new solar system objects.
-    keys:
-      - name: name
-        type: string
-        length: 64
-    storage:
-      cls: lsst.daf.butler.registry.dimensions.table.TableDimensionRecordStorage
-
-  ssp_hypothesis_bundle:
-    doc: >
-      A subset of hypothesized r, r', r'' over which a single quanta of
-      HelioLinC searches for new SSOs.
-    keys:
-      - name: id
-        type: int
-    requires:
-      - ssp_hypothesis_table
-    storage:
-      cls: lsst.daf.butler.registry.dimensions.table.TableDimensionRecordStorage
-
-  ssp_balanced_index:
-    doc: >
-      Dummy index for parallelization of linkPurify and other tasks.
-    keys:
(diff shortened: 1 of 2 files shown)


python/lsst/daf/butler/configs/old_dimensions/daf_butler_universe8.yaml
```diff
@@ -0,0 +1,462 @@
+version: 7
```
andy-slac: This needs to be `8`. Also `version` in `dimensions.yaml` needs an increment.

python/lsst/daf/butler/configs/dimensions.yaml
```diff
@@ -435,6 +435,42 @@ elements:
...
+      cls: lsst.daf.butler.registry.dimensions.table.TableDimensionRecordStorage
+
+  ssp_hypothesis_bundle:
+    doc: >
+      A subset of hypothesized r, r', r'' over which a single quanta of
+      HelioLinC searches for new SSOs.
+    keys:
+      - name: id
+        type: int
+        length: 64
```
andy-slac: `length` is not used for integer types. If you want 64-bit integer, `type` should be `long`. Same for `ssp_balanced_index` below.

TallJimbo (approved):
Please address Andy's comments, but I don't have any others.

python/lsst/daf/butler/configs/dimensions.yaml
TallJimbo: :+1: to dropping `length`, but in [this context](https://github.com/lsst/daf_butler/blob/e71ca121f20624d7d4cef26d10c05b38adee75cd/python/lsst/daf/butler/ddl.py#L296) I think 64-bit integer is implied by `int`, and `long` is not allowed (for better or worse, all integers in dimension tables are currently 64-bit).

andy-slac: Sorry, I'm confusing myself (too much felis exposure). Butler only has `int` type, which is 64-bit. Just drop `length`.

Gerenjie: Got it -- sticking to 32-bit integer since I can't imagine we parallelize over billions of hypotheses.

Gerenjie: Excuse me. Somehow didn't load the previous comments. Dropping length :)

python/lsst/daf/butler/configs/dimensions.yaml
```diff
@@ -1,4 +1,4 @@
-version: 7
+version: 9
```
andy-slac: Sorry if I confused you, this has to be `8` too.

andy-slac: And `old_dimensions/daf_butler_universe8.yaml` has to be a copy of this file version 8 with the new tables included.

PR #1367, opened 2026-05-05
DM-54558: Add DetectorV2 storage class and detector VisitImage component.

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -38,6 +38,8 @@ storageClasses:
     pytype: lsst.afw.cameraGeom.Camera
   Detector:
     pytype: lsst.afw.cameraGeom.Detector
+    converter:
+      lsst.images.cameras.Detector: lsst.images.cameras.Detector.to_legacy
   Box2I:
     pytype: lsst.geom.Box2I
   Extent2I:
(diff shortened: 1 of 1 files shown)


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -510,13 +512,21 @@ storageClasses:
       # extra argument.  But we need to declare it anyway to give the
       # legacy formatter an opportunity to do the conversion on read.
       lsst.afw.detection.Psf: lsst.images.psfs.PointSpreadFunction.from_legacy
+  DetectorV2:
+    pytype: lsst.images.cameras.Detector
+    converters:
+      # This converter function doesn't actually work, because it takes an
```
czwa, line 518: I don't understand this comment.  I think I see that this is fine for now, as this only gets used in the `VisitImage` type, so if those don't have an `lsst.afw.cameraGeom.Detector` attached.  How will this work after the transition?

python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -510,13 +512,21 @@ storageClasses:
       # extra argument.  But we need to declare it anyway to give the
       # legacy formatter an opportunity to do the conversion on read.
       lsst.afw.detection.Psf: lsst.images.psfs.PointSpreadFunction.from_legacy
+  DetectorV2:
```
czwa, line 515: Will these eventually become `Detector` and the current one `DetectorV1`?

czwa (approved):
I think both of these are "I don't know what the plan is" questions, and not blockers.

python/lsst/daf/butler/configs/storageClasses.yaml
TallJimbo: The butler code checks to see if there is a converter in this section in order to see whether it should consider two storage classes compatible at all.  But then it gives the formatter hook a chance to do the conversion directly on read before actually using the converter, so if we have the formatter special-case the `.detector` case, we can still make it work.  I haven't actually done that yet, because I'm deferring a lot of storage-class conversion work and testing to [DM-54556](https://rubinobs.atlassian.net/browse/DM-54556).

python/lsst/daf/butler/configs/storageClasses.yaml
TallJimbo: I'm not sure.  We'd have to add some new aliasing functionality to the butler to make that work at the very least, and since these names do get baked into butler database tables, it might not ever be worth it.

PR #1370, opened 2026-05-13
DM-54919: Improve import of DatasetRefs from different universe

andy-slac:
A couple of improvements were added:
- In transfer_from() we try to convert DatasetType to a target universe before registering it. If after conversion its dimensions are different it means that dimensions changed in incompatible way and we raise an exception.
- When deserializing DatasetRefs in ingest_zip() we do similar checks and generate a specific exception when things look incompatible.

There is still cases that fail with pydantic ValidationError, e.g. when metadata is different in different universes.

## Checklist

- [X] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_dataset_ref.py
+++ b/python/lsst/daf/butler/_dataset_ref.py
@@ -63,6 +63,7 @@ from lsst.utils.classes import immutable
 
 from ._config_support import LookupKey
 from ._dataset_type import DatasetType, SerializedDatasetType
+from ._exceptions import InconsistentUniverseError
 from ._named import NamedKeyDict
 from ._uuid import generate_uuidv7
 from .datastore.stored_file_info import StoredDatastoreItemInfo
@@ -1084,7 +1085,7 @@ class SerializedDatasetRefContainerV1(SerializedDatasetRefContainer):
             return []
 
         if universe.namespace != self.universe_namespace:
-            raise RuntimeError(
+            raise InconsistentUniverseError(
                 f"Can not convert to refs in universe {universe.namespace} that were created from "
                 f"universe {self.universe_namespace}"
             )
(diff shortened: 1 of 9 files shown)


python/lsst/daf/butler/_dataset_ref.py
```diff
@@ -1084,24 +1085,44 @@ def to_refs(self, universe: DimensionUniverse) -> list[DatasetRef]:
...
+            # or new dimensions can be aedded to conforming set.
+            for name, dtype in self.dataset_types.items():
+                try:
+                    dataset_type = DatasetType.from_simple(dtype, universe=universe)
+                except KeyError as exc:
+                    raise InconsistentUniverseError(
+                        f"Source dimensions {dtype.dimensions} are not compatible with "
+                        f"target universe dimensions {universe}."
+                    ) from exc
+                if set(dataset_type.dimensions.names) != set(dtype.dimensions or []):
```
TallJimbo: It looks like we usually just put the `required` subset into a `SerializedDatasetType`, so this should be:
```suggestion
                if set(dataset_type.dimensions.required) != set(dtype.dimensions or []):
```
And I think that's good, because I think having the same required dimensions is actually sufficient for a valid dataset transfer, and that covers a pretty important space of dimension universe changes (e.g. the additions of `day_obs` and `group` we did in the past, at least for the vast majority of dataset types that only see those as implied).

python/lsst/daf/butler/_exceptions.py
```diff
@@ -222,6 +223,14 @@ class UnknownButlerUserError(ButlerUserError):
     error_type = "unknown"
 
 
+class InconsistentUniverseError(Exception):
+    """Raised when an imported dataset has a dimension universe that is
+    incompatible with the target butler universe.
+    """
+
+    pass
```
TallJimbo:
```suggestion
```
You actually don't need `pass` if you have a docstring.

TallJimbo (approved):

PR #1378, opened 2026-05-19
Avoid passing pre-formatted strings to time_this (DM-54999)

andy-slac:
That methods expects message string to be a format string for `logging.log`. When we pre-format that message it could end up with a percent sign in it with causes exceptions in `logging`.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_butler_metrics.py
+++ b/python/lsst/daf/butler/_butler_metrics.py
@@ -29,9 +29,9 @@ from __future__ import annotations
 
 __all__ = ["ButlerMetrics"]
 
-from collections.abc import Callable, Iterator
+from collections.abc import Callable, Iterable, Iterator
 from contextlib import contextmanager
-from typing import Concatenate, ParamSpec
+from typing import Any, Concatenate, ParamSpec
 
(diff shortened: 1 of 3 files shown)


timj (approved):
Perfect. That's how the API is meant to be used.

andy-slac:
@timj, I had to add a fix for mypy2, could you check my last commit [acf6d5f](https://github.com/lsst/daf_butler/commit/acf6d5fcbf464a95f96574bb793f1eaacd410358)?

TallJimbo (approved):
I think you could have just removed the `cast` calls, but I guess with the `asserts` we get compatibility with new and old.

andy-slac:
> I think you could have just removed the cast calls, but I guess with the asserts we get compatibility with new and old.

Without assert old mypy complains about `.name` being None when we want it to `str`. It's probably going to be the same with mypy2.

TallJimbo:
I think that suggests that this was triggered by a `click` annotation upgrade, not MyPy (I think the MyPy 2 updates already happened).  But the change is fine either way.

PR #1386, opened 2026-06-12
Remove the ScarletModelData to LsstScarletModelData converter (DM-55150)

fred3m:
A base ScarletModelData cannot be promoted to an LsstScarletModelData: the latter requires a single catalog-wide observed PSF and band list, while a ScarletModelData carries them per blend. The conversion never produced a usable model in the science pipelines, so drop the storage-class `converters` entry (and the now-removed `scarlet_model_to_lsst_scarlet_model` it referenced, deleted in a companion commit on this ticket).

Old ScarletModelData datasets continue to load at their own storage class.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -431,8 +431,6 @@ storageClasses:
     delegate: lsst.meas.extensions.scarlet.io.ScarletModelDelegate
   LsstScarletModelData:
     pytype: lsst.meas.extensions.scarlet.io.LsstScarletModelData
-    converters:
-      lsst.scarlet.lite.io.model_data.ScarletModelData: lsst.meas.extensions.scarlet.io.utils.scarlet_model_to_lsst_scarlet_model
     parameters:
       - blend_id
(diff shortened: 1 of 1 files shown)


timj:
@TallJimbo I am guessing we need to keep this converter around for a while given #1411

TallJimbo:
I'm not sure; #1411 is about the other direction, which in practice is about supporting new pipelines/tasks on old repos.  That's important because most extant repos qualify as "old" - i.e. they have the old storage class for the dataset type where this is used.

So I _think_ this PR is just for supporting old pipelines/tasks (which now just means older software versions) against new repos (`dp2*` at USDF).  And that means it can probably follow a regular-software deprecation cycle.

TallJimbo:
Oh, no, nevermind: we do need this in the long term, because if #1411 makes a new `PipelineTask` write the old storage class in an old repo, a consuming new `PipelineTask` will then need the conversion provided by this PR.  Which, I suppose, might not work given the limitations that motivated this PR.  So it's possible this whole scheme is doomed to failure, and we will have to do something else like a dataset type rename for backwards compatibility.

PR #1387, opened 2026-06-17
DM-55220: add difference kernel and template provenance components to DifferenceImage

TallJimbo:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -562,11 +562,15 @@ storageClasses:
       masked_image: MaskedImageV2
     converters:
       lsst.afw.image.Exposure: lsst.images.VisitImage.from_legacy
+  ConvolutionKernel:
+    pytype: lsst.images.convolution_kernels.ConvolutionKernel
   DifferenceImage:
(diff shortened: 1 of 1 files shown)


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -551,11 +551,16 @@ storageClasses:
...
+  ConvolutionKernel:
+    pytype: lsst.images.convolution_kernels.ConvolutionKernel
   DifferenceImage:
     inheritsFrom: VisitImage
     pytype: lsst.images.DifferenceImage
     converters:
       lsst.afw.image.Exposure: lsst.images.DifferenceImage.from_legacy
+    derivedComponents:
+      kernel: ConvolutionKernel
+      templates: StructuredDataList
```
isullivan, line 574: Do you need to add `backgrounds` here?

isullivan (approved):

python/lsst/daf/butler/configs/storageClasses.yaml
timj: backgrounds is present in VisitImage so doesn't have to be added here.

PR #1391, opened 2026-06-24
DM-55297: Make parquet reading much faster by avoiding parsing the astropy header yaml

erykoff:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/configs/storageClasses.yaml
+++ b/python/lsst/daf/butler/configs/storageClasses.yaml
@@ -148,6 +148,7 @@ storageClasses:
     parameters:
       - columns
       - filters
+      - strip_astropy_meta_yaml
   DataFrameIndex:
     pytype: pandas.Index
     converters:
@@ -174,6 +175,7 @@ storageClasses:
     parameters:
       - columns
       - filters
+      - strip_astropy_meta_yaml
   ArrowColumnList:
     pytype: list
     converters:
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/configs/storageClasses.yaml
```diff
@@ -148,6 +148,7 @@ storageClasses:
     parameters:
       - columns
       - filters
+      - no_strip_astropy_meta_yaml
```
TallJimbo: I think this would be a little easier to reason about if the option was `strip_astropy_meta_yaml` and the default was `True`.

TallJimbo (approved):

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -337,7 +352,9 @@ def arrow_to_numpy(arrow_table: pa.Table) -> np.ndarray | np.ma.MaskedArray:
             has_mask = True
 
     if has_mask:
-        array = np.ma.mrecords.fromarrays(numpy_dict.values(), dtype=dtype)
+        import numpy.ma.mrecords as mrecords
+
+        array = mrecords.fromarrays(list(numpy_dict.values()), dtype=dtype)
```
TallJimbo, line 359: I hate that MyPy isn't smart enough to figure out this pattern; it is if you use a ternary expression, but then of course you can't guard the import.

Just do
```
array: Any
```
(and `from typing import Any`) before the `if` block to shush it.  If you really want to, you can import `mrecords` at module scope (or maybe do that within an `if TYPE_CHECKING` guard) and do
```
array: np.recarray | mrecords.MaskedRecords
```
but that's a lot of verbosity for negligible gain.

PR #1392, opened 2026-07-08
DM-55436: Update parquet code to support pandas3

erykoff:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/formatters/parquet.py
+++ b/python/lsst/daf/butler/formatters/parquet.py
@@ -168,12 +168,7 @@ class ParquetFormatter(FormatterV2):
                     if len(md["column_indexes"]) > 1:
                         has_pandas_multi_index = True
 
-                    if isinstance(md["index_columns"][0], dict):
-                        # Pandas3
-                        index_columns = [col["name"] for col in md["index_columns"]]
-                    else:
-                        # Pandas2
-                        index_columns = md["index_columns"]
+                    index_columns = _get_pandas_index_columns(md)
 
                 if not has_pandas_multi_index:
                     # Ensure uniqueness, keeping order.
@@ -835,13 +830,7 @@ def arrow_schema_to_pandas_index(schema: pa.Schema) -> pd.Index | pd.MultiIndex:
         len_indexes = len(indexes)
 
         if len_indexes > 0:
-            # Get index columns; pandas2 or pandas3 methods.
-            if isinstance(md["index_columns"][0], dict):
-                # Pandas3
-                index_columns = [col["name"] for col in md["index_columns"]]
-            else:
-                # Pandas2
-                index_columns = md["index_columns"]
+            index_columns = _get_pandas_index_columns(md)
     else:
         len_indexes = 0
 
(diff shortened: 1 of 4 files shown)


python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -1582,3 +1627,11 @@ def compute_row_group_size(schema: pa.Schema, target_size: int = TARGET_ROW_GROU
...
 
     return target_size // byte_width
+
+
+def _is_string(t: pa.DataType) -> bool:
+    return pa.types.is_string(t) or pa.types.is_large_string(t) or pa.types.is_string_view(t)
+
+
+def _is_binary(t: pa.DataType) -> bool:
+    return pa.types.is_binary(t) or pa.types.is_large_binary(t) or pa.types.is_binary_view(t)
```
dhirving: Not sure if it comes up in the context of pandas, but it might make sense to also cover fixed size binary via `is_fixed_size_binary` here.

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -795,15 +828,27 @@ def arrow_schema_to_pandas_index(schema: pa.Schema) -> pd.Index | pd.MultiIndex:
...
         len_indexes = len(indexes)
+
+        if len_indexes > 0:
+            # Get index columns; pandas2 or pandas3 methods.
+            if isinstance(md["index_columns"][0], dict):
+                # Pandas3
+                index_columns = [col["name"] for col in md["index_columns"]]
+            else:
+                # Pandas2
+                index_columns = md["index_columns"]
```
dhirving: Might make sense to pull this out to a small utility function, since it's duplicated above.

python/lsst/daf/butler/formatters/parquet.py
```diff
@@ -161,11 +162,19 @@ def _read_parquet(
...
                     md = json.loads(schema.metadata[b"pandas"])
                     if len(md["column_indexes"]) > 1:
                         has_pandas_multi_index = True
 
+                    if isinstance(md["index_columns"][0], dict):
+                        # Pandas3
+                        index_columns = [col["name"] for col in md["index_columns"]]
+                    else:
+                        # Pandas2
+                        index_columns = md["index_columns"]
```
dhirving: Is the implication here that any files written in the future using pandas3 will be incompatible with pipelines versions prior to this PR?  If so let's make sure that @TallJimbo knows about this PR for backporting/etc.

dhirving (approved):
Looks good to me

python/lsst/daf/butler/formatters/parquet.py
erykoff: That is a great question, and it may be the case.  I hate pandas, and I don't want to get into the business about keeping pandas backwards compatibility if they don't.  I know for sure that the strings are going to get completely type-changed if you write pandas3 and read pandas2.  But the index stuff ... might be something that arrow is doing, and I'm going to check.

erykoff: Okay, I've dug a bit more and this was not what I was expecting.  This change in index columns is happening at the *write* stage in pandas3, using the default way of writing parquet files (which we test explicitly just in case something like this happens).

So our *current* parquet reader will have trouble knowing about index columns by name because it isn't aware of this change.  At the same time, the version of pyarrow that we have in rubin-env does understand pandas3 index syntax, so doing a full read does work.  I assume older versions of pyarrow will lose the index.

As we do support ingestion of parquet files not written by the butler, we need to be able to support this.  But I don't think it's a direct backwards compatibility issue within our middleware.  And this PR does now work fully with pandas3-written parquet files.

erykoff:
Okay, I ran the CI with pandas3 and everything passed (including the coverage).  I have reverted that commit.

PR #1397, opened 2026-07-17
DM-55542: Fixes for dimension universe compatibility

timj:
In python mode the model_dump can return set() which can then not be dumped back to json using the DimensionConfig.dump method.

Also unify DatasetType conformance across universes (needed by lsst/pipe_base#572)

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/_dataset_type.py
+++ b/python/lsst/daf/butler/_dataset_type.py
@@ -38,7 +38,7 @@ from typing import TYPE_CHECKING, Any, ClassVar, Self, cast
 from pydantic import BaseModel, StrictBool, StrictStr
 
 from ._config_support import LookupKey
-from ._exceptions import UnknownComponentError
+from ._exceptions import InconsistentUniverseError, UnknownComponentError
 from ._storage_class import StorageClass, StorageClassFactory
 from .dimensions import DimensionGroup
 from .json import from_json_pydantic, to_json_pydantic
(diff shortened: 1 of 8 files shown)


python/lsst/daf/butler/_dataset_type.py
```diff
@@ -343,6 +343,59 @@ class for this dataset type that can convert the python type associated
...
+            raise InconsistentUniverseError(
+                f"Dataset type {self._name!r} has universe {source_universe} with a different "
+                f"namespace than target universe {universe}."
+            )
+        try:
+            dimensions = universe.conform(self._dimensions.names)
+        except KeyError as exc:
+            raise InconsistentUniverseError(
+                f"Dimensions {self._dimensions} of dataset type {self._name!r} do not all exist "
+                f"in target universe {universe}."
```
andy-slac: Maybe also include the version of the source universe to make it more obvious where differences come from?

tests/test_datasets.py
```diff
@@ -1004,6 +1005,123 @@ def test_dataset_provenance(self) -> None:
...
+        self.assertIs(conformed.dimensions.universe, other)
+        self.assertEqual(conformed, dataset_type)
+
+    def test_dataset_type_missing_dimension(self) -> None:
+        """Test that conforming to a universe that lacks one of the dataset
+        type's dimensions fails.
+        """
+        other = self._make_universe(1001, subfilter=None)
+        dataset_type = DatasetType("test", self.universe.conform(["subfilter"]), self.storageClass)
+        with self.assertRaises(InconsistentUniverseError):
```
andy-slac: Could use `assertRaisesRegex` to validate specific exception in addition to its type?

andy-slac (approved):
Looks great.

andy-slac:
But do we want Claude as a main contributor to almost everything?

timj:
> But do we want Claude as a main contributor to almost everything?

Claude did do the code reorganization (it moved your code to a new location) but we don't have a project policy of whether we declare that and LLM was used in commits or not.

PR #1403, opened 2026-07-23
DM-55436-v3: Update parquet formatter to support pandas3

erykoff:
## Checklist

- [ ] ran Jenkins

### DIFF
--- a/doc/changes/DM-55436.feature.md
+++ b/doc/changes/DM-55436.feature.md
@@ -1,0 +1,1 @@
+Update parquet code to support both pandas2 and pandas3.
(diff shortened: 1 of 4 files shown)


dhirving (approved):
Looks good to me, though I want to be clear I understand the compatibility situation.  

It looks like this time around, you are fixing up the indexes on write instead of read.  So am I correct in understanding that files written after the stack switches to pandas3 will be readable by older stacks using pandas2?

erykoff:
The way pandas3 materializes indexes when converting to arrow/parquet is different than pandas2, and this is a known issue.  Any dataframe written with an older/current stack with pandas2 will read in just fine using this reader, whether to arrow or astropy or pandas.  But in the future dataframe written with pandas3 *by our code* should also be readable by an older stack just fine.  (Though we don't guarantee that sort of compatibility afaict).  However, even now, any dataframe written with pandas3 *by somebody outside DM* and then ingested may not be able to see the index column if they ask for the column names.  I don't know how to work around this generally (nor is it easy to test).

PR #1405, opened 2026-07-24
DM-55626: Make the formatter ref consistent with remote and direct butler

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/remote_butler/_get.py
+++ b/python/lsst/daf/butler/remote_butler/_get.py
@@ -20,7 +20,8 @@ from .server_models import FileAuthenticationMode, FileInfoPayload, FileInfoReco
 
 
 def get_dataset_as_python_object(
-    ref: DatasetRef,
+    written_ref: DatasetRef,
+    read_ref: DatasetRef,
     payload: FileInfoPayload,
     *,
     auth: RemoteButlerAuthenticationProvider,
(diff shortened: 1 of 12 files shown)


python/lsst/daf/butler/remote_butler/_get.py
```diff
@@ -31,8 +32,15 @@ def get_dataset_as_python_object(
...
-        Metadata about this artifact.
+    written_ref : `DatasetRef`
+        Metadata about this artifact using the `StorageClass` that the file
+        was written with.  This is the ref given to the `Formatter`, so that
+        it always sees the write `StorageClass` regardless of any read-time
+        override, matching the behavior of `DirectButler`.
+    read_ref : `DatasetRef`
+        Metadata about this artifact using the `StorageClass` that the caller
+        wants returned.  This differs from ``written_ref`` only in its
+        `StorageClass` when a read-time override has been requested.
```
TallJimbo: Can this differ in whether it's a component as well?  Worth documenting either way.

TallJimbo (approved):

python/lsst/daf/butler/remote_butler/_get.py
timj: I don't think component is relevant at all. It flows through the system attached to whatever dataset type is in play. Will add to the docs. The component override storage class is a bit broken as described in DM-55708.

PR #1407, opened 2026-07-25
DM-55631: Ensure streaming queries close on disconnect

timj:
Starlette can cancel its response task while the query output generator is suspended at a yield. In that state the producer may remain blocked while writing the end-of-stream marker to the one-element queue, retaining the user's query-limit slot until garbage collection and producing pending-task or coroutine-reuse errors.

Attach an asynchronous response background task that explicitly closes the output generator when streaming finishes or the client disconnects. Use a separate async helper because an async generator's built-in aclose method is not recognized as an asynchronous callable.

Add an ASGI 2.3 regression test that coordinates a two-page query with an early disconnect, retains the response object, and verifies that the active-query count is restored without relying on garbage collection.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/remote_butler/server/handlers/_query_streaming.py
+++ b/python/lsst/daf/butler/remote_butler/server/handlers/_query_streaming.py
@@ -28,10 +28,11 @@
 from __future__ import annotations
 
 import asyncio
-from collections.abc import AsyncIterator, Callable, Iterator
+from collections.abc import AsyncGenerator, AsyncIterator, Callable, Iterator
 from contextlib import AbstractContextManager
 from typing import Protocol, TypeVar
 
+from fastapi import BackgroundTasks
 from fastapi.responses import StreamingResponse
 
 from lsst.daf.butler.remote_butler.server_models import (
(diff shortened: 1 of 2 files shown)


dhirving:
Your AI is sort of close on the underlying cause but its solution does not work.  Background tasks only run if the stream is NOT canceled, so this does nothing.

The relevant code in Starlette is here: https://github.com/Kludex/starlette/blob/5174d4c8358a6f06aa8056bafd14c2272dab8dd1/starlette/responses.py#L277

In the event that the caller disconnects, an asyncio cancellation exception is raised on task group exit and it never gets to the background tasks.

There are actually a few different issues at play generating similar errors in the logs:
1. If a caller disconnects BEFORE the initial HTTP response gets sent, it never starts iterating the generator at all -- this is benign because no resources are acquired, but does generate some log messages about un-awaited tasks.
2. If a caller disconnects AFTER we start iterating, we run into the problem your AI tried to fix where an unfinished async generator gets cleaned up by the garbage collector.  The garbage collector does trigger a call to aclose(), but it's somewhat nondeterministic.  In practice the cleanup seems to be getting triggered immediately so that's why this hasn't caused any major issues.
3. There may be some issues with the INNER async generators related to async cancellation of task groups  (as described in [PEP-789](https://peps.python.org/pep-0789/).  This should be benign because everything of interest related to resources happens in either the outer generator or synchronous function running in another thread, which gets cancelled when the top-level queue is closed.

As of a month or two ago, FastAPI natively supports the JSON Lines response format, so we can eliminate the StreamingRepsonse entirely and drastically simplify this code: https://fastapi.tiangolo.com/tutorial/stream-json-lines/

We can deal with the task group async cancellation issue using the pattern recommended [here](https://anyio.readthedocs.io/en/stable/cancellation.html#avoiding-cancel-scope-stack-corruption)

dhirving:
Closing since this will be solved a different way

PR #1410, opened 2026-08-07
DM-55708: Attempt to fix component gets with storage class overrides

timj:
## Checklist

- [x] ran Jenkins

### DIFF
--- a/python/lsst/daf/butler/remote_butler/_remote_butler.py
+++ b/python/lsst/daf/butler/remote_butler/_remote_butler.py
@@ -75,6 +75,7 @@ from ._query_results import convert_dataset_ref_results, read_query_results
 from ._ref_utils import (
     apply_storage_class_override,
     get_component_override,
+    make_read_ref,
     normalize_dataset_type_name,
     simplify_dataId,
     split_dataset_type_name,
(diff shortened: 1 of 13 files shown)


python/lsst/daf/butler/remote_butler/_remote_butler.py
```diff
@@ -300,19 +301,17 @@ def get(
         with self._metrics.instrument_get(log=_LOG, msg="Retrieved remote dataset"):
             model = self._get_file_info(datasetRefOrType, dataId, collections, timespan, kwargs)
 
-            written_ref = DatasetRef.from_simple(model.dataset_ref, universe=self.dimensions)
+            server_ref = DatasetRef.from_simple(model.dataset_ref, universe=self.dimensions)
```
TallJimbo: I don't understand how there can be only one server ref.  Isn't the server the source of truth for both the registry dataset type's storage class and the "written" storage class stored in the datastore tables?  Is the former sent to the client at startup or something?

Those being different isn't just an edge case - it's how the `dp2_prep` repo pairs provide backwards- and forwards-compatibility.

TallJimbo (approved):
Looks good as a necessary improvement, though it's also a reminder that the storage class override logic was quite a twisty ride, even before this.  Hopefully after disassembly is gone we can dig in on a refactor to simplify it.

python/lsst/daf/butler/remote_butler/_remote_butler.py
timj: I've clarified things. "server_ref" is referring to the ref with the original dataset type definition (which includes the config override). "written_ref" is the ref with the storage class that was actually used to write it. "read_ref" applies to the "server_ref" (which is now "registry_ref").

PR #1412, opened 2026-08-12
Experimental: Add python 3.15 to build matrix

timj:
## Checklist

- [ ] ran Jenkins


timj:
Since the tests currently use conda we can not test 3.15. This ticket needs to rewrite the tests to use standard python+pip to see how much works. There might be issues with missing binary wheels.

PR #1415, opened 2026-08-14
DM-55824: Use fake remote URI for more efficiency than using moto

timj:
Drop moto for datastore tests and use new faked remote URI that is a local file. This also gives us speed ups in the tests.

Requires lsst/resources#153

## Checklist

- [x] ran Jenkins


dhirving (approved):
Looks great, much simpler and more robust.

PR #1417, opened 2026-08-20
DM-55886: Switch to uv for dependency management

dhirving:
Switched library management to use `uv.lock` instead of multiple `requirements.txt` files.  This makes the dependencies easier to update, makes installation faster for running GitHub Actions and Docker build, and also makes it trivial to set up a local development environment with a known-good set of libraries.

We now use pinned versions of libraries when running tests in GitHub Actions, so PRs will no longer be blocked by "surprise" dependency changes.  There is now an extra unpinned variant of the `build_and_test` GitHub action so we can still passively get warnings when new library versions are incompatible.

The workflow for pinning a development branch of other LSST packages has changed slightly and is documented in the README.

## Checklist

- [x] ran Jenkins


dhirving:
I'll need someone with admin rights to adjust the settings for which checks are required, to point at the new mypy action and point to the pinned=True `build_and_test` action.

timj:
Is your plan to move the mypy code to the shared workflow so that other packages can use it?

dhirving:
> Is your plan to move the mypy code to the shared workflow so that other packages can use it?

I could if you want, but it's literally one line so I'm not sure how much value there is in it.

timj:
> literally one line

isn't it 16 lines of diff? Confused.

dhirving:
> isn't it 16 lines of diff? Confused.

... of interesting executable code :D Otherwise it's just swapping the boilerplate to run the shared action with the boilerplate to install uv.  

If we decide we like this and extend it to other packages we can pull it out then.

mypy.ini
tcjennings, line 1: Might as well fold this mypy config into `pyproject.toml` and git rid of another file.

.github/workflows/do_not_merge.yaml
```diff
@@ -28,10 +28,10 @@ jobs:
 
       - uses: actions/checkout@v7
 
-      - name: Check requirements.txt for branches
+      - name: Check lockfiles for branches
```
tcjennings, line 31: I don't think this works like you'd expect. The lock file is going to resolve any VCS sources to commit hashes, so you won't see the branch names, and `pyproject.toml` could legitimately have matching lines that aren't dependency refs.

dhirving: I think that's incorrect on both counts:
- `uv.lock` includes the branch name in the git URI in the `[manifest]` section, as well as in each individual resolved dependency.  It additionally includes the commit hash, but it records the branch name to keep track of how it resolved the commit hash.
- In the Rubin workflow, `tickets/xxxxx` branches are ephemeral and so nothing merged to main should ever reference them in any way, because they likely won't be there a few minutes later.

timj: It seems like this is easy to demonstrate. Temporarily add a ticket branch to the utils entry and see if this triggers. @tcjennings is  trying to standardize this checker in the reusable workflows so it seems worth confirming.

dhirving: Oh, it was there in its raw form for uv.tool.sources but it turns out it URI-encodes it for overrides-dependencies. So I'll have to adjust the regex slightly.

dhirving: image

.github/dependabot.yml
```diff
@@ -2,13 +2,13 @@ version: 2
 updates:
   # Each ecosystem batches all its version updates into a single pull request.
   # Security updates are not grouped and still open immediately.
-  - package-ecosystem: "github-actions"
```
tcjennings, line 5: Is this intentional? (removing github-actions ecosystem from dependabot)

dhirving: Oh, good catch.  I misunderstood what that was doing.  I should have added the `uv` section separately.

.github/workflows/do_not_merge.yaml
tcjennings: Oh, right, it gets urlencoded in the lock file. It's when you `uv export --format requirements.txt` it trades the vcs ref info for a commit hash.

mypy.ini
dhirving: I think this PR already has enough moving parts without reworking the mypy config too

dhirving:
I had tagged @timj on this review but it seems that @tcjennings has reviewed most of it at this point.  Could one or the other of you approve it so that I can get it merged?  

After it's approved I will also need the "required" checks adjusted so that it is mergeable.

README.md
```diff
@@ -23,3 +23,43 @@ Recipients may choose which of these licenses to use; please see the files gpl-3
...
+```
+to pull in the updated version for testing, where `<package name>` is
+`lsst-utils` or whatever package you are overriding.
+ (You can also just do `uv lock --upgrade` if you don't mind upgrading all the
+ other libraries at the same time.)
+
+You will need to re-do this lock update any time you change the code, so that
+it resolves to the new version.
+
+After the other library's development branch has been merged, reset the
```
timj, line 63: Does this imply we shouldn't be dropping the commit from the history like we do now but instead be committing a new edit with `main` and re-running the lock?

timj (approved):
Thanks for locating the uv issue at https://github.com/astral-sh/uv/issues/11452

I am assuming that there is a desire to make these changes across all the middleware packages and so we would be wanting to have re-usable workflows where possible for mypy and others (the do not merge checker is being worked on I think) even if we have a mypy classic and a mypy uv variant in rubin_workflows.

This is a fundamental change to package dependency tracking so we should likely have discussed this at the weekly team meeting so @andy-slac and @TallJimbo could have signed off on it as well.

dhirving:
> This is a fundamental change to package dependency tracking so we should likely have discussed this at the weekly team meeting

I did bring this up at the last team meeting and nobody objected, although Andy was on vacation I think.

README.md
dhirving: Either way works, as long as the effect is that you make a commit in which you ran `uv lock` while the branch you want to build against was merged to main.

PR #1418, opened 2026-08-21
DM-55885: Speed up tests with cached butler

timj:
## Checklist

- [x] ran Jenkins


test-performance-proposal.md
```diff
@@ -0,0 +1,304 @@
+# Speeding up the daf_butler test suite
```
timj: I will aim to remove this before merging.

python/lsst/daf/butler/tests/_repo_template_cache.py
```diff
@@ -0,0 +1,373 @@
...
+
+    _stats["served"] += 1
+    return written
+
+
+def _make_butler_config(
+    root: ResourcePathExpression,
+    config: Config | str | None,
+    forceConfigRoot: bool,
+    resource: ResourcePath,
```
andy-slac: `resource` does not seem to be used?

python/lsst/daf/butler/tests/_repo_template_cache.py
```diff
@@ -0,0 +1,373 @@
...
+    """Return counts of cache activity, for tests and diagnostics.
+
+    Returns
+    -------
+    stats : `dict` [`str`, `int`]
+        Keys are ``templates`` (databases actually built), ``served``
+        (requests handled by this helper), ``bypassed`` (requests that
+        went straight to `lsst.daf.butler.Butler.makeRepo`), and
+        ``reused_database`` (requests whose database was copied from an
+        earlier identical one).
```
andy-slac: There are more keys in the list below than documented here.

python/lsst/daf/butler/tests/_repo_template_cache.py
```diff
@@ -0,0 +1,373 @@
...
+            path = os.path.join(holder, "butler.yaml")
+            shutil.copyfile(os.path.join(root_uri.ospath, "butler.yaml"), path)
+            _configs[key] = path
+        return written, root_uri
+
+    _stats["reused_config"] += 1
+    root_uri = ResourcePath(root, forceDirectory=True)
+    root_uri.mkdir()
+    destination = os.path.join(root_uri.ospath, "butler.yaml")
+    shutil.copyfile(cached, destination)
```
andy-slac: I hope no one runs `rm -rf` on `$TMPDIR` while the tests are running 🙂

python/lsst/daf/butler/tests/_repo_template_cache.py
```diff
@@ -0,0 +1,373 @@
...
+    key : `str` or `None`
+        A hash of the inputs, or `None` if they cannot be rendered
+        deterministically.
+    """
+    try:
+        rendered = json.dumps(
+            [
+                config.toDict() if isinstance(config, Config) else config,
+                forceConfigRoot,
+                os.environ.get("DAF_BUTLER_CONFIG_PATH"),
```
andy-slac: Just for my education - there are probably cases when `forceConfigRoot` or `DAF_BUTLER_CONFIG_PATH` do not matter, but they do in our tests? Maybe add a comment as from the context it's may not be clear to other people why they are here.

python/lsst/daf/butler/tests/_repo_template_cache.py
```diff
@@ -0,0 +1,373 @@
...
+        Butler._make_repo_registry(written, dimensionConfig=dimensionConfig, root_uri=root_uri)
+        return written
+
+    cached = _databases.get(db_key)
+    if cached is None:
+        _stats["templates"] += 1
+        Butler._make_repo_registry(written, dimensionConfig=dimensionConfig, root_uri=root_uri)
+        holder = tempfile.mkdtemp(prefix="butler-registry-template-")
+        _tmpdirs.append(holder)
+        cached = os.path.join(holder, "gen3.sqlite3")
```
andy-slac: "gen3.sqlite3" should have been a constant defined somewhere. Will be difficult now to switch to gen4 🙂

python/lsst/daf/butler/tests/_repo_template_cache.py
```diff
@@ -0,0 +1,373 @@
...
+    """
+    if config is None:
+        # The default registry is SQLite inside the repository.
+        return True
+    if not isinstance(config, Config):
+        config = Config(config)
+    db = config.get(("registry", "db"))
+    if db is None:
+        return True
+    return str(db).startswith("sqlite")
```
andy-slac, line 164: I'm always worried about substring match (feels like Perl), but it may be OK in this particular case.

python/lsst/daf/butler/_config.py
```diff
@@ -72,17 +72,84 @@
...
+            if type(lhs) is not dict and not isinstance(lhs, MutableMapping):
                 lhs = {}
-            d[k] = _doUpdate(lhs, v)
+            d[k] = _mergeInto(lhs, v)
         else:
             d[k] = v
     return d
 
 
+# Types that are common in configurations and are never mappings. Checking
```
andy-slac, line 111: Could `list` or `tuple` contain mappings?

andy-slac (approved):
Looks good. I have a few minor comments, and I leave it to you to comment on Copilot comments 🙂

python/lsst/daf/butler/_config.py
timj: I think one thing I learned from this is that lists in configs do not get deep copied so nothing cares what they contain.

python/lsst/daf/butler/tests/_repo_template_cache.py
timj: I think in this case it seems unlikely that there will be a different non-sqlite URI that starts with sqlite. Parsing the URI and looking at the scheme seems like overkill.

python/lsst/daf/butler/tests/_repo_template_cache.py
timj: Maybe this should be a pydantic BaseModel so we don't have to list all the keys explicitly.

python/lsst/daf/butler/tests/_repo_template_cache.py
timj: I'm pretty sure that if TMPDIR is wiped out mid test there will be failures for most test suites.

python/lsst/daf/butler/tests/_repo_template_cache.py
timj: We only really use DAF_BUTLER_CONFIG_PATH in tests of DAF_BUTLER_CONFIG_PATH so there aren't many tests using it. I suppose we could simply not use the cached butler for those tests and remove the env var from the cache key. If it's only one test it's probably fine.

PR #1421, opened 2026-08-28
DM-53347: Report malformed dataset type names as user errors

timj:
DatasetTypeExpressionError is now a ButlerUserError, so Butler server forwards it to its clients instead of raising an internal server error. It moves to the top-level exceptions module and is re-exported from lsst.daf.butler.registry, alongside the other user errors that used to live there.  It inherits from the legacy DatasetTypeError so existing handlers still catch it.

Client/server Butler now validates a dataset type name before putting it on the wire.  A name such as "..." previously split into an empty parent name, producing an HTTP 404 rather than a useful message.

## Checklist

- [x] ran Jenkins

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -55,7 +55,7 @@ from lsst.daf.butler import (
     UnknownComponentError,
 )
 from lsst.daf.butler.datastore import DatasetRefURIs
-from lsst.daf.butler.registry import RegistryDefaults
+from lsst.daf.butler.registry import DatasetTypeExpressionError, RegistryDefaults
 from lsst.daf.butler.tests import DatastoreMock, addDatasetType
 from lsst.daf.butler.tests.dict_convertible_model import DictConvertibleModel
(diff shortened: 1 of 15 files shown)


tests/test_server.py
```diff
@@ -198,6 +198,36 @@ def tracking_get(path, **kwargs):
...
+
+        def tracking_get(path, **kwargs):
+            requested_paths.append(path)
+            return original_get(path, **kwargs)
+
+        with patch.object(self.butler._connection, "get", side_effect=tracking_get):
+            with self.assertRaises(DatasetTypeExpressionError):
+                self.butler.get_dataset_type("...")
+
+        self.assertEqual(requested_paths, [])
```
dhirving:
```suggestion
        with patch.object(self.butler._connection, "get") as mock:
            with self.assertRaises(DatasetTypeExpressionError):
                self.butler.get_dataset_type("...")
        mock.assert_not_called()
```

The existing test is fine, but that pattern is also available in the standard library if you want to save a few lines.

dhirving (approved):
Looks good to me

PR #1422, opened 2026-08-28
DM-55989: Only measure coverage in the pinned Python 3.14 CI job

timj:
Before Python 3.14 coverage.py must use its sys.settrace core, which disables bytecode specialization and inlining of Python-to-Python calls. Measured against this test suite that costs around 65% extra runtime, compared with around 10% for the sys.monitoring core used from 3.14 onwards, making the 3.13 job by far the slowest in the matrix.

Coverage is now collected from a single variant, flagged in the matrix entry itself. The junit report is still produced by every job, so test results analytics is unaffected.

## Checklist

- [x] ran Jenkins


dhirving (approved):
Looks good to me

PR #1424, opened 2026-09-03
Include extras by default in dev environment

dhirving:
Include all the extras referenced by commonly-used features and tests in the dev environment by default, so that `uv run` works out of the box without needing to specify extras.

## Checklist

- [ ] ran Jenkins

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -255,6 +255,7 @@ exclude = [
 [dependency-groups]
 dev = [
     "cryptography>=49",
+    "lsst-daf-butler[postgres,remote,server,s3,https]",
     "fsspec>=2026.7.0",
     "matplotlib>=3.11.1",
     "moto>=5.2.2",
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -255,6 +255,7 @@ exclude = [
 [dependency-groups]
 dev = [
     "cryptography>=49",
+    "lsst-daf-butler[postgres,remote,server,s3,https]",
```
timj, line 258: Depending on itself reads a bit odd. I assume this realizes it does not have to install lsst-daf-butler itself from PyPI and will use the local checkout?

dhirving: Correct, when it's resolving packages `lsst-daf-butler` is already in scope as a pointer to the local checkout.  (And you can confirm this by checking the diff for the lock file to see that this didn't cause it to lock some external version of the package.)

timj (approved):

PR #1431, opened 2026-09-15
DM-56097: Improve performance of trash emptying

timj:
We were seeing it was taking 90 minutes to delete 50k datasets. We had assumed it was file system taking the time but in fact it was the emptyTrash. The subquery that was checking that there isn't a another ref pointing at the file was resulting in a full scan of the billion row tables and then a deduplication of billions of rows in python.

Changing the query gives us 9000x speed up.

## Checklist

- [x] ran Jenkins


andy-slac (approved):
Looks good to me. The old and new queries are logically equivalent, but it looks like Postgres tends to choose bad plan for the old query. I think it could also be improved with the old query if you change the order of tables in the last join. But rewritten query is fine too.

